PR #1802, opened 2013-09-11
add scale tool

### LINKED ISSUE #1793: Add scale tool
It could be useful to add a feature scale button other than rotate when a line or area are selected.
https://github.com/openstreetmap/josm/blob/mirror/src/org/openstreetmap/josm/command/ScaleCommand.java

bhack:
implementation of the scale tool. This try to resolve #1793 but it is my first hack on the iD code so please take a look a give me a feedback I've tried to follow rotate implementation.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -116,6 +116,13 @@ en:
       annotation:
         line: Rotated a line.
         area: Rotated an area.
+    scale:
+      title: Scale
+      description: Scale this object around its center point.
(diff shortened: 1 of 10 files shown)


hb0nd:
@tyrasd can this PR be reconsidered? IMHO the reason it was declined is a bit silly, and definitely doesn't align with iD's role as one of the most used OSM editors today.

PR #2004, opened 2013-11-21
Reversing for cardinal directions in relation member roles

mvexel:
This adds smart reversing for north<->south and east<->west in relation member roles when reversing a way.

### DIFF
--- a/js/id/actions/reverse.js
+++ b/js/id/actions/reverse.js
@@ -14,6 +14,8 @@
 
     Relation members:
        role=forward ⟺ role=backward
+         role=north ⟺ role=south
+          role=east ⟺ role=west
 
    In addition, numeric-valued `incline` tags are negated.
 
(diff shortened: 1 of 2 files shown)


simonpoole:
Long done deed I know. But I fail to see why this would make sense, why would the cardinal direction role of a member of a relation change when the (OSM) way direction changes? Contrary to forward/backward the use as a member of route with the role north, south, etc is not relative to the ways drawing direction.

bhousel:
@simonpoole This is probably a case where iD does a thing because JOSM also does it.  We link to the JOSM code in a comment, and my reading of it is that it reversed the "Combined switchers" when reversing relation roles.

https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/corrector/ReverseWayTagCorrector.java#L265

simonpoole:
Even JOSM devs are known to be wrong now and then :-)

bhousel:
I've removed the cardinal direction reversals in 2e6799a
Per https://twitter.com/bhousel/status/1025805098063208448 it looks like JOSM will stop doing this too.  Thanks @simonpoole

HolgerJeromin:
Done https://josm.openstreetmap.de/changeset/14085/josm

Issue #1994, opened 2013-11-20
Radial menu gets in the way


hajo4:
e.g. when you want to select several points that are close to each other,
the radial menu is covering the next point you want to select.

This is a case where the UI makes it really hard to do a simple task.

It would be nice to have an alternative to fiddling around,
like clicking on different places so the menu is not in the way
for the next point you want.

E.g. an alternative menu (straight dropdown, to the left / right of cursor),
or simple switch off the radial menu, and use keyboard-shortcuts.

tmcw:
I vote to close this issue. The radial menu has drawbacks but I don't see a viable alternative solidifying on this issue. At this point in iD's development a change to this interaction would be likely be disruptive to documentation and user habits in a way that outweighs the potential benefit.

Anyone else 👍 to close?

bhousel:
sure we can close it

PR #2145, opened 2014-03-03
Replace radial menu with boxy edit menu

bhousel:
After much frustration with the radial menu, I decided to take a shot at #1994 - Radial Menu gets in the way.

I changed the menu to be a more tightly packed box.  This is nice because it tends to avoid nearby points, and we can add many more operations to the menu without it becoming too unwieldy.

I also added the ability to control which direction the menu pops up.  The calling code in select.js could check for nearby points on the selected entity, and tell the edit menu to avoid them, although I haven't implemented this yet.

editmenu
(In this screenshot I added extra dummy "delete" operations to test column packing code and show what a larger menu would look like).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,8 +96,8 @@ Use `make` to build the translations with the local changes.
 ## Contributing Documentation
 
 Documentation is maintained as a series of [Markdown](http://daringfireball.net/projects/markdown/)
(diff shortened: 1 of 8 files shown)


tmcw:
Can we close this as stale?

bhousel:
> Can we close this as stale?

sure :cry:

Issue #2341, opened 2014-08-27
Status field for life-cycle tagging


talllguy:
**FR** Related to #1922 

![2014-08-27 12_55_00-openstreetmap](https://cloud.githubusercontent.com/assets/1827881/4063830/5277a03a-2e0b-11e4-854a-cc595c38d898.png)

Note the tags in the screenshot above. The yellow hashed road (Campbell Boulevard) has the `highway=construction` and `construction=tertiary`. The white hashed roads have `highway=construction` and `construction=residential`. It is nice that iD (and OSM.org) consider these tags for rendering, but it'd be great if there was a dropdown preset selection for the various types of `construction=*` tags that are applicable to highways.

[Read more on the wiki](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dconstruction)

ignaciolep:
Perhaps a simpler solution could be implemented to solve the most common uses of the construction tag: `building=construction` and `highway=construction`.

Adding an "Status" field to the `building=*` and `highway=*` presets with just two options, "Ready for use" (or "Operational") and "Construction", would be enough. Being "Ready for use" the default option.

By selecting the "Construction" status, iD should decompose the `building=*` or `highway=*` tag in `building=construction` + `construction=*` or `highway=construction` + `construction=*`

Examples:

Residential road under construction

	Status: Ready for use (Default) -> highway=residential
	Status: Construction -> highway=construction + construction=residential

Apartments building under construction

	Status: Ready for use (Default) -> building=apartments
	Status: Construction -> building=construction + construction=apartments

matkoniecz:
> Perhaps a simpler solution could be implemented to solve the most common uses of the construction tag: building=construction and highway=construction.

Or maybe even just highway=construction where it is both widely used and useful.

Issue #1922, opened 2013-10-26
Improve handling of *=construction/construction=*


eliasp:
Currently it is very cumbersome to edit complex construction sites, as every item has to be edited manually from e.g.

    highway=primary

to

    highway=construction
    construction=primary

and the other way round once a construction site has been removed.

I'd like to see this improved by a function which just applies this changes to a node/way with a single click.

bhousel:
#2341 will track development of a field for lifecycle tagging.
In the mean time, 7288806 at least sorts out the styling of construction=* and *=construction, as well as other similar-working tags.

PR #2347, opened 2014-09-03
Construction Presets

bhousel:
Added `construction=*` presets (for #2341, #1922)
Not sure how I feel about this.
Good idea?  Dumb idea?

If it's worth continuing on this, I'll improve the styles too before merging...


bhousel:
withdrawing this, in favor of eventually making a 'status' field instead

Issue #2352, opened 2014-09-09
feature request: prioritize local GPX file layer in drawing order


sbma44:
Decided to get familiar with iD by tracing the upper portion of the Met Branch Trail and adding bicycle:designated to relevant streets. But using a GPX trace to revise tags on existing geometry proved somewhat difficult due to the poor visibility of the trace.

![example of GPX visibility difficulty](http://f.cl.ly/items/2R1r3F3o3Y0a2Q2A1W0X/Image%202014-09-09%20at%203.05.56%20PM.png)

A number of related tickets have raised similar issues:

https://github.com/openstreetmap/iD/issues/2335
https://github.com/openstreetmap/iD/issues/1626
https://github.com/openstreetmap/iD/issues/2083
https://github.com/openstreetmap/iD/issues/1871

Occluding the trace is probably preferable when creating new geometry, so I will certainly understand if this suggestion is rejected. OTOH, if a user has uploaded a trace it will presumably be a central focus of that editing session and consequently might deserve visual priority.

Issue #2335, opened 2014-08-24
GPX-Track in a forest


MyIgel:
The GPX-Track should be on a higher layer then the forest, because it is verry hard to see such a thin line in the green forest
![gpx-track_forest](https://cloud.githubusercontent.com/assets/1448057/4024732/45809430-2bd7-11e4-82b6-328bbdcb382e.png)

Issue #2033, opened 2013-11-28
Building nodes are invisible


pnorman:
Nodes with `building=*` appear to be invisible. See for example http://www.openstreetmap.org/edit?editor=id#map=18/32.30267/36.31946

cc @mikelmaron

Issue #1871, opened 2013-10-06
Add filters for data to make it easier to edit dense multilevel areas


mkv123:
In areas where features are in multiple levels, it would make editing much easier and clearer if there was a way of filtering shown features.

PR #2357, opened 2014-09-16
Feature filtering

bhousel:
This addresses #1299, #1871, #2033 

Adds:
- flexible feature definition
- sidebar to allow user to toggle features (hotkey 'F')
- info bar on bottom to warn user when features are hidden
- supports auto-hiding features (currently auto-hides >100 points, >100 buildings)

Not ready to merge yet, but looking for feedback.
I really like how this is coming together - dense areas are way easier to edit now.

TODO:
- [X] tests
- [X] replace hamburger icon
- [X] adjust text/tooltip language?
- [X] better handling of multipolygons? (e.g. should really hide inner 'holes' when hiding an 'outer' landuse)
- [X] restrict some edits if a feature is hidden? (e.g. moves, deletes?)
- [X] add wireframe mode? (#813)
- [X] make gpx tracks features instead of background? (related #2352, #2335)

Updated:
- [x] @samanpwbb's design suggestion for bottom bar
- [x] exclude selected ids from being hidden
- [x] profile and improve performance where possible

Skipping this item for now, but I definitely want to revisit it later.  Would like to refactor some of the things in features and graph into a more proper graphics pipeline with stat/cull/render passes.
- cull features determined to be "too small" for current zoom (i.e. area is cached as a transient, so we can test quickly before rendering if a feature would only end up being a few sq px)

screenshot 2014-09-16 16 02 48

### DIFF
--- a/js/id/id.js
+++ b/js/id/id.js
@@ -205,7 +205,7 @@ window.iD = function () {
     context.background = function() { return background; };
 
     /* Features */
-    var features = iD.Features(context);
+    var features = iD.Features();
     context.features = function() { return features; };
     context.hasHiddenConnections = function(id) {
         var graph = history.graph(),
--- a/js/id/modes/select.js
+++ b/js/id/modes/select.js
@@ -137,9 +137,16 @@ iD.modes.Select = function(context, selectedIDs) {
             .call(keybinding);
 
         function selectElements() {
-            context.surface()
-                .selectAll(iD.util.entityOrMemberSelector(selectedIDs, context.graph()))
-                .classed('selected', true);
+            var selection = context.surface()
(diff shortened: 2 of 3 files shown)


cgiovando:
Am I missing something or there's still no switch to turn off the data layer altogether? A keyboard shortcut  would be super useful too, without having to go through the feature filtering menu and turn on/off each feature type manually. I looked through other issues - like #2083, #2290 and #1299 for example - but they all have being closed referencing this one.

tyrasd:
@cgiovando No, there is no such option yet. Do you mind opening a new ticket for this?

bhousel:
Hi @cgiovando, there isn't currently a way to turn off all the data layers quickly but my hope was that wireframe mode, with hockey 'w', would meet that need.

Issue #2427, opened 2014-10-27
Edit large features exclusively.


geohacker:
Related to #1520.

Extending from @bhousel's work on [feature-filtering](https://github.com/openstreetmap/iD/pull/2357), can we think about being able to edit large features at lower zoom levels?

Here's one approach -
1. If the way is referenced in the URL, add a checkbox along with other features in the feature-filter and disable everything else other than the feature in question.
2. Make the zoom level for edit configurable based on the feature.

PR #2442, opened 2014-11-11
Support for editing a single focused feature at any zoom level

jfirebaugh:
Continuing from https://github.com/openstreetmap/iD/issues/2427.

Remaining tasks:
- [x] Title/tooltip for focused feature checkbox should be the name of the feature itself (`iD.util.displayName`)
- [x] "Zoom in to edit" button and feature editor should not both appear in the sidebar when zoomed out
- [ ] Shouldn't do expensive intersect/filter of features when < minimum editable zoom
- [x] Should work with relations

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -63,7 +63,8 @@
         "osmAuth": false,
         "sexagesimal": false,
         "toGeoJSON": false,
-        "marked": false
+        "marked": false,
+        "simplify": false
     },
 
     "env": {
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ dist/iD.js: \
 	js/lib/sexagesimal.js \
 	js/lib/togeojson.js \
 	js/lib/marked.js \
+	js/lib/simplify.js \
 	js/id/start.js \
 	js/id/id.js \
 	js/id/services/*.js \
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -317,6 +317,9 @@ en:
     others:
       description: Others
       tooltip: "Everything Else"
+    focused:
+      description: Focused Feature
+      tooltip: "Feature that is focused via the URL"
   area_fill:
     wireframe:
       description: No Fill (Wireframe)
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -387,6 +387,10 @@
         "others": {
             "description": "Others",
             "tooltip": "Everything Else"
+        },
+        "focused": {
+            "description": "Focused Feature",
+            "tooltip": "Feature that is focused via the URL"
         }
     },
     "area_fill": {
--- a/index.html
+++ b/index.html
@@ -33,6 +33,7 @@
         <script src='js/lib/sexagesimal.js'></script>
         <script src='js/lib/togeojson.js'></script>
         <script src='js/lib/marked.js'></script>
+        <script src='js/lib/simplify.js'></script>
 
         <script src='js/id/id.js'></script>
 
--- a/index.html
+++ b/index.html
@@ -95,7 +95,7 @@
         <script src='js/id/ui/contributors.js'></script>
         <script src='js/id/ui/help.js'></script>
         <script src='js/id/ui/geolocate.js'></script>
-        <script src='js/id/ui/notice.js'></script>
+        <script src='js/id/ui/zoom_in_to_edit.js'></script>
         <script src='js/id/ui/flash.js'></script>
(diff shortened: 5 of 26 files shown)


geohacker:
@jfirebaugh Just getting to finish this. I caught up with master. The tests pass in the browser but travis complains about 'simplify' being undefined. Something to do with the way @mourner's [Simplify being exported](https://github.com/mourner/simplify-js/blob/master/simplify.js#L127-L131) I guess? Thoughts?

cc @aaronlidman

mourner:
@geohacker it's JSHint complaining, because you reference `simplify` as a global variable JSHint doesn't know about. You need to add it to JSHint globals config.

geohacker:
Slowly switching from ops to dev mode. 

>  "Zoom in to edit" button and feature editor should not both appear in the sidebar when zoomed out

Can't seem to reproduce this.

>  Shouldn't do expensive intersect/filter of features when < minimum editable zoom

@jfirebaugh - tell me more please? 

>  Should work with relations

Just checked and seem to work. Any specific edge cases?

cc @aaronlidman

jfirebaugh:
@geohacker -- This feature is not working at all for me right now. I load a page with an object ID in the URL, zoom out, and the feature is hidden along with everything else. Am I missing something or does f17fbf0b516efbab0a2317bbadb8ac4aeb338b56 basically disable this feature?

jfirebaugh:
> Shouldn't do expensive intersect/filter of features when < minimum editable zoom

This refers to [this line](https://github.com/openstreetmap/iD/blob/large-object/js/id/renderer/map.js#L100-100) specifically. This deserves benchmarking, but when zoomed very far out I expect this `context.intersects(map.extent())` will be expensive and basically return everything that's currently loaded. Which then will turn around and be filtered down to just the focused feature anyway. If profiling confirms my hunch, "drawing zoomed out focused feature" should be a special case in `drawVector`.

> Should work with relations

Try loading https://localhost:3000/#id=r162069

geohacker:
>  Am I missing something or does f17fbf0 basically disable this feature?

This is true. Sorry. I was a bit confused to see the return of `editOff` while merging from master. I see that we skipped those commits to keep it clean. To add to the confusion, I completely forgot how this feature worked during < minEditableZoom.

Fixing this right away.

jfirebaugh:
Recapping from chat: main tasks to enabling editing for large feature relations are 1) modify the filter for focused feature to include child members and 2) ensure that the simplification step for large-feature rendering gets applied to the child members.

geohacker:
@jfirebaugh I'm [filtering](https://github.com/openstreetmap/iD/commit/0e0bc95a1479678abe402a5acffc0df3339212c1#diff-7339ab11d6ade149f6a5c5c0c270c79eR177) the relation, members and relations that are boundaries when focused. Nodes that make up the relation are being filtered and preserved but ways are not.

Tagging @bhousel here for some help as well -

What we are trying to do - Get relations to work with the "focused" feature, to let people edit large relations. But the member ways doesn't seem to get picked up. Where do you think this should go? Might gathering connections be useful? Thoughts or guidance?

geohacker:
@bhousel ping. Let me know when we can look at this? Thank you!

bhousel:
Hey @geohacker I'm adding a thing called `iD.restrictions` that will help us write a custom rule to prevent users from editing anything but the focused feature. 
WIP on this: 8467433

geohacker:
Ooo that's great @bhousel. Thank you!

bhousel:
@geohacker we need to restart this PR anyway after the iD v2 refactor.  It's a useful idea and we should try to do it.

Per #3681, let's set very clear limits on what people should be allowed to do to large features.   

* I'm ok with someone editing a single selected feature zoomed to around z10 or z11 or so, but probably no lower.
* As long as the features are z11-tile-sized features, no larger.  This is like the size of Manhattan.
* I'm ok with people moving nodes if we limit the distance that the nodes can be moved.


Tile sizes for example:
screenshot 2016-12-26 23 57 48

quincylvania:
Hi everyone, I'm closing this since it's fallen very far behind. I separately implemented the basic functionality of viewing single selected features at any zoom level: see https://github.com/openstreetmap/iD/issues/5001#issuecomment-510131936. It doesn't cover the full scope of this PR—you can edit the tags but not the geometry—but at this point I think we should move on and create new issues for the other components if needed.

Thanks for all your work on this, sorry it never got merged!

PR #2459, opened 2014-11-29
Add drive_through preset to fast_fast, atm, etc..

brianegge:
http://wiki.openstreetmap.org/wiki/Key:drive_through

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -173,6 +173,12 @@ en:
         label: Denotation
       description:
         label: Description
+      drive_through:
+        label: Drive through
+        placeholder: "Yes, no..."
+        options:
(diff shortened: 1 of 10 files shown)


jfirebaugh:
Sorry, I somehow missed the notifications for this PR.

Shouldn't the drive_through field be `"type": "check"`? Doesn't seem like values other than `yes` and `no` are widely used.

brianegge:
Yes, type: check makes sense. I didn't know about that field.

bhousel:
Thanks @brianegge!
Because the fields have moved a bit since this PR started, it couldn't be automatically merged, so I just made the changes in cc9bcba, using a checkbox field instead of combo.

Issue #2475, opened 2014-12-12
Error while saving: version-mismatch / "save" refuses to delete already deleted way


hajo4:
and after that message, refuses to save/upload the rest of the data.

Errormessage here:
"The way with the id 12524388 has already been deleted"

It wouldn't be an issue if this was just a warning / information,
but blocking the upload of all the other data is not acceptable.

It looks like this is a long-standing problem, and still unresolved.
Basically, behaviour like this prevents mappers in the same area
from working efficently, because it forces everyone to save after
every small bit of edit, to prevent loss of data / work done.

Also, the last time I reported a problem like this, I proposed an
"option to exclude just this conflicting data".

Adding to that, it might be a first step for a fix
to split the upload into several parts, and doing the things
least likely to cause conflicts first (e.g. first adding new objects, like houses).

Maybe (automatically) split an upload from the mapper into several changesets,
for each part that was completed sucessfully.

Issue #1053, opened 2013-03-15
Conflict resolution


skorasaurus:
A notification to a user who is about to upload changes that there is newer data on the OSM server would be cool. This would prevent people from doing duplicate work and/or breaking OSM Data. 

For what it's worth, it's in josm (and possibly potlatch), and would be really useful, especially for mapping parties where people may be editing the same areas; and will be more important as iD's user base increases.

PR #2489, opened 2014-12-23
Conflict resolution

bhousel:
This implements Conflict Resolution as discussed on #1053 

* At save time, retrieve latest versions of all modified nodes
* Merge non-conflicting changes to tags
* Merge non-conflicting changes to way nodelist (using diff3)
* ~~If a modified entity has been deleted remotely, restore it (related #2475)~~

Still to do
- [X] some minimal UI for listing the unmergable conflicts for the user
- [X] allow user to download the changeset if automatic merge fails
- [ ] merge non-conflicting changes to relation members for some types? (I think the `route` type would be safe to diff3)

@jfirebaugh, @aaronlidman want to take a look?


js/id/core/graph.js
```diff
@@ -16,10 +16,7 @@ iD.Graph = function(other, mutable) {
 
     this.transients = {};
     this._childNodes = {};
-
-    if (!mutable) {
-        this.freeze();
-    }
+    this.frozen = !mutable;
```
jfirebaugh: What's the rationale for this change?

js/id/actions/merge_remote_changes.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+        else if (target.type === 'relation') {
+            target = mergeRemoteMembers(target);
+        }
+
+        target = mergeRemoteTags(target);
+        return target ? graph.replace(target) : graph;
+    };
+
+    action.withOption = function(opt) {
```
jfirebaugh: Only `safe` is used in practice, right? What's the purpose of `force_local` and `force_remote`?

js/id/core/graph.js
bhousel: It was bothering me that a `Graph` could be initialized with `frozen` left undefined:
```
> g = iD.Graph()
> g.frozen
true
> g2 = iD.Graph(g, true)
> g2.frozen
undefined
```

js/id/actions/merge_remote_changes.js
```diff
@@ -0,0 +1,152 @@
...
+            return target.update({nodes: remote.nodes});
+        }
+
+        var o = base.nodes || [],
+            a = local.nodes || [],
+            b = remote.nodes || [],
+            nodes = [],
+            hunks = Diff3.diff3_merge(a, o, b, true);
+
+        for (var i = 0, imax = hunks.length; i !== imax; i++) {
```
jfirebaugh: [Don't cache array lengths](http://mrale.ph/blog/2014/12/24/array-length-caching.html).

js/id/actions/merge_remote_changes.js
```diff
@@ -0,0 +1,152 @@
...
+            var hunk = hunks[i];
+            if (hunk.ok) {
+                nodes.push.apply(nodes, hunk.ok);
+            }
+            else {
```
jfirebaugh: No line break before `else`.

js/id/actions/merge_remote_changes.js
```diff
@@ -0,0 +1,152 @@
...
+            return target.update({tags: remote.tags});
+        }
+
+        var keys = _.reject(_.union(_.keys(base.tags), _.keys(remote.tags)), ignoreKey),
+            tags = _.clone(target.tags),
+            changed = false,
+            fail = false;
+
+        function ignoreKey(k) {
+            return k.indexOf('tiger:') === 0 || _.contains(iD.data.discarded, k);
```
jfirebaugh: What's the rationale for excluding non-discarded `tiger:*` tags? It looks like the current logic will ignore all remote changes to such keys, which isn't correct.

js/id/actions/merge_remote_changes.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        for (var i = 0, imax = keys.length; i !== imax; i++) {
+            var k = keys[i];
+            if (remote.tags[k] !== base.tags[k]) {  // tag modified remotely..
+                if (target.tags[k] && target.tags[k] !== remote.tags[k]) {
```
jfirebaugh: In the case of a local delete + remote modify tag conflict, this always resolves in favor of the modification. Is this intended?

js/id/actions/merge_remote_changes.js
bhousel: I ignored them here because they are ignored by [Entity#hasInterestingTags](https://github.com/openstreetmap/iD/blob/master/js/id/core/entity.js#L111)

js/id/core/graph.js
bhousel: ok

js/id/actions/merge_remote_changes.js
bhousel: Yes, I did intend to prioritize modifications over deletes.

(This was also why I originally tried to restore locally modified entities when they were deleted on the server.  I still feel like the downside losing a modification is worse than the downside of needing to re-delete something.)

js/id/actions/merge_remote_changes.js
jfirebaugh: Gotcha. `hasInterestingTags` serves a slightly different purpose though; for conflict resolution I think just `iD.data.discarded` is the right set to ignore.

js/id/actions/merge_remote_changes.js
bhousel: I'm probably going to change this to raise the conflict rather than silently restoring the tag, since we are probably moving in the direction of giving the user buttons to resolve the conflict anyway.

I think the most common scenario here goes something like:   UserA sees an address that he knows is wrong and deletes it, meanwhile UserB knows the correct street address and enters it. I'm ok with telling UserA about the conflict.

bhousel:
Here's how the conflict resolution ui looks now.. 

screenshot 2015-01-12 22 35 21

Each item can be expanded for more info and to present options for the user to resolve the conflict:

screenshot 2015-01-12 22 32 42

Choosing "Keep Mine" or "Keep Theirs" will redo the merge and update the graph, and transition out the item from the list.

Alternatively, the "Download Changes" button will make generate a OsmChange xml and open it as a `data:` url in a new tab.

tristen:
@bhousel whoa this is awesome work! I'd like to make a couple recommendations for the UI.

- __Download changes__ is a power user utility and I would hesitate to add this as a primary action.  It might be better focused as linked text in the description of the conflict message.
- I'd remove __Try again__. If a user closes or cancels the modal screen, makes new edits and clicks save that is effectively doing the same thing. This eliminates managing any expectation around what that button does and simplifies the choices in the dialog.
- Downplay the Cancel action and make this a secondary action.
- Surface _who_ made the change.
- If feasible, it would be great to break the diff review into separate steps. Here's a wireframe walkthrough:








This provides a few affordances. Rather than providing a link to a node/relation/way id  (Like, _node w264620391_) the visual feature could be highlighted on the map as a user steps through the different changes. It's a better flag of what's been changed. It also allows us to treat these's ids [as links](http://www.openstreetmap.org/way/264620391) in the description text of each issue which can be helpful. Lastly, focusing on one thing (ie. _surface tag changed_) makes it easier to land on a decision of whether to keep your version or the recent saved version remotely.

bhousel:
Thanks @tristen, I like the suggestions!

One thing that you mentioned:
>Rather than providing a link to a node/relation/way id (Like, node w264620391) the visual feature could be highlighted on the map as a user steps through the different changes. It's a better flag of what's been changed. It also allows us to treat these's ids as links in the description text of each issue which can be helpful.

The main reason I didn't do this initially is because the modal box covers up the map, so highlighting features isn't too helpful.  But I think I can put the conflict resolution UI into the sidebar and make this work. Especially if we are stepping the user through each conflict rather than just presenting them all, it could work well.

tristen:
> But I think I can put the conflict resolution UI into the sidebar and make this work.

This sounds great!

samanpwbb:
@bhousel @tristen I'm going to spend a day or two on tightening up the UI here :+1:

samanpwbb:
If there's a conflict when current user changes a feature  that has been deleted, current user is hit with blocking error message and can't save:

screen shot 2015-02-10 at 1 07 31 pm

 Ideally in this situation user is given the option to either keep their changes to the feature or delete the feature.

bhousel:
continued on #2525

Issue #642, opened 2013-02-05
Copy and paste features


tmcw:
Wishlist! User selects node, C-C C-V duplicates it. Same tags, new osm id.

bhousel:
Copy and Paste of geometries is now merged, and I'm closing this issue.
Copy and Paste of tags (and Repeat Tags), still planned, is being tracked on #839

Issue #815, opened 2013-02-18
point-area merge action is not discoverable


jups23:
Inexperienced users make a point for a village, add data to it. Later, advanced mappers want to make it correct and mark the area of that village and have to transfer data. 
Therefore it would be nice to have a comfortable way to copy that data from the point to the area or be able to turn a point into an area.

polarbearing:
This is bad editing style anyway. 
It would be better to keep the node alive so it can be used prominently, e.g. as the entrance. This preserves the history of the node and makes the changeset better reviewable.

bhousel:
> It would be better to keep the node alive so it can be used prominently, e.g. as the entrance. This preserves the history of the node and makes the changeset better reviewable.

Hey @polarbearing - you are commenting on a very old issue!
Since #3683, iD does now preserve the history of the merged node.

BjornRasmussen:
Isn't this fixed by the new ribbon UI?

quincylvania:
@BjornRasmussen Good find! The Merge operation is now shown in the toolbar when a point and area are selected, so I think that's more discoverable. See #6680.

PR #2498, opened 2015-01-02
Copy paste features

bhousel:
This is for #642 
(these commits were split off from #2489)

Foundation and tests for entity copy/paste is there, need to add:
- [X] behaviors (keybindings)
- [X] handle circular, self-referencing relation graphs

### DIFF
--- a/index.html
+++ b/index.html
@@ -171,6 +171,7 @@
 
         <script src='js/id/behavior.js'></script>
         <script src='js/id/behavior/add_way.js'></script>
+        <script src='js/id/behavior/copy.js'></script>
         <script src='js/id/behavior/drag.js'></script>
         <script src='js/id/behavior/draw.js'></script>
         <script src='js/id/behavior/draw_way.js'></script>
@@ -178,6 +179,7 @@
         <script src='js/id/behavior/hash.js'></script>
         <script src='js/id/behavior/hover.js'></script>
         <script src='js/id/behavior/lasso.js'></script>
+        <script src='js/id/behavior/paste.js'></script>
         <script src='js/id/behavior/select.js'></script>
         <script src='js/id/behavior/tail.js'></script>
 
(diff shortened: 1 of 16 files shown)


bhousel:
[These tests](https://github.com/openstreetmap/iD/blob/90147b23dbf314f975c2cae56b97b9f8508984ac/test/spec/core/relation.js#L94-L121) fail and I have no idea why.  

Entering the same commands into a browser dev console works fine.

bhousel:
This is working really well now!  I have it bound to ⌘-C and ⌘-V for now, just as behaviors (not operations). 

I have to say it is pretty awesome being able to do this in iD:
copy_paste

@jfirebaugh, @aaronlidman check it out when you have a chance..

aaronlidman:
@bhousel This works fantastically.

I only noticed one weird behavior. After pasting items, they don't follow the hidden behavior on zoom as other items of the same type.

hihizoom

bhousel:
> I only noticed one weird behavior. Pasted items don't follow the same hidden behavior on zoom as other items of the same type.

Oh, that's a weird behavior with feature filtering.  We don't hide things that the user just created (i.e. anything with a negative id).

aaronlidman:
Ah, makes sense :+1:

talllguy:
Very exciting feature! Would it let me copy the attributes of a point to a polygon?

bhousel:
> Very exciting feature! Would it let me copy the attributes of a point to a polygon?

Actually iD can already do that..  Just select both the point and polygon and click the Merge button.  The point will be removed and its tags will get merged into the polygon.

talllguy:
@bhousel Bryan thanks for the heads up on that!

bhousel:
@talllguy Yeah it's super useful!  There is an open issue #815 to make point-area merging more discoverable.

nighto:
Excellent, that's the only feature I missed from other editors for the everyday mapping. Thanks!

PR #2505, opened 2015-01-13
Fix German and French translation of 'flush toilet'

georgschoelly:
The original translation for flush were flach (flat) and toilettes publiques (public toilettes).


georgschoelly:
I just found out that translations need to be done through transifex.

PR #2506, opened 2015-01-13
fix German translation for building:detached

rjoberon:
see discussion on http://wiki.openstreetmap.org/wiki/DE:Tag:building%3Ddetached


jfirebaugh:
Hi, thanks for the report. Our translations are handled through [Transifex](https://www.transifex.com/organization/ideditor/dashboard/id-editor) and you can fix this yourself if you sign up for an account there.

PR #2507, opened 2015-01-14
Add amenity=public_bookcase preset

guillaumep:
This commit adds support to amenity=public_bookcase, a recently approved tag:

http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpublic_bookcase

Contributing to iD is actually easy. Thanks for it!


guillaumep:
@bhousel Wow, that was fast, thanks!

PR #2516, opened 2015-01-24
Avoid zorroing connected ways when moving a way

bhousel:
The simple case wasn't too bad:
zorro1

These get tricker, handling endpoints, closed ways, etc
zorro2

zorro3

I'm mostly happy with how it works, may make a few improvements before merging.

### DIFF
--- a/js/id/actions/move.js
+++ b/js/id/actions/move.js
@@ -2,25 +2,46 @@
 // https://github.com/openstreetmap/potlatch2/blob/master/net/systemeD/halcyon/connection/actions/MoveNodeAction.as
 iD.actions.Move = function(moveIds, delta, projection, cache) {
 
-    function addIds(ids, nodes, ways, graph) {
+    function setupCache(graph) {
+        if (!cache) {
+            cache = {};
+        }
+        if (!cache.ok) {
+            cache.moving = {};
+            cache.replacedVertex = {};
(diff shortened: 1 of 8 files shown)


tyrasd:
Wow, this is soo neat. :+1:

jfirebaugh:
This is awesome!

One thing I notice is that the end result is path-dependent. Sometimes even when you move the line back to the original position, the intersection winds up in a different place:

nozorro

jfirebaugh:
This could also avoid un-intersected crossings:

nozorro

bhousel:
I did have a thought that instead of just allowing the non-intersection vertices to move freely, I should make a scale matrix and apply it to them instead.  I think that might make them move more naturally.

So maybe.. 
Intersections = figure out best placement and snap it there
Endpoints = move freely
All other vertices = scale

I did a refactor last night that is doing a better job at picking where the intersection should be (e.g. when 2 ways cross, fixing the zorro on the moving way can introduce one on the fixed way and v.v.).  I'll push soon..

bhousel:
This is done today too!
@jfirebaugh the issues that you noticed before are working much better now.

This was one of those features where I had to try it every possible wrong way before settling on something that works well.

bhousel:
Planning to merge this tomorrow if no further comment.

PR #2525, opened 2015-02-10
Conflict resolution

samanpwbb:
Reroll of https://github.com/bhousel/iD/pull/2

This branch reworks the conflict resolution UI:

- Moved the conflict resolution UI out of a modal and into a sidebar overlay
- Clicking on conflicting features centers +highlights features.
- After resolving one feature, next feature is automatically highlighted.
- Save button is disabled until all conflicts are resolved.

In addition, I couldn't restrain myself from righting some design/css wrongs a younger version of myself made a couple years ago in iD.

@bhousel could you give this a review? Let me know what you think of the design adjustments, and I wouldn't be surprised if the code could be significantly cleaned up, I'm not that experienced with D3.

cc: @jfirebaugh @aaronlidman


bhousel:
Thanks @samanpwbb this looks really awesome!  I think the D3 code looks great..

I'm probably going to change these things: 
- [ ] Thinking of adding some text to show the user how many total they need to click through (e.g. "Conflict 1 of 5") 
- [ ] Make cancel button undo to the pre-save state (right now it leaves all the conflict decisions intact)
- [ ] What do you think about @tristen's suggestion to put the other user's username in the message?  I could do that too..
- [ ] For childnodes of a way/area: If the user moved the way/area, I'm going to present it a single conflict on the parent way/area (instead of asking the user to resolve lots of little conflicts for each moved childnode).

samanpwbb:
Yeah, all those bullets sound good great.

One other suggestion, after letting this sit for a while: would be nice to give user the ability to check what the other user's change is before picking one. This could be done by a) adding a confirm step to each conflict or b) after selecting 'mine'/'theirs', the listing is preserved in the dom and user can return to it to reverse their decision.

bhousel:
> One other suggestion, after letting this sit for a while: would be nice to give user the ability to check what the other user's change is before picking one. This could be done by a) adding a confirm step to each conflict

Yeah I like the idea of letting the user see the difference visually between mine/theirs.. Will add it.

bhousel:
Hey everyone, I think I'm finally happy with how this is working.. Want to take a final code and design pass, and then we can merge it into master and get more people testing it?

Resolving conflicts is now almost *fun*!

conflict resolution

cc @jfirebaugh @samanpwbb @tristen @aaronlidman

tristen:
Awesome!!

jfirebaugh:
This looks great. I think it's time to merge.

:shipit:

danstowell:
This looks lovely. One tiny feedback based purely on your GIF: the link "Download your changes" - I know why it's there, but it looks a bit like it might be the first thing I should do after I've read the instructions in the box. Perhaps it could say "Or download your changes"

samanpwbb:
ooh, you went all the way! :ship: :ship: :ship:

iandees:
What happens when multiple people have modified an object you're trying to modify? Or more importantly, what happens if there is more than one edit?

iandees:
(PS this is awesome)

bhousel:
> What happens when multiple people have modified an object you're trying to modify? Or more importantly, what happens if there is more than one edit?

I can summarize how it works:

As soon as the user decides to save, iD will fetch the latest version from the server before we assemble the changeset.  If the server returns a newer version of an entity, we first try to merge in non-conflicting changes (e.g. if I changed the address and you changed the phone number, these changes will be merged safely and automatically). 

If there is a conflict that can't be merged automatically (e.g. you and I both change the same tag, or I move something and you delete it), then we present this interface to the user and let them pick which version they want to keep.  The default choice is to keep the other user's version.

After the user has reviewed all of the conflicts and presses "Save" it actually restarts the process.  So if somebody did sneak in another change while the user is making their decisions, we would catch it and the user would see the warning again and need to make another choice.

iD won't actually finalize the changeset until it's in a state where all the locally modified entities are version-consistent with the server entities.

iandees:
Great summary! I was mostly curious what the UI looks like when multiple revisions (potentially from different users) exist between the "edited" version and the "currently on server" version.

i.e. I move version 2 of an existing node. Then you come along and move it, save to the server, and move it again and save to the server. Then I try to save my changes. The server reports the node at version 4.

It sounds like your flow handles that just fine, but I was curious how you handled the UI/copy peculiarities in that case.

bhousel:
> i.e. I move version 2 of an existing node. Then you come along and move it, save to the server, and move it again and save to the server. Then I try to save my changes. The server reports the node at version 4.

Yes, in that case the UI will ask you to choose between: 
"Keep mine" - your moved v2
"Use theirs" - current v4

iandees:
Very cool! I'm looking forward to giving this a try.

PR #2526, opened 2015-02-12
Don't delete ways when part of route/boundary relations

### LINKED ISSUE #1461: Warn before deleting a member of a relation [$15]
User can delete a relation member without warning.
Either deletion should not be allowed at all, or an explicit message should pop.

systemed:
Simple patch to prevent ways being unthinkingly deleted from route or boundary relations.

The user can still delete a way which is part of such a relation by manually removing the way from the relation first, then deleting the way.

Fixes #1461.

### DIFF
--- a/js/id/actions/delete_way.js
+++ b/js/id/actions/delete_way.js
@@ -32,18 +32,16 @@ iD.actions.DeleteWay = function(wayId) {
     };
 
     action.disabled = function(graph) {
-        var way = graph.entity(wayId);
-        var reltypes = ['route','boundary','multipolygon'];
-        var required_roles = { 'multipolygon': 'outer' };
         var disabled = false;
-        graph.parentRelations(way)
-            .forEach(function(parent) {
-                if (reltypes.indexOf(parent.tags.type)>-1) {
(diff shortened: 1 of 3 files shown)


tmcw:
LGTM: it does swing the dial from overly lax to pretty strict, but there's a clear way to act on the disabled-message which is good. @jfirebaugh want to review for merge?

KathleenLD:
:dancers:

jfirebaugh:
Looks good to me. Based on the feedback in #1461 I suggest we expand this to at least the _outer_ ways of multipolygon relations.

I'd also like to put in a modifier key escape hatch for experienced editors. E.g. hold down alt or shift to bypass the prohibition. If you want to tackle that let me know, otherwise I can merge this and add it myself later.

systemed:
No problem - I've added a check for multipolygon outers. I'll let you do the modifier key bit!

jfirebaugh:
I really confused GitHub with some force pushing, but this was merged.

systemed:
Excellent, thank you!

KathleenLD:
Nice work, guys!

brycenesbitt:
Thank you for this, it was much needed, and will save significant future hassle.

PR #2549, opened 2015-03-08
Add/Prioritize network=* tag in public transport presets

gileri:
There are multiple companies involved in public transport (PT), but the one which the user faces (and mostly cares about) is the *network*, and the operator is at best not clearly visible to users.

However, operator=* is often used to store the PT network name instead of network=*.
I suspect it might be caused by the ordering (and sometimes lack) of the network=* tag in presets, which this PR tries to fix.

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -3781,6 +3781,7 @@
     "highway/bus_stop": {
         "icon": "bus",
         "fields": [
+            "network",
             "operator",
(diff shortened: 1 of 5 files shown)


bhousel:
Looks good, thank you!

gileri:
Wow, that was fast ! Thank you for accepting my PR :)

Issue #2515, opened 2015-01-24
Display location in the footer of interface


tristen:
With tools like http://osmlab.github.io/to-fix when you select an error to edit it could any location in the world. It would be nice to get a sense of its geographic spot. ie:

![screen shot 2015-01-23 at 11 44 43 pm](https://cloud.githubusercontent.com/assets/61150/5886154/2c626640-a35a-11e4-8e79-bd91eaab6a50.png)

tas50:
+1 this would be incredibly useful

bhousel:
We're planning to add a small "map-in-map" popup pretty soon to help users better know where they are editing.  (Think like a little zoomed out map with a square representing the current viewport bbox).  I think I'll find a way to include the location text in this feature.

bhousel:
see #2554

1ec5:
A breadcrumb would be useful on the main map as well: openstreetmap/openstreetmap-website#848.

bhousel:
I have been using this gist to test out nominatim reverse geocode:
https://gist.github.com/bhousel/05be464d7f53c95c4eab

Even `zoom=10`, which I would think should return 'cities', nominatim is still returning hamlets, and the hamlet data in the USA is [pretty bad](http://www.openstreetmap.org/user/samely/diary/34541). 

If I go out to `zoom=9`, it returns county/region data, which is less useful.

pnorman:
I think it would be useful to go ahead, even if there is bad data in a region which makes it less useful.

bhousel:
Coming soon - A location pane that toggles with <kbd>⌘L</kbd>

<img width="776" alt="screenshot 2017-06-29 01 29 10" src="https://user-images.githubusercontent.com/38784/27672664-687412de-5c6a-11e7-825d-1b3083af6fe9.png">

PR #2554, opened 2015-03-12
Map-in-Map overview box

bhousel:
* press 'M' to toggle
* shows current zoom - 6
* with locator overlay and bounding box

Pretty sweet!

map-in-map3

@aaronlidman

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -2107,7 +2107,7 @@ img.wiki-image {
 /* Map-In-Map
 ------------------------------------------------------- */
 
-.map-in-map-wrap {
+.map-in-map {
     position: absolute;
     overflow: hidden;
     top: 60px;
@@ -2117,13 +2117,25 @@ img.wiki-image {
     background: #000;
     border: #aaa 1px solid;
     box-shadow: 0 0 2em black;
-    pointer-events: none;
 }
 
-.map-in-map-wrap svg {
+.map-in-map-tiles {
+    transform-origin:0 0;
+    -ms-transform-origin:0 0;
+    -webkit-transform-origin:0 0;
+    -moz-transform-origin:0 0;
+    -o-transform-origin:0 0;
+    -moz-user-select: none;
+    -webkit-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+}
+
+.map-in-map-svg {
     position: relative;
     overflow: hidden;
     height: 100%;
+    width: 100%;
 }
 
 .map-in-map-bbox {
--- a/css/app.css
+++ b/css/app.css
@@ -609,7 +609,7 @@ a:hover .icon.out-link   { background-position: -500px -14px;}
     top:0;
     right:0;
     height:60px;
-    z-index: 1;
+    z-index: 9;
     min-width: 768px;
 }
 
@@ -2113,7 +2113,7 @@ img.wiki-image {
     top: 60px;
     width: 200px;
     height: 150px;
-    z-index: 9;
+    z-index: 5;
     background: #000;
     border: #aaa 1px solid;
     box-shadow: 0 0 2em black;
--- a/css/app.css
+++ b/css/app.css
@@ -2145,6 +2145,10 @@ img.wiki-image {
     shape-rendering: crispEdges;
 }
 
+.map-in-map-bbox.thick {
+    stroke-width: 5;
+}
+
 /* About Section
 ------------------------------------------------------- */
 
--- a/css/app.css
+++ b/css/app.css
@@ -683,7 +683,7 @@ a:hover .icon.out-link   { background-position: -500px -14px;}
     float: left;
     height: 100%;
     overflow: hidden;
-    z-index: 2;
+    z-index: 10;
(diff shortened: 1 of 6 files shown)


aaronlidman:
Very cool. @bhousel what is left to do before this is in a state to merge?

Also are all the d3 changes you made in and final? I'm having trouble getting this working locally. This is as far as loading gets me on map-in-map branch and master:

screen shot 2015-03-12 at 10 43 59 am

`Cannot read property 'matches' of undefined` at https://github.com/openstreetmap/iD/blob/master/js/id/renderer/features.js#L308

bhousel:
> what is left to do before this is in a state to merge?

I want to have the mini-map capture some mouse events, rather than passing them all through to the main map.  Right now you can actually edit stuff underneath the mini-map, which is kind of hack.  I'm going to see how complicated it is would be to implement some reasonable zoom/pan behavior.

I guess I can also try to do a nominatum lookup for some "Where am I" text (#2515)

Anything else you can think of?



> Also are all the d3 changes you made in and final? I'm having trouble getting this working locally. This is as far as loading gets me on map-in-map branch and master:

Ah, thanks for finding that, it's something I broke in 3308b55, not D3's fault.   I'll fix it in master.  
(I can see how it could happen if a way has a parent relation that doesn't have any matched features.  Like a site relation or something).

aaronlidman:
Sounds great, looking forward to it.

bhousel:
This is in pretty good shape now.. Panning the mini-map will recenter the main map, which is nice.

I did experiment with using nominatim to reverse geocode a description of where the map is centered, but I wasn't happy with the results.  Most of the time it would return a "hamlet" that is either too obscure or total fiction.  I had to go all the way out to the state level just to get it to not do that.

So I think we should just stick to the locator overlay tiles that we're displaying - they are better anyway.

aaronlidman:
This works great, perfect.

Should the left margin match the buttons in the top bar?
screen shot 2015-03-15 at 4 29 11 pm

@bhousel This just needs tests and it's good to merge?

bhousel:
> Should the left margin match the buttons in the top bar?

I guess we could?  My thought was actually to just set aside this region where extra UI widgets/plugins could render themselves.  Want to loop @samanpwbb in tomorrow and see what he thinks?   The Mapillary photos render themselves to the bottom right corner.. Maybe there is a coherent way we could organize these widgety things. 

> This just needs tests and it's good to merge?

We don't really have test coverage of any of the other UI classes.. I'm not sure how to do it.

Also, I just realized - I think I need to pick a different hotkey.  'M' is used for move too.

css/app.css
```diff
@@ -683,7 +683,7 @@ a:hover .icon.out-link   { background-position: -500px -14px;}
     float: left;
     height: 100%;
     overflow: hidden;
-    z-index: 2;
+    z-index: 10;
```
samanpwbb: This change messes up z-index order:
screen shot 2015-03-15 at 10 17 48 pm

samanpwbb:
Since this is a navigational aid, what about adding it as a fourth item in the navigation toolbar below 'show my location':

screen_shot_2015-03-15_at_10_41_14_pm

I'd also suggest moving the mapillary overlay to the top right, with a higher z-index than the minimap, so it occupies the same space as the minimap when active. This will open up more workspace if user is using both minimap and mapillary.

bhousel:
I don't have really strong feelings either way, but I was kind of thinking longer term of evolving it into something like this:

plugin all the things

Where plugins might include minimap, Mapillary, distance measurement tool, task manager integration, etc etc.  It looks kind of like Photoshop I guess.

Then we wouldn't add any more buttons to iD, except one for installing and managing the user's plugins/widgets/addons.  With the goal eventually pulling more of these things *out* of iD.

(yes I have plugins on the brain, and I'm probably getting ahead of myself)

samanpwbb:
> Where plugins might include minimap, Mapillary, distance measurement tool, task manager integration, etc etc. It looks kind of like Photoshop I guess.

> Then we wouldn't add any more buttons to iD, except one for installing and managing the user's plugins/widgets/addons. With the goal eventually pulling more of these things out of iD.

Moving towards a plugin system will definitely be great for keeping complexity at bay for beginners but enabling features for power users. I _am_ afraid the design sketch you posted would be a too inflexible though. Different plugins will need design solutions that take different forms. I guess we can cross that bridge when we need to.

Anyway, if this work is going to be refactored soon, probably fine to ship the minimap without too much fuss over the specifics until then.

aaronlidman:
> We don't really have test coverage of any of the other UI classes

Ok, didn't know that.

This works really well, seems solid to me. If you're happy with it I say merge away. :+1: 

cc @jfirebaugh

bhousel:
Thanks! @samanpwbb I agree that plugins will be need ability to hook into different parts of iD and have very different UI depending on what they do. Definitely needs more thought.

One final thing I added today was making it so that you can zoom/unzoom with the mousewheel, to control how far zoomed out the minimap is.  Very useful.

Issue #2449, opened 2014-11-20
Add locale and host info to changeset tags


ImreSamu:
iD Editor only saving the version information, like: created_by=iD 1.6.2
JOSM save the locale settings, like : created_by=JOSM/1.5 (7724 hu)

for QA - I would like to propose to add 
* 'locale' ( like: en,de,it,es,hu ) and
* hosting 'website' url ( like  strava.github.io/iD, openstreetmap.us/iD/master , ... )

proposal example:  
* created_by=iD 1.6.2 (hu strava.github.io/iD )
* created_by=iD 1.6.2 (de openstreetmap.us/iD/master) 
* created_by=iD 1.6.2 (en openstreetmap.us/iD/release)

business cases:

* I would like to monitor the using of the locale=hungarian  iD Editor:

* If I found a "bad" translating error -on "iD Editor" - and I would like analize the problematic edits.
  but I am interested only the 'hu' localisation changesets.
  
* If NOW - somebody use very old iD Editor version ( like  1.0.0 ) and I would like to find the hosting website.    like: created_by=iD 1.5.4 (hu mapillary.github.io/iD)
so we can alert the website owner to update the iD Editor code base ( and the translations ) to the newer version.

this proposal is related to translating problems :  https://github.com/openstreetmap/iD/issues/2448

ImreSamu:
:(
http://www.openstreetmap.org/user/aseerel4c26/diary/34899

PR #2559, opened 2015-03-20
Add basic browser and platform into to changeset tags

bhousel:
This adds some simple info to the changeset tags about browser and platform used.
Reporting this info is useful for some things (see #2449).

* Browser version is detected as major.minor only (no patch included).
* Platform is very simple (Windows/Macintosh/Linux) with no version.
* These fields are namespaced with 'iD:' to not conflict anything

Looks like this in practice:

screenshot 2015-03-20 14 27 37

cc @pnorman

### DIFF
--- a/js/id/id.js
+++ b/js/id/id.js
@@ -330,17 +330,39 @@ iD.version = '1.7.0';
     var detected = {};
 
     var ua = navigator.userAgent,
-        msie = new RegExp('MSIE ([0-9]{1,}[\\.0-9]{0,})');
+        m = null;
 
-    if (msie.exec(ua) !== null) {
-        var rv = parseFloat(RegExp.$1);
-        detected.support = !(rv && rv < 9);
+    m = ua.match(/Trident\/.*rv:([0-9]{1,}[\.0-9]{0,})/i);   // IE11
+    if (m !== null) {
+        detected.browser = 'msie';
+        detected.version = m[1];
+    } else {
+        m = ua.match(/(opera|chrome|safari|firefox|msie)\/?\s*(\.?\d+(\.\d+)*)/i);
+        if (m !== null) {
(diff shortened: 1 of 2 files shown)


pnorman:
If the tags aren't iD specific and are clear in their meaning, I think avoiding namespaces are best.

P2 can run on multiple sites, and all editors have a locale. Browser is less relevant for P2, but any future HTML or JavaScript editors will probably find it relevant.

bhousel:
hmm.. `platform` is already a key..  Maybe change to `operating_system` and remove the namespaces?

pnorman:
> hmm.. `platform` is already a key

Used in changeset tags?

ImreSamu:
statistics: https://gist.github.com/ImreSamu/0fc3b8015e6dd16595c5

js/id/id.js
```diff
@@ -350,11 +372,22 @@ iD.version = '1.7.0';
...
+    else if (nav('Mac')) {
+        detected.os = 'mac';
+        detected.platform = 'Macintosh';
+    }
+    else if (nav('X11') || nav('Linux')) {
+        detected.os = 'linux';
+        detected.platform = 'Linux';
+    }
+    else {
+        detected.os = 'win';
```
tilmanb: This does seem odd. If we cant detect anything, we set the detected os to windows? Why?

bhousel: `detected.os` is used for only one thing in iD, and that is to setup the keyboard shortcuts in [cmd.js](https://github.com/openstreetmap/iD/blob/master/js/id/ui/cmd.js).  

I suppose `win` was just picked as a conservative default for the keyboard shortcut setup.  In reality, if the code runs here, the user is probably on a device without a keyboard - like a game console or an Android phone.

PR #2561, opened 2015-03-23
Fix max scaleExtent for custom overlays. Fix #2558

### LINKED ISSUE #2558: scaleExtent of custom overlay not respected

olafveerman:
This should fix the issue where custom overlays are not obeying to the max scaleExtent.


olafveerman:
While this solves the issue with the custom overlay, it doesn't show the locator overlay on all zoom levels. Closing and opening another PR that should address it well.

PR #2562, opened 2015-03-24
Display overlays on correct zoom level. Fix #2558

### LINKED ISSUE #2558: scaleExtent of custom overlay not respected

olafveerman:
Add a test for the Locator Overlay and one for the other overlays.


jfirebaugh:
This isn't the correct fix either as it disables overzooming for all imagery sources. Discussion continuing on #2558.

Issue #2504, opened 2015-01-12
Introductory help text


pnorman:
The introductory help text (help on the side-bar) states

```markdown
Edits that you make on this map will be visible to everyone who uses
OpenStreetMap. In order to make an edit, you'll need a
[free OpenStreetMap account](https://www.openstreetmap.org/user/new).
```

Although technically correct, when viewed on OpenStreetMap.org this text can only be seen by someone who already has an account, so directing them to /user/new is not helpful.

A second more minor problem is that it directs to www.openstreetmap.org, and I was using the dev site, so it was the wrong link anyways.

harkirat96:
Kindly check the pull request
https://github.com/openstreetmap/iD/pull/2563
Redirected the person directly to login rather than new user

PR #2563, opened 2015-03-24
Fixes issue #2504

harkirat96:
Instead of redirecting to new user login , it redirects to login page


jfirebaugh:
Hi @harkirat96, thanks for contributing.

The correct place to make this change is [data/core.yaml](https://github.com/openstreetmap/iD/blob/master/data/core.yaml). `dist/locales/en.json` gets overwritten by the build script.

Also, you've checked in an editor backup file `dist/locales/en.json~`. Please add a rule to your global .gitignore so that these files are ignored.

harkirat96:
@jfirebaugh:Made changes in core.yaml.Ignored en.json

jfirebaugh:
There are even more editor backup files in this PR now. Can you try again with a clean PR?

ashish0kumar:
> Instead of redirecting to new user login , it redirects to login page

Great things started here :)

PR #2567, opened 2015-03-24
fixes issue #2504

### LINKED ISSUE #2504: Introductory help text

harkirat96:
Changes core.yaml to fix #2504

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -400,8 +400,8 @@ en:
       better for everyone.
 
       Edits that you make on this map will be visible to everyone who uses
-      OpenStreetMap. In order to make an edit, you'll need a
-      [free OpenStreetMap account](https://www.openstreetmap.org/user/new).
+      OpenStreetMap. In order to make an edit, you'll need to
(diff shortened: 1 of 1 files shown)


harkirat96:
Apologies for the previous pull request.

jfirebaugh:
Thanks!

pnorman:
I'm not sure this really fixes the issue. Anyone who sees this text will have already logged in.

harkirat96:
If it is so then this text is irrelevant.We could delete this part from the file.

PR #2568, opened 2015-03-25
Allow overzoom to be set on overlays. Fix #2558

### LINKED ISSUE #2558: scaleExtent of custom overlay not respected

olafveerman:
The default behaviour is to overzoom overlays past their max scaleExtent.
When 'overzoom':false is defined in imagery.json, the layer will disappear on the max scaleExtent.

### DIFF
--- a/data/imagery.json
+++ b/data/imagery.json
@@ -10515,6 +10515,7 @@
             0,
             16
         ],
+        "overzoom": false,
         "terms_url": "http://www.mapbox.com/about/maps/",
         "terms_text": "Terms & Feedback",
         "default": true,
--- a/js/id/renderer/background_source.js
+++ b/js/id/renderer/background_source.js
@@ -4,6 +4,7 @@ iD.BackgroundSource = function(data) {
         name = source.name;
 
     source.scaleExtent = data.scaleExtent || [0, 20];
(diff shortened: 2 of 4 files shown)


data/imagery.json
```diff
@@ -10515,6 +10515,7 @@
             0,
             16
         ],
+        "overzoom": false,
```
jfirebaugh: This will get overwritten by `make imagery`. Can you add some code to `data/update_imagery.js` that adds `imagery: false` to the entry for the locator overlay as a special case?

js/id/renderer/background_source.js
```diff
@@ -4,6 +4,7 @@ iD.BackgroundSource = function(data) {
         name = source.name;
 
     source.scaleExtent = data.scaleExtent || [0, 20];
+    source.overzoom = data.overzoom !== false ? true : false;
```
jfirebaugh: No need for a ternary here; `data.overzoom !== false` is already a boolean value.

olafveerman:
Above commit should address the issues @jfirebaugh

jfirebaugh:
Looks good, thanks! Committed as 89b354ddbaf896b408d65c0a95c61ae6316882b4.

Issue #2519, opened 2015-02-03
Add distance/area measurement tool


mobina:
I am interested to add 'measure area' for area digitized.. Any clue to implement this tool ?

bhousel:
Yeah this is a feature that would be really useful, and we do want to build it.. 
Can you describe how you might use the tool?  What kinds of features are you measuring?

pnorman:
I recall this having come up in the past as being out of scope for iD, but cannot find the issue.

jfirebaugh:
See #2039 -- I'm not convinced that this is a common enough need to justify permanent UI space in iD. At this point the bar is very high for adding features with visible UI. Every new feature adds complexity, and reduces the relative prominence of other, more commonly used, features.

mobina:
To measure the length of the road or area of the building..

pnorman:
> To measure the length of the road or area of the building..

I think you need to explain why in the context of a **mapping** workflow. Collecting distance measurements is not conventional. I have done it with a laser rangefinder, but it's a very specialized task.

hajo4:
E.g. estimating the capacity of a car-parking-space from length or area.

pnorman:
That sounds like a data consuming task, not a mapping task. If you're estimating the capacity, the parking lot is already mapped, and it's not like you should be putting an estimate into the tags.

hajo4:
While mapping, after marking the area of some carparking, I would like to also fill in the capacity.
But I usually don't bother to count more than 5-10 cars/spaces.
So, to make a reasonable estimate, it would be nice to have a tool 
for measuring/displaying length/width/area.

hajo4:
Regarding UI:
This could use a shortcut (say "i" for "info"), but even that could be left out.
Options to consider for displaying of the values:
* Area at the top and to the right of the "save-counter"
* Tooltip - say, after hovering 5 seconds over selected object --> no shortcut needed
* new "Info"-section in the right bar, say below "All tags" --> could be "displayed" at all times,
 but is only visible when scrolling down --> no button or shortcut needed

hajo4:
Also, in some areas it is hard to estimate the size of objects.

There, a measurement-tool would help to see if some building
is a proper house, or just a hut/shed/garage, or a trailer/container.

bhousel:
merged now, closing..

Issue #2039, opened 2013-12-03
Display stats about selected points / line / area


hajo4:
For taking measurement on the map:
- 2 selected points: coordinates, distance, direction, 
- line: length, coordinates of endpoints
- line/way: number of points, running length, length between endpoints
- area: number of points, area (as square-meters), max. width, max. height

This would also be more flexible as a (still missing) scale.

PR #2573, opened 2015-04-02
Infobox / distance measurement

bhousel:
This adds an information widget with some info about the selected items and distance/area measurement capability per #2519 

* Show/hide with 'i' key
* Toggles metric/imperial

cc @aaronlidman @jfirebaugh 

distance-measurement

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,7 @@ D3_FILES = \
 	node_modules/d3/src/behavior/zoom.js \
 	node_modules/d3/src/core/index.js \
 	node_modules/d3/src/event/index.js \
+	node_modules/d3/src/geo/length.js \
 	node_modules/d3/src/geo/mercator.js \
 	node_modules/d3/src/geo/path.js \
 	node_modules/d3/src/geo/stream.js \
--- a/css/app.css
+++ b/css/app.css
@@ -2149,6 +2149,20 @@ img.wiki-image {
     stroke-width: 5;
 }
 
+/* Info Box
+------------------------------------------------------- */
+.infobox {
+    position: absolute;
+    overflow: hidden;
+    top: 220px;
+    width: 200px;
+    height: 150px;
+    z-index: 5;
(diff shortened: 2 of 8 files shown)


pnorman:
Does this change https://github.com/openstreetmap/iD/issues/2039#issuecomment-56242986?

bhousel:
It may end up being a disabled feature that only advanced users can turn on via forking iD or developer console. Not sure yet.

bhousel:
@aaronlidman and @tristen want to add your thoughts on this?  It's intended for power users to measure distance, perimeter, area.   

I know it feels a bit hacky in its current state.  I would eventually like to change the "widgety" things (map-in-map, mapillary, and this) to be plugins, and give them a clearer plugin UI, i.e. give them a dedicated place in the dom to add themselves, allow users to drag/reposition to their liking, and maybe some UI with the name of what it is, and 'X' to close.  But that's for later.

aaronlidman:
The functionality is good as is we just need a design pass to make it fit in with the rest of iD's aesthetic.

tristen:
Made some small adjustments. In summary:

- Variable height depending on the context list
- Added width that alligns with the grid
- Re-use classnames shared by the interface
- Changes to the unit toggle:
    - Shortened name to current unit of measurement (Accounted for this change in i8n)
    - Display as a button to match other toggle/button like elements in the interface

Looks like this:

infobox

bhousel:
Looks nice, thanks @tristen !

aaronlidman:
Nice!

@bhousel Good to merge?

bhousel:
Sure, looks good to merge to me..

sisskind:
@bhousel @tristen I noticed that when you create a polygon feature, the measurement tool only measures the lines created by the user.  If the user double-clicks to create polygon, the last segment is not included in the perimeter measurement.  Double-checked with line tool.  Will open an issue - but wanted to document here.

PR #2576, opened 2015-04-05
Presets : access=yes instead of access=public for toilets

gileri:
Hello,

In general on OSM access=yes is used to mean that an amenity/feature is accessible by everyone.

Toilets seems to be the only exception where access=public is used with no apparent reason. This PR replace access=public by access=yes for toilets.

### DIFF
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -66,7 +66,7 @@
         "type": "combo",
         "label": "Access",
         "options": [
-            "public",
+            "yes",
             "permissive",
             "private",
             "customers"


jfirebaugh:
I assume you made [this change](http://wiki.openstreetmap.org/w/index.php?title=Tag%3Aamenity%3Dtoilets&diff=1156006&oldid=1152826) to the wiki as well. Have you gotten consensus for making this change from the wider OpenStreetMap community, e.g. by posting on the tagging mailing list?

gileri:
You're right I made the wiki edit, and tagging needs consensus. In this instance the consensus was made [long ago](http://wiki.openstreetmap.org/wiki/Key:access#Values).

brycenesbitt:
As the person who wrote "public" on the toilet page: this is a reasonable change for the entry level editor.  Some advanced mappers use 'public' where there is a legal right to use a facility, such as a public park, and "yes" where no such requirement exists.  That's a lot to explain in iD however.

matkoniecz:
See also https://lists.openstreetmap.org/pipermail/tagging/2014-May/thread.html#17501

To quote the entire thread

>> Is there any difference between access=public and access=yes?

> I am not aware of any, seems synonymous to me in the access context, suggest to normalize to yes

i concur.

bhousel:
Merged as d45613b

gileri:
Thank you !

PR #2604, opened 2015-04-22
Added farm as preset for Place

Stalfur:
Place=farm has over 26 thousand entries so far


bhousel:
Looks good, thanks @Stalfur

Issue #537, opened 2013-01-26
Ability to reset entity


ansis:
If you notice you've made a bad edit to an entity after you've already made a lot of good edits, there is no way to reset an entity to its original values without losing all the other changes.

bhousel:
Bumping this issue because I want to implement it to solve an annoying thing about conflict resolution.  If a user makes an conflicting edit and chooses to keep the remote version, what I currently do is replace their local entity with one that is like the remote version.  But this is dumb because it needlessly includes it in the changeset and bumps the entity version number.  What I'd rather do is a proper revert.  Meaning, remove it from the graph so that it doesn't get included in the changeset at all.

I'm planning to the simplest thing possible, which is to implement a `Graph#revert` like this:
```
    revert: function(entity) {
        return this.update(function() {
            this._parentWays[entity.id] = _.clone(base.parentWays[entity.id]);
            this._parentRels[entity.id] = _.clone(base.parentRels[entity.id]);
            delete this.entities[entity.id];
        });
    },
```

That should take the entity out of the edited graph completely and revert to the base version, then it won't be included in the changeset anymore.

@tmcw & @jfirebaugh it looks like you have thought about this a while back - am I missing anything?

PR #2611, opened 2015-04-26
Add ability to revert entity back to base version

bhousel:
see [#537](https://github.com/openstreetmap/iD/issues/537#issuecomment-96315941)

I did not create an operation for it, but may be useful later.  For now I just want it for conflict resolution.

This seems ok to me..
@jfirebaugh @tmcw see any issues?

### DIFF
--- a/index.html
+++ b/index.html
@@ -169,6 +169,7 @@
         <script src='js/id/actions/rotate_way.js'></script>
         <script src='js/id/actions/restrict_turn.js'></script>
         <script src='js/id/actions/reverse.js'></script>
+        <script src='js/id/actions/revert.js'></script>
         <script src='js/id/actions/straighten.js'></script>
         <script src='js/id/actions/split.js'></script>
         <script src='js/id/actions/unrestrict_turn.js'></script>
--- a/js/id/actions/revert.js
+++ b/js/id/actions/revert.js
@@ -1,0 +1,5 @@
+iD.actions.Revert = function(entity) {
+    return function(graph) {
+        return graph.revert(entity);
+    };
+};
--- a/js/id/core/graph.js
+++ b/js/id/core/graph.js
@@ -251,6 +251,16 @@ iD.Graph.prototype = {
         });
     },
 
+    revert: function(entity) {
+        if (this.entities[entity.id] === this.base().entities[entity.id])
+            return this;
+
+        return this.update(function() {
+            this._updateCalculated(entity, this.base().entities[entity.id]);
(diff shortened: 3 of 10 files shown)


jfirebaugh:
What happens in the following scenarios:

* A new node is added to an existing way, and then the node is reverted?
* A new node is added to an existing way, and then the way is reverted?
* A new way is created and added to an existing relation, and then the way is reverted?
* An existing node that's a member of way A is dragged onto way B, connecting them. Then:
  * The node move is reverted
  * Or, one of the ways is reverted

You can see I'm getting at scenarios where multiple entities are involved, but only one of them is reverted -- you can probably think of other scenarios in this vein.

bhousel:
I think calling [`_updateCalculated()`](https://github.com/openstreetmap/iD/commit/6b18066dd695c5b60f998204bf26b0a5aa55edc5#diff-af670b2e1a04df9f14bc9375d65a426fR259) with the base version of the entity (which will be `undefined` if it's a new entity) handles all of these cases.

I did write some tests for a few of these, but I do want to make sure all the scenarios are handled. 
(most of these are really similar)

> A new node is added to an existing way, and then the node is reverted?

New node gets removed, and existing way's nodelist is updated to not include it.

> A new node is added to an existing way, and then the way is reverted?

New node stays (it turns into a point), and existing way's nodelist is updated to not include it.

> A new way is created and added to an existing relation, and then the way is reverted?

New way gets removed and existing relation's memberlist is updated to not include it.

> An existing node that's a member of way A is dragged onto way B, connecting them. Then:
> The node move is reverted
> Or, one of the ways is reverted

I think in all cases `_updateCalculated()` will do the right thing, but just for our sanity I will expand the tests.

jfirebaugh:
` _updateCalculated()` handles bookkeeping internal to `Graph`, but it doesn't update a containing way's `nodes` or relation's `members`.

bhousel:
Good point...   I think `Graph#revert` is ok but I definitely need to fix `iD.actions.Revert`

bhousel:
7adb0dd now updates parent way's `nodes` / parent relation's `members`, for new entities.

We should continue discussion of this on #537 before deciding if it's solid enough for an operation or shortcut key.

I'm pretty sure it is safe for the purpose that I am using it for in [save.js](https://github.com/openstreetmap/iD/commit/a565b72f7da030bbe5ca798a38f1926ec626223b) - to exclude certain local edits from the changeset where the user chooses "keep theirs" in conflict resolution.

PR #2617, opened 2015-04-30
fix: build.js tests wrong fields in validatePresetFields()

species:
validatePresetFields() tested old fields from file, not newly aggregated ones.


almereyda:
Discovered during the http://dossier.allmende.io/WitzSprint/

bhousel:
Thanks!

PR #2618, opened 2015-04-30
Added oil well preset

1ec5:
This PR adds a preset for [`man_made=petroleum_well`](http://taginfo.openstreetmap.org/tags/man_made=petroleum_well) called “oil well”, along with various synonyms. It uses the existing oil well icon.

@gregcrago also asked for a way to tag oil fields – areas where one can find many of derricks or pumpjacks. Unfortunately, there [doesn’t appear to be consensus](http://wiki.openstreetmap.org/wiki/WikiProject_Oil_and_Gas_Infrastructure#Tags_currently_in_use_for_Oil_and_Gas_Infrastructure) around a particular tagging scheme for oil fields yet. The only common tag is specifically for [offshore fields](http://wiki.openstreetmap.org/wiki/Tag:waterway%3Doffshore_field). So I only added a preset for individual wells, to be applied to points.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -1570,6 +1570,9 @@ en:
       man_made/observation:
         name: Observation Tower
         terms: "<translate with synonyms or related terms for 'Observation Tower', separated by commas>"
+      man_made/petroleum_well:
+        name: Oil Well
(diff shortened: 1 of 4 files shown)


bhousel:
Thanks @1ec5!

PR #2621, opened 2015-05-01
Corrected oil well preset tag

1ec5:
@gregcrago pointed out that the oil well preset added for #2618 actually adds `man_made=mast` instead of `man_made=petroleum_well`, due to my copy-pasta. :disappointed:

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -5449,7 +5449,7 @@
             "pumpjack"
         ],
         "tags": {
-            "man_made": "mast"
+            "man_made": "petroleum_well"
         },
         "name": "Oil Well"
     },
--- a/data/presets/presets/man_made/petroleum_well.json
+++ b/data/presets/presets/man_made/petroleum_well.json
@@ -13,7 +13,7 @@
         "pumpjack"
     ],
     "tags": {
-        "man_made": "mast"
+        "man_made": "petroleum_well"
     },
     "name": "Oil Well"
 }
--- a/data/taginfo.json
+++ b/data/taginfo.json
@@ -1281,7 +1281,7 @@
         },
         {
             "key": "man_made",
-            "value": "mast"
+            "value": "petroleum_well"
         },
         {
             "key": "man_made",


bhousel:
Thanks for the fix, Minh..

1ec5:
@bhousel, do the strings need to be pushed to Transifex manually somehow? osm.org picked up the new release before translators got any opportunity to update their localizations.

tyrasd:
Actually, strings should be synced automatically (daily if I remember correctly). But it looks like the last update of the source file at transifex was on March 25. Maybe there was some hiccup with the sync process and it therefore it stopped working and just needs a kick-start? (At least that's what happened for my *overpass turbo*'s project on transifex a few weeks ago.) //cc @jfirebaugh

bhousel:
It should check for the latest `core.yaml` and `presets.yaml` files from our Github repository. And it looks like this is working.  On the dashboard I see `core.yaml` updated May 20th and `presets.yaml` updated May 25th.

Before each release, we run `make translations` to pull in the latest translations from Transifex.  I remember doing this release pretty quickly after the preset got added, so that's why the translations aren't there.  Going forward I'll try to wait a bit longer between changing the presets and cutting a release.

tyrasd:
so, why aren't the changes from [#2618](https://github.com/openstreetmap/iD/pull/2618/files) (from Apr 30) not on Transifex yet? Also, to me on transifex' dashboard the last update of the presets resource appears to have happened on March 19th:

selection_046

bhousel:
That is weird.. I look on the "Resources" tab and I see this:

screenshot 2015-05-27 07 59 44

bhousel:
Oh maybe this is the issue.. Transifex thinks these URLs are no good:

screenshot 2015-05-27 08 47 51

I don't understand why it would say that.. Github seems to serve these from a varnish cache with 200 OK, so it's not even like there is a redirect.

tyrasd:
Yeah, that's exactly what happened to me with overpass turbo on transifex, too. I don't know what the problem was, but as far as I remember the error got away after entering the same URL once again in the settings.

bhousel:
I updated the URLs to point to `openstreetmap/iD` so we'll see if that helps.   They don't say "URL not found" anymore, but we may have to wait a day or so for the next time Transifex tries to sync.

bhousel:
Good news, it looks like updating those URLs fixed the issue!   Transifex picked up a current `core.yaml` and `presets.yaml` last night, and these do include the oil well preset, as well as the other new presets I added yesterday.

tyrasd:
:+1:

Issue #2633, opened 2015-05-04
show "changeset tags" to the user before saving changes.


ImreSamu:
related to #2449  ( + privacy )

For  greater transparency,  it would be useful showing to the user  the  "changeset tags" before saving changes.  ( or create an options to see )  

statistics:
-   https://gist.github.com/ImreSamu/0fc3b8015e6dd16595c5

pnorman:
I'd like to be able to add changeset tags so we can properly tag changesets in iD, but believe the privacy issue is a red herring. Presumably the full changeset tags aren't going to be shown by the standard workflow, only if a user decides to add/remove tags, vaguely similar to the JOSM way of not showing changeset tags by default. Most users won't see them, so it doesn't change their awareness of the details of what is uploaded.

bhousel:
@ImreSamu very cool!

I do see in the report that we should probably be more consistent with the locale (e.g. having both 'en-us' and 'en-US' in the results is sloppy, so I can just lowercase them all).

Agree, @pnorman  I think we addressed the privacy concerns already in #2559.  Maybe I can find  a way to put a collapsed raw tag editor on the save screen, so users can actually add their own changeset tags.

ImreSamu:
My latest osm changeset statistics : https://gist.github.com/ImreSamu/2b31315ef2328cb482c7

PR #2643, opened 2015-05-13
Update connection.js

schwindp:
Removed browser and platform as changeset tags. I don't really see a need for pushing this information into the OSM database

(better force the need for a changelog comment when uploading, but that's another issue)

### DIFF
--- a/js/id/core/connection.js
+++ b/js/id/core/connection.js
@@ -249,8 +249,6 @@ iD.Connection = function() {
                 imagery_used: imageryUsed.join(';').substr(0, 255),
                 host: (window.location.origin + window.location.pathname).substr(0, 255),
                 locale: detected.locale,
-                browser: detected.browser + ' ' + detected.version,
-                platform: detected.platform
             };
 
         if (comment) {


bhousel:
@ImreSamu is producing a very useful set of statistics (see #2633) which helps us know how iD is being used.  We do sometimes run into platform or browser bugs that we need to workaround, and it's good to know how many users are affected by these.  To limit the amount of data gathered, we only log the major browser version (not all the way down to patch level) and don't capture the OS version at all.  In fact, these tags that iD saves in the changeset are far less uniquely identifying than the JOSM version tag.

All that said, I do understand that some users are uncomfortable about seeing information about their browser and OS saved in their changeset tags.  

As a compromise how about this:  instead of removing the tags completely, I can honor the user's `doNotTrack` preference (supported by all major browsers now) and if it is set we will save nothing but the iD version tag?

bhousel:
looping in @pnorman because I value his opinion here...

schwindp:
While I understand the reason to have user statistics (or at least statistics how many % of users might be affected by a code change), isn't it at least possible to anonymously sample that data? Writing those tags into changesets is unfortunately fully user-traceable.

bhousel:
> While I understand the reason to have user statistics (or at least statistics how many % of users might be affected by a code change), isn't it at least possible to anonymously sample that data?

iD and its various forks are hosted on several websites other than openstreetmap.org.  Some of those sites use analytics to gather anonymous statistics and some don't.  Either way, we don't have or want access to our partners' analytics data.


> Writing those tags into changesets is unfortunately fully user-traceable.

Again, I'm willing to exclude the tags if a user has `doNotTrack` set.  Is this something that would satisfy your concerns, or are we going to continue talking past each other?

pnorman:
I think we should always be recording iD version, hosting site, and locale. These are useful for mappers for seeing why a user made a particular mistake. If any third party hosting an iD instance doesn't want it disclosed, they can patch out the hosting site as they are by definition hosting their own copy of iD.

I don't see the value for mappers in knowing the OS and browser a user mapped from. I don't think we've had any bugs that would manifest themselves in changesets that have been specific to a browser or OS. The browser-specific bugs tend to be ones where something doesn't work.

Full browser version is already recorded in the server logs by means of user-agent, not just for iD on osm.org, but any iD uploading to osm.org. If we need usage information, I'd rather get them there, even if they might be mixed with an insignificant percentage of P2.

schwindp:
I'm fully satisfied with the proposal by @pnorman - normal users just don't know the "donotTrack" feature.
No one should have the need to opt out in order to not have personal data collected.

ImreSamu:
I am  _also_  fully satisfied with the proposal by @pnorman
For  "Translation QA" -  the 2 most important info :  "iD version" and "locale".

And my latest osm changeset statistics : https://gist.github.com/ImreSamu/2b31315ef2328cb482c7

bhousel:
Sounds good, if everybody is happy with dropping these then I'll merge it...

PR #2681, opened 2015-06-07
Use HTTPS if location protocol is HTTPS in iD.Connection

frewsxcv:
Relevant to https://github.com/strava/iD/issues/5#issuecomment-109800372

### DIFF
--- a/js/id/core/connection.js
+++ b/js/id/core/connection.js
@@ -1,12 +1,17 @@
-iD.Connection = function() {
+iD.Connection = function(useHttps) {
+    if (typeof useHttps !== 'boolean') {
+      useHttps = window.location.protocol === 'https:';
+    }
+
     var event = d3.dispatch('authenticating', 'authenticated', 'auth', 'loading', 'loaded'),
-        url = 'http://www.openstreetmap.org',
+        protocol = useHttps ? 'https:' : 'http:',
+        url = protocol + '//www.openstreetmap.org',
         connection = {},
         inflight = {},
         loadedTiles = {},
         tileZoom = 16,
         oauth = osmAuth({
-            url: 'http://www.openstreetmap.org',
+            url: protocol + '//www.openstreetmap.org',
             oauth_consumer_key: '5A043yRSEugj4DJ5TljuapfnrflWDte8jTOcWLlT',
             oauth_secret: 'aB3jKq1TRsCOUrfOIZ6oQMEDmv2ptV76PA54NGLL',
             loading: authenticating,
--- a/test/spec/core/connection.js
+++ b/test/spec/core/connection.js
@@ -2,11 +2,23 @@ describe('iD.Connection', function () {
     var c;
 
     beforeEach(function () {
-        c = new iD.Connection({});
+        c = new iD.Connection();
     });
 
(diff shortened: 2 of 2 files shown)


pnorman:
Looks like the tests need revising too.

frewsxcv:
I fixed all the tests except one. If anyone has ideas about what is going wrong there, let me know

pnorman:
```
  iD.modes.AddPoint
    clicking the map
XMLHttpRequest cannot load file://www.openstreetmap.org/api/0.6/map?bbox=-77.02514648437499,38.89958342598271,-77.01965332031249,38.90385833966776. Cross origin requests are only supported for HTTP.
  1) "before each" hook
```

Ugh. This is a nasty problem.

You can't assume that the OSM API is accessible on the same protocol that iD is loaded by, because iD could be loaded locally over file://

jfirebaugh:
Chrome has some restrictions for resources loaded via `file://` that generally make loading it that way unadvisable. That said, it's probably better to do something like `location.protocol === 'http' ? 'http' : 'https'`.

frewsxcv:
How does it look now? I added an optional parameter to the `iD.Connection` constructor mainly because I didn't want to figure out how to mock `window.location` in order to write a regression test

frewsxcv:
Alternatively, we could just make all connections secure and disallow HTTP...

frewsxcv:
Anyone have thoughts on this?

pnorman:
Not all API instances allow HTTPS.

frewsxcv:
My current changes do not force HTTPS, unless the user is already on an HTTPS site

pnorman:
> My current changes do not force HTTPS, unless the user is already on an HTTPS site

Yes, but you were suggesting disallowing HTTP

frewsxcv:
> Yes, but you were suggesting disallowing HTTP

Yeah, I understand. Sorry I didn't make that clear in my previous message.

bhousel:
I notice that using the [source switcher](https://github.com/openstreetmap/iD/blob/master/index.html#L261-L272) on an https site will break this, but it's broken anyway because `api06.dev.openstreetmap.org` doesn't support https, so I guess not that big a deal.

Thanks for adding this, @frewsxcv

js/id/core/connection.js
```diff
@@ -1,12 +1,17 @@
...
-        url = 'http://www.openstreetmap.org',
+        protocol = useHttps ? 'https:' : 'http:',
+        url = protocol + '//www.openstreetmap.org',
         connection = {},
         inflight = {},
         loadedTiles = {},
         tileZoom = 16,
         oauth = osmAuth({
-            url: 'http://www.openstreetmap.org',
+            url: protocol + '//www.openstreetmap.org',
```
1ec5: This is unnecessary: just start the URL with `//`, and the current protocol will be used (except in Netscape Navigator 4, I suppose).

frewsxcv: Look through he discussion the associated pull request. Sometimes, we are retrieving this js file via the `file://` protocol which results in an undesired outcome

1ec5: Doh! Sorry about that. :smile:

PR #2686, opened 2015-06-09
Add cycleway preset field for highways with bike lanes

ebrelsford:
Add a cycleway preset field for highways with bike lanes. This can be complicated as there are numerous types of bike lanes and it's helpful to include which side of the highway the bike lane is on.

Hope this is okay, please let me know if there is anything I can do to improve it.


bhousel:
I think this is mostly pretty good!

Some questions:
* Should we really show this field for `highway=path` and `highway=track`?
* It looks like the tagging is either `cycleway` *or* `cycleway:left`/`cycleway:right`.  Is it an issue if users fill it in both ways?  Or does it make more sense to disable the part of the field that isn't being used?

@samanpwbb what do you think? 

screenshot 2015-06-12 15 59 33

ebrelsford:
Thanks for looking at it. I would be fine making both of the changes you're
suggesting. When left or right is set it would be easy enough to set
cycleway to none.

samanpwbb:
This form is a bit confusing. Would it be possible to remove the 'both sides' input and have only two fields. Having that _as well as_ a left side and right side input makes this a bit hard to parse. If user inputs the same value for left and right, can could then automatically set the `cycleway` tag to the value both inputs share.

ebrelsford:
Sure, I could see that working. I suppose at the time I thought the way it
is would be more convenient, but I agree that less confusing is likely more
important here.

ebrelsford:
I removed cycleway from `highway=path` and `highway=track`, removed the "both sides" dropdown, and added some logic to decide when to use which tags. Thoughts?

cycleway

bhousel:
Looks great, thanks for building this!

ebrelsford:
Great, thanks for the feedback and for merging!

Issue #2457, opened 2014-11-28
Use CLDR for language list source


kmilos:
Currently iD only supports the 'name:sr' tag in the locale dropdown list, and wrongly labeled as 'Српски / Srpski' at that.

This locale is not intended to be used with both scripts, but Cyrillic only, and should be labeled as 'Српски' only, or 'Serbian (Cyrillic)'.

For Serbian written in Latin script, we have a separate locale 'name:sr-Latn' which should be added to iD dropdown as 'Srpski' only, or 'Serbian (Latin)'.

1ec5:
The `wikipedia` tag's value specifies a Wikipedia edition, not necessarily a language. Other examples of mismatches include `zh-min-nan` instead of the ISO code `nan` and `simple`, which doesn't correspond to any ISO language code. So it isn't a great source for language codes.

tmcw:
@1ec5 is CLDR the answer, then?

1ec5:
CLDR would work, certainly, but it’s a very large package that contains much more than language names. Someone may’ve already created a subset of it specifically for this use case.

bhousel:
Just a quick followup to this issue that it's now easy to pull in just the parts of CLDR that we want.  It has been split into subpackages, automatically converted from xml to json, and published on npm:
https://github.com/unicode-cldr/cldr-json#package-organization

1ec5:
Per #4632, we’ll need to introduce some iD-specific overrides to support `name:*` tags that are qualified with an ISO 15924 or ad-hoc script code, such as [`name:sr-Latn`](https://taginfo.openstreetmap.org/keys/name:sr-Latn) or [`name:zh-hant`](https://taginfo.openstreetmap.org/keys/name%3Azh-hant).

PR #2688, opened 2015-06-11
Switch Wikipedia list to osmlab/wmf-sitematrix

### LINKED ISSUE #2682: Wikipedia list is outdated
[wikipedia.json](https://github.com/openstreetmap/iD/blob/master/data/wikipedia.json) lists 285 Wikipedias, but [Wikimedia’s site matrix](https://meta.wikimedia.org/wiki/Special:SiteMatrix?uselang=en) lists 288 and new Wikipedias are added somewhat regularly as they graduate from Wikimedia Incubator. iD’s build process should call the [MediaWiki API](https://en.wikipedia.org/wiki/Special:ApiSandbox#action=sitematrix&format=json) to regenerate the list more easily.

/cc @aaronlidman

1ec5:
This PR switches the list of Wikipedias to the one maintained in [osmlab/wmf-sitematrix](https://github.com/osmlab/wmf-sitematrix). That list comes from the MediaWiki API, so it’ll be easy to update. Some things to note, however:

 - The MediaWiki API sorts the wikis by ISO 639 language code rather than article count.
 - Closed wikis are filtered out. Do we need them in iD?

Fixes #2682.

/cc @aaronlidman


bhousel:
> The MediaWiki API sorts the wikis by ISO 639 language code rather than article count.

[This line](https://github.com/openstreetmap/iD/blob/master/js/id/ui/preset/wikipedia.js#L77) seems to expect English first, but could probably just be replaced with an 'en'.

> Closed wikis are filtered out. Do we need them in iD?

I think not.  They're usually closed because they have no meaningful pages.

1ec5:
> [This line](https://github.com/openstreetmap/iD/blob/master/js/id/ui/preset/wikipedia.js#L77) seems to expect English first, but could probably just be replaced with an 'en'.

@bhousel, good point. I beefed up the logic to fall back to the browser’s locale, or, failing that, English.

1ec5:
Another consequence of using this list is that some language names (español) start with a lowercase letter, per CLDR. Is that desirable, or should we title-case these names in iD?

bhousel:
@1ec5 I'm really sorry, I thought this was merged already!

bhousel:
> Another consequence of using this list is that some language names (español) start with a lowercase letter, per CLDR. Is that behavior desirable, or should we title-case these names in iD?

Not a big deal, we'll fix it with #2457 when we switch the language list to something better.

PR #2690, opened 2015-06-12
Better look for mapillary markers

pgiraud:
In my opinion, the markers used for layer mapillary are ugly.
With this pull request, the styling is a bit better, well I hope.

Before
screenshot from 2015-06-12 16 48 39

After
screenshot from 2015-06-12 16 48 51


bhousel:
This looks really nice, thank you!!

PR #2691, opened 2015-06-13
Adds increased zooming/panning keyboard shortcuts

rowanhogan:
The following suggestions have come up from previous crisis mapping from individuals as a part of [RHoK Brisbane](http://rhokbrisbane.org/) and [MSF](http://www.msf.org/) [mapathons](http://wiki.openstreetmap.org/wiki/Previous_Missing_Maps_parties).

This adds extra keyboard shortcuts (with modifier keys) for increased zooming and panning.

When scanning areas for crisis mapping tasks from [tasks.hotosm.org](http://tasks.hotosm.org/), it can be quite tedious to move around large areas with little/no human-made features with the existing controls.

- The increased zoom functionality (Ctrl & +/-) increases/decreases the zoom level by 4 rather than 1.
- The increased pan functionality (Shift & arrow keys) pans the map by the map viewport width/height.

The development was done as a part of [RHoK Brisbane Winter Hackathon 2015](http://www.meetup.com/RHoK-Brisbane/events/221665857/)


bhousel:
I played around with this a bunch today and I really like it...

My one question is that it's a little inconsistent using "Control" as the modifier key for fast zooming and "Shift" as the modifier key for fast panning...

Is it ok if I change this code to use `iD.ui.cmd('⌘' + key)`, so that the shortcuts use the same modifier that we use elsewhere in iD?  (this function will map to "Command" on Mac and "Control" on Windows/Linux).  This would make it a more consistent experience and easier to to document on the keyboard shortcut wiki too.

I have a feeling you might have used "Shift" for the panning because, by default, browsers use Command-arrow as forward and backward navigation shortcuts - but by adding `d3.event.preventDefault();` we can override this browser behavior that isn't really helpful for an editor like iD anyway.

(Also worth mentioning that the `d3.keybinding` code is smart enough to not grab these events when a user happens to be using an input or textarea field.)

rowanhogan:
Hi @bhousel,

Thanks for your feedback! I'm very open to suggestions for the modifier key.

My reasoning for the usage of "Shift" for panning stems from the forward/back shortcuts for Ctrl/Cmd as you mentioned, but also from graphics editors such as Adobe Illustrator (for example) that use the same modifier for increasing the amount panned.

Also with the zoom shortcuts, I felt "Shift" may have been awkward for some users as it was available previously to be used for the regular zoom function – as you'd normally have to use "Shift" in addition to the "=" key to get the "+" character.

I agree, consistency is the best way forward. Personally I would probably lean towards "Shift" for both, but I understand your reason for disabling the back/forward shortcuts in this app (not a bad idea to prevent pressing it accidentally anyway).

bhousel:
> My reasoning for the usage of "Shift" for panning stems from the forward/back shortcuts for Ctrl/Cmd as you mentioned, but also from graphics editors such as Adobe Illustrator (for example) that use the same modifier for increasing the amount panned.

Sounds good, let's keep "Shift" for fast panning.

> Also with the zoom shortcuts, I felt "Shift" may have been awkward for some users as it was available previously to be used for the regular zoom function – as you'd normally have to use "Shift" in addition to the "=" key to get the "+" character.

Yes, the Shift key was used for that reason for normal zooms.. 
So I'm making it so that:

* Cmd will speed up both zooms and pans
* Shift will speed up pans only

rowanhogan:
Sounds good, did you want me to make that change?

bhousel:
> Sounds good, did you want me to make that change?

No it's ok, I'm doing it now (sorry for the confusion!)

bhousel:
All merged, thanks again for your contributions! :)

rowanhogan:
Thanks!

PR #2692, opened 2015-06-14
Error handling is added to the GPX loading callback.

duemir:
Previously erroneous response was corrupting the GpxLayer, which could
potential corrupt the whole Map, if user were to use 'Zoom to GPX'
feature. Callback logic is wrapped in the error guard to fix this.


jfirebaugh:
Thanks! Committed as be1e3738f1b5e3a3aa2448164ed51be726571421.

PR #2693, opened 2015-06-14
Add custom GPX layer to minimap

rowanhogan:
_The following suggestion has come up from previous crisis mapping from individuals as a part of [RHoK Brisbane](http://rhokbrisbane.org/) and [MSF](http://www.msf.org/) [mapathons](http://wiki.openstreetmap.org/wiki/Previous_Missing_Maps_parties)._

This adds custom GPX layers (from either URL param or file upload) to the minimap.

This is very useful when working on a task from [tasks.hotosm.org](http://tasks.hotosm.org/) (see below screenshot). Seeing the task boundary (as a GPX layer) in the minimap enables a user to be able to easily locate themselves within a given task.

Screenshot

A distribution with this functionality (and that from #2691) can be viewed at [id.rhokbrisbane.org](http://id.rhokbrisbane.org).

An example based on a recent HOTOSM TM task [can be viewed here](http://id.rhokbrisbane.org/#background=Bing&gpx=https://s3-ap-southeast-2.amazonaws.com/id.rhokbrisbane.org/123.gpx&map=16.10/85.84287/27.70792). Press `/` to open the minimap.

This PR also adds a visual option to toggle the minimap under the background tab.

The development was done as a part of [RHoK Brisbane Winter Hackathon 2015](http://www.meetup.com/RHoK-Brisbane/events/221665857/)


bhousel:
Awesome work!  I added a tooltip to the minimap toggle control, but otherwise left it as-is.. Thank you!

rowanhogan:
Thanks @bhousel! Yeah good call on the tooltip. We didn't even know iD had a mini-map until we started digging around the codebase, so we thought it was worth making more prominent as it's a great feature (and now shows custom GPX layers as well, so even more useful).

PR #2694, opened 2015-06-14
Tighten up Wikipedia URL handling

1ec5:
If you paste a full Wikipedia URL into the Wikipedia preset field, iD tries to detect the Wikipedia language edition and normalize the tag value to the `xy:Foo` format. It also round-trips from this format back to a URL for the link button. This PR fixes a few edge cases around these normalization steps:

- Some Wikipedia subdomains contain hyphens, e.g. zh-min-nan.wikipedia.org.
- Some Wikipedia subdomains have variants, so paths such as zh.wikipedia.org/zh-hk/ are possible.
- When a map feature is covered by a section of a Wikipedia article, but not a full-blown article, the tag value can contain an anchor. MediaWiki has a special format for anchors that looks like URL encoding ([example](https://zh.wikipedia.org/zh-hk/%E8%8C%AB%E6%9B%B2%E9%95%87#.E8.A1.8C.E6.94.BF.E5.8C.BA.E5.88.92)), but the `wikipedia` tag value usually contains the unencoded form.

ref #2688

### DIFF
--- a/js/id/ui/preset/wikipedia.js
+++ b/js/id/ui/preset/wikipedia.js
@@ -84,12 +84,22 @@ iD.ui.preset.wikipedia = function(field, context) {
 
     function change() {
         var value = title.value(),
-            m = value.match(/https?:\/\/([-a-z]+)\.wikipedia\.org\/wiki\/(.+)/),
-            l = m && _.find(iD.data.wikipedia, function(d) { return m[1] === d[2]; });
+            m = value.match(/https?:\/\/([-a-z]+)\.wikipedia\.org\/(?:wiki|\1-[-a-z]+)\/([^#]+)(?:#(.+))?/),
+            l = m && _.find(iD.data.wikipedia, function(d) { return m[1] === d[2]; }),
+            anchor;
 
         if (l) {
             // Normalize title http://www.mediawiki.org/wiki/API:Query#Title_normalization
-            value = m[2].replace(/_/g, ' ');
+            value = decodeURIComponent(m[2]).replace(/_/g, ' ');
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Can you fix the lint error?

1ec5:
All set.

bhousel:
I tried pasting your [example zh-hk url above](https://zh.wikipedia.org/zh-hk/%E8%8C%AB%E6%9B%B2%E9%95%87#.E8.A1.8C.E6.94.BF.E5.8C.BA.E5.88.92) into the wikipedia field, but it didn't pick up the correct locale (it was left as English).

It looks like the [regexp here](https://github.com/openstreetmap/iD/pull/2694/files#diff-e10680be77be83f468321b6890893e17R87) doesn't match a url like this.. Is this expected?

1ec5:
@bhousel, what browser are you testing in? It works for me in Firefox 40.0a2 (2015-06-07), Safari 8.0.6, and Chrome 44.0.2376.0. I tested both `en-US` and `vi` as my `navigator.language`.

1ec5:
Note that the conversion only happens when pasting into the Wikipedia preset field, not the field in All Tags.

bhousel:
> what browser are you testing in?

Chrome 43.0.2357.124 on Mac OS X 10.10.3

I'm going to test again in a bit.  I just tried the regex and the url in the dev console and it matched it fine.  I was probably on the wrong branch when I tested the first time.

bhousel:
Yes, sorry about that - it's working totally fine.  I must have done a dumb thing before..

Issue #1761, opened 2013-08-28
Add/modify the same tag to multiple selected objects


smsm1:
Would it be possible to select multiple road segments and then to edit the tags for all of the selected items, such as setting the maxspeed or sidewalk tag?

mapninja:
+1 to this idea. Would cut editing time in half on creating building footprints for HOT tasks, like the current Nepal effort.

bhousel:
> Would cut editing time in half on creating building footprints for HOT tasks, like the current Nepal effort.

I was actually doing the same thing last night to test the new release, and I totally agree - we can do better.

I'm bumping this issue and #528 to the front of the my queue, because of the time it can save in tracing building outlines.

cromlyngames:
@bhousel just had to square off then label as building 150 odd buildings in one of the missingmaps tasks. 
Let me know if there's any way I can help on this. Happy to code but coming in cold.

sharazam:
So yeah, what happened to this? It's been a year ... UI wise, it should just auto-apply the last selected type instead of the white, feature-less area. Or at least have an option in the options menu to enable this behaviour.

bhousel:
> It's been a year

Yeah, we've done a lot in the last year, check the CHANGELOG.

> Or at least have an option in the options menu to enable this behaviour.

iD don't have an options menu.  I'm going to add one in #3123, then this feature will have a place to be.

Anyway, we're tracking these issues as priorities on the [project tracker](https://github.com/openstreetmap/iD/projects/1).  I'll get to them soonish.

pietervdvn:
Hello,

What is the current state of this issue? Has work already started on this? If so, where can we find it?

We are willing to spend some time on it - we badly need the feature. This would also mean that the code would be pull-requested back if there is still an interest in this.

However, then we would like a little guidance on how to implement it.

bhousel:
Hey @pietervdvn ... If you really badly need the feature, use an editor like JOSM.

We're really not able to hand-hold new contributors right now.   The dev team is heads-down preparing the v2.15 release, and after that, our focus for the forseeable future will be working on iD v3.

pietervdvn:
No problem, I do understand your sentiment.

is this feature still on the roadmap? (As: if I would be able to figure it all out by myself and send a PR, would it be accepted?)

If not, perhaps this issue should be closed.

bhousel:
> is this feature still on the roadmap? 

Yes

[...]

Issue #435, opened 2013-01-18
Operations: merge


tmcw:
This should be a 'unified merge' I think, which works for (uses different algorithms) for lines & for areas, but has the same UI.

homersimpsons:
Sorry for #3074 didn't find this tread when research...
I hope this will exist, in France a lot of building are extracted from 'cadastre' (government map) but there are a lot of buildings composed of multiple area like [here (google maps)](https://www.google.fr/maps/@44.6656798,2.1594132,20.6z) , I have corrected this in [OSM](http://www.openstreetmap.org/#map=19/44.66561/2.16006) [I kept some fragments because I'm not there to check if it's really one building or differents ones]. But it's ugly and constitute a lack of space. This tool could be apply on 70% of multiple area I think, so this would be really usefull.
Another helpfull tool could be split (from 2 non consecutive points of an area) in 2 area, this is usefull when mapping adress a group of building.

egonzale:
Hi, I think this is the right place to mention a problem we had yesterday.
Mapping riverbanks as indicated in wiki, we make them by sections, so we draw polygons that have a common border where they continue (see example [here](http://www.openstreetmap.org/edit?editor=id&#map=18/10.22995/105.35048) ).
Problem is that if you select them and merge them (using the + symbol), it merges "ways" instead of "area" which is what you would expect... mergin these areas as ways simply breaks it all (you can test in iD and then try to make some sense of it in JOSM...)
Would it make sense to change the algorithm when two areas are selected (and have a common border) they are merged into one single area? Does it make sense?
As reference, the action currently implemented in the + symbol seems to do the same as in JOSM Tools>Combine ways ... the one I would expect is JOSM Tools>Join overlaping areas

homersimpsons:
@egonzale This is exactly what I was asking for in my previous answer :) Hope this will be realised.

bhousel:
Hi @egonzale and @homersimpsons, yes I'd really like to see iD have an area merge tool.  Ideally my plan is to bring in turf's polygonal operations like `turf-union`, `turf-difference`, `turf-intersect` after they finish removing the JSTS dependency, see https://github.com/Turfjs/turf/issues/88

This is something that the Turf team could use help with and would benefit not just iD but a lot of projects. Maybe someone whose skills lie at the intersection of JavaScript and computational geometry could move this across the finish line (paging dr. @paulmach?).

EdLoach:
[...]

PR #2699, opened 2015-06-17
Orthogonal / Building drawing tool

bhousel:
This adds a special drawing mode that lets you draw orthogonal shapes in just 2 clicks!
Drag along one wall and then click the opposite corner..

buildings

It is also smart enough to connect shapes to one another, inserting midpoints where needed..

buildings2

CC @jfirebaugh @aaronlidman @samanpwbb @tristen 

I could use some code review and design advice before it is ready to merge.

TODO
- [ ] restore hover behavior
  I had to disable hover behavior when in orthogonal drawing mode because the existing hover behavior relies heavily on mouseover.  I think I can reimplement just by classing the right things in the move event handler.
- [ ] figure out UI for this
  Currently for the UI, I just assigned shift-click to start drawing ways and areas in this mode.  I feel that is a bit cheating (and won't work on tablets), so I think what might be nicer is a way to cycle through the available drawing options by clicking on the Way/Area buttons and having the icon/text change, or maybe repeated presses of '2' or '3'.  Or maybe make those into dropdowns.  Not really sure what would be best.
- [ ] add to help/walkthrough
  Self explanatory
- [ ] support for circles too?
  The draw_way code is now mostly setup to accept any number of click targets, so it's not a big leap to make options to draw other shapes.


tristen:
@bhousel this is awesome!

> Currently for the UI, I just assigned shift-click to start drawing ways and areas in this mode. I feel that is a bit cheating (and won't work on tablets), so I think what might be nicer is a way to cycle through the available drawing options by clicking on the Way/Area buttons and having the icon/text change, or maybe repeated presses of '2' or '3'. Or maybe make those into dropdowns. Not really sure what would be best.

Personally, I don't think there needs to be a UI around this. shift-click is a great option for exposing it and we can do a good job documenting it under a key bindings section in the documentation. Tablet is an issue but perhaps it's enabled via gesture? (i.e pinch when the editor is in a polygon draw mode)

quarksteilchen:
Hi,
just an idea to make this a bit more intuitive and with less clicks:

You could just click one corner, then the diagonally opposite corner. a rectangle could then be automatically generated and you do not have the limitation of having to be very exact on the first line.

houses

Starting this mode with shift is a good idea i think.

bhousel:
> You could just click one corner, then the diagonally opposite corner. a rectangle could then be automatically generated and you do not have the limitation of having to be very exact on the first line.

Drawing a side is needed because two clicks are not enough to uniquely identify a rectangle..

You can fit ∞ rectangles to a diagonal segment:

twoclicks

quarksteilchen:
you are right :-). sorry.

slhh:
A drawing mode requiring two clicks only in most cases seems to be possible. In other cases an dragging one node will be required in addition to this.
Most buildings are likely parallel/orthogonal to the next building or the street. A rectangle is well defined by one diagonal segment if the angle of the rectangle is known.
Assume the rectangle is parallel/orthogonal to the next building if it is potentially touching.
Otherwise assume the rectangle is parallel/orthogonal to the next street, and if the next street is too far away assume the rectangle is a square, which is well definded by one diagonal segment.
This is likely a good guess, but of course it is sometimes wrong. Thus the user needs to be able to modify it. This can be done by dragging one of the other nodes not belonging to the diagonal segment. In case of dragging such a node immediately after drawing the diagonal segment dragging should be restricted along the circumcircle of the rectangle. Refitting the rectange to the diagonal segment the fourth node should be moved automatically.

slhh:
For the UI the lower right corner of the Area button could contain a small dropdown arrow opening a menu with multiple drawing modes. The Area button should indicate the active mode. The button text "Area" can change to a multiline text "Area - 3-click" or "Area - 2-click" if the normal mode is not active.

iandees:
In JOSM, you get two clicks by selecting an existing "orthogonal-ish" polygon whose angles will be used to make further polygons.

That took me a really long time to figure out, so I thought I'd add it here. :smile:

bhousel:
rebased to 1.9.4  😏

pnorman:
Many buildings are L shaped. Can this help with those buildings?

On your second animation, do the 1st and 2nd buildings share a common node at the southeast of their common wall?

bhousel:
> Many buildings are L shaped. Can this help with those buildings?

No, this just draws rectangles.  But maybe someday we'll support merging 2 rectangles together with #435 (likely), or possibly extrude edges with #2986.

> On your second animation, do the 1st and 2nd buildings share a common node at the southeast of their common wall?

Yes, they do - I didn't hover long enough over it to see, but [this code](https://github.com/openstreetmap/iD/blob/a0744d234bf95c6fa06c701b366912bd060b4c59/js/id/behavior/draw_way.js#L180) connects nodes to existing lines, unless you hold down the 'Alt' key.

pnorman:
>>  On your second animation, do the 1st and 2nd buildings share a common node at the southeast of their common wall?
> Yes, they do - I didn't hover long enough over it to see, but this code connects nodes to existing lines, unless you hold down the 'Alt' key.

How far off from 90 degrees can you get with the first line of the second building before they no longer join?

bhousel:
> How far off from 90 degrees can you get with the first line of the second building before they no longer join?

IIRC the snapping threshold is something like within 3px.

aawiseman:
@bhousel I'm excited for this! Any ETA? I think it will help a lot with mapathon data quality and such.

bhousel:
> Any ETA? I think it will help a lot with mapathon data quality and such.

Sorry no ETA, I have to pretty much start over after the iD v2 refactor - it's not possible to merge the old code.  I think early 2017 is do-able though.

homersimpsons:
Any news on this ?

bhousel:
> Any news on this ?

Nope, news would be here.  In my mind I have an idea of how #3123 could unblock it.

dalekunce:
@bhousel wanted to plug putting some time in on this. @mikelmaron and I looking at a bunch of ways to improve quality of edits during mapathons and this is for sure one of the top ways.  We are having a huge upswing in new mappers at mapathons and one of our biggest cleaning/validation issues is non-squared buildings. Having this tool available would save literally thousands of man hours a year and make many of the people who validate/clean new mapper data a lot happier.

slhh:
This is just a modification of the area mode, we should add a have a right-click context menu to the area mode button in order to configure the mode. To make the user aware of the context menu, we can add "(right-click to configure)" to the tooltip of the area mode button. The special configurations of the mode could be indicated by a slightly modified area mode button and/or a toolbar icon.

If desired, there could be controls to set the building mode from the toolbar drawer duplicating the area button context menu function, but the toolbar drawer seems to be the wrong place for the only control of the special area mode because it isn't associated with the area mode.

sisskind:
@bhousel if you are looking to offload this task for iD 2.0, happy to take a look at it.  Unsure if you were still cranking at it or not.

tordans:
I would love to see this feature @bhousel @sisskind.

Some suggestions for the interface:

**Interface goals / interaction goals:**

(a) Introduce a way to draw square blocks as a second option to free form areas.
(b) When drawing a square block a set of tags should be applied to the new area form for the usecase "draw many houses".
(c) The UI should explain (new) users whats going on.

**Wireframe 1 @bhousel** 

See https://github.com/openstreetmap/iD/pull/2699#issue-88903359
- I think the interaction on the drawing board in on point. Especially with the interaction between two shapes (second screenshot).
- I think the buttons could be improved - see below.

**Wireframe 2**

There are two ways to add shapes now, "Area" which can be any form and behave the same way as before and "Block" (naming todo) which behave like Wireframe 1 shows.

One option would be to expose those two button at the same level.

(Just to be sure: This is not a design, the form of buttons are just a scribble. :-))

bildschirmfoto 2017-12-31 um 08 54 02

- Advantage: Give the new feature a lot of attention
- Disadvantage: I feel like its a pro user feature that could be hidden a bit more for those who discover it. Especially since the "Area" tool can be used for the same purpose but with more editing steps.

**Wireframe 3**

Another option would be to introduce a split button (https://getbootstrap.com/docs/3.3/components/#btn-dropdowns-split) that allows to select the secondary editing option. Both edit shapes, so they are in one dropdown but they behave differently.

bildschirmfoto 2017-12-31 um 08 52 41

**Wireframe 4**

Both solutions need to explain (b) and solve (c) which they cannot do without more text since the UI element is nothing people really know from office software and such.

One way to solve it would be to add a information overlay once you click the "Block" tool.

- There is a "hide forever" link to remove the tip
- The text explains the usecase
- The text explains the "copy tags from previous shape" behaviour 
- Everything else should be on a separate help page. Including the click 1, click 2 behaviour since those can be discovered easily by just using the tool.

bildschirmfoto 2017-12-31 um 08 54 08

**How to handle tags**

About (b): How should the Block-Tool handle tags.
1. The same way as area, so its just blank or area=yes?
    - IMO this is fine once there is https://github.com/openstreetmap/iD/issues/1761 to edit all new shapes at once
2. Copy tags from previously selected shape, fallback to the same as the area-tool
     - My favorite ATM, but I think this is best understood in a prototype. This would introduce a new pattern of behaviour into the editor which should be considered carefully. I expect the interaction to be quite complex to grasp so it would only be something for pro users. Which is fine as long as other users have a good experience (expected behaviour, feel in control, helpful/fast) as well.
3. Copy tags from previously selected shape, fallback to building
4. Always apply basic building tags
    - IMO both 3 and 4 are too close to the one use case which I dont like.

bhousel:
> I would love to see this feature @bhousel @sisskind.
> Some suggestions for the interface:

I appreciate the time you spent on those wireframes @tordans but I don't like any of them.  The primary persona that I design for is someone who has never used a map editor before.  We meet a lot of people like this at mapathons.

Adding more buttons to the top bar doesn't scale, and it introduces confusion for new users.  If they have to ask "why would I use block instead of area" then we've failed them.

tordans:
@bhousel Well I guess I am out then, soon. We now touch a level of interaction design discussion which does not lead to results in a comment box - in my experience.

Lets see what other options are proposed that are discoverable, usable and easy to understand.

slhh:
I agree with @bhousel , except that iD should also support the persona of an typical intermediate local mapper well. It doesn't make sense for OSM if users have to switch to a totally different editor (JOSM) in order to program productive. We need to find a way to support both personas well.

Reproducing tags seems to be a must have feature, and we should also have an orthogonal mode according to #3641 and #3637. @edpop Why did you closed these?

The three click drawing mode seems to target mappers, who intend to do large scale edits, and who are hopefully more advanced. I would consider these to be out out scope of iD. Maybe, the 3-click drawing mode is still a nice to have feature for the intermediate users, but it's of limited use because many buildings have L- or T-shape, and it might led to wrong mapping of such shapes.

>Lets see what other options are proposed that are discoverable, usable and easy to understand.

A right click on the mode buttons should open a configuration dialog. The tooltip should make user aware of the right-click dialog.
Instead of right-click dropdown arrows as shown in wireframe 3 might be used.

In case of the area mode, the configuration dialog should contain:

- A checkbox to assign predefined tags to subsequently drawn areas
- An editor for the predefined tags
- A button to clone the tags of the current selected feature to the editor, and there could be a warning when cloning from an non area feature.
- Controls for pretty angles: force multiples of 45° or 90°, invert SHIFT behavior (always force pretty angles, but allow arbitrary angle with SHIFT). These controls can also help to make the SHIFT function discoverable.
- Optionally: checkbox for fast (3-click) rectangle drawing mode.

The user need to be made aware that a special behavior of the mode is active.
Therefore, the button should slightly change, maybe to a diffenent colour, just to indicate that something is special. In addition, when in the draw mode the sidebar should indicate special settings of the mode including any predefined tags to be applied. The user should be able to deactivate any special behavior directly in the sidebar. This also makes it easier for unexperienced users, who have accidentially activated a special setting, to get to get rid of it, even if they don't know where.

edpop:
Hi, @slhh. I waited almost a year for merge or any feedback, so I forked.

homersimpsons:
@slhh Maybe it could support via an "advanced" button (you get the idea), but if it's too hard and time consuming it's better not to think about thoses features. The goal of iD is "just" to bring new users in my mind, experienced users should swith to others editors if they need more tools.
I usually map easy things using iD and complete more complicated things with JOSM. Sometimes just drawing with iD and applysing tags with JOSM. Plus using shortcuts and tag suggestion it can be fast.

slhh:
>Maybe it could support via an "advanced" button (you get the idea)
Using a right-click context menu/dialog to configure, as I have proposed above, is already something like an advanced button. The features would be invisible for new users except of a small hint in the tooltip.
Also the sidebar would be unaffected (empty in the drawing modes) as long as no special mode modification is activated.

>but if it's too hard and time consuming it's better not to think about thoses features.
I would never have requested the 3 click drawing mode, but I think HOT projects would like it. Btw, I would prefer if HOT projects were using JOSM, because they generate mass data.

Replicating tags is a widely requested and agreed on feature #1761. Buildings are an important usecase, but not the only one. Therefore, it needs to be separated from from the building tool.
As iD is intended mainly for less experienced users we should use the safest and most intuitive way to replicate the tags. We should avoid the risk of leaving buildings untagged because of a the workflow of adding many untagged buildings first and appying tags later. And we should also reduce the risk of copying unwanted tags. Tags which are accidentially copied can be hard to discover on the destination object because the all fields section doesn't cover all tags, and all tags section might be collapsed.
Therefore, the user needs to be made aware which tags a copied.

manfredbrandl:
I suggest that user can use a modifier key (e.g. Shift) to change the area mode into a rectangle mode before the placement of the third node. When the modifier Key is pressed after the placement of the second node the rectangle should be shown and the rectangle area should finish with the placement of the third node @bhousel what do you think?

slhh:
@manfredbrandl 
We should not waste the SHIFT modifier to optimize the special usecase of rectangular buildings only.
>Alternatively the modifier key could restrict drawing to rectangular angles only.

When placing the third or any further node, this makes sense. It does also support L or T-shaped buildings, etc.. Maybe, we can allow any multiple of 45° instead of 90° to make it more flexible.

When placing the second node while SHIFT is down, the placed edge should be forced to be rectangular or parallel to one of the edges (of any other way) which are connected to the first node of the area. E.g. to draw a building parallel to the next one.

When clicking the last or first node again to finish the area while SHIFT is pressed, the area should be completed with two othogonal edges using a right angle in the clicked node.

Placing a rectangle would need two normal clicks to place the first and second node, and two SHIFT-clicks on the third node to place it and finish the area. One click more than suggested above, but more flexible.

Pressing SHIFT while placing the first node would not have any meaning. In case a 3-click rectangle drawing mode is really desired, it might be activated by pressing shift while placing the first node.

manfredbrandl:
> Maybe, we can allow any multiple of 45° instead of 90° to make it more flexible.
Multiples of 45° is a good idea.

> ... a 3-click rectangle drawing mode is really desired, it might be activated by pressing shift while placing the first node.
That's not neccesary, starting rectangular at connected area is more important.

When we introduce Shift restricts angled to multiple of 45° we should also allow this for adding nodes by taking the midpoint triangles.

edpop:
@manfredbrandl, it's done in https://github.com/openstreetmap/iD/pull/3641.

MRVDH:
What's the status on this? Still really looking forward to this feature, but it seems inactive.

quincylvania:
> What's the status on this?

@MRVDH We've actually been working on this recently! See #6453. You can test it out on the [preview site](http://preview.ideditor.com/master). It doesn't do everything that this PR does yet, but it will.

MRVDH:
So I assume this is not going to happen any time soon?

quincylvania:
@MRVDH I closed this incidentally by renaming the `master` branch, but it was never going to get merged anyway on account of being almost 5 years stale.

However, I am aiming to get a square drawing tool out this year.

aawiseman:
Checking on this

rmx90210:
A few years later, just +1-ing that this would be awesome to have 😄

PR #2702, opened 2015-06-17
Add recycling_type preset field

### LINKED ISSUE #2689: please add recycling_type option for every amenity=recycling
It is a big difference if the recycling feature is a [centre or just a container](http://wiki.openstreetmap.org/wiki/Key:recycling_type).

Thanks!

ebrelsford:
Fixes #2689

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -739,6 +739,9 @@ en:
       recycling/paper:
         # 'recycling:paper=*'
         label: Accepts Paper
+      recycling/type:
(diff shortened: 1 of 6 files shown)


bhousel:
Thanks @ebrelsford 

I provided strings for the `recycling_type` choices, because it's really just limited to `centre` and `container`, and this way they are translatable.

ebrelsford:
Got it, thanks for merging!

PR #2710, opened 2015-06-26
Some natural/* and man_made/* translations

jmespadero:
Translation for some useful keys for natural/* and man_made/* nodes.

### DIFF
--- a/dist/locales/es.json
+++ b/dist/locales/es.json
@@ -2651,6 +2651,10 @@
                 "name": "Estructura artificial",
                 "terms": "manufactura, industrial, fabricado, manufacturado, fábrica, industria, hecho por el hombre"
             },
+            "man_made/adit": {
+                "name": "Entrada a una mina",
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Please submit these translation corrections via Transifex, our localizations management tool. See [this guide](https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#translating) for instructions. Thank you!

bhousel:
Also, I noticed that you included some translations for things that aren't actually presets in iD yet (adit, monitoring_station, volcano).  Adding them as translations won't make them available as presets - but they would be good additions.  

You can open up a new GitHub issue to have presets added, or you can try to add them yourself.  If you do want to try this, take a look at the [presets README](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md) which explains where all the preset files are stored - it's as simple as just adding the new files and running `make`.

jmespadero:
Thanks, I will try to do both.
Regards

PR #2715, opened 2015-06-29
Some natural and man_made presets (Volcano, saddle, adit, tree_row)

jmespadero:
Hi,

This is a proposal for some new presets in iD, for some natural and man_made features. The preset are accepted in the OSM wiki.

I just uploaded the .json files. I'm not sure if I must run "make" and upload all the changes that the command make to the project folder. By the way... if I run the "make" command, I got the error :

make: *** No rule to make target node_modules/maki/www/maki-sprite.json', needed bycss/feature-icons.css'. Stop.

### DIFF
--- a/data/presets/presets/man_made/adit.json
+++ b/data/presets/presets/man_made/adit.json
@@ -5,7 +5,6 @@
     ],
     "fields": [
         "operator",
-        "resource",
         "depth"
     ],
     "terms": ["entrance",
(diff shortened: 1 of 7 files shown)


bhousel:
Awesome, thank you!   committed as 720b261

`npm install` should fetch the maki file that `make` is looking for.  I'll add a rule so the `Makefile` knows what to do.

PR #2724, opened 2015-07-08
Added landuse=plant_nursery preset

jmespadero:


jmespadero:
Oppps. Travis failed on a field...

PR #2729, opened 2015-07-13
Add Irish postcode field

rory:
Ireland got postcodes (aka 'eircode') today. This (untested) patch adds the "postcode" field to Irish address format using the method recommended.

See the example here: http://www.irishtimes.com/news/ireland/irish-news/eircode-q-a-things-you-need-to-know-about-the-new-system-1.2283200

### DIFF
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -8,7 +8,7 @@
     },
     {
         "countryCodes": ["ie"],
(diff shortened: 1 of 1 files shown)


bhousel:
Just tested it, and it looks good, thank you!

PR #2762, opened 2015-08-29
Change "Access" to "Legal access"

### LINKED ISSUE #2761: Rename "Access" section to "Legal access"
It would reduce confusion of how this tags are applied.

It is common mistake to tag for example bicycle=no because way is not suitable for bicycles (but using it is perfectly legal) or using bicycle=yes to mark way where using bicycle is illegal but infrastructure itself is friendly for cyclists.

See http://wiki.openstreetmap.org/wiki/Key:access "For describing the legal accessibility of a element."

Inspired by https://lists.openstreetmap.org/pipermail/tagging/2015-August/026211.html

M1dgard:
to avoid mappers tagging suitability for a transportation mode instead of legal access.

Fixes #2761.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -71,7 +71,7 @@ en:
           motor_vehicle: Motor Vehicles
       access_simple:
         # 'access=*'
-        label: Access
+        label: Legal access
         # access_simple field placeholder
         placeholder: 'yes'
       access_toilets:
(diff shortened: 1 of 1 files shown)


M1dgard:
@d1g Fixed

bhousel:
Thanks for the contribution, but I decided on "Allowed Access" because "Legal Access" is unclear in English  (see #2761)

In case you want to do this again, to change the caption of a preset field, the better place to make the change like this is in the `data/presets/fields/access.json` file.  Then run `make` and it will generate `data/presets.yaml` and some other files.

Issue #2218, opened 2014-05-11
Level field for features likely to be within buildings


1ec5:
Below the tag presets, there’s a button to set the selected feature’s elevation in the `ele` tag. Because new users tend to map POIs like shops, they are likely to think this field indicates what floor the POI is on. For example, [changeset 22192377](http://www.openstreetmap.org/changeset/22192377) set the elevation of a shop to “1” (1 m) in a city that sits at an elevation of 482 feet.

There should be a separate button for setting the `level` tag for POIs likely to be inside buildings. It might be too confusing to have it replace the elevation button, though.

pavlo-dudka:
It seems that lot of users confuse elevation with level or height. Since it is used quite rarely, I would remove elevation as optional tag.

bhousel:
@pavlo-dudka That does seem odd that a user would enter building levels in the `elevation` field, especially when we provide a visible `building:levels` field for buildings, and the `elevation` field is hidden.  Can you (or somebody who knows) check on the translations for these fields in this area (Ukraine?)

pavlo-dudka:
@bhousel Some ukrainian people filled ele-tag with building height(height-tag), not with building levels. You are right, that was probably done because of incorrect translation of *elevation* to ukrainian.

Incorrect understanding of elevation also occured in USA. It was confused it with level(shop level above the ground floor). See initial post of @1ec5.
http://www.openstreetmap.org/node/2842085937/history

1ec5:
The user who incorrectly put the level in `ele` on that building was likely using iD in English. I think they probably were aware that the `building:levels` tag was for the total number of floors. Being a new mapper who was mainly there to add Bitcoin tags, they wouldn't've known about the ability to specify custom tags.

In the U.S. at least, the floor can be considered an informal part of the address, so that would be a good place for a `level` tag. But we'd have to account for regional differences about how to number the ground floor.

bhousel:
I added this as a 'universal' field, so mappers can add this to any feature that happens to be indoors.

PR #2765, opened 2015-09-01
Add field "level" to preset vending_machine

henningvs:
I would suggest to add the field level to the amenity preset, as many vending_machines are in buildings at a defined level.

### DIFF
--- a/data/presets/presets/amenity/vending_machine.json
+++ b/data/presets/presets/amenity/vending_machine.json
@@ -1,7 +1,8 @@
 {
     "fields": [
         "vending",
-        "operator"
+        "operator",
+        "level"
     ],
     "geometry": [
         "point"


bhousel:
I don't think this is necessary.  Anyone can already add the 'level' field just by using the "Add field" dropdown below all of the other fields.  It's a universal field, just like "address", "phone", "wheelchair", "wikipedia" - so it is always available for any feature.

henningvs:
I agree to this. Level is in fact an universal category. I would still suggest to show it directly to the user, as it is a lot easier to identify than for example an address. 
Also it is quite annoying to select the field every time. Right now it's not even showing up in the dropdown of "add field". Is this probably a bug?

bhousel:
> Also it is quite annoying to select the field every time. Right now it's not even showing up in the dropdown of "add field". This might be a bug?

It was added in 58cb66f for #2218
I'll try to make a release this week so that it shows on the main openstreetmap.org site.

PR #2772, opened 2015-09-05
Fix "terms" translations (closes #2756)

### LINKED ISSUE #2756: Terms preset does not work properly

M1dgard:
"terms" lists that are not fully lowercase or used combinations of whitespace and commas as separators (instead of just commas) could not be handled properly. This PR fixes that.

Improved version of #2767 after feedback there.

### DIFF
--- a/js/id/presets/preset.js
+++ b/js/id/presets/preset.js
@@ -46,7 +46,7 @@ iD.presets.Preset = function(id, preset, fields) {
     };
 
     preset.terms = function() {
-        return preset.t('terms', {'default': ''}).split(',');
+        return preset.t('terms', {'default': ''}).toLowerCase().split(/\s*,+\s*/);
     };
 
     preset.isFallback = function() {


bhousel:
Thanks!  Making the search terms case insensitive also fixes a bunch of search terms in English that I didn't realize were not working:   "WC, EV, Supercharger, God, Elsan"

bagage:
There are still one bug because it does not trim `terms` white-spaces (see output of `grep 'terms": " ' . -r` and `grep 'terms": ".* "' . -r`), so all these bad formatted strings will still have issues.

By the way lowercasing `a` and `b` is useless [at line 38](https://github.com/openstreetmap/iD/blob/master/js/id/presets/collection.js#L38) since there are already lowercased.

PR #2774, opened 2015-09-08
Make raw tag editor show docs for key=value

### LINKED ISSUE #2754: link to page descriping tag on OSM wiki if able (currently it leads to page describing key)
see for example https://www.openstreetmap.org/edit?editor=id#map=18/52.61392/24.36614 and select object.

In section "All tags" select info button (with "i" icon). 

"View on OpenStreetMap Wiki" links page describing key, even for tags where there is a separate page for tag value.

For example for landuse=forest http://wiki.openstreetmap.org/wiki/Tag:landuse%3Dforest is preferable over current target - http://wiki.openstreetmap.org/wiki/Key:landuse .

M1dgard:
instead of just key.

* Pass the value to iD.ui.TagReference in the `context` function of iD.ui.RawTagEditor
* Let iD.taginfo query the docs for the key when its callback returns false.
* Let the callback function defined inside iD.ui.TagReference return false when both of these are met:
  * it isn't happy with the docs results AND
  * it was called with its softfail argument set to true.

Closes #2754.

### DIFF
--- a/js/id/services/taginfo.js
+++ b/js/id/services/taginfo.js
@@ -122,8 +122,21 @@ iD.taginfo = function() {
         if (parameters.value) path = 'tag/wiki_pages?';
         else if (parameters.rtype) path = 'relation/wiki_pages?';
 
+        var decoratedCallback;
+        if (parameters.value) {
+            decoratedCallback = function(err, data) {
+                // The third argument to callback is the softfail flag, to
+                // make the callback function not show a message to the end
+                // user when no docs are found but just return false.
+                var docsFound = callback(err, data, true);
+                if (!docsFound) {
+                    taginfo.docs(_.omit(parameters, 'value'), callback);
+                }
+            }
+        }
+
         request(endpoint + path +
-            iD.util.qsString(parameters), debounce, callback);
+            iD.util.qsString(parameters), debounce, decoratedCallback||callback);
     };
 
     taginfo.endpoint = function(_) {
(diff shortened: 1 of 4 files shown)


js/id/services/taginfo.js
```diff
@@ -122,8 +122,21 @@ iD.taginfo = function() {
...
+        if (parameters.value) {
+            decoratedCallback = function(err, data) {
+                // The third argument to callback is the softfail flag, to
+                // make the callback function not show a message to the end
+                // user when no docs are found but just return false.
+                var docsFound = callback(err, data, true);
+                if (!docsFound) {
+                    taginfo.docs(_.omit(parameters, 'value'), callback);
+                }
+            }
```
bhousel: need ; here

js/id/services/taginfo.js
```diff
@@ -122,8 +122,21 @@ iD.taginfo = function() {
...
+        }
+
         request(endpoint + path +
-            iD.util.qsString(parameters), debounce, callback);
+            iD.util.qsString(parameters), debounce, decoratedCallback||callback);
```
bhousel: please add spaces around ||

bhousel:
This works really well!  Thanks for fixing the style issues so quickly.

Can you also add another test to `#docs` [around here](https://github.com/openstreetmap/iD/blob/522e593f112d3c628604a54a2d0ca0bae11dc5c3/test/spec/taginfo.js#L106) so that callback is tested with and without `softfail`..

M1dgard:
No problem. :)
I added a test case.

tyrasd:
There's a minor issue after this merge: When one opens the documentation of a particular tag in the taglist – say `amenity=parking` – and then immediately selects another object with a tag of the same key but different value – say `amenity=restaurant` – the taglist will still show the documentation of the former tag:

id-taginfo-tags o3

PR #2777, opened 2015-09-14
preset.js: trim terms when retrieving translations to avoid white spaces

bagage:
Terminates #2756 and PR #2772 - trimming terms before using them is needed for some translations where whitespaces are inserted at string start/end. Note that function `leading` tend to handle that but not for everything actually (only spaces).


bhousel:
Thanks for this contribution, looks good..

Issue #2766, opened 2015-09-01
Incomplete preset translations are a Really Bad Thing™


M1dgard:
Okay, incomplete core translations look unprofessional, but they can be used.

On the other hand, preset translations where the labels have not all been translated, result in a very frustrating editing experience, potentially scaring away new mappers.

Consider this scenario:

You use iD in Norwegian (`no`) and search for a swimming pool by typing the Norwegian word `svømmebasseng`. You find what you want to map.

But if you want to map a vending machine, and search for `salgsautomat`, you don't find anything.

You decide to try the English term `vending machine` and find what you are looking for, because iD obviously falls back to English if no translation is present yet.

Now you may have realised that the translation is not complete<sup>[note 1]</sup>, and that you will not find everything in Norwegian. Then you want to map a town hall and based on this experience you decide to search for the English term right away. **However, you won't find the preset**, because the label has already been translated (`Rådhus`).

It was upon realizing this that I completed iD's translation to Dutch<sup>[note 2]</sup>. Unfortunately, I am not sufficiently fluent in Bosnian, Chinese (Yue), Korean, Chinese (Hong Kong), Arabic, Greek, Persian, Turkish, Serbian, etc to translate to those languages as well, and I don't know if and when they will be completely translated.

___

**Notes**

1. Not everyone may realise this. It would be useful to display some warning somewhere if not all labels are translated.
2. The attentive reader may note that Dutch has a translation completeness of 86% on Transifex. This is because of the terms lists, which were deliberately left blank if I could not think of synonyms and related terms.

bhousel:
I agree that they are a Really Bad Thing.  Is there anything actionable we can do about it?  Maybe to improve `Rådhus` we could also include the English tag value 'town_hall' in the terms list? 

I did send an update on Transifex to the translators today to let them know we are releasing soon.  We need to do a better job keeping them engaged, I'm not sure best way to do it.

M1dgard:
Well, it's a lot of work if you're alone. If you're not, you have to make some agreements on how you will translate or you will end up with am incoherent translation. This is basically out of your control of course. The glossary function on Transifex can be a huge help.

Firstly I'd not include translations with too low a completion rate (say 62%). You'd need to know English anyway if you wanted to use them.

Secondly, I'd just include all English terms and labels as well in searches, be it with a somewhat lower priority rate. (Is that feasible?)

M1dgard:
[...]

PR #2779, opened 2015-09-15
Languages

kaipedia:
Hallo! Using iD in OpenStreetMap I only found attributes etc. in english. 
Is it possible to define different languages in iD so each user can select, which one he/she wants to use?
Example: I had to learn to use "allotments" to create a garden named "Kleingartenanlage" in german OpenStreetMap...
Thanks Kaipedia

### DIFF
--- a/languages
+++ b/languages
@@ -1,0 +1,4 @@
+Hallo! Using iD in OpenStreetMap I only found attributes etc. in english. 
+Is it possible to define different languages in iD so each user can select, which one he/she wants to use?
+Example: I had to learn to use "allotments" to create a garden named "Kleingartenanlage" in german OpenStreetMap...
+Thanks Kaipedia


bhousel:
Hi @kaipedia, there are a few things going on with languages in iD, hopefully this helps clear things up:

1. iD tries to detect your language based on which languages set up in your browser preferences.
2. The detected language can be overridden by JavaScript.  For example, the iD that is embedded in openstreetmap.org will get your openstreetmap language preference passed in, and iD will try to honor it.
3. As you know all the languages aren't translated completely.  We are tracking ways to improve this on issue #2766.  So it's definitely possible that "Allotments" is not translated well to German.  (oddly, we call them "Community Gardens" in the US, and most people here would not understand the term "Allotments" either.)

If you see anything in iD that is missing a translation or translated improperly, check out our project on [Transifex](https://www.transifex.com/organization/ideditor/dashboard/id-editor) and you sign up for an account there and make the improvement.

Issue #2406, opened 2014-10-18
"comment" URL pre-fill param fails to work if comment contains equals sign (%3D)


danstowell:
Thanks for adding the URL parameter to pre-fill changeset comment (commit 21da76cf8060832e36, closing issue #2311). It works, but it seems to fail when there's an equals sign included in the comment (URL-encoded as %3D) ... which happens in quite a lot of HOT Tasks!

Here's the URL which I originally discovered the issue in, a URL constructed by the HOT Tasking Manager: http://www.openstreetmap.org/edit?editor=id&#map=13/-0.968384792013/20.7398745496&comment=DRC%2C+%23hotosm-ebola-drc-661%2C+%23MapGive%2C+source%3DWorldView-2%2C+DigitalGlobe%2C+NextView%2C+8+March+2011+or+16+September+2014&gpx=http://tasks.hotosm.org/project/661/task/48.gpx&background=custom:http://hiu-maps.net/hot/1.0.0/boende-2-flipped/{z}/{x}/{y}.png

The `comment` param is quite long. So I experimented to find a minimal reproducer and I found the following:

* This works: http://www.openstreetmap.org/edit?editor=id#map=19/51.56473/-0.04403&comment=new%20changeset%20comment%20taken
* This works: http://www.openstreetmap.org/edit?editor=id#map=19/51.56473/-0.04403&comment=new%20changeset%20comment%23taken (it has a URL-encoded hash sign in the comment)
* This fails: http://www.openstreetmap.org/edit?editor=id#map=19/51.56473/-0.04403&comment=new%20changeset%20comment%3Dtaken (it has a URL-encoded equals sign in the comment)

Here, "works" means that if you make simple edit and press save, you see the desired comment pre-filled in the text box. "fails" means you just see the last comment you used in iD, whatever that was.

(BTW, I'm using Firefox v33)

jfirebaugh:
The Firefox bug will be fixed in FF 41: https://www.fxsitecompat.com/en-US/docs/2015/urlutils-hash-no-longer-decodes-fragment/.

jfirebaugh:
Browser fix has landed; this is now working in FF.

PR #2781, opened 2015-09-15
Allow for equals in background hash

danielfdsilva:
In a recent use case we needed to have query parameters for the background. Something like:

```
background=custom:http://a.tiles.mapbox.com/v4/astrodigital.LC81170532015080LGN00_bands_432/{z}/{x}/{y}.png?access_token=token_here
```
If you insert this string as a custom background source it works flawlessly but when it needs to be parsed from the url it breaks.

This happens because the `.split('=')` would convert this string to a 3 element array and that doesn't meet the criteria. This PR fixed the problem by allowing everything after the first `=` sign.

### DIFF
--- a/js/id/util.js
+++ b/js/id/util.js
@@ -40,7 +40,8 @@ iD.util.displayType = function(id) {
 
 iD.util.stringQs = function(str) {
     return str.split('&').reduce(function(obj, pair){
-        var parts = pair.split('=');
+        var i = pair.indexOf('=');
+        var parts = i === -1 ? pair :[pair.slice(0, i), pair.slice(i + 1)];
         if (parts.length === 2) {
             obj[parts[0]] = (null === parts[1]) ? '' : decodeURIComponent(parts[1]);
         }


jfirebaugh:
I believe if you URL-encode the `background` value, i.e. use `%3d` instead of `=`, this won't be necessary. Can you try that?

danielfdsilva:
That won't do it. The url gets decoded before reaching the function.
When you use `location.hash` the url is automatically decoded.

jfirebaugh:
Ah, that's the Firefox bug described in #2406. Looks like there's been some movement on that and it will be fixed in FF 41: https://www.fxsitecompat.com/en-US/docs/2015/urlutils-hash-no-longer-decodes-fragment/

danielfdsilva:
Ok, but shouldn't you be looking at supporting "older" browsers as well?

jfirebaugh:
For "evergreen" (autoupdating) browsers like Chrome and Firefox, we only support the most recent release -- supporting out-of-date releases when new ones are coming on a monthly basis would be prohibitively expensive.

danielfdsilva:
Ok, fair enough. Closing

PR #2784, opened 2015-09-18
Added space bar "click" when drawing ways

brandonreavis:
This makes the space bar function like a mouse click when drawing a way. When drawing long ways like roads it is so much more efficient because you don't have the mouse click stopping the motion of your mouse.

### DIFF
--- a/js/id/behavior/draw.js
+++ b/js/id/behavior/draw.js
@@ -1,6 +1,7 @@
 iD.behavior.Draw = function(context) {
     var event = d3.dispatch('move', 'click', 'clickWay',
-        'clickNode', 'undo', 'cancel', 'finish'),
+        'clickNode', 'undo', 'cancel', 'finish', 
+        'spacedown', 'spacedownNode', 'spacedownWay'),
         keybinding = d3.keybinding('draw'),
         hover = iD.behavior.Hover(context)
             .altDisables(true)
@@ -8,7 +9,8 @@ iD.behavior.Draw = function(context) {
         tail = iD.behavior.Tail(),
         edit = iD.behavior.Edit(context),
         closeTolerance = 4,
-        tolerance = 12;
+        tolerance = 12,
+        lastDatum = {};
 
     function datum() {
         if (d3.event.altKey) return {};
@@ -55,7 +57,8 @@ iD.behavior.Draw = function(context) {
     }
 
     function mousemove() {
-        event.move(datum());
+        lastDatum = datum();
+        event.move(lastDatum);
     }
 
     function click() {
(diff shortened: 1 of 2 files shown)


brianreavis:
@bhousel Can we get some activity on this?

bhousel:
I like the idea - will test over the next few days..

bhousel:
Just a quick followup, I'm going to cleanup and merge a version of this soon.  I think I'm going to make the spacebar work like a left click anywhere on the map, because it would be too confusing to have it work like a click only in certain modes.

slhh:
I believe making the spacebar a duplicate of left-click would be confusing too because of different behavior when iD has lost focus.

In addition I think the spacebar should be reserved for additional functions, but not wasted as a pure duplicate of the left click.

Drawing a line the spacebar could add nodes like the left click, but without connecting to exiting ways or nodes. This is a dual-purpose function. Sometimes the user needs unconnected nodes/ways and it is useful for the application of placing nodes during mouse movement as requested. Nodes being set during mouse move would connect to existing objects only at random anyway because of position tolerance.

If you wan't to avoid accidently unconnected line ends you might prohibit using spacebar click for the starting and ending click of the line. If the user wants to have an unconnected start or end node he can place it using the space bar, but place an additional node with left click elsewhere for starting or ending the line. The user has to delete this additional node later. Using the mouse for the first node would also prevent the issue of lost focus. ID losing the focus during drawing the line seems to be very unlikely.

Being a modified click the spacebar working only in certain modes should not be too confusing so long as other modes don't need such modified click.

bhousel:
> Sometimes the user needs unconnected nodes/ways and it is useful for the application of placing nodes during mouse movement as requested.

Alt/Option key already does this - when pressed, snapping to existing objects is disabled.
(It's not very obvious and I didn't know this for a while)

slhh:
>Alt/Option key already does this - when pressed, snapping to existing objects is disabled.
(It's not very obvious and I didn't know this for a while)

Before commenting I had already tried the typical click modifiers Shift, Ctrl, and Alt alt without success.
This was likely due to a nasty interference with Firefox (tested on Windows  10) menu toggling by Alt.

Using Firefox full sceen mode (by F11) it works well. 

Otherwise, if Firefox menu is hidden it works too, but it behaves somewhat nasty if you decide not to click after pressing Alt down. In this case the menu is switched on and the map is shifted down on Alt key release.

The worst case is with Firefox menu already been shown. In this case the the menu is already switched off on the Alt key down event, shifting the map up under the cursor before you can click. 
Your intended function still works after repositioning the mouse, but this is difficult to discover because users don't expect a second function to be bound to the same key.

I had also check the shortcut list http://wiki.openstreetmap.org/wiki/ID/Shortcuts, which is linked from iD's help. The Alt-key wasn't mentioned there, but I have added it yet. Please can you check if some other hidden key functions of iD exist, which are not mentioned there? Shortcuts being shown in iD don't matter, but totally hidden functions like the Alt key would be nice to know.

bhousel:
Thanks @brianreavis for getting this started, and I'm sorry I took so long for me to merge it..  This PR exposed some gnarly issues with in iD's keybinding code, but I was finally able to spend a few days to clean those up, and I'm really happy with how it turned out.

Now in draw mode, spacebar will simulate clicks.  It's fun because you can either discretely tap to add points, or just hold it down and drag around.  There is a small distance threshold so the points won't stack on top of each other.

space_click

brandonreavis:
@bhousel Beautiful! I'm excited to try it out. It should be a big time-saver!

PR #2798, opened 2015-10-07
Fetch proper tag-references for relation types in raw tag editor

### LINKED ISSUE #2797: More relevant wiki link for type=multipolygon preset
Still present in openstreetmap.us/iD/master/:
https://wiki.openstreetmap.org/wiki/Key:type - current link won't help them
https://wiki.openstreetmap.org/wiki/Relation:multipolygon - better link with images and translations

tyrasd:
Implements special casing of relation type tags as described in https://github.com/openstreetmap/iD/issues/2797#issuecomment-145907014. Fixes #2797

### DIFF
--- a/js/id/ui/raw_tag_editor.js
+++ b/js/id/ui/raw_tag_editor.js
@@ -85,7 +85,12 @@ iD.ui.RawTagEditor = function(context) {
         $items.order();
 
         $items.each(function(tag) {
-            var reference = iD.ui.TagReference({key: tag.key, value: tag.value}, context);
+            var isRelation = (context.entity(id).type === 'relation'),
+                reference;
+            if (isRelation && tag.key === 'type')
+                reference = iD.ui.TagReference({rtype: tag.value}, context);
+            else
+                reference = iD.ui.TagReference({key: tag.key, value: tag.value}, context);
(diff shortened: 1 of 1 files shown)


js/id/ui/raw_tag_editor.js
```diff
@@ -85,7 +85,11 @@ iD.ui.RawTagEditor = function(context) {
         $items.order();
 
         $items.each(function(tag) {
-            var reference = iD.ui.TagReference({key: tag.key, value: tag.value}, context);
+            var reference;
+            if (id[0] === 'r' && tag.key === 'type')
```
tyrasd: I'm a bit unsure about inferring the object's type from it's id – feels a bit hacky…

pnorman: Well, it's ID is just a number, if the `id` variable starts with r it is encoding both object type and ID.

bhousel: @tyrasd I'm ok with what you did, it's clear to me.  

I guess if you want to make the code a little clearer, you can do (before the loop)
```js
var isRelation = (context.entity(id).type === 'relation');
```

PR #2809, opened 2015-10-18
Disable joining ways with conflicting tags

### LINKED ISSUE #2358: iD should not combine ways with different tags
Way to reproduce (example):
- search for two connected highway=track with different tracktype (for example tracktype=grade2 and tracktype=grade3)
- select them and combine them
- the result in iD is one way that has tracktype=grade2;grade3

I expect that iD can not combine these ways, or give the possibility for choosing one of the values of tracktype (like in JOSM).

Another example:
- search for two ways, of which one is part of a relation
- select both ways and combine them
- the result is that the new combined way (actually both ways!) are part of the relation

I expect the same as stated above: iD should not give the possibility to combine these ways, simply because they are different.

For some time I had much work to restore ways and relations that were combined with iD and now had double values in keys. Restoring the ways was relatively easy, reverting the relations was much more difficult because it is not easy to see to which way the relation belonged to. I think I even couldn’t restore all the relations, so good information was destroyed because of the incorrect edits with iD!

I think that the best solution for this problem is to make the + (combine) icon not available when ways have different tags or are different member of a relation (select two not connected ways to see what I mean with the + icon: the icon is there, but not available). If the combine icon is hidden and newbies (or other iD users) see the message that it isn't possible to combine the ways, I think it doesn't matter them much. In any case it is better than damaging the data! Besides that, because of the aimed simplicity of iD, I think it isn't very important that in some situations ways couldn't be combined. Indeed this makes iD more simple!

jfirebaugh:
Fixes #2358.

"Uninteresting" tags are exempted -- this includes "attribution", "created_by", "source", "odbl", and "tiger:*". The values of such tags will continue to be joined with semicolons.

The rule also takes into account tag values that will automatically be reversed when the ways are joined. For example, if you have a `oneway: yes` way from a to b, and a `oneway: -1` way from c to b, joining is permitted, and will produce a `oneway: yes` way going a to b to c.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -114,6 +114,7 @@ en:
       not_adjacent: These lines can't be merged because they aren't connected.
       restriction: These lines can't be merged because at least one is a member of a "{relation}" relation.
       incomplete_relation: These features can't be merged because at least one hasn't been fully downloaded.
+      conflicting_tags: These lines can't be merged because some of their tags have conflicting values.
     move:
       title: Move
       description: Move this to a different location.
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -143,7 +143,8 @@
             "not_eligible": "These features can't be merged.",
             "not_adjacent": "These lines can't be merged because they aren't connected.",
             "restriction": "These lines can't be merged because at least one is a member of a \"{relation}\" relation.",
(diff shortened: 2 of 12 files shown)


bhousel:
Thanks @jfirebaugh!

PR #2810, opened 2015-10-19
better support for browser language preferences

kriscarle:
Use navigator.languages preferences when available. This allows a preferred language to be set in Chrome without needing to change the primary language of the OS.  This appears to only apply when iD is running outside of openstreetmap.org, as we are doing for the loggingroads.org project. When it is framed inside osm, the osm user profile language settings appear to override the browser settings.

also see: https://alicoding.com/detect-browser-language-preference-in-firefox-and-chrome-using-javascript/ and http://www.w3.org/html/wg/drafts/html/master/webappapis.html#dom-navigator-languages


bhousel:
Thanks for this fix!  I didn't realize Chrome did a different thing here.

PR #2814, opened 2015-10-22
Corrected some brands to match ad copy

zeromap:
Removed Hess gas station
The Hess brand is no longer used in retail fuel sales. Signs may still exist, but are being converted to Speedway brand.
http://www.usatoday.com/story/money/business/2014/05/25/hess-stations-name-change/9445589/

Dunkin' Donuts
http://news.dunkindonuts.com/

Lowe's
http://media.lowes.com/about-lowes/

Baskin-Robbins
http://news.baskinrobbins.com/about

### DIFF
--- a/data/name-suggestions.json
+++ b/data/name-suggestions.json
@@ -115,9 +115,6 @@
             "Shell Express": {
                 "count": 131
             },
-            "Hess": {
-                "count": 127
-            },
             "Flying V": {
                 "count": 129
             },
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks!  But we pull this file from the [name-suggestion-index project](https://github.com/osmlab/name-suggestion-index#contributing) automatically.  You will need to submit these changes there, and then it can be included in the next release of iD.

zeromap:
That makes sense, thanks!

PR #2819, opened 2015-10-27
Correct iD.Way test to use Way() not Entity()

gmaclennan:
Whilst going through the tests I noticed that the core/way.js test for #copy is using Entity() for some reason rather than Way(). I'm pretty sure this should be Way().


bhousel:
Yes you are right.. Thank you!

Issue #2813, opened 2015-10-21
Locator overlay in minimap sometimes misaligned


tyrasd:
After panning around a bit by panning directly in the minimap, I got to a state where the locator overlay is largely misaligned from the minimap's content, see below:

![selection_083](https://cloud.githubusercontent.com/assets/1927298/10639085/91cd2cf2-780e-11e5-9555-f7170b0dd534.png)

(tested in Firefox and Chromium)

brianreavis:
To reproduce: toggle any overlay. This causes `map-in-map-overlay` to be placed *after* `map-in-map-gpx` (which has the block svg in it, which bumps the overlay layer down the height of the SVG)

PR #2829, opened 2015-11-10
Support for multiple overlays in minimap + fix for #2813

brianreavis:
Refs https://github.com/openstreetmap/iD/issues/2813

### DIFF
--- a/js/id/ui/map_in_map.js
+++ b/js/id/ui/map_in_map.js
@@ -4,9 +4,9 @@ iD.ui.MapInMap = function(context) {
     function map_in_map(selection) {
 
         var backgroundLayer = iD.TileLayer(),
+            overlayLayers = {},
             dispatch = d3.dispatch('change'),
             gpxLayer = iD.GpxLayer(context, dispatch),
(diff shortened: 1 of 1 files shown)


js/id/ui/map_in_map.js
```diff
@@ -159,37 +159,39 @@ iD.ui.MapInMap = function(context) {
...
 
             overlay.enter()
                 .append('div')
                 .attr('class', 'map-in-map-overlay');
 
-            overlay.exit()
-                .remove();
+            var overlays = overlay
+                .selectAll('div')
+                .data(activeOverlayLayers, function(d) { return d.source().name(); });
```
bhousel: Does it need to be keyed on name?   (I don't think it's bad, just not sure it is necessary).

brianreavis: See https://github.com/openstreetmap/iD/blob/master/js/id/renderer/background.js#L67-L68

bhousel: Cool, makes sense to keep it working the same way the main map keys them.

PR #2838, opened 2015-11-21
Fix iD.detect when navigator.languages is empty.

jleedev:
Fix iD.detect when navigator.languages is empty.


bhousel:
Thank you!

Issue #2826, opened 2015-11-06
Suggest alternate imagery based on map location


planemad:
Recent investigations into the [data issues in Japan](https://www.openstreetmap.org/user/PlaneMad/diary/36058) revealed the Bing imagery is poorly orthorectified with both distortions and offsets that vary throughout the country. Data traced from Bing requires correction which is a very resource intensive effort.

New mappers who join OSM are completely unaware of these issues and start putting in a lot of effort to essentially contribute data that requires rework, which is a massive waste of community resources. Another issue is new users realigning correct data to incorrect imagery.

If iD can have a database for alternate imagery to be used for different regions, we could prompt the mapper to use the correct background layer. Otherwise there is no mechanism to inform the user who is bound to be confused on why there is a mismatch between the imagery and data.

![](https://cloud.githubusercontent.com/assets/126868/10997540/dd8e612c-84b3-11e5-969f-2e0e53a29fed.gif)
*Imagery distrotions in Bing*

cc @bhousel @lxbarth @mikel

lxbarth:
> If iD can have a database for alternate imagery to be used for different regions, we could prompt the mapper to use the correct background layer. 

Such a db should be available to any editor. See also https://github.com/osmlab/editor-imagery-index cc @tmcw

bhousel:
There is a very old issue for this in #1124 .. @Zverik what's the status of this?  I assume the Imagery Offset Database isn't "upcoming" anymore?

Agree that this really needs tie-in with `editor-imagery-index` so that we can pick the best imagery for the user's loc/zoom.  (yes, zoom matters)

Zverik:
The offset database does not provide alternative imagery information, but it is very active for 2.5 years with more than 11 thousand offsets recorded.

bhousel:
Hey @Zverik I did some looking through the [API documentation](http://wiki.openstreetmap.org/wiki/Imagery_Offset_Database/API) and it looks like a decent initial attempt at recording offsets.

Some questions:
* Can you make the source available for the server component?
* Will you continue to host this server, or does it make more sense for OSM's OWG to host it somewhere (obviously if iD starts auto-adjusting all its imagery based on your offset database, it could generate a ton of traffic)
[...]

Issue #1310, opened 2013-04-15
Per country default background


Firefishy:
For South Africa it would be great if iD used http://aerial.openstreetmap.org.za/ [ngi-aerial] as the default background layer for South Africa.

ngi-aerial has 100% coverage of South Africa at 50cm resolution. Positional accuracy is consistent good to excellent.

Bounding polygon can be found here: http://josm.openstreetmap.de/maps under "South Africa CD:NGI Aerial" entry.

PR #2841, opened 2015-11-24
Prefer imagery marked as `best` in editor-imagery-index

bhousel:
* Sorts the `best` imagery at the top of the background list
* Pick it automatically if the url hash doesn't specify an imagery source
* Also bump "none" to the bottom of the list
* re: #2826 #1310

cc @pnorman @planemad @lxbarth thoughts?

Before|After
--- | ---
#map=14.96/8.2276/47.5272 | #map=14.96/8.2276/47.5272
screenshot 2015-11-24 11 18 27|screenshot 2015-11-24 11 18 16

### DIFF
--- a/data/imagery.json
+++ b/data/imagery.json
@@ -2625,6 +2625,660 @@
         "terms_text": "Ortofoto public domain ÚHUL, year 2000"
     },
     {
+        "name": "FR-BAN",
+        "type": "tms",
+        "description": "French address registry or Base Adresses Nationale",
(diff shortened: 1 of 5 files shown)


planemad:
@bhousel awesome! does it switch by default to the best imagery on opening iD for an area?

bhousel:
Yes it will try to pick the best one - as long as the url includes `map` coordinates and doesn't include something like `background=Bing`

planemad:
:+1:

Issue #2827, opened 2015-11-06
Add different types of vending machines as presets


henningvs:
As the title states I wanted to ask if it is possible to add the different types of vending machines to the presets (I would also do it myself). I did not find anything about this in https://github.com/openstreetmap/iD/blob/master/data/presets/README.md so I thought asking your opinion about this first would be the best idea before doing something.

What I would like to do is to create a preset for every different kind of vending machine (seperated by the goods they are vending) as we have here in Germany (and I think in Japan also) a lot of specialized vending machines. For example there are already more than 10.000 cigarette vending machines (vending=cigarettes) taged.

What do you think about this idea?

bhousel:
> What I would like to do is to create a preset for every different kind of vending machine (seperated by the goods they are vending) as we have here in Germany (and I think in Japan also) a lot of specialized vending machines.

How many is "every kind"?

I guess I could see doing this for a few categories ("snacks", "soda", "cigarettes", "tickets")
but the [OSM wiki lists quite a lot of them](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dvending_machine), and I wouldn't want to see a preset for every uncommon kind of vending machine.

[Taginfo shows](http://taginfo.openstreetmap.org/keys/vending#values) that only a few types have gained a lot of traction.. (surprised to see `excrement_bags` so high on that list).

henningvs:
Yes I agree, it would make sense to just add the most common categories. Maybe everything which is tagged more than 1000 times? 

I guess excrement_bags is so high on the list because you are "forced" by german law to pick up the excrement of your dogs (or other animal) when you are in a public place.
Thanks for the quick response!

PR #2842, opened 2015-11-25
New presets for the different types of vending machines

henningvs:
This is a pull request regarding Issue #2827.


bhousel:
Thanks!

Issue #2635, opened 2015-05-08
Two background layers at once and image slider tool


cgiovando:
It would be nice to be able to add two background layers at once and have an image slider tool to swipe between the images. The main use of this feature would be for before and after comparison when two imagery layers of the same areas are available for different moments in time. This would allow the user for quick change detection visual analysis and understand what new features could be mapped.

The feature would be implemented either with a "swiping" handle or by flickering (rapidly switching) between the two background images with a keyboard shortcut.

Here is an [example](http://demosthenes.info/blog/819/A-Before-And-After-Image-Comparison-Slide-Control-in-HTML5) of what I mean by image slider.  It is also a feature used or [wished for](http://gis.stackexchange.com/questions/87770/qgis-quickly-switching-overlapping-raster-images) in GIS and image analysis software (see [swipe](http://resources.arcgis.com/en/help/main/10.1/index.html#/Interactively_reveal_rasters_using_the_Swipe_Layer_tool/009t00000099000000/) and [flicker](http://resources.arcgis.com/en/help/main/10.1/index.html#/Revealing_rasters_using_the_Flicker_tool/009t0000009m000000/) methods in ArcGIS).

bhousel:
We didn't build a slider tool, but #2854 provides the ability to quickly switch between 2 backgrounds.

Issue #2095, opened 2014-01-09
Add method for switching between two custom tile layers


oiva:
Use case: using two different custom tile layers for mapping. For example, one map layer and one aerial imagery layer.

Now switching between two custom layers requires you to first change the background to some other layer and then back to custom layer, entering the URL for the second custom tile layer.

One option that might solve this (and #2094) would be to add new custom layers to the layer stack, retaining the "Custom" option, which could then be used to add another custom layer.

Another way to solve this could be having two separate custom layer options and make  both of them remember their last value.

bagage:
More generally, it would be convenient if we could add as many custom layers as we wish: in France there are at least 3 rendering that might be useful at the same time (IGN aerial photos, cf #3420, Strava heatmap and [BANO](http://tile.openstreetmap.fr/~cquest/leaflet/bano.html)). 
It would mean a bit of UI reworking, but maybe something like that?

![capture d ecran de 2017-02-22 19-07-22](https://cloud.githubusercontent.com/assets/1451988/23225900/ea0b8830-f933-11e6-9fc3-d3761a633d6b.png)

Abbe98:
Reposting my comment from #3888, 

> Save all of the past ones in localstorage and display a dropdown in the dialog?
> 
> A new dialog window would be needed, is there any existing element that could be reused?

verdy-p:
How would you customize the transparency if multiple layers are used ? They are not necessarily transparent.
Opaque non-transparent layers (such as Mapnik) would be exclusive of each other (radio buttons) unless there's an additional slider to set their respective transparency.
Some layers may need to be tuned also to implement a filter (such as converting white to transparent with some conversion curve from lightness to alpha, and inverting lightness), in advanced options.

As well the possibibilty of dragging up or down layers in list (or the use of "up/down" arrow buttons) could easily switch the order of opaque RGB layers (with no alpha channel).

You can see these in effect for example on the (non-free) French GeoData portal (from IGN) or the Swiss Geoportal, or in Mapbox.

Still we should keep the interface simple by enabling these only with minimum interface when we have a single layer: radio buttons are useful to switch between single opaque background layers, and a last "custom" radio button would have the possibility of compositing one or more layers with various viewing options for each of them and their respective order. In that case a single "Custom" radio button will be needed, or jsut a single "+" button to add a new selectable view where we can compose the layers as wanted and switch between multiple composite views or with preenabled opaque views.

verdy-p:
[...]

PR #2854, opened 2015-12-02
Add 'I' shortcut for switching between the last two backgrounds.

RoPP:
It's sometime convenient to quickly switch between two backgrounds, this PR adds a 'S' shorcut for this purpose.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -1808,6 +1808,10 @@ div.full-screen > button:hover {
     background: #E8EBFF;
 }
 
+.layer-list li.switch {
+    background: #fffce8;
+}
+
 .layer-list label {
     display: block;
     padding: 5px 10px;
--- a/js/id/ui/background.js
+++ b/js/id/ui/background.js
@@ -8,7 +8,8 @@ iD.ui.Background = function(context) {
             ['bottom', [0, 1]]],
         opacityDefault = (context.storage('background-opacity') !== null) ?
             (+context.storage('background-opacity')) : 1.0,
-        customTemplate = context.storage('background-custom-template') || '';
+        customTemplate = context.storage('background-custom-template') || '',
+        latestBg = [context.background().baseLayerSource(), null];
 
     // Can be 0 from <1.3.0 use or due to issue #1923.
     if (opacityDefault === 0) opacityDefault = 1.0;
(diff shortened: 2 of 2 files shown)


bagage:
I'd love having that! Any ETA for integration and/or comments?

bagage:
Bump a last time :+1:

bhousel:
I like the idea, but we can't use 'S' because that key is used for straighten and square operations.

RoPP:
I was thinking I'd checked, I changed for 'I' (interchange), as it seems to be an available shortcut (at least from http://wiki.openstreetmap.org/wiki/ID/Shortcuts).

RoPP:
@bhousel any comments on the PR now that the shorcut is on "I" key ?

bhousel:
'I' is kind of a strange choice too.  I was really thinking of letting the user quick switch between backgrounds using the row of number keys 6,7,8,9,0 or something, maybe with shifted number keys to change backgrounds on the minimap.

Is that maybe too complicated?  If you think it sounds good, I can put that in instead.

RoPP:
In my opinion, it's probably too complicated for our use case. Our need is just to compare data between two backgrounds. But if you think there is a need for shortcuts of all backgrounds, why not (note: non alphabetic shortcuts seem to not work properly, for instance zoom in ('+') and zoom out ('-') do not work on my layout).

By the way, this should address #2635.

bhousel:
See also #2095

pnorman:
> I was really thinking of letting the user quick switch between backgrounds using the row of number keys 6,7,8,9,0 or something,

JOSM uses Alt+number to turn on/off backgrounds. This is a bit different because JOSM has multiple backgrounds at once. I find that I never know which background I'm turning on/off, so have to try multiple ones until the imagery I want is adjusted.

With iD, it'd have to instead use the list of potential backgrounds that editor-imagery-index supplies, right? This would be worse because that list isn't fixed, so what hotkey triggers what imagery would change as you pan.

bagage:
What is the status of this PR as of today? Is there any plan to add one shortcut per background? If not, what must be improved in this PR to be merged in iD?

bhousel:
I thought more about using number keys to switch backgrounds today, but decided against it because there are often too many backgrounds to choose from, and as @pnorman mentioned, the background list is not fixed so it would be confusing.

So I ended up keeping this PR mostly intact, but I changed it to use ⌘B for quick background switching, as this key combination isn't used by anything else and isn't a common browser shortcut.

I also added some tooltips to make this discoverable, and changed the variables and style a little..

Thanks, @RoPP!

PR #2879, opened 2015-12-18
Add Breathe behavior for interpolated select halos

### LINKED ISSUE #1814: Invisible colors in iD
8% of the male population is (to a certain measure) color-blind. Red is weak color for them.
I'm one of those people - using iD is plainly impossible: I cannot see what is selected or even if something is selected.
There is an easy solution: take care take something that has to be distinguished not only has a different colour, but also a different luminance!

Accessibility is a basic requirement for software. iD needs improvement in this aspect. Who should I turn to??

bhousel:
This adds an opacity/stroke interpolator behavior which should make it a lot easier for users with colorblindness to see when things have been selected..

(closes #1814)

merging over the weekend if no opposition..

### DIFF
--- a/css/map.css
+++ b/css/map.css
@@ -43,12 +43,6 @@ path.shadow {
     pointer-events: stroke;
 }
 
-.shadow {
-    -webkit-transition: 200ms;
-       -moz-transition: 200ms;
-            transition: 200ms;
-}
-
 /* points */
 
 g.point .stroke {
--- a/css/map.css
+++ b/css/map.css
@@ -43,6 +43,7 @@ path.shadow {
     pointer-events: stroke;
 }
 
+
 /* points */
 
 g.point .stroke {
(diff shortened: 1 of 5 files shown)


bhousel:
breathe

slhh:
Visibility of the halo is bad here:

selection halos
https://openstreetmap.us/iD/master/#background=Bing&id=w111263466&map=17.52/7.00132/49.22868

Visibily of the ends of the way is good, but these might be outside of the visible area in general.

I believe selection halos need to be drawn on top of all objects except of selected objects.

slhh:
Due to the breathe behavior the selection halos are quite invisible for some time. This might be an issue if the user is not working slowly.
How about making the halo more intense during the time of being thin?

bhousel:
I agree that looks bad there - I'll try to adjust the draw order so that the halos appear above neighboring features.

> How about making the halo more intense during the time of being thin?

I did try this originally, an early version looked more like the "radar ping" that iOS Maps app uses to show the user's location - their halo starts out bright and then drops off as it spreads out.  I found that although that style looked "prettier" (less "throbby") it was also much less visible during most of the animation.

bhousel:
@slhh btw can you open a new issue for this?  I dont want it lost in the comments here.

slhh:
Do you want to have a single new GitHub issue for both select halo visibility issues (drawing order and visibility during being thin) together?

bhousel:
Just a single issue mentioning both things is fine.

slhh:
>Just a single issue mentioning both things is fine.

I have copied the comments concerning the visibility issues to #2914.

PR #2886, opened 2015-12-24
Reorder categories by popularity according to taginfo

bagage:
Basically this PR reorders category-route, category-road and category-path according to taginfo most used values.

The previous order was problematic because most important roads were the firsts in the list. But as a mapper, you do not often map motorways - it's often the inverse. That's why you have now tertiary road then secondary road then primary road and finally motorway (basically the order is inverted).

I also removed `type/route/pipeline` since taginfo reports it only used 34 times worldwide.

### DIFF
--- a/data/presets/categories.json
+++ b/data/presets/categories.json
@@ -48,12 +48,12 @@
         "name": "Path",
         "icon": "category-path",
         "members": [
-            "highway/pedestrian",
             "highway/footway",
-            "highway/cycleway",
-            "highway/bridleway",
             "highway/path",
-            "highway/steps"
+            "highway/cycleway",
+            "highway/steps",
+            "highway/pedestrian",
+            "highway/bridleway"
         ]
     },
     "category-rail": {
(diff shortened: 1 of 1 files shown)


bhousel:
I kind of think it's helpful for new users to see these things in order of logical hierarchy, but haven't really questioned this assumption.

Is this really an issue for people?  I'm inclined to leave the order the way it is unless we get feedback from new mappers that they can't find what they are looking for.

samanpwbb:
Logical hierarchy should be maintained because it is going to be how beginners learn the hierarchy. The point about how frequently tags lower in the hierarchy are used vs those higher in the hierarchy is valid. Maybe a better change would be to simply flip the core tertiary ->  motorway sequence for roads as well as links and leave everything else as is?

bagage:
> Is this really an issue for people? I'm inclined to leave the order the way it is unless we get feedback from new mappers that they can't find what they are looking for.
When I started using iD, I found what I was looking for by selecting nearby routes and get their name; the `route` category is a bit disorganised in my opinion but it is difficult to know what majority of newcomers think of the current sort...

> a better change would be to simply flip the core tertiary -> motorway sequence as well as links
Maybe this would be enough - my major issue yet is having links in the top 10 while I actually never used them nor see them... do you agree for that change @bhousel?

bagage:
My other concern is also that Residential route is at first position - if logical order is kept, residential route should be at least after tertiary roads, should not it be?

bhousel:
No I don't agree with this change.  I think the current road category is fine, with residential listed first (as most frequently used) and the rest of the logical road hierarchy following.  

current road category:
```
            "highway/residential",
            "highway/motorway",
            "highway/trunk",
            "highway/primary",
            "highway/secondary",
            "highway/tertiary",
            "highway/service",
            "highway/motorway_link",
            "highway/trunk_link",
            "highway/primary_link",
            "highway/secondary_link",
            "highway/tertiary_link",
            "highway/unclassified",
            "highway/track",
            "highway/road"
```

proposed changes in this PR:
```
            "highway/residential",
            "highway/service",
            "highway/track",
            "highway/unclassified",
            "highway/tertiary",
            "highway/secondary",
            "highway/primary",
            "highway/trunk",
            "highway/motorway",
            "highway/motorway_link",
            "highway/road",
            "highway/trunk_link",
            "highway/primary_link",
            "highway/secondary_link",
            "highway/tertiary_link"
```

bagage:
If you are fine with current order, I guess this is a dead-end PR.

PR #2890, opened 2015-12-28
added diaper to amenity=toilets

morray:
http://wiki.openstreetmap.org/wiki/Key%3Adiaper
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dtoilets

### DIFF
--- a/data/presets/fields/diaper.json
+++ b/data/presets/fields/diaper.json
@@ -1,0 +1,6 @@
+{
+    "key": "diaper",
+    "type": "combo",
+    "label": "Diaper changing available",
+    "options": ["yes", "no", "room", "1", "2", "3", "4", "5"]
+}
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks, this looks good!

Can you change the label to "Diaper Changing Available"  (we tend to Proper Case the field labels).
Then run `make` and commit the changes to the generated files?

morray:
Hi bhousel, OK changed the case. For you second coment: is there an easy way to do this from within github, or do I have to setup a build system?

bhousel:
Need a build system to run `make`, but that's ok, I'll take care of that step for you later when I'm back at my computer. 

Thanks again!

morray:
Thanks. Over the Christmas  holidays I only have my windows-notebook on me and there it's really a pain to set this up ;-).

bkil:
There is an ongoing proposal to enhance tagging of baby changing tables. Please share with whoever may be concerned and feel free to comment:
https://wiki.openstreetmap.org/wiki/Proposed_features/baby_changing_tables

Issue #2893, opened 2016-01-02
Add presets for various tags (2)


math1985:
Would it be possible to add presets for the following tags? They all have more than 10,000 occurrences.

* [ ] man_made=works
* [ ] man_made=surveillance
* [ ] amenity=waste_disposal
* [ ] man_made=chimney
* [ ] man_made=groyne
* [ ] man_made=gasometer

bhousel:
Sure, can you please add wiki links - makes things easier for us..
This will probably go in after the next release which is any day now, because I like to give the translators some time to catch up.

manfredbrandl:
Hope that helps:
http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dworks
http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dsurveillance
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwaste_disposal
http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dchimney
http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dgroyne
http://wiki.openstreetmap.org/wiki/Tag:man_made%3Dgasometer

manfredbrandl:
Now that 1.8.4 is out I would like to help including new presets. Is there something like a "Howto create a PR to add presets"? I did not find that,

math1985:
I don't know, but at least here you have an example PR:

https://github.com/openstreetmap/iD/commit/da0319677497d69f7036d8e8bd8cac00e6591741

bhousel:
> Now that 1.8.4 is out I would like to help including new presets.

Great!  I appreciate the help..

There is a [README in the presets folder](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md) with more info about how the presets work.

tyrasd:
done in #2897

PR #2897, opened 2016-01-07
Adding presets according to issue #2893

manfredbrandl:
Adding presets according to issue #2893

### DIFF
--- a/data/presets/presets/amenity/waste_disposal.json
+++ b/data/presets/presets/amenity/waste_disposal.json
@@ -1,0 +1,17 @@
+{
+    "icon": "waste-basket",
+    "geometry": [
+        "point",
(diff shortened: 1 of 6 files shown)


manfredbrandl:
@bhousel please review this change.

tyrasd:
Thanks, Manfred!

Btw: Next time, you could also run [`make`](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md#building) after adding the preset(s) and commit the produced asset files in the `data` (and `dist`) directory. That way your changes would be immediately live on http://openstreetmap.us/iD/master (and immediately sumbitted to *transifex* for translation) for example.

math1985:
Thanks!

PR #2901, opened 2016-01-08
Don't dumbly assume svg-sprite is in $PATH

pnorman:


pnorman:
> I could be wrong, but I don't think you need to list node_modules/.bin/svg-sprite as a source here? If you have it listed, make will expect there to be a rule to build the source, and that's not really necessary because the rule is just the npm install that $(MAKI_SOURCES) requires

It will only fail if it doesn't exist, and if it doesn't exist it'll fail anyways. The rule is depending on `node_modules/.bin/svg-sprite`

I suppose a more correct way would be to have a rule for `node_modules/.bin/svg-sprite` which calls `npm install`

jfirebaugh:
You can use the `node_modules/.install` prerequisite to ensure that `npm install` has been run.

PR #2907, opened 2016-01-12
add comment

viclm:


bhousel:
Hi, thank you for this, though I'm confused why you want to add it.  Most of the code in `js/lib` isn't really iD code, but rather things that we have found other places and bundle with iD.

I would totally welcome more source documentation of the iD code and API, but we should first open a proper issue to track that, decide on a JSDoc tool, and add it to the build process, etc.  

As it stands, this PR doesn't really make sense to me.  If you want to discuss the reasons behind it, maybe open an issue first and we can talk about what you are trying to achieve.

PR #2931, opened 2016-01-24
Autocomplete changeset comments from previous changeset comments

### LINKED ISSUE #2002: History for Commit message
Just like here, for entering the title of new issues.

That would make it more easy to re-use 
one of the previous commit messages,
if you revisited+fixed an earlier edit,
after some other saves.

jfirebaugh:
Fixes #2002

:eyes: @bhousel

### DIFF
--- a/js/id/core/connection.js
+++ b/js/id/core/connection.js
@@ -322,6 +322,23 @@ iD.Connection = function(useHttps) {
         oauth.xhr({ method: 'GET', path: '/api/0.6/user/details' }, done);
     };
 
+    connection.userChangesets = function(callback) {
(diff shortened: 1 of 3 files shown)


bhousel:
This is a cool way to do it - I like it.

CCing the [macrocosm](https://github.com/developmentseed/macrocosm) folks a heads up, as they will need to support a new changeset API endpoint.

cc @mojodna @olafveerman

mojodna:
@bhousel thanks for the heads up! For now, POSM has switched to using openstreetmap-website + CGImap (mainly because we need to extract edits from the history tables), but we may be back to look at macrocosm in the future and this will be good to have known about.

Issue #2303, opened 2014-07-25
Easy to create single node ways in iD


SomeoneElseOSM:
I did this by accident in https://www.openstreetmap.org/changeset/24343839

What I did was:

Click Line

Click an existing node

Move the map down a bit 

Click the map to draw an "end" node

Decide that that wasn't quite the right place

Press delete

Try and click where the end node should have gone - but I'm not drawing a line any more!

So click line, draw the thing in the right place, tag it etc. - but the original aborted attempt is added as single-node way https://www.openstreetmap.org/way/294294495

Browser is Chrome Version 36.0.1985.125 m on Windows 7.

pnorman:
Another case

http://www.openstreetmap.org/changeset/35473803

This one is with iD 1.8.0, and had one [tagless way](http://www.openstreetmap.org/way/381605679/history) and [one single-node way](http://www.openstreetmap.org/way/381605677/history), as well as a bunch of Haiti mapping. The errors were in Washington, DC near the default start location.

jfirebaugh:
I can't reproduce creating single node ways using the steps given.

1. Click Line - OK
1. Click an existing node - OK
1. Move the map down a bit - Shouldn't make any difference, but OK
1. Click the map to draw an "end" node - Ambiguous. A single click would not create an "end" node, it would continue drawing the line. Did you mean "click the map to add a vertex" or "double click to finish the line"?
1. Press delete
  * If you meant "click the map to add a vertex" in the previous step, pressing delete now ends the draw operation and creates a two-vertex line. No bug.
  * If you meant "double click to finish the line" in the previous step, pressing delete now deletes the created two-vertex line. No bug.

It sounds like you may have been expecting the behavior requested in #2057, but that's not implemented in iD yet, and in any case is independent of any potential single-node-way creation bug.

SomeoneElseOSM:
I've managed to reproduce a single-node way again doing this - see http://api06.dev.openstreetmap.org/way/4299893242 .

To clarify:

Firstly the initial node must (now) have a tag on it, hence the fake place of worship in the test change I just made.

When I said "end node" I just meant "vertex" - single click, not double.

Moving the map down a bit isn't needed to cause the problem.

(and for the avoidance of doubt) When I mean "press delete" I mean "delete" not "backspace".

jfirebaugh:
OK, so:

1. Starting condition: have a tagged vertex visible.
2. Click the "Line" button
3. Click the tagged vertex to begin drawing a line.
[...]

PR #2936, opened 2016-01-26
make sure DrawLine mode is called with a clean pre-operation graph

tyrasd:
Potential fix for #2303.

### DIFF
--- a/js/id/modes/add_line.js
+++ b/js/id/modes/add_line.js
@@ -14,7 +14,7 @@ iD.modes.AddLine = function(context) {
         .on('startFromNode', startFromNode);
 
     function start(loc) {
-        var graph = context.graph(),
+        var baseGraph = context.graph(),
             node = iD.Node({loc: loc}),
             way = iD.Way();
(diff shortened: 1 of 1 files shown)


tyrasd:
@jfirebaugh can you please take a look and review this PR (considering the changes in https://github.com/openstreetmap/iD/commit/8a215cac81a88e2346ac0e0fbe4f55140d1845cc#diff-dca22f16b0f009e9ff9c946ca989cb99)?

jfirebaugh:
Yes, I see what's going on -- [`drawWay.cancel`](https://github.com/openstreetmap/iD/blob/4ce9bedb9c1866839a4d23539a5f4cffba623edb/js/id/behavior/draw_way.js#L182-L182) restores this `baseGraph`, so it needs to be the graph state before adding the pending way.

The reason I could never reproduce this issue is that Mac keyboards do not have a key that triggers this action. They have a "delete" key but it does what a PC keyboard calls "backspace".

Issue #2927, opened 2016-01-23
Use of the leisure=swimming_pool tag


math1985:
The tag leisure=swimming_pool is mainly used by mappers to mark the water area of a swimming pool. However, some mappers also use this tag to map an area that includes a swimming pool's facilities (for example a building that houses a swimming pool). A separate tag to capture the latter concept was [rejected](http://wiki.openstreetmap.org/wiki/Proposed_features/Aquatics_centre#Voting). However, the discussion did point out that the overwhelming consensus is to map the facilities of a swimming pool as leisure=sports_centre, sport=swimming.

Users of iD tend to use the tag for both the water area and the entire facility, which is understandable as the word 'Swimming pool' in English can mean both. The data would be more useful if iD could better support users to choose the correct tag.

I have some suggestions which might improve data quality. They are not mutually exclusive, but some of them might perhaps be better than others:

* [x] Render leisure=swimming_pool with a blue rather than a white outline.
* [ ] Call leisure=swimming_pool something like 'Swimming pool (water area only)'
* [ ] Add a searchable preset 'Swimming pool (facility)' or 'Swimming pool (sports centre)' or something of the sort for the tag combination leisure=sports_centre, sport=swimming.

If you have any other ideas how users can be helped to pick the right tag here that would be good too!

Thanks for having a look at this issue!

bhousel:
These are good ideas..

I think 
* "Swimming Pool" is a good name for `leisure=swimming_pool`.. blue rendering will help.
* "Swimming Pool Facility" is a good name for `leisure=sports_centre + sport=swimming`

math1985:
Yes, I think that would work!

slhh:
The tag combination leisure=sports_centre + sport=swimming has been suggested for sport focused facilities.
>"Swimming Pool Facility" is a good name for leisure=sports_centre + sport=swimming
This would led to using it also for the more play focused facilities, where leisure=water_park should be used.

bhousel:
> This would led to using it also for the more play focused facilities, where leisure=water_park should be used.

So we can also make a preset for:
- [ ] [`leisure=water_park`](http://wiki.openstreetmap.org/wiki/Tag:leisure%3Dwater_park)

And make sure that the terms for that include "swimming pool" so that anyone searching for a preset like that will see both options.

HolgerJeromin:
Amenity =swimming_pool should not be used anymore. But your recent commit adds blue color to this tag...

bhousel:
[...]

PR #2939, opened 2016-01-28
Call leisure=swimming_pool something like 'Swimming pool (water area …

kepta:
Adds swimming pool facility to presets

Issue:https://github.com/openstreetmap/iD/issues/2927


kepta:
@bhousel review ?

bhousel:
This is really good!

A few minor nitpicks:
* Can you make a subfolder under `leisure` called `sports_centre`, then rename `swimming_facility.json` to `swimming.json` and put it in there.  (If we get more requests for subtypes of sports centres, they will go in the subfolder.  The `pitch` subtypes follow this convention too.)
* Rename `sports_center.json` to `sports_centre.json`.   (I know you didn't create this, it was this way already.  But we try to follow the convention of having the file named the same thing as the tag, so we may as well correct it now.)
* Remove `vertex` geometry type from the Swimming Pool and Swimming Pool Facility presets.
* Add a `leisure=water_park` preset (requested in #2927) <br/>
(icon = `swimming`, fields = `operator`, `address`; geometry = `point`, `area`)
* Add `terms` (to help place them properly in search results):
  *  "Swimming Pool" :   `"terms": ["dive", "water"]`
  *  "Swimming Pool Facility" :   `"terms": ["dive", "water"]`
  *  "Water Park" :   `"terms": ["swim", "pool", "dive"]`

kepta:
@bhousel check now?

bhousel:
Changes look good!  I'll merge them today.

The conflicts are because of differences in the minified json files..  This is something I did in c2833ba that is causing a lot of problems for everyone lately, so I think I'm just going to revert that change.  `git` can't diff those files correctly when the file is just condensed down to a single line.

bhousel:
merged in fe246c6

PR #2965, opened 2016-02-03
HTTPS sensitive URIs

### LINKED ISSUE #2960: Access Nominatim via HTTPS for location search
When trying to do a location search using the "Search worldwide" button, my browser is blocking the request because I loaded osm.org via HTTPS but iD is trying to contact Nominatim via HTTP. I am using Chromium but I think all browsers have become more strict about this in the last year or two.

I see for reverse geocoding Nominatim's URL was changed in 16a2449 but the search feature is still pointing at HTTP. I believe the simplest fix might be to just change [line 248 in feature_list.js](https://github.com/openstreetmap/iD/blob/master/js/id/ui/feature_list.js#L248) to point at https://

kepta:
change of all http:// URIs in js/id folder to //.

(closes #2960)

### DIFF
--- a/js/id/renderer/background_source.js
+++ b/js/id/renderer/background_source.js
@@ -117,7 +117,7 @@ iD.BackgroundSource.Bing = function(data, dispatch) {
     };
 
     bing.logo = 'bing_maps.png';
-    bing.terms_url = 'http://opengeodata.org/microsoft-imagery-details';
+    bing.terms_url = 'https://blog.openstreetmap.org/2010/11/30/microsoft-imagery-details';
 
     return bing;
 };
--- a/js/id/renderer/mapillary_layer.js
+++ b/js/id/renderer/mapillary_layer.js
@@ -16,7 +16,7 @@ iD.MapillaryLayer = function (context) {
             .attr('src', 'https://d1cuyjsrcm0gby.cloudfront.net/' + image.key + '/thumb-320.jpg');
 
         div.selectAll('a')
-            .attr('href', 'http://mapillary.com/map/im/' + image.key);
+            .attr('href', 'https://www.mapillary.com/map/im/' + image.key);
     }
 
     function hide() {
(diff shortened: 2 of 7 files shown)


js/id/renderer/background_source.js
```diff
@@ -117,7 +117,7 @@ iD.BackgroundSource.Bing = function(data, dispatch) {
     };
 
     bing.logo = 'bing_maps.png';
-    bing.terms_url = 'http://opengeodata.org/microsoft-imagery-details';
+    bing.terms_url = '//opengeodata.org/microsoft-imagery-details';
```
bhousel: Looks like this has moved to:
https://blog.openstreetmap.org/2010/11/30/microsoft-imagery-details/

js/id/ui/splash.js
```diff
@@ -23,8 +23,8 @@ iD.ui.Splash = function(context) {
             .append('p')
             .html(t('splash.text', {
                 version: iD.version,
-                website: '<a href="http://ideditor.com/">ideditor.com</a>',
-                github: '<a href="https://github.com/openstreetmap/iD">github.com</a>'
+                website: '<a href="//ideditor.com/">ideditor.com</a>',
```
bhousel: This one isn't available over https
(we should really just move it to github pages)

bhousel:
see also discussion on #2681 

For outbound links and api endpoints, we need to make sure the resource is available over both `http` and `https`.  I commented in the diff the places where that's not true.

Given that iD can sometimes be loaded locally, and `file://` urls can break in surprising ways, maybe we should instead add the `location.protocol === 'http' ? 'http' : 'https'` check to `iD.detect()`?

Thoughts @jfirebaugh ?

kepta:
This `//` will load the hyperlink depending on the iD's current connection status 
if plain then // will go to http://
if encypted then // will go to https://

jfirebaugh:
Please don't use protocol relative URLs -- as @bhousel said they fail for `file://`, and https is preferred even when loading the referring page on http. Therefore:

* If the site supports https, use https unconditionally.
* If not, look for an alternative URL.
* Failing that, bug the site owner for https support.
* Failing that, stick to http.

kepta:
@bhousel I have updated the patch.
Please have a look.

bhousel:
Thanks this looks good!

Issue #2615, opened 2015-04-29
Fix highway=track


lxbarth:
Ongoing Nepal mapping highlights the massive confusion around `highway=track`. 

> This tag represents roads for mostly agricultural use, forest tracks etc.; usually unpaved (unsealed) but may occasionally apply to paved tracks as well, that are suitable for two-track vehicles, such as tractors or jeeps.

[![screen shot 2015-04-29 at 4 54 56 pm](https://cloud.githubusercontent.com/assets/98233/7401238/8f464d9a-ee90-11e4-9cdf-697cb769ce54.png)](http://www.openstreetmap.org/#map=17/28.00016/84.62408
)

http://www.openstreetmap.org/#map=17/28.00016/84.62408

Can we call track "agricultural track"?

jfirebaugh:
What exactly is the confusion?

bhousel:
@lxbarth since the time you posted that, it looks like they have all been changed to "unclassified road"?

The HOT tasking manager does contain pretty detailed instructions on how to tag different roads, so I'm not sure how renaming Track to something else helps here.

pnorman:
> Can we call track "agricultural track"?

They're not purely agricultural - see forest tracks.

lxbarth:
> What exactly is the confusion?

All across OpenStreetMap and in the particular example above, dirt roads are tagged `highway=track` where they should be tagged `surface=unpaved` together with the appropriate road classification. `highway=track` is reserved for roads for agricultural or similar use.

> Do not use tracks to represent public unpaved roads in built-up areas [1], that would be considered tagging for the renderer. In this situation, classify them as usual according to the conventions in your country, and also provide a surface=* tag.

http://wiki.openstreetmap.org/wiki/Tag:highway%3Dtrack

brycenesbitt:
In the USA they are definitely not "agricultural track".
Many residential roads and hiking routes are marked highway=track.
The highway tag is a mess unfortunately.

pmailkeey:
Totally agree with you Bryce, the highway tag is a complete mess and I take it, it's the core of OSM.

lxbarth - tag the roads in Nepal as you think fit and is appropriate. Get a consensus among all Nepal map editors.

RicoElectrico:
This is a problem that really annoys me, frankly. It's more about tagging for the renderer and possibly bad translation. You would have to check specific locale of the user who added it. (which is possible from some time as locale is tagged in the changeset). 
[...]

PR #2982, opened 2016-02-08
[Preset] Clarification: "Track" → "Two-Track Road (Primitive)"

brianreavis:
**Purpose:** to clarify that `highway=track` is not a trail / single-**track** trail.

I've noticed that particularly in the MTB community there's a trend of classifying trails as `highway=track` because "Track" is a common word to refer to single-track trails (e.g. [changeset/29897324](https://www.openstreetmap.org/changeset/29897324#map=16/37.6216/-119.0007))

Most of the road presets have a "Road" suffix that make it clear you're adding a road, but the track preset does not. In its current form it's too abstract, I feel. This PR aims to make it a little more clear what you're adding.

Relevant: [BLM Road Classifications](http://www.blm.gov/id/st/en/prog/planning/Travel_Management/Craters_of_the_Moon/Road_Trail_Classifications.print.html) (Class D)

### DIFF
--- a/data/presets/presets/highway/track.json
+++ b/data/presets/presets/highway/track.json
@@ -22,5 +22,5 @@
         "woods road",
         "fire road"
     ],
-    "name": "Track"
+    "name": "Two-Track Road (Primitive)"
 }


bhousel:
"Two-Track Road (Primitive)" is still not really clear.. 

How about "Vehicular Off-Road Track"?

brianreavis:
Hmm, I'm not sure "Off-Road" is a good description, given it is a road... it's just quite primitive. Just throwing some words out here: unmaintained, high-clearance, ruts, Jeep Road, 4WD Track, 4x4 Track, Track (Road)

bhousel:
I like "Unmaintained"

bhousel:
related: https://github.com/openstreetmap/iD/issues/2615#issuecomment-104828550

brianreavis:
I don't want to start bikeshedding here given there's already been long discussions like the one you linked. At the end of the day it needs "Track" in the name and also "Road". Any other descriptors (or none at all) are fine by me.

brianreavis:
Another approach altogether is to not modify the preset at all and instead introduce a "Common Confusion" info box that comes up below the selected preset that can nudge people to the right preset.

bhousel:
"Unmaintained Track Road" it is..

> Another approach altogether is to not modify the preset at all and instead introduce a "Common Confusion" info box that comes up below the selected preset that can nudge people to the right preset.

Ideally, the description and photo that we fetch from the OSM wiki should clear up all confusion about how to use the preset.

screenshot 2016-02-16 23 19 42

(sorry @1ec5)

brianreavis:
Cool, thanks @bhousel!

PR #2983, opened 2016-02-08
README: correct wrong path

zbycz:
After running make, the presets are actually in `dist/presets.js`.

### DIFF
--- a/data/presets/README.md
+++ b/data/presets/README.md
@@ -89,8 +89,7 @@ names when identifying the icon to be used for a given preset.
 To build presets, all you need to do is run `make`.
 
 This command will take care of running the build script, which packages all presets
-together with imagery data, and deprecated or discarded tags into one file, `data/data.js`,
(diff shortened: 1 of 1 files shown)


data/presets/README.md
```diff
@@ -89,7 +89,7 @@ names when identifying the icon to be used for a given preset.
 To build presets, all you need to do is run `make`.
 
 This command will take care of running the build script, which packages all presets
-together with imagery data, and deprecated or discarded tags into one file, `data/data.js`,
+together with imagery data, and deprecated or discarded tags into one file, `dist/presets.js`,
```
bhousel: Good catch, the presets are in their own file now.  Can you also remove mention of the "imagery data, and deprecated or discarded tags"?

zbycz: Sure - see the new "ammended" commit.

bhousel:
Thanks!

PR #2984, opened 2016-02-09
Test with phantomjs 2.1 and mocha-phantomjs-core

### LINKED ISSUE #2905: Mocha and PhantomJS
PhantomJS has transitioned to v2 but the version of Mocha we use as a testrunner is somewhat capped at PhantomJS v1.9.7, which is starting to show its age.  

A consequence of this:  running our tests was generating a bunch of ["SSL handshake failed" errors](
https://github.com/nathanboktae/mocha-phantomjs/issues/173) because PhantomJS has better TLS support starting in v1.9.8.  

I worked around the issue by adding 
~~`--ssl-protocol=any --ignore-ssl-errors=true`~~
`--ssl-protocol=tlsv1`  ([better] (http://stackoverflow.com/a/26417660/7620))
to the [test command](https://github.com/openstreetmap/iD/blob/1afab94a321e504044105d758764bfca287110a2/package.json#L11), but I don't view this a permanent fix.

Just opening this issue to keep track of the situation until either Mocha resolves the issue (ideal) or we switch to a different testrunner (bigger lift).

jfirebaugh:
Fixes #2905

:eyes: @bhousel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,9 @@
 language: node_js
 sudo: false
+before_install:
+  # https://github.com/travis-ci/travis-ci/issues/3225
+  - mkdir travis-phantomjs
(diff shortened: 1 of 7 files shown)


bhousel:
This is great, thank you!

PR #2988, opened 2016-02-12
Adding ▶/▼ to category presets to make it clearer

### LINKED ISSUE #2972: Make it more clear that "Building" is a category
I've been to a few mapathons lately where newbies are confused about tagging buildings: often they'll click Building (the category) and assume that it's tagged, rather than scrolling down and clicking Building again (which then makes it pink and building=yes.) It's confusing to them to pick it twice, even when I have explained the first is the category and the second is the specific type. (Recognition versus recall and all that.)

Maybe call the category "Types of buildings" or "Building types" or something that makes it more apparent that the first time you click you're selecting a category and the second is the specific type? There's the little paper overlap thing on the categories, but I think that's too subtle for new people to understand or notice. 

Or maybe even have it so when you click the building category it automatically tags it as building=yes, and then you can choose another kind if you want from the dropdown.

I wonder if part of the confusion is that on smaller laptops or lower resolutions, the generic "building" is at the bottom of the list, and thus off the page, so they don't think to pick it.

kepta:
(closes #2972)

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -2,25 +2,25 @@ en:
   presets:
     categories:
       category-building:
-        name: Building
+        name: Building Features
       category-golf:
-        name: Golf
+        name: Golf Features
       category-landuse:
-        name: Land Use
+        name: Land Use Features
       category-path:
-        name: Path
+        name: Path Features
       category-rail:
-        name: Rail
+        name: Rail Features
       category-restriction:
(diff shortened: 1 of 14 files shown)


kepta:
@bhousel review?

bhousel:
Nice, this will make the difference between categories and presets much clearer!  :+1:

PR #2990, opened 2016-02-13
Live reload documentation

kepta:
I often use `live-server` while working on iD and thought it might be use for others as well.
Let me know if it is necessary 🙃


kepta:
> can you change it to say "local" server page
I didn't quite get this part.

bhousel:
Oh, it just says "locally server page", and "local server page" sounds a bit better.

kepta:
Oh ya, let me fix that.

kepta:
@bhousel the commit is updated.

bhousel:
Thanks!

PR #2996, opened 2016-02-18
Post office should have opening_hours instead collection_times

HolgerJeromin:


bhousel:
You are right - thank you!

gileri:
Couldn't they have both ? Usually there is a specific time each day when the collected mail is sent to a sorting/distribution site, and past this time the mail would be collected/sent the next day; which is very much like a post_box.

HolgerJeromin:
This is probably not easy to verify.

gileri:
In france, [post office collection times are available on lapostefr.fr](http://imgur.com/ZwTL3VF), the official website for post offices (see bottom of picture).

I think that you can get collection times in post offices by asking at the counter, and I trust it can be qualified as "easily verifiable", even if it need a survey.

PR #2998, opened 2016-02-21
map features master checkbox

### LINKED ISSUE #2904: add map features master checkbox
For (Shortcut F) map features, there are checkboxes for "points", "major roads",
"minor roads", etc. but no additional checkbox at top to turn them all off or on at once.

kepta:
Adds a master checkbox which selects/deselects all map features.

(closes #2904)

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -285,6 +285,7 @@ en:
     data_layers: Data Layers
     fill_area: Fill Areas
     map_features: Map Features
+    select_all: Select All
     autohidden: "These features have been automatically hidden because too many would be shown on the screen.  You can zoom in to edit them."
   feature:
     points:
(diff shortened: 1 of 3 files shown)


kepta:
@bhousel review?

bhousel:
So, this code is awesome and does exactly what it was supposed to do... I really like what you did.

BUT, I thought about this some more and I'm going to add OpenStreetMap as a data layer here instead:

screenshot 2016-02-29 12 16 28

It makes more sense to me, post-#3001 layer refactor, and I want to reinforce the idea that OSM is just one of many layers that iD can support.  

We have more projects now where people are starting to fork iD and swap out OSM for something else, and I think this signals to them conceptually the cleanest way to do it..

PR #3005, opened 2016-02-27
Preventing the jumping of point markers

kepta:
In this patch, I am saving the difference between mouse coordinates and the node.
and then using this offset to compensate for jumping.

However, I am not really sure whether to put it before or after the `if else` clause in the subsequent lines.

**drag_node.js: 112**
``` Javascript
if (d.type === 'node' && d.id !== entity.id) {
    loc = d.loc;
} else if (d.type === 'way' && !d3.select(d3.event.sourceEvent.target).classed('fill')) {
    loc = iD.geo.chooseEdge(context.childNodes(d), context.mouse(), context.projection).loc;
}
```

Are there any edge cases I should look into?

### DIFF
--- a/js/id/modes/drag_node.js
+++ b/js/id/modes/drag_node.js
@@ -8,6 +8,7 @@ iD.modes.DragNode = function(context) {
         activeIDs,
         wasMidpoint,
         cancelled,
+        nodeOffset,
         selectedIDs = [],
         hover = iD.behavior.Hover(context)
             .altDisables(true)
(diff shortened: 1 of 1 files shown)


kepta:
@bhousel review?

bhousel:
I took a look at this yesterday but it's not quite right.. The code that you added in `move()` takes the `nodeOffset` location from the `datum()` - which really contains the location of the thing "under" the point being dragged.  The code _seems_ to work because when there is nothing under the point being dragged, you end up with the mouse coordinates on the surface.  But in cases where there are coincident points (i.e. something under the point) `nodeOffset` will have the wrong value, or where the user is holding down the "alt" key to disable feature snapping, `nodeOffset` will have no value and throw an exception.

(The `if else` clause you mentioned above is for snapping the dragged point onto another point, or onto a line.)

This was not really obvious to me either, but I realized this morning that the mousemove `d3.event` is [already decorated with a `point` property that has the offset built into it](https://github.com/openstreetmap/iD/blob/master/js/id/behavior/drag.js#L110).  

So I did d130c51 which swaps `mouseCoordinates()` for this projected `d3.event.point` and seems to work ok.

kepta:
Oh my god,
That's a one line fix.

I wonder why I couldn't see that. 😭

bhousel:
Don't feel bad, I only saw it after trying to understand the `drag.js` code for a few hours yesterday.  Also a good nights sleep, and some coffee in the AM.

PR #3012, opened 2016-03-03
expand globs in Makefile to fix ordering issues on make >=3.82

substack:
gnu make does not guarantee the ordering of glob expansions: https://bugzilla.redhat.com/show_bug.cgi?id=635607

Prior to 3.82, make happened to sort glob results. On make version 4, I get inconsistent results that cause the built ideditor to fail due to ordering issues.


bhousel:
Yuck.  Thanks for fixing this!

PR #3014, opened 2016-03-03
adding MapillaryJS, first stab.

peterneubauer:
This is a first add of the MapillaryJS viewer, a (non scientifically hosted, so icons are wrong) version can be seen at http://mapillary.github.io/iD/#background=Bing&map=18.70/13.00314/55.59456

I need some help with the activating/deactivating of the Viewer component on hover, since we only want to activate it once, then just hide/show the viewer. WDYT @bhousel ?


bhousel:
> I need some help with the activating/deactivating of the Viewer component on hover, since we only want to activate it once, then just hide/show the viewer. WDYT @bhousel ?

I'll check it out.. D3's ["general update pattern"](https://bl.ocks.org/mbostock/3808234) is our friend here.  

The D3 way of doing this is to bind a selection like this to some data.  D3 runs:
* the `enter()` selection for elements that exist in the data and not the DOM  
* the update selection for elements that exist in both the data and the DOM
* the `exit()` selection for elements that exist in the DOM and not the data.

```js
var divs = selection.selectAll('.something')
    .data(isVisible ? [0] : []);

divs.enter()
    .append('div')
    .attr('class', 'something');

divs
    .attr('foo', currentValue);

divs.exit()
    .remove();
```

peterneubauer:
@bhousel did you have any time to look at this yet? Would be nice to get this in, too - I think MapillaryJS provides much better UI experience.

peterneubauer:
@bhousel any chance you could take a look for some of the interaction parts? Thanks a lot!

bhousel:
Hey @peterneubauer,  yes @kepta is working on merging this - he just started this morning..

peterneubauer:
Perfect - I think in the traffic signs there is some interaction problems, but I think they should be solvable by an D3 expert :)

kepta:
Closing this PR and continuing this feature request at  #3128

PR #3048, opened 2016-03-24
FAQ.md: api url config in standalone iD

zbycz:
It took me a while to find it out. I guessed it would be in a separete config file, but it wasnt.

btw, I really wonder, how is the API url set up in the embeded iD in ruby-port. It uses correctly the my own server. Do you know how? :)

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -53,7 +53,7 @@ and [configure](https://github.com/openstreetmap/openstreetmap-website/blob/mast
 an instance of the Rails Port, the server that runs the OpenStreetMap website and API.
 
 Once you have the Rails Port running, you may edit as normal using the version of iD that
-is bundled with it. Your changes will be saved to your own database.
+is bundled with it. Your changes will be saved to your own database. To use a standalone iD with your own api, you may edit the [connection.js](https://github.com/openstreetmap/iD/blob/master/js/id/core/connection.js) file.
 
 Depending on your requirements, you may also want to set up [cgimap](https://github.com/openstreetmap/cgimap)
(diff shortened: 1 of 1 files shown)


bhousel:
> It took me a while to find it out. I guessed it would be in a separete config file, but it wasnt.

Yes, when running a custom stack `connection.js` is the right place to change the OSM endpoints.  I am leaning strongly towards moving this file into the `iD.services` namespace, like where Taginfo, Nominatim, Mapillary, Wikipedia live.

> I really wonder, how is the API url set up in the embeded iD in ruby-port. It uses correctly the my own server. Do you know how?

I'm not sure what you're asking, but maybe the answer is in here:
https://github.com/openstreetmap/openstreetmap-website/blob/master/CONFIGURE.md

zbycz:
Thanks for merging.

To clarify my question:
- when i run standalone instance - I have to edit connection.js
- when i install openstreetmap-website, then the embeded iD instance is already configured to use not "http://openstreetmap.org/api", but rather something like "/api".   ... So, where in the build process is the value changed?

But its just beeing curious, dont worry :-)

bhousel:
Sorry, I have no idea..

zbycz:
Just in case somebody googles here.

Its possible to add this code to `index.html` (inspired by [openstreetmap.org/id's index.html](http://www.openstreetmap.org/id)): 

```js
id.connection().switch({
                    "url": "http://localhost:3000",
                    "oauth_consumer_key": "xx",
                    "oauth_secret": "xx"
});
```

PR #3052, opened 2016-03-29
added diaper to some of the most common occurrences besides toilets 

morray:
added diaper to some of the most common occurrences besides toilets 
see also http://wiki.openstreetmap.org/wiki/Key:diaper


jfirebaugh:
I don't see how "diaper" is a relevant term for the "Cafe" or "Restaurant" presets. Is there a particular reason you think it should be added?

jfirebaugh:
Ah sorry, I misread this as adding to the search terms.

morray:
I would say it is relevant since these are one of the top places to find these.

bhousel:
Adding the `diaper` field to `amenity=toilets` in #2890 makes a lot of sense and I think that was a great addition to the preset.  But I think adding a `diaper` field to other things that happen to have toilets (like restaurants and cafes) is unnecessary, and it makes it more work to edit those other features.  We should be encouraging people to tag consistently, especially given that `diaper` is a relatively new tag with only about 1000 uses.  If a person really wants to add a diaper changing table to a restaurant, we should encourage them to add a node for the restroom, then tag the restroom details on that node.

bkil:
There is an ongoing proposal to enhance tagging of baby changing tables. Please share with whoever may be concerned and feel free to comment:
https://wiki.openstreetmap.org/wiki/Proposed_features/baby_changing_tables

PR #3055, opened 2016-04-03
add multiselect preset

kepta:
ezgif com-resize

This is what it looks like.

Currently I have these issues,
- Combobox doesnt show dropdown if only one item is remaining.
- On pressing enter editing exits.

@bhousel what do you think about this implementation ?


bhousel:
moving this to new branch and new pr #3080 so I can push commits to it..

Issue #2739, opened 2015-07-22
Error message preventing save:  Relation not found for reference -9 in relation -4.


SHARCRASH:
Hello,

I'm having an error message while trying to save changes i've made:

<b>Errors occurred while trying to save:</b>
Placeholder Relation not found for reference -9 in relation -4.
Server returned status code 400

These changes are worth about <b>4 hours of work and 186 changes</b> according ID editor. I got into a massive attempt to correct a large zone, this made me convert many areas overlapping each other into proper MPs. I still have the ID window opened in Firefox. What can i do to avoid loosing these changes and upload the changes??? Please help... I think i'll become crazy if this is lost...

I've posted this issue in [OSM help](https://help.openstreetmap.org/questions/44371/error-message-while-trying-to-save) too.

bhousel:
Hmm, this is a duplicate of #2485 but I would really like to track it down..  Can you still open iD in Firefox and restore your changes?  If so, it would be really helpful to send me your iD history so I can see what happened.

* Open developer console
* Change "main frame/top frame" to "id-embed" if necessary, eg. running from openstreetmap.org
* Run this:
```
(function getHistory() {
    function getKey(n) { return 'iD_' + window.location.origin + '_' + n; }
    return id.storage(getKey('saved_history'));
})();
```

The output will be really long, but you could save it to pastie.org / pastebin.com or a place like that.

SHARCRASH:
Last wednesday I had to close Firefox because it crashed. Don't know if it's related. 

But... I encountered something weird yesterday, right after OSM's maintenance... ID recovered another previous "lost changes" session, i recognized clearly because i was going to start to plot many empty lines. So... I'm saying this because maybe there is a way to retrieve the "186 edits".

I couldn't try your steps. I don't see what you mean with "main frame/top frame"? Is that in javascript? And how will i run the script code you gave me?

bhousel:
> But... I encountered something weird yesterday, right after OSM's maintenance... ID prompted to recover another previous "lost changes" session, i recognized clearly because i was going to start to plot many empty lines. So... I'm saying this because maybe there is a way to retrieve the "186 edits".

Upon starting, iD will always offer to restore your edits with the "You have unsaved changes" message, if the edits are still stored in your browser's local storage.  Choose "Restore" and you should be put right back to your pre-save state with 186 edits, and saving will always give you the same error (this is what I hope you can still do).  Choose "Reset" and the edits are unfortunately gone forever.

[...]

PR #3062, opened 2016-04-15
Fix member argument to ChangeMember when using raw member editor

bhousel:
There is a nasty bug when using the Raw Member Editor to set the "role" of a relation member.  The `member` argument to `iD.actions.ChangeMember` was constructed wrong, causing the relation to become corrupted. 

I'm not 100% sure that this caused https://github.com/openstreetmap/iD/issues/2739#issuecomment-210112137 but it seems possible.

:eyes: @jfirebaugh


jfirebaugh:
👍 

Seems very likely that this was the cause of #2739.

PR #3064, opened 2016-04-15
add a warning to the changeset page if a user mentions google

### LINKED ISSUE #3063: Add unobtrusive warning if a Changeset comment includes the word "Google"
This is something we're increasingly flagging in our validation systems, and I believe it would be useful to build into iD itself.

This would consist of:

* At the save stage, check the text for google (in any capitalization)
* If google is found, show a warning with a link to the copyright page
* We should still allow saving with this message, imho, and keep it only a warning in the first iteration. A second iteration might include a more obtrusive flow, like requiring a checkbox to be checked.

tmcw:
google

Fixes #3063

This moves a check that's very common in our post-changeset validation up before the user even makes the mistake. This _doesn't_ prevent the user from saving the changeset yet. We may want to do that in a second pass, but there are innocuous ways to mention Google, like "This feature isn't in Google Maps, but it does exist.", so blocking save entirely is probably _too_ strong.

cc @geohacker @bhousel for the review

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -206,6 +206,8 @@ en:
     created: Created
     about_changeset_comments: About changeset comments
     about_changeset_comments_link: //wiki.openstreetmap.org/wiki/Good_changeset_comments
+    google_warning: "You mentioned Google in this comment: remember that copying from Google Maps is strictly forbidden."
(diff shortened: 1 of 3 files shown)


bhousel:
Thanks, this is a great check to have..

PR #3072, opened 2016-04-18
Studio

### LINKED ISSUE #3061: Use `studio=*` instead of `type=*` for `amenity=studio`
from irc

screenshot 2016-04-13 09 41 30

Looks like this was a recent (2014) change:
https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dstudio#problematic_use_of_the_.22type.22_key

But it makes sense so we should do it.

kepta:
(closes #3061)


bhousel:
perfect, thank you!!

PR #3073, opened 2016-04-18
Add prison preset

### LINKED ISSUE #3070: Add a Preset for Prisons
**Preset Request**. [Prisons](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dprison) are currently shown on the map, [in a stylish gray zebra pattern](https://github.com/gravitystorm/openstreetmap-carto/pull/772), yet there is no easy way to tag them in iD.

The related tags for prison are `name`, `operator`, and `access`. Users often trace the barrier fence as the prison boundary, so the `barrier=fence` tag is useful in combination.

2016-04-17 17_56_58-openstreetmap

kepta:
(closes #3070)

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -1507,6 +1507,25 @@
         },
         "name": "Post Office"
     },
+    "amenity/prison": {
+        "icon": "prison",
+        "fields": [
+            "operator",
+            "access"
+        ],
+        "geometry": [
+            "point",
+            "area"
+        ],
+        "terms": [
+            "cell",
+            "jail"
+        ],
+        "tags": {
+            "amenity": "prison"
+        },
+        "name": "Prison Grounds"
+    },
     "amenity/pub": {
         "icon": "beer",
         "fields": [
--- a/data/presets/presets/amenity/prison.json
+++ b/data/presets/presets/amenity/prison.json
@@ -1,0 +1,19 @@
+{
+    "icon": "prison",
+    "fields": [
+        "operator",
+        "access"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "cell",
+        "jail"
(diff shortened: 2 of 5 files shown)


pnorman:
You should avoid having a linear tag (barrier=fence) on the same object as an area (amenity=prison), and this preset would encourage that.

data/presets/presets/amenity/prison.json
```diff
@@ -0,0 +1,21 @@
...
+        "area"
+    ],
+    "terms": [
+        "prison",
+        "cell",
+        "jail"
+    ],
+    "tags": {
+        "amenity": "prison",
+        "barrier": "fence"
```
bhousel: Yeah what @pnorman said - remove `barrier: fence` from here.
People should tag the fence as a separate element, or of they really want to they can add this tag to the amenity using the raw tag editor, but we can't automatically assume that prisons are also fences.

data/presets/presets/amenity/prison.json
```diff
@@ -0,0 +1,21 @@
...
+    "fields": [
+        "operator",
+        "access"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "prison",
```
bhousel: Don't need to add "prison" as a search term if it is the tag value - it's already included.  The other terms you added are 👍 ..

data/presets/presets.json
```diff
@@ -1507,6 +1507,25 @@
         },
         "name": "Post Office"
     },
+    "amenity/prison": {
+        "icon": "prison",
+        "fields": [
+            "operator",
+            "access"
```
bhousel: I think this should be `address` not `access`?

bhousel: I made this change and merged..

Issue #2748, opened 2015-08-11
feature request: autocompletion for tags


javiribera:
These tags are not being autocompleted when typing in the iD editor:

- button_operated https://wiki.openstreetmap.org/wiki/Key:button_operated
- traffic_signals:*** https://wiki.openstreetmap.org/wiki/Tag:highway%3Dtraffic_signals#Traffic_signals_for_pedestrians

![screenshot from 2015-08-10 21 38 45](https://cloud.githubusercontent.com/assets/5376776/9188261/52234daa-3fa8-11e5-9e50-cb77a401b840.png)

bhousel:
The code in [`taginfo.js`](https://github.com/openstreetmap/iD/blob/522e593f112d3c628604a54a2d0ca0bae11dc5c3/js/id/services/taginfo.js) is doing a few things that we should maybe adjust:

* `shorten` cuts the query down to only the first 3 characters, so above examples would search `but` or `tra`.   This is done to cut down on extra requests to taginfo, but because of debouncing and caching, seems unnecessary.
* `popularKeys` only shows a key if it has at least 10000 uses.  This seems too high to me, and both `button_operated` and `traffic_signals` have a little over 5000.
* `popularValues` will also show less popular values if they have a wiki page. `popularKeys` should have this behavior too.

PR #3080, opened 2016-04-23
Add multiselect preset

bhousel:
continued from #3055


bhousel:
This is really good!  Awesome work @kepta!

I fixed a thing in 1a9c111:
- [X] Use `field.keys` so that undo and reset buttons work

>Currently I have these issues,
>Combobox doesnt show dropdown if only one item is remaining.

This is because d3.combobox defaults to `minItems=2` 
- [ ] we should probably change that to 1, because what if taginfo only fetches 1 thing

> On pressing enter editing exits.

A few things with keypresses:
- [ ]  enter should accept the value but send focus to the input field
- [ ]  tab should do the same 

Also I'll replace the input placeholder text with something translateable
- [X]  "Type here" can just say "Add..."

bhousel:
> and we should check that the taginfo queries are looking up enough values

^ related to this, I did some testing today and it looks like we need to adjust the popularity threshold values in the [taginfo service code](https://github.com/openstreetmap/iD/blob/master/js/id/services/taginfo.js#L37-L45)
We adjusted them before in #2748, but I think they're still too high.

In this case, we need to do a `key` lookup - but the key lookups have a higher threshold for popularity than the value lookups.

bhousel:
> combobox should probably fetch data on typing (like other combobox fields), not fetch data once on field initialization

^ related to this, I realized today that the `combo` field does not fetch data on typing, but the raw tag editor value field does.  This surprised me - I assumed they both had the [bindTypeahead behavior](https://github.com/openstreetmap/iD/blob/4c9da0d518be526dfab5aa7f9ce1cc1fe70142f8/js/id/ui/raw_tag_editor.js#L129-L169) from the raw tag editor..

So, I'm now thinking I want to add typeAhead to combo boxes too, and maybe merge together some of the code in `combo.js` and `multiselect.js`.

kepta:
@bhousel after looking (at)[https://github.com/openstreetmap/iD/commit/c592acb759da44f956b326e21d6ec05555a0de4e],

it was changed to 5000, so should we half it again?

bhousel:
> it was changed to 5000, so should we half it again?

For normal keys it should probably be smaller than 5000.  This usecase only applies in the raw tag editor when the user selects a key, so it's not as visible.  The purpose is, ostensibly, to filter out garbage keys, but in practice some of the most garbage keys in OSM are also the most popular (`tiger:*`, `ref:bag`).   This is also why we have code to sort keys with colons below keys without colons.

But.. for these multiselectable keys (`recycling:*=yes`) the threshold needs to be set very low, or changed to a popularity filter like value uses.  It's a different usecase, so I want to think of a clean way to have `taginfo.keys()` serve both purposes - maybe pass another parameter (we already abuse `parameter` pretty bad).

bhousel:
This is coming along really well!   I refactored combo.js and moved the multiselect code into there, because it turns out a lot of the code needed to be shared anyway.  

Hoping to improve the tabbing behavior and merge soon!

kepta:
That's interesting, so now the combo type will be handling single or multiple values. 

Nice job merging them into one, @bhousel :D

bhousel:
Looks awesome.. merging!  :shipit: 

multicombo1

PR #3084, opened 2016-04-25
phone placeholder from country being edited

### LINKED ISSUE #3083: phone number placeholder text confusing in United States
I'm looking at phone numbers in OSM in the United States and and seeing some users incorrectly entering phone numbers starting with `+01`.  The placeholder text is `+31 42 123 4567`, which doesn't really look like an [NANP](https://en.wikipedia.org/wiki/North_American_Numbering_Plan) phone number.  The [preferred form of a phone number in the United States](http://wiki.openstreetmap.org/wiki/Key:phone) is `+1-503-555-6789`.

dobratzp:
This changes the behavior for setting the placeholder text of the phone field and attempts to resolve #3083.

Prior to this, the phone placeholder text was read from the users locale and not specific to the area of the world being edited.  Many of the locales use the default placeholder text `+31 42 123 4567`.  However, some of the locales change this phone number with a different country calling code.  For example, pt-BR uses `+55 11 0982 1098` so that users with Brazilian computers will see a more appropriate country code if they are editing in their home country.

It seems like it would make more sense to specify the placeholder text based on the area of the world being edited rather than the being different based on the users' language settings.  This changes the behavior so that if I add a phone number to an object in the United States, the placeholder is `+1-202-555-1234` and if I then add a phone number in the UK, I get the placeholder of `+44 207 123456` even if I happen to have set my computer's language to German.

I've added phone-placeholders.json which maps [ISO 3166-1 alpha-2 country codes](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) to default phone numbers.  I've taken the phone numbers from the locale files and I've added entries for fictitious phone numbers in the US, Canada, and the rest of the NANP countries.  The list is sorted in hierarchical order by [country calling code](https://en.wikipedia.org/wiki/List_of_country_calling_codes).  This list could certainly be expended to cover more countries, but I wasn't sure about the rules for creating fictitious phone numbers in each country.

I've updated input.js to query nominatim for the current country code before creating the phone field with placeholder text.  If the country is not listed, then the placeholder from the locale is used as a fallback.

Please bear with me as I'm new to this type of JavaScript code.  I've used address.js as a guide for these changes.


bhousel:
This looks excellent, thanks!  Using `address.js` as a guide is a good approach.  

Because you're only adjusting the placeholder attribute, you might be able to get by with always creating the `input` field on initialization and setting only the placeholder asynchronously in the taginfo callback.  This can simplify the code, avoiding the need for `isInitialized`.  Another issue with creating `input` asynchronously is that you need to add guard code in `i.focus()` because `input.node()` can be null.  (see `address.js`). 

Also I think I'd change `phone-placeholders` to `phone-formats` just to keep consistent with `address-formats`.

But overall it's really good.  Do you want to try to make those changes?  (If you'd prefer, I can do it too).

dobratzp:
Thanks for the feedback.  I'll take another look at the code and make some updates as you suggest.

Does it make sense to keep the phone specific changes in `input.js` rather than moving it to its own file (`tel.js`)?

bhousel:
Squashed and merged as:
7eb964a 2016-05-12 Switch only phone placeholder asynchronously <Bryan Housel>
2d9327c 2016-05-12 phone placeholder from country being edited <Peter Dobratz>

PR #3086, opened 2016-04-25
Look up languages by English or local names [closes #3023]

### LINKED ISSUE #3023: Language auto-completion for Multilingual name doesn't work the same for all languages
While selecting languages for adding multilingual name, the auto-completion for language names doesn't work as expected for all languages when English names are used. In below example, Hindi works but not Tamil and Telugu -

Latin script

But when respective language scripts are used to add languages, it works as expected by completing the correct language name -

Indic scripts

Since English is the default preferred language if the user hasn't set his own language preferences, I think the auto-completion should be working the same across all languages when typed in English.

mapmeld:
This was partly implemented on the placenames feature (adding a placename in "Chinese" shows you a list of localized options such as 中文)

The issue is that the box closes if it falls below d3.combobox minItems threshold . So if you start typing "ta" for Tamil, you will see many results whose English names contain "ta" (Tibetan at the top, Tamil close to the bottom if you scroll down). When you type "tam", there is only one matching option, so the combobox hides and your string cannot be completed.

This PR sets the minItems to zero, so you will see தமிழ் left over. If there are no results, the box will disappear.  Fix for #3023

### DIFF
--- a/js/id/ui/preset/localized.js
+++ b/js/id/ui/preset/localized.js
@@ -127,7 +127,7 @@ iD.ui.preset.localized = function(field, context) {
         innerWrap.attr('class', 'entry')
             .each(function() {
                 var wrap = d3.select(this);
-                var langcombo = d3.combobox().fetcher(fetcher);
+                var langcombo = d3.combobox().fetcher(fetcher).minItems(0);
 
                 var label = wrap.append('label')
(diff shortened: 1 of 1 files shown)


bhousel:
Good catch, thanks!

Issue #3007, opened 2016-02-28
Support RTL layout


SafaAlfulaij:
Application layout should support RTL layout for languages written from right to left.
This is what I mean:
LTR layout (Default):
![LTR layout - Default](http://4.bp.blogspot.com/-EzUf6UQhhus/UR6ATD23lZI/AAAAAAAAB88/_APE31dL4zw/s1600/setings-ltr.png)
RTL layout:
![RTL layout](http://2.bp.blogspot.com/-txMrX6kMxtY/UR6AW28Jx9I/AAAAAAAAB9E/DhVAtPhkPSI/s1600/settings-rtl.png)
Notice how the layout is mirrored.

HolgerJeromin:
Can this be closed?

bhousel:
Yes!

PR #3087, opened 2016-04-25
Add right-to-left language layout to CSS

mapmeld:
Addresses #3007 - right-to-left languages (Arabic and Persian are available in iD editor)

Here's what it looks like (using English locale and ```?rtl-test-rtl``` in the URL)

screen shot 2016-04-25 at 4 25 01 pm

I might be missing some tooltips... if someone is an Arabic or Persian speaker and can check this out with the language locale on, that'd be great

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -3399,6 +3399,16 @@ img.tile-removing {
   border-radius: 3px 0 0 3px;
 }
 
+/* search */
+[dir='rtl'] .feature-list-item .label {
+  text-align: right;
+}
+
+[dir='rtl'] .feature-list-item .entity-name {
+  padding-left: 0;
+  padding-right: 10px;
+}
+
 /* preset form */
 [dir='rtl'] .form-label {
   padding: 5px 10px 5px 0;
--- a/css/app.css
+++ b/css/app.css
@@ -3318,8 +3318,6 @@ img.tile-removing {
     margin: auto;
 }
 
-/* Mapillary
-------------------------------------------------------- */
 .mapillary-wrap {
     position: absolute;
     bottom: 30px;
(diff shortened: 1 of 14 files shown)


pnorman:
The layout of the bottom bar is unreversed, so the scale ends up underneath the background selection.

Should the triangle for Adjust Imagery offset be on the right?

The imagery_source_faq text (Where does this imagery come from?) appears to be wrong, as the question mark has ended on the wrong side. It might help to translate this text

1ec5:
The header bar that says “Edit feature” should be reversed, so that it contains the checkmark on the left, then “Edit feature” in the center, then finally the back button on the right. Similarly, the magnifying glass button next to the “Custom” option should be on the left, not the right.

> Should the triangle for Adjust Imagery offset be on the right?

Other disclosure triangles are also affected (“All fields”).

Any arrow that doesn’t indicate an absolute direction should be reversed. That includes the disclosure triangles as well as the back button next to “Edit feature”.

> The imagery_source_faq text (Where does this imagery come from?) appears to be wrong, as the question mark has ended on the wrong side.

That’s probably an artifact of displaying English as RTL.

Ideally the pie menu would be reversed, too, so that the trash can would appear down and to the right of the selected point, not down to the left.

mapmeld:
Yes, question mark and colon moving is a side effect of English in a dir="rtl" setting. For example Twitter hashtags appear on the right: https://twitter.com/hashtag/%D9%85%D8%B5%D8%B1?src=hash

These other points are good; I have some other work but I'll see when I can resume this

bhousel:
Wow, impressive work @mapmeld!   

Like @pnorman said, there are a few small UI things that need to be cleaned up, but overall it looks pretty good.  I'll start a list of things just from what I see in the screenshot, and we can test and add to it if needed.. This is really minor stuff that we could probably adjust easily in code.  

##### Supporting code
- [ ] maybe move textDirection testing code from `ui.js` into `context.loadLocale()` callback  
- [ ] maybe add `iD.detect().textDirection` property for other parts of iD to use

##### UI
- [ ] scale bar label placement
- [ ] bottom about bar links not reversed
- [ ] expand/collapse arrows
- [ ] radial menu not reversed
- [ ] are undo/redo buttons ok?
- [ ] background "custom" option "magnify" button placement
- [ ] punctuation looks funny, e.g. question mark and "(OpenStreetMap (Standard"
- [ ] entity editor preset picker navigation i.e. "◄ Edit Feature" 
- [ ] anything else absolutely positioned (minimap?)
- [ ] d3 transitions

1ec5:
Yeah, I don't think there's anything we can do about the way punctuation looks. English just isn't meant for RTL. But for RTL languages, it'll be just fine.

mapmeld:
ok, I've revised this to a commit which makes several of the suggested changes

screen shot 2016-04-26 at 5 42 41 pm

I'm not sure about the radial menu and two of the arrows mentioned: geolocation, undo/redo.  I hadn't even thought of the order of the radial menu... I think it might be more valuable for the icon order to match up with tutorials in other languages than to have them reversed

mapmeld:
I've made a couple of fixes so the tooltips and ```id.detect().textDirection``` work

Need help with a bug: when you draw a new node, it appears far left of where you clicked. That's because it's still adjusting the event X for the width of the menu, which is now on the right side of the window. Somehow this doesn't affect the code for selecting and dragging existing map data. Any suggestions?

mapmeld:
ok, I solved the drawing and menu issue with a change to the ```iD.util.fastMouse``` function. Any idea why ```getBoundingClientRect``` is confused about left and right boundaries?

pnorman:
I think we need to get some feedback from native users of RTL scripts

bhousel:
Thanks for the updates, I'm going to play around with this today and see about the `getBoundingClientRect` issues.   It could be because iD [caches the clientrect in the selection](https://github.com/openstreetmap/iD/blob/master/js/lib/d3.dimensions.js), to avoid triggering reflow.   We might need to invalidate this cache by calling something like `id.surface().dimensions(null)` whenever iD decides to switch LTR <-> RTL.

bhousel:
I think you fixed the mouse position issue?  It looks ok when I test it..

Here are some other things I saw:

- [ ] Wikipedia field link button:

screenshot 2016-05-13 15 06 10 

- [ ] address and raw tag editor fields seem to have styled border-left / border-right:

screenshot 2016-05-13 15 06 33
screenshot 2016-05-13 15 06 48

- [ ] help navigation arrows:

screenshot 2016-05-13 15 07 38

- [ ] GPX track nav button:

screenshot 2016-05-13 15 08 21

- [ ] ⌘I info pane:

screenshot 2016-05-13 15 10 09

- [ ] Preset category arrow:

screenshot 2016-05-13 15 10 59

- [ ] Save buttons and User icon:

screenshot 2016-05-13 15 43 29

- [ ] spinner placement:

screenshot 2016-05-13 15 46 07

- [ ] Most preset icons look fine, but Multipolygon is off:

screenshot 2016-05-13 15 46 46

- [ ] Change placement of GPX labels?:

screenshot 2016-05-13 15 50 41

- [ ] This text that says "top right" is wrong, but this is something that would be addressed in translation.

screenshot 2016-05-13 15 53 00

- [ ] images in walkthrough:

screenshot 2016-05-13 15 53 31

mapmeld:
@bhousel thanks for finding these!
yes I was able to fix the mouse position issue, but I don't know if there's an easier solution than the kind of hacky solution which I have now

mapmeld:
I'm making a demo of this PR on https://mapmeld.github.io/iD?rtl-test-rtl#background=Bing&map=17.32/10.33630/36.85405 so it can be reviewed by people on the OSM Tunisia list or similar community

Currently when I select an object on the map, I can't see any of its attribute fields, and as best I can tell, it's a problem in the master branch, too. Is there something I can do to patch this?

bhousel:
> Currently when I select an object on the map, I can't see any of its attribute fields, and as best I can tell, it's a problem in the master branch, too. Is there something I can do to patch this?

Thanks @mapmeld!   `master` branch is a bit messed up currently because of the big refactoring to use rollup as a packaging tool.  We're also upgrading to the latest d3, which is close to getting merged into master and will fix a bunch of the odd issues.

For now, your best bet is probably to apply this PR to the [1.9.x branch](https://github.com/openstreetmap/iD/tree/1.9.x)

mapmeld:
I prefer to make it compatible with 2.0 rather than get the JS working on 1.9 and then have to rewrite it when 2.0 is ready. Most of the changes are in CSS so it hasn't been so difficult.

I made a patch to my demo so that the name field is passed as an object instead of the string '[Object object]'. Works OK

bhousel:
> I prefer to make it compatible with 2.0 rather than get the JS working on 1.9 and then have to rewrite it when 2.0 is ready. Most of the changes are in CSS so it hasn't been so difficult.
> 
> I made a patch to my demo so that the name field is passed as an object instead of the string '[Object object]'. Works OK

Great!  Yes I'd prefer to merge it directly into v2 also - I just wasn't sure how quickly you needed to get something working for your demo.

marrouchi:
Hi Nick, very nice job :) 

You will find some glitches that i found (Tested on Chrome Version 49.0.2623.87 (64-bit)).

layers-parenthesis
road-name-reversed
road-search

mapmeld:
Hi @marrouchi - I've fixed up the search results and background layer list. The update is visible on https://mapmeld.github.io/iD/?rtl-test-rtl#background=Bing&map=17.32/10.33630/36.85405

Arabic labels in iD are broken in the current version on OSM, too. I tried a couple of different options but there isn't a quick fix. I think we should give it a try in a separate PR

marrouchi:
Yes perfect! Let me know if i can help in correcting the reversed arabic labels in OSM.

mapmeld:
@bhousel I've got this up-to-date with master branch. What do you think?

bhousel:
> @bhousel I've got this up-to-date with master branch. What do you think?

This looks really good!  I made a few minor tweaks today and merged it, and we can ticket any other items that come up in followup issues.  Thanks so much for your work on this @mapmeld!  🎉

mapmeld:
Thanks! As for the question of Arabic labels on ways @marrouchi , it's a bug with how Chrome displays text along a path using SVG textPath; it works in Firefox. So there isn't much iD can do (I tried adding lang, dir, other attributes and it wouldn't budge): 
http://stackoverflow.com/questions/31362171/how-to-use-svg-textpath-in-persianrtl-text
https://bugs.chromium.org/p/chromium/issues/detail?id=379149

Issue #2225, opened 2014-05-12
Right now it is impossible to select between multiple overlapping polygons


a1668753:
Description: if multiple polygons overlap you can only select one of them. This is missing feature in JOSM too.

Solution is to implement "what is under here?" feature found in some GIS systems. Prompt user to select particular polygon that contains point requested by user.

Bonus points if it will order polygons in reverse layer level, i.e.:
1. house polygon
2. landuse residential
3. district 
4. city boundary 
5. etc

You may also query not point but circle with small radius (.1 m) - this will be more usable with bad input devices (touch screens, poor mouse pouting).

Please note that this functionality shouldn't replace current "select" function but add new one. Usually GIS systems use left mouse button as "select" and place "what under here" in right click menu.

pbb72:
I found two workarounds for this problem:

- If the object has a name, use the Search function to find (and automatically select) it.

- Use the OpenStreetMap slippy map to find and highlight the object, and then switch to iD. The object will still be selected.

Not the best workarounds, and it will not help in every case, but hopefully this helps some people that are struggling.

waldyrious:
> Polygons are sorted smallest-on-top, so it's almost always possible to select the larger one underneath by clicking where there isn't overlap. If you have a specific situation where this isn't possible I'd like to take a look.

@jfirebaugh this is the case with tagging of buildings for 3D rendering, as described in https://wiki.openstreetmap.org/wiki/Simple_3D_buildings (emphasis mine):

> The building outline is a closed way or multipolygon tagged with `building=*`. It represents the area of land covered by the union of all parts of the building. The outline is also referred to as the building footprint.
> (...)
> **Cover the whole building outline** with `building:part=*` areas, tagged with their respective height and other attributes.

verdy-p:
Yes we need to be able to select alternate polygons. This is a frequent problems in iD, notably when these polygons are adjascent or we need to split a polygon in several parts, and still be able to adjust the largest one, not just the new smallest one!

The logic "using the smallest polygon" is simply corrupted, badly thought with a limited vision of what is needed. For this reason, iD users will frequenty break existing polygons by only fixing one, or deleting the nodes of parts still needed by the largest one; they redraw new objects and delete all existing ones, not caring about preserving their tags or necessary nodes with additional tags.

This is a *major* problem.

bhousel:
[...]

PR #3097, opened 2016-05-03
Indoor mode

zbycz:
Hi,

I am submitting a little larger pull request. I am not sure if its possible to be merged as is, but I did my best and will be happy to cooperate further to make it so.

I’ve been engaged in OpenStreetMap since 2008 and for the time being I’ve led many [workshops](http://osmap.cz/skoleni), written some school works and worked on getting the Czech community together on [openstreetmap.cz](http://openstreetmap.cz). Last summer I had to come up with a topic for my final engineer thesis, so I decided to use the effort already invested in indoor mapping and chose to update the iD editor accordingly. Let me elaborate...

Current work-in-progress proposal is called [Simple Indoor Tagging](http://wiki.openstreetmap.org/wiki/Simple_Indoor_Tagging). I didn’t want to rely on WIP, so I realized there is already an accepted solution – the [`level=*` tag](wiki.openstreetmap.org/wiki/Key:level). 

There are two challenges for indoor:
1. making underlying background map (this is addressed by Simple Indoor’s `indoor=*` area features)
2. display ways – the accessibility of places – and this kind of was the nature of OSM from the beginning. Until we had all the `landuse=*` areas, OSM started with highways + pois. 

Drawing `highways=*` for indoor feels very natural – corridors usually repeats every floor even in multi-storey buildings, it displays clear accessibility information, can be already used for routing and also works with parking highways. The logic is also very simple - when a `level=*` feature is in viewport - show “Indoor” button, and let the user filter on these. 

**Several examples** (my build):
- [Shopping mall in Prague](https://openstreetmap.cz/edit/#id=w382843342&map=18.06/14.40284/50.07310) - underground parking, main foot passages and few shops POIs
- [University building in Aachen](https://openstreetmap.cz/edit/#map=20.02/6.07485/50.77945) - complete room plans with Simple Indoor (older proposal version with optional `level=*` xor `repeat_on=*`). Another building [here](http://openstreetmap.cz/edit/#background=Bing&map=19.00/-1.63590/42.79960).
- [Berlin main railway station](https://openstreetmap.cz/edit/#map=18.74/13.41165/52.52148) - several underground floors + indoor=\* features
- [University campus Prague](https://openstreetmap.cz/edit/#map=18.21/14.39011/50.10425) - mapping in progress ;-)
- [Another shopping mall](https://openstreetmap.cz/edit/#map=19.20/14.46135/50.07867) - only POIs

Well, i think I made my point - let’s discuss the technical background :-)

---
### Some thoughts / questions:
1. Is it ok, to call it indoor _mode_? It doesn't fit with current “modes” terminology, where modes are exclusive.
2. I think the philosophy of iD is to show disabled buttons for “availible” actions. I chose to hide the button completely, where there isn't any `level=*` features on screen. Would you agree?
3. Current `level=*` allows only separate values by semicolon, I borrowed three useful tagging from SimpleIndoor. Is it ok to use these? Should i maybe make new official proposition?
   - ranges – `level=0-12` or `-3--1` for features spaning several floors. 
   - `repeat_on` – for repeating the same (possibly multilevel) feature on other floors - ie. doors,toilets
   - `min/max_level` – when set on building, the building spans through these levels. It helps as a visual cue, and also for the renderers.
4. I had to unclip the building-outline shape, because a building completely covered with `indoor=*` areas, couldn't be clicked anyway. Kind of addresses the issue #2225.
5. I submitted my original development commits, rebased to current master. Is it appropriate to squash the commits into one or few?
6. Although I tried to abide the current code style and wrote some test, tell me if i should change something globally. Easiest would be to comment the code in the specific commits.. 
### Something to do in future
- if agreed, I would like to add level indicator as shown here https://github.com/zbycz/iD/wiki/devnotes-zbycz – but rather use another svg layer instead of filters.
- once proposition is accepted – enable switching over floating point level values (ie. `level=1.5`). But even now - any string value could be written in the level chooser, so it “works”.
- add icon for ‘highway=elevator`and`indoor=corridor+stairs=yes` as these had to be distinguished easily.
- Unfortunately I found later, I developed very similar changes as Panier’s [id-indoor fork](https://git.framasoft.org/PanierAvide/iD-indoor). Although he added many more indoor presets, which could be eventually merged here.
- Create a level switcher in osmbuildings.org 3D viewer ;-)   .. well, distant future indeed.
### Changelog until now:
- added `iD.ui.IndoorMode` control ui
- added filtering using `features.indoor_different_levels`
- added `.indoor-mode` class for svg surface
- added general preset for `indoor=*` when indoor-mode enabled
- added several map styles for enabled `.indoor-mode`
- added `indoor_levels` as general field for building, using `iD.ui.preset.range`

btw, it was a pleasure to work with iD codebase, also d3 is so very amazing :-)

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -485,6 +485,12 @@ button.save.has-count .count::before {
     border-right:  6px solid rgba(255,255,255,.5);
 }
 
+.indoormode-level-combo .combobox-input,
+.indoormode-level-combo .combobox-caret {
+    height: 40px;
+    border-radius: 4px 0 0 4px;
+}
+
 /* Icons */
 
 .icon {
--- a/css/app.css
+++ b/css/app.css
@@ -485,10 +485,22 @@ button.save.has-count .count::before {
     border-right:  6px solid rgba(255,255,255,.5);
 }
 
-.indoormode-level-combo .combobox-input,
-.indoormode-level-combo .combobox-caret {
+.indoormode-level-combo .combobox-input {
     height: 40px;
     border-radius: 4px 0 0 4px;
+}
+.indoormode-level-combo .combobox-caret {
+    display: none;
+}
+.indoormode-control .spin-control {
+    height: 40px;
+    width: 41.6666%; /* force .col5 */
+    border-right: 1px solid rgba(0,0,0,.5);
+    margin-left: 0;
+    background: white;
+}
+.indoormode-control .spin-control button {
+    border-right: 0;
 }
 
 /* Icons */
--- a/css/app.css
+++ b/css/app.css
@@ -1071,7 +1071,7 @@ button.save.has-count .count::before {
     border-radius: 0 0 4px 4px;
     overflow: hidden;
 }
-.form-field > input:last-child {
+.form-field > input:nth-child(3) {
     border-left: 0;
 }
 
--- a/css/app.css
+++ b/css/app.css
@@ -1071,6 +1071,9 @@ button.save.has-count .count::before {
     border-radius: 0 0 4px 4px;
     overflow: hidden;
 }
+.form-field > input:last-child {
+    border-left: 0;
+}
 
 .form-field textarea {
     height: 65px;
--- a/css/app.css
+++ b/css/app.css
@@ -488,6 +488,11 @@ button.save.has-count .count::before {
 .indoormode-level-combo .combobox-input {
     height: 40px;
     border-radius: 4px 0 0 4px;
+    font-weight: bold;
+    text-align: center;
+}
+.indoormode-level-combo .combobox-input::-webkit-input-placeholder  {
+    color: #000;
 }
 .indoormode-level-combo .combobox-caret {
     display: none;
--- a/css/map.css
+++ b/css/map.css
@@ -1325,7 +1325,14 @@ path.fill.tag-indoor {
 .indoor-mode path.stroke.tag-indoor {
     stroke: rgb(169, 169, 169);
 }
+.indoor-mode text {
+    display: none;
+}
 
+.indoor-mode use.icon.areaicon.tag-building,
+.indoor-mode text.arealabel.tag-building {
+    display: none;
+}
 
 
 /* Labels / Markers */
--- a/css/map.css
+++ b/css/map.css
@@ -1311,7 +1311,7 @@ path.fill.tag-amenity-shelter {
 path.fill.tag-indoor {
     display: none;
 }
-.indoor-mode path.fill.tag-building {
+.indoor-mode path.fill.tag-building.tag-max_level {
     clip-path: none !important;
     stroke-width: 30px;
 }
@@ -1330,6 +1330,9 @@ path.fill.tag-indoor {
 .indoor-mode text.arealabel-halo.tag-building,
 .indoor-mode text.arealabel-halo.tag-amenity {
     display: none;
+}
+.indoor-mode.indoor-underground .way:not(.tag-level):not(.tag-building) {
+    opacity: 0.1;
 }
 
 
--- a/css/map.css
+++ b/css/map.css
@@ -1306,6 +1306,27 @@ path.fill.tag-amenity-shelter {
     background-color: rgba(224, 110, 95, 0.3);
 }
 
+.indoor-mode path.fill.tag-building {
+    clip-path: none !important;
+    stroke-width: 120px;
+}
+
+
+/* Indoor features */
+path.fill.tag-indoor {
+    display: none;
+}
+
+.indoor-mode path.fill.tag-indoor {
+    display: block;
+    fill: rgba(169, 169, 169, 0.3);
+    stroke-width: 30px;
+}
+.indoor-mode path.stroke.tag-indoor {
+    stroke: rgb(169, 169, 169);
+}
+
+
 
 /* Labels / Markers */
 
--- a/css/map.css
+++ b/css/map.css
@@ -1306,17 +1306,15 @@ path.fill.tag-amenity-shelter {
     background-color: rgba(224, 110, 95, 0.3);
 }
 
-.indoor-mode path.fill.tag-building {
-    clip-path: none !important;
-    stroke-width: 120px;
-}
-
 
 /* Indoor features */
 path.fill.tag-indoor {
     display: none;
 }
-
+.indoor-mode path.fill.tag-building {
+    clip-path: none !important;
+    stroke-width: 30px;
+}
 .indoor-mode path.fill.tag-indoor {
     display: block;
     fill: rgba(169, 169, 169, 0.3);
@@ -1325,12 +1323,9 @@ path.fill.tag-indoor {
 .indoor-mode path.stroke.tag-indoor {
     stroke: rgb(169, 169, 169);
 }
-.indoor-mode text {
-    display: none;
-}
-
 .indoor-mode use.icon.areaicon.tag-building,
-.indoor-mode text.arealabel.tag-building {
+.indoor-mode text.arealabel.tag-building,
+.indoor-mode text.arealabel-halo.tag-building {
     display: none;
 }
 
--- a/css/map.css
+++ b/css/map.css
@@ -1324,8 +1324,11 @@ path.fill.tag-indoor {
     stroke: rgb(169, 169, 169);
 }
 .indoor-mode use.icon.areaicon.tag-building,
+.indoor-mode use.icon.areaicon.tag-amenity,
 .indoor-mode text.arealabel.tag-building,
-.indoor-mode text.arealabel-halo.tag-building {
+.indoor-mode text.arealabel.tag-amenity,
+.indoor-mode text.arealabel-halo.tag-building,
+.indoor-mode text.arealabel-halo.tag-amenity {
     display: none;
 }
 
--- a/css/map.css
+++ b/css/map.css
@@ -1315,6 +1315,9 @@ path.fill.tag-indoor {
     clip-path: none !important;
     stroke-width: 30px;
 }
+.indoor-mode path.way.area.stroke.tag-building.tag-max_level {
+    stroke-width: 6px;
+}
 .indoor-mode path.fill.tag-indoor {
     display: block;
     fill: rgba(169, 169, 169, 0.3);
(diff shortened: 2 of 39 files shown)


bhousel:
Wow, impressive work @zbycz!    I really like this and would like to see this merged in some form. 

I'll need some time to think about how to present the UI in a way that doesn't clutter things up too much.  

I'll try to answer your questions now:
> Is it ok, to call it indoor mode? It doesn't fit with current “modes” terminology, where modes are exclusive.

I think it's ok to call it indoor mode.  (People will call it that anyway).  You're right it's not technically a "mode" like the other exclusive drawing modes.

> I think the philosophy of iD is to show disabled buttons for “available” actions. I chose to hide the button completely, where there isn't any level=* features on screen. Would you agree?

I'm leaning towards having the button smaller but available, maybe just an icon with tooltip, and it can transition open to reveal the indoor drawing controls when active.  (I'm struggling with this same problem on #2699 - it's cool and useful, but where to put it?)

> Current level=* allows only separate values by semicolon, I borrowed three useful tagging from SimpleIndoor. Is it ok to use these? Should i maybe make new official proposition?
> `ranges` – `level=0-12` or `-3--1` for features spaning several floors.
> `repeat_on` – for repeating the same (possibly multilevel) feature on other floors - ie. doors,toilets
> `min/max_level` – when set on building, the building spans through these levels. It helps as a visual cue, and also for the renderers.

The right thing to do is to make a proposal and loop in the OSM tagging list.  At this point, indoor editing is not solidly established, so I think you have a good chance of pushing the community in a sane direction.

> I had to unclip the building-outline shape, because a building completely covered with indoor=* areas, couldn't be clicked anyway. Kind of addresses the issue #2225.

Makes sense.. Yes I agree that in "indoor mode" the building outline should not render as an area, as this covers up the indoor features.

> I submitted my original development commits, rebased to current master. Is it appropriate to squash the commits into one or few?

It would be better to squash them into fewer commits, but I'm not super picky about this.

> Although I tried to abide the current code style and wrote some test, tell me if i should change something globally. Easiest would be to comment the code in the specific commits..

I'll try to review it in more detail, but what I saw at a quick glance looks great!  And the tests are much appreciated...

zbycz:
Thanks, @bhousel !

I will make the proposal next week, in the meantime I am looking forward to your code comments 👍  Once we're done, I will sqash the commits in fewer.
 
**ad button positioning)** the more buttons, the more the user think how complex the software is (aka Norman: The design of everyday things) - and I think iD should stay "simple". But also its a bad practise for controls to disappear unexpectedly.

I completely agree with a smaller icon expanding to full control. What about placing Indoor-mode button in a dropdown menu "Tools" on the right of the top toolbar. It could be left closed forever, or left toggled opened. It would be a nice drawer for more useful tools in the future (including the #2699).

bhousel:
> What about placing Indoor-mode button in a dropdown menu "Tools" on the right of the top toolbar. It could be left closed forever, or left toggled opened. It would be a nice drawer for more useful tools in the future

I've been playing around with making the UI more responsive and supporting something like a "drawer" for more commands or options on existing commands.

I don't really like it.. Maybe @samanpwbb has ideas?

The problem is it either bumps down the other things in the layout, or it overlaps them.  So we'd either need to 1. position the other UI stuff differently or 2. be ok about having them move (eh) or 3. something else. 

responsive_bar

css/app.css
```diff
@@ -1048,6 +1071,9 @@ button.save.has-count .count::before {
     border-radius: 0 0 4px 4px;
     overflow: hidden;
 }
+.form-field > input:nth-child(3) {
+    border-left: 0;
+}
```
zbycz: Because the ui.type.range needs second input without border. 
Would you suggest a better selector maybe?

css/map.css
```diff
@@ -1307,6 +1307,44 @@ path.fill.tag-amenity-shelter {
...
+    opacity: 0.1;
+}
+.indoor-mode use.icon.areaicon.tag-building,
+.indoor-mode use.icon.areaicon.tag-amenity,
+.indoor-mode text.arealabel.tag-building,
+.indoor-mode text.arealabel.tag-amenity,
+.indoor-mode text.arealabel-halo.tag-building,
+.indoor-mode text.arealabel-halo.tag-amenity {
+    display: none;
+}
```
zbycz: I`m not sure this is the right way to hide the building icon+label. It possibly breaks if other tags get precedence over building=* or amenity=* and shows its icon. Then it wouldnt be hidden.

Can you think of a better way, please? Or do you think its fine this way?

js/id/modes/add_area.js
```diff
@@ -11,13 +11,20 @@ iD.modes.AddArea = function(context) {
...
-        defaultTags = {area: 'yes'};
+            .on('startFromNode', startFromNode);
+
+    function defaultTags() {
+        var tags = {area: 'yes'};
+        if (context.indoor().enabled()) {
+            tags.level = context.indoor().level();
+        }
+        return tags;
+    }
```
zbycz: I dont particulary like this code, b/c it repeats in all add_* files. Still i dont think it should use other dependency.. So maybe its fine.

js/id/renderer/features.js
```diff
@@ -100,10 +100,34 @@ iD.Features = function(context) {
...
+        if (entity.tags.max_level && entity.tags.min_level) {
+            if (parseFloat(entity.tags.max_level) < current || parseFloat(entity.tags.min_level) > current)
+                return true;
+        }
+        else if (entity.tags.building || entity.tags['building:part']) {
+            if (current < 0 || current >= parseFloat(entity.tags['building:levels'] || 0))  //one level <=> level=0
+                return true;
+        }
+    });
+
```
zbycz: Now it shows all the time in the "map features" panel. But maybe it doesnt make sense to show this particular feature. It could be added/removed when indoor is toggled, but i like it this way better. (Also something is using it lower down in this file..)

Conclusion - maybe just hide the checkbox when indoor mode is not active?

js/id/renderer/features.js
```diff
@@ -360,7 +384,10 @@ iD.Features = function(context) {
 
     features.isHiddenChild = function(entity, resolver, geometry) {
         if (!_hidden.length) return false;
-        if (!entity.version || geometry === 'point') return false;
+        if ((!context.indoor().enabled() && entity.isNew()) || geometry === 'point') return false;
+
+        //hide vertices with different level
+        if (geometry === 'vertex' && context.indoor().enabled() && _features.indoor_different_level.filter(entity, resolver, geometry)) return true;
```
zbycz: Default behaviour is iD shows new entities regardless the feature filter. But here i need to hide different level always.. 

Also for the code clarity i replaced `!entity.version` for `entity.isNew()` – is it correct?

js/id/renderer/indoor.js
```diff
@@ -0,0 +1,96 @@
...
+                indoor.toggle();
+        }
+        return indoorLevel;
+    };
+
+    indoor.toggle = function () {
+        if (!context.surface()) { // hash calls before surface is initialized
+            setTimeout(indoor.toggle, 200);
+            return false;
+        }
```
zbycz: I dont like this solution, but havent found better..

js/id/svg/labels.js
```diff
@@ -166,7 +166,7 @@ iD.svg.Labels = function(projection, context) {
 
         icons.enter()
             .append('use')
-            .attr('class', 'icon areaicon')
+            .attr('class', function(d, i) { return 'icon areaicon ' + labels[i].classes; })
```
zbycz: To be able to hide icons with same css selectors as labels.

PanierAvide:
I'm glad to see the opportunity of integrating indoor editing in main iD version. Feel free to reuse any custom part of iD-indoor, like presets or [level parser](https://git.framasoft.org/PanierAvide/iD-indoor/blob/master/js/id/core/entity.js#L45) (which handles lot of level-related tags) ;)

samanpwbb:
Indoor mode would work well as a Map data filter:

screen shot 2016-05-10 at 2 40 40 pm

If the auto-tagging piece of this PR is mandatory, we can add a note about it in the filter panel:
screen shot 2016-05-10 at 2 46 06 pm

js/id/behavior/hash.js
```diff
@@ -36,6 +36,11 @@ iD.behavior.Hash = function(context) {
                 '/' + center[0].toFixed(precision) +
                 '/' + center[1].toFixed(precision);
 
+        if (context.indoor().enabled())
+            q.level = context.indoor().level();
+        else
+            delete q.level;
```
jirutka: @zbycz Please never ever omit curly braces around if-else, it’s generally considered as a bad practice and there are many documented cases when it caused a security issue.

js/id/ui/indoor_mode.js
```diff
@@ -0,0 +1,171 @@
+iD.ui.IndoorMode = function (context) {
+
+    var enterButton, indoorControl;
+
+    function createControls(selection) {
+
+        enterButton = selection.append('button')
+            .attr('class', 'indoormode-enter-button col12 ')
+            .on('click', toggleIndoor)
+            .call(buttonTooltip(t('indoor_mode.help'), '⌘⇧I'));
```
jirutka: I think that this shortcut will not work on other platforms than OS X…

js/id/renderer/indoor.js
```diff
@@ -0,0 +1,129 @@
...
+        bg.transition().style('opacity', d);
+        if (!iD.detect().opera) {
+            iD.util.setTransform(bg, 0, 0);
+        }
+    }
+
+
+    // ---- tagging related -----
+
+    var rangeRegExp = /^(-?\d+(?:\.\d+)?)(?:(-)(-?\d+(?:\.\d+)?)|(;-?\d(?:\.\d+)?)+)?$/;
```
jirutka: It’d be helpful to add some examples of inputs that this regexp parses…

zbycz:
Hi, just letting you know, we have moved a little further. I created [official proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/CoreIndoor) and submited [RFC to tagging list](https://lists.openstreetmap.org/pipermail/tagging/2017-February/031123.html). 

In near future i will create new pull-request based on current master, thus I am closing this one.

bhousel:
> Hi, just letting you know, we have moved a little further. I created official proposal and submited RFC to tagging list.
> 
> In near future i will create new pull-request based on current master, thus I am closing this one.

Great!  We are thinking a lot about plugins lately and I would love to find a good way to merge your work into iD to make indoor editing easy for everyone.

TheAdventurer64:
Progress?

quincylvania:
@TheAdventurer64 As I've mentioned before, any progress on an issue will be shown there. If you have any substantive questions or comments feel free to voice them.

PR #3104, opened 2016-05-09
Use npm scripts to build iD

kepta:
This commit liberates iD from makefile.

List of things related to this issue:
- `make translations` doesn't currently work
- Use of `http-server` npm package for `npm start`
- Use of config key in package.json to store all the file name

Let me know if I missed or did something wrong.

Also, these are the options I considered,
- Option A replace make with npm scripts, which is what this PR addresses
- Option B use gulp or any other task runner to simplify the build steps
- Option C stick to what we already have
 
@bhousel @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,23 @@
     "test": "test"
   },
   "scripts": {
-    "test": "eslint js/id && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html dot && make && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index_packaged.html dot"
+    "test": "eslint js/id && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html dot && make && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index_packaged.html dot",
+    "build:lodash": "lodash --development --output js/lib/lodash.js include=\"$npm_package_config_lodash_include\" exports=\"global,node\"",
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Is there an issue with the current build system that this PR is aiming to fix?

bhousel:
I would also prefer to drop the `make` prerequisite if we can.  

Getting a build environment set up on Windows is painful mostly because of this, because it requires Cygwin/MinGW or something similar, only to run some commands that node can run just as well.

bhousel:
> Option B use gulp or any other task runner to simplify the build steps

I think I like this option rather than just using npm scripts..  Any reasons not to use gulp?

It also means you don't need to store all the filenames in the config key..

Abbe98:
It would be great to be able to build iD without `make`, currently when I'm trapped with my Windows school computer I do it in a Virtual Machine... Cygwin/MinGW also have a hard time playing well with Node...

tmcw:
These `npm` scripts still use `cat`, `cp`, and other linuxisms, so they will not grant us Windows compatibility.

I'd prefer not to use gulp - it tends to introduce a lot of complexity with requiring a 'gulpy' version of every utility, and doesn't save complexity in any real way. I think it makes more sense to wait on this refactor until we refactor into modules, use browserify, and can make these build steps simpler.

kepta:
@tmcw what if we replace the Linux commands with their equivalent from npm packages ?

bhousel:
> I think it makes more sense to wait on this refactor until we refactor into modules, use browserify, and can make these build steps simpler.

^^ this is happening soon, so closing this PR

PR #3105, opened 2016-05-09
save button alerts user to save changes.

tanerochris:
Save button background color gradually changes as more changes are done by user.This warns user to save changes.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -451,10 +451,7 @@ button.secondary-action:hover {
 
 button.save.has-count {
     padding: 9px;
-    -webkit-transition: background-color 8s;
-    transition: background-color 8s;
 }
-
 
 button.save .count {
     display: none;
@@ -472,10 +469,7 @@ button.save.has-count .count {
     border-radius: 4px;
     margin: auto;
     margin-left: 8.3333%;
-    -webkit-transition: background-color 8s;
-    transition: background-color 8s;
 }
-
 
 button.save.has-count .count::before {
     content: "";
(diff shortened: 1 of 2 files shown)


bhousel:
Hmm that's a good start!  It's not quite what I had in mind..

Could you instead set an [inline style](https://github.com/mbostock/d3/wiki/Selections#style) like:

```js
button
   .style('background', getBackground(numChanges));
```

That will make the button change very slightly with every edit, and avoid the need to use the css transitions.  (It's not very good for it to transition slowly from the disabled style at 0 edits to enabled style at 1 edit.)


Some other things: 
1. Set `opacity: 0.5` on the [count](https://github.com/openstreetmap/iD/blob/master/css/app.css#L460), so it doesn't look like button once it gets background style.
2. Try to style the little triangle too.. [it's sneaky](https://github.com/openstreetmap/iD/blob/master/css/app.css#L485)

screenshot 2016-05-10 08 34 47
screenshot 2016-05-10 08 35 04

css/app.css
```diff
@@ -2042,17 +2071,6 @@ div.full-screen > button:hover {
     right: -10px;
 }
 
-.nudge-surface {
```
bhousel: Please don't remove this..

tanerochris:
I had some problems with actually updating the background colors across the range of colors @bhousel specified,tried getting a percentage of numChanges over our range of colors ,using hexadecimal codes ,but colors went out of range.I have done using rgb and I think it is what you want. Any other modification or suggestion you wish to make , I am waiting to have them. As for the tiny triangle ,could not think much of a style to it.I increased the opacity to make it more visible.

bhousel:
Hey @tanerochris thanks for this!  I made some changes and merged it today.

Your `getBackground()` function was pretty good, but I realized that d3 already includes a color interpolator that we can use, so I swapped that in for simplicity.   I should have checked this before when I gave my initial feedback...

```js
 function getBackground(numChanges) {
        var step;
        if (numChanges === 0) {
            return null;
        } else if (numChanges <= 50) {
            step = numChanges / 50;
            return d3.interpolateRgb('#fff', '#ff8')(step);  // white -> yellow
        } else {
            step = Math.min((numChanges - 50) / 50, 1.0);
            return d3.interpolateRgb('#ff8', '#f88')(step);  // yellow -> red
        }
    }
```

Some other things to keep in mind.
1. try to match the [code style we use](https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#javascript) (e.g. 4 space indent, no trailing spaces, etc).
2. try to avoid global or class variables where local variables will do..

Thanks again!

PR #3106, opened 2016-05-10
Fix greediness of autocompletion

### LINKED ISSUE #3077: textinput field are too greedy with autocompletion
I found this problem problem with the raw tag editor and the changeset comment dialog:

For example we have a `shop` entry and want to [prefix](http://wiki.openstreetmap.org/wiki/Lifecycle_prefix) it with `disused:`
I focus the cursor at start of the shop field and start typing:
First key (d) works, i have `dshop` but the second char kills the output with replacing the full text with 
`diameter_crown` while `ameter_crown` is marked. My old value shop is lost.

I often have this problem with the changeset comment dialog which replaces the current value, too.

My browser is the latest chrome

kepta:
(closes #3077)

### DIFF
--- a/js/lib/d3.combobox.js
+++ b/js/lib/d3.combobox.js
@@ -180,11 +180,16 @@ d3.combobox = function() {
             });
         }
 
+        function isCursorNotAtEnd() {
+          return input.property('selectionStart') < input.property('value').length;
+        }
+
         function autocomplete() {
             var v = caseSensitive ? value() : value().toLowerCase();
             idx = -1;
-            if (!v) return;
 
+            if (!v) return;
+            v = isCursorNotAtEnd() ? input.property('value'): v;
             for (var i = 0; i < suggestions.length; i++) {
                 var suggestion = suggestions[i].value,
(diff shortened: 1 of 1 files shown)


kepta:
@bhousel as you mentioned about the following 2 points
1. typing is happening anywhere other than at the end of the string or 
2. typing is overwriting a text selection

I think is a person is overwriting a text selection that would automatically become the 1st case.

what do you suggest ? and I am having trouble writing test case for this one, the mochajs doesn't autocomplete in any case.

bhousel:
@kepta I ended up changing this a little bit..
You had it entering that `autocomplete()` function but then completing with the current field value..  I changed it to avoid entering `autocomplete()` if possible.  

It's a little different in this way:
1. user creates a restaurant, then in `cuisine` field enters "chinese".
2. selects "chi" and types in "japa"

Your code would autocomplete them to the end of the field, my version leaves the cursor where they stopped typing.   It's really similar but I think it works a little more how people expect typeover selection to work.. What do you think?

kepta:
@bhousel I think it makes sense. You covered 2nd case I thought would not be possible.

PR #3110, opened 2016-05-13
Point url to a local macrocosm server

dereklieu:
1. Use d3.xhr instead of oauth.js to propagate changesets.
1. Puts user id, display name in changeset.
1. Points url at local macrocosm setup.


dereklieu:
Ignore that!

PR #3128, opened 2016-05-23
Mapillary js

kepta:
Continuing from #3014 , thanks to @peterneubauer.

### Things remaining

- [x] Toggle signs layer if and only if image layer is enabled.
- [ ] Look into https://github.com/mapillary/mapillary-js/issues/133
- [x] Async loading of CSS
- [x] Async loading of JS


kepta:
Build Process Issues:

1. The `mapillary-js.min.js` file is **1.1MB** in size. To give some perspective, the entire iD.js file is roughly 2MB in size. If I try to add `mapillary-js.min.js` to `js/lib` folder the make process takes 3-4 minutes to compile. Currently, I am directly copying the `mapillary-js.min.js` to avoid concatenating it to iD.js.

2. The svg dependencies of mapillaryJS require them to be hosted on the same level as that of the Mapillary css file. So I am directly copying these files to `dist` folder as of now :(.

@bhousel @peterneubauer your views on this?

peterneubauer:
Hi there,
I upgraded to MapillaryJS 1.3 in https://github.com/openstreetmap/iD/pull/3014/commits/a2f3063227cf943aee3ff7f918c25add04faf979 which makes the local artifacts unnecessary, also the SVGs can be removed.

Regarding the API call- the image you are looking for is from today, we have some lag in the system. Try anything older than a day, e.g. https://a.mapillary.com/v2/nav/im/1gE6B94DP0iDmUHjtRFKcQ?client_id=NzNRM2otQkR2SHJzaXJmNmdQWVQ0dzo1ZWYyMmYwNjdmNDdlNmVi

/peter

js/id/services/mapillary.js
```diff
@@ -206,9 +205,9 @@ iD.services.mapillary = function() {
         return iD.services.mapillary.sign_defs[country][type];
     };
 
-    mapillary.showThumbnail = function(imageKey, position) {
+    mapillary.showThumbnail = function(imageKey, position, interactive) {
         if (!imageKey) return;
-
+//         console.log('hello world');
```
pnorman: commented out debug comment left in here

peterneubauer:
I see that the in line mapillary icons and svg is still in here. Is that mandatory or could we just refer to the js and css in index.html for easier upgrade?

js/id/services/mapillary.js
kepta: oh thanks for catching that pnorman 👍

kepta:
@peterneubauer we generally keep js/css libraries in the code base itself.

@bhousel to asynchronously load css/js do we want to use their CDN or host a copy within iD.

bhousel:
> @bhousel when asynchronously loading css/js do we want to use their CDN or host a copy within iD?

Probably better to send this stuff to the CDN..  That helps the OSM folks a bit.

kepta:
The implementation is almost complete.
Here is a gif to show how it looks like currently 

id_demo_mapillary

bhousel:
The gif doesn't do it justice, but `centerEase` is now 👌 
Before it was just firing off 10 little map moves.. Now we smoothly interpolate to the new location, adjusting the projection each animation frame and queueing redraw for later. 

centerease

bhousel:
This is working alright!   
Thanks @kepta and @peterneubauer for your work on this...
Merging 👍 

mapillary-viewer

peterneubauer:
Super work @kepta ! Also, is this working even for street signs correctly?

peterneubauer:
I guess we should soon implement submission of changes to the locations/angles to Mapillary :)

bhousel:
> Also, is this working even for street signs correctly?

Yes, it's working for the signs layer also..  The one change we made is that now the photo layer must be activated before the signs layer can be activated.  

> I guess we should soon implement submission of changes to the locations/angles to Mapillary :)

Sure!  If there are some more API calls that would be useful to make, like for tagging or blurring or moving the markers, open a new issue so we can keep track of it..

Issue #2980, opened 2016-02-07
Option to make invisible relations visible


slhh:
The inspector should have a control to make invisible relations visible.
A graphical representation of routes seems to be easy. For other relation a net of lines connecting the members could be drawn.
If a relation has lost all of its members, it should be represented by a point. Of course this point and the relation must not be uploded. The user should get a warning with option to delete or rework instead. 

When the last member relationship is removed, the user should have an option to delete the relation or keep it temporarily as the virtual point. I believe this could simplify rework of relations.

jfirebaugh:
We already have issues covering route relations: #659, #1575. When you select a relation, all members are highlighted on the map.

iD isn't going to gain a bunch of features for advanced relation editing, as you seem to want. Sorry. That's just not it's domain.

Issue #659, opened 2013-02-06
indicate route membership with shadow


jfirebaugh:
P2 does this:

![image](https://f.cloud.github.com/assets/98601/132436/b25bd034-7084-11e2-9962-0935cace5027.png)

@samanpwbb what do you think? Useful?

NopMap:
I think there is an important feature missing here.

Potlatch2 is capable of highlighting ways which are members of routes on the map (without selecting anything). This is very helpful to see where route relations are "hidden". It is also possible to distinguish different types of route relations by different color highlighting.

It is probably necessary to switch the highlighting of different types of relations on and off in the sidebar, similar to the "Map Features" checkboxes. Most users will be focusing on one type of relations, e.g. when I am mapping hiking routes, I am not interested in seeing turn restrictions or public transportation.
The default should probably be "off" as new users are ususally not yet familiar with relations and would be confused by the additional highlighting.

NopMap:
In my fork I did a basic implementation of the behaviour possible in P2:
- membership of ways in hiking relations (foot or hiking) is translated in an attribute **route-hiking** on lines
- such lines are marked with a white semi-transparent highlight shadow unless selected or hovered upon, then the original selection shadow is used
- horse routes are similarly highlighted with a bluish shadow
This makes it easy to note the presence of hiking relations and follow their turns.
You can try it out on www.wanderreitkarte.de. My implementation statically supports two route relations to match the focus of my fork, for the iD master it probably should be user-configurable which routes are of interest.

bhousel:
I'll probably do something stylistically similar to what I just added in #4768 in the Turn Restriction editor.

It's not hard, but we need to render another stroke and add some design elements so that users can easily identify them - either by hovering on relations in the sidebar or when one is selected.

![tr_preview](https://user-images.githubusercontent.com/38784/37379792-b3c5921e-270b-11e8-8aef-62079f8295f5.gif)

(Going to tag this as a priority item)

NopMap:
I think there are two sides to this issue
- showing the associated members of a particular relation when selected or hovering, as shown above
- always marking all members of all instances of a type of route relation even if nothing is hovering/selected. E.g. all foot/hiking route relations. This is very helpful to understand routes.

Do you also plan a statical highlighting of the different route types?

bhousel:
> Do you also plan a statical highlighting of the different route types?

Not sure what you mean by this, but probably yes.

Issue #2946, opened 2016-01-30
Highlight relations/members when mouse is over item in all members or all relations


slhh:
- [x] When the mouse is over an item in all relations section the relation or all of its members should be highlighted in the map instead of the current selection.

- [X] When the mouse is over an item in all members section, the respective member should be highlighted in the map instead of the current selection. ( done in #5402 )

- [x] When the mouse is over an item of the selection list for adding of members/relations the same applies.

That would make identifying members and relations much easier.

sun-geo:
Nice :-)
Thnak you!

PR #3140, opened 2016-05-31
Highlight related members

Sinya:
I notice the issue:  indicate route membership with shadow #659 and related issues #2980 . I add 13 lines in raw_membership_editor.js. 

This make when selecting a relation, all members (including points, ways) are highlighted on the map.

highlight related members

### DIFF
--- a/js/id/ui/raw_membership_editor.js
+++ b/js/id/ui/raw_membership_editor.js
@@ -4,6 +4,19 @@ iD.ui.RawMembershipEditor = function(context) {
     function selectRelation(d) {
         d3.event.preventDefault();
         context.enter(iD.modes.Select(context, [d.relation.id]));
+
+        d.relation.members.forEach(function(member, index) {
+            if (member.type === 'way') {
+                var element = document.getElementsByClassName(member.id + ' ' + member.type + ' shadow');
+            }
+            else if (member.type === 'node') {
+                var element = document.getElementsByClassName(member.id + ' ' + member.type);
+            }
+            else {
+                return;
+            }
+            element[0].classList.add("selected");
+        });
     }
 
     function changeRole(d) {


bhousel:
I'm not sure if it's a good idea to class multiple things as `selected` without actually selecting them.. Doesn't this make the operations on radial menu misleading?

For example, in your screenshot above, it looks like 3 things are selected.  If the user clicks the trash can on the menu, they'd end up deleting only the relation, leaving the member items around.

Sinya:
Or I should new a class different from "select" that just has highlight effect? Thanks for suggestions.

slhh:
@bhousel, we have three different cases: 
- Relation members of a selected relation are already shown as `selected` in case of a multiselection (1 relation and at least one other feature selected). Unfortunately, this is the case where I would really agree on being misleading.
- In case of a single relation being selected showing members as `selected` might become an issue in future, but currently I would disaree on being misleading. The radial menue is not available in this case, and the feature type in the inspector is indicating that a relation is selected.
- One exception is a multipolygon relation because radial menu is available and the inspector is not clearly showing that a relation is selected. For example a multipolygon forest and a closed way forest are shown with the same feature type. Unfortunately, iD is already showing members as `selected` in the more misleading case of a selected multipolygon.

I suggest to always use a new `member` style to differentiate members of selected relations from selected objects. I believe differentiating `member` and `selected` style might be based on color only, because importance of the difference doesn't seem to be high enough to require considering colorblindness.

slhh:
@Sinya Your pull request does support identifying members as a whole, but identifying members individually would still be an issue. How about integrating the related issue #2946 ?
In addition, hovering a member of the selected relation on the map should highlight the respective member in the all members section.

bhousel:
from @Sinya 
> Thanks for suggestions :) I think I should new a class different from selected that just has highlight effect. And when selecting relation, it will lead editorPane to "relation edit mode". All members (including points, ways) are highlighted (maybe with color yellow?) with same relation. As the editorPane leave "relation edit mode" should cancel the highlight at the same time.

from @slhh 
> I suggest to always use a new member style to differentiate members of selected relations from selected objects. I believe differentiating member and selected style might be based on color only, because importance of the difference doesn't seem to be high enough to require considering colorblindness.

Yes, any attempt at relation highlighting should really use a different style than `selected`.. It probably needs additional svg strokes drawn too, so that it can coexist with the selected halo.  I think we should be able to find a way to make it recognizable by colorblind users.


from @slhh
> One exception is a multipolygon relation because radial menu is available and the inspector is not clearly showing that a relation is selected.

Yeah, I realized this after I wrote my comment above - iD already does style multipolygon members as selected.  Deleting a selected multipolygon with the trash can icon does leave the inners around.

NopMap:
I tried out this approach in my fork. Unfortunately it is not realistic for several reasons:

- use of select (easy to solve)
- it throws an exception if parts of the relation are not on screen (easy to solve)
- static approach: the visible members of the relation are highlighted permanently. Additional members appearing by zooming or scrolling are not highlighted. All members remain highlighted when the relation is unselected. Multiple relations selected one after the other add up.

slhh:
@NopMap 
> static approach: the visible members of the relation are highlighted permanently. Additional members appearing by zooming or scrolling are not highlighted. 

I see two possible solutions:
1. Let the code adding downloaded elements check whether they a member of a selected relation, and class them accordingly.
2. Let mode select download missing members in mode.enter(), and let the callback add the member to class "member".
In addition, we would likely need to prevent members of selected relations from getting unloaded. The class "member" might be useful to do this.

@bhousel How do you think about downloading missing members. It is clearly too much for relation highlighting alone, but would be benificial for example in the membership editor or for operations like splitting a multipolygon.
Maybe, we can add direct members only to the graph, but keep their out of view childs in a separate cache. Some operation would need to be able to access that cache, but D3 doesn't seem to need access.

> All members remain highlighted when the relation is unselected. Multiple relations selected one after the other add up.

Mode.exit() of mode selected does already clear the "selected" and "related" (related parent for vertex navigation) class and can also clear a new "member" class.

The code should be moved into modes/selected.js anyway. Otherwise, it doesn't work in case a relation is one of multiple selected elements, because the membership editor wouldn't be called.

Issue #3115, opened 2016-05-17
Add 'direction' key as default to some tags


homersimpsons:
It would be usefull for:
'highway'='stop','give_way','traffic_signals'
to add a field 'direction' with option: 'forward' or 'backward'
for 'traffic_signals' the key need to be: 'traffic_signal:direction' (don't know really why).

bhousel:
`highway=stop`
Yes, any street sign preset should probably have a forward/backward field.  Stop should probably also have an "all way" checkbox.

`highway=give_way`
We don't have a preset for this yet, but yeah, same as stop sign.

`highway=traffic_signals`
This makes no sense.. They kind of need to face all directions, otherwise what's the point?

homersimpsons:
Maybe you can create a "big" category named "sign" to regroup those 3
And same create a 'big' category named "traffic calming" to add all type of traffic calming (island, cushion, hump, bump etc...)

slhh:
> highway=traffic_signals
This makes no sense. They kind of need to face all directions, otherwise what's the point?

I agree in case of tagging traffic_signals on the junction node, but in case of mapping the traffic_signals of an larger junction separately, traffic_signals:direction makes sense, due to typically facing to a single direction for car traffic.

Preferably, traffic_signals:direction should be made available in case highway=traffic_signals is tagged on a virtex of a single non-oneway way.

bhousel:
These were added for `highway=stop` and `highway=give_way`... Not doing `traffic_signals` for now - I think it would confuse people more than help them tag correctly.

homersimpsons:
Yes, I read your comment on PR, can iD detect if it's on a intersection node (>=3 ways) to give the correct preset?

bhousel:
> can iD detect if it's on a intersection node (>=3 ways) to give the correct preset?

Not currently.  `vertex` geometry will tell us that the node is on a line, but not the degree of how many edges connect to it.

Codain:
Hi @bhousel 

Thank you for having put this fields for stops and give ways, very handy because it increases the quality of routings.

I would like to argue in favor of having a direction field proposed also for traffic signals. The only difference between a stop, a give way and a traffic signal is the type of rule imposed by the authorities and the physical "sign". Moreover in term of mapping in OSM it is the same mechanism.

[...]

Issue #3144, opened 2016-06-05
Improve Hotel Preset


g4rpch:
Dear all

Could you please improve the Hotel Preset with following tags:

- Stars
- Rooms
- Internet_Access
- Internet_Access:fee

Thanks a lot
G4rp

PR #3147, opened 2016-06-06
Update some presets.

homersimpsons:
A pull request to close a part of my request https://github.com/openstreetmap/iD/issues/3115
I added presets for sop and traffic_signals, and improve a translation (because it was impossible to make the difference beetween zebra crossing and non zebra one).

I edited on web github then I can't run Cmake...


data/presets/presets/tourism/hotel.json
```diff
@@ -4,7 +4,11 @@
         "operator",
         "address",
         "building_area",
-        "smoking"
+        "smoking",
+        "stars",
+        "rooms",
+        "internet_access"
+        "internet_access/fee"
```
homersimpsons: I don't know if this line is correct

homersimpsons:
I added others to match https://github.com/openstreetmap/iD/issues/3144
But again, no cmake

bhousel:
Thanks @homersimpsons!   I cherry picked about half of these commits and also added "rooms", "internet access", "internet access fee" fields to guest_house, hostel, motel presets ..

A few comments:
1. The field name should just be title cased words, no punctuation ("Internet Access Fee" not "Internet access fee ?")
2. You shouldn't update the `fr.json` translation file in the repository here.  These will be available for translation on [Transifex](https://www.transifex.com/ideditor/id-editor/dashboard/) after tonight.  Whatever is saved in the repository gets overwritten by the data pulled down from Transifex when we make a new release of iD.
3. I skipped the traffic signals direction tag because I don't think adding direction to traffic signals is useful.  Or rather - I get that it might be useful when people place the signals not at the junction, but I think it's more likely to confuse people than lead to better data in OSM.
4. I skipped `parallel_direction.json` because I was confused, but I will add it.

Thanks again!  I know you were editing on github - I do encourage you to try to set up a command line environment if you want to do more of this...  It will make things much easier for you!   Try out the instructions in [README](https://github.com/openstreetmap/iD/blob/master/README.md#prerequisites) and let me know if you get stuck..

homersimpsons:
Thanks you

1. Okay, I just looked some example to release those changes then maybe I missinterpreted some values
2. I'm okay but I want to fast correct, this was always annoying because I culdn't distinct zebra and non-zebra crossing preset
3. Okay, as I said in #3115 is it possible do detect if it's an intersection node to display the correct preset?
4. This was to mach cardinal and clock direction name. I didn't find a correct word to explain what is it. Feel free to change this :)

I know a few language and only base (java, html, css, php, python, c++) then I just fix some little issue that I can easily fix without trying to understand the whole code... But maybe later I will come back to help coding when I will get more knowledge about coding.

bhousel:
Great, thanks!

Issue #891, opened 2013-03-04
Refactor core/ into module


tmcw:
This is the 'holy grail' goal of post-launch refactoring - a module that's only reliant on d3 and/or a few other libraries and can be tested / used separately.

gmaclennan:
Any interest in revisiting this and https://github.com/openstreetmap/iD/pull/889?

We've been forking iD editor for different applications - offline, custom databases outside OSM - but the lack of modular structure makes modifications and re-use hard. It would be great to meet some of iD's [early goals](https://github.com/openstreetmap/iD/blob/master/ARCHITECTURE.md#core) of being "a reusable, modular library to kickstart other JavaScript-based tools for OpenStreetMap".

I know that [Moabi](http://moabi.org), [Development Seed](https://developmentseed.org/) and [American Red Cross](https://github.com/AmericanRedCross) are all looking at using customized iD for different use-cases. Perhaps we could coordinate an effort to refactor iD to make it more re-usable / modular / hackable?

@kriscarle @hallahan @smit1678

bhousel:
Yes let's make it happen.  Lets coordinate over email and pick a time in the next month or so for me to come down to DC for a few days of code sprint.

hallahan:
We're currently heads down on doing some mods on iD to handle fetching OSM XML from OpenMapKit Server, applying filters on that data, staging it, and then submitting it to either the main OSM or a Macrocosm API.

I'm making do with being as plugin-y like as I can. Breaking things down into modules and creating a plugin architecture, I think, would be a great idea. I'd like to make sure that our work leverages developments with this, because right now I have to inject in some less than ideal spots to make things happen.

We're working on a _staging concept_ where we sort of "fast forward" the history graph model to have a given save based on what came from our data edit source ready for submission. With that, we are having to do some interesting stuff with generating SHA-1 checksums on properties of OSM entities, since new elements have negative IDs and these will be rewritten by the OSM server they are submitted to.

https://github.com/AmericanRedCross/iD/tree/omk2macrocosm/js/id

tmcw:
Merging into newer issue #3118

PR #3148, opened 2016-06-06
Baby steps mode

tmcw:
This turns js/id/actions into CommonJS modules and creates js/id/actions.js with browserify. Refs #891 - this is an approach that would let us gradually transition to modules, by porting them into `modules/`, including them from `dist/modules`, and gradually introducing `require()`

cc @bhousel for the review


kepta:
thats awesome @tmcw ⭐️⭐️⭐️

tmcw:
Re-ran the tests and they pass now. @kepta and @bhousel want to review, see if this approach will let us do the gradual path? Given that other PRs will land soon, I want to avoid conflict city as soon as someone PRs anything related to actions.

bhousel:
I kind of feel weird generating the file into `dist/modules/actions.js`.  Anything we put into `dist`, people might build on and expect to stick around.

Would it make more sense to put it under something like `js/lib/modules` instead?   When this is done, most of `js/lib` will go away and be replaced by modules, right?

bhousel:
Also, per chat with @kepta, any thoughts on using [rollup](http://rollupjs.org/) instead of browserify?  It seems to support more kinds of module patterns, maybe could be useful if we ever want proper es6 modules.  Downside is it's newish and doesn't have great documentation. 

We aren't opinionated about this, just wondering if you have experience with it.

tmcw:
> Would it make more sense to put it under something like js/lib/modules instead? When this is done, most of js/lib will go away and be replaced by modules, right?

That was my first iteration, but we have `eslint` running on all files in `src`, and browserified bundles do not mix with eslint. We could make an exception, but it does underscore that a bundle is a generated file, not a source file.

> Also, per chat with kepta, any thoughts on using rollup instead of browserify?

I'm totally open to it as an experiment, but I have doubts. Rollup supports _only_ ES6 modules on the way in, and [ES6 modules, while great, have a very uncertain future](https://medium.com/@nodesource/es-modules-and-node-js-hard-choices-2b6995e4d491?utm_source=javascriptweekly&utm_medium=email). There's zero node.js support, and zero browser support, so any node-based or browser-based iD tests, even of pure, non-browser dependent code, would need transpilation. The same with if we ever wanted to provide iD components as standalone modules: they would always need to go through rollup in order to have a `main` entry that's loadable in node. And if we want to provide modules that provide 'es6 goodness' and let people take advantage of the tree-shaking abilities of rollup, we'd need both a transpiled `main` and non-transpiled `jsnext:main`.

The benefits of ES6 modules - slightly smaller filesize and enhanced export semantics - are real, but I'm not convinced that they outweigh the cost of integrating with any other JavaScript and the uncertainty around the standard's future.

But on the other hand ¯\_(ツ)_/¯ it could be fun

bhousel:
Thanks for helping with this @tmcw!  
I moved the intermediate file from `dist/modules` to `js/lib/id`.. For now this is less disruptive to downstream projects like openstreetmap-website.

Sticking with `browserify` for now - we can switch to something else later if it makes sense to do so.

PR #3153, opened 2016-06-08
D3 mousewheeled fix

### LINKED ISSUE #2773: Zooming under heavy load can translate the map center far away
I find that occasionally when I'm editing, and zooming in/out at a time when the browser is for whatever reason under heavy load (usually zoomed out a little and there's a lot of editable features on-screen), a small zoom/scroll gesture during that second of high-CPU will send my viewport to some random place on the earth.

e.g. just now it went from 33ºS 151ºE (in Sydney) to about 38ºS 178ºE (somewhere in the ocean, many thousands of kilometres away).

I get that zooming might go 'too far' in or out when scrolling, but to translate the viewport vast distances across the world doesn't really make sense. I'm seeing it quite a bit if I zoom in and out a lot, almost always ending up looking very closely at a random section of the ocean—at first I thought the rendering had just failed and would have to exit the editor and reload OSM, not realising it'd just moved to an area with just a lot of near-black tiles.

I've seen this on both Safari and Chrome (up-to-date versions of both as of July/August 2015) on Mac OS X.

kepta:
(closes #2773)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,7 +315,7 @@ D3_FILES = \
 
 d3:
 	node_modules/.bin/smash $(D3_FILES) > js/lib/d3.v3.js
-	@echo 'd3 rebuilt. Please reapply 7e2485d, 4da529f, and 223974d'
+	@echo 'd3 rebuilt. Please reapply 7e2485d, 4da529f, 223974d and 71a3d3e'
 
 lodash:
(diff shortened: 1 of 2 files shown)


bhousel:
Awesome analysis.. It will be great to have this bug fixed.. Thank you! 🙇

PR #3161, opened 2016-06-13
Start inter-requiring actions

tmcw:
Take another step towards modularity by using require, not globals, to have actions call each other

Well, this may give a point to @kepta's idea of using ES6 modules :) - iD does a _lot_ of circular dependencies, and that currently breaks this branch.

Other notes:

* Rollup's umd and iife patterns don't mix with the global `use strict` in iD + phantomjs, so it's removed from start.js for now.
* This upgrade phantomjs, making the bind polyfill unnecessary


tmcw:
Okay, ready for review @bhousel

kepta:
@tmcw I wanted to know is adding babel to the build step to use the goodness of `ES6` a good idea?

package.json
```diff
@@ -37,17 +37,19 @@
...
+    "rollup": "0.31.2",
     "sinon": "~1.6",
     "sinon-chai": "~2.3.1",
     "smash": "0.0",
     "svg-sprite": "1.2.19",
     "uglify-js": "~2.4.16",
     "xml2js": "~0.4.13",
-    "xmlbuilder": "~4.2.0"
+    "xmlbuilder": "~4.2.0",
+    "phantomjs-prebuilt": "2.1.7"
```
bhousel: I have to check this - `phantomjs-prebuilt` is something I thought about before but haven't tested.  If it works, that's great because it could simplify the steps for getting started (also testing on travis).

tmcw: Latest commit removes the travis manual build pattern - we'll see if that works.

bhousel:
> I wanted to know what do you think about adding a babel transpilation step to use the sugary ES6 🙃.

per chat - yes eventually but not today! 😄

PR #3163, opened 2016-06-14
Making spec/actions lint free

kepta:
To prepare for modularization of iD, I believe it would be good if we could lint all our test folder.

### DIFF
--- a/test/spec/actions/merge_remote_changes.js
+++ b/test/spec/actions/merge_remote_changes.js
@@ -1,4 +1,6 @@
-describe("iD.actions.MergeRemoteChanges", function () {
+/* global locale: true */
+/* eslint no-console: 0 */
+describe('iD.actions.MergeRemoteChanges', function () {
     var base = iD.Graph([
             iD.Node({id: 'a', loc: [1, 1], version: '1', tags: {foo: 'foo'}}),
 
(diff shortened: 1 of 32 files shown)


test/spec/actions/merge_remote_changes.js
```diff
@@ -1,4 +1,6 @@
-describe("iD.actions.MergeRemoteChanges", function () {
+/* global locale: true */
```
kepta: @bhousel I would want you to take a closer look at this diff.

test/spec/actions/merge_remote_changes.js
```diff
@@ -367,27 +369,27 @@ describe("iD.actions.MergeRemoteChanges", function () {
...
                     remoteTags = {foo: 'foo', bar: 'bar_remote'},  // same foo, added bar
                     remoteLoc = [2, 2],                            // moved node
                     local = base.entity('a').update({tags: localTags}),
                     remote = base.entity('a').update({tags: remoteTags, loc: remoteLoc, version: '2'}),
                     localGraph = makeGraph([local]),
                     remoteGraph = makeGraph([remote]),
-                    action = iD.actions.MergeRemoteChanges('a', localGraph, remoteGraph),
-                    result = action(localGraph);
+                    action = iD.actions.MergeRemoteChanges('a', localGraph, remoteGraph);
+                    action(localGraph);
```
kepta: @bhousel remote was unused var here, so I removed it.

test/spec/actions/merge_remote_changes.js
bhousel: Yes, I think that's right..

bhousel:
This is awesome!  I've wanted to do it for a while..

Can you also make sure the spec files get linted [when `npm run test` is called](https://github.com/openstreetmap/iD/blob/master/package.json#L12)?

kepta:
@bhousel sure.

kepta:
@bhousel updated the PR.

bhousel:
npm run commands look great, merge it if you are happy with it.. :shipit:

Issue #3118, opened 2016-05-18
Refactor iD into ES6 modules (phase 1)


bhousel:
---

Brief ( @tmcw )

- The work to modularize iD is underway and we are welcoming contributors!
- The approach is incremental so that we don't have to freeze iD development.
- We're using ES6 modules and rollup to bundle them.

Step 1 was https://github.com/openstreetmap/iD/pull/3148 and step 2 was https://github.com/openstreetmap/iD/pull/3161 .

We are using ES6 modules because iD's codebase includes many circular dependencies, for instance: `DeleteMultiple -> DeleteWay -> DeleteRelation -> DeleteMultiple`.

**How you can help**: if this feature push is important to you, please invest the time helping us make it real. This is an open source project and relies on its contributors. This is a great opportunity to get involved with OpenStreetMap and learn about this module system by following existing patterns.

**How to help**:

iD is split into multiple directories under [js/id/](https://github.com/openstreetmap/iD/tree/master/js/id):

* [x] actions #3161
* [X] behavior #3178
* [x] core #3174
* [x] geo #3172 
* [x] modes #3169
* [x] operations #3173 
* [x] presets #3168
* [X] renderer #3178
* [x] services #3175
* [x] svg #3177 
* [x] ui *(has PR #3167)*
* [x] util #3166
* [x] validations #3165 

These folders contain components of iD that are attached to each folder's namespace: for instance, everything under operations/ ends up as parts of iD.operations. Luckily we've followed a good pattern for how each component works: each one attaches a function to the namespace. With ES6 modules, we're instead going to `export` each function, and then import them.

So the flow is:

* Move a directory full of files from js/id/namespaceName to modules/namespaceName
* Convert each file to ES6 export and import style
* Add a new bit in the [Makefile](https://github.com/openstreetmap/iD/blob/master/Makefile) patterned off of the `actions.js` generation that generates a bundle file for the newly-modularized section of iD
* Update [test/index.html](https://github.com/openstreetmap/iD/blob/master/test/index.html) to point to this bundled file instead of the individual files from that part of iD.
* Run tests
* Submit a PR!

This is an incremental process, so this flow does not include everything that will eventually be part of this effort, only the things we can parallelize.

_After_ the initial batch of modularization is done and all the checkboxes above are checked, then we will move from multiple bundle files to one, and then move dependencies like lodash and d3 from globals to `import`'ed modules.

**How to start**:

[...]

PR #3165, opened 2016-06-14
refactor validations into ES6 modules for #3118

tyrasd:


tmcw:
🎉  Awesome! I owe you one ☕

PR #3167, opened 2016-06-14
Modularize iD.ui

tyrasd:
re: #3118


modules/ui/confirm.js
```diff
@@ -1,5 +1,7 @@
-iD.ui.confirm = function(selection) {
-    var modal = iD.ui.modal(selection);
+import { modal as modalModule } from './modal';
```
tyrasd: I had to do a named import of the "modal" module in a couple of places because of collisions with the local `modal` variable (see below)… :unamused:

Makefile
```diff
@@ -43,11 +44,18 @@ $(BUILDJS_TARGETS): $(BUILDJS_SOURCES) build.js
...
 
 MODULE_TARGETS = \
-	js/lib/id/actions.js
+	js/lib/id/actions.js \
+	js/lib/id/ui.js
 
 js/lib/id/actions.js: modules/
 	node_modules/.bin/rollup -f umd -n iD.actions modules/actions/index.js --no-strict > $@
 
+js/lib/id/ui.js: modules/ui/*
```
tyrasd: I think the actions.js target above also needs proper dependencies defined, i.e. `modules/actions/*`, right?

tyrasd:
Some remarks (in addition to the stuff I've inlined to the commit above):
* I left most of `js/id/ui.js` as is because I didn't find a solution to make it work in the rollup bundle (`export default` did't work, unfortunately). I think it makes most sense to refactor these higher layer modules later on, doesn't it?
* I wasn't sure what to do with the nested `ui/preset` and `ui/intro` directories. For now, I've added an additional bundle target in the Makefile which contains only the preset ui fields that have tests defined for. That's very hacky, but I guess the tests are going to be refactored to use the ES6 modules directly later on, right?

kepta:
@bhousel @tmcw I picked up @tyrasd 's work.
I have broken down the `ui module` into 4 sub modules.
```Makefile
js/lib/id/ui/index.js: $(shell find modules/ui -type f)
	@rm -f $@
	node_modules/.bin/rollup -f umd -n iD modules/ui/ui.js --no-strict -o $@

js/lib/id/ui/core.js: $(shell find modules/ui/core -type f)
	@rm -f $@
	node_modules/.bin/rollup -f umd -n iD.ui modules/ui/core/index.js --no-strict -o $@

js/lib/id/ui/intro.js: $(shell find modules/ui/intro -type f)
	@rm -f $@
	node_modules/.bin/rollup -f umd -n iD.ui.intro modules/ui/intro/index.js --no-strict -o $@

js/lib/id/ui/preset.js: $(shell find modules/ui/preset -type f)
	@rm -f $@
	node_modules/.bin/rollup -f umd -n iD.ui.preset modules/ui/preset/index.js --no-strict -o $@
```

- `modules/ui/ui.js` used to be `js/iD/ui.js`, since it attached to global iD, I made a separate `rollup` build step.

-  `modules/ui/core` is composed of all the files inside `js/iD/ui` and it attaches itself to `iD.ui`.

- `modules/ui/intro` & `modules/ui/preset` folders attach to `iD.ui.intro` & `iD.ui.preset` respectively.

Let me know what you guys think of this PR.

tmcw:
Approach lgtm! Stage 2 unlocked

PR #3168, opened 2016-06-15
migrate presets to ES6 modules for #3118

beaugunderson:
The only change I had to make to namespaces was that `iD.presets` itself used to be a function; I moved it to `iD.presets.presets`.

### DIFF
--- a/modules/presets/category.js
+++ b/modules/presets/category.js
@@ -1,4 +1,4 @@
-iD.presets.Category = function(id, category, all) {
+export function Category(id, category, all) {
     category = _.clone(category);
 
     category.id = id;
@@ -22,4 +22,4 @@ iD.presets.Category = function(id, category, all) {
     };
 
     return category;
-};
+}
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -1,0 +1,5 @@
+export { Category } from './category.js';
+export { Collection } from './collection.js';
+export { Field } from './field.js';
+export { Preset } from './preset.js';
+export { presets } from './presets.js';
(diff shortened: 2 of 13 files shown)


tmcw:
👏  Awesome work, thanks @beaugunderson !

modules/presets/index.js
```diff
@@ -0,0 +1,5 @@
+export { Category } from './category.js';
+export { Collection } from './collection.js';
+export { Field } from './field.js';
+export { Preset } from './preset.js';
+export { presets } from './presets.js';
```
kepta: @beaugunderson why do we have lower case `presets` ?

modules/presets/category.js
```diff
@@ -1,4 +1,4 @@
-iD.presets.Category = function(id, category, all) {
+export function Category(id, category, all) {
     category = _.clone(category);
```
kepta: @tmcw should we be using `iD.presets.Collection` internally?

modules/presets/index.js
beaugunderson: as I mentioned in the PR description, `iD.presets` was itself a function; it also contained the other functions as properties, e.g. `iD.presets.Field`, etc.

I didn't think it was wise (or possible without being kludge-y) to carry that forward to the ES6 modules approach, so I moved it to the same level as the other functions... maybe I'm wrong though :)

kepta: Imho I feel we should have iD.presets modularized to 'presets/index.js'.
I have seen a general pattern in es6 modules, if you import a folder let say 'presets' the bundler automatically looks for 'presets/index.js'. So it would be better have all the functionality with iD.presets inside 'index.js'. And if you want you can have another file inside this folder called 'presets/presets' which has the functionality of iD.presets.presets or you can simply not have this file and do all the stuff in index.js. 

Let me know what are your views regarding this.

PR #3170, opened 2016-06-15
Fix spec lint

kepta:
This PR will finally allow us to lint our tests.😋


kepta:
@bhousel before you look `package.json` to get disappointed like last PR, 
we have `eslint js/id test/spec` which covers the entire spec folder :P .

bhousel:
Awesome!   That went faster than I expected - feel free to merge it.. 👍

PR #3171, opened 2016-06-15
Allow linter to do static analysis on modules

kepta:
Currently linter doesn't support static analysis of modules( like checking if they actually exist or not).

This PR adds [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) to dev dependency and will allow us to throw error if a module is malformed / undefined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "d3": "3.5.5",
     "editor-layer-index": "git://github.com/osmlab/editor-layer-index.git#gh-pages",
     "eslint": "^2.12.0",
+    "eslint-plugin-import": "^1.8.1",
     "glob": "~3.1.21",
     "happen": "0.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,8 +48,7 @@
     "svg-sprite": "1.2.19",
     "uglify-js": "~2.4.16",
     "xml2js": "~0.4.13",
-    "xmlbuilder": "~4.2.0",
```
kepta: for some unknown reasons my `npm` loves to jumble package.json

bhousel: alphabetical listing is nicer anyway... 🔤

bhousel:
Sounds good to me, thanks!

package.json
pnorman: but it's all the same to json ;)

PR #3174, opened 2016-06-15
Modularize iD.core

kepta:
@tmcw @bhousel review?

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,6 +34,7 @@
         <script src='js/lib/marked.js'></script>
 
         <script src='js/id/id.js'></script>
+        <script src='js/id/id/core.js'></script>
         <script src='js/lib/id/actions.js'></script>
         <script src='js/lib/id/modes.js'></script>
         <script src='js/lib/id/presets.js'></script>
@@ -158,17 +159,6 @@
         <script src='js/id/behavior/paste.js'></script>
         <script src='js/id/behavior/select.js'></script>
         <script src='js/id/behavior/tail.js'></script>
-
-        <script src='js/id/core/connection.js'></script>
-        <script src='js/id/core/difference.js'></script>
-        <script src='js/id/core/entity.js'></script>
-        <script src='js/id/core/graph.js'></script>
-        <script src='js/id/core/history.js'></script>
-        <script src='js/id/core/node.js'></script>
-        <script src='js/id/core/relation.js'></script>
-        <script src='js/id/core/way.js'></script>
-        <script src='js/id/core/tree.js'></script>
-        <script src='js/id/core/tags.js'></script>
 
         <script src='js/lib/locale.js'></script>
         <script src='data/introGraph.js'></script>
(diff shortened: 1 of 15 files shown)


index.html
```diff
@@ -34,6 +34,7 @@
         <script src='js/lib/marked.js'></script>
 
         <script src='js/id/id.js'></script>
+        <script src='js/id/id/core.js'></script>
```
bhousel: js/lib/id

modules/core/index.js
```diff
@@ -0,0 +1,10 @@
+export { Connection } from './connection';
+export { Difference } from './difference';
+export { Entity } from './entity';
+export { Graph } from './graph';
+export { History } from './history';
+export { iD_Node as Node } from './node';
```
bhousel: Why is this one `iD_Node`?  Can it be just `Node` to be consistent with `Way` and `Relation`?

kepta: Umm I am worried about this guy conflicting with the browser [Node API](
https://developer.mozilla.org/en/docs/Web/API/Node).

bhousel: Oh good point.. Maybe can you export them as `OSMNode`, `OSMWay`, `OSMRelation`?

kepta: ya sure, but eventually we still are reexporting them as `Node`, `Way`, `Relation`, because most of our code uses `iD.Node`. 
```Javascript
export { OSMNode as Node } from './node';
````
or do you want me to do it 
```Javascript
export { OSMNode } from './node';
````
and refactor all the iD.Node to iD.OSMNode, that can be a big task...

Comment (bhousel) [inline modules/core/index.js]: I prefer the first one: 
```js
export { OSMNode as Node } from './node';
```
as long as `Node` `Way` and `Relation` "look" the same, as these are the OSM primitives.

I agree that refactoring all the iD.Node would be a pain and not worth it now.

But why is it ok to export them as `Node` here but not in `./node.js`

kepta: umm good point, because I believe the only thing using this export
```Javascript
export { OSMNode as Node } from './node';
```
is rollup which will eventually make this function `iD.Node`, so I had to re-export it this way.
Although, I don't know the best solution. It feels like a hack. 
what do you suggest ?

bhousel: Does anything actually break if you export it as `Node` in both places?   I feel like the modules should scope their exports somehow so that they don't really conflict with anything in practice.  

(e.g. I think Browserify would that `Node` into something like `11_Node`)

kepta: Not it didn't break anything. But it might break if we want the use `Node API` inside the scope of `./node`. But that is rarely going to happen. So, I guess we should stick to `Node` then...

bhousel: Yes, and in that case it would be up to `import`ing code to rename that to avoid the conflict, right?
```js
import {Node as 'OSMNode'} from '/js/lib/id/node'
```
(still trying to wrap my head around ES6 modules)

kepta:
@bhousel @tmcw updated the code. This PR might need a careful review since it is the `core` of iD.

bhousel:
Looks ok to me!  I'll resolve conflicts and merge..

index.html
kepta: oops !

PR #3175, opened 2016-06-15
Module for iD.services

bhousel:
Also I did some Makefile cleanup, because it was not re-running the build steps after changing source files.

Tests fail when run on commandline with mocha, but pass when run in browser.
No idea why.. It looks like not everything is getting put in the global namespace like it should.

Any ideas @kepta @tmcw ?


bhousel:
I think I figured it out.  I can't use module level variables (yet).

tmcw:
I think it was because PhantomJS doesn't reliably [support const](https://github.com/karma-runner/karma/issues/1621)? Regardless, PR looks 👍

PR #3177, opened 2016-06-16
refactor svg into ES6 modules for #3118

beaugunderson:
Also moved functions from `js/id/svg.js` into their own modules.


tmcw:
👍 Awesome, thanks! I owe you two cups of coffee next time I'm in Seattle

PR #3181, opened 2016-06-17
Add sidewalk preset to Path presets category

willemarcel:
sidewalk

I think it's important to have this preset in the Path Category. More information about this tag: 

* https://www.mapbox.com/blog/mapping-sidewalks/
* https://wiki.openstreetmap.org/wiki/Sidewalks


bhousel:
Thanks @willemarcel, I agree!
I added Pedestrian Crosswalk too and merged..

PR #3182, opened 2016-06-18
Ext modules for core, actions, geo

kepta:
After a discussion with @bhousel, we decided it would be a better idea for all of us to do phase 2 in steps.
I ran tests locally and will continue to do more ui testing on browser.
cc: @tmcw


tmcw:
@kepta this fails with a reference to `iD.geo.RawMercator` - should the `work-in-progress` tag still be on it?

kepta:
@tmcw that was unintended, maybe when I break the PR it slipped in. It should be a small fix.

PR #3184, opened 2016-06-22
External modules part 2/3

kepta:
In this PR I am targettting behavior, modes, util, operations, presets, validations, renderer.
I removed svg from this so that we can do the ui, svg together in the next PR.

I am still not sure on how to go about with services.
@tmcw @bhousel what do you guys say?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -52,7 +52,7 @@
 
     "globals": {
         "d3": false,
-        "iD": false,
+        # "iD": false,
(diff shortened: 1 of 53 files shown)


kepta:
why is travis not detecting this?

bhousel:
Not sure why Travis skipped it, but I tested a bit with it and it works fine.  Going to merge it in as is, so we can continue to work incrementally..

PR #3187, opened 2016-06-22
When setting a form’s background colour, also set the text colour

### LINKED ISSUE #3100: Text is hard to read on Firefox on Linux with dark GTK3 theme
I’m running Firefox on Linux using a dark GTK3 theme. This means without any styling a form input would have white text on a dark background.

This is what trying to edit my local primary school looks like: 
screenshot from 2016-05-04 18-07-40

jonnybarnes:
This should fix #3100, where the background colour of form inputs had been set to white, and firefox was then using the dark theme’s light text colour which had poor contrast.

A solution is to also define a sane `color` property.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -159,6 +159,7 @@ input[type=url],
 input[type=tel],
 input[type=email] {
     background-color: white;
+    color: #333;
     border:1px solid #ccc;
     padding:5px 10px;
     height:30px;


bhousel:
Thanks this makes sense now that I see it.  By not setting color, we were leaving it up to the browser's builtin stylesheet.

Issue #1756, opened 2013-08-27
Remember selected imagery between sessions


pauliuszaleckas:
Each time I start editing I have to change background imagery.
It would be nice if iD remembered last selected imagery and set it automatically if it is available for area being edited.

afmenez:
Part of this was recently implemented on  #2566, the last imagery is remembered between sessions.  Now only the auto-select is missing.

bhousel:
Oh thanks!  I didn't even see this issue. I suppose we can auto-select the user's preferred imagery, first checking that:
* it is valid for where the user happens to be looking, and
* an imagery parameter has not been supplied in the url hash

bhousel:
This was done in #4720   🎉

PR #3188, opened 2016-06-22
Save/restore background

edpop:
User did not have to choose background again or retype custom

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -188,8 +188,9 @@ export function Background(context) {
         }
 
         var q = stringQs(location.hash.substring(1)),
-            chosen = q.background || q.layer || context.storage('background'),
+            chosen = q.background || q.layer,
             extent = parseMap(q.map),
+            validRecent,
             best;
 
         backgroundSources = imagery.map(function(source) {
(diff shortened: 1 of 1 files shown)


modules/renderer/background.js
```diff
@@ -187,7 +188,7 @@ export function Background(context) {
         }
 
         var q = stringQs(location.hash.substring(1)),
-            chosen = q.background || q.layer,
+            chosen = q.background || q.layer || context.storage('background'),
```
bhousel: The problem with putting it here is that now the user's preferred background might not be visible where they started editing the map.  This is especially a problem if a user is entering iD via a task manager or QA tool, and editing in a place that is not where they normally use iD.

Could you move the `context.storage` code down to the block where `best` gets set... Then in there set a variable called `recent` that contains this BackgroundSource only if it is valid in the current visible extent.

Then a few lines later do `background.baseLayerSource(findSource(chosen) || recent || best || ...etc)`

edpop:
Ok, now there some much code.
On the other hand I can add parameter extent to "findSource" function if you like.

edpop:
There is a wrong behaviour when we read a custom background from storage. We can't check it for extent. What should to do in this case?

modules/renderer/background.js
```diff
@@ -202,13 +204,18 @@ export function Background(context) {
         backgroundSources.unshift(BackgroundSource.None());
 
         if (!chosen && extent) {
-            best = _.find(this.sources(extent), function(s) { return s.best(); });
+            var recent = findSource(context.storage('background'));
```
edpop: context.storage('background') can be /^custom:/

bhousel:
Nice work!  this will close #1756..
I think it's ok to restore the user's last used imagery, as long as it is valid where they are viewing the map.  (You're correct in your comment above that we can't do this validity check on a custom imagery source.)

Issue #3180, opened 2016-06-17
ES6 Modules: Phase 3


tmcw:
Phase 2 is documented in https://github.com/openstreetmap/iD/issues/3179

---

In Phase 2, we condense the modules into one bundle and eliminate reliance on the iD global. Phase 3, we eliminate reliance on libs: we stop using a lodash global, same with ohauth and other lil libraries. End of the day, we have no global vars whatsoever, we remove vendored libraries from the source tree and instead use npm to install them.

This happens after phase 2 and can be @kepta or whoever wants to take it on :)

bhousel:
Vendored libraries TODO:

- [ ] bootstrap-tooltip.js
- [ ] d3-compat.js
- [ ] d3.combobox.js
- [ ] d3.curtain.js
- [ ] d3.dimensions.js
- [ ] d3.geo.tile.js
- [ ] d3.jsonp.js
- [ ] d3.keybinding.js
- [ ] d3.one.js
- [ ] d3.trigger.js
- [ ] d3.v3.js  🌊 
- [ ] d3.value.js
- [ ] diff3.js
- [ ] jxon.js
- [ ] locale.js
- [ ] lodash.js 🌊 
- [ ] marked.js
- [ ] osmauth.js
- [X] rbush.js (#3196)
- [X] sexagesimal.js (#3191)
- [ ] togeojson.js

bhousel:
I think all the remaining items were rolled into #3344 and this can finally be closed..
Thoughts @tmcw ?

tmcw:
👍 Yep, the future is now!

PR #3191, opened 2016-06-24
Eliminate vendored sexagesimal library

bhousel:
Starting to remove the vendored libraries in js/lib and replace them with modules installed via npm..
Sexagesimal is a simple one to start with - how did I do?  

cc @tmcw @kepta  
re: #3180


tmcw:
LGTM! As iD starts using modules, we'll need to come to an approach for semver ranges - the new default is the ^ range, this PR uses ~. My gut feeling is shifting towards _no semver specifiers at all_ since semver-automatic upgrading is rarely desired and incidental automatic upgrades are rarely bulletproof.

bhousel:
Good point about semver ranges.   I agree I'd rather not use them at all, because I don't like surprises.

@tmcw can I drop all the range specifiers and set up [Greenkeeper](https://greenkeeper.io/)?

tmcw:
👍 sounds like a plan. I haven't used Greenkeeper before, interested to hear how it goes.

kepta:
Green keeper looks awesome ✌

PR #3194, opened 2016-06-25
Trunk ui

kepta:
Ok so this is the dreaded ui submodule. 
I really want you guys to rigorously test and scrutinise this PR as much as possible.

cc: @bhousel @tmcw @tyrasd


modules/ui/core/intro.js
```diff
@@ -1,3 +1,8 @@
+import { Icon } from '../../svg/index';
+import { Entity, Graph } from '../../core/index';
+import { Browse } from '../../modes/index';
+import { Intro } from '../intro/index';
```
kepta: I couldn't think of a better name. So added an uppercase.

modules/ui/core/intro.js
```diff
@@ -93,7 +98,7 @@ export function intro(context) {
         }
 
         var steps = ['navigation', 'point', 'area', 'line', 'startEditing'].map(function(step, i) {
-            var s = iD.ui.intro[step](context, reveal)
+            var s = Intro[step](context, reveal)
```
kepta: This is where `Intro` is being used.

modules/ui/core/intro.js
```diff
@@ -154,3 +159,36 @@ export function intro(context) {
     }
     return intro;
 }
+
+intro.pointBox = function(point, context) {
```
kepta: These guys are hanging like this. 
Should I export them ?

modules/ui/index.js
```diff
@@ -0,0 +1,111 @@
+import { ui } from './ui';
```
kepta: This is the part where it gets interesting.

modules/ui/index.js
```diff
@@ -0,0 +1,111 @@
...
+} from './core/index';
+export {ui};
+
+ui.preset = preset;
+Object.keys(presetObj).map(function(p) {
+    ui.preset[p] = presetObj[p];
+});
+
+// Need to do for test cases
+ui.intro = intro;
```
kepta: I had to add these keys to ui as `tests` still use it.
If we can create  separate build for tests and have a node.env variable to conditionally bundle this code. We can very much remove it for our production build.

modules/ui/core/save.js
```diff
@@ -1,4 +1,6 @@
+import { Save as SaveMode } from '../../modes/index';
```
kepta: Name check again.

modules/ui/core/preset.js
```diff
@@ -1,3 +1,9 @@
+import { Icon } from '../../svg/index';
+import { TagReference } from './tag_reference';
+import { Browse } from '../../modes/index';
+import { presetObj } from '../preset/index';
```
kepta: Again, I have this naming problem. So this time I named it `presetObj`. Let me know if something better fits.

modules/ui/intro/index.js
```diff
@@ -1,5 +1,13 @@
...
+import { area } from './area';
+import { line } from './line';
+import { navigation } from './navigation';
+import { point } from './point';
+import { startEditing } from './start_editing';
+
+export var Intro = {
+    area: area,
+    line: line,
+    navigation: navigation,
```
kepta: This I had to do as `core/intro.js` expects an object and then does `Intro[steps]`. 

Let me know what you think of this approach.

modules/ui/preset/index.js
```diff
@@ -1,13 +1,36 @@
...
+import { text, url, number, email, tel } from './input';
+import { localized } from './localized';
+import { lanes } from './lanes';
+import { maxspeed } from './maxspeed';
+import { radio } from './radio';
+import { restrictions } from './restrictions';
+import { textarea } from './textarea';
+import { wikipedia } from './wikipedia';
+
+export var presetObj = {
```
kepta: over here again I had to export them as an object, because of the way `core/preset.js` uses them.

bhousel:
I don't understand why there are naming problems.   Can you elaborate?

kepta:
Let's take core/intro.js. 
There world be two variables named intro.  One internal and one from ui/intro/index.js

kepta:
closed in favour of #3222

PR #3203, opened 2016-06-29
Eliminate vendored jxon library

bhousel:
Unfortunately this branch works great in the browser but won't run any tests because of a mysterious PhantomJS issue.

I've spent the better part of today on this and I'm pretty much ready to dump PhantomJS at this point.  It is not a real browser and has only basic debugging capability.


tmcw:
Looked into this issue - this PR is sprawling pretty far from its stated purpose, I think it's better to focus on #3222 instead

bhousel:
> this PR is sprawling pretty far from its stated purpose, I think it's better to focus on #3222 instead

ok, I cherry-picked the unrelated stuff and we can revisit this later..

PR #3207, opened 2016-06-29
Add license title

waldyrious:
It's not strictly required, but it's useful metadata, and part of the recommended license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+ISC License
+
 Copyright (c) 2015, iD Contributors
 
 Permission to use, copy, modify, and/or distribute this software for any


bhousel:
Thank you!

waldyrious:
Thank you for such a swift reaction! And kudos for picking this license, it's my favorite actually :)

huangyingjie:
I have a personal question about License:

This project use d3.js whose license is BSD-3. BSD-3 asks people to copy its copyright notice when use it. 
````js

Redistribution and use in source and binary forms, with or without modification,
are permitted provided that the following conditions are met

````
But why iD.js does not contains?  I have found that many project also don't contain their dependencies's copyright, ie. npmlog.  But why?

bhousel:
We do mention d3's BSD license here:
https://github.com/openstreetmap/iD/blob/master/README.md#license

However that is a throwback to when the iD source code included our own custom d3 bundle, which we don't do anymore.  We just depend/require d3 now, not redistribute its source code, so we probably don't need to mention the license anymore.

Issue #3183, opened 2016-06-18
add man_made=bridge to presets


matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dbridge

bhousel:
implemented in #3215

PR #3215, opened 2016-07-02
Adding man_made=bridge preset issue# 3183

SatyaSudheer:
add man_made=bridge to presets #3183

### DIFF
--- a/data/presets/presets/man_made/bridge.json
+++ b/data/presets/presets/man_made/bridge.json
@@ -1,0 +1,9 @@
+{
+  "geometry": [
+    "line"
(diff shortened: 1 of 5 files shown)


data/presets/presets/man_made/bridge.json
```diff
@@ -0,0 +1,9 @@
+{
+  "geometry": [
+    "line"
```
bhousel: This should be `area`.  The `man_made=bridge` tag is used for bridge outlines..

bhousel:
Thank you!

PR #3226, opened 2016-07-05
Use local jxon

tmcw:
* Q: why not the jxon module?
* A: We're using [algorithm 2](https://developer.mozilla.org/en-US/docs/Archive/JXON#Usage), which has a `@` shorthand.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -56,10 +56,6 @@
         "t": false,
         "bootstrap": false,
         "Diff3": false,
-        "JXON": false,
-        "osmAuth": false,
-        "toGeoJSON": false,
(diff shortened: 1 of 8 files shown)


bhousel:
> Q: why not the jxon module?
> A: We're using algorithm 2, which has a @ shorthand.

Can't we specify that in [config](https://github.com/tyrasd/jxon#configcnf)?

tmcw:
I'm not totally sure - I tried but didn't succeed initially. This is an in-between step that gets us closer to a one-shot compile, so going to work with this as-is and switch to the module later if appropriate.

PR #3228, opened 2016-07-06
[wip] use diff3 like a module

tmcw:
TODO: there's a test for the diff internals that we can't really run if they aren't global


bhousel:
Oh!  After I vendored this, somebody else independently took that same code and packaged it up.
https://www.npmjs.com/package/node-diff3

Let's comment out the test for now, then someday soonish I'll switch it over to using the module and contribute the tests upstream.

PR #3229, opened 2016-07-06
Sort modules [eslint rule]

kepta:
- use the sort modules rule to make the code more readable
- ignore the js/lib/id folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist/*.js
 dist/*.css
 dist/img/*.svg
 data/data.js
+js/lib/id/*.js
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1,4 +1,27 @@
 {
+    "access_simple": {
+        "key": "access",
+        "type": "combo",
+        "label": "Allowed Access",
+        "placeholder": "yes",
+        "options": [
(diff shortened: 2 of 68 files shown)


kepta:
This PR is pretty much trivial so I will go ahead and merge it.

.gitignore
```diff
@@ -6,3 +6,4 @@ dist/*.js
 dist/*.css
 dist/img/*.svg
 data/data.js
+js/lib/id/*.js
```
bhousel: To preserve the ability for people to check out the project and use it without a build step, I think we want to keep `js/lib/id/index.js` checked in for now. Even though it is a generated file, it's used by `index.html` in project root.

kepta: Okay. I totally got it wrong.
As per our conversation last week weren't we planning to ignore the bundles
generated by rollup ?

bhousel: No problem!  I want them ignored too, but I realized yesterday that `index.html` uses `js/lib/id/index.js`.

bhousel: see https://github.com/openstreetmap/iD/commit/90cd412bbe0bf0cfbe0631a8073a5c8d415093f3#commitcomment-15780969 for context

PR #3230, opened 2016-07-06
Phase II svg and services

kepta:
This also modifies some of the test cases to get services working.

Note:
2 test cases for `wikipedia` field fail,
`test/spec/ui/fields/wikipedia.js`
 cc: @bhousel @tmcw


bhousel:
Merged the `svg` part of this in b3492b1
`services` part of this will be continued on #3231

PR #3231, opened 2016-07-06
External modules for services

bhousel:
TBD issues with 
* wrapping closure module in `init()` function
* Wikipedia field tests failing because `sinon.stub` can't stub the wikidata lookup function


bhousel:
Thanks @tmcw!

PR #3236, opened 2016-07-08
Replace diacritics when doing fuzzy searches

### LINKED ISSUE #3159: equal ss to ß in the searchfield
Hi
and thanks for this absolutely awesome editor!

people in germani use the "double s"  > ß
people in switzerland and i also believe austrio don't.

so everytime someone from these countries tries to find "fussball feld" (soccer field)
they won't find it because it's written as "Fußball fled"

so if you could making appear while typing "fuss" like it does when i search for "soccer" would make lots of peoples life easier :)

thanks

bhousel:
tldr: it means that strings like "fussball" will fuzzy match strings like "fußball"

more details:
In `collection.js` search results are returned in the following order:

1. match name
1. match terms
1. match tag value
1. similar name
1. similar terms

Diacritical marks are only replaced when calculating the "similar" search results string distance, so this is a fallback strategy from strict string matches.

(closes #3159)

I also 
- [X] Expanded the tests to make sure results are returned in the right order
- [x] Added a few tests to test the diacritical mark replacement on fuzzy matches

cc @1ec5

### DIFF
--- a/modules/presets/collection.js
+++ b/modules/presets/collection.js
@@ -1,4 +1,5 @@
 import { editDistance } from '../util/index';
+
 export function Collection(collection) {
     var maxSearchResults = 50,
         maxSuggestionResults = 10;
@@ -22,6 +23,20 @@ export function Collection(collection) {
         search: function(value, geometry) {
             if (!value) return this;
 
+            function leading(a) {
+                var index = a.indexOf(value);
+                return index === 0 || a[index - 1] === ' ';
+            }
+
+            function suggestionName(name) {
+                var nameArray = name.split(' - ');
+                if (nameArray.length > 1) {
+                    name = nameArray.slice(0, nameArray.length - 1).join(' - ');
+                }
+                return name.toLowerCase();
+            }
+
+
             value = value.toLowerCase();
 
             var searchable = _.filter(collection, function(a) {
@@ -31,10 +46,6 @@ export function Collection(collection) {
                     return a.suggestion === true;
                 });
 
-            function leading(a) {
-                var index = a.indexOf(value);
-                return index === 0 || a[index - 1] === ' ';
-            }
 
             // matches value to preset.name
             var leading_name = _.filter(searchable, function(a) {
@@ -57,10 +68,10 @@ export function Collection(collection) {
 
 
             // finds close matches to value in preset.name
-            var levenstein_name = searchable.map(function(a) {
+            var similar_name = searchable.map(function(a) {
                     return {
                         preset: a,
-                        dist: editDistance(value, a.name().toLowerCase())
+                        dist: editDistance(value, a.name())
                     };
                 }).filter(function(a) {
                     return a.dist + Math.min(value.length - a.preset.name().length, 0) < 3;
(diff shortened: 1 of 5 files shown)


1ec5:
I tried this branch out with a few preset search terms that have been inconvenient in the past. The change produces a slight improvement over master. I haven’t found any real problems yet.

As a further test, I took this branch and removed the manually diacritic-folded Vietnamese terms using the regular expression `(, [-a-z ,]+)(?=")` on the `presets` object in vi.json. (I had to perform the find and replace manually because a handful of terms, like “TV”, “boong ke”, and “ga ra”, are supposed to lack diacritics.) It’s a wash, with significant improvements in cases where the search term mostly matches the diacritic-folded preferred term, but regressions in cases where the search term only partly matches the diacritic-folded preferred term or exactly matches a diacritic-folded secondary term. It may be possible to get better results by fiddling with the orders of terms in individual presets. I’ve pushed my changes to the [diacritics-vi-folding-removed](https://github.com/1ec5/iD/tree/diacritics-vi-folding-removed) branch in my fork, but I understand that I’d have to apply the changes in Transifex for them to stick.

preferred | master | this branch | my branch<br><sup>sans manually folded terms</sup>
----|----|----|----
⬆️ sân chơi<br>⬇️ công viên<br>⬆️ sân cỏ<br>⬆️ sân bóng đá | san choi, master | san choi, diacritics | san choi, simplified
⬆️ trường học<br>⬆️ nhà trường<br>⬇️ sân bay<br>⬇️ ga sân bay<br>⬆️ tường | truong, master | truong, diacritics | truong, simplified
⬇️ trung tâm bơi lội<br>⬇️ tiệm dụng cụ bơi lội | boi loi, master | boi loi, diacritics | boi loi, simplified
⬆️ nhà ở<br>⬇️ tòa nhà dân cư<br>⬇️ đất dân cư<br>⬆️ nhà thờ<br>⬆️ nha sĩ | nha o, master | nha o, diacritics | nha o, simplified

bhousel:
@1ec5 Thanks for digging into this!  

> It’s a wash, with significant improvements in cases where the search term mostly matches the diacritic-folded preferred term, but regressions in cases where the search term only partly matches the diacritic-folded preferred term or exactly matches a diacritic-folded secondary term. It may be possible to get better results by fiddling with the orders of terms in individual presets.

Yes, this change definitely won't solve all problems, but it should be possible to work around the more common issues by using the "preset terms".

[`collection.js search()`](https://github.com/openstreetmap/iD/blob/diacritics/modules/presets/collection.js#L23) contains the code which prioritizes how matches happen.

Exact matching on the leading part of name and leading part of terms are stuff you can control by adjusting the strings in Transifex.  Exact matching of the leading tag value might cause more problems in other languages.  Maybe we should disable that unless the locale is `en`?  In your examples above, I don't think it's affecting the results.

I think the diacritic replacement has a more pronounced effect on 'ß' -> 'ss', because it normalizes the Levenshtein distance between the search strings.  For example before this change, "grass" looks 3 chars different from "glaß", and after this change, they only differ by 1 character.

1ec5:
> it should be possible to work around the more common issues by using the "preset terms".

To clarify, the Vietnamese localization is already (ab)using the preset terms to include the main preset name, any synonyms, the main name diacritic-folded, and the synonyms diacritic-folded, in that order. I believe that's why this change has little effect.

Removing the diacritic-folded terms results in some results getting a lot better and some getting a lot worse, which in my opinion shows that the language-agnostic diacritic folding may be weighted too high for Vietnamese. (It would ideally count less toward the edit distance than base letter changes, whereas for other languages it should count more or the same.) To the extent that the workaround works, it's because we've specified a lot of synonyms in the Vietnamese presets.

So I'll probably keep the workaround in the Vietnamese localization (despite the bloat) and hold out for a more sophisticated solution in the future.

Issue #3179, opened 2016-06-17
ES6 Modules: phase 2


tmcw:
Phase 1 of iD's modularization is captured in https://github.com/openstreetmap/iD/issues/3118 and completed by our lovely community! Phase 2 is one task that @kepta is leading out and @bhousel is helping land:

So, phase 1 involved a lot of work and was split by the natural split points in iD: each module, like iD.ui and iD.svg. We have some degree of 'using modules for requiring their dependencies' but not 100% - the modules in `modules/*` still rely on the iD global all over the place. And we're creating a bundle per part of iD.

Phase 2 means

* Remove the iD global for the eslint configuration of `modules/*`, so that relying on global iD is forbidden
* Using `import` instead to require dependencies
* Instead of making a bunch of bundles, create one entry file that imports _all_ of iD, and goes through one rollup run to create the iD bundle.

---

I'll open a ticket for a moment for phase 3; the two tasks are/should be separable so that we can keep shipping little bits of work at a time.

kepta:
@tmcw how should we tackle the file `js/id/id.js`....

tmcw:
@kepta looks like that file serves 2 purposes: providing window.iD and then the [iife](http://benalman.com/news/2010/11/immediately-invoked-function-expression/) for feature detection. It should probably end up in two files serving those two purposes.

kepta:
@tmcw will we be needing `window.iD` anymore?

tmcw:
@kepta I _think_ so - do some research into the [openstreetmap/openstreetmap-website](https://github.com/openstreetmap/openstreetmap-website) repo to see if it uses the iD global to manipulate the map.

tmcw:
Merged the core PR! @kepta want to keep rolling on this work? Next step could be to [remove the iD global](https://github.com/openstreetmap/iD/blob/master/.eslintrc#L55) and get all those errors fixed.

kepta:
@tmcw, while working on `services` module, some of the test cases are failing. Because some test cases  (spec/directly manipulate `iD.services` eg `spec/ui/preset/wikipedia.js:32` ) and rollup freezes all the import objects.

bhousel:
With #3231 merged I think this can _almost_ be closed.

I think we missed this part, so I'll do it now..
>@kepta looks like that file serves 2 purposes: providing window.iD and then the iife for feature detection. It should probably end up in two files serving those two purposes.

kepta:
This is looking good.
[...]

PR #3243, opened 2016-07-12
Module for iD.Detect()

bhousel:
Re #3179 
This PR moves the `iD.detect()` iife into a module under `util`.
(I still export it as `iD.Detect()` so that embedding code like the rails site can test whether iD is supported.)

This PR also removes the `start.js` and `end.js` wrappers around `id.js`, because there is almost nothing left to wrap.

@tmcw @kepta for the review?


kepta:
It looks good to me.

My only concern is that we might give `util/detect.js` a special treatment. 
Till now, every other submodule is expected to be picked off from its `index.js` file.

bhousel:
> My only concern is that we might give util/detect.js a special treatment. 
> Till now, every other submodule is expected to be picked off from the submodule/index.js file.

`util/jxon` and `util/diff3` are like this too..

I don't really know why we still have the `index.js` files unless we are planning to break them back up into submodules again.  I hope so, because that way made more sense to me.

Put another way, I liked the namespaces we used to have, so I kind of prefer code like:
```js
import * as actions from '../actions/index';
...
context.perform(actions.Disconnect(params), ...)
```

instead of the current:
```js
import { Disconnect as DisconnectAction } from '../actions/index';
...
context.perform(DisconnectAction(params), ...);
```

Having every file make up it's own names for the things in the other files is just weird to me.

tmcw:
> Put another way, I liked the namespaces we used to have, so I kind of prefer code like:

I prefer it too on an aesthetic level and it is a bit easier to write, but it comes with disadvantages that I think outweigh the benefits:

* `import * as actions` means that rollup cannot do tree-shaking at all, so there's no chance for the file to optimize any part of actions _out_ - you're importing everything, so it'll include everything. Specifying specific parts of actions lets tree-shaking happen.
* `import { Disconnect } from './actions'` will throw a _compile time error_ if you misspell the import as disconnect or Disconect. Importing an object and then referencing a method on it, like `actions.Disconnect`, will delay this error to runtime, which may be also the time a user is using iD.

kepta:
Imho,
we should consider 'util' as a sub module and abstract away its internal
directory structure.
By that I mean,
To keep the index.js as a place import anything from the 'util' sub module.

Put another way,
Is there any disadvantage to using

import { Detect } from 'util/index'

Instead of

import { Detect } from 'util/detect'

tmcw:
I'm not terribly opinionated either way. Right now, for jxon and diff3, as @bhousel pointed out, we're using `util/` as a stopping place before those modules become external and we start to say `import merge from 'diff3'` instead of touching util at all. In those cases, I don't see much reason to import them _via_ `util/index`, since that import will soon change anyway, to the external module.

I can see `util/index` as having the advantage that you can import multiple utils at the same time, but that advantage doesn't really come into effect here since there's only one place we import Detect in the entire codebase.

kepta:
Ok I wasn't aware of the fact that `util` is just a transition thing.
Lets go ahead with this PR 😎.

tmcw:
:shipit:

PR #3254, opened 2016-07-19
Replace iD global with iD.Context() module

### LINKED ISSUE #3179: ES6 Modules: phase 2

bhousel:
This PR replaces the iD global window object with a Context module.

@kepta, @tmcw Does this seem right?  This might close #3179 

(though we still have to do something with `iD.data` / `data_dev.js`)

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -1,9 +1,21 @@
-(function () {
+import { Background } from '../renderer/background';
+import { Connection } from './connection';
+import { Detect } from '../util/detect';
+import { Features } from '../renderer/features';
+import { History } from './history';
+import { Map } from '../renderer/map';
+import { RawMercator } from '../geo/raw_mercator';
+import { presets as presetsInit } from '../presets/presets';
+import { init as uiInit } from '../ui/init';
 
-/*global iD*/
-window.iD = function () {
-    window.locale.en = iD.data.en;
-    window.locale.current('en');
+export function Context(root) {
+    if (!root.locale) {
+        root.locale = {
+            current: function(_) { this._current = _; }
+        };
+    }
+    root.locale.en = iD.data.en;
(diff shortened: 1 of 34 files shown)


modules/core/context.js
```diff
@@ -328,15 +340,16 @@ window.iD = function () {
 
 
     /* Init */
+    context.version = '2.0.0-alpha.1';
```
kepta: excited

kepta:
@bhousel looks good to me.
Lets merge it already !!

tmcw:
👍 Tested, LGTM. Doesn't close #3179 yet - we'll need to remove iD from the eslint globals in order to lock that down.

bhousel:
> we'll need to remove iD from the eslint globals in order to lock that down.

ok, I'll take care of `.eslintrc`

Also what should we do about `iD.data`?
I know we import/export the .json files too, so my plan was to put an `index.js` in the data folder that does that.

PR #3256, opened 2016-07-19
Inter-require iD instead of relying on the global

tmcw:
These are the two remaining cases of relying on iD _funcitonality_ via the global iD variable. The remainder are cases where we rely on iD.data or something similar, like iD.areaTags.


bhousel:
Nice!  Thanks for catching the ones we missed..
(btw, lets get rid of that lodash method chain.. not in this PR but soon)

PR #3257, opened 2016-07-19
Require lodash

tmcw:
cc @bhousel for the review

Refs #3180


bhousel:
Awesome!!   

The size impact is less than I expected, and I know we can probably do more targeted imports of lodash submodules to get the size down later.  Let's merge it..

Before:
```
-rw-r--r--   1 bryan  staff   2.1M Jul 19 12:01 iD.js
-rw-r--r--   1 bryan  staff   1.2M Jul 19 12:01 iD.min.js
```

After:
```
-rw-r--r--   1 bryan  staff   2.4M Jul 19 11:55 iD.js
-rw-r--r--   1 bryan  staff   1.3M Jul 19 11:55 iD.min.js
```

PR #3260, opened 2016-07-20
Add Taiwan (R.O.C.) address format.

david082321:

PR #3261, opened 2016-07-20
Add Taiwan (R.O.C.) address format

david082321:
Sorry. I forget a comma in #3260, so I pull request again.

### DIFF
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -36,6 +36,6 @@
     },
     {
         "countryCodes": ["tw"],
-        "format": [["postcode", "city", "district"], ["place", "village", "street"], ["housenumber", "floor"]]
+        "format": [["postcode", "city", "district"], ["place", "street"], ["housenumber"]]
     }
 ]
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -33,5 +33,9 @@
     {
         "countryCodes": ["ca"],
         "format": [["housenumber", "street"], ["city", "province", "postcode"]]
+    },
+    {
+        "countryCodes": ["tw"],
+        "format": [["postcode", "city", "district"], ["place", "village", "street"], ["housenumber", "floor"]]
     }
 ]
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -36,6 +36,6 @@
     },
     {
         "countryCodes": ["tw"],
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks!  It looks like some of the `addr:` keys that you want to use aren't yet added to `address.json`.
Specifically, `addr:village` and `addr:floor` keys seem to be missing:

screenshot 2016-07-21 00 21 52

It might be correct, and I can add them if so, but I'd like to see if these keys are really used anywhere before we "standardize" the tagging of ROC addresses this way.  I did look around in the map but didn't see any tagged this way - all current examples I could find use `addr:city`, `addr:street`, `addr:housenumber`.

Are `addr:village` and `addr:floor` real tags being used?

pnorman:
Both have low usage. `addr:floor` has worldwide usage and a wiki page, `addr:village` isn't documented and may overlap with `addr:city`. 25% of uses of addr:village are in one place.

I'd recommend against adding addr:village and asking a wider audience about addr:floor

david082321:
you're right. 
addr:village is a hide option, only show in full address.
And it would not have same address when hide it.

david082321:
OK! I've removed them.

Supaplextw:
Yeal addr:floor is not well documented but it is use in urban area in Taiwan like Taipei. Maybe use key:level instead of addr:floor for shop or amenity in multilevels building.

david082321:
BUT there are exception.

1.
addr:village
https://www.openstreetmap.org/node/2536763307
default

2.
addr:place
the road without street names
[http://overpass-turbo.eu/s/hpR](url)

3.
addr:floor
 to indicate different floors of the restaurant or company
[http://overpass-turbo.eu/s/hpU](url)

Supaplextw:
Actually "乾華派出所" is not in "阿里荖",  "阿里荖" is the smaller place under the village. "阿里荖" should be place in the addr:place form.

bhousel:
Looks like the latest version includes `addr:floor` and leaves off `addr:village` (which is not a standard documented tag).  This sounds fine to me.

bhousel:
This look ok to everyone?

screenshot 2016-07-21 10 33 30
screenshot 2016-07-21 10 33 36

david082321:
Yes, it's ok.

Supaplextw:
Yes, @bhousel it's ok

bhousel:
Just merged this, thanks everyone!

PR #3263, opened 2016-07-20
create amenity = ice_cream issue #3253

### LINKED ISSUE #3253: Create preset for amenity=ice_cream
Could you add support to the tag [`amenity=ice_cream`](http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dice_cream)? It is an approved tag and is being used in more than 7,000 elements, according to Taginfo. Maki has already [an icon](https://github.com/mapbox/maki/blob/master/icons/ice-cream-15.svg) for that too.

ankit-m:
(closes #3253)

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -1356,6 +1356,11 @@ en:
         # amenity=hunting_stand
         name: Hunting Stand
         terms: '<translate with synonyms or related terms for ''Hunting Stand'', separated by commas>'
+      amenity/ice_cream:
+        # amenity=ice_cream
+        name: Ice Cream Shop
+        # 'terms: gelato,sorbet,sherbet,frozen,yogurt'
+        terms: '<translate with synonyms or related terms for ''Ice Cream Shop'', separated by commas>'
       amenity/kindergarten:
         # amenity=kindergarten
         name: Preschool/Kindergarten Grounds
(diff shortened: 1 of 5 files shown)


data/presets/presets/amenity/icecream.json
```diff
@@ -0,0 +1,21 @@
...
+        "opening_hours",
+        "takeaway",
+        "delivery"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "ice-cream"
```
bhousel: You don't have to put "ice cream" in the terms, because it is part of the preset name and also the value.  The `terms` are more for synonyms or related words like "gelato", "sorbet", "sherbet", "frozen", "yogurt", "self serve"..

bhousel:
Looks pretty good, thanks!  Can you rename the file to `amenity/ice_cream.json`?  We try to have the filename match the preset value when that is possible.

data/presets/presets/amenity/icecream.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "tags": {
+        "amenity": "ice_cream"
+    },
+    "name": "Ice Cream"
```
bhousel: Let's call it "Ice Cream Shop".    Without shop, it just sounds like a pile of ice cream on the ground.

ankit-m:
@bhousel I have made the changes.

bhousel:
Looks great, thanks!  🍦

PR #3264, opened 2016-07-21
fix critical error on save changes

edpop:
A small fix


bhousel:
Good catch, thank you!  Not sure how that didn't make it into 7ca6d06..

PR #3270, opened 2016-07-24
Add Taiwan emap NLSC opendata version as an image source

Supaplextw:
There's a emap from Taiwan surveying authority, NLSC emap, which is compatible with CC-BY 4.0 (http://data.gov.tw/license). It's already add in JOSM imagery, but still need to add in iD.


Supaplextw:
Just ignore the part of adding zh.js. I'm not quite familiar with Git

bhousel:
Hi, thanks for the report. iD's background imagery sources come from the [editor-layer-index](https://github.com/osmlab/editor-layer-index) repository. Can you open the request there?  By adding it there, it can also be used in other editors.   Thanks!

PR #3274, opened 2016-07-25
add second_hand tag in preset and second_hand field for shop=car

skorasaurus:
Hi, 

I added the second_hand= tag to the preset. The tag is used to denote for shops whether the merchandise is used. It's used over [4,000 times ](http://taginfo.openstreetmap.org/keys/second_hand#values) so far in the planet. 

One common use of the second_hand tag is to designate a used car lot. I've also added the second_hand tag in this pull request.

### DIFF
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1096,7 +1096,7 @@
     "second_hand": {
         "key": "second_hand",
         "type": "combo",
-        "label": "Second Hand / Used / Previously Owned ",
+        "label": "Sells Used",
         "placeholder": "Yes, No, Only",
         "strings": {
             "options": {
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1093,6 +1093,19 @@
         "type": "check",
         "label": "Seasonal"
     },
+    "second_hand": {
+        "key": "second_hand",
+        "type": "combo",
+        "label": "Second Hand / Used / Previously Owned ",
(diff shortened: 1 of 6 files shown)


data/presets/fields.json
```diff
@@ -1093,6 +1093,19 @@
         "type": "check",
         "label": "Seasonal"
     },
+    "second_hand": {
+        "key": "second_hand",
+        "type": "combo",
+        "label": "Second Hand / Used / Previously Owned ",
```
bhousel: Is it ok to make this label shorter, like "Sells Used"

data/presets/fields/second_hand.json
```diff
@@ -0,0 +1,14 @@
+
+{
+    "key": "second_hand",
+    "type": "combo",
+    "label": "Second Hand / Used / Previously Owned ",
```
bhousel: > Is it ok to make this label shorter, like "Sells Used"

This is the actual line to make the change..

data/presets/fields.json
skorasaurus: Sure, I have no problem with that.

data/presets/fields/second_hand.json
skorasaurus: thanks for tip. Do I run ```make``` again then?

bhousel:
Thanks!  This looks great  👍

Issue #2949, opened 2016-01-31
With multiple features selected, indicate which one is mouse-hovered in the sidebar multiselect list 


slhh:
Edit: The remaining task is highlighting the list item when hovering the selected feature on the map, which is discussed and agreed on starting here: https://github.com/openstreetmap/iD/issues/2949#issuecomment-431540750
Original issue:

Identification of items in the multiselect list should be improved:

When the mouse is over an item of the list, selection halos of all other selected elements should be disabled temporarily.

This will likely require similar software infrastructure as #2946 , and is a prerequisite for some further potential enhancements of iD.

bhousel:
I'm not removing the halo from the other selected items.  They are still selected and the user shouldn't forget about that just because they happen to put their mouse over that selection list. 

But I agree that:
* when mousing over a list item, that feature should get a `hover` style (maybe it does already)
* the `hover` style should be more distinguishable from the select style with consideration for colorblind users (I think it may only get a slight opacity bump)
* breathe behavior needs to be aware of `hover` style if it is mucking with opacity

slhh:
> I'm not removing the halo from the other selected items.

I believe the user wants to go through the list quite fast searching for an item. Anything delaying the visisual detection of the hovered feature seems to be bad.

> They are still selected and the user shouldn't forget about that just because they happen to put their mouse over that selection list.
I would disagree in this specific case. While the mouse is over the buttton the users can't do anything harmful with the mouse. The only hamful thing they can do is using some shortcut keys, but they can't really work on the map for any significant time period without moving the mouse away from the hovered item; therefore they don't really have time to forget.

If the users were able to fill in the inspector with the keybord during hovering the item, I would agree; but the inspector is unavailable while the multiselect list is open.

In addition multiple items being selected is still visible, because the multiselect list is shown.

If you still want to protect something, you can consider disabling shortcut keys which are changing map data according to the hidden selection; but I think the risk is very low anyway.

slhh:
In adition to the hover function, which has been discussed above, a reserse hover function would be useful. If the mouse is over one or more selected items on the map, the respective item of the multiselect list should be highlighted

slhh:
@quincylvania, @bhousel, 
How do you think about my last comment here?

Everything else was done in #5402, thanks @quincylvania.

quincylvania:
[...]

PR #3275, opened 2016-07-25
Remove an item from the multiselect list

### LINKED ISSUE #2950: Multiselect list: Icon to remove an item from list and selection
Each item of the multiselect list should have an icon which is removing the item from the list, and the associated feature from the selection.

This could be used to refine shift+drag selections.

This would also solve  Handle coincident ways #1239 if all coincident ways were selected on mouse click.

A prerequisite for this issue is #2949

ebrelsford:
Fixes #2950

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -735,6 +735,14 @@ button.save.has-count .count::before {
     opacity: .5;
 }
 
+.feature-list-item .label .close {
+    float: right;
+}
+
+.feature-list-item .label .close .icon {
+    opacity: 1;
+}
+
 .feature-list-item:hover .label {
     background-color: #ececec;
 }
--- a/modules/ui/selection_list.js
+++ b/modules/ui/selection_list.js
@@ -10,6 +10,14 @@ export function SelectionList(context, selectedIDs) {
         context.enter(Select(context, [entity.id]).suppressMenu(true));
     }
 
+    function deselectEntity(entity) {
+        d3.event.stopPropagation();
+        var index = selectedIDs.indexOf(entity.id);
+        if (index > -1) {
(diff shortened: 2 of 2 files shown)


bhousel:
Thanks @ebrelsford for contributing this at SOTM-US hack day!   

I made a slight change to the code so that the geometry and close icons would be classed differently.   With the old code `items.selectAll('use')` it was possible to trigger a list redraw, causing things like this:

screenshot 2016-07-26 15 35 56

So I added a class and replaced that selection with `items.selectAll('.entity-geom-icon use')`, so if the redraw does happen, it won't break the close buttons:

screenshot 2016-07-26 16 03 42

ebrelsford:
Makes sense, thanks for the merge @bhousel !

slhh:
Thanks @ebrelsford for contributing this.

Unfortunately, it doesnt work with Firefox. The close icons are positioned too low and don't seem to be active buttons (neither in the shown position, nor in the expected position) . Clicking there just selects the item from the list.
multiselect close icon issue

I have tested with Firefox 47.0 and 47.0.1 on Windows 10.
Clearing the cache has not solved the issue.

With Edge it works fine here.
The usability is somewhat limited, because it might not be possible to find the correct item to be removed in case of multiple items of the same type, for example residential buildings in @bhousel 's screenshot.
This needs #2949 to be solved.

ebrelsford:
@bhousel Fixing the styling is easy enough, but it seems that Firefox is not going to allow children of buttons to listen for clicks (didn't realize!). Okay for me to make these elements `div`s rather than `button`s?

bhousel:
cool, I didn't realize that either - yes make them divs!

PR #3276, opened 2016-07-25
Do not attach dragNode behavior in browse mode

gmaclennan:
Currently when in 'browse' mode (i.e. when nothing is selected) it is very easy for a user to accidentally drag a point when they are trying to move the map. This is not currently the case for lines or areas: they must be selected before it is possible to move anything.

This PR changes the browse mode so that it does not attach the DragNode behavior to points. The new behavior is that you need to select a point first before dragging to move it.

This is to reduce the possibility of users accidentally moving points without realizing when they are just trying to move the map. This is a particular issue in areas of the map where there is a high density of points, which makes it almost impossible to drag the map without accidentally moving a point. This fixes that.

### DIFF
--- a/modules/modes/browse.js
+++ b/modules/modes/browse.js
@@ -1,6 +1,5 @@
 import { t } from '../util/locale';
 import { Hover, Lasso, Paste, Select } from '../behavior/index';
-import { DragNode } from './index';
 
 export function Browse(context) {
     var mode = {
@@ -15,8 +14,7 @@ export function Browse(context) {
         Hover(context)
             .on('hover', context.ui().sidebar.hover),
         Select(context),
-        Lasso(context),
-        DragNode(context).behavior];
+        Lasso(context)];
 
     mode.enter = function() {
         behaviors.forEach(function(behavior) {


bhousel:
This is a great idea.. I've been frustrated before too when I try to drag the map and end up dragging a node by accident.

I checked it out and one thing I noticed is that with this change, yes you can't drag points in browse mode, but you can drag _any_ point in select mode, even a point that's not actually selected.  It's a little strange.  

Maybe [here in DragNode](https://github.com/openstreetmap/iD/blob/018e7f207e1797c560d3561f0aedbe87fc06511d/modules/modes/drag_node.js#L60) can you try to cancel the drag behavior if the entity is not in the `selectedIDs` array?   You might also need to expand the whitelist to include nodes that are children of way seletedIDs or members of relation selectedIDs, I'm not sure.

jfirebaugh:
> You might also need to expand the whitelist to include nodes that are children of way seletedIDs or members of relation selectedIDs, I'm not sure.

Pretty sure this will be necessary. When a way is selected, you should be able to move its vertices without first selecting them -- otherwise realigning roads will be extremely tedious.

bhousel:
This was finished in #3739

PR #3280, opened 2016-07-27
Allow uppercase in network taginfo

1ec5:
When suggesting values for `network` or `cycle_network` from taginfo, include values that contain uppercase letters. This effectively enables taginfo suggestions for the`cycle_network` tag where previously they were completely disabled, and suggestions for the `network` tag are more accurate as well.

cycle_network

Presumably the uppercase filter was originally intended to block suggestions for freeform tags like `name`. But `network` and `cycle_network` aren’t entirely open-ended, even though their values contain uppercase letters.

/cc @bhousel

### DIFF
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -49,9 +49,10 @@ function filterMultikeys() {
     };
 }
 
-function filterValues() {
+function filterValues(allowUpperCase) {
     return function(d) {
-        if (d.value.match(/[A-Z*;,]/) !== null) return false;  // exclude some punctuation, uppercase letters
+        if (d.value.match(/[;,]/) !== null) return false;  // exclude some punctuation
+        if (!allowUpperCase && d.value.match(/[A-Z*]/) !== null) return false;  // exclude uppercase letters
         return parseFloat(d.fraction) > 0.0 || d.in_wiki;
     };
 }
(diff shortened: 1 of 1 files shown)


bhousel:
> Presumably the uppercase filter was originally intended to block suggestions for freeform tags like name. But network and cycle_network aren’t entirely open-ended, even though their values contain uppercase letters.

Looks like I added the uppercase filter in 03bb916.  

I couldn't remember why, but thankfully I added [a test for it](https://github.com/openstreetmap/iD/commit/03bb9162870e562fc6481b55b1c09bb13eb36086#diff-a2d4d83d15b9d06ca799ef758794b0a8R161) that gives some explanation.  This is for values retrieved from taginfo that we might show as suggestions in a dropdown field.  If the user has entered key `amenity` and types `pa` in the value box, and taginfo returns both `parking` and `PArking`, we just want to suggest the lowercase version.  (They can still type whatever they want in the box).

You're right that for some fields this should be configurable - a handful of tags like `network` use uppercase or mixed case values.

1ec5:
Ah, that makes sense. Mixed-case suggestions would also apply to `opening_hours`, and for that matter any conditional tag. But I actually couldn't stand it when I used to get any suggestions for those tags.

I'll add a test for the network tags.

Issue #3202, opened 2016-06-28
Add emergency=defibrillator to iD presets


lakedistrictOSM:
Would be ideal if this came up when either "defibrillator" or "AED" was searched for. An important node to map - has potential to be life saving.
OSM wiki page [here](https://wiki.openstreetmap.org/wiki/Tag:emergency%3Ddefibrillator).

PR #3283, opened 2016-07-27
Add AED as term for Defibrillator

M1dgard:
AED, short for Automated External Defibrillator, is a term known by many people and often marked on defibrillators. Was also originally requested in #3202:

> Would be ideal if this came up when either "defibrillator" or "AED" was searched for.


bhousel:
Yes good idea, thanks! 👍

PR #3284, opened 2016-07-27
Add indoor=yes/no to preset for Defibrillator

M1dgard:
Doing this separately from #3283 because this adds a new field and I'm not sure you're OK with it.

I add `indoor=yes` to all AEDs I map that are located inside a building.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -447,6 +447,9 @@ en:
           down: Down
           # incline=up
           up: Up
+      indoor:
+        # indoor=*
+        label: Indoor
       information:
         # information=*
         label: Type
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -618,6 +618,11 @@
         "type": "combo",
         "label": "Incline"
     },
+    "indoor": {
+        "key": "indoor",
+        "type": "check",
+        "label": "Indoor"
+    },
     "information": {
         "key": "information",
         "type": "typeCombo",
--- a/data/presets/fields/indoor.json
+++ b/data/presets/fields/indoor.json
@@ -1,0 +1,5 @@
+{
+    "key": "indoor",
+    "type": "check",
+    "label": "Indoor"
+}
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -4047,6 +4047,7 @@
         "fields": [
             "access",
             "opening_hours",
+            "indoor",
             "phone"
         ],
         "geometry": [
--- a/data/presets/presets/emergency/defibrillator.json
+++ b/data/presets/presets/emergency/defibrillator.json
@@ -2,6 +2,7 @@
     "fields": [
         "access",
         "opening_hours",
+        "indoor",
         "phone"
     ],
     "geometry": [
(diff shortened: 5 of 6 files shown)


bhousel:
That doesn't seem to be documented, but I do see it as a combo in taginfo.
http://taginfo.openstreetmap.org/tags/?key=emergency&value=defibrillator#combinations

Why is it helpful to tag them as indoor?  Doesn't positioning them inside a building make that clear?

M1dgard:
> Doesn't positioning them inside a building make that clear?

Most of the time, yes. But there are complex situations where it isn't as clear. Of course, those are the most intriguing to map, and my government finds those the most intriguing to build as well lately. :smile:

Examples:

* [Marie Popelinplantsoen, Bruges](https://www.openstreetmap.org/#map=19/51.19669/3.21966)
The ways through [the park](https://i296.photobucket.com/albums/mm182/Keveve/IMG_0406.jpg) on the map are outdoor. There is building underneath them. layer tags are also used IIRC.

* [Railway station of Bruges](https://www.openstreetmap.org/#map=19/51.19707/3.21750)
The tracks and platforms are outdoor, the shops are in the corridor underneath (indoor).

* [Near the railway station of Ghent](https://www.openstreetmap.org/#map=19/51.03681/3.70891)
The most complicated fully public layered environment I know, there's a [90m/300ft tall building](https://www.mapillary.com/app/?lat=51.03637738&lng=3.71097505&z=17&pKey=gvMUc79MVGd-Q7GOg5dbPw&focus=photo) with a [tram going underneath](https://www.mapillary.com/app/?lat=51.03690675412234&lng=3.708463265680197&z=17&pKey=a0i7VEwAlky_MkKMfXyJiQ&focus=photo), with a [bicycle parking underneath](https://www.mapillary.com/app/?lat=51.03682654&lng=3.71020806&z=17&pKey=tqhrjWZumJrsz_Poagw2hA&focus=map), with a car parking underneath.

Granted, there are no AEDs on these locations, but there might be some day. I hear that in Japanese cities there are even crazier situations.

So no, you can't know whether something is indoor or not. Maybe you think layer tags suffice.

Some AEDs may be on the rooftop. Dang, maybe we'd better add [`location=*`](https://wiki.openstreetmap.org/wiki/Key:location) instead?

bhousel:
> So no, you can't know whether something is indoor or not. Maybe you think layer tags suffice.
> Some AEDs may be on the rooftop. Dang, maybe we'd better add location=* instead?

Why not use [`level`](http://wiki.openstreetmap.org/wiki/Key:level) for features that are indoors (or located at a place where a building is).

`layer` tags are really more to keep bridges / tunnels / waterways sorted.

M1dgard:
> Why not use level for features that are indoors (or located at a place where a building is).

I do. I'm not using layer indoor.

> layer tags are really more to keep bridges / tunnels / waterways sorted.

I also use layer tags for outdoor overlapping things. I gave the Marie Popelinplantsoen and the ways through it `layer=1`, because they are higher than the building underneath.

I'd give an AED in the park `layer=1` as well as `indoor=no`. So maybe you think layer would be enough. I think the indoor tag helps to avoid any disambiguity – when locating an AED every second counts, so you have to know if it's indoor or not.

bhousel:
I guess since it has around 1500 uses on taginfo, it's ok.. It just seemed kind of weird to me.

PR #3287, opened 2016-07-27
Fix deselect entity on Firefox

ebrelsford:
Includes splitting select/deselect elements into separate buttons, fixes issue brought up in #3275


bhousel:
Thanks @ebrelsford!  I added a little padding to the buttons (they were scrunched pretty close to the right side of the div), and tested in Chrome, Firefox, and Safari, and all looks good..

Issue #3285, opened 2016-07-27
Multiple checkboxes


M1dgard:
I was trying to write a preset for [blood donor centers](https://wiki.openstreetmap.org/wiki/Tag:healthcare%3Dblood_donation). The tagging scheme is easy to apply manually, but I'm struggling to make the part about which blood components are collected look clear in iD. I was thinking of this:

![id_field_blood_components](https://cloud.githubusercontent.com/assets/4283376/17179216/ab63fba0-5418-11e6-8fa7-33305f4c943f.png)

But I didn't find a field type that looks like this. (It's like a hybrid of `access` and `check`.)

Would it be OK if I tried to make it? What name should it have? Something like "multiCheck"?

bhousel:
For that I'd use the multiselect field that we added in #3080:
example: https://github.com/openstreetmap/iD/pull/3080#issuecomment-215933840

M1dgard:
I did consider this, but it *feels* a bit odd. When I'm mapping a donor center, do I have to search for all components I can donate there?

Also, you have to be able to express that you *cannot* donate a specific component there (e.g. `blood:platelets=no`). Unlike `recycling:*=yes` it defaults to unknown, even when some components are given.

bhousel:
> I did consider this, but it feels a bit odd. When I'm mapping a donor center, do I have to search for all components I can donate there?

We wrote the multiselect field in a way that (like combo fields) it can either be a fixed list of values or use free-tagging with suggestion from taginfo.

> Also, you have to be able to express that you cannot donate a specific component there  (e.g. blood:platelets=no).

Yes that's true.  I'm ok about this.  I think data consumers should just assume 'no' if a thing is not tagged with a 'yes'.  Encouraging users to tag the 'no's is unnecessary.

M1dgard:
> Encouraging users to tag the 'no's is unnecessary.

I guess you're proposing to make no distinction between `no` and unknown. Or would you use `blood:platelets=unknown`? (not intended cynically)

I have mapped blood donor centers without knowing whether you can have stem cell samples taken there, and I quite like the fact that absent=unknown.

bhousel:
In iD's multiselect fields we're only supporting `yes`.  It's ok if people want to use the raw tag editor to add `no` or other values.  But our main goal is getting people to tag the `yes` values (and not being confusing to new users).

> I guess you're proposing to make no distinction between no and unknown.

[...]

PR #3290, opened 2016-07-28
Preset for blood donor centers

M1dgard:
Preset for [healthcare=blood_donation](https://wiki.openstreetmap.org/wiki/Tag:healthcare%3Dblood_donation) :heart::droplet:, as promised in #3285.

I didn't even think about adding the part about which compensation you receive, because we already had problems enough with the blood components. :stuck_out_tongue:

Some mini golf changes ended up here as well, due to executing `make`.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -178,6 +178,18 @@ en:
       bin:
         # bin=*
         label: Waste Bin
+      blood_components:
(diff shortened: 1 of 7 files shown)


bhousel:
Sure this looks good, thank you!

PR #3298, opened 2016-07-31
Fix taginfo for relation presets

### LINKED ISSUE #3297: No documentation for relation presets
#1862 has regressed: taginfo documentation is unavailable for any relation preset.

boundary

1ec5:
taginfo’s relation/wiki_pages method no longer places results in a `data` property.

boundary

Also, only use relation/wiki_pages for top-level relation presets. More specific presets behave like non-relation presets. For example, the bicycle route relation has the tags `type=route` `route=bicycle`, so `route=bicycle` is what we’re interested in.

bicycle

Fixes #3297.

/cc @bhousel

### DIFF
--- a/modules/presets/preset.js
+++ b/modules/presets/preset.js
@@ -61,8 +61,15 @@ export function Preset(id, preset, fields) {
             value = preset.tags[key];
 
         if (geometry === 'relation' && key === 'type') {
-            return { rtype: value };
-        } else if (value === '*') {
+            if (value in preset.tags) {
+                key = value;
+                value = preset.tags[key];
+            } else {
+                return { rtype: value };
+            }
+        }
+
+        if (value === '*') {
             return { key: key };
         } else {
             return { key: key, value: value };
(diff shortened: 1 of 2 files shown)


bhousel:
Is this something that should maybe be reported as a bug in taginfo rather than us working around it in iD?

1ec5:
taginfo has formatted the results this way for years. I’d have to trawl through history to be sure, but I suspect iD used to be more lenient about the lack of a `data` property. Meanwhile, the second part of this change never used to work correctly, since there were only top-level relation presets when I implemented #1863.

1ec5:
Oh, I see what you mean, though: the output format doesn’t match [the documentation](https://taginfo.openstreetmap.org/taginfo/apidoc#api_4_relation_wiki_pages). I’ve filed joto/taginfo#179 to track that issue, but I think the second part of this PR can be merged.

1ec5:
I removed the workaround for joto/taginfo#179. Until that upstream bug is addressed, the first example above doesn’t work, but the second one does.

joto:
I just fixed taginfo so that the API now behaves as the documentation says.

bhousel:
Thanks @1ec5 for documentation improvement and @joto for the quick fix on the taginfo side!

PR #3299, opened 2016-07-31
taginfo tooltips for combo values

1ec5:
Each combo value obtained from taginfo now has a tooltip describing the value, if available.

culvert

(taginfo suggestions in the “All tags” section already have descriptive tooltips.)


bhousel:
Nice, this is great, thanks @1ec5!

PR #3300, opened 2016-07-31
taginfo for relation roles

1ec5:
Enabled taginfo suggestions for roles on relation members and memberships.

roles

### DIFF
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -9,11 +9,25 @@ var taginfo = {},
         area: 'count_ways',
         line: 'count_ways'
     },
+    tag_sort_members = {
+        point: 'count_node_members',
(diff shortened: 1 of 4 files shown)


modules/services/taginfo.js
```diff
@@ -146,6 +179,23 @@ export function init() {
             });
     };
 
+    taginfo.roles = function(parameters, callback) {
```
bhousel: Can you add a test for `roles` to [`test/spec/services/taginfo.js`](https://github.com/openstreetmap/iD/blob/master/test/spec/services/taginfo.js)?

1ec5: Done.

bhousel:
Awesome, thanks!!

Issue #1524, opened 2013-05-25
Render address points and addresses


lxbarth:
Address points (screenshots) and addresses aren't rendered currently. This would be great to include to encourage more address mapping.

[![](http://cl.ly/image/060i0v42182U/Screen%20Shot%202013-05-25%20at%2010.25.58%20AM.png)](http://openstreetmap.us/iD/release/#background=Bing&id=w67231791&map=19.10/-77.01914/38.91357)

tordans:
Could the the feature that shows `ref=*` next to a tree(…) node be re-used for this?
So, instead of only showing `ref` next to a node, we also display `addr:housenumber` next to it.

Screenshot: Above a `addr:housenumber-node`, below a `natural=tree + ref=123` node.
<img width="161" alt="Bildschirmfoto 2019-04-13 um 07 37 15" src="https://user-images.githubusercontent.com/111561/56075262-e85c2380-5dbf-11e9-8a2d-3cef6bd82d4a.png">

quincylvania:
I'm okay with just showing the house number and street for features without any other name. I know #3840 had greater ambitions but I think this would be a fine intermediate step.

<img width="807" alt="Screen Shot 2019-04-15 at 1 57 32 PM" src="https://user-images.githubusercontent.com/2046746/56168649-a33b1b80-5f90-11e9-8324-5ba23496a96d.png">

Nakaner:
Rendering the both house numbers and street names adds a lot of labels to the map hiding a lot of other information (background layer). It seems more reasonable to render house numbers only. Otherwise you will have to deal with `addr:place` and other special cases addresses can have.

tordans:
@quincylvania I see a merge commit in your PR. Will it be live at http://preview.ideditor.com/master/#background=Berlin-2018&disable_features=boundaries&map=19.37/52.47279/13.44774 soon to try out (isn't the master synced automatically with "preview"?)

quincylvania:
@tordans Nope, I closed the PR instead of merging. We're not going this direction at this time after all.

bagage:
Thanks @quincylvania for trying a first partial solution! What were the reasons to finally not merge it?

Also #3840 is closed too, so yet there is no plan for solving that issue yet, is that correct?

jidanni:
Odd, are there any other names that carto renders,
https://www.openstreetmap.org/?#map=19/24.23887/120.83488
but iD doesn't?!

Filip10:
[...]

PR #3301, opened 2016-07-31
Include network in relation display name

1ec5:
If a relation is tagged with a `network`, the `network` is needed alongside the `ref` to identify the relation.

For example, this Ohio State Route relation is tagged `type=route` `route=road` `network=US:OH` `ref=28`, while its members are tagged `ref=SR 28`:

membership

changes

Displaying the network will make mappers less inclined to redundantly tag the network-ref combination in `name`.

### DIFF
--- a/modules/util/util.js
+++ b/modules/util/util.js
@@ -29,12 +29,13 @@ export function entityOrMemberSelector(ids, graph) {
 }
 
 export function displayName(entity) {
-    var localizedNameKey = 'name:' + Detect().locale.toLowerCase().split('-')[0];
-    var name = entity.tags[localizedNameKey] || entity.tags.name;
+    var localizedNameKey = 'name:' + Detect().locale.toLowerCase().split('-')[0]
+        name = entity.tags[localizedNameKey] || entity.tags.name || '',
+        network = entity.tags.cycle_network || entity.tags.network;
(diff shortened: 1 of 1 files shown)


bhousel:
I'll check this tomorrow to make sure it doesn't display strange things for cycle/hiking/horse networks (e.g. `ncn ECG` for the "East Coast Greenway"), but I think those routes usually have something in the `name` field.

1ec5:
That’s a good point: relations tagged with `cycle_network` should use that tag instead of `network`. I’ll need to address that issue before this can be merged.

edpop:
Perhaps, many people want to see also house numbers.

1ec5:
This PR is about how relations are summarized in the "All relations" section of the inspector and in the summary of changes when saving. It may make sense for the summary to fall back on the house number or street address for unnamed buildings, but I think that should be addressed in a separate PR.

bhousel:
@edpop rendering of address numbers is ticketed on #1524  👍

bhousel:
Thanks @1ec5!  I tested this with the East Coast Greenway and it looks good.

It would only display "ncn ECG" if a route has no values in preferred fields like `name` or `cycle_network`.  I think this is fine.

PR #3281, opened 2016-07-27
Combo Network Type field for bike/hike routes

1ec5:
Replaced the generic Network field (`network`) with a specialized Network Type field (`network`) in the bicycle, foot, and hiking route relation presets. Added a Network (`cycle_network`) field to the cycling route relation preset.

bicycle
_Cycling route relation fields_

bicycle-open
_Cycling route network types_

The combo values for the cycling route relation preset conform to [wiki guidelines](http://wiki.openstreetmap.org/wiki/Tag:route%3Dbicycle) and common usage:

* Local (`network=lcn`)
* Regional (`network=rcn`)
* National (`network=ncn`)
* International (`network=icn`)

The combo values for the walking and hiking route relation presets also conform to [wiki guidelines](http://wiki.openstreetmap.org/wiki/Tag:route%3Dhiking):

* Local (`network=lwn`)
* Regional (`network=rwn`)
* National (`network=nwn`)
* International (`network=iwn`)

Depends on #3280.

/cc @bhousel

PR #3302, opened 2016-07-31
Combo field with taginfo for road networks

1ec5:
Added a variation of the combo field for road networks. If the field’s value is blank or shares a prefix with the current country code, search taginfo for values beginning with the country code and a colon.

combo

`network` values for road routes generally begin with the country code (either uppercase or lowercase) followed by a colon. The one big exception is Ontario: 2,023 are relations tagged [`network=ca_on_county`](http://taginfo.openstreetmap.org/tags/network=ca_on_county), so overall there are 1,500 elements tagged `network=CA:*` versus 2,429 tagged `network=ca_*`. On the other hand, 2,478 bus stops are tagged [`network=co:dc:sitp`](http://overpass-turbo.eu/s/hBi), which does begin with the country code and a colon. Also, a few popular values in Europe are unqualified letters. Despite these exceptions, I think it’s reasonable to suggest values beginning with the country code and a colon; if local standards differ, the user is free to enter any other value.

/cc @bhousel

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -688,6 +688,9 @@ en:
           rhn: Regional
         # network_horse field placeholder
         placeholder: 'Local, Regional, National, International'
+      network_road:
+        # network=*
+        label: Network
       note:
         # note=*
         label: Note
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -906,6 +906,11 @@
             }
         }
     },
+    "network_road": {
+        "key": "network",
+        "type": "networkCombo",
+        "label": "Network"
+    },
     "network": {
         "key": "network",
         "type": "text",
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -367,7 +367,7 @@
     },
     "cycle_network": {
         "key": "cycle_network",
-        "type": "combo",
+        "type": "networkCombo",
         "label": "Network"
     },
     "cycleway": {
--- a/data/presets/fields/cycle_network.json
+++ b/data/presets/fields/cycle_network.json
@@ -1,5 +1,5 @@
 {
     "key": "cycle_network",
-    "type": "combo",
+    "type": "networkCombo",
     "label": "Network"
 }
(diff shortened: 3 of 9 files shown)


1ec5:
This same field could be used for `cycle_network`, where there is even stronger consensus on prefixing networks with the country code and a colon.

1ec5:
Added the same functionality to the `cycle_network` field.

I also noticed taginfo is returning matches anywhere in the tag values, not just prefix matches. Here’s what the `cycle_network` field looked like in Canada:

canada-california

So I fixed that by filtering after the fact:

canada-after

bhousel:
I played around with this and I think it's great, thanks @1ec5 👍 

Aside: surprised that `US:TX:FM` ranks so highly.

1ec5:
Yes, by sheer number (and mileage), the Texas Farm to Market and Kentucky Route networks are huge. Unfortunately, there isn’t a good way to rank the U.S. networks by importance just by doing string processing.

pnorman:
> `network` values for road routes generally begin with the country code (either uppercase or lowercase) followed by a colon.

This really varies. When I looked previously colons were mainly used in the US, but usage has shifted somewhat.

Of course, the most common network tag values are lwn, rwn, rcn, and lcn.

1ec5:
> When I looked previously colons were mainly used in the US, but usage has shifted somewhat.

Colons are used almost everywhere these days. In the past, the U.S. was unique for using colons to delimit a sort of “reverse DNS” rather than the older [`XY:national`](http://wiki.openstreetmap.org/wiki/Key:network#Basic_format) scheme. But other countries have adopted the U.S. scheme too (Brazil, Canada, India, Italy, and Slovenia the last time I checked).

> Of course, the most common network tag values are lwn, rwn, rcn, and lcn.

iD uses a different field to represent the `network` tag for non-road route relations: #3281. A typical user should never see the letters “lwn”. With this PR, the separate `cycle_network` field pulls the most common values for that tag.

PR #3316, opened 2016-08-02
Add 75, 80 mph speed limits

1ec5:
75 and 80 mph speed limits are [common in several U.S. states](https://en.wikipedia.org/wiki/Speed_limits_in_the_United_States).


pnorman:
I'd add 85 mph (Texas) and 5, 10 and 15, all of which are used for residential areas and sometimes service roads. Anything higher only appears to be used on rail lines.

bhousel:
I merged this and added 5, 10, 15 because...

1. It's not so bad to scroll past the values. 
2. Some mappers are reluctant to enter values that don't appear in the list
3. People who are adding values quickly won't use the list, they will just type them.

PR #3321, opened 2016-08-03
Add recreation_ground preset

boothym:


bhousel:
Thanks @boothym!  I'm surprised we did not have a preset for this already..

I squashed your 2 commits into 1, then ran `make` and merged as:
b3d8d13 2016-08-03 Run make <Bryan Housel>
d47fa11 2016-08-03 Add recreation_ground preset <boothym>

PR #3333, opened 2016-08-08
Turn lanes: Lane editor

kepta:
This PR adds complete turn lanes parsing and test cases.


bhousel:
Merged this today so it doesn't fall too far behind... Thanks @kepta 👍 
I'm going to do a little mini refactor with #3497 that will also split the osm lane code out into its own module (similar to how we have multipolygon code).  We knew it would outgrow `way.js` eventually!

kepta:
@bhousel thanks for taking out time to merge it. 
These days I am a bit busy with  school work. 
Will pick up this soon.

PR #3334, opened 2016-08-08
Fix defaults

mdrouhard:
fixed defaults bug


mdrouhard:
Merging fix-defaults branch

PR #3339, opened 2016-08-10
Fixed validate function

McKaneAndrus:
Preset validation failed to identify schema errors because `jsonschema.validate(instance, schema).length` is always `undefined`.

This PR checks `jsonschema.validate(instance, schema).errors` instead.

### DIFF
--- a/build.js
+++ b/build.js
@@ -25,10 +25,10 @@ function rp(f) {
 }
 
 function validate(file, instance, schema) {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks this is great!
I was wondering why we didn't start seeing schema validation errors after 4922a37, which added `networkCombo` as a field type, but didn't update the schema.

PR #3360, opened 2016-08-18
Performance improvement. Debounce redraw call.

guillaume:
This enhances panning speed especially when tiles are already loaded. 
My understanding is that the previous code will `clearTimeout()` only the last call to `queueRedraw` since the `timeoutId` variable gets overwrote upon each call

## verify

in both this branch and `origin/master`:

* go to an urban area for instance http://localhost:8080/#background=Bing&map=16.52/-71.28537/46.77116
* Pan around the buffer to make sure you get all the tiles in browser cache
* activate google chrome profiling
* pan around for a couple of seconds make sure the panning is small enough so that no new tiles are loaded (otherwise this ain't scientific)

Screenshots below show for a panning of `7000ms`. 

In `origin/master`, notice that the several redraw calls are stacked even under 750ms

screen shot 2016-08-18 at 1 05 37 am

in this branch, `redraw` calls are "debounced"
screen shot 2016-08-18 at 1 02 40 am

This also speeds up when zoomed further up like zoom 18 and such since less tiles need to be downloaded in normal interaction.

### DIFF
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -217,8 +217,6 @@ export function Map(context) {
     function redraw(difference, extent) {
         if (!surface || !redrawEnabled) return;
 
-        clearTimeout(timeoutId);
-
         // If we are in the middle of a zoom/pan, we can't do differenced redraws.
         // It would result in artifacts where differenced entities are redrawn with
         // one transform and unchanged entities with another.
@@ -254,10 +252,7 @@ export function Map(context) {
         return map;
     }
 
-    var timeoutId;
-    function queueRedraw() {
-        timeoutId = setTimeout(function() { redraw(); }, 750);
-    }
+    var queueRedraw = _.debounce(redraw, 750);
 
     function pointLocation(p) {
         var translate = projection.translate(),


guillaume:
zoom 18 profile under `6000ms` quick short panning interaction was done around this area http://localhost:8080/#background=Bing&map=18.38/-71.29213/46.77116

`origin/master`
screen shot 2016-08-18 at 1 19 45 am


this branch
screen shot 2016-08-18 at 1 19 49 am

modules/renderer/map.js
```diff
@@ -217,8 +217,6 @@ export function Map(context) {
     function redraw(difference, extent) {
         if (!surface || !redrawEnabled) return;
 
-        clearTimeout(timeoutId);
```
bhousel: This overall looks really good to me, I can't wait to try it out.

I still think you'd need a `queueRedraw.cancel()` in here, because `redraw()` can be called from other places (e.g. data loaded from osm or other layers) - and you wouldn't want a stray debounced `redraw()` to sneak in after another one has already happened.

guillaume: i think you're right. In practice this would only do one more redraw call (from the panning) after the difference/extent call and would still be a perf improvement when comparing to the previous implementation. I'll dig in some more later and do an amendment to the PR.

bhousel: I thought about it a little more, and it probably needs to be:
`if (!difference && !extent) { queueRedraw.cancel(); }`
Because a partial redraw shouldn't cancel a full redraw.

Maybe `queueRedraw` could be named `debouncedFullRedraw` to make it clear. (sorry, I actually hate nitpicking on variable names in PRs).

bhousel:
Just a quick update, I merged this today, thank you again!.. 

I'm using `_throttle` instead of `_debounce`, because I want redraw calls to happen occasionally, not to wait until all events have stopped.   The reason for this is because iD supports eased transitions (e.g. `id.map().centerEase([-73.9649, 40.7851], 10000)`) - these are implemented using d3 zoom transition, which just fires a bunch of events at the zoom/pan handler, and we need some of those to result in draws.

PR #3361, opened 2016-08-18
Remove population from locality

boothym:
As per the wiki, place=locality is for unpopulated places


bhousel:
screenshot 2016-08-18 15 05 56

boothym:
lol, didn't notice that - I've removed it from the wiki

PR #3363, opened 2016-08-18
Add Bowling Green preset

boothym:
Came across many bowling greens in the UK but there's no preset.


bhousel:
Looks good to me, thanks!

PR #3364, opened 2016-08-18
Change instructions to use node's http-server instead of going into python land

guillaume:
http-server for node.js is very convenient, and is a lot faster than Python's SimpleHTTPServer. This is primarily because it uses asynchronous IO for concurrent handling of requests, instead of serialising requests.

> 🎶  it's a node world after all 🎶 

Enjoy the ride! https://www.youtube.com/watch?v=iJFGAX77zw4

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "scripts": {
     "test": "npm run lint && make && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html dot",
     "start": "rollup --config=./rollup.config.js -w --input ./modules/id.js --output dist/iD.js",
+    "web": "node_modules/http-server/bin/http-server",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "test": "npm run lint && make && phantomjs node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js test/index.html dot",
     "start": "rollup --config=./rollup.config.js -w --input ./modules/id.js --output dist/iD.js",
+    "web": "node_modules/http-server/bin/http-server",
```
tmcw: `"web": "http-server",`

Should be sufficient. Node scripts have the local `node_modules/.bin/` automatically in their path

guillaume: correct 👍

tmcw:
👍 Thanks!

Issue #3076, opened 2016-04-20
When reversing a line, reverse tags on nodes


homersimpsons:
If someone reverse a line (because he want or make a mistake), all the direction become false. The it would be good if, in iD, when someone reverse a line the direction change to math the previous (or open a box to ask if it need to be done). This need to be reverse too:
backward <> forward
left <> right

bhousel:
iD already does this.

homersimpsons:
Sorry, misunderstanding, it's about POI on the road like stop sign.

homersimpsons:
Thank you :) that wasn't too explicit in fact.

bhousel:
Ok, that makes sense, although I don't think any of the other editors do this. 

The stop sign example is a good one.  I'm wondering if reversing the tags on child nodes would break any of the commonly used tags, but I can't think of any issues.

cc @pnorman

pnorman:
I had to check taginfo, but the direction=forward/backward taghas about 35k uses. It's not the clearest data model, as a node can be a child of multiple ways.

stephan75:
Maybe see http://wiki.openstreetmap.org/wiki/Forward_%26_backward,_left_%26_right as a kind of reference?

But as far as I have seen there is no mention of stop signs as node with direction=forward or =backward ... maybe wiki page needs updating?

homersimpsons:
@stephan75 have a look to [stop on wiki](http://wiki.openstreetmap.org/wiki/Tag:highway%3Dstop )

bhousel:
or http://wiki.openstreetmap.org/wiki/Key:traffic_sign#As_part_of_a_way
(279k nodes mapped currently)

bhousel:
closed by #3376, thanks @Psigio!

PR #3376, opened 2016-08-24
Reverse tags on nodes belonging to way on reversal (#3076)

Psigio:
Hi,

I've created this pull request for an implementation to address issue #3076.  I've added some unit tests for operation and also tested it out in a local iD build and it seems to work.  Can you please review and let me know if it's ok?

Thanks!

Jon


bhousel:
Thanks @Psigio this is fantastic!  I'm sorry I didn't get to merging it sooner..

Issue #3025, opened 2016-03-09
Support different icon sets


bhousel:
One actionable thing we can do to get ready for #3024 would be to support icon namespacing in the presets.

For example, we currently do:

``` js
{
    "icon": "bicycle",
    ...
    "tags": {"shop": "bicycle"},
    "name": "Bicycle Shop"
}
```

And we use a little bit of magic to know that `bicycle` is a Maki icon, and not one found in the iD spritesheet.

We should namespace this string to:
1. eliminate magic "is it maki or not" code 
2. support both maki1 and maki2
3. open up the possibility for people with custom presets to use other iconsets:

``` js
    "icon": "maki1:bicycle"
    "icon": "maki2:bicycle"
    "icon": "fontawesome:bicycle"
```

kepta:
We change all the icons which we get from maki to "x" => "maki1:x",
and then in the file,
*id/ui/preset_icon.js:12*
``` Javascript
  icon = p.icon || (geom === 'line' ? 'other-line' : 'marker-stroked'),
```
we do some logic to determine whether it is coming from maki1, maki2 or fontawesome.
Is my understanding correct ?

bhousel:
@kepta, it's more complicated than that.. I started to list out the places in code we would need to change, but it got to be quite a lot of things, and I'm still not sure myself if what I am thinking is the best way to support this.  I need to think about it for a few days to decide if it's even worth doing.

Key challenges to supporting multiple icon sets are:
- code in `defs.js` needs to namespace all of the svg id's to avoid DOM conflicts.
- different icon sets support different icon resolutions.. e.g. maki1 is `-12`, `-18`, `-24`, maki2 is `-11`, `-15`
- changes in `preset_icon.js`, `labels.js`, etc..
- do this in a way to support splitting out the presets into separate projects per #2656

Also forget what I said about fontawesome - it's really a [webfont, not a svg icon set](http://www.sitepoint.com/icon-fonts-vs-svg-debate/).  But I do hope more designers move towards offering svg icon sets now that GitHub has done it.   Even if an icon repository has the svg source files in it, we could package them up ourselves into a sprite they way we already do with maki.

bhousel:
For now let's just support maki2 and we will revisit later the idea of supporting multiple spritesheets if an iD fork project needs this.

bhousel:
This is happening anyway for the cleanups for #3924
I've split the internal iD poi icons out into a new library: https://github.com/bhousel/temaki

Issue #3024, opened 2016-03-09
Support maki2


bhousel:
This is just a placeholder issue to keep an eye on 
~~https://github.com/mapbox/maki-2/issues/46~~
~~Maki2 will likely be merged into Maki, which will get a major semver bump.~~

^UPDATE: this happened already, see https://github.com/mapbox/maki/pull/241

  This is good!

iD is currently pinned to Maki 0.5.0, so we don't need to do anything right away.

We might be able to [install both versions](https://github.com/npm/npm/issues/5499#issuecomment-129481232), and build 2 spritesheets so that we don't lose any of the Maki1 icons that aren't yet available in Maki2.

bhousel:
related ticket in Maki:  https://github.com/mapbox/maki/issues/270

pnorman:
https://github.com/mapbox/maki-2/issues/46 gives a 404

bhousel:
Thanks @pnorman I updated the issue.
`maki2` repo is gone and its contents got folded into the main `maki` repo.
see https://github.com/mapbox/maki/pull/241

nickidlugash:
@bhousel I just wanted to flag that Maki 2 icons are different sizes than Maki 1: they are 11px/15x versus 12px/18px/24px. It looks like you're currently using Maki 1 `-24` for the UI, resized to 32px, and Maki 1 at actual size for the map? Will it work for your purposes to use both versions at the same time?

/cc @natslaughter @aparlato

bhousel:
> Will it work for your purposes to use both versions at the same time?

It would probably be ok, but we'll need to add some code to store what icon sizes are available and have iD pick the best size.  See #3025 for more info

PR #3380, opened 2016-08-27
Add fire-station icon for fire-hydrant

bagage:
Fire hydrants are for firemen anyway, and it help distinguish them from other no-icon point on the map. 

By the way, is there any plan to upgrade maki from version 0.5.0 to latest version 2.0.x? Is that OK if I migrate iD by updating icons' name or there is a particular reason to use maki 0.5.0?

### DIFF
--- a/data/presets/presets/emergency/fire_hydrant.json
+++ b/data/presets/presets/emergency/fire_hydrant.json
@@ -1,4 +1,5 @@
 {
+    "icon": "fire-station",
     "fields": [
         "fire_hydrant/type"
     ],


bhousel:
> By the way, is there any plan to upgrade maki from version 0.5.0 to latest version 2.0.x? Is that OK if I migrate iD by updating icons' name or there is a particular reason to use maki 0.5.0?

I'd like to upgrade to maki2, but there's a bit of work for us to be able do this.  see #3024, #3025 
Because it uses different icon sizes, it's not a drop in replacement.  We have a bunch of places in the iD code that pick the right icon and construct a name with the right size (e.g. it turns `fire-station` into `fire-station-12`, `fire-station-18`, or `fire-station-24`).

bagage:
Ooooh, it looks like a considerable work actually! I subscribed to these issues to keep track of any progress here (and for any `get-started` label if any!). Thanks @bhousel.

PR #3381, opened 2016-08-28
Add preset for amenity=coworking_space

willemarcel:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcoworking_space


bhousel:
Thanks @willemarcel!

PR #3399, opened 2016-09-02
parseInt cleanup

HolgerJeromin:
parseInt should always have a radix.
Even if i have not found a problematic value (leading zero) in Taginfo. Surprising :-)

Removed one call of parseInt which was called on a number.

### DIFF
--- a/modules/core/way.js
+++ b/modules/core/way.js
@@ -146,11 +146,11 @@ _.extend(Way.prototype, {
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


HolgerJeromin:
I am surprised that eslint on travis did not complained as the radix rule is [default active](http://eslint.org/docs/rules/radix).

bhousel:
Looks like the `"extends": "eslint:recommended"` flag turns the rule "off".
I agree we should leave this rule on to catch issues with `parseInt`.

PR #3405, opened 2016-09-08
Fix way disappearing due to invalid "layer" tag

edpop:
Hello.

Now layer() returns only a finite number, not NaN.

/* Chrome 53 */


bhousel:
Thanks for finding this!   Wondering if there are other places in our code where we try to coerce tag values to a numeric this way.

Just flagging for @kepta, to check if your lanes parsing code has similar issues.

PR #3456, opened 2016-09-26
Add capacity field to amenity=charging_station preset

boothym:
Suppose there should be more fields with all the socket options etc, but this will do for now.


bhousel:
Thanks!  I really appreciate all the work you are doing to improve iD's presets..

boothym:
No probs, now we just need a release to use them :P

PR #3466, opened 2016-09-29
update installation instructions and prerequisites

tyrasd:
This updates the readme to reflect recent changes in the project structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Come on in, the water's lovely. More help? Ping `jfire` or `bhousel` on
 
 ## Prerequisites
 
-* [Node.js](http://nodejs.org/) version 4 or newer, ([npm](https://www.npmjs.com/package/npm) version 3 or newer)
+* [Node.js](http://nodejs.org/) version 4 or newer
 * Command line development tools (`make`, `git`, and a compiler) for your platform
   * Ubuntu:
     * `sudo apt-get install build-essential git`
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Come on in, the water's lovely. More help? Ping `jfire` or `bhousel` on
 
 ## Prerequisites
 
-* [Node.js](http://nodejs.org/) version 4 or newer
+* [Node.js](http://nodejs.org/) version 4 or newer, ([npm](https://www.npmjs.com/package/npm) version 3 or newer)
 * Command line development tools (`make`, `git`, and a compiler) for your platform
   * Ubuntu:
     * `sudo apt-get install build-essential git`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,7 +28,7 @@ Come on in, the water's lovely. More help? Ping `jfire` or `bhousel` on
 
 ## Prerequisites
 
-* [Node.js](http://nodejs.org/) version 4 or newer
+* [Node.js](http://nodejs.org/) version 4 or newer, ([npm](https://www.npmjs.com/package/npm) version 3 or newer)
```
tyrasd: node v4 comes with npm v2 by default (AFAICT) – so it could also make sense to shorten this requirement to something like `node.js version 5 or newer` (so that users don't have to worry about checking and eventually manually upgrading their npm version)?

README.md
```diff
@@ -28,7 +28,7 @@ Come on in, the water's lovely. More help? Ping `jfire` or `bhousel` on
...
 * Command line development tools (`make`, `git`, and a compiler) for your platform
```
tyrasd: We don't actually require a *compiler* to be installed, right?

tyrasd:
some stuff in [RELEASING.md](https://github.com/openstreetmap/iD/blob/tyrasd-update-installreadme/RELEASING.md#update-master-branch) probably needs an update as well (e.g. we don't have a `make suggestions` command anymore).

README.md
bhousel: Yeah, I dont understand why the npm 3 requirement was added.  node 4 / npm 2 should be fine.  I set Travis to node 5 for reasons unknown, so let me put it back to 4 and confirm..

bhousel: Looks good with node 4, 5, 6:  https://travis-ci.org/openstreetmap/iD/builds/163742016

README.md
bhousel: hmm, we probably did need it for something at one point (osmium?), but we're really close to dropping this entire bullet point.  Still using `make` for a few things, but it can be dropped.

README.md
tyrasd: Wow, that's odd: When I tried to run [these tests](https://travis-ci.org/openstreetmap/iD/jobs/163742017) here with node v4.6.0 + npm v2.15.9 (installed via `n`), I got an error:

```
> node build.js && node development_server.js

Rebuilding
[Error: Could not resolve 'd3-transition' from …/iD/modules/behavior/breathe.js]
Error loading resource file:///…/iD/dist/iD.js (203). Details: Error opening …/iD/dist/iD.js: No such file or directory
ReferenceError: Can't find variable: iD
```

In fact, after `npm install`ing with npm v2, the d3-transition module is nested inside the d3 one: `iD/node_modules/d3/node_modules/d3-transition`, while it's a top level module in npm v3: `iD/node_modules/d3-transition`

tyrasd: Oh, wait. You fixed that one in https://github.com/openstreetmap/iD/commit/348a973f688d2921f37d2cf5c4ef7cabbe7bc941. (Why you no tell me? :stuck_out_tongue_closed_eyes: )

README.md
tyrasd: Hmm. As far as I see, you added that comment relatively recently in 3f5d7ec7. It's not really important, though. I'm just curious whether or not I'm missing something. (Maybe some dependency needs to be built from source on some OS'es? phantomjs?)

README.md
bhousel: > Oh, wait. You fixed that one in 348a973. (Why you no tell me? 😝 )

I'm sorry! I thought you would see it.. 😬

PR #3471, opened 2016-10-02
Add pastry shop preset

### LINKED ISSUE #3444: Add shop=pastry
See https://wiki.openstreetmap.org/wiki/Tag:shop%3Dpastry

Dgleish:
(closes #3444)

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -3650,6 +3650,7 @@ en:
       shop/pastry:
         # shop=pastry
         name: Pastry Shop
+        # 'terms: patisserie,cake shop,cakery'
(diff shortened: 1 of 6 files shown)


bhousel:
Thanks for adding this!  Looks ok, but can you name the file: 
`data/presets/presets/shop/pastry.json`

By convention we try to make the name of the file the same as the value of the tag.

Also it looks like the `terms` array is missing from there.
```
"terms": [
        "patisserie",
        "cake"
    ],
```

The terms were in `en.json` only - maybe you manually added those?  You only need the add the `pastry.json` file and then run `npm run build` and it will make the other files automatically.

Dgleish:
Sorry, missed that section of the readme

bhousel:
Thanks, this looks great!

PR #3475, opened 2016-10-04
Added Anime Terms to Anime Shop

mbrickn:
Added terms include things you might purchase at an anime shop, plus a catch all "Japan" term.

### DIFF
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -3569,7 +3569,7 @@
             },
             "shop/anime": {
                 "name": "Anime Shop",
-                "terms": ""
+                "terms": "anime, manga, figurine, cosplay, dakimakura, japan"
             },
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks!  But `en.json` is a generated file that gets overwritten when rebuilding the presets.

The correct place to add this is directly in the [`anime.json` preset file](https://github.com/openstreetmap/iD/blob/master/data/presets/presets/shop/anime.json).

```
    "terms": [
        "manga", "figurine", "cosplay", "dakimakura", "japan"
    ],
```
(You don't need to include "anime" in the terms, because it's the tag value and gets included as a search term automatically).

mbrickn:
Ah, sorry about that! I misunderstood things. Thank you!

PR #3478, opened 2016-10-04
Updated Anime Terms

mbrickn:
Updated Anime Shop preset with anime related terms. ^_^


bhousel:
Looks good, thanks for the quick fix!

PR #3516, opened 2016-10-23
add preset for nail salon

skorasaurus:
Here's a pull request that adds nail salons (where people have their fingernails lacquered and/or manicured) to the generic preset. 

The common tags for tagging these are shop=beauty and beauty=nails. [taginfo](http://taginfo.openstreetmap.org/search?q=nails#values) and [wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbeauty)

I wasn't sure on the correct file location for this preset but I created a new folder within the shop folder, and 
named it after its parent tag, beauty. 

This structure is based the practice of presets whose meaning are not defined so much by their first tag but rather by their second tag. This occurs with a mosque which has amenity=place_of_worship and religion=muslim)[https://github.com/openstreetmap/iD/tree/master/data/presets/presets/amenity/place_of_worship]. 

Bryan, would you like iD continue with this precendence? 

selection_081
My other question regarding this PR is that after adding this preset, rebuilding the presets, and testing it locally, there's an additional box of 'beauty' as indicated in the image below. Is there a way to hide this?

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -3345,8 +3345,13 @@ en:
       shop/beauty:
         # shop=beauty
         name: Beauty Shop
-        # 'terms: nail spa,spa,salon,tanning'
+        # 'terms: spa,salon,tanning'
         terms: '<translate with synonyms or related terms for ''Beauty Shop'', separated by commas>'
+      shop/beauty/nails:
+        # 'shop=beauty, beauty=nails'
+        name: Nail Salon
+        # 'terms: Nail Salon,nails'
+        terms: '<translate with synonyms or related terms for ''Nail Salon'', separated by commas>'
       shop/bed:
         # shop=bed
         name: Bedding/Mattress Store
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -8495,7 +8495,6 @@
             "area"
         ],
         "terms": [
-            "nail spa",
             "spa",
             "salon",
             "tanning"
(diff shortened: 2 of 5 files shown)


bhousel:
Thanks for this @skorasaurus!  You are right that this preset works a lot like the religion one, so adding a subfolder under `shop` makes sense (just like for religion, there is a subfolder under `amenity`).

To answer your question, that extra "type" field was because you added the `shop` field.  I removed it, so it won't show anymore for the Nail Salon preset.   The reason we have generic fields like this is in case someone creates a shop with a type that we don't recognize as a preset.  For example, somebody could tag a business as `shop=halloween` and even though we don't have a preset for that specific kind of shop, it will show the "generic" shop, and that type field will say "halloween".  We can leave it off from the specific presets.

A few other points:

* I added a Tanning Salon preset for `beauty=tanning` also, it is the next most common preset after `beauty=nails`.
* You don't need to list "Nail Salon" or "Nails" in the terms list for `beauty=nails` because it's in the value already.  I replaced those with "manicure" and "pedicure" which are related words that people might search for.

Again, thanks so much for your contribution!

1ec5:
> I added a Tanning Salon preset for beauty=tanning also, it is the next most common preset after beauty=nails.

[`leisure=tanning_salon`](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dtanning_salon) is documented on the wiki and used almost five times as much.

bhousel:
> leisure=tanning_salon is documented on the wiki and used almost five times as much.

Yes, but the wiki page for `leisure=tanning_salon` is pretty thin, and the talk page is just one comment saying "what is this doing here?".  That tag looks like a mistake to me..

PR #3525, opened 2016-10-26
Add basic docker-compose support

Firefishy:
Add basic docker support.

To run: docker-compose up
(and it will run on port 8080)


.dockerignore
```diff
@@ -0,0 +1 @@
+# TODO
```
pnorman: Is this ready yet?

Firefishy: They contain builds and runs as expect, but likely could have a few things excluded. Eg: .git

Happy for you to suggest or I will look properly in a day or 2.

pnorman: Maybe copy `.npmignore`?

bhousel:
Thanks @Firefishy, but I'm going to close this for now.. I think using Docker is great when the development environment is complicated and has a lot of dependencies, but we've already done a bunch of work to make sure that the only requirement for building iD is to have a Node.js 4.0 or newer, which is a pretty easy requirement to satisfy.

PR #3537, opened 2016-11-01
Validate preset defaults at build step

gmaclennan:
Adds an additional presets consistency check at build time to ensure `defaults.json` refers to categories and presets that actually exist.


bhousel:
good idea, thanks @gmaclennan!

Issue #3250, opened 2016-07-18
"Viewport Following" feature in iD


chtnha:
If we want to map very long `highway/river`s  features, we need to pan imagery a lot. It's pain and time consuming. It'd be better if we have tool like [Viewport Following](https://josm.openstreetmap.de/wiki/Help/Action/ViewportFollowing) feature in iD which exists in Java OpenStreetMap Editor (JOSM).

I have shared the illustration here: 

![viewport_following](https://cloud.githubusercontent.com/assets/6770741/16921317/b2109a62-4d2e-11e6-8858-d552822de5bb.gif)

@bhousel @kepta

Abbe98:
I could work on this,  mainly to learn all this new modular iD stuff. If anyone got the time to give me some pointers and be accessible on IRC if I run into something.

I guess we would like a checkbox and if that checkbox is checked we would like to center the map view on every (map)click?

bhousel:
> I guess we would like a checkbox and if that checkbox is checked we would like to center the map view on every (map)click?

Yes, but the checkbox would be a toggle under whatever taskbar comes out of #3123, so it's probably at least a month out.  

If you want to learn modular iD, we could use help with #3180.. Lets start replacing our custom lodash build with only the lodash modules we really need!  (Comment there to coordinate)

bhousel:
Just a quick update that some things in iD do viewport follow now:

* Deleting vertices
* Undo/Redo
* Vertex navigation 
  `{`: first
  `}`: last, 
  `[`: previous, 
  `]`: next, 
  `\`: reparent

Leaving this issue open because we don't yet do viewport following in draw mode.   I have to try it out, and we might want to make it an option.  One concern is what it will do if someone holds down the spacebar and draws a lot of nodes.

Issue #2812, opened 2015-10-21
Cannot open Minimap on Non-US keyboards


tyrasd:
The keyboard shortcut to open the minimap (#2554) is the slash `/` character, which is input on a German keyboard layout via `shift`+`7`. iD doesn't trigger the minimap-toggle when pressing this key combo. Strangely, pressing the `#` character on a German keyboard opens the minimap on Chrome, but not on Firefox.

I'm assuming this is because of using the `keyCode` property [here](https://github.com/openstreetmap/iD/blob/0543f684d2b431eb64f3918bad344f50d3c17733/js/lib/d3.keybinding.js#L75), which seems to be a [bit broken](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode):

> Web developers shouldn't use the keycode attribute for printable characters […] especially those input with the Shift or Alt key pressed. When implementing a shortcut key handler, the keypress event is usually better.

bhousel:
Hmm, yeah keycode detection is kind of broken.  

It's hacky, but maybe for now (in lieu of changing `d3.keybinding`) we can find a keycode on the German keyboard that isn't used for anything.

What codes do you get when you visit this jsfiddle, and try keys like ä,ö,ü,ß
http://jsfiddle.net/YKeye/

tyrasd:
character | keycode (chromium) | keycode (firefox) | keycode (old presto based opera)
------------ | ------------- | ------------- | -------------
`/` | 55 (then 16 on releasing the shift key) | 55 (then 16) | 55 (then 16)
`ä` | 222 | 0 | 222
`ö` | 192 | 0 | 186
`ü` | 186 | 0 | 219
`ß` | 219 | 63 | 109

It's a mess! :poop:

Abbe98:
This is broken using Swedish and Norwegian keyboards too.

bhousel:
I did some research today, and yeah it's a mess.  This will hopefully get better when [DOM Level 3 Keyboard Events](http://www.w3.org/TR/DOM-Level-3-Events-key/) are implemented in all browsers, but that's a long way off.

For now some decent resources are [Unixpapa's summary](http://unixpapa.com/js/key.html) and [this implementation](http://stackoverflow.com/questions/5306132/translate-javascript-keycode-into-charcode-for-non-u-s-keyboard-layout-i-e-az) by the noVNC developer.

bhousel:
Check French keyboard too, per #3660

[...]

Issue #1293, opened 2013-04-11
Problems with presets and tagging on closed ways drawn with line tool instead of area tool


bussed:
I draw a closed way using the line drawing tool (not area tool).

Now i want to tag it with "building=yes".

At first i try using the Building preset but it wont show up when i search for it. No chance.

So i fall back using the "Other" preset.
At the bottom in the inspector view i add the tag building=yes. But this tag is never applied. After reopening the inspector view the tag is lost.

So i end up that i cannot apply a building=yes on a closed way drawn with the line tool.

I do really like that users get a push using the area drawing tool instead of using the line drawing tool. But to have no chance to apply the tag at all is a bit too strict in my opinion.

Issue #1239, opened 2013-04-02
Handle coincident features (i.e. multiple things one spot)


ansis:
This is a hard one.

Trams are drawn as separate ways that share the nodes of the road. http://www.openstreetmap.us/iD/master/#id=w33910429&layer=Bing&map=19.72/-79.41092/43.66473

Its impossible to select way rendered on the bottom (sometimes the tram, sometimes the road).

I'm seeing a couple possibilities.
1. Offset the line paths so that they don't overlap. 
2. Clicking on a line segment selects all parent ways. Add inspector for multiple entities, a list that lets you choose one of them.
3. Not handle this at all. Let users disconnect and move if they need to edit.

~~P2 and JOSM don't handle this at all.~~

SomeoneElseOSM:
Just checking to see if anyone's implemented a fix to this and forgotten to update the issue :)

The context is the forum discussion about this changeset http://www.openstreetmap.org/changeset/35882220 (see the links from the changeset discussion) - I'd like to be able to say "just press <something>" to switch between contiguous ways rather than suggesting using another editor.

Shift-click "sometimes" works - it seems to depend on draw order.  There's a couple of test overlaps at http://api06.dev.openstreetmap.org/#map=18/51.50434/-2.65264 if anyone wants to experiment.  Another options seems to be clicking the end node, clicking "+" on the rotary menu, getting told there are several lines to extend, clicking the only place you can and extending one of the duplicates, then in the same way extending the _other_ duplicate in a different direction. Once you can select each way individually, you can delete the duplicate that you don't want, and remove the extra bit of the other one that you added, but it's not a workflow to suggest to someone who's just started mapping.

bhousel:
No we haven't implemented this yet.

I'm not really sure the best way to handle this either.  The current rendering can make it impossible to know that there even are coincident ways.

slhh:
In order to make the user aware of coincident ways the cursor symbol could show multiple lines (preferably three lines) instead of a single line when hovering over the coincident ways. In addition the colour of the cursor symbol can change for example to red making the cursor symbol change more evident.

The user might stiil fail to recognize it. Therefore iD should not use something similar to the P2 "/" or JOSM Alt-Click method, but open a selection menu when clicking the coincident ways.

slhh:
[...]

PR #3539, opened 2016-11-02
Add vertex keyboard navigation

### LINKED ISSUE #1917: An option to move to another end of the selected line
I've tried to map bus routes with iD and the most time-consuming action is finding another end of the selected line.
I think having a shortcut to move to the end of selected line that's farther away from the view center would help a lot.

bhousel:
Today I played around with keyboard navigation for vertexes.  

This closes #1917 and adds some useful keyboard shortcuts:

`[` - jump to previous vertex
`]` - jump to next vertex
`⌘[` - jump to first vertex
`⌘]` - jump to last vertex

Also it: 
* handles circular ways (can jump next/previous across the connecting node)
* adds interpolated viewport following  (see #3250 - which is this but in the draw modes)
* when navigating across a vertex with multiple parents, remembers which line to stay on

keyboard node navigation

There are still a few weird things to clean up - e.g. it's possible for the vertex to go unselected when jumping to one that's offscreen.

Also I'd like to try to support a shift modifier, which woud let you add or remove nodes from the selection.

### DIFF
--- a/css/map.css
+++ b/css/map.css
@@ -37,6 +37,7 @@ g.point .shadow {
     stroke-opacity: 0;
 }
 
+g.point.related:not(.selected) .shadow,
 g.point.hover:not(.selected) .shadow {
     stroke-opacity: 0.5;
 }
@@ -104,7 +105,9 @@ g.vertex.vertex-hover {
     display: block;
 }
 
+g.vertex.related:not(.selected) .shadow,
 g.vertex.hover:not(.selected) .shadow,
+g.midpoint.related:not(.selected) .shadow,
 g.midpoint.hover:not(.selected) .shadow {
     fill-opacity: 0.5;
 }
@@ -144,6 +147,7 @@ path.shadow {
     stroke-opacity: 0;
 }
 
+path.shadow.related:not(.selected),
 path.shadow.hover:not(.selected) {
     stroke-opacity: 0.4;
 }
@@ -1600,6 +1604,7 @@ text.gpx {
     stroke-width: 8;
 }
 
+.fill-wireframe path.shadow.related:not(.selected),
 .fill-wireframe path.shadow.hover:not(.selected) {
     stroke-opacity: 0.4;
 }
--- a/modules/lib/d3.keybinding.js
+++ b/modules/lib/d3.keybinding.js
@@ -58,39 +58,44 @@ export function d3keybinding(namespace) {
         return keybinding;
     };
 
-    keybinding.on = function(code, callback, capture) {
-        var binding = {
-            event: {
-                keyCode: 0,
-                shiftKey: false,
-                ctrlKey: false,
-                altKey: false,
-                metaKey: false
-            },
-            capture: capture,
-            callback: callback
-        };
+    keybinding.on = function(codes, callback, capture) {
+        var arr = [].concat(codes);
+        for (var i = 0; i < arr.length; i++) {
+            var code = arr[i];
+            var binding = {
+                event: {
+                    keyCode: 0,
+                    shiftKey: false,
+                    ctrlKey: false,
+                    altKey: false,
+                    metaKey: false
+                },
+                capture: capture,
+                callback: callback
+            };
 
-        code = code.toLowerCase().match(/(?:(?:[^+⇧⌃⌥⌘])+|[⇧⌃⌥⌘]|\+\+|^\+$)/g);
+            code = code.toLowerCase().match(/(?:(?:[^+⇧⌃⌥⌘])+|[⇧⌃⌥⌘]|\+\+|^\+$)/g);
 
-        for (var i = 0; i < code.length; i++) {
-            // Normalise matching errors
-            if (code[i] === '++') code[i] = '+';
+            for (var j = 0; j < code.length; j++) {
+                // Normalise matching errors
+                if (code[j] === '++') code[i] = '+';
 
-            if (code[i] in d3keybinding.modifierCodes) {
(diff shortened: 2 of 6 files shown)


bhousel:
Just hammering down on the `]` key...

keyboard node navigation 2

tyrasd:
Does it work on non-English keyboard layouts, too (see #2812)? Also, on a German keyboard `⌘[` is quite awkward to hit :unamused:

bhousel:
> Does it work on non-English keyboard layouts, too (see #2812)? Also, on a German keyboard ⌘[ is quite awkward to hit 😒

Hmm, yeah I do want to pick something easy to hit on QWERTZ.  Maybe we can use Ö and Ä?
What key shortcuts are used to indent / unindent code blocks in editors like sublime or xcode?

tyrasd:
> What key shortcuts are used to indent / unindent code blocks in editors like sublime or xcode?

I'm used to press Tab / Shift-Tab for that (or `>` / `<` in vi).

`vi` (and other unix cli tool) use `j` and `k` to [navigate](https://en.wikipedia.org/wiki/Arrow_keys#HJKL_keys) up/down in lists or lines of text. Another idea might to use the ubiquitous `wasd` navigation as an inspiration…

slhh:
The following keys might be used instead: 
Page Up - jump to previous vertex
Page Down - jump to next vertex
Home - jump to first vertex
End - jump to last vertex

The vertex navigation is nice, but the usecases seem to be very limited without some more new shortcuts:
-  toggle selection between the vertex and the line to stay on.
-  open the radial menu of the selected vertex or line.
- at a connecting vertex: cycle connected lines as the line to stay on. 

This would also be a (partial) solution for #1293,

The line to stay on should be made visible during vertex navigation. This can be done by temporarily flashing a halo around it when any vertex navigation shortcut is used.

bhousel:
> Page Up - jump to previous vertex
> Page Down - jump to next vertex
> Home - jump to first vertex
> End - jump to last vertex

Sounds good to me.. I was kind of hoping to provide keys that exist on a laptop keyboard, but at least offering these as alternatives helps for the keyboard layouts that don't have handy `[` and `]`.


>The vertex navigation is nice, but the usecases seem to be very limited without some more new shortcuts:
> toggle selection between the vertex and the line to stay on.

Maybe.. I will think about it and try a few ideas.

> open the radial menu of the selected vertex or line.

`spacebar` does this  (although everything on the radial menu already has a keyboard shortcut).  I did try it and amusingly the radial menu is in the wrong place (because of the follow behavior moving the view) - so that's something else to fix..

> at a connecting vertex: cycle connected lines as the line to stay on.

This cycling through parent ways (and the halo flash idea needed to make it visible) would add a bit more complexity than I'm comfortable with.

> This would also be a (partial) solution for #1293,

Did you mean to link a different issue here?  not sure how that one is related..

slhh:
> <spacebar> does this (although everything on the radial menu already has a keyboard shortcut).

That's good. Unfortunately, it was a secret shortcut. It  was not mentioned in https://wiki.openstreetmap.org/wiki/ID/Shortcuts, which is linked in iD's help function. I have added the Spacebar and also the eight vertex navigation shortcuts.

> Did you mean to link a different issue here? not sure how that one is related.

Sorry, I have meant the issue to handle coincident ways #1239 .
Of course this needs cycling through parent ways, but doesn't need the halo flash idea.
We just need to do the cycling a bit differently. The cycling shortcut needs to select the next way, which is connected to the currently selected vertex or the previously selected vertex (in case a way is currently selected). This does not only show the geometry, but also the properties in the inspector. This helps identifying the correct way. In many cases selecting the connected way is the desired result. In the context of a way being selected due to this cycling function, the vertex navigation shortcuts should automatically switch back to the recently selected vertex, before performing their normal operation.

bhousel:
>> at a connecting vertex: cycle connected lines as the line to stay on.

>This cycling through parent ways (and the halo flash idea needed to make it visible) would add a bit more complexity than I'm comfortable with.

👆 ok I lied.. I added slight hover styling for the related parent way, and added the key `\` to toggle which parent is active, and it's awesome.

whee 
keyboard node navigation 3

bhousel:
Also for anyone wondering why this is useful..
1. jumping to first/last node of a way is helpful to see where it goes.
2. when you delete a node in iD, sometimes the next node that is selects is not the one that you want.
3. cleaning up overnoded ways is great

keyboard node navigation 4

slhh:
I have tested in master deployment:

1. Styling looks good.
2. Home/End/PgUp/PgDown mainly do what they should, but sometimes PgUp/PgDown left the way to stay on at the end vertex and continues with the next way.
3. Selecting a way and a connecting vertex afterwards sometimes the wrong way is selected as the way to stay on.
4. \, [, ], Ctrl-[, Ctrl-] don't do anything on Firefox and Edge under Windows 10. This is likely because I have to use Alt Gr due to the German keyboard.
Maybe ß can be defined as an alternate for \ because it is on the same key, but without requiring Alt Gr.

tyrasd:
> added the key `\` to toggle which parent is active

This would be super useful, but that's also another key which is impossible to press on (at least some) non-English keyboard layouts. :cry:

@slhh: Just hacking in the `ß` key (keycode 219), where the `\` is located on German keyboards, doesn't help much, since the `\` can be on yet another key on other language's keyboards (for example on a French keyboard it's `AltGr+8` behind keycode 56). Another complication is that the keycode for the German `ß` key is 219 (on chrome at least) which is the same as the English `[` key.

Issue #3375, opened 2016-08-24
Feature Suggestion: Flip


Treemanhero98:
I would like to suggest a feature which mirrors (vertical or horizontal) selected structures, especially buildings.
Is this possible to implement it into the editor?

bhousel:
Yes, definitely possible.  Flip Horizontal / Flip Vertical are tools that people are accustomed to seeing,  from working with other vector drawing tools.  This is not a priority for iD, but I'll leave this open as a "maybe someday".

slhh:
I don't see a real usecase of flipping existing objects, but in case of copying objects flippng could be useful. Some keys could have a special function in order to flip, rotate, or scale the copied object in the period between pressing Ctrl-V and placing the copied object by mouse click.

bhousel:
Yes, the main usecase would be for copy/pasting of symmetrical building outlines.

<img width="609" alt="screenshot 2016-08-26 01 33 34" src="https://cloud.githubusercontent.com/assets/38784/17995142/33d5d244-6b2d-11e6-8523-848de0869eb1.png">

bhousel:
This was done in #3375 (thanks @Psigio!)
Try it out in the mirror:  http://openstreetmap.us/iD/master

PR #3555, opened 2016-11-05
Add flip horizontal/vertical operations (for issue #3375)

Psigio:
This is my first attempt at adding a flip horizontal & flip vertical operation for an area-type way (in issue #3375).  It doesn't use the current projection, I'm not sure if it's necessary or not - happy to add if it is!  I've also not been able to add any artwork for the buttons, is there a standard open-source icon repository in use already to source the images from?

I don't think that there's a "copy" operation in iD already, and it was mentioned in the original issue.  Perhaps this should be added as a separate issue/feature request?


slhh:
A "copy" does already exist in ID, but it is available using a shortcut sequence only: Ctrl-C (to copy), Ctrl-V (to paste), click (to place). In the period between Ctrl-V and click some shortcuts can be made available.
The existing rotate "R" shortcut needs also to be made available here, because the reqired flip axis will rarely be the x or y axis. With the rotate operation being available, a flipping at one fixed axis is sufficient, because flipping at any axis can be acheived by flipping at that fixed axis and rotating the object afterwards.

Optionally the spacebar can open a radial menu with tools to modify the temporary, unplaced copy.

Psigio:
Ah yes, that makes sense; I've just tried the Ctrl-C/Ctrl-V shortcuts in iD and I see how it works.  It makes sense to introduce the rotate and make flip work on one axis only as you say.  I'll remove the logic from my PR for the vertical/horizontal distinction and see if I can work out how to attach the flip operation (and existing rotate operation) to the temporary state after the paste operation.

bhousel:
Thanks @Psigio!  I haven't tried it out yet but I can't wait to.. This is something that I would use a lot.

Yes, we do have copy/paste already.. They aren't yet on the menu.

Agree with @slhh that we can get by with just a flip along the vertical axis - we can call it "Reflect" which is the word used in Illustrator.   We can use use `T` as the shortcut key (which looks like a vertical reflected thing and happens to be next to `R`).

For the icon, we'd need to add it to the spritesheet, which is a bit fragile right now.  I can take care of it.  We usually design our own icons for ui elements, in order to make sure they all look cohesive.  (I tend to look at http://fontawesome.io/icons/ for inspiration, or for a starting point, but I don't see anything on there that would be suitable for this).

Psigio:
Thanks - I've made some changes so that "flip" is now called "reflect", it only operates around the vertical axis (ie, it flips an area from left/right as it is viewed on screen), it's bound to the shortcut key T and also it now won't operate if the shape is off-screen or linked to anything hidden.

I've not yet worked out how to attach it to the "after-paste" state, but in it's current state it could be usable as the user can move, rotate & reflect the object after completing a paste operation.

slhh:
I don't like the name/shortcut combination refect/T.
Other ideas are: transpose/T, flip/F, mirror/M
M is already used for move, but in the  "after-paste" state, we could reuse it for mirror, because move is not required in this state.
The icon for an after-paste state menu could be a vertical line with an double-sided horizontal arrow across.

Psigio:
I've had a go at adding the post-paste mode.  I looked at the existing modes and tried to follow the same sort of pattern.  I've not yet made any changes to the keybindings, but that should be easy to do.  There doesn't appear to be much in the way of unit tests for the existing modes, but if this implementation looks ok I'm happy to have a go at adding some!

The post-paste mode pops up the radial window on paste and gives you delete, merge, move & flip operations.  I'm not sure that merge makes much sense, so I can explicitly remove it from the list of operations if necessary.  You can toggle the menu with space, or return to browse with escape.  Any operation will also return to browse after completion, and panning the map also returns to browse.

Let me know what you think!

bhousel:
> The post-paste mode pops up the radial window on paste and gives you delete, merge, move & flip operations. I'm not sure that merge makes much sense, so I can explicitly remove it from the list of operations if necessary. You can toggle the menu with space, or return to browse with escape. Any operation will also return to browse after completion, and panning the map also returns to browse.

I don't understand post paste mode.  Post-paste the user is in Move mode.  They can drop the pasted object anywhere and it should stay selected and get whatever operations are valid for that object.

I'm going to back out the last commit, and focus just on getting Reflect operation merged.

bhousel:
This is awesome work @Psigio.. 🙌 

One thing I might change - It doesn't _quite_ do what I would expect.  I understand that it reflects around the y axis, but I think an untrained user might expect us to derive an axis based on the shape and its orientation.  I'm going to try to add a little bit of magic to the code and see if we can make this even better.

reflect

bhousel:
Quick update on this, I made some progress and I think it's close to being done.

Plan is to:
- [X] Generate smallest surrounding rectangle (done, see below)
- [ ]  Reflect across whichever axis of symmetry is more vertical
- [ ] Update tests
- [ ] Add icon

ssr

bhousel:
Working great now, added 2 icons for reflecting around a shape's long and short axis:

screenshot 2016-12-16 11 14 53

reflect2

bhousel:
This is working really great now, so I'm going to merge it.. I will open a separate ticket for improving the post-paste behavior.

PR #3556, opened 2016-11-05
Custom color for farmyards.

Thue:
A change that turns everything red - well actually only farmyards - and those orange.

Right now, farmyards are not defined in iD CSS, which result in them being colored forest-green while editing. Green is not an intuitive color for a human-centric area. And farmyards are fairly common when mapping, so it seems important that they should have a distinct color.

I used a modified copy of the CSS code for residential areas, and changed the color to the same orange used for farmyards on the standard openstreetmap layer on osm.org.

### DIFF
--- a/css/map.css
+++ b/css/map.css
@@ -408,6 +408,18 @@ path.stroke.tag-landuse-farmland {
     background-color: rgba(140, 208, 95, 0.2);
 }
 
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks I like it!

screenshot 2016-11-05 19 13 35

PR #3564, opened 2016-11-08
add missing keycode for underscore

tyrasd:
The underscore key is referenced in [zoom.js](https://github.com/openstreetmap/iD/blob/master/modules/ui/zoom.js#L78), but not defined as a keycode. This resulted in a keycode=0 definition [here](https://github.com/openstreetmap/iD/blob/master/modules/lib/d3.keybinding.js#L67-L87), which is potentially triggered by special keyboard keys (for example, on my keyboard the [Fn](https://en.wikipedia.org/wiki/Fn_key) key produces such an event).

### DIFF
--- a/modules/lib/d3.keybinding.js
+++ b/modules/lib/d3.keybinding.js
@@ -169,7 +169,9 @@ d3keybinding.keyCodes = {
     '=': 187, 'equals': 187,
     // Comma, or ,
     ',': 188, comma: 188,
-    'dash': 189, //???
+    // Dash / Underscore key
+    'dash': 189,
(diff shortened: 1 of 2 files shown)


tyrasd:
I've now dropped the definitions for the `_` keys again after realizing that there would have been duplicated definitions for the same key in the zoom-out handler. Should be fine now.

bhousel:
Thanks good catch @tyrasd!   I think my original goal in `zoom.js` was to try to capture every possible combination that these keys can send (note that Firefox sends different keycodes for some of the symbols).  That explains the array of key codes, each shifted and unshifted.
 
keyboard copy

PR #3566, opened 2016-11-08
.travis.yml: validate consistency of generated files

bkil:
If one forgets to commit changes to generated files, the build should fail
and the requested diff will be posted in the log.

Hence one could also simply apply the displayed patch in order to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,13 @@ node_js:
   - "4"
   - "6"
 sudo: false
+
+script:
+  - npm test
+  - if ! git diff --patch --stat --color=never --exit-code; then
+        echo "";
+        echo "#### You can apply this patch via:";
+        echo "git apply -";
+        echo "git commit --all --message 'fix up generated files via `npm run build`'";
+        false;
+    fi >&2


bhousel:
I'm not sure I really want this.  Sometimes new contributors make their edits directly in GitHub and don't run `npm run build`.  I can always see when reviewing their pull requests whether the generated files were updated or not. 

I think for now I'd prefer to have travis error on actual test failures only.

bkil:
How about a new matrix element listed under `allow_failures`?

https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

Those editing over GitHub could at least get feedback about what else they would need to change.

bhousel:
The bigger pain point for me is that our test coverage needs to be much better.

1. We don't test anything under `modules/ui`, which is a significant chunk of iD.  
2. We rely on PhantomJS to test the browsery parts of iD.  This is an old fork of webkit that, now in 2016, is even more outdated than IE11. 

People forgetting to `npm run build` after changing presets isn't an actual problem I need solved right now.

PR #3568, opened 2016-11-08
Preset/field additions: ssid lunch mirror outdoor_seating terms

bkil:
_I could split it even further if you think it is still too big._

Support handling of some common fields:
* `internet_access:ssid`
* `lunch`
* `outdoor_seating`

New preset:
* `highway=traffic_mirror`

Extend search terms:
* `amenity=pub`
* `amenity=hunting_stand`
* `man_made=surveillance`

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -498,6 +498,9 @@ en:
       internet_access/fee:
         # 'internet_access:fee=*'
         label: Internet Access Fee
+      internet_access/ssid:
+        # 'internet_access:ssid=*'
+        label: SSID (Network Name)
       kerb:
         # kerb=*
         label: Curb Ramp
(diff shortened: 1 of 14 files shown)


boothym:
If you adding "alpr" then you need to add "anpr" as it more commonly known: https://en.wikipedia.org/wiki/Automatic_number_plate_recognition

bkil:
@boothym Thanks for the suggestion, I've added some more variants and rebased to master.

bhousel:
Thanks @bkil I merged most of these just now, and added the new internet access and outdoor seating fields to a lot of related presets.  

I still don't think `lunch` is a useful tag, so lets skip that one for now.  It's a bigger lift, but I'd prefer to build a better way for people to tag opening hours.

9dfdf09 Add internet_access, outdoor_seating to presets that might use them
94a546c data/presets/presets/highway/traffic_mirror.json: new
504ed75 data/presets/presets/man_made/surveillance.json: add search terms
a0128e8 data/presets/presets/amenity/pub.json: add search terms
2b44f9e data/presets/presets/amenity/hunting_stand.json: add search terms
ed2f693 data/presets/fields/outdoor_seating.json: new (for pub, restaurant, ...)
53e2fff data/presets/fields/internet_access/ssid.json: new

bkil:
I have clarified what `lunch` means:
https://wiki.openstreetmap.org/wiki/Key:lunch

Thanks.

PR #3572, opened 2016-11-08
implement keyboard shortcuts using KeyboardEvent.key api

tyrasd:
This is a try on improving compatibility across different keyboard layouts (e.g. used by different language groups): Use the new [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) API (if available, but most browsers already have it, only Safari seems to be a bit behind on this – for which old-style keyCodes are used as a fallback).

I've tested it on Firefox (Linux) and Chrome and IE11 (both on Windows) and there it seems to work very well. But there's a strange bug in Chrome on Linux (see this ~~[ticket](https://bugs.chromium.org/p/chromium/issues/detail?id=663339)~~ [ticket](https://bugs.chromium.org/p/chromium/issues/detail?id=633838) (which will apparently be fixed in Chrome v56+) which renders the important undo/redo and zoom shortcuts `Ctrl + z/Z` and `Ctrl + +/-` unusable (well, one can still use them but they are behind different keys there: `Ctrl + y/Y` and ``Ctrl + ß/` ``). Hmm…

### DIFF
--- a/modules/ui/map_in_map.js
+++ b/modules/ui/map_in_map.js
@@ -325,7 +325,7 @@ export function uiMapInMap(context) {
         redraw();
 
         var keybinding = d3keybinding('map-in-map')
-            .on(key, toggle);
+            .on([key, '⇧'+key], toggle);
 
         d3.select(document)
             .call(keybinding);
--- a/modules/lib/d3.keybinding.js
+++ b/modules/lib/d3.keybinding.js
@@ -1,4 +1,5 @@
 import * as d3 from 'd3';
+import _ from 'lodash';
 
 
 /*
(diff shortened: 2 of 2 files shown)


tyrasd:
Maybe we could also use https://github.com/marijnh/w3c-keyname for this which looks to have nice fallbacks for older browsers.

modules/ui/map_in_map.js
```diff
@@ -325,7 +325,7 @@ export function uiMapInMap(context) {
         redraw();
 
         var keybinding = d3keybinding('map-in-map')
-            .on(key, toggle);
+            .on([key, '⇧'+key], toggle);
```
bhousel: I'd rather not bind <kbd>Shift-/</kbd> here, because on US keyboards that's <kbd>?</kbd>, which we're saving for #1481

bhousel:
Hey @tyrasd just checking in with this..  The approach looks good, as long as it's supported in all the main browsers.  While we wait for Chrome 56, is there anything else that we should do in the meantime?

bhousel:
I tested this today in Safari 10.1 on Mac and Chrome 58 on Windows, and both seem fine.  I'm going to merge it 👍

bhousel:
Thanks again for this @tyrasd!

I added a few commits after, to make matching with `shiftKey` less strict.

634002b  Remove shifts from the keybinds that don't need them
94c705e  Don't match `shiftKey` strictly, but prioritize shifted over unshifted

PR #3585, opened 2016-11-15
update docu - use {zoom} instead of {z} for TMS URLs

Klumbumbus:
{zoom} is the default and should be used. See also osmlab/editor-layer-index/issues/226


bhousel:
Thanks @Klumbumbus!

PR #3599, opened 2016-11-19
data/presets: add surveillance and camera related properties

bkil:
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dsurveillance
https://wiki.openstreetmap.org/wiki/Proposed_features/Extended_tags_for_Key:Surveillance

These are mandatory tags:
* `surveillance`
* `surveillance:type`

I do not feel that the following are important, they can be selected from
the drop down manually anyway:
* `operator`
* `name`

These tags are well established:

* `camera:type` has 15k users
* `camera:mount` has 13k users
* `surveillance:zone` has 17k users (mostly traffic is noted)
* `camera:direction` has 6k users

These tags are in the proposed status:
* `contact:webcam` usage is only starting
* `height`

People used various tags instead of `contact:webcam` in the past
because of it being a recent tag, and/or a lack of knowledge or being
mislead by the dropdown.

Thus it is difficult to quote an exact number, but by using complex
regexp queries, I could find thousands of usages for webcam links
when properly combining man_made/surveillance/camera and various
(sometimes invalid) keywords in the key or value part of a tag:

* `website`
* `image`
* `uri`, `url`, `link`, `http`, `contact`
* `stream`
* `webcam`

### DIFF
--- a/data/presets/fields/camera/type.json
+++ b/data/presets/fields/camera/type.json
@@ -1,0 +1,6 @@
+{
+    "key": "camera:type",
+    "type": "combo",
+    "label": "Camera type",
+    "options": ["fixed", "panning", "dome"]
+}
--- a/data/presets/fields/contact/webcam.json
+++ b/data/presets/fields/contact/webcam.json
@@ -1,0 +1,7 @@
+{
+    "key": "contact:webcam",
+    "type": "url",
+    "icon": "website",
+    "label": "Webcam feed URL",
+    "placeholder": "http://example.com/"
+}
--- a/data/presets/fields/surveillance.json
+++ b/data/presets/fields/surveillance.json
@@ -1,0 +1,6 @@
+{
+    "key": "surveillance",
+    "type": "combo",
+    "label": "Surveillance kind",
+    "options": ["outdoor", "public", "indoor"]
+}
--- a/data/presets/fields/surveillance/type.json
+++ b/data/presets/fields/surveillance/type.json
@@ -1,0 +1,6 @@
+{
+    "key": "surveillance:type",
+    "type": "combo",
+    "label": "What is watching",
+    "options": ["camera", "guard", "alpr"]
+}
--- a/data/presets/fields/surveillance/zone.json
+++ b/data/presets/fields/surveillance/zone.json
@@ -1,0 +1,6 @@
+{
+    "key": "surveillance:zone",
+    "type": "combo",
+    "label": "Zone",
+    "options": ["traffic", "building", "town", "parking", "shop", "bank"]
+}
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -206,6 +206,17 @@ en:
       building_area:
         # building=*
         label: Building
+      camera/direction:
+        # 'camera:direction=*'
+        label: Camera direction
+        # camera/direction field placeholder
+        placeholder: 90 (degrees clockwise)
+      camera/mount:
+        # 'camera:mount=*'
+        label: Camera mount
+      camera/type:
+        # 'camera:type=*'
+        label: Camera type
       capacity:
         # capacity=*
         label: Capacity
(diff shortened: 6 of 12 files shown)


data/presets/fields/contact/webcam.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "contact:webcam",
+    "type": "url",
+    "icon": "website",
+    "placeholder": "http://example.com/",
+    "universal": true,
```
bhousel: This should not be a universal field

data/presets/fields/camera/type.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "camera:type",
+    "type": "combo",
+    "label": "Camera type",
+    "default": "fixed",
```
bhousel: We should not default this.. better to let users enter it..

data/presets/fields/camera/features.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "camera:features",
```
bhousel: Let's not add `camera:features`.. it's not an established tag.

data/presets/fields/camera/features.json
```diff
@@ -0,0 +1,6 @@
...
+    "type": "combo",
+    "label": "Camera features",
+    "placeholder": "Webcam, Microphone, Nightvision, Heatvision, Zoom, Motion, Wireless"
```
bhousel: Should not suggest Uppercase placeholders for a lowercase tag

data/presets/fields/surveillance.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "surveillance",
+    "type": "combo",
+    "label": "Surveillance kind",
+    "default": "outdoor",
```
bhousel: Should not default this

data/presets/fields/surveillance.json
```diff
@@ -0,0 +1,8 @@
...
+    "placeholder": "Public, Outdoor, Indoor",
```
bhousel: No need to add placeholder here, this will be computed automatically from available values.. Don't add Uppercase placeholders for a lowercase tag..

data/presets/fields/surveillance/type.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "surveillance:type",
```
bhousel: Let's skip `surveillance:type`.. 98% of the values are just "camera", and the other values are noisy.

data/presets/fields/surveillance/zone.json
```diff
@@ -0,0 +1,6 @@
+{
+    "key": "surveillance:zone",
+    "type": "combo",
+    "label": "Zone",
+    "placeholder": "Town, Parking, Traffic, Shop, Building"
```
bhousel: No placeholder here.. let it be computed from available values.

bkil:
Thank you for the insightful comments. I will probably have some time tomorrow to do the fix-ups.

data/presets/fields/surveillance/type.json
bkil: Although I default to `camera` as well when I map and I don't view the other variants meaningful either, however it has become kind of de facto mandatory.

For example, this visualization also requires many fields to work (described on right panel):
https://kamba4.crux.uberspace.de/

Although, I'll probably add `camera:direction` to the form as well because I find it easy to add in practice.

bhousel: Ok, I see.  Yes I have heard of Surveillance Under Surveillance.  I didn't realize they were consuming this data and required `surveillance:type` to be filled in.  In that case, what you have for this file is fine 👍

bkil:
Requested changes done. Added `camera:direction` as well and ordered options by occurrence.

bhousel:
Thanks @bkil I made a few small cleanups and merged..

BTW, you don't need to include the filename in the commit message - GitHub makes it really easy for me to see what files were changed.   It's better to use those words to include a summary that I can scan quickly when I'm looking over the commits with `git log`.

[How to Write a Git Commit Message](http://chris.beams.io/posts/git-commit/)

PR #3600, opened 2016-11-19
data/presets: taxon, taxon:genus added to help identify plants

bkil:
https://wiki.openstreetmap.org/wiki/Key:taxon

Add taxon to wood, tree*, hedge, forest, scrub, heath
(These were the kind of objects indicated at the definition)

Marsh was skipped because it is a deprecated tag and iD has no
preset for it.

Height is also a common tag for trees.

Display legacy fields.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -431,6 +431,11 @@ en:
       generator/type:
         # 'generator:type=*'
         label: Type
+      genus:
+        # genus=*
+        label: Biological Genus
+        # genus field placeholder
+        placeholder: Scientific Name
       golf_hole:
         # ref=*
         label: Reference
@@ -1011,6 +1016,11 @@ en:
       source:
         # source=*
         label: Source
+      species:
+        # species=*
+        label: Biological Species
+        # species field placeholder
+        placeholder: Scientific Name
       sport:
         # sport=*
         label: Sport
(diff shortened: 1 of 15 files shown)


bhousel:
Not sure if I want to add these fields to the sidebar.  Is it fair to say that only a trained botanist can input data properly into them?

bkil:
Well, if one wants to fill in all the fields they must probably use Wikipedia. According to the OSM wiki, the previous `genus`/`species` tags were invented for more precise categorization, however the new `taxon:*` scheme could theoretically relax the restrictions.

At the most basic level, when fruits are visible in summer, any lay person should be capable of tagging at least `taxon:en` for the most common kinds in a given area like apple, plum, hazelnut, mulberry, cherry, figs or walnut. The other fields could usually, but not always be inferred by this, or a more experienced person could give a second visit to add further detail.

https://taginfo.openstreetmap.org/keys/taxon:en#values

bhousel:
My worry is that people will type wrong things into here and then other people will complain that "iD is encouraging people to put in bad data".  `taxon:en` would be ok for the common names of things, but only works for English, which makes the field not useful to a lot of the world.  

@SK53, you know more about these tags than I do.  Do you think the benefits of making it _easy_ for casual mappers to enter `taxon` and `taxon:genus` outweigh the risks of having bad data entered there?

bhousel:
So I've thought about this more, and I still think the risks of adding these biology fields outweigh the benefits of adding them to the iD sidebar.  

We haven't said so before, but I believe that every field in iD should be something that an **untrained** beginner mapper with local knowledge can understand and contribute to OSM.  I think most of our fields meet that standard.  

Some fields that are a bit iffy:
* Things like `sac_scale`, `piste:difficulty`, `smoothness` are a arbitrary, but at least these fields have a short fixed list of translatable options to choose from. 
* `opening_hours` is just terribly complicated, but we have nothing better.

PR #3607, opened 2016-11-22
improve release algorithm

edpop:
Escape '\*' characters for correct display in Markdown.
Do not add force dist/locales/\*.json files, that are not ignored in master branch.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -29,7 +29,7 @@
 - [ ] git checkout release
 - [ ] git reset --hard master
 - [ ] npm run all
-- [ ] git add -f dist/*.css dist/*.js dist/img/*.svg dist/locales/*.json dist/mapillary-js/
+- [ ] git add -f dist/\*.css dist/\*.js dist/img/\*.svg dist/mapillary-js/
 - [ ] git commit -m 'Check in build'
 - [ ] git tag vA.B.C
 - [ ] git push origin -f release vA.B.C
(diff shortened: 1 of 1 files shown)


bhousel:
Hmm, I kind of like being able to copy and paste these lines into the shell.. Can we instead wrap the whole thing in a `<code>` block or something?

pnorman:
> Hmm, I kind of like being able to copy and paste these lines into the shell..

You should be able to with this change

image

> Can we instead wrap the whole thing in a <code> block or something?

Yes, a different format might be more useful since the checkboxes aren't really useful.

edpop:
Still needs escaping :(

```<code>git add -f dist/*.css dist/*.js dist/img/*.svg dist/mapillary-js/</code>```

### Update and tag release branch
- [ ] <code>git add -f dist/*.css dist/*.js dist/img/*.svg dist/mapillary-js/</code>

Strange behaviour, star character shows before .svg

**Comment by pnorman** [comment]
> Strange behaviour, star character shows before .svg

The first two mark a section of italics.

**Comment by edpop** [comment]
You right.
I found only this solution.

    - [ ] ```git add -f dist/*.css dist/*.js dist/img/*.svg dist/locales/*.json dist/mapillary-js/```

### Update and tag release branch
- [ ] ```git add -f dist/*.css dist/*.js dist/img/*.svg dist/locales/*.json dist/mapillary-js/```

But now the source text is harder to copy.

**Comment by bhousel** [comment]
>Yes, a different format might be more useful since the checkboxes aren't really useful.

Agree, I really don't need the checkboxes.. This file is mostly for my own use to make sure I do all the steps.  It doesn't really need to look pretty.. Even this would be fine:

```bash
$ thing1
$ thing2
$ thing3
$ thing4
```

PR #3621, opened 2016-11-26
add presets to amenity=food_court and amenity=crematorium

samely:
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcrematorium
http://wiki.openstreetmap.org/wiki/Tag:amenity%3Dfood_court


bhousel:
Thanks @samely!

Issue #3603, opened 2016-11-21
Vertex Inspector: Add a connected ways section


slhh:
In case of showing a vertex, the inspector should habe an additional section (like all relations section) listing all connected ways:
When clicking a way there, 
- the way is made the way to stay on in view of vertex navigation, 
- the associated halo is shown, which makes identifying the way easier not only for vertex navigation,
- a toolbar is shown below the way to stay on in the list.

The toolbar offers operations on the specific pair of vertex and way.
Some ideas are:

1. `Jump to way`, which would offer access to coincident ways
2. `Delete vertex from way`
3. `Disconnect`: Same as above, but replace it by a new vertex in the same position
4. `Jump to first/last/next/previous vertex`, as a graphical equivalent of the vertex navigation keys, making the discoverable.
5. `Continue way`, this might remain a duclicate to the radial menu operation, but accessing from this toolbar might be easier in dense environments of the vertex, and especially in case of coincidence.

The impact on the user interface would be small, because it does effect the inspector in the case of a vertex only, where typically very few information has to be displayed. In addition the connected ways section can be shown collapsed by default in most cases, but should be automatically unfolded where coincidence exist, or where unconnected ways are around in an invisible short distance. This can make users aware of the specific situation.

@bhousel If the impact on the UI is still deemed to be too complex let's find a way to make it available for the intermediate users as a task or plugin. I this case I hope you will change your mind in view of complexity when testing the final code.

If it is acceptable, I will try to make my first pull request. I think this is a good issue to start with because I expect to find examples for most required things in the existing code. I might need some help later, but I have already found the relevant sources and I managed to insert a second raw membership editor as an exercise and preparation for this issue.

bhousel:
This seems ok to add something like an "All Parents" section, down near where the relation membership UI would go. 

Being able to see the parents of a vertex and select it from the list (the same way parent relations work) seems simple enough.  You could probably add next/previous/first/last somehow too.  You might be able to make a textual stripmap that shows where the current vertex sits along the way in context with these other vertices - kind of like the digital map we have in some of our subways in nyc:

![nycsubway](https://cloud.githubusercontent.com/assets/38784/20471228/f09d1dd0-af7d-11e6-87b1-e1bd85878e7c.jpg)


I am unsure about adding operations there.  I kind of think adding stuff like "Disconnect", "Delete", "Continue" might be be too much.

PR #3631, opened 2016-12-04
Add a parent ways section to the inspector for vertexes, and improve vertex navigation

slhh:
This adds a section to the inspector which shows the connected ways of a vertex #3603 .
It enables the user to select a specific way in case of coincidence, which makes it a partial solution for #1239. 
It provides an visual equivalent of the keyboard vertex navigation, which can also make the respective shortcuts discoverable.

In addition the vertex keyboard navigation has been improved.

This pull request is currently a first preview version for discussion.

Known issues:
- npm test does still fail on a few tests, but I have the same issue with the current master (on Windows 10).
- The visual indications of the current related parent in the parent ways section of the inspector does not update as long as the selected element is not changed. Update of the sidebar seems to be prohibited when mode select is reentered without changing the selectedIDs. @bhousel How to address this? Update the sidebar more often, or implement some callbacks for updating the related parent? 
- some CSS styling to be done
- preparations for translation to be done

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -1756,14 +1756,36 @@ img.wiki-image {
 }
 
 .raw-parent-way-editor .parent-way-vertex-index {
-    border-radius: 15px 15px 15px 15px;
-    border: 3px solid #cfcfcf;
     width: 15%;
+    padding: 0px;
 }
 
-.raw-parent-way-editor .rp-active {
-    background: rgba(0,0,0,.5);
+.raw-parent-way-editor .button-parent-way-vertex-border {
+    border-radius: 15px 15px 15px 15px;
+    border: 3px solid #cfcfcf;
+    padding: 4px;
+    width: 100%;
+    min-width: 34px;
 }
+
+.raw-parent-way-editor .rp-active .button-parent-way-vertex-border {
+    background: rgba(0,0,0,.5);
+    color:  white;
+}
+
+.raw-parent-way-editor .rp-active .button-parent-way-vertex:hover .button-parent-way-vertex-border {
+    background: rgba(0,0,0,.8);
+}
+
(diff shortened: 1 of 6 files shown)


bhousel:
It's going to be a few days before I can do a thorough review.. To help me out, can you rebase to master and squash away all the merge commits?

to your points:
> npm test does still fail on a few tests, but I have the same issue with the current master (on Windows 10).

Yes I noticed this too 😞  If you have time, it would be awesome if you could look into the failing tests on Windows and make a separate PR for it.  I think it's something to do with PhantomJS and xmlhttprequest and maybe side effects from testing the osm service leaving the xmlhttprequest in a bad state.  _👈 stuff like this is more important than adding new features to iD._

> The visual indications of the current related parent in the parent ways section of the inspector does not update as long as the selected element is not changed. Update of the sidebar seems to be prohibited when mode select is reentered without changing the selectedIDs. @bhousel How to address this? Update the sidebar more often, or implement some callbacks for updating the related parent?

There is no obvious way that I can think of.  You're right that the sidebar wouldn't normally redraw unless a tag is changed or a different feature is selected.  It sounds like this new sidebar component would need to listen to an event dispatched whenever "current related parent" is changed by the user.  Maybe we don't need that feature?  Could it just display all the related parents?  You might be able to indicate which one is current with css.

slhh:
@bhousel I have rebased it. The windows build issue seems to be a too hard one for me due to missing knowledge of such web server things.

I have managed to get the related parent display updated by css.

I have also added support for relations in view of keyboard navigation. This requires autoloading of selected elements which are not already in context. This is included here, but results in failure of one of your recently added tests. ModeSelect does no longer reject invalid ids, but does return to to modeBrowse after receiving the callbacks of loadEntity. I think your test might just wait some time before checking the mode, but I don't know how to do it in the test code.

Issue #3637, opened 2016-12-08
Orthogonal / isogonal angle drawing


edpop:
Hello.

I'll be glad to share some solution for https://github.com/openstreetmap/iD/pull/2699, https://github.com/openstreetmap/iD/issues/1957.
I made a feature in my iD for drawing angles multiple of 45°. It works for _next_ segments of lines, _next_ and _last_ segments of areas.

![default](https://cloud.githubusercontent.com/assets/13479292/20991887/98aff07c-bd2c-11e6-9c55-541e810c4378.png)

If you need It, I'll try to make my code/architecture not so ugly that it is now and send PR then.

bhousel:
Cool, can you put a demo of it online?  I'd love to take a look and try it..

edpop:
I think it is possible, I'll notify.

edpop:
This dirty hack will be able for some time → [pretty-angles](http://api.map.vl.ru/assets/iD/).

bhousel:
Nice, I tried it out and it does feel intuitive to have angles snap when holding down shift..

I did notice that after drawing some ways with this, the graph got corrupted, and I saw lots of stuff in the console like `graph.js:38 Uncaught Error: entity w-51 not found`.  But if you clean it up a bit, I'd be happy to merge it in..

edpop:
I tried to repeat this error in different browsers and places, but my console still clear, except some 404s of mapbox tiles. My cartoghaphers are using this tool for a 5 months, so it should be stable.
Can you tell me, how to reproduce this error, please?

bhousel:
> Can you tell me, how to reproduce this error, please?

I tried some more and I did find the error, but it turns out to be unrelated to your code.  It's in master too.  Sorry about that!  But at least trying out the drawing mode helped me find this other thing.  I'll ticket now separately.

edpop:
Do you have additional wishes to merge it? Maybe you have some thoughts about not to use this function twice?

bhousel:
> Do you have additional wishes to merge it?

I think it's good and I do want to merge it, just working on other things right now.   I'm making substantial changes to how draw_area/draw_way works, so I don't want to start merging the angles code until that other work is finished.

tordans:
@bhousel @edpop did this code/idea make it into the live version?
I would love to see it there. It would make HOTOSM tasks soooo much easier.
If not, is there a change to revive it?

bhousel:
[...]

PR #3641, opened 2016-12-09
Pretty angles (orthogonal / isogonal angles)

edpop:
A new feature for drawing orthogonal / isogonal angles with _shift_ key https://github.com/openstreetmap/iD/issues/3637.

Perhaps you can refactor this to not import **d3** in _draw_way.js_.
May be I should delete comments in _pretty_angles.js_?


bhousel:
> May be I should delete comments in pretty_angles.js?

The code comments are awesome, honestly I want that everywhere in iD.

PR #3642, opened 2016-12-09
Public bath scheme

Natsuyasumi:
A simple public bath scheme, as specified [on OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpublic_bath).


bhousel:
Thanks!  I proper cased the labels (e.g. "Sand Bath" not "Sand bath") and merged.

PR #3643, opened 2016-12-09
Further address customization

Natsuyasumi:
Continuation of PR #3639.

I wanted to create Japanese address scheme, but the generic placeholders doesn't fit with the Japanese tags (especially city, suburb, quarter and neighbourhood). Additionally there are up to 3 numeric address parts, so the `123` housenumber placeholder will mislead users.

The Vietnamese-specific placeholders are imported from (PR #2252)[https://github.com/openstreetmap/iD/pull/2252]


modules/ui/fields/address.js
```diff
@@ -154,16 +149,21 @@ export function uiFieldAddress(field, context) {
...
+              var countryInserter = '';
+              if (addressFormat.customPlaceholders.indexOf(d.id) !== -1) { countryInserter = '!' + countryCode; }
+              return field.t('placeholders.' + d.id + countryInserter); })
             .attr('class', function (d) { return 'addr-' + d.id; })
             .style('width', function (d) { return d.width * 100 + '%'; });
 
         // Update
-        var addrTags = [
+        // setup dropdowns for common address tags
+        if (typeof addressFormat.dropdowns !== 'undefined') { addrTags = addressFormat.dropdowns; }
```
bhousel: Hey @Natsuyasumi I'd like to get this code merged this week, so I'm taking a look through it now..

Is the reason for `addressFormat.dropdowns` so that you can turn `block_number` ("Building No.") and `housenumber` ("Lot No.") into dropdown fields?

Natsuyasumi: I've created the `addressFormat.dropdowns` for simplicty, so that you don't have to add this fields-that-should-have-dropdowns in address,js. The `housenumber` field doesn't have dropdown.

bhousel: @Natsuyasumi Ok, makes sense, thanks!

bhousel:
Just merged this.. thanks @Natsuyasumi!

PR #3654, opened 2016-12-15
Add address field to works preset

abdeldjalil09:


bhousel:
Thank you!  I also added the `product` field and merged.

PR #3655, opened 2016-12-16
Add Taiwan phone format

Supaplextw:


data/phone-formats.json
```diff
@@ -49,6 +49,7 @@
         "ru": "+7 495 1234567",
         "kz": "+7 495 1234567",
         "vn": "+84 1 234 5678",
-        "hk": "+852 12345678"
+        "hk": "+852 12345678",
+        "tw": "+886 2 2720 8889",
```
bhousel: I get a parse error because of this trailing comma, can you remove it?

bhousel:
Is "+886 2 2720 8889" an example number or somebody's actual phone number?  I'd expect an example number to look like "+886 123456789".  This format looks a bit different from [what Wikipedia suggests](https://en.wikipedia.org/wiki/Telephone_numbers_in_Taiwan).

Supaplextw:
Sure, I'll fix it.

bhousel:
Thanks! Merged as:

6ed2dce Use fake phone number for placeholder
8f40627 fix format: remove comma
ec61344 Add Taiwan phone format

Supaplextw:
Wow, I didn't notice it use fake number, I've just pick the phone number from Taipei City Government. Thanks for transformation of fake number.

PR #3664, opened 2016-12-20
Improved Post Paste Behaviors

### LINKED ISSUE #3656: Improve post-paste behaviors
Followup from #3555.. 

When pasting objects, there are several things that we can improve:

- [X] When pasting, the user is dropped into Move mode.  This is great, but iD currently multiselects _all_ pasted objects, and we should really not select child vertices of parent selected ways.  They will get moved along with the parent way.  If you're copy/pasting one building, you should end up with only one pasted building way selected, not all the child vertices.

- [X] Even though Move is the active mode, there are useful keystrokes for other operations.  A user should be able to reflect, orthogonalize, circularize, etc with the keyboard shortcuts while moving.  To accomplish this, I think what we really want to do is expose _behaviors_ for these operations, and bind the behaviors that are appropriate in move move.

- [x] This is minor but I noticed that after pasting, you are not guaranteed to have the mouse pointer located on top of the actual pasted objects.  I have to check that clicks that don't land on any object still complete the move.  (You can still hit `enter` to finish the move).

bhousel:
This exposes behaviors for all of iD's operations so we can support them in more places other than select mode.  Being able to perform operations while moving and rotating shapes (e.g. post-paste) is super useful.

This will close #3656


bhousel:
You can now use many of the operations from Select mode in Move and Rotate modes, and can seamlessly switch between the modes with the `M` and `R` hotkeys.  Also many things that previously did not work with multiselections (like reflect or rotate around centroid) do work now.

I have a few more checks to add before merging, but this is feeling 👌 

paste buildings2

Issue #3211, opened 2016-06-30
iD sometimes creates ways with duplicated nodes (way lenght = 0)


naoliv:
While fixing some geometries I found that sometimes people editing with iD are creating ways with duplicated nodes (ie, the initial and final nodes are the same)

For example, https://www.openstreetmap.org/api/0.6/way/427739762

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<osm version="0.6" generator="CGImap 0.4.3 (6182 thorn-05.openstreetmap.org)" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
 <way id="427739762" visible="true" version="1" changeset="40334708" timestamp="2016-06-28T04:06:07Z" user="AjBelnuovo" uid="1799626">
  <nd ref="1840070954"/>
  <nd ref="1840070954"/>
  <tag k="highway" v="residential"/>
  <tag k="name" v="Rua Hermes Salgado Vasconcelos"/>
 </way>
</osm>
```

Is it possible to avoid this, please?

pnorman:
> While fixing some geometries I found that sometimes people editing with iD are creating ways with duplicated nodes (ie, the initial and final nodes are the same)

What you describe is not an error by itself or necessarily a duplicated node, circular ways must have the same initial and final nodes.

This particular case is an error because it's a two node ways where both nodes are the same.

bhousel:
^  That does give me a clue though.. Maybe it's a circular way that got simplified down to just it's start/end point.  I feel like I saw this recently in another recent issue.

naoliv:
Sure, my description was incomplete, sorry.
What I was trying to say is "iD is creating highways with only 2 nodes, and they (initial and final) are the same".

slhh:
The needle replacement in way.js will be a source of duplicated nodes if it is called for a way containing replacement as one of the neighbors of needle.

    replaceNode: function(needle, replacement) {
        if (this.nodes.indexOf(needle) < 0)
            return this;
        var nodes = this.nodes.slice();
        for (var i = 0; i < nodes.length; i++) {
            if (nodes[i] === needle) {
                nodes[i] = replacement;
            }
        }
        return this.update({nodes: nodes});
    },

I'm not sure if this can really happen. I will make a pull request to prevent this if desired.

slhh:
[...]

Issue #1296, opened 2013-04-11
consecutive identical nodes in way


o5k6:
Please make sure not to create ways containing multiple consecutive references to the same node (e.g. A-B-B-C).

Simple solution: before uploading, iterate through the each way's list of nodes and discard those which are identical to their immediate predecessor. However, such a situation should not even arise in the first place, so something is rotten elsewhere in the code if such a way is found.

Proof that the issue is not hypothetical:

http://www.openstreetmap.org/browse/way/216250602       v1      2013-04-07      iD 0.0.0-beta1
http://www.openstreetmap.org/browse/way/207430096       v1      2013-02-27      iD 0.0.0-alpha2
http://www.openstreetmap.org/browse/way/207323200       v1      2013-02-26      iD 0.0.0-alpha2
http://www.openstreetmap.org/browse/way/181500153       v5      2013-03-11      iD 0.0.0-alpha2
http://www.openstreetmap.org/browse/way/178471693       v2      2013-04-08      iD 0.0.0-beta1
http://www.openstreetmap.org/browse/way/44173536        v8      2013-04-05      iD 0.0.0-beta1

simonpoole:
A recent changeset which created a dup node in way 234752583 https://www.openstreetmap.org/changeset/28292892

stachek:
It is possible to combine any ways that have common first or last nodes. So if these ways have any another common nodes we can create selfoverlapping ways with duplicated nodes, even like this:
![screenshot - 05182015](https://cloud.githubusercontent.com/assets/12500479/7686442/75852602-fd9e-11e4-8f0a-99ec38ca0cd4.png)

drkludge:
In the changeset, https://www.openstreetmap.org/changeset/28292892, that simonpoole reported
 - iD 1.6.2 was used.
 - The way with the duplicate node is https://www.openstreetmap.org/way/234752583/history.
 - Node https://www.openstreetmap.org/node/1650627794/history in the northeast corner of the closed way and is the terminating node ( real duplicate for an area ).
 - The termination node was not modified in the change.
 - The duplicate node is in the southeast corner https://www.openstreetmap.org/node/1650627780/history.
 - The duplicate node appears to be the last node in the way before the way closes.
 - The way is using a counter clockwise rotation.
 - Both the coordinates in the duplicate node were modified.
 - The JOSM history diff says that the node was moved by 8.88 m.
[...]

Issue #1369, opened 2013-04-24
DrawWay is churning node and way IDs


jfirebaugh:
For every vertex added, DrawWay uses three node IDs and one way ID. Each of the used IDs creates an (id, undefined) entry in subsequent graphs.

PR #3676, opened 2016-12-24
Tests and fixes of potential duplicate consecutive node issues in way.js

slhh:
I have added some tests for replaceNode, updateNode, addNode, and removeNode. These are partially failing with the original code, due to some potential issues in view of duplicate nodes.

Reworked code passing these tests is in the second commit.

In addition I have added some code comments.

Related issues are #1296 and #3211.

### DIFF
--- a/modules/osm/way.js
+++ b/modules/osm/way.js
@@ -188,17 +188,16 @@ _.extend(osmWay.prototype, {
     },
 
     // Adds a node (id) in front of the node which is currently at position index.
-    // If index is negative, it will be counted from the end of the way.
-    // If index is 0 or < -length, the node (id) will be added at the start of the way.
-    // If index is undefined or >= length, the node (id) will be added at the end of the way.
+    // If index equals length, the node (id) will be added at the end of the way.
+    // If index is negative or > length, it will throw an error.
     // Generating consecutive duplicates is silently prevented
     
     addNode: function(id, index) {
         var nodes = this.nodes.slice(),
             spliceIndex = index === undefined ? nodes.length : index;
-        if (spliceIndex > nodes.length) spliceIndex = nodes.length;
-        if (spliceIndex < 0) spliceIndex = nodes.length + index;
-        if (spliceIndex < 0) spliceIndex = 0;
+        if (spliceIndex > nodes.length || spliceIndex < 0) {
+            throw new Error('addNode: index ' + spliceIndex + ' is invalid');
+        }
         if (nodes[spliceIndex] !== id&& nodes[spliceIndex-1] !== id) {
             nodes.splice(spliceIndex, 0, id);
         }
@@ -206,14 +205,15 @@ _.extend(osmWay.prototype, {
     },
 
     // Replaces the node which is currently at position index with the given node (id). 
-    // If index is negative, it will be counted from the end of the way.
-    
+    // If index is negative or >= length, it will throw an error.   
     // Consecutive duplicates are eliminated including existing ones.
 
     updateNode: function(id, index) {
         var nodes = [];
         
-        if (index < 0) index = this.nodes.length + index;
+        if (index === undefined || index >= this.nodes.length || index < 0) {
+            throw new Error('updateNode: index ' + index + ' is invalid');
+        }
 
         for (var i = 0; i < this.nodes.length; i++) {
             var node = this.nodes[i];
--- a/modules/osm/way.js
+++ b/modules/osm/way.js
@@ -187,34 +187,70 @@ _.extend(osmWay.prototype, {
         });
     },
 
-
+    // Adds a node (id) in front of the node which is currently at position index.
+    // If index is negative, it will be counted from the end of the way.
+    // If index is 0 or < -length, the node (id) will be added at the start of the way.
+    // If index is undefined or >= length, the node (id) will be added at the end of the way.
+    // Generating consecutive duplicates is silently prevented
+    
     addNode: function(id, index) {
-        var nodes = this.nodes.slice();
-        nodes.splice(index === undefined ? nodes.length : index, 0, id);
-        return this.update({nodes: nodes});
-    },
-
-
-    updateNode: function(id, index) {
-        var nodes = this.nodes.slice();
-        nodes.splice(index, 1, id);
-        return this.update({nodes: nodes});
-    },
-
-
-    replaceNode: function(needle, replacement) {
-        if (this.nodes.indexOf(needle) < 0)
-            return this;
-
-        var nodes = this.nodes.slice();
-        for (var i = 0; i < nodes.length; i++) {
-            if (nodes[i] === needle) {
-                nodes[i] = replacement;
-            }
+        var nodes = this.nodes.slice(),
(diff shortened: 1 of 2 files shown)


bhousel:
Hey @slhh sorry for taking so long to look at this.  I think it's a pretty good improvement.  I agree that range checking the index argument and removing the duplicates is worth doing.

One thing that worries me is that there may be places elsewhere in the code where we are calling `addNode` and `updateNode` expecting them to just use `Array.splice`, so any change in behavior from what `Array.splice` does might affect something else.  

I normally hate exceptions, but I think in this case I'd almost rather throw an out of range exception when trying to call these functions with a weird index value.  Rather than silently succeed by adding the node to the beginning or end or something like that.  That way we can track down places in the code where the assumptions fall short.  

* with exceptions:  "I tried to circularize a way and it made an error and iD stopped working" 
* silent failures: "I tried to circularize a way and the corner got a bit mangled but I just continued editing and told nobody."

I guess it's a judgement call, but I think the case for exceptions is better for us for debugging and better for OSM not getting bad data put into it.  What do you think?

Again, overall this looks totally fine.. My worry is about other code in iD that assumes the methods internally use `Array.splice` or that contains subtle bugs.

slhh:
> One thing that worries me is that there may be places elsewhere in the code where we are calling addNode and updateNode expecting them to just use Array.splice, so any change in behavior from what Array.splice does might affect something else.

This worried me too, therefore, I tried to reproduce the splice behavior. All the index checks are intended to do this, and don't limit the useable index range.
Of course one difference remains which is the behavior in case of duplicates. A calling function expecting the updated node at a certain position or a certain length can fail. I think this pull request should not be merged just before a release.

I an earlier version I haven't supported the negative index range. The one direct test with negative index failed, but no additional higher level tests failed. And a little manually testing hasn't shown an issue.
Therefore, I think the negative index range is not used, but I can't be absolutely sure.

In case we want to limit the useable index range, which makes sense in my opinion, an exception seems to be useful. Can we throw it in case of debugging only? An exception in the field is only useful in case we want to address the resulting issue with high priority. We should not impair user experience just to fill the bug tracker without working on it.

Provided we prevent duplicate nodes in incoming data from the database, we may also throw an exception in case the functions have to remove duplicates. We do only need to compare the resulting lenght with the expected length. I'm not sure if we want to let calling code rely on way.js for duplicate node prevention or want to have all required rules for prevention in the calling code. In the later case an exception can be useful to find the bug.

bhousel:
> In case we want to limit the useable index range, which makes sense in my opinion, an exception seems to be useful. Can we throw it in case of debugging only? An exception in the field is only useful in case we want to address the resulting issue with high priority. We should not impair user experience just to fill the bug tracker without working on it.

The other notable place we throw an exception in iD is in [`graph.entity`](https://github.com/openstreetmap/iD/blob/master/modules/core/graph.js#L38), when calling code requests an entity which is no longer in the graph.  This is always a programming error, and we don't want the user to continue editing with the graph in a bad state.

So yes, anytime a user reports something like that, it's very high priority to get it fixed.

> Provided we prevent duplicate nodes in incoming data from the database, we may also throw an exception in case the functions have to remove duplicates.

I don't think we want an exception when removing duplicate consecutive nodes because I'm sure there are some cases of this in existing OSM data, which is pretty messy from imports, old bugs, even people just messing around with the api.  For a little while there were even some ways referencing nodes that had been deleted (there is no referential integrity in the OSM database).

slhh:
@bhousel I have added a commit changing it to throw exceptions. Feel free to use it as you like.
The Travis build failed, but maybe you know why. It seems to have similar problems as the Windows build is used to have even with current master.

> with exceptions: "I tried to circularize a way and it made an error and iD stopped working"

The best would be if iD were catching the exception, reporting the error, and resuming from the end of the last successful action. I don't know how complex this would be, at least my current knowledge is too limited to do it.

> I don't think we want an exception when removing duplicate consecutive nodes

I would agree, but for the reason of making actions easier due to relying on way.js for duplicate prevention.

> because I'm sure there are some cases of this in existing OSM data, 

Of course they are, but that's not a reason, we just had to add some code eliminating the duplicate e.g. in the OSM service. That's why I wrote: "Provided we prevent duplicate nodes in incoming data from the database"

>which is pretty messy from imports, old bugs, even people just messing around with the api.

You have missed iD's still existing bugs here. I have watched iD generating a duplicate when testing for my other pull request, which is nicely monitoring such duplicates.
It was just a somewhat failed click on an connecting node of three starting or ending highway, just the first click after starting iD. I recognized a little movement of the node and the undo stack shows something like way connected to line (in German). Selecting the node the new parent ways sections showed that the node has index 1 and 2 of the same one of the three ways. It wasn't a duplicate from the database and I haven't been able to reproduce the bug.

I don't think the problem was due to my code modification. There were only the modification of the sidebar, but not the changes to mode select at that stage. And the sidebar haven't been opened when the bug occurs.

bhousel:
> You have missed iD's still existing bugs here. I have watched iD generating a duplicate when testing for my other pull request, which is nicely monitoring such duplicates.
> It was just a somewhat failed click on an connecting node of three starting or ending highway, just the first click after starting iD. I recognized a little movement of the node and the undo stack shows something like way connected to line (in German). Selecting the node the new parent ways sections showed that the node has index 1 and 2 of the same one of the three ways. It wasn't a duplicate from the database and I haven't been able to reproduce the bug.

Good find though, can you open a separate issue for this?  Maybe it's possible to double click around the endpoint of a way and have iD try to add a midpoint there that ends up merging into the existing node.

bhousel:
Just a quick update, I'm adding a lot of tests (like _hundreds_) and changing slightly the logic to handle different kinds of duplicates in different places along the nodelist, possibly involving connecting nodes on circular ways.  Once we throw those into the mix, things get kind of complicated.  It might take another day or two to merge fully, but this work should make these functions a lot safer in the end.

slhh:
> Good find though, can you open a separate issue for this?

@bhousel 
I hope it it is fixed by this pull request. Do you really want a new issue? Maybe I can add the information to #1296 instead.
> Maybe it's possible to double click around the endpoint of a way and have iD try to add a midpoint there that ends up merging into the existing node.

That was my first idea too, but I haven't been able to reproduce it. Maybe its needs to be too fast to be reproduced intentionally.

bhousel:
> I hope it it is fixed by this pull request. Do you really want a new issue? Maybe I can add the information to #1296 instead.

Yes, reading #1296 again, it sounds like you found one of the pathways to cause it.  You can add the info there and it hopefully will be fixed by this PR.

I wanted to merge this yesterday, but I realized that our area drawing mode [does intentionally splice the node array using index `-1`](https://github.com/openstreetmap/iD/blob/master/modules/modes/draw_area.js#L18) which breaks the assumption that `addNode` is always called with a positive range.  This is related to #1369, so I'm going to attempt to fix that also by having the area drawing code add the node to an existing temp way where it belongs rather than recreating it each time.

bhousel:
Thanks @slhh for kicking this off.. I just merged the updated version!

Issue #3682, opened 2016-12-27
Midpoints adding by double click


edpop:
Hello.
Don't you think, that it would be awesome to add a vertex by double-clicking to midpoint, instead of zooming? I have a few lines of code to this feature and can send PR.
I'm afraid to overload you, imho [orthogonal angles](https://github.com/openstreetmap/iD/pull/3641) is more important.

bhousel:
hmm, I thought this did work, but I realize now that double clicking on the midpoint triangle itself does still count as a zoom intent.  Double-clicking anywhere else on a line adds a point.

Yes send the PR, if it's just a few lines, I can merge it quickly. 👍

PR #3687, opened 2016-12-28
Make vertex by midpoint double-clicking

edpop:
For discussion https://github.com/openstreetmap/iD/issues/3682.
Double-click to a midpoint makes a vertex instead of zooming.


bhousel:
Thanks again, just merged!

PR #3688, opened 2016-12-28
Fix reflect-typo

edpop:
Fix errors in console after reflect-operation on single object.

> Missing en translation: operations.reflect.annotation.short.single
> Missing en translation: operations.reflect.annotation.long.single


bhousel:
Thanks!

Issue #2414, opened 2014-10-21
Display left and right names of administrative boundaries on the map


1ec5:
When you select an administrative boundary in iD, it’s often hard to tell which side is which. Sometimes you can zoom way out, but often the administrative area is far too large to see completely. It would help if iD displayed both the left and right names of the boundary on the map itself, just like the openstreetmap-carto style does.

Nakaner:
This might have a positive influence on iD-caused boundary damages.

matkoniecz:
do we have any other feature which changes rendering when selected? (beyond generic selection marker)

k-yle:
yes, nodes with a diameter/radius tag, if #9732 goes ahead

PR #3707, opened 2017-01-01
fix persian characters on chrome and safari

### LINKED ISSUE #3491: RTL text along SVG textPaths is wrong in Chrome
Just re-ticketing this comment from @mapmeld so we don't lose track of it:
https://github.com/openstreetmap/iD/pull/3087#issuecomment-252453324

> Thanks! As for the question of Arabic labels on ways @marrouchi , it's a bug with how Chrome displays text along a path using SVG textPath; it works in Firefox. So there isn't much iD can do (I tried adding lang, dir, other attributes and it wouldn't budge): 
> http://stackoverflow.com/questions/31362171/how-to-use-svg-textpath-in-persianrtl-text
> https://bugs.chromium.org/p/chromium/issues/detail?id=379149

Thanks @mapmeld for tracking this down.. We can keep an eye on the issue and close once it's fixed upstream.

miladkdz:
To achieve this purpose, we created a mapping for characters in fix-string.js.

In order to display the names correctly, a temporary tag is created named real_name which is terminated while putting changesets to the server.

Closes #3491

### DIFF
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -7,8 +7,6 @@ import { geoExtent } from '../geo/index';
 import { osmEntity, osmNode, osmRelation, osmWay } from '../osm/index';
 import { utilDetect } from '../util/detect';
 import { utilRebind } from '../util/rebind';
-import { fixArabicScriptTextForSvg } from '../util/svg_paths_arabic_fix';
-
 
 var dispatch = d3.dispatch('authLoading', 'authDone', 'change', 'loading', 'loaded'),
     useHttps = window.location.protocol === 'https:',
@@ -71,12 +69,7 @@ function getTags(obj) {
         var attrs = elems[i].attributes;
         tags[attrs.k.value] = attrs.v.value;
     }
-    var isFirefox = utilDetect().browser.toLowerCase().indexOf('firefox') > -1
-    var arabicRegex = /[\u0600-\u06FF]/g
-    if(tags.name && tags.highway && !isFirefox && arabicRegex.test(tags.name)){        
-        tags.real_name = tags.name;
-        tags.name = fixArabicScriptTextForSvg(tags.real_name);            
-    }
+    
     return tags;
 }
 
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -7,7 +7,7 @@ import { geoExtent } from '../geo/index';
 import { osmEntity, osmNode, osmRelation, osmWay } from '../osm/index';
 import { utilDetect } from '../util/detect';
 import { utilRebind } from '../util/rebind';
-import { fixTextForSvg } from '../core/fix-string';
+import { fixArabicScriptTextForSvg } from '../util/svg_paths_arabic_fix';
 
 
 var dispatch = d3.dispatch('authLoading', 'authDone', 'change', 'loading', 'loaded'),
@@ -71,10 +71,11 @@ function getTags(obj) {
         var attrs = elems[i].attributes;
         tags[attrs.k.value] = attrs.v.value;
     }
-    var isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') > -1
-    if(tags.name && tags.highway && !isFirefox){
+    var isFirefox = utilDetect().browser.toLowerCase().indexOf('firefox') > -1
+    var arabicRegex = /[\u0600-\u06FF]/g
+    if(tags.name && tags.highway && !isFirefox && arabicRegex.test(tags.name)){        
         tags.real_name = tags.name;
-        tags.name = fixTextForSvg(tags.real_name);            
+        tags.name = fixArabicScriptTextForSvg(tags.real_name);            
     }
     return tags;
 }
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -349,7 +349,7 @@ export default {
     putChangeset: function(changes, version, comment, imageryUsed, callback) {
         if(changes.modified && changes.modified.length > 0){
             for(var i = 0, l = changes.modified.length; i < l; i++){
-                if(changes.modified[i].tags.highway && changes.modified[i].tags.real_name){
+                if((changes.modified[i].tags.highway || changes.modified[i].tags.railway) && changes.modified[i].tags.real_name){
                     changes.modified[i].tags.name = changes.modified[i].tags.real_name; 
                     delete changes.modified[i].tags.real_name;
                 }
@@ -357,7 +357,7 @@ export default {
         }
         if(changes.created && changes.created.length > 0){
             for(var i = 0, l = changes.created.length; i < l; i++){
-                if(changes.created[i].tags.highway && changes.created[i].tags.real_name){
+                if((changes.modified[i].tags.highway || changes.modified[i].tags.railway) && changes.created[i].tags.real_name){
                     changes.created[i].tags.name = changes.created[i].tags.real_name; 
                     delete changes.created[i].tags.real_name;
                 }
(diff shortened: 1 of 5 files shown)


pnorman:
There's no comments, text in this PR, or issue links which describe what the current problem is that this is supposed to fix.

bhousel:
This is for #3491 

@miladkdz, Thank you!!   I took a quick look and have a few questions...

* Is there a reason that the code only looks for highway name tags?
* I'm not opposed to including the replacement characters in fix-string directly into iD, but is there any existing JavaScript library that we could import that does this for us? 
* You won't want to create a tag called `real_name` for fear of collisions.  Really it's best to assume that OSM can use "any tags they like".  We can work around this by adding a transient graph property for the "display tags" to the `osmEntity` class (like how we [compute and cache object extents](https://github.com/openstreetmap/iD/blob/84537fa1c8fd7c6b3f0713d1e712a635ae63321d/modules/osm/way.js#L47)), I think it probably makes sense to just transliterate (is that the right word?) all the tags, since that's a small dataset and they all get displayed in the sidebar.
* should use `utilDetect()` instead of inspecting [`navigator.UserAgent` directly](https://github.com/openstreetmap/iD/pull/3707/files#diff-84c6ddaf5cf5e08471e7537f8c7acbf6R74) (minor nitpick)
* It probably belongs in the `util` module and should have a better name than `fix-string` (minor nitpick)

@mapmeld can you take a quick pass at this and let me know if it seems like a good approach?  I'm totally happy to finish up the changes to the code listed above.

mapmeld:
1. We should apply this to any line because on points and polygons, the right-to-left text is already displaying correctly. The only issue is when Webkit-based browsers are placing text along an SVG textPath.
2. Renaming ```fix-string``` and ```fixTextForSvg``` would go a long way to making it clear
3. There should be a regular expression to run this function only when it contains characters in Arabic Unicode block... otherwise we are using it on every character in every label
4. There should be a test for how this works with bi-directional text or numbers (equivalent of: '38th Street')

miladkdz:
Hi @bhousel,

- The reason I checked only for highway tags is that the SVG rendering bug happens only on <path> elements. Other elements were fine. If there are any other tags that render as <path>, we should look for them too.
- I didn't find any libraries that would help us here. Of course there are many more characters that we should look for but I settled for Persian standard characters only as my Arabic knowledge is very limited and I have no Idea where to look. But I believe that there are not more than 20 characters that should be added to the file. Maybe an Arab speaking fellow can help us here?
- I will look more into this. I didn't like the approach either but it was fast.
- Sure. Will replace.
- I will move it and apply the suggestion from @mapmeld about naming the file.

Many thanks for your complete review.

miladkdz:
@bhousel 
I pushed another commit fixing most of the things you listed. Though the approach for replacing tags(real_name) has not changed yet. I will do it soon. In the meanwhile, I would appreciate if you guide me a little bit more. Maybe a document or another issue that used the approach?

mapmeld:
Is it possible to change the flip to happen in the code where the textPath is being created, rather than changing the name of the object?

Here are some issues which I saw in Tunisia (looking for Arabic language and bidi compatibility)

- (RN 3 طو) at coordinates 8.49/36.53602/9.99751 -- Arabic letters remain LTR
- (Avenue Habib Bourguiba شارع الحبيب بورقيبة) at 18.42/36.79970/10.18295 -- Arabic and Latin letters are flipped, including the 'name' shown in the editor is "ةﺒيﻗﺭﻮﺑ ﺐيﺒﺤﻟا ﻉﺭﺎﺷ abiugruoB bibaH eunevA" -- "iin" shows both "ii" and "n" in word-ending forms
- I am not able to edit the name property
- These two should be written 'شارع الحرية' (iiyah at end) and 'نهج الباكستان' (odd k)

screen shot 2017-01-03 at 5 29 36 pm

miladkdz:
Fixed kaf.

miladkdz:
The only problem now is when the name parameter has both persian/arabic and other latin characters.

mapmeld:
I think I have a PR for @miladkdz which fixes latin + Arabic character issues.

Main concerns for me at this point would be:
- verifying that we have all Arabic letters and their forms right (especially since the LTR/RTL system will have trouble with missing letters)
- including short vowel signs which go above or below the letter (a i and u)

modules/services/osm.js
```diff
@@ -347,6 +347,30 @@ export default {
...
+            }
+        }
+        if(changes.deleted && changes.deleted.length > 0){
+            for(var i = 0, l = changes.deleted.length; i < l; i++){
+                if((changes.modified[i].tags.highway || changes.modified[i].tags.railway) && changes.deleted[i].tags.real_name){
+                    changes.deleted[i].tags.name = changes.deleted[i].tags.real_name; 
+                    delete changes.deleted[i].tags.real_name;
+                }
+            }
+        }
```
mapmeld: do we still need this change? I don't think so

mapmeld:
@pnorman @bhousel here's my write-up (mostly for non-OSM / non-i18n people) on *what this problem is* and *what this solution is*

https://gist.github.com/mapmeld/556b09ddec07a2044c76e1ef45f01c60

jfirebaugh:
@mapmeld Awesome writeup, thank you! @miladkdz Thanks for the PR!

I've been lurking here but I want to jump in and note the overlap in this problem to [one we've been working on over in Mapbox GL](https://github.com/mapbox/mapbox-gl/issues/4), and cc @chrisloer, who has been working on it over there. In particular, [the current proposal to improve Arabic text rendering in Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js/pull/3758) uses the same approach here, of replacing normal Arabic code points with the appropriate presentation forms. It does so using [ICU](http://site.icu-project.org/), which has a robust algorithm for this, [compiled to JavaScript via emscripten](https://github.com/mapbox/mapbox-icu-js).

The drawback to this approach is that the compiling the ICU C and C++ code to JavaScript has substantial size overhead. Maybe we could combine forces here and port [the relevant ICU code](http://source.icu-project.org/repos/icu/trunk/icu4c/source/common/ushape.cpp) to native and idiomatic JavaScript in a standalone library. That would address the issue of comprehensiveness that @mapmeld brings up, and the size issue that we're facing in GL JS.

modules/services/osm.js
miladkdz: Nope. I will remove them.

miladkdz:
@mapmeld Thanks a lot for your quick fix. 👍

bhousel:
Thanks @mapmeld for the writeup and @miladkdz for the fixes too!  

In order to move this forward I'll try to merge the code to master later today.  Then people try it out in their browser to see if there are any follow-on tasks per https://github.com/openstreetmap/iD/pull/3707#issuecomment-270128922.  Sound good?

miladkdz:
@bhousel Perfect.
@mapmeld Thanks for your efforts on this :-)

ChrisLoer:
@mapmeld @miladkdz This is awesome! I don't have a lot to add but here are some of the things ICU gives us that may or may not be necessary for iD:

* Full support for "Tashkeel", aka Arabic Diacritics. This adds a lot of extra characters and is probably something you can get away without... but they do show up. "Yemen" for instance is (often?) written "اليَمَن"
* ICU BiDi supports explicit order markers which are present in some OSM labels
* ICU BiDi supports line breaking bidirectional text, which is pretty complicated and probably not necessary for this specific use case

bhousel:
Thanks again, I just merged this!   I've tested on Chrome and I do see the difference.  Haven't tested with Safari / Edge / IE11 (will leave it up to those fluent in Arabic to test).

We have a mirror of the current master code here that can be used to test the Arabic linestring labels:  https://openstreetmap.us/iD/master
Please open new follow-on issues if anything looks off..

I did refactor this into 2 functions: `utilDisplayName` and `utilDisplayNameForPath` - because `utilDisplayName` gets called many places, and for most of them we don't want the reversal code. 

This also allowed me to remove the "highway" and "railway" conditionals, because I think the purpose of these was to perform the reversals only on the linestrings.  We also label waterways and aeroways with this code, and some highway and railway features may be point or vertex labeled, and we might eventually label other linestrings in the future (like boundaries #2414) so it's better to not hardcode these tags.

PR #3708, opened 2017-01-01
Add fire_hydrant:position field to presets

wopfel:
http://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_hydrant


bhousel:
thank you!

PR #3712, opened 2017-01-02
Japanese address scheme

Natsuyasumi:
Just a temporary Japanese address scheme. Still country-specific placeholders should be defined, look at #3643.


bhousel:
Thanks, it looks ok, as far as I can tell.. 

I did notice something strange going on with address field combo boxes - it seems like d3 isn't always removing them.  I'll dig into this today and try to find out what's happening.  (It's not an issue with the code you added. I think it's an existing bug in the address field we didn't see before).

screenshot 2017-01-02 14 03 13

screenshot 2017-01-02 14 03 23

bhousel:
The weird field formatting has something to do with Helvetica Neue's handling of kanji glyphs.  I don't understand the root cause, but I'm following some of the guidance [here](https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/) to use modern system fonts, which are all better than Helvetica Neue these days.

Most people probably will not notice the difference, but the new fontstack looks a little nicer..

bhousel:
This was merged

bhousel:
Whoops, I didn't see that you added postcode in fde26f8

I added it in aba1ab1

PR #3714, opened 2017-01-03
Adjust coworking preset

iandees:
Move the `amenity=coworking_space` preset to a more sensible `office=coworking`.

From the Department Of Not Everything's An Amenity.


bhousel:
SGTM, thanks @iandees 👍

abdeldjalil09:
office=coworking is undocumented tag

pnorman:
> office=coworking is undocumented tag

With 34 uses vs the 500ish for amenity=coworking_space

talllguy:
This doesn't qualify as documentation?

tuxayo:
I agree that `office=coworking` is better and that not everything should be an amenity.

But why is it a good idea to make iD switch the preset from amenity to office when the wiki have not?

Discussion about changing the tag on the wiki: https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dcoworking_space

bhousel:
> But why is it a good idea to make iD switch the preset from amenity to office when the wiki have not?

The last time I looked, the wiki had been changed over and people were using the new tag, but I guess someone changed the wiki back?

To be honest, if a tag has only like 500 uses, I don't care too much.  Anyone who knows overpass and JOSM and has an hour of free time can just update the tag to whatever they want, right?  It seems like everyone agrees `office=coworking` is better, so.. why not just change it?

polarbearing:
> The last time I looked, the wiki had been changed over and people were using the new tag, but I guess someone changed the wiki back?

Not sure when you looked. However when this issue started here, somebody had _moved_ the wiki page form amenity= to office= which is against the policy that the wiki documents how a tag is used, not vice versa. Thus the _move_ was reverted after discussion on the tagging list, and the office page currently redirects to the amenity. Please note that there might be coworking spaces currently tagged that are not offices. Of course the page office=coworking can be filled with documentation of this tag. 

> To be honest, if a tag has only like 500 uses, I don't care too much. Anyone who knows overpass and JOSM and has an hour of free time can just update the tag to whatever they want, right? It seems like everyone agrees office=coworking is better, so.. why not just change it?

You are proposing an [Automated Edit](https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct) which is only allowed in special cases, [read here why](https://wiki.openstreetmap.org/wiki/What's_the_problem_with_mechanical_edits%3F). Migrating a tag is not such a case, thus please don't encourage a mechanical edit here. As said above, there might be places that are not offices, thus each place needs to be reviewed with local knowledge.

PR #3717, opened 2017-01-04
fix a bug with space

miladkdz:
Along with the recent changes in #3707, found a bug with various cases when the string contains a space. So wrote a quick fix for it.


bhousel:
I see the additions of `|| nextCode === 32`

Should we maybe do `/\s/.test(String.fromCharCode(nextCode))` to catch any kind of whitespace (tabs, spaces, etc)?

miladkdz:
Are we supposed to support Tabs? Regex test might be much more expensive here.

bhousel:
> Are we supposed to support Tabs? Regex test might be much more expensive here.

I'm not sure whether it matters, so I'll merge it as-is and we can reticket later if there are issues with other whitespace characters.. Thanks!

PR #3727, opened 2017-01-07
Add amenity=coworking_space as non-searchable preset

M1dgard:
In 23a37183762adb2b1fe044b9bee3b50d5095eff3 `amenity=coworking_space` was changed into `office=coworking`.

I'd propose to add a **non-searchable preset** for `amenity=coworking_space` so it doesn't show up as a blank Amenity in the Inspector.


bhousel:
Good idea, thank you!

PR #3730, opened 2017-01-08
Add outdoor_seating options

mertemin:
This PR adds `outdoor_seating` options following [OSM wiki](http://wiki.openstreetmap.org/wiki/Key:outdoor_seating).

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -817,11 +817,6 @@ en:
       outdoor_seating:
         # outdoor_seating=*
         label: Outdoor Seating
-        options:
-          # outdoor_seating=yes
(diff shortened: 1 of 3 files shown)


bhousel:
> This PR adds outdoor_seating options following OSM wiki.

[Looking at taginfo](https://taginfo.openstreetmap.org/keys/?key=outdoor_seating#values), I agree that it seems like `yes`/`no` are the only meaningful values for `outdoor_seating` and everything else is noise.  Can we just make this field a checkbox?

mertemin:
Good call @bhousel. I updated the PR.

PR #3734, opened 2017-01-08
Highways under construction

M1dgard:
💡 **Motivation**: Someone changed a cycleway under construction to a finished cycleway, presumably because iD showed `Highway` where it should have shown `Cycle Path Under Construction`.

👎 **Cons**: This will generate redundant work for the translators, but there may not be a cleaner way to handle all language's quirks. If you can think of a better way, feel free to improve.

➕➖ Because the diff is a bit unwieldy, here's a **change summary**:
* Add a preset file for each road type under construction, in a new subdir `.../highway/construction`. Service road types are under `.../highway/construction/service`.
* Create `.../highway/_construction.json`, for `highway=construction` without a recognized `construction=*`.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -2421,6 +2421,124 @@ en:
         # highway=bus_stop
         name: Bus Stop
         terms: '<translate with synonyms or related terms for ''Bus Stop'', separated by commas>'
+      highway/construction:
+        # highway=construction
+        name: Highway Under Construction
+      highway/construction/bridleway:
+        # 'highway=construction, construction=bridleway'
+        name: Bridle Path Under Construction
+        # 'terms: bridleway,equestrian,horse,construction'
+        terms: '<translate with synonyms or related terms for ''Bridle Path Under Construction'', separated by commas>'
+      highway/construction/cycleway:
+        # 'highway=construction, construction=cycleway'
+        name: Cycle Path Under Construction
+        # 'terms: bike,construction'
+        terms: '<translate with synonyms or related terms for ''Cycle Path Under Construction'', separated by commas>'
+      highway/construction/footway:
+        # 'highway=construction, construction=footway'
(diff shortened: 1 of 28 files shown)


afmenez:
A better solution would be to map the "construction" tag to a checkbox.  For example:

highway=trunk -> highway=trunk + construction=unchecked
highway=construction;construction=trunk -> highway=trunk + construction=checked

bhousel:
Thanks, @M1dgard! A while ago I made presets for the construction versions of the highways, see #2347.  It does make a lot of extra work for the translators, and it also adds a lot of options to the preset list, and it's not flexible enough to handle other tags like "proposed" or "disused", so that's why we decided against it at the time.  So, thank you for the PR but unfortunately I'm going to close it..

There is some good info on #2347 about why this is a tricky problem, and part of it is that there are several competing ways to tag lifecycle stuff in OSM.   #2341 covers adding a special "status" field to sort out all the various tags and make it easy for people to map this.  It's a bigger lift than just using presets to work around the problem, but it will be a better user experience for users in the end.

@afmenez, a `construction=yes/no` checkbox won't catch all cases, and it currently the less preferred way of the 3 kinds of lifecycle tagging (because `construction=no` is a trolltag, and it leaves `highway=trunk` there, which renderers, routers, consumers tend to treat like a normal trunk road). Still, some people do tag this way, and whatever field we build would still need to recognize it.

afmenez:
@bhousel, I believe I was not clear explaining my proposal.  I don't want a construction=yes/no tag, I proposed a construction checkbox that would be translated to the current OSM tags.
Example:
On OSM: highway=trunk -> On ID : highway=trunk + construction=unchecked
On OSM: highway=construction; construction=trunk -> On ID: highway=trunk + construction=checked

I believe there are other instances where the ID fields don´t translate directly to OSM tags.

PR #3735, opened 2017-01-08
Add universal email field, make fax universal

### LINKED ISSUE #3710: Add "fax" and "email" to "add field".

M1dgard:
Fixes #3710.

I don't know if it's desirable to have fax as a universal field. Perhaps it'd be better to separately add it to more features? In that case this PR has to be modified. Otherwise: ready to merge!

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -418,6 +418,11 @@ en:
       elevation:
         # ele=*
         label: Elevation
+      email:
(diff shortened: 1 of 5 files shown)


bhousel:
Looks good, thanks!

I think having fax under the "add field" universal list is better than adding it to a bunch of presets.

PR #3739, opened 2017-01-10
Do not attach dragNode behavior in browse mode (try №2)

edpop:
Good day.
This is a solution for https://github.com/openstreetmap/iD/pull/3276.


bhousel:
Thanks @edpop!  I made a slight change to this in 0fb506f to ensure that the new midpoints also get added to `activeIDs`.. Without this code, the new `osmNode` was constantly trying attach to itself during dragging.

PR #3748, opened 2017-01-12
fix: properties exported are redefined

huangyingjie:
````
export * from './ui/fields/index';
export * from './ui/intro/index';
export * from './ui/index';
````
uiFields and uiIntro are redefined through `export * from ..`. I remove them them from `./ui/index`, because nobody call them from `./ui/index`


bhousel:
Thanks!  I agree they aren't needed..

Issue #1734, opened 2013-08-23
Better indication when an action is selected


jfirebaugh:
Sometimes users select actions from the radial menu without noticing (#1731). The visual indication of selecting an action should be more prominent.

Possibilities:
- Blink the menu item a couple times
- Animated transitions for the change
- Others?

bhousel:
Closing.. Flash messages were added in #3753

PR #3753, opened 2017-01-13
Replace radial menu with a context menu

### LINKED ISSUE #3671: Radial Menu is full
For closed ways, the radial menu is full.

The addition of the Reflect commands pushes the menu to 8 items, which is really too many.  

The radial menu was already problematic, and it's something we've put off dealing with, but now we really should do something about it. 

ccing @samanpwbb for advice

Not looking to bikeshed, but in absence of concrete suggestions or help, I'll probably just change it to be a wider circle, or a fixed menu someplace.

screenshot 2016-12-22 16 06 38

Also tooltip placement on the left side of the menu is broken:
screenshot 2016-12-22 16 19 57

prior discussion
#2145
#1994
#548  
#226

kepta:
Todo
- [x] Fix the tooltip
- [x] Remove the redundant variables for adjusting the position of svgs in edit_menu.js
- [x] Fix the failing test cases

(closes #3671 )


bhousel:
Hey @samanpwbb can you check the screenshot here https://github.com/openstreetmap/iD/issues/3671#issuecomment-272365384
and let us know if you have any thoughts?

samanpwbb:
I think the whole form can be vertically condensed, and simplified. I would lose the semitransparent background – it's distracting and less legible than a solid background. If you make the background solid, you will be able to remove the round background fills from the action icons, which will allow you to cut a good amount of vertical space from the design. End result will be cleaner looking and easier to use.

kepta:
@bhousel I feel this is done.

Could you look over it ?

kepta:
I tried a white style 
white

kepta:
As per talk with @rasagy, he suggested on removing the operations which are always disabled.
We took a bunch of screenshots.

screen shot 2017-02-08 at 12 33 42 pm
screen shot 2017-02-08 at 12 34 00 pm
screen shot 2017-02-08 at 12 34 19 pm
screen shot 2017-02-08 at 12 36 46 pm
screen shot 2017-02-08 at 11 13 42 am

bhousel:
> As per talk with @rasagy, he suggested on removing the operations which are always disabled.
> This would help reduce the height of context menu.
> We took a bunch of screenshots.

I know that the disabled icons were always there for discoverability.  I still think messages like "This feature can't be made circular because it's not a loop" are for users to see.  

Maybe we can just make the icons smaller?  This menu doesn't need to be so much bigger than the browser's default contextmenu.

I really like the light style! 👍

slhh:
@bhousel 
Discoverability is important, but we don't need to overdo it. Always showing every operation seems to be too much, but this doesn't mean that we should always hide disabled operations.
We should always think about what is really required to make the specific operation sufficiently discoverable.

For example "Circularize":
Someone who is not working with nearly circular features doesn't need "Circularize". 
Someone who is working with such will eventually select such a feature and discover the "Circularize" operation, even if it is shown in case of nearly circular closed features only.

> I still think messages like "This feature can't be made circular because it's not a loop" are important for users to see.

Showing this message can be limited to the case of a nearly closed, nearly circular way.

> Maybe we can just make the icons smaller? This menu doesn't need to be so much bigger than the browser's default contextmenu.

Doesn't this make the icons too unspecific? Otherwise I would agree.
Maybe we can reduce the height of the menu rows, but place the icons aligned in two colums alternatingly (zigzag pattern).

rasagy:
> I know that the disabled icons were always there for discoverability. I still think messages like "This feature can't be made circular because it's not a loop" are important for users to see.

@bhousel I think the idea was not to hide _all_ the disabled icons, but only ones that didn’t seem relevant for a particular feature. On exploring a bit with @kepta, we realized that this already happens. Here’s what we felt: actions that might be disabled due to the feature not being in the screen **should** be shown, since they explain why something can’t be done, but actions that won’t be possible for a feature (straight road would never be able to Circularize), it only adds more confusion.

Do you think there are more cases — like Circularize for straight roads — that would be disabled regardless of the context? If this isn’t common, then maybe we can take a quick call on hiding Circularize when it’s not relevant.

bhousel:
The iD architecture document mentions disabled operations a bit here:  https://github.com/openstreetmap/iD/blob/master/ARCHITECTURE.md#operations-module

I don't know who the original author of this section was, but I think it's still important to show disabled operations for discoverability.

The menu is there for teaching in addition to using.  So please keep Circularize on the menu, but disabled for straight roads.

modules/behavior/select.js
```diff
@@ -23,6 +23,12 @@ export function behaviorSelect(context) {
 
 
     function click() {
+        var rtClick = d3.event.type === 'contextmenu';
```
bhousel: `contextmenu` is not _quite_ right click.  On a Mac it fires on `mousedown`, and on Windows it fires on `mouseup`. 

The code here might be totally ok, this is just a reminder to myself to check the reselection behavior on Windows.

modules/modes/select.js
```diff
@@ -139,42 +139,45 @@ export function modeSelect(context, selectedIDs) {
...
 
         var entity = singular();
         if (entity && context.geometry(entity.id) === 'relation') {
             suppressMenu = true;
-        } else if (entity && entity.type === 'node') {
-            radialMenu.center(context.projection(entity.loc));
         } else {
             var point = context.mouse(),
-                viewport = geoExtent(context.projection.clipExtent()).polygon();
+                viewport = geoExtent(context.projection.clipExtent()).polygon(),
```
bhousel: I have to check, I think there is a `context.map().trimmedExtent()` that can be used to get a smaller viewport avoiding UI controls around the edges.

modules/renderer/map.js
```diff
@@ -311,7 +311,7 @@ export function rendererMap(context) {
     function resetTransform() {
         if (!transformed) return false;
 
-        surface.selectAll('.radial-menu').interrupt().remove();
+        surface.selectAll('.edit-menu').interrupt().remove();
```
bhousel: If we're going to keep `.radial-menu` around for backward compatibility and fork projects, we need to include it in the selector here.

modules/ui/intro/point.js
```diff
@@ -143,7 +143,7 @@ export function uiIntroPoint(context, reveal) {
             context.history().on('change.intro', deleted);
 
             setTimeout(function() {
-                var node = d3.select('.radial-menu-item-delete').node();
+                var node = d3.select('.edit-menu-item-delete').node();
```
bhousel: Reminder to 1. keep `radial-menu` classes around for backward compat.  2. test what this actually does when we run through the tutorial.

modules/modes/select.js
kepta: Yes, it totally makes sense to use `context.map().trimmedExtent()`.
I also created this `context.map().trimmedExtentPadding()` for the lane work, which accepts additional padding, should we have such a thing?

modules/renderer/map.js
kepta: Well, so we will end up having two of these (one for radial, other for edit).
I am thinking if about all the cleaning that would need to be done for a major release ..

bhousel: I think that is ok.. `selectAll` takes a CSS selector, so it can be written like:
```js
surface.selectAll('.radial-menu, .edit-menu').interrupt().remove();
```
We can tell people that the radial menu is deprecated and will be removed in iD v3.

modules/modes/select.js
bhousel: > I also created this context.map().trimmedExtentPadding(p) for the lane work, which accepts additional padding, should we have such a thing?

Instead of adding another method, can we just have `trimmedExtent` accept an optional argument with a padding rect?

bhousel:
I edited a lot with this today to test it out. Overall it's great.
I really don't miss the menu except in one situation, and that's when operations are disabled.  

One example of this is:  I'd draw a building and hit 'S' to square the corners of it, and nothing would happen because too much of the building is off screen.  Normally you would be able to see the menu with the disabled icons and realize that something is preventing you from running the command.  But without a menu, I'd find myself wondering whether the command ran or not, and just keep hitting 'S', then eventually right-click to see the menu and realize what was going on.

This is a hard problem to solve elegantly.  #1734 touches on this - "Better indication when an action is selected".  That issue is, sometimes a user will do something by accident and not realize that the action happened.  But now without a menu, it's also possible for a user to try to do something on purpose and not realize when the action _didn't_ happen.

I'm thinking for keyed operations initiated by the user, we should maybe take the message that we already use for undo/redo (e.g. "Made an area circular") and flash it in the corner of the screen.  If the operation is disabled, instead flash the disabled message "This can't be made circular because it isn't a loop".  

_Flash messages are really hard to get right,_ in a way that is tasteful, enjoyable, and delights the user instead of being annoying.  But it would close #1734 and help us replace the visual clues that we lose by not popping up the context menu on every select.  I'm going to prototype this idea and see how it turns out.

kepta:
@bhousel I like the idea of notifying the user that something is not allowed. In the Mac they play that famous sound when an operation is disabled. 
I just want to make sure, by flash messages, we are not planning to blink or do something fancy to grab the users attention right? 

I really like the way `visual studio code` shows its alerts. They just overlay them on the top.
screen shot 2017-02-11 at 12 24 21 am

Maybe we could do something similar for iD? Showing the alert at the top (even if it hides something). Some of these alerts would go off and some wouldn't , depending on the severity. 
This is also similar to how iOS shows notifications.

bhousel:
> Maybe we could do something similar for iD? Showing the alert at the top (even if it hides something). Some of these alerts would go off and some wouldn't , depending on the severity.
> This is also similar to how iOS shows notifications.

Yeah I agree, I'm going to start off displaying something small at the top and see how that looks.  I agree we don't want to be overly distracting or blinky, or have these things stack up and need to be dismissed.  And this would be _only_ be an immediate feedback response to an action performed by the user, not just random "helpful" popups.

bhousel:
@kepta Just trying out different ideas.. It needs work, but it's not terrible..

flash

kepta:
@bhousel looks pretty awesome. I was looking at how other people have implemented it.
I like the way they do it in this screenshot.
toast

Link: [dialogs](http://materializecss.com/dialogs.html)

bhousel:
> Will we be showing this dialog for every operation, or just for keyboard shortcuts?

I'm just planning to try it out with keyboard shortcuts for now.. 

Latest version:
- added the icons
- support succeeded and failed operations 
- configurable showDuration and fadeDuration
- failed operations show for slightly longer (3sec), because there more text for the user to think about

Still unsure about placement.  It's just centered at the top, which would be ok except that the buttons just above the message are left aligned.

I'll probably edit with it like this for an hour and see how it feels.

flash_w_icon

kepta:
@bhousel looks really good. 
I would like to see it in the bottom or top right though.....

bhousel:
>> Will we be showing this dialog for every operation, or just for keyboard shortcuts?
>
> I'm just planning to try it out with keyboard shortcuts for now..

Hmm maybe we should just have this happen on every operation?  It might be weird for it to only happen if the user is using the keyboard shorctuts.   It seems like the sort of thing someone might open an issue about later, like "Why does the flash show up only sometimes?"

bhousel:
re: placement.. There are a bunch of other UI elements that are in the corners.. topleft Minimap, bottomleft Mapillary, bottomright InfoBox, topright Nav buttons.   For now I'm going to move it to top right beside the buttons like @kepta suggested..

screenshot 2017-02-16 10 58 52


I do wish we had a better UI for these addons, because they are all over the place and not at all consistent.  Other tools like Illustrator or Photoshop have nice panels for stuff like this.

three-main-panels

samanpwbb:
the iD UI has gotten spawly. I recommend placing the notice messages here:

screen shot 2017-02-16 at 11 15 59 am

If a black background doesn't make the flash stand out enough, try a background color:

screen shot 2017-02-16 at 11 16 11 am

bhousel:
@samanpwbb That works, and I like how unobtrusive it is..  I'll try it today..

slhh:
@bhousel 
> I edited a lot with this today to test it out. Overall it's great.
> I really don't miss the menu except in one situation, and that's when operations are disabled.
Relying on shortcuts seems to be ok for power users, but I don't think this is good for the casual users.
 
The radial menu is annoying when selecting a feature without intent to use the menu, but it is fine when selecting a feature with intent to use the menu. Requiring left click to select, and right click to open the new menu afterwards would be annoying too.

What we can do instead is offering two different ways to select a feature:
A left click selection which is not opening a menu, and a right click selection which is opening the menu.
In addition we can offer a right button drag based menu operation working on the current selection without selecting a new feature. There is a little issue, as we can't open the menu on right button down, because the selection to work on is unknown at this time. Instead we have to open the menu on right button up or start of dragging, but I think users would get used to it easily.

bhousel:
> What we can do instead is offering two different ways to select a feature:
> A left click selection which is not opening a menu, and a right click selection which is opening the menu.

Yes, this is how it works currently..  I also have to test this on Windows because `contextmenu` fires a little differently there, but I think it should be ok.  `contextmenu` is just the event that would fire when the default browser context menu pops up on right click.. We're stealing that event to pop up an iD menu instead.

> In addition we can offer a right button drag based menu operation working on the current selection without selecting a new feature.

If we ever do support anything on right button drag, it would probably be to rotate the map.  Lots of maps (Google, Mapbox, etc) support this now, so it's pretty common.

bhousel:
Latest version.. I like it!  Thoughts @samanpwbb @kepta ?

footer_flash_w_icon

kepta:
@bhousel, I would love to see a different colour on it. (😅 pink looks good to me)

bhousel:
Final stuff:
- [ ] Show operation title above tooltip
- [ ] Check contextmenu/right-click especially on Windows
- [ ] Check right click preventdefault (right click seems to bring up menu but not activate frame?)
- [ ] Preserve backward compatibility of radial menu
- [ ] Update walkthrough

bhousel:
This was merged!  🎉 
Thanks @kepta @rasagy @samanpwbb & @slhh!

kepta:
Awesome🤘

gmaclennan:
Great work! Testing this in Chrome 56.0.2924.87 when I select an element and hit <kbd>SPACE</kbd> to bring up the edit menu, it appears in the top right of the screen, away from the selected element.

bhousel:
> Great work! Testing this in Chrome 56.0.2924.87 when I select an element and hit SPACE to bring up the edit menu, it appears in the top right of the screen, away from the selected element.

Good catch, I notice that too.. I'll open a new issue for it..

PR #3756, opened 2017-01-13
Upgrade to maki 4, fixes #3024

### LINKED ISSUE #3024: Support maki2

ajithranka:
Updates maki to the latest version and changes some of the icon names in various presets.

Related [discussion](https://github.com/mapbox/maki/issues/270).

cc: @bhousel @kepta


bhousel:
This is great and the new icons look sooo good.. Thanks for taking this on @ajithranka 🚀 
cc @samanpwbb @slaughternat @aparlato

PR #3759, opened 2017-01-13
Do not suggest duplicate commit messages

Abbe98:
This makes sure no duplicate commit messages are suggested.

The issue illustrated:
screen shot

I had to add a container variable to be able to use `indexOf`, the alternative would have been a loop as far as I know...

### DIFF
--- a/modules/ui/commit.js
+++ b/modules/ui/commit.js
@@ -59,13 +59,18 @@ export function uiCommit(context) {
             if (err) return;
 
             var comments = [];
+            var addedComments = [];
 
             for (var i = 0; i < changesets.length; i++) {
                 if (changesets[i].tags.comment) {
-                    comments.push({
-                        title: changesets[i].tags.comment,
-                        value: changesets[i].tags.comment
-                    });
+                    if (addedComments.indexOf(changesets[i].tags.comment) === -1) {
+                        comments.push({
+                            title: changesets[i].tags.comment,
+                            value: changesets[i].tags.comment
+                        });
+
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks!  I changed it a little in 132a5b5 to use lodash `_.uniqBy` for simplicity.

Abbe98:
Thanks @bhousel didn't know about `_.uniqBy` :-)

Abbe98:
@tyrasd changeset comment suggestions to be working over at https://openstreetmap.us/iD/master/ 
image

tyrasd:
ok, I see it now.

But, apparently, the API-0.6 [changesets-call](http://wiki.openstreetmap.org/wiki/API_v0.6#Query:_GET_.2Fapi.2F0.6.2Fchangesets) can be quite slow for users with many changesets. E.g. for [me](https://www.openstreetmap.org/api/0.6/changesets?user=115612) it takes around 5 to 6 seconds to load my last 100 changesets (for @manfredbrandl's account the situation is similar). And only after that the dropdown arrow will be shown.

Suggestion: Assuming that most users want to save their changes eventually, we could also load a user's changesets right after iD starts up (caching the contents locally), so that they're available immediately when the save-dialog is displayed.

bhousel:
> Suggestion: Assuming that most users want to save their changes eventually, we could also load a user's changesets right after iD starts up (caching the contents locally), so that they're available immediately when the save-dialog is displayed.

This is a good idea...

Issue #3755, opened 2017-01-13
Add language debugging mode


bhousel:
This is an idea from today's show-and-tell which would help translators a lot.

It would be great to be able to switch iD into a mode where instead of the translation strings, you just see the translation keys.  This in itself isn't too hard to do by changing what [ the code in `locale.t()`](https://github.com/openstreetmap/iD/blob/master/modules/util/locale.js#L25) does.

The tricky part is that there are probably lots of components of iD that are only entered once and never updated again.  When switching the language, we'd really need to rebuild the ui from scratch.  [This code is in `ui/init.js`](https://github.com/openstreetmap/iD/blob/master/modules/ui/init.js) and is really not designed to be rerun.  It might be ok to preserve some of the map state and throw out all that stuff and reenter it - I don't know.  It's like how iD can switch in and out of the walkthrough, but probably more complicated.

This would enable #3120 too (switching languages on the fly).

bhousel:
So, I did this today:

<img width="1279" alt="screen shot 2017-01-26 at 11 48 18 am" src="https://cloud.githubusercontent.com/assets/38784/22341036/74e7f088-e3bd-11e6-86a9-64e13cb9be9a.png">

It's kind of weird but it works.. Is this something that would be useful to people doing translations?

The idea is that now iD can switch languages on-the-fly, and I added a special language called `_tkeys_` that shows the keys instead of the strings.

Right now toggling languages can only be done through the dev console (we can add a proper field to pick the language later, closing #3120):

```js
id.ui().restart('it')
id.ui().restart('ru')
id.ui().restart('zh')
id.ui().restart('_tkeys_')
```

cc @1ec5 @manfredbrandl @slhh @tyrasd

1ec5:
Yes, that could be useful for zeroing in on an incorrect translation I spot while using iD. I sometimes use a similar mode when translating MediaWiki ([example](https://en.wikipedia.org/wiki/OpenStreetMap?uselang=qqx)), although fortunately Transifex’s search engine is much more usable than Translatewiki.net’s.

More commonly, I need to go in the other direction while translating in Transifex: figuring out where a string marked for translation is used within iD. For that purpose, I usually head over to this repository’s commit history and root around for the relevant change. (Thank you for descriptive commit messages!) One way to improve this workflow would be a mode where untranslated UI elements are highlighted prominently or greeked (replaced with lookalike ¢ħª®ä©¯ţǝ®§ or emoji).

magol:
[...]

Issue #3120, opened 2016-05-18
Toggle between local language and English


slhh:
Translations are error-prone and potentially misleading. In addition English field names do better correspond to the wiki because the OSM wiki is not translating keys and tag values.

Therefore, the user should be able to toggle between his local language and English as iD's original language during the edit session.

By the way this could also improve the quality of iD because of simplifying translation verification and general bug reporting, because using local language names of iD's elements in bug reports here doesn't make much sense.

matkoniecz:
<img width="430" height="127" alt="Image" src="https://github.com/user-attachments/assets/033623e4-7448-49d5-a9c2-0213149f5814" />

this button can be now used

it reloads page, but it is fine as session is preserved

PR #3764, opened 2017-01-16
Change translation on the fly

kepta:
Need to fix the history and map flying to some random area.

cc: @bhousel


bhousel:
awesome, this will unblock #3755 and #3120

modules/ui/init.js
```diff
@@ -228,6 +228,7 @@ export function uiInit(context) {
 
         var mapDimensions = map.dimensions();
 
+
```
pnorman: extra newline

modules/ui/init.js
```diff
@@ -247,6 +248,7 @@ export function uiInit(context) {
             };
         }
 
+
```
pnorman: extra newline

bhousel:
> How can I test this?

I'm pretty close to merging it, but you can check out this branch and when running, open a developer console call `restart` with a language code:

* `restart('ru')`
* `restart('fr')`
* `restart('ar')`

etc...

modules/ui/init.js
bhousel: fwiw, I've been adding 2 spaces before function blocks and anywhere else I want documentation to go.

Issue #3777, opened 2017-01-18
Make it clearer that the user needs to do something to save


bhousel:
Sometimes a new user hits "Save", then thinks they are done and just continues editing, not realizing that they need to click another "Save" button.

We could callout the save panel somehow (motion?), and/or darken the map.  Also maybe consider renaming the "second" save button to "Upload"

![screenshot 2017-01-18 18 44 04](https://cloud.githubusercontent.com/assets/38784/22065576/ddaae162-ddae-11e6-9e8a-a607bf5b2c3a.png)

boothym:
> Sometimes a new user hits "Save", then thinks they are done and just continues editing,

But you're unable to click on an object in the editing screen until you either save or cancel - or do they just close the tab? (though that should give a warning about unsaved changes from the browser)

kepta:
@bhousel, since we are talking about the save button. 
It would be really nice to add an icon next to  `save` button (or whatever we rename it) to make it consistent with other buttons in the `bar`.

bhousel:
> It would be really nice to add an icon next to save button (or whatever we rename it) to make it consistent with other buttons in the bar.

Yes I agree.. Just not a floppy 💾

bhousel:
see also #2378

bhousel:
Here's what I did:

- [x] When entering Save mode, map content transitions inactive: appears grayscale and dims slightly so colorblind users perceive a change.  The sidebar controls still let the user review their edits, or return to editing to address any warnings.
- [x] Adjusted the language a little bit "Save" -> "Upload" to make it clear that an Upload needs to happen.  No more 2 Save buttons. 👍 

This along with the previous changes in #2378 should make it impossible for a user to ignore that they have an extra step to complete in order for their changes to save.

![save inactive](https://cloud.githubusercontent.com/assets/38784/22258328/2a3337bc-e22f-11e6-8fda-bbbb7bb5316a.gif)

bhousel:
Leaving this open so we can 
- [ ] Add Icon to Save Button

tyrasd:
> Adjusted the language a little bit "Save" -> "Upload" to make it clear that an Upload needs to happen.

:+1: That should also improve the occasional confusion of users who don't realize that by clicking on "save" their changes actually get published for everyone (like this recent example: https://www.openstreetmap.org/changeset/45246084).

bhousel:
Added the icon, looks nice 👌 
[...]

PR #3788, opened 2017-01-22
show "You have unsaved changes" message also in save mode

tyrasd:
While it makes sense to not save the state of the history in save mode (see 15bc08795d5a0daaeaddd804cdc6729e1dfff848) we should still display the *"You have unsaved changes"* browser message. This should remind people that changes are not actually saved unless the second save button is also pressed (see #3777).


bhousel:
Good catch, thanks!

PR #3791, opened 2017-01-25
Display a shortcuts modal

ajithranka:
Shows a subset of shortcuts from this [wiki page](https://wiki.openstreetmap.org/wiki/ID/Shortcuts) on pressing `?`.

screen shot 2017-01-25 at 12 48 19 pm

cc: @bhousel @kepta


bhousel:
This is a great start!   Next step is to move the English strings from `shortcuts.json` to `core.yaml`.  Then `shortcuts.json` with contain translation keys and `shortcuts.js` will use `t()` to display them.

ajithranka:
@bhousel I couldn't make enough space in the modal (without making the width larger, allowing for a two column format) to show all shortcuts in the wiki. Is it ok to show only the most used shortcuts or should we display all of them?

slhh:
@ajithranka We shouldn't generate secret shortcuts by omitting them from the display.

In case we want to omit something from the modal, we can omit all shortcuts for buttons which are directly visible on the UI, e.g. radial menu operations, or undo/redo/save. We can do that, because the operation is easily accessible by using the button, and the shortcut is discoverable due to the hover text.

In case we want to display all shortcuts the modal could have multiple tabs, or collapsible sections plus scrollbar.

boothym:
I assume that the Mac command key will change to control when using a browser on Windows?

bhousel:
> I assume that the Mac command key will change to control when using a browser on Windows?

Yes, `iD.uiCmd` ([link](https://github.com/openstreetmap/iD/blob/master/test/spec/ui/cmd.js#L45)) contains the code to adjust the keyboard shortcuts between Mac / Windows / Linux.

modules/ui/shortcuts.js
```diff
@@ -0,0 +1,69 @@
...
+            shortcuts
+                .selectAll('kbd')
+                .data(function(d) { return d.shortcut; })
+                .enter().append('kbd')
+                .text(function(d) { return uiCmd(d); });
+
+            var description  = p
+                .append('span')
+                .attr('class', 'col8')
+                .text(function(d) { return t('shortcuts.' });
```
kepta: @bhousel I want to form for e.g.  `t('shortcuts.display.zoom')` over here. d would have `d.key == 'zoom'`. But what would be the best way to get `display`. Is the approach fine?

modules/ui/shortcuts.js
```diff
@@ -0,0 +1,144 @@
+import * as d3 from 'd3';
+import { uiCmd } from './cmd';
+import { uiModal } from './modal';
+import { d3keybinding } from '../lib/d3.keybinding.js';
+import { t } from '../util/locale';
+import { shortcuts as shortcutsData } from '../../data';
+
+export function uiShortcuts() {
+    var key = uiCmd('⇧/');
```
bhousel: I think we only need uiCmd when there's a ⌘ involved.

kepta:
@bhousel lets merge this?

bhousel:
> lets merge this? 😎

Sorry I haven't looked at it lately..  My focus right now is merging the edit menu, which is a big change to iD but will unblock like 10 other issues.  Some questions about this PR:
- does it include all the keyboard shortcuts?
- is `shortcuts.json` now the single source for the keyboard shortcuts across iD (good) or do we need to update shortcuts in multiple places (less good but not dealbreaker)
- export it as `dataShortcuts`, not `shortcuts`
- does it work RTL?
- does it work on smallish screens?  is it responsive if I resize the browser window?

kepta:
rtl version
screen shot 2017-03-09 at 9 23 26 pm

small screen
screen shot 2017-03-09 at 9 17 21 pm

kepta:
> lets merge this? 😎

Sorry I haven't looked at it lately..  My focus right now is merging the edit menu, which is a big change to iD but will unblock like 10 other issues.  Some questions about this PR:
- does it include all the keyboard shortcuts?
- is `shortcuts.json` now the single source for the keyboard shortcuts across iD (good) or do we need to update shortcuts in multiple places (less good but not dealbreaker)
- export it as `dataShortcuts`, not `shortcuts`
- does it work RTL?
- does it work on smallish screens?  is it responsive if I resize the browser window?

bhousel:
Just rebased this branch to master.  Looks like a great start - getting it merged will be my focus this week.

magol:
How do I view this? When I press `?` i only zoom in on the map (The `+` sign is on the same key on a Swedish keyboard)

bhousel:
Thanks @magol - I just made a change that should fix this, but I'm going to open a separate issue just in case it didn't.

Issue #3799, opened 2017-01-26
Issue with dragging nodes on multipolygon outers


bhousel:
The nodes on this multipolygon can't be dragged.
http://localhost:8080/#background=Bing&id=r276895&map=16.77/40.37391/-74.93380

slhh:
Here is a non-local link:
https://openstreetmap.us/iD/master/#background=Bing&id=r276895&map=16.00/40.3769/-74.9077

The nodes can't be dragged when the multipolygon is selected, but they can be dragged when the way is selected.

PR #3801, opened 2017-01-27
Fix dragging nodes on multipolygon

edpop:
Fix for https://github.com/openstreetmap/iD/issues/3799. That was my fault in https://github.com/openstreetmap/iD/pull/3739.
Sorry for commits, I'm a git lamer :{. Take this changes by compare.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -839,11 +839,6 @@ en:
       outdoor_seating:
         # outdoor_seating=*
         label: Outdoor Seating
-        options:
-          # outdoor_seating=yes
-          yes: Outdoor seating is possible
-          # outdoor_seating=no
-          no: No outdoor seating
       par:
         # par=*
         label: Par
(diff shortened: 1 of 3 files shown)


bhousel:
Thanks for the quick fix @edpop this looks great!.. 

I did a bunch of editing yesterday and I think I might also extend "editable" to include sibling vertices (other children with same parent).  This is because sometimes I'd find myself editing where I would select a vertex and want to be able to drag other nearby vertices along that way.  Overall though this is working great.

cherry-picked & merged it as:
ad06e34 2017-01-27 do not use tabs
56dbebd 2017-01-27 fix array concat
a2f9f51 2017-01-27 Allow to drag a node when a parent relation is selected

bhousel:
@edpop Quick followup, check out 0440054... I realized that vertices.js already does all the hard work of figuring out what the children and siblings vertices are, so I just added a "sibling" class and look for that in drag_node to decide whether or not to allow the drag.

Issue #2586, opened 2015-04-12
Support multiple layers for managing imports


mikelmaron:
JOSM allows multiple data layers, which is handy when doing a merge of import data. Makes it easy to compare vs existing OSM data, and you can copy/paste features or tags between layers. For iD, simply allowing to have an OSM layer, and a "reference" layer (perhaps just a GeoJSON file passed by URL or from the local computer), could suffice.

bhousel:
As an intermediate step towards this, I think it makes sense to give the GPX layer a better name (suggestions welcome) and support a few other file formats. I'm kind of surprised iD doesn't support geojson here. 

Then, actually giving the user some ability to import those features as OSM objects would be cool too, if we can keep it simple and do it in a way that doesn't break anything.

mikelmaron:
@bhousel makes sense. first step sounds pretty straightforward? and second step, what would it take to make the "(suggestions welcome)" layer interactive, clickable, and copyable?

bhousel:
> what would it take to make the "(suggestions welcome)" layer interactive, clickable, and copyable?

I don't think that we would need to make the reference layer editable.. We could just add a button to import whatever is there (gpx/kml/geojson) onto the main map where it will be editable.  Some potential issues:

* Imported features need to have presets assigned
* Imported features often need simplification and we don't have an operation for that (yet)
* Ensure that the user understands not to import improperly licensed data
* Consensus that this is a feature that should be built in the first place.  OSM community tends to be pretty anti-import lately.

But anyway I like the idea.  I have a bunch of gpx tracks sitting on my computer that I am too lazy to trace.

pnorman:
Is this in scope for a basic editor? I recall imports an external data sources being out of I'd scope in the past.

ajturner:
I think that this would be very helpful - at least from the concept story: 

**"As a User, when editing in iD, I can add layers from other Services that I can edit and push into OpenStreetMap"**

- iD should keep it's core data model focused on OSM nodes/ways/relations.
- a plugin would be loaded at runtime that adds support for API conversion
- a URL to a GeoJSON file would be handled by this plugin
- plugin parses the GeoJSON and converts it into OSM nodes/ways and hands it to iD core - output is same as if the user had hand-drawn the objects
- User can now interact with the objects and commit to OSM

This is something I'd like to help out on.

systemed:
[...]

PR #3808, opened 2017-01-31
Add geojson data layer

mertemin:
I had some geojson data from local sources that I wanted to use to create new objects and edit/improve existing ones. Rather than doing this programmatically (bulk imports or other ways of import etc) I wanted to have a way to display data on the map and work on it.

Even though iD had a data layer concept, I couldn't find a way to get some geojson data displayed on the map. I saw the `gpx` one though and used that as an example to create geojson data layer. Unlike `gpx`, geojson would have more information than some location points, but I think for now this is a good start.

I am not sure if using geojson as a base to edit is a common scenario or not. But I wanted to share this with the community anyway. I am using this feature locally and it has been helpful so far.

Looks like this is related to https://github.com/openstreetmap/iD/issues/2586

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -3588,7 +3588,8 @@ img.tile-removing {
   border-radius: 0 0 4px 0;
 }
 
-[dir='rtl'] .list-item-gpx-browse svg {
+[dir='rtl'] .list-item-gpx-browse svg,
+[dir='rtl'] .list-item-geojson-browse svg {
   transform: rotateY(180deg);
 }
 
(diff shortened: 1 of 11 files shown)


bhousel:
Yes, we did discuss it a bit in #2586.  I think a better approach, which I mentioned on that issue, is to just have the "gpx" layer also support geojson and kml.  In fact we just use the togeojson library internally to convert the gpx into geojson for d3 to render.  So there really isn't a need to take all that gpx layer code and duplicate it when we could just change the existing code and help text slightly.

PR #3811, opened 2017-02-01
Add kml and geojson support to gpx layer

mertemin:
After @bhousel's comment on https://github.com/openstreetmap/iD/pull/3808 I updated existing gpx layer, added ability to read in .kml and .geojson files based on file extension.

### DIFF
--- a/modules/svg/gpx.js
+++ b/modules/svg/gpx.js
@@ -106,6 +106,33 @@ export function svgGpx(projection, context, dispatch) {
     }
 
 
+    function getExtension(file) {
+        if (_.isUndefined(file)) {
+            return '';
+        }
+
+        var lastDotIndex = file.name.lastIndexOf('.');
+        if (lastDotIndex < 0) {
+            return '';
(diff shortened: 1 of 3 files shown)


bhousel:
Awesome!  Yes this is what I had in mind 👍   I'll try it out tonight or tomorrow...

bhousel:
Works perfectly!  Just merged, thanks again!

modules/svg/gpx.js
```diff
@@ -139,7 +166,8 @@ export function svgGpx(projection, context, dispatch) {
     drawGpx.url = function(url) {
         d3.text(url, function(err, data) {
             if (!err) {
-                drawGpx.geojson(toGeoJSON.gpx(toDom(data)));
+                var extension = getExtension(url);
```
tyrasd: FYI: this caused #3820 since url is just a string and getExtension() expects a File object. I fixed it by changing getExtension() to expect a string instead.

PR #3821, opened 2017-02-05
2 new presets and add living street to Road preset category

willemarcel:
Add presets to:
* noexit=yes -> https://wiki.openstreetmap.org/wiki/Key:noexit
* shop=watches -> https://wiki.openstreetmap.org/wiki/Tag:shop%3Dwatches

and include living street prest to the road category.


bhousel:
Kind of iffy on `noexit`.  It's really only appropriate in certain situations, and it's purpose is mostly just to silence Q/A tools, it doesn't actually add any information about the endpoint of a way.  A user would need to read the wiki just to learn when to use it.

Everything else seems ok.

PR #3826, opened 2017-02-06
improve autozooming on non-gpx background data layers

tyrasd:
The original implementation only worked for Points and LineStrings, but GeoJSON and KML can contain any geometry.

Aso, d3 is a bit [special](https://github.com/d3/d3-geo#d3-geo) when working with GeoJSON Polygons: it does require clockwise winding, but typical GeoJSON uses either counterclockwise winding ([RFC7946](https://tools.ietf.org/html/rfc7946#section-3.1.6)) or "winding doesn't matter" mode (old geojson.org spec). When inputing such GeoJSON (or KML) input, iD would zoom out all the way, which is not the intended behavior.

### DIFF
--- a/modules/svg/gpx.js
+++ b/modules/svg/gpx.js
@@ -203,11 +203,25 @@ export function svgGpx(projection, context, dispatch) {
             viewport = map.trimmedExtent().polygon(),
             coords = _.reduce(geojson.features, function(coords, feature) {
                 var c = feature.geometry.coordinates;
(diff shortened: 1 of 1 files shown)


bhousel:
> Also, d3 is a bit special when working with GeoJSON Polygons: it does require clockwise winding,

Whoa I did not know this, good catch.. All the GeoJSONs I tried this with contained LineStrings..

PR #3830, opened 2017-02-08
Use cache to speed up development

kepta:
I noticed rollup provides [cache](https://github.com/rollup/rollup/wiki/JavaScript-API) to speed up subsequent builds.
I am noticing 30-40% improvement.  😎 ( from 8500ms to 5000ms)


bhousel:
Nice!  for some reason I thought we were using this already..

Issue #3803, opened 2017-01-27
"Save" button deformation when count span appears


edpop:
Hello.
In some languages, as 'ru' and 'fr', "save" button deforms, when a count-span appears.

![-2](https://cloud.githubusercontent.com/assets/13479292/22362301/5b76ee88-e4ad-11e6-9bdd-316d085b74b2.png)
![-1](https://cloud.githubusercontent.com/assets/13479292/22362216/cb1cdb22-e4ac-11e6-93fb-9fd625cba17a.png)

I'm not a frontend developer, but I tried to do something like this, and It looks fine.
```css
/*button.save.has-count {
    padding: 9px;
}*/

button.save .count {
    display: none;
}

button.save.has-count .count {
    display: block;
    position: absolute;
    top: 5px;
    background: #fff;
    border-color: #fff;
    opacity: 0.5;
    color: #333;
    padding: 10px;
    height: 30px;
    line-height: 12px;
    border-radius: 4px;
    margin: auto;
    margin-left: 9.1%; /*margin-left: 8.3333%;*/
}
```
![default](https://cloud.githubusercontent.com/assets/13479292/22362215/cb19374c-e4ac-11e6-9c20-6b58c022726c.png)

bhousel:
Thanks for the fix, just merged it 👍

PR #3834, opened 2017-02-10
Save -> Save... ; Save -> Review and save

tbertels:
Make it clear that clicking on the Save button doesn't actually save the changes immediately and that the user will be able to review them, as stated in ### Saving Edits

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -402,8 +402,8 @@ en:
     restore: Restore
     reset: Reset
   save:
-    title: Save
-    help: "Save changes to OpenStreetMap, making them visible to other users."
+    title: Save...
+    help: "Review your changes and save them to OpenStreetMap, making them visible to other users."
     no_changes: No changes to save.
     error: Errors occurred while trying to save
     status_code: "Server returned status code {code}"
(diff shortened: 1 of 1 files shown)


manfredbrandl:
I prefer "Upload..." and "Review your changes and upload them to" ... and avoiding save as that may mean local/private Save.

bhousel:
Since we just changed this recently in #3777, I'd prefer not to keep changing the save language unless we notice an uptick in users that find it confusing.  (People have complained before when we change too many things too quickly, because they need to update translations and training materials).

Also, I agree conceptually with ellipses on the save button (as this usually signals another step will follow) but I'm worried about making the Save caption any longer because of issues raised in #3803, so I want to leave it how it is for now.

One frustrating thing about Transifex is that translators can't easily preview how their translated strings will look in iD, so I could imagine a situation where somebody adds the ellipses and then that makes the button caption look weird, but we don't find out about it until next release.

Thanks for the suggestions though!  Keep them coming!

tbertels:
Couldn't at least the tooltip be changed? Maybe to "Review your changes and upload them to" like suggested by manfredbrandl.
That shouldn't require a documentation change and would improve consistency (since the next button is "Upload", not "Save" anymore).

bhousel:
> Couldn't at least the tooltip be changed? Maybe to "Review your changes and upload them to" like suggested by manfredbrandl.

Ok, yes, we can change the tooltip.

PR #3835, opened 2017-02-10
Add ".." to "Uploading changes to OpenStreetMap."

tbertels:
Since it's in progress.


bhousel:
Agree, this is ok, thanks!

PR #3837, opened 2017-02-13
Update UK/IE phone placeholders

boothym:
The UK phone number placeholder was a digit out, so I've updated the format with a fictitious number as specified here: https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama

Also added Ireland's format, with a fictitious number from here: https://en.wikipedia.org/wiki/Telephone_numbers_in_the_Republic_of_Ireland#Special_numbers


bhousel:
Thank you!

PR #3840, opened 2017-02-15
Display housenumbers in labels

edpop:
Hello.
There is housenumbers in labels. Discussion in https://github.com/openstreetmap/iD/issues/1524.
default
Building icons taken a lot of space, so many housenumbers are not shown. Probably we should remove icons or improve zoom options.

---

TODO*
---
- [x] In `svg/labels.js`: `label` → `nameLabel`, `houseNumberLabel`
- [ ] Trigger housenumbers vision
  - depending on zoom
  - depending on total numbers count (for not dense places)?
- [x] Get housenumbers from special function near `utilDisplayName()`
  - needs localize through `address-formats.json`?
- [ ] Unique housenumbers by area buildings (_group of a building, its housenumbered vertices and interior points has one housenumber label_)
- [x] Place housenumber label on a building edge close as possible to its street

*discussion


bhousel:
> Building icons taken a lot of space, so many housenumbers are not shown. Probably we should remove icons or improve zoom options.

Nice, thanks for taking on this issue!  This is definitely a challenging one because of how much space things take up.   Also we'd want the housenumbers to show up in addition to the labels..

I haven't thought about this too much but I wonder if it would make more sense to have a totally separate drawing function for the house numbers, rather than having `labels.js` draw them?   It would give us more flexibility to put them on their own layer, style them differently from the labels, and have them only render at high zoom.

Thinking about it more, I'm going to add to a followup comment..

bhousel:
This part of the map in NYC shows how addresses can be mapped a bunch of different ways... http://www.openstreetmap.org/edit#map=18/40.74292/-74.00632

screenshot 2017-02-15 10 57 06 copy

How iD renders things in this screenshot is interesting.

So the **building** is 431 West 16th Street.  We would want to put a _single_ 431 somewhere in there, ideally at a building entrance.

There is a supermarket **POI** in there called Western Beef, and it also has an address tag for 431 West 16th Street.

The Highline Ballroom is mapped as a **vertex** on the edge of that building and it also has the address tag for 431 West 16th Street.  It looks silly because the vertex label appears under the supermarket POI, but this is just chance of how the names got inserted into the label collision rbush.  Really this nightclub should probably be mapped as a node, not a vertex.  (But building entrances are regularly mapped as vertices on the edge of buildings and have address tags).

And of course, a user can just put an address tag anyplace as a **standalone node**, we have the Address preset for this.  There are lots of these scattered around too.

I think the ideal solution might:
- not display duplicates
- prioritize address points and entrances over addresses tagged on other features
- place the labels near the edges if possible
- avoid the labels?  (should addresses take priority over labels or the other way around?)
- maybe use a different marker style (not the rounded marker) for address points
- only attempt this at higher zoom (z18+?)

I haven't even thought about addr:interpolation ranges, but they complicate things too.

edpop:
Thank you for explanation! I added some todos, they needs review from different countries.
I aggree very much to change standard point marker wherever it possible, especially addresses, because they are many.

slhh:
I would limit housenumber rendering of the building area to z18 or higher, and rendering of housenumbers on nodes to z20 and higher. At lower zoom levels housenumbers on nodes might be rendered misleadingly in the wrong building othererwise. At z20 housenumber labels seem to be sufficiently attached to the node to make it clear.

I think it is more important to indicate all elements which have address tags, than to show the housenumber of some addresses. Maybe we can use a yellow background for all nodes with address tags (POI, pure address node, or entrance), and line style with yellow dots for a building outline with address tags.

In case we want to suppress duplicate housenumbers, we should only suppress duplicates within a single building, and center the remaining housenumber label in the building. Errors, like the same housenumber tagged on different buildings, should not be hidden.

edpop:
I explored the `svg/label.js` module and thought about the task..

- We use `utilDisplayName()` for a 4 times for each entity. Is there any problems with memory to cache it at [first iteration](https://github.com/openstreetmap/iD/blob/v2.1.2/modules/svg/labels.js#L289)?
- We can reuse `getPointLabel()` and `drawPointLabels()` for housenumbers.
- If housenumbers will also have halos, there will be `nameLabels` and `housenumberLabels` with halos into.

The only ambiguity is where to put housenumber in a variety of situations.
Anyway, this module will be greatly rewritten, if we want housenumbers in a distinct layer.

We can do a small update for now to show housenumbers in buildings and points, if they have not a `name` tag and get a feedback from community then.

bhousel:
> We use utilDisplayName() for a 4 times for each entity. Is there any problems with memory to cache it at first iteration?

It's probably ok to cache.

> We can reuse getPointLabel() and drawPointLabels() for housenumbers.

I don't think we can, as doing so would put the housenumbers in the same place as the point labels, right?  I think for code clarity I'd rather have `getHousenumberLabel` and `drawHouseNumberLabels` methods, even if the code does look mostly similar.  I don't think we'll end up putting the housenumbers at the entity centroid.  It would be nice to put them near the polygon edge in the direction of whatever street the address goes to.

> If housenumbers will also have halos, there will be nameLabels and housenumberLabels with halos into.

Yes, I think they'll need halos.

> Anyway, this module will be greatly rewritten, if we want housenumbers in a distinct layer.

Yeah thinking about it more, it's probably best to just draw them in `svg/labels.js` on the same layer as the other labels. Trying to figure out cross-layer label collision would be annoying.

> We can do a small update for now to show housenumbers in buildings and points, if they have not a name tag and get a feedback from community then.

I really don't want to show the housenumbers where the names are currently drawn.  This will confuse people.  New mappers will start adding housenumbers into the name field just to get them to render.

This is one of those issues where - we haven't had this feature before in the several years that iD has been around, so it's ok to move slow, make it perfect, and not rush something into production just to close an issue and get feedback on it.  I see this as a hard problem.

edpop:
Thanks again for detailed explanation.
Updated TODO. I hope I will have enough time to finish.

edpop:
The logic in `svg/labels.js` is based on different arrays of the same size: `positions` & `labelled`. This trick uses in all subfunctions `drawPointLabels()`, `drawLineLabels()`, `drawAreaLabels()`, `drawAreaIcons()`. Following this logic, if I want to cache displayNames (and displayHousenumbers, in the future), I should create a new array and put it as next (already 6!) parameter in all this functions. Do you think that it is necessary to normalize all rendering parameters to one jdocumented collection?

bhousel:
> Following this logic, if I want to cache displayNames (and displayHousenumbers, in the future), I should create a new array and put it as next (already 6!) parameter in all this functions.

Definitely feel free to refactor this code however you want 👍 
It's probably better to just have one datastructure with all the data inside it rather than just adding more arrays of arrays, and arguments to every function..

edpop:
Nice, it works for areas with street and housenumber.
default

Now we should solve some cases:
- different addresses of points into one building

bhousel:
Wow @edpop this looks great!  I like how it's coming along...

edpop:
Now the most difficult part - what to do in set of cases and their combinations.
There is very simple address tagging in my city, so I need your help to know, which cases should be solved and which cases should be ignored.

- What is building (for POI/vertex parent searching)?
  - closed area with white-list tags (building, landuse=construction, other?)
  - any closed area
- Whole building is numbered - ignore points and vertexes on it?


For example the case: http://localhost:8080/#background=Bing&map=21.44/40.74212/-74.00634.

Reasoning
--------------
_Suppose for example, that amenity=car_sharing POI tagged right to number `422` and street `West 15th Street`._
So we have three addresses on one edge and two of them (412 and 418) are very close to each other - interpolate each address to edge, split it to three segments and label each segment by its address.

There is also Meatpacking District Hotel located outside the building - is it an error (should be ignored or snapped)?

manfredbrandl:
I would prefer to display all housenumbers. display those stored as vertex, line/area, relation. display those stored independend of other tags. So everyone can see if there are possible errors and correct them without clicking on every object that may contain housenumbers.

PR #3849, opened 2017-02-19
add preset to landuse=aquaculture

willemarcel:
https://wiki.openstreetmap.org/wiki/Tag:landuse%3Daquaculture


bhousel:
Thanks @willemarcel!

I also styled these blue like other watery things and added the "Produce" field.

willemarcel:
Great! Thank you!

PR #3854, opened 2017-02-23
fixed gitignore

MindFreeze:
Line 14 (img) caused the entire /dist/img directory to be ignored even though it is already in git, so new images were ignored. Prefixing it with / prevents this very general rule from applying to the subdirs. The same logic applies to the other rules for top level dirs.


bhousel:
Thanks for this!

PR #3861, opened 2017-02-26
Fix some spelling mistakes and translation errors

mikini:


bhousel:
Hi, thanks for the report. Our translations are handled through [Transifex](https://www.transifex.com/ideditor/id-editor/dashboard/) so you will need to sign up for an account there to correct the translations.  

The files in the iD repository get overwritten by the latest translations from Transifex whenever we release a new version of iD.

mikini:
Ok, sorry for not having studied the procedures beforehand. Have submitted through Transifex now.

bhousel:
Thanks @mikini!

PR #3871, opened 2017-03-02
Order to-be-created relations by dependencies in the changeset

### LINKED ISSUE #3208: Can't create a new master relation for new relations
Hi,

I wanted to add a bus route is OpenStreetMap today using iD, but I got the following error message : 

> Placeholder Relation not found for reference -3 in relation -2. (Server returned error code 400.)

Here is what I tried to do : 
- Create a relation for a direction (A -> B)
- Create an other relation for the other direction (B -> A)
- Create a relation (route master) containing the two previous relations
- Mapping
- Sending to the server ... uh oh ...

Here is what I've done : 
- Delete route master
- Sending to the server ... it worked !
- Finally, re-creating the route master in a new changeset (it worked too)

I don't really know what happened, nor why it worked when deleting the route master ... Hope this helped.

mstn:
This patch fixes #3208. In details:

* New relations are ordered in the changeset so that children appear in the file before their parents.
* Circular dependencies are not ordered because it is not possible. In this case, an osmChange will be generated, but the API will return an error. Ideally, the user interface should prevent users from creating circular dependencies. Hence, this case should not happen.
* The bug is specific to creations. I do not think that  we need a similar fix also for modifications and deletions.

Hope everything ok. Otherwise please let me know! :)

### DIFF
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -319,7 +319,8 @@ export default {
           // find a referenced relation in the current changeset
           function resolve(item){
             return _.find(relations, function(relation) {
-              return item.keyAttributes.ref === relation['@id'];
+              return item.keyAttributes.type === 'relation'
+                && item.keyAttributes.ref === relation['@id'];
             });
           }
           // a new item is an item that has not been already processed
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @mstn!  This looks really good 👍 
I will try it out over the weekend...

bhousel:
I tested this out and it does reorder the relations.. Thanks again!

PR #3881, opened 2017-03-07
reorder path preset category

willemarcel:
I ordered better the `path` presets category, based on type and most used.

captura de tela de 2017-03-07 14-40-53

### DIFF
--- a/data/presets/categories.json
+++ b/data/presets/categories.json
@@ -99,13 +99,13 @@
             "geometry": "line",
             "name": "Path Features",
             "members": [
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @willemarcel, but I'd prefer not to order the preset lists by most used.  This comes up for discussion sometimes and in the past we have decided these category lists are more about teaching new users what they _can_ map, rather than speeding up data entry by putting common presets near the top.

However I will move "Pedestrian Street" further down the list, since those are somewhat more rare than the other presets.

PR #3894, opened 2017-03-13
fix text padding in key-value input field

51114u9:
When you input/retrive long texts in the ket-value box, the combobox caret hide the end of it

captura de pantalla - 130317 - 01 15 49


bhousel:
Looks pretty good to me.  Can you check this also with `&rtl=true` in the url, just to make sure it looks ok right-to-left.

51114u9:
@bhousel no, I will check.

51114u9:
@bhousel ok, now right-to-left looks good.

bhousel:
Can you double check the css rules?  Now it looks like all the fields have extra padding.. 
(I'm in LTR right now):

screenshot 2017-03-15 16 18 53

51114u9:
my error, but I fixed. I tested with two directions and looks good.

PR #3896, opened 2017-03-13
Normalize form with that from OsmHydrant.org

ToastHawaii:


bhousel:
Thanks, its a good start!  In order for the fire hydrant preset to show those new fields, you'll also need to create files for the fields under `data/presets/fields`.  Check the [presets README](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md) for details.

PR #3899, opened 2017-03-14
Update package.json to fix dist:min script on Windows

slibby:
`npm install` was failing on Windows due to the single quotes added to `dist:min` in https://github.com/openstreetmap/iD/commit/bb9b4b87251dd4f28f358dc6ef543e4f43e11b73. Removing the single quotes makes it work install again successfully, and does not cause problems on OSX builds either.


bhousel:
Thanks for this!

PR #3900, opened 2017-03-14
Add ref and operator to fire hydrant.

ToastHawaii:


ToastHawaii:
@bhousel Looks great, thanks for the good work.

Issue #2087, opened 2014-01-07
Add the layer= tag to the editor presets


bborkmiller:
It would be handy to have a dedicated field for the layer= tag, rather than have it buried down in the "All tags" section. It would be _handiest_ if it the value of the layer tag showed up in the Structure box, since it's most likely to be used in conjunction with a bridge or tunnel.

Codain:
I tag `man_made=bridge` a lot and it is tipically something missing.

It can be displayed as soon as one of those tags are set:

- `man_made=bridge`
- `bridge=*` (with a value different from 'no')
- `tunnel=*` (with a value different from 'no')
- `layer=*` (as already the case I assume)

Which also means that if I manually add one of those tags I would expect this field to be automatically shown.

bhousel:
This was done in #3911

PR #3911, opened 2017-03-18
Add type and layer subfields below structure radiobutton field

bhousel:
(for #2087)

screenshot 2017-03-18 02 11 51

There are still a few cleanups to do before merging, and I want to make the `type` field a combo.


bhousel:
This is good enough for now, merging..

PR #3916, opened 2017-03-20
Switch to `contact:`

### LINKED ISSUE #3915: Switch to `contact:`
Hello,

I think that, according to [this wiki](http://wiki.openstreetmap.org/wiki/Key:contact) iD should now use `contact:` as prefix for contact object.
At least this three item need to be switched :

- `phone`=> `contact:phone`
- `fax`=> `contact:fax`
- `website`=> `contact:website`

homersimpsons:
Switched 3 field to `contact:` schema
this fix #3915 unless I've forgot others...

### DIFF
--- a/data/presets/fields/fax.json
+++ b/data/presets/fields/fax.json
@@ -1,5 +1,5 @@
 {
-    "key": "fax",
+    "key": "contact:fax",
     "type": "tel",
     "label": "Fax",
     "universal": true,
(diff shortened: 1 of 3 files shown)


bhousel:
I do appreciate you wanting to improve the tag situation, but it would be irresponsible to deprecate the old contact tags that are in much wider use, just based on what the osm wiki says.

https://github.com/openstreetmap/iD/issues/3915#issuecomment-287945284

Issue #2914, opened 2016-01-13
Visibility of Select Halos when objects very close together


slhh:
The visibility of the select halos with the current breathe behavior is not always good.
Copied from #2879:
- Visibility of the halo is bad here:
  
  ![selection halos](https://cloud.githubusercontent.com/assets/13684947/12213216/f96290a6-b674-11e5-8865-ead8b081b6e7.PNG)
  https://openstreetmap.us/iD/master/#background=Bing&id=w111263466&map=17.52/7.00132/49.22868
  
  Visibily of the ends of the way is good, but these might be outside of the visible area in general.
  
  I believe selection halos need to be drawn on top of all objects except of selected objects.
  
  > Comment bhousel:
  > I agree that looks bad there - I'll try to adjust the draw order so that the halos appear above neighboring features.
- Due to the breathe behavior the selection halos are quite invisible for some time. This might be an issue if the user is not working slowly.
  How about making the halo more intense during the time of being thin?
  
  > Comment bhousel:
  > I did try this originally, an early version looked more like the "radar ping" that iOS Maps app uses to show the user's location - their halo starts out bright and then drops off as it spreads out. I found that although that style looked "prettier" (less "throbby") it was also much less visible during most of the animation.

slhh:
>I'll try to adjust the draw order so that the halos appear above neighboring features.

Selecting multiple neighboring parallel ways it's likely looking bad. Therefore I believe the selected objects need to be drawn on top of the halos.

bhousel:
closed by #3917

PR #3917, opened 2017-03-21
Draw selected objects after not-selected objects

bhousel:
This is a fix for #2914 to draw any selected ways in a separate group after the unselected ways.
Code still needs some cleanup

halo order

### DIFF
--- a/css/50_misc.css
+++ b/css/50_misc.css
@@ -55,13 +55,13 @@ path.casing.tag-bridge {
 }
 
 path.shadow.tag-bridge {
-    stroke-width: 22;
+    stroke-width: 24;
 }
 path.casing.tag-bridge {
     stroke-width: 16;
 }
 .low-zoom path.shadow.tag-bridge {
-    stroke-width: 14;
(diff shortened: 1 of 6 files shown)


bhousel:
Merging this as "good enough".  It's not worth spending more than a day on it, and I'll think about ways to improve it later.  The new code reorders the drawing for selected lines only.  It doesn't do anything special for the halos that are drawn around hovered lines or related parent when a vertex is selected.  In order to make this work, I needed to trigger a redraw when a user selects a singular way.

This is a tricky catch-22 problem because:
- We tend to identify DOM elements by classing them `selected`, `hovered`, `related`, etc.
- These things can't be classed until they are first drawn
- So they get drawn, then classed, then need to be redrawn in the proper order
- This is because SVG doesn't have any concept of a z-index (yet).  You just have to put them into groups to get the draw order right.
- Our map.js code really needs a way for external code to trigger full/partial/extent redraws without firing events.  Those redraw methods are internal and should be exposed.

Issue #3680, opened 2016-12-27
Teach Undo in walkthrough


bhousel:
I think at least some of the bizarre edits that we see by new users (e.g. unintentional circularization) are because users don't know realize that they can undo things.  

We should at least mention in the walkthrough that Undo is a thing.  Not only is it useful, but it makes editing less scary.

Issue #3675, opened 2016-12-24
Can't close path in tutorial


ysangkok:
I didn't make my first point the right place, and now I can't close the area, since the first point is outside the 'valid' area...

![screenshot](http://i.imgur.com/ptCKnYf.png)

bhousel:
Thanks for the report!  I think I can change the curtain to track the shape, rather than leaving it stuck that size.

Users can also click the navigation on the bottom of the screen to skip around to different parts of the tutorial if they get stuck, but that's not obvious.  Maybe I can add some text about that..

Issue #3403, opened 2016-09-06
Slim down intro_graph


tmcw:
After #3344, iD.min.js is 2mb. Not a terribly large bundle - not top priority to slim it down. But intro_graph.json takes up 25% of that size - 500kb without whitespace. Ideally we can reduce the size of intro_graph significantly.

bhousel:
Yeah, I thought about this.. https://github.com/openstreetmap/iD/issues/1336#issuecomment-166441749

I think it's mostly the river geometries that extend far beyond the parts of town where the tutorial happens.

Check out `imagery.json` - we can probably save some space there too..

bhousel:
One weird but necessary thing `rollup-plugin-json` does is generate "gunky" output when importing json.  (see https://github.com/rollup/rollup-plugin-json/issues/5)

It creates a lot of extra keys -- but only the toplevel ones..

So instead of importing json with a lot of toplevel properties:
```js
{'node1':{stuff}, 'node2':{stuff}, ... many more }
... then ...
import { default } from '../../../data/intro_graph.json';
```
...we get far less "gunk" in the generated bundle if we wrap each json file in a toplevel property:
```
{ 
  introGraphRaw: {'node1':{stuff}, 'node2':{stuff}, ... many more }
}
... then ...
import { introGraphRaw } from '../../../data/intro_graph.json';
```

This might be a good workaround for your issue https://github.com/rollup/rollup-plugin-json/issues/11 too.

bhousel:
Yeah, good suggestion @iandees.. I was actually thinking of adding some "paper streets" somewhere in a farm field and having the user to delete them in the Lines chapter.  Sound good?

iandees:
Yea, that makes sense.

bhousel:
> After #3344, iD.min.js is 2mb. Not a terribly large bundle - not top priority to slim it down.

So an update!

These days iD has grown to be an 11mb bundle, which is now "terribly large" and a "top priority to slim it down".  So I'm doing that.

Most of the bloat is from data files like the presets, imagery, community index, and other stuff.  I made a data loader that will let us fetch these things at runtime and as needed instead of having them bundled in.  

Another side benefit will be that we can even fetch some of these assets from a CDN if we want to.  JSDelivr lets us request a file with appropriate semver range - so that we could have things like the community index or name suggestion index on a separate release schedule from the core iD code.  iD will just fetch the latest compatible version (#4994!)

[...]

Issue #3121, opened 2016-05-18
Change suggested text in "name" field


aawiseman:
I've seen after a lot of mapathons that people put bad data into the name box, like a description ("house" or "dirt road") or some other duplicative or descriptive tag. Right now the placeholder/help text inside the name field is "Common name (if any)" 

I think we should change it to something like "Official name (if not known, leave blank)" or "Proper name (if known)" -- something that implies that it's a specific name and can be empty. I think "Common name" may be why people are putting things like "house" in there.

Thanks!

aawiseman:
I should add that this name data is coming from remote mappers who definitely don't know what the actual name for that thing is.

pnorman:
Name isn't for the official name of something. The problem is that "house" and similar aren't names at all.

bhousel:
I also noticed that in the walkthrough we make users add names to everything, even the playground (we probably shouldn't).  And the walkthrough uses overly simple language, like "Add a name".   I think we can keep this simple, but add another sentence there about names.

bhousel:
BTW, this was discussed in #1417.  Because this is a very common new user misconception, adjusting the walkthrough makes sense.

aawiseman:
@pnorman Right, I'm just trying to make it seem more like a specific name than a description. Maybe "proper name" "actual name" ? 

@bhousel That's interesting about the walkthrough, I wonder if that's part of it.

PR #3921, opened 2017-03-22
Walkthrough updates

bhousel:
I'm going to work over the next few ~~days~~ weeks to close a bunch of issues with our walkthrough.  This will involve some bug fixing, but also I'll add some new content to teach the new right-click menu and address some of the more common OSM newbie errors..

see: [issues tagged walkthrough-help](https://github.com/openstreetmap/iD/issues?q=is%3Aopen+is%3Aissue+label%3Awalkthrough-help)

- [x] #2381
- [x] #2386
- [x] #3067
- [x] #3068
- [x] #3085
- [x] #3121 
- [x] #3403
- [x] #3675
- [x] #3680
- [x] #3925
- [x] #3936


iandees:
Some feedback as I'm going through:

- The walkthrough asks you to "Select road from …" but the entry in the presets menu is "road features" because it's a group of presets. That might be confusing.
- The last step before it asks you to practice with more lines says to name the road, but the attributes pane isn't focusable because the "fog of tutorial" blocks it out.

bhousel:
> The walkthrough asks you to "Select road from …" but the entry in the presets menu is "road features" because it's a group of presets. That might be confusing.

Good point.. I did notice that some steps of the walkthrough we token-replace the name into the message (e.g. "Choose {name}") and other places it's part of the translation string itself (e.g. "Choose Cafe").  I'm going to switch them all over to doing the former.

> The last step before it asks you to practice with more lines says to name the road, but the attributes pane isn't focusable because the "fog of tutorial" blocks it out.

Thanks, just fixed it in b923a6d 👍

bhousel:
This is coming along really well and almost finished.. A few more things:

- [ ] Adjust timings (some things go too slow or fast)
- [ ] For playground step, make sure user really added Description field and not some other field
- [ ] Use token replace for preset/category names
- [ ] Try to shrink graph some more by renumbering ids
- [ ] More localization of names, ideally _all_ the names
- [ ] Adjust buildings tooltip illustration (it's not terrible but I don't love it)
- [ ] Teach realigning roads by moving nodes, adding midpoints (this is #2381)
- [ ] Teach deleting roads too (e.g. create some paper street development and have user delete it)

iandees:
One more thing to consider is to review the user testing report that I did with CUTGroup in Chicago a few years ago. These people were newbies to iD and went through the walkthrough.

http://www.smartchicagocollaborative.org/cutgroup-6-openstreetmap-editor/

bhousel:
Ran through this again today with a few people new to editing.. Here are my notes:

Jargon:
- [ ] People don't like the word "features"
- [ ] "main pane" as in "The main pane shows us the feature's attributes.." (Navigation Chapter)
- [ ] People don't like the word "nodes" (Areas Chapter)
- [ ] We say "Feature Editor" but sidebar says "Edit Feature"
- [ ] Define "presets" - or don't say "explore the presets" (play step, Areas Chapter)

Confusing:
- [ ] Teach zoom in/out
- [ ] Play step should find a way to remove the curtain before instructing the user to "Try moving the map and clicking..." or make it clearer that nothing will happen until they click OK.
- [ ] In point placement step, make sure to say that the mouse needs to be positioned where the point should go.. It's not _quite_ as simple as "hit spacebar to place the point" (Points Chapter)
- [ ] Teach that you can finish an area by clicking again on the last point placed?
- [ ] Be less strict about where the endpoint can go, when updating the line shape (Lines Chapter)

Bugs:
- [ ] In Buildings step, Adjust scrollTop so that "Building Features" is not all the way at the bottom
- [ ] Point hit target is still too big, causing overeager point snapping. We should not extend the hit target beyond the rendered points.  This causes problems when tracing areas/buildings.
- [ ] Undo while in draw-line mode leaves the `wTemp` temporary end segment.

lukasmartinelli:
I can already edit/move stuff around in Navigation step (probably ok?).

q5gnyk2oom

You can move features out of the focus box and then no longer select it.

hvxilvsnsb

After the Navigation step I should click on Points. If I instead click on OK and then start dragging points/polys around - I am stuck without instructions until I click on the Points step again.

rqobillg19

bhousel:
Thanks for your feedback @lukasmartinelli !
> I can already edit/move stuff around in Navigation step (probably ok?).

Yes this is ok, we reset the editing graph to known states at certain points in the tutorial - this is just in case the user decides to delete the thing right before we ask them to select it.

> You can move features out of the focus box and then no longer select it.

Good catch, I'll improve this so the box tracks the item as the user drags it.

> After the Navigation step I should click on Points. If I instead click on OK and then start dragging points/polys around - I am stuck without instructions until I click on the Points step again.

Yes this confused some of my other testers too.  I'll need to adjust the messaging a bit to make sure users don't forget that they need to click the next chapter to continue.

PolyglotOpenstreetmap:
I'm going through the tutorial for the first time of my life, a little bit more and I'll be weaned of JOSM :-)

I'm having a problem in the lines step. I can't seem to add a node on Flower street to start adding a street going north. It's also not 'highlighted'.

Oh, now I notice that this is because I hadn't been following the instructions to the letter.

When deleting 12th Avenue, it reappeared (I used del key, maybe that wasn't right)

Moving on to buildings

I noticed that I can also finish a building shape by clicking enter or esc. space adds another node. Of course I accidentally pressed space with the mouse in the wrong place, then went on to tag it as a building. Then I got to the explanation of how to square the building, but that failed. If I had been a new user that would probably have confused me. Admittedly I don't have any suggestions how it would be possible to get back on track, when things start going south like that.

Anyway, being able to betatest like that is nifty feature (of github?)

I think what I would like to see is a dedicated way to draw buildings (accessible using 4 maybe?).Clicking on three corners would create a rectangular shape and then the mapper can move on. Simply tag them building=yes automatically.

The other thing I'd like is to use less nodes for circular buildings. Couldn't that number be made dependent on the size/surface of the building? 9-11 nodes per hut is more than enough to show that they are round shapes.

Maybe I'll propose to mentor a GSoC project to accomplish that next year, but it would be better if it would appear in iD sooner than that. HOT validators are really overwhelmed with the flood of buildings that are mapped poorly. My pet peeve is that I can not press 's' to  square the shape, while drawing it. I'm afraid that this is inherent to iD though, so it would be hard to solve. 
Pressing 's' starts searching in the tags.
So I'm still working on the geometry, but iD all of a sudden moves to tagging the shape. This is probably due to me being used to JOSM where keys pressed are always shortcut keys performing some action and tags are either added in the dedicated 'pane' or looking through the presets, after pressing F3.

Polyglot

M1dgard:
This is great! So much better. \o/

Mostly it seems to be pretty fool-proof, but not in all situations:

* [ ] The most serious one: adding a name before you're asked to causes your beloved cafe to disappear.  
Adding a name before you're asked to causes the point to be deleted

* [ ] No, it's not.  
Changing the type of the Town Hall doesn't cause any reaction. Not disastrous but not very clean either  
Changing the type of the Town Hall doesn't cause any reaction. Not disastrous but not very clean either


* [ ] Selecting the building instead of the cafe.  
It's possible to (accidentally) select the building instead of the cafe, leaving you in the Feature Editor with the wrong feature

* [ ] First creating a non-intersection, and then an intersection is still possible. Extremely silly shapes are possible too, maybe you would like to catch this. (Both things depicted.)  
A completely messed up road, it has non-intersections with the other road and a ridiculous shape

bhousel:
Thanks for the feedback everyone!  Over the past few days I've added a bunch of extra code to try to handle all of the situations where a user might get stuck.  This is a _much_ bigger challenge than with the old walkthrough code!  For example, the trainee can start adding some information to a feature, then undo all the way back to where the feature is an untagged area, which brings back the "choose preset" screen.  The previous walkthrough disabled undo in order to prevent situations like this, but I like better being able to teach about undo, and having the walkthrough work mostly the same way "real" editing works.

There are still a few things that need cleanup, but I'm getting pretty close to merging this so that the translators can get to work.

If you want to share this new walkthrough with anyone, please do so, and encourage them to post their feedback here on this PR, or reach out to me directly on OSMUS Slack or IRC (see [README](https://github.com/openstreetmap/iD#participate) for how to get on these).

Here is a tweet that you can share:
https://twitter.com/bhousel/status/853227516676501504

mmdolbow:
Comments/feedback:
- In "Navigation", the last dialog, it says "Try moving the map and clicking on some other features...." but you can't move the map or click at that point in the walkthrough. NBD because when you click "OK", you can. This repeats at the end of each section. Maybe a small "Click OK to practice...." or a tooltip?
- When finishing a line, clicking again on the last node is suggested. Double-clicking also works I believe. I'm not sure which is better to suggest, but I usually do the double-click thing, because it's how I finish an area.

Bug? After clicking the checkbox to clear the attributes dialog for a building, I got stuck in a screen like shown below. Nothing worked at that point - I was getting a cursor change when hovering over buttons, making them appear clickable, but clicking didn't produce anything. It's almost like I got stuck between the transitions.
id-walkthrough-stuck

M1dgard:
I remember seeing a similar situation to the one @mmdolbow describes ("stuck in a screen") but couldn't reproduce it.

samanpwbb:
This tutorial is a huge improvement. With a track pad, "right click" isn't meaningful:
 
screen shot 2017-04-15 at 2 23 41 pm

EIther say "right click or control + click" or do feature detection and cater the text to the user's setup.

bhousel:
> Bug? After clicking the checkbox to clear the attributes dialog for a building, I got stuck in a screen like shown below.

Thanks @mmdolbow!  I fixed that today.

bhousel:
> This tutorial is a huge improvement. With a track pad, "right click" isn't meaningful:
> Either say "right click or control + click" or do feature detection and cater the text to the user's setup.

Good suggestion, @samanpwbb 👍   I added some additional words to the welcome chapter, and a small interactive button-pressing task to make sure the user can left and right click.

bhousel:
Just a heads up, I'm planning to merge this today! 🎉 

There are a few bugs in the snapping and undo behaviors, but I can fix those after it merges.
If there are any more last minute suggestions for adjusting the language before we unleash the translators on all this new content, let me know!

magol:
@bhousel 
Great job to create this walkthrough. I'm now working to localize it, and I have found some issues

- When I select a node, It do not view the localized street name.
- The adress format should be localized to. In Sweden, the attribute "state" has no meaning and the fields is in the wrong ordet. 
- In addition, I have noted that it is somewhat inconsistent with period in the end of sentence of some instructions.

image

bhousel:
Thanks for the feedback @magol!  I moved your comments to a new issue #3976

althio:
Is there some robust link we can use to refer to the walkthrough, a 'start the walkthrough' url ?

bhousel:
> Is there some robust link we can use to refer to the walkthrough, a 'start the walkthrough' url ?

Not currently - users are prompted to start it their first time using iD, or they can start it by opening the Help screen.  But we don't have a url parameter to autostart the walkthrough, if that's what you mean?

althio:
> we don't have a url parameter to autostart the walkthrough, if that's
what you mean?

Yes. That would be handy.

bhousel:
>>Not currently - users are prompted to start it their first time using iD, or they can start it by opening the Help screen and clicking a button there. But we don't have a url parameter to autostart the walkthrough, if that's what you mean?
>
> Yes. That would be handy.

ok added #4111 for this

PR #3923, opened 2017-03-23
Support Arabic vowel signs, include more languages in generic RTL fix

mapmeld:
This takes the Arabic-specific regex and fix which we implemented earlier, and extends it to two other right-to-left scripts (Hebrew, and Thaana from the Maldives)

These two languages don't need special text-shaping like we did for Arabic, just right-to-left reversal to handle the Chrome bug, and keeping vowel signs together with the letter that they modify (sort of like ```N + ̃  = Ñ``` - reversing the string should not change which letter gets the tilde)

The Arabic code also gets the fix for vowel signs


bhousel:
Cool, thanks @mapmeld!

PR #3958, opened 2017-04-11
Fix misleading connectivity error

### LINKED ISSUE #3952: Wrong tooltip message for Merge
This is a weird message because the lines are in fact connected.  I think what the tooltip means to say is something like "These lines can't be joined because they aren't connected end-to-end." 

The message is string key `operations.merge.not_adjacent`

screenshot 2017-04-09 21 14 19

slibby:
fixes #3952.

I think it's appropriate to fix this in /en.json and /en-GB.json, and it will update Transifex to request new translations for the other languages? I read the contributing.md but didn't see how that was supposed to be handled, let me know if this is wrong.

### DIFF
--- a/dist/locales/en-GB.json
+++ b/dist/locales/en-GB.json
@@ -138,7 +138,7 @@
                 "key": "C",
                 "annotation": "Merged {n} features.",
                 "not_eligible": "These features can't be merged.",
-                "not_adjacent": "These features can't be merged because they aren't connected.",
+                "not_adjacent": "These features can't be merged because their endpoints aren't connected.",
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks!  Sorry that the translations aren't explained very clearly..  Can you do these steps instead:

1. Edit the string in `core.yaml` - This is the file that Transifex will read from.
2. Do `npm run build` - This will generate the `en.json` file.

(You don't need to worry about `en_GB` - this file we will pick up from Transifex)

slibby:
closing and will re-open when I've done it properly 🥇

PR #3961, opened 2017-04-11
Update licensing info

waldyrious:
According to @bhousel in #3207, the mention of BSD "is throwback to when the iD source code included our own custom d3 bundle, which we don't do anymore."

Also, add extension to the license file, to match the other standard text files at the root of the repo.


bhousel:
Thank you!

PR #3964, opened 2017-04-13
Updated Merge Messaging + CONTRIBUTING.md

slibby:
Hopefully this is the corrected version of https://github.com/openstreetmap/iD/pull/3958

I also updated the CONTRIBUTING.md to explain how to make these updates to `core.yml` in the future, as I think it was missing that info.

I'm not sure why it moved `place/city` beneath `place/farm` but I think it was part of the npm build script?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,7 +90,7 @@ These are separate translations for uniformity reasons and because some language
  may translate "type" differently in "type of aeroway" and "type of amenity", for
  example.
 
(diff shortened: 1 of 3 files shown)


pnorman:
> I'm not sure why it moved place/city beneath place/farm but I think it was part of the npm build script?

That's a change in the serialization of the json, but not of the json, so it's just diff noise.

bhousel:
Thanks for this!

PR #3995, opened 2017-04-27
Added fields for address in Ukraine

Andygol:
As Ukrainian OSM Community don't fill in `addr:city` for each building, It is better to decrease amount of fields for adding addresses.
>Теґ addr:city=* в Україні зазвичай не використовується.

☝️ https://wiki.openstreetmap.org/wiki/Uk:Addresses


bhousel:
Looks great, thanks @Andygol!

PR #4026, opened 2017-05-08
Change ordering in highway fields presets

willemarcel:
The `structure` and `access` fields takes too much space in the screen and are less used than `surface` and `lanes`, so I change the ordering to show more fields on the screen without the need of scrolling. 

**Before:**
image

**After:**
image
image


bhousel:
Thanks @willemarcel, I also like this arrangement better.. 👍

PR #4032, opened 2017-05-10
Omit global search UI when no geocoder service is available

mojodna:
After `delete iD.services.geocoder`.


bhousel:
thanks, good catch!

PR #4052, opened 2017-05-17
Adding "football" terms to soccer pitch

JamesKingdom:
Hi,
We like to call it football instead of soccer, so I have added an extra search term to make this possible.
Thanks :+1:


bhousel:
Looks good, thanks!  ⚽️

PR #4058, opened 2017-05-20
Create electronic repair shop preset

JamesKingdom:


bhousel:
It looks like you didn't check in the `craft/electronics_repair.json` file.  Can you do that, and then squash all the commits down to a single commit to get rid of the accidental change to the development server port?

PR #4084, opened 2017-06-04
update Taiwan's imagery sources

happyman:
1. update NLSC's tms url 
2. add Taiwan historical map with gpx traces.


bhousel:
Hi, @happyman  thanks for the report. iD's background imagery sources come from the [editor-layer-index](https://github.com/osmlab/editor-layer-index) project - can you make the pull request over in that project, and it will get accepted into iD automatically when we release a new version?  Thanks!

happyman:
@bhousel thank you!

PR #4109, opened 2017-06-17
Adding pavement term to sidewalk

JamesKingdom:
A sidewalk is commonly referred to as a "pavement" here in the UK.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -2491,6 +2491,7 @@ en:
       footway/sidewalk:
         # 'highway=footway, footway=sidewalk'
         name: Sidewalk
+        # 'terms: pavement'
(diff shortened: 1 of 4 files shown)


pnorman:
I think this is fairly specific to British English, so we could add it to the en_GB translation instead.

JamesKingdom:
@pnorman That is a very good idea. :+1:

bhousel:
I made the changes on Transifex here:
https://www.transifex.com/openstreetmap/id-editor/translate/#en_GB/presets/18729418
https://www.transifex.com/openstreetmap/id-editor/translate/#en_GB/presets/18729419

@JamesKingdom, you can sign up for Transifex and review or change the `en_GB` translations if needed, to make sure that all the terms we use in iD are easily understood.

PR #4116, opened 2017-06-24
Add terms to power generator

boothym:
Make it easier for people to add wind turbines etc. - maybe better to create a proper preset for it, but getting people to the power generator preset is better than nothing.

### DIFF
--- a/data/presets/presets/power/generator.json
+++ b/data/presets/presets/power/generator.json
@@ -11,6 +11,12 @@
         "vertex",
         "area"
     ],
+    "terms": {
+        "hydro",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks!  I agree it might be nice to create presets for the more common generator... sources?  methods?  the tagging for these things is a bit wonkish.

boothym:
I think the combination of tags is more useful than presets for sources, methods etc. 
For example you could have power=generator, generator:source=wind, generator:method=wind_turbine and generator:output:electricity=yes appear as the wind turbine preset.

Issue #3761, opened 2017-01-14
View history link


boothym:
When editing a feature there is a link at the bottom to "View on opensteetmap.org" - how about adding a link to the history as well? Would save a click!

bhousel:
Per https://github.com/openstreetmap/iD/issues/2273#issuecomment-233331887 ,  I kind of want to add some information about the history of a way to the "secret" <kbd>⌘I</kbd> object inspector.  If we put a link in here does that work for you?  I don't want it to be a permanent UI feature on the sidebar, but the inspector box is a good place for it.

boothym:
![history](https://cloud.githubusercontent.com/assets/16965772/22039816/b0ac716e-dcf8-11e6-9b07-116c8cb03881.jpg)
I was thinking about something like this. Generally the use case is where you can quickly look at the tags between the last two versions, and maybe go back to when it was first created. Not sure you'd be able to do that in the secret menu.

slhh:
The history view of openstreetmap.org isn't really user friendly.
The history view of JOSM is better, but we can't link JOSM.
Maybe we can link the following OSM history viewer, or even integrate it because it seems to be JavaScript:
https://github.com/PeWu/osm-history
Example: https://pewu.github.io/osm-history/#/way/37519344
This one looks very nice, at least for ways and nodes.
For tag history OSM deep history is very useful too, but it is already linked by above OSM history viewer.
Example: https://osmlab.github.io/osm-deep-history/#/way/48005102

Issue #2492, opened 2014-12-23
Show date of aerial data


hajo4:
When mapping in areas where demolition & construction is taking place,
and the map is different from the image, it would be useful to see
the date of aerial images shown.

cgiovando:
:+1:   Would be nice to be able to pass some imagery metadata automatically, especially when adding custom background imagery (WMS/TMS from HOT-OSM Tasking Manager) and automatically add a source=* and source:date=YYYY-MM-DD to whatever element is created or modified with that specific imagery layer turned on.

jfirebaugh:
Even if the image tile comes with response headers indicating the vintage, we can't make use of them -- there's no way to get response headers for an `<img>` element. The only way to do this is to make a _separate_ request: either an ajax request for the same image (doubling the transfer cost for background imagery) or to another API that provides vintage in another format (hopefully for all tiles in a given area, so that we can make a single request).

pnorman:
Based on experience, the metadata for the middle is sufficient. The per-tile info is not perfectly accurate at imagery boundaries anyways.

We should also make use of editor-imagery-index supplied dates.

aawiseman:
Hi @bhousel I talked with @gmaclennan about this, he has worked on something similar, here's what he had to say: 

> I had some code from a Leaflet Bing Layer plugin that I created for parsing metadata from Bing imagery. That code is here:
> 
> https://github.com/gmaclennan/leaflet-bing-layer/blob/gh-pages/index.js#L160-L246
> 
> I mentioned this because from what somebody said at Sotmus I thought that iD did not currently display imagery date for Bing, but I see that it does actually display some metadata which you can see here: https://github.com/openstreetmap/iD/blob/af776501d3e8445c94d8d5f4198df5a06ccf1ba2/modules/renderer/background_source.js#L89-L133
> 
> It looks like iD is not currently reading metadata specific to the current center point, which my Leaflet plugin does. I'm not sure how this data is different, or whether it is useful - it's a while since I worked on this. The relevant docs from Bing are linked in my code comments.

bhousel:
Nice!  Thanks @aawiseman for pointing me at @gmaclennan's code...  We could definitely perform that jsonp query to fetch the bing metadata.

dalekunce:
👍 for this feature.

bhousel:
Coming soon!
Added the code today to get the vintage from Bing's REST service, and use whatever is in editor-layer-index for all other sources..

[...]

Issue #2273, opened 2014-06-24
Display "last edited by" for objects in iD


Mashin6:
I think it would help to make editing in iD better to have a way how to browse through history of the objects. Now the only way is to follow the link to View on osm.org and then select history, which is not always easy to understand.

I would imagine that upon choosing history for selected object, iD would show list of previous versions in the left panel. Each version item would expand on click to reveal tags from given time in the past.

Maybe too ambitious would be to change the shape of the object in real time to reflect how it looked in given version.

bhousel:
I don't want to add a full history browser, but there are some situations where seeing "last edited by" for a selected feature would be useful.  We can sneak this into the "secret" `Ctrl-I` infobox.

slhh:
@bhousel "Last edited by" seems to be useless information without knowing what has been changed. A deteted or added node? A tag change? It doesn't help to find out if a specific user has changed the object because someone else might have added/deleted a node later. Which usecase do you have?

The original request is useful. The community would benefit from encouraging users to view the history, at least in regions with an active community.

bhousel:
> Which usecase do you have?

If a feature doesn't appear on aerial imagery, and it was "last edited by" bot-mode 4 years ago, it's probably garbage.  If it was "last edited by" some actual user 4 months ago, it's probably a real feature that is just newer than the imagery.  There are a handful of active users around where I live, so if I recognize that it's something added by one of them, I know it's fine.

> The community would benefit from encouraging users to view the history, at least in regions with an active community.

Yes, so we can put a link to the history viewer on openstreetmap.org for people that want to see the full history.

bhousel:
Sneak preview:

![history](https://user-images.githubusercontent.com/38784/27759878-c27b7e1c-5e08-11e7-8510-6623c977d081.gif)

PR #4121, opened 2017-06-29
Info box refactor to panels

bhousel:
I've been working on this for a few days and it's coming along well.  We had the infobox before <kbd>⌘I</kbd>, but this refactor wraps it into a panel and extends it so we can add more of them for other purposes.

The best thing about these panels is that you don't need to see them or know about them.  It's a handy way to add "pro" features to iD without complicating the UI in other ways, and they can end up as plugins eventually.  I'm going to use this tactic to close out a handful of other old issues over the next few days - watch this PR for more.

infobox

- [x] #2183
- [x] #2273
- [x] #2492 
- [x] #2515 
- [x] #3761


bhousel:
This is good enough for now.. merging!  :shipit:

PR #4144, opened 2017-07-11
Service workers are here

kepta:
This adds a simple build step to add service workers using [workbox](https://workboxjs.org).
The workbox's documentation is pretty bad and their rollup example doesn't work.

### Notes
- iD.js is pretty huge.(~4mb). Service workers get scared seeing such huge files. We need fix this sometime. ( Code splitting would help, but `rollup` doesn't [support it yet](https://github.com/rollup/rollup/issues/372) ).
- Some of API's that iD uses can have a [runtime caching]( https://workboxjs.org/reference-docs/latest/module-workbox-runtime-caching.html).
- Need some input on how to make it work with `openstreetmap.org` website.
- Figure out how to tell the user a new version is available. Currently it just does `windows.alert`. 
- Some nice [points](https://developers.google.com/web/fundamentals/instant-and-offline/offline-ux#inform_the_user_when_the_app_is_ready_for_offline_consumption) on offline first.
- Currently to test if service worker is updating assets correctly I have setup an iD instance at https://languid-alley.surge.sh . To update it with some local changes I run `surge -p dist -d languid-alley.surge.sh`. (seems to work fine and also updates the assets properly)

### What could go wrong.
Service worker doesn't actually interfere with the actual code, but it might keep sending the stale data if something goes wrong.
- Need to make sure service worker doesn't eat up children URLs like `openstreetmap.org/iD/xyz` more on this [here](https://github.com/GoogleChrome/sw-precache#navigatefallback-string)

### Bright Side
- Huge improvements in load time for users.
- Offline support.
- We can decide on a better caching strategy for APIs.
- Would make iD more aligned with [PWA](https://developers.google.com/web/progressive-web-apps/)

### DIFF
--- a/build-sw.js
+++ b/build-sw.js
@@ -1,0 +1,15 @@
+const workboxBuild = require('workbox-build');
+
+workboxBuild
+    .generateSW({
+        maximumFileSizeToCacheInBytes: 2097152 * 3,
+        clientsClaim: true,
+        skipWaiting: true,
+        swDest: 'dist/service-worker.js',
+        globDirectory: 'dist',
+        staticFileGlobs: ['**/!(*map*)'],
+        globIgnores: ['**/service-worker.js']
+    })
+    .then(() => {
+        console.log('The production service worker has been generated.');
+    });
--- a/build-sw.js
+++ b/build-sw.js
@@ -7,7 +7,7 @@ workboxBuild
         skipWaiting: true,
         swDest: 'dist/service-worker.js',
         globDirectory: 'dist',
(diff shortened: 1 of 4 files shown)


bhousel:
Hey @kepta just checking in on this.  Do you still think this is worth doing, or better to spend time on other things? 

E.g. if service workers are kind of a non-starter with the code the way it is and rollup we can close, and keep in mind for iD v3.

bhousel:
Closing as stale.. Moving to a more modular iD would make this better.

PR #4148, opened 2017-07-12
Adds mapillary detections

### LINKED ISSUE #3772: Highlight detected objects in Mapillary images
There might be a way to do this automatically in the Mapillary viewer, but the idea is to draw some box around a detected feature.

kepta:
- I couldn't find an example of Point shape, do we need a point shape?
- The code doesn't cache the network requests and makes one whenever the user clicks an Image. I guess that is fine.
(closes #3772)


bhousel:
Screenshot for anyone who is wondering what this PR is - to highlight the detected objects in the images.

screenshot 2017-07-12 17 42 59

bhousel:
I added a few things..
* results from the `/detection` api calls are cached now
* make it so we can show many detections per image
* style the detections for better visibility
* cleanup the names

This is 🔥 now!  Thanks @kepta for getting this started!

detections

Issue #4162, opened 2017-07-19
Custom background imagery info lost


pgiraud:
When iD is launched with information about custom background imagery (for example using the tasking manager), it looks like it's impossible to switch to an other background and then switch back to the custom imagery.

Steps to reproduce:
 - open the following url http://www.openstreetmap.org/edit?editor=id&#map=18/-1.66488101692/29.2366790719&comment=%23foss4ge17%20%23hotosm-project-2608%20%23Goma%20%23DRC%20%23MissingMaps%20%23MapGive%20Source%3DGeoEye-1%2C%20WorldView-02%2C%20DigitalGlobe%2C%20NextView%2C%2022%20February%202017%20%26%2028%20February%202017&gpx=http://tasks.hotosm.org/project/2608/task/834.gpx&background=custom:http%3A%2F%2Fhiu-maps.net%2Fhot%2F1.0.0%2Fgoma-28feb2017-flipped%2F%2F%7Bz%7D%2F%7Bx%7D%2F%7By%7D
 - map should be centered on Goma DRC with custom imagery loaded,
 - switch to any other background imagery,
 - switch back on custom imagery.

Result:
 - on Chrome nothing happens and background is black,
 - on Firefox, the user is asked to provide a URL for custom imagery.

Sorry if it has already been reported.

pgiraud:
This doesn't happen if the url is manually typed in the custom imagery URL input field.

bhousel:
Thanks for the report @pgiraud, I'll try to get this fixed quickly.

bhousel:
closed by #4165

PR #4165, opened 2017-07-20
Remember the chosen custom background when set by url

pgiraud:
Here's a quick workaround for #4162.
@bhousel I'm not sure if it's the best way to fix the issue but it seem to work.

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -248,7 +248,9 @@ export function rendererBackground(context) {
         }
 
         if (chosen && chosen.indexOf('custom:') === 0) {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks so much @pgiraud - this is pretty good!

Your quick fix worked because it was creating a Custom imagery layer and setting the `localStorage` template - then the _real_ Custom layer would get created later when the Background ui was added and that template would get picked.

I added a commit after to cleanup the ownership of these objects.  Now all the layers, including Custom and None are 'owned' by `rendererBackground()`, and the `uiBackground()` does not create a Custom layer, it just finds the existing one by id.  The new code is a little easier to reason about.

Issue #1141, opened 2013-03-26
WMS support


tmcw:
via http://www.microformats.dk/2013/03/22/se-et-preview-af-det-nye-id-openstreetmap-redigeringsv%C3%A6rkt%C3%B8j/ + http://www.microformats.dk/2013/03/14/sadan-kalder-du-geodatastyrelsens-wms-tjenester-fra-josm/

Question is really whether there will be many WMS providers that support EPSG:900913

bhousel:
Let's support the easy WMSs via [whoots-js](https://github.com/mapbox/whoots-js). 

TODO:
- [ ] add whoots-js library to dependencies
- [ ] edit `update_imagery.js` to 
  - [ ] include WMSs that support EPSG:3857
  - [ ] with a tile url including a replacement token like `bbox={bbox-epsg-3857}` - [like how mapbox-gl-js does it](https://www.mapbox.com/mapbox-gl-js/example/wms/)
- [ ] add whoots-js call to `tile_source.js` to replace the token
- [ ] make sure these WMS sources that support EPSG:3857 are listed that way in `editor-layer-index`.  There is a [property for available projections](https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L122), but most WMS sources don't have it filled in.

simonpoole:
@bhousel I suspect you are opening a can of worms. While I have now and then seen WMS servers supporting EPSG:3857, it is in my experience, the absolute exception (hey for once I agree with @tmcw). Which implies that you are not going to get around at least EPSG:4326 support.

pnorman:
> TODO:

You might need to handle deduplication for sources which have both TMS and WMS

> While I have now and then seen WMS servers supporting EPSG:3857, it is in my experience, the absolute exception

There's about 15 or so in editor-layer-index which have 3857 or 900913 listed. The JOSM list has more, with about 20 900913 and 30 3857 (and duplication between them I'm sure).

As a ballpark guess I'd estimate 20 new sources, since some of those will have TMS equivalents.

geowOSM:
> Which implies that you are not going to get around at least EPSG:4326 support (the request that kicked off reopening this issue, is BTW, exactly such a case).

Exactly, would be nice if ID will support EPSG:4326 in the short to medium term and best flag Bavaria DOP80.

bhousel:
> Exactly, would be nice if ID will support EPSG:4326 in the short to medium term and best flag Bavaria DOP80.

[...]

PR #4166, opened 2017-07-20
Imagery Offset Database

### LINKED ISSUE #1124: Use the upcoming imagery offset database
Manually adjusting every layer is hard, non-obvoius and usually unneeded task which duplicates the efforts of previous mappers in an area. In a week or two I'm announcing the imagery offset server which allows users to determine an offset for each area once, and then just use the shared value.

See [the description](http://wiki.openstreetmap.org/wiki/Imagery_Offset_Database) and [API documentation](http://wiki.openstreetmap.org/wiki/Imagery_Offset_Database/API). There is also [a map](http://offsets.textual.ru/map) and [a list](http://offsets.textual.ru/all) (refreshes every hour).

Also I'd like you to check [the alpha-version of a JOSM plugin](https://www.dropbox.com/s/abqlwa57pkrck81/imagery_offset_db.jar) (it should be put into `~/.josm/plugins`). It shows that the user interface mostly consists of informational messages and a dialog for choosing an offset. The latter is the most important part: its design should give users enough information to correctly choose an offset and rid them of doing the same work twice.

I plan to release beta on Monday or Tuesday, and before that I'm open to any suggestions, even those which could reqiuire structural changes to the database and the plugin.

kepta:
@bhousel would need help in figuring out certain things.

- What do we keep the padding of imagery points ?
- Where should we invoke the check for imagery offset ? Should it happen automatically or be shown in the imagery adjustment UI, like click to apply saved imagery offset.

(closes #1124 )

### DIFF
--- a/modules/services/imagery_offset.js
+++ b/modules/services/imagery_offset.js
@@ -23,18 +23,6 @@ export default {
     offsetCache = rbush();
   },
 
-  countryCode: function(location, callback) {
-    this.reverse(location, function(err, result) {
-      if (err) {
-        return callback(err);
-      } else if (result.address) {
-        return callback(null, result.address.country_code);
-      } else {
-        return callback("Unable to geocode", null);
-      }
-    });
-  },
-
   search: function(location, callback) {
     var paddedLocation = geoExtent(location).padByMeters(500);
     var cached = offsetCache.search({
--- a/modules/services/imagery_offset.js
+++ b/modules/services/imagery_offset.js
@@ -22,8 +22,95 @@ export default {
         inflight = {};
         offsetCache = rbush();
     },
+    getImageryID: function(url, imageryType) {
+        if (url == null) return;
+        console.log('url=', url)
+        debugger;
+        url = url.toLowerCase();
+        if (imageryType === 'bing' || url.indexOf('tiles.virtualearth.net') > -1)
+            return 'bing';
+        // if (url.indexOf("scanex_irs") > )
+        if (
+            imageryType === 'TMS' &&
+            url.match(/.+tiles\.mapbox\.com\/v[3-9]\/openstreetmap\.map.*/)
+        )
+            return 'mapbox';
 
-    search: function(location, callback) {
+        // does iD support WMS? if yes, how to detect
+        var isWMS = false;
+
+        // Remove protocol
+        var i = url.indexOf('://');
+        url = url.substring(i + 3);
+
+        var query = '';
+        // Split URL into address and query string
+        var questionMarkIndex = url.indexOf('?');
+        if (questionMarkIndex > 0) {
+            query = url.slice(questionMarkIndex);
+            url = url.slice(0, questionMarkIndex);
+        }
+
+        var removeWMSParams = [
+            'srs',
+            'width',
+            'height',
+            'bbox',
+            'service',
+            'request',
+            'version',
+            'format',
+            'styles',
+            'transparent'
+        ];
+
+        var qparams = {};
+        var qparamsStr = query.length > 1 ? query.slice(1).split('&') : '';
+
+        qparamsStr.forEach(function(param) {
+            var kv = param.split('=');
+            console.log(kv);
+            kv[0] = kv[0].toLowerCase();
+            // WMS: if this is WMS, remove all parameters except map and layers
+            if (isWMS && removeWMSParams.indexOf(kv[0]) > -1) {
+                return;
+            }
+            // TMS: skip parameters with variable values and Mapbox's access token
+            if (
+                (kv.length > 1 &&
+                    kv[1].indexOf('{') >= 0 &&
+                    kv[1].indexOf('}') > 0) ||
+                kv[0] === 'access_token'
+            ) {
+                return;
+            }
+            console.log('here')
+            qparams[kv[0].toLowerCase()] = kv.length > 1 ? kv[1] : null;
+        });
+
+        var sb = '';
+        Object.keys(qparams).forEach(function(qk) {
+            if (sb.length > 0) sb += '&';
+            else if (query.length > 0) sb += '?';
+            sb += qk + '=' + qparams[qk];
+        });
+
+        // TMS: remove /{zoom} and /{y}.png parts
+        url = url.replace(/\/\{[^}]+\}(?:\.\w+)?/g, '');
+
+        // TMS: remove variable parts
+        url = url.replace(/\{[^}]+\}/g, '');
+
+        while (url.indexOf('..') > -1) {
+            url = url.replace('..', '.');
+        }
+
+        if (url.startsWith('.')) url = url.substring(1);
+
+        return url + query;
+    },
+    search: function(location, url, callback) {
+        console.log(this.getImageryID(url))
         var cached = offsetCache.search({
             minX: location[0],
             minY: location[1],
--- a/modules/services/imagery_offset.js
+++ b/modules/services/imagery_offset.js
@@ -1,71 +1,78 @@
-import * as d3 from "d3";
-import _ from "lodash";
-import rbush from "rbush";
-import { geoExtent } from "../geo/index";
-import { utilQsString } from "../util/index";
+import * as d3 from 'd3';
+import _ from 'lodash';
+import rbush from 'rbush';
+import { geoExtent } from '../geo/index';
+import { utilQsString } from '../util/index';
 
-var apibase = "http://offsets.textual.ru/get?",
-  inflight = {},
-  offsetCache;
+var apibase = 'http://offsets.textual.ru/get?',
+    inflight = {},
+    offsetCache;
 
 export default {
-  init: function() {
-    inflight = {};
-    offsetCache = rbush();
-    window.offsetCache = offsetCache;
-  },
+    init: function() {
+        inflight = {};
+        offsetCache = rbush();
+        window.offsetCache = offsetCache;
+    },
 
-  reset: function() {
-    _.forEach(inflight, function(req) {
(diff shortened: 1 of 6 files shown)


bhousel:
This is a great start - just a reminder to use [4 space indent](https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#javascript), like we do everywhere else in iD.  (we should probably have an [`.editorconfig`](http://editorconfig.org/) file in our project root that enforces this.)

> What do we keep the padding of imagery points ?

Ideally we don't want to pad the lookup and store in a location cache - it would be better to cache the results of each tile url that we lookup.  I'm not sure whether the service could handle the load of all iD users asking for offset for every tile.  What do you think @Zverik ?

> Where should we invoke the check for imagery offset ? Should it happen automatically or be shown in the imagery adjustment UI, like click to apply saved imagery offset.

Ideally automatically.  We store the [user supplied offset here](https://github.com/openstreetmap/iD/blob/master/modules/renderer/background_source.js#L29) in `background_source`.  We could add a bool variable there to say whether the offset is automatically fetched, and we could add a corresponding checkbox on the Background ui screen.

bhousel:
@kepta As a next step, can you add a test here:  https://github.com/openstreetmap/iD/tree/master/test/spec/services

Zverik:
Thanks for the pull request. Automatically checking the offset for every tile is an overkill and will quickly put the server offline. I suggest doing that when opening the editor, and after that when a user scrolls the map more than a kilometer (at least 500 m) off the last offset checking point.

kepta:
I have added some simple tests.
Next actions would be, to change the way search works. Currently the backend gives the whole list of imagery's without the provided radius. We would like to cache all of it but return back only the imagery offset matching with the active imagery. 
The backend gives the data_url of the imagery, I guess that can be used to map the active imagery to the one in the cache?.

bhousel:
> My bad forgot to check the indentation, I have fixed it in the latest commit. @bhousel should I rebase and squash the commits, to make it look better?

No problem.  I don't care much about how the commit history looks.

> I have added some simple tests and also added a padding of 1km to each imagery pos as per @Zverik 's recommendation. I am still not sure about the radius to search in.

You could probably use a large radius, since we're caching the results  (so that would mean fewer calls) - and the results packets are small, and there probably aren't too many.

I guess try it with a few different search radii and see?

kepta:
> I guess try it with a few different search radii and see?
That makes sense, I will go for an increased radius. Any thoughts on how to match the active imagery with the one in cache?

bhousel:
> That makes sense, I will go for an increased radius. @bhousel Any thoughts on how to match the active imagery with the one in cache?

Yes, I think we can do this similarly to how I sample [imagery vintage in the Background panel](https://github.com/openstreetmap/iD/blob/master/modules/ui/panels/background.js#L65-L81).

The background layer is a bunch of `` elements, and one of them which is closest to the center of the screen will be classed with `.tile-center`.

Here is some code that you can stuff into [the background panel](https://github.com/openstreetmap/iD/blob/master/modules/ui/panels/background.js).  There is already a `getVintage` function, so this would be how you might write a `getOffset` function.

```js
var tile = d3.select('.layer-background img.tile-center');   // tile near viewport center
if (tile.empty()) return;

var d = tile.datum();   // the data for that tile
if (!d || d.length < 4) return;

var zoom = d[2];
var url = d[3];
var center = context.map().center();

// service.getOffset will work somewhat like our nominatim service reverse():
// 1. use magic to turn the url, center, and zoom level into an identifier, per
//     https://wiki.openstreetmap.org/wiki/Imagery_Offset_Database/API#TMS
// 2. query the offset database with those parameters (or return cached result)
// 3. cache result and call the callback
service.getOffset(url, center, zoom, function(err, result) {
   // console.log something?
   // or update a div someplace with the numbers
});
```

modules/services/imagery_offset.js
```diff
@@ -0,0 +1,165 @@
...
+    },
+
+    reset: function() {
+        _.forEach(inflight, function(req) {
+            req.abort();
+        });
+        inflight = {};
+        offsetCache = rbush();
+    },
+    getImageryID: function(url, imageryType) {
```
kepta: @Zverik could I get your review for this piece of code, I tried to convert the Java [code](https://trac.openstreetmap.org/browser/subversion/applications/editors/josm/plugins/imagery_offset_db/src/iodb/ImageryIdGenerator.java) you wrote in Javascript.

If you could also point me to some unit test cases for this function in your code base ..

kepta:
@bhousel do we support `WMS` type of imageries in iD, since I couldn't find one in `imagery.json`?

bhousel:
> do we support WMS type of imageries in iD, since I couldn't find one in imagery.json?

Not yet.. that's #1141

modules/services/imagery_offset.js
```diff
@@ -22,8 +22,95 @@ export default {
...
+        if (url == null) return;
+        console.log('url=', url)
+        debugger;
+        url = url.toLowerCase();
+        if (imageryType === 'bing' || url.indexOf('tiles.virtualearth.net') > -1)
+            return 'bing';
+        // if (url.indexOf("scanex_irs") > )
+        if (
+            imageryType === 'TMS' &&
+            url.match(/.+tiles\.mapbox\.com\/v[3-9]\/openstreetmap\.map.*/)
```
pnorman: A sed pattern (`s/foo/bar/`) can use any character to delimit, and it's normal to use something other than `/` when dealing with URLs, e.g. `s@foo@bar@`. This avoids excessive escaping. Is this possible with Javascript's regex implementation?

modules/ui/background.js
```diff
@@ -185,7 +185,7 @@ export function uiBackground(context) {
         function update() {
             backgroundList.call(drawList, 'radio', clickSetSource, function(d) { return !d.overlay; });
             overlayList.call(drawList, 'checkbox', clickSetOverlay, function(d) { return d.overlay; });
-
+
```
pnorman: Stray whitespace

modules/ui/panels/background.js
```diff
@@ -81,6 +86,35 @@ export function uiPanelBackground(context) {
...
+        var tile = d3.select('.layer-background img.tile-center'); // tile near viewport center
+        if (tile.empty()) return;
+
+        var d = tile.datum();
+        var zoom = d[2];
+        var url = d[3];
+        var center = context.map().center();
+        console.log(d);
+        searchOffset.search(center,url, console.log);
+        //     zoom = (d && d.length >= 3 && d[2]) || Math.floor(context
```
pnorman: Should this commented code be removed?

modules/ui/panels/background.js
```diff
@@ -1,14 +1,17 @@
...
+var searchOffset = services.imageryOffset;
+searchOffset.init();
 
 export function uiPanelBackground(context) {
     var background = context.background();
     var currSource = null;
     var currZoom = '';
     var currVintage = '';
-
+    debugger;
```
pnorman: This seems out of place

modules/services/imagery_offset.js
```diff
@@ -22,8 +22,95 @@ export default {
...
+            // TMS: skip parameters with variable values and Mapbox's access token
+            if (
+                (kv.length > 1 &&
+                    kv[1].indexOf('{') >= 0 &&
+                    kv[1].indexOf('}') > 0) ||
+                kv[0] === 'access_token'
+            ) {
+                return;
+            }
+            console.log('here')
```
pnorman: Left over debugging statement?

modules/services/imagery_offset.js
```diff
@@ -22,8 +22,95 @@ export default {
...
+        ];
+
+        var qparams = {};
+        var qparamsStr = query.length > 1 ? query.slice(1).split('&') : '';
+
+        qparamsStr.forEach(function(param) {
+            var kv = param.split('=');
+            console.log(kv);
+            kv[0] = kv[0].toLowerCase();
+            // WMS: if this is WMS, remove all parameters except map and layers
```
pnorman: Not needed for iD

modules/services/imagery_offset.js
```diff
@@ -22,8 +22,95 @@ export default {
...
+
+        var query = '';
+        // Split URL into address and query string
+        var questionMarkIndex = url.indexOf('?');
+        if (questionMarkIndex > 0) {
+            query = url.slice(questionMarkIndex);
+            url = url.slice(0, questionMarkIndex);
+        }
+
+        var removeWMSParams = [
```
pnorman: Not needed for iD

modules/services/imagery_offset.js
pnorman: JavaScript probably contains more (or better) files for handling URLs. I'm wondering if it makes sense to write this differently, particularly when parsing the parameters part of the URL

pnorman:
> > do we support WMS type of imageries in iD, since I couldn't find one in imagery.json?
> Not yet.. that's #1141

I wouldn't worry about adding any WMS code here. #1141 is probably a long way off, and there's no way to currently test any WMS code you write.

bhousel:
> I wouldn't worry about adding any WMS code here. #1141 is probably a long way off, and there's no way to currently test any WMS code you write.

I'd prefer to leave it in actually.  This isn't WMS code, but rather code for generating an identifier to query the the Offset Database.

kepta:
Hey @Zverik , do you have test cases for this [code](https://trac.openstreetmap.org/browser/subversion/applications/editors/josm/plugins/imagery_offset_db/src/iodb/ImageryIdGenerator.java)

bhousel:
stale

PR #4172, opened 2017-07-21
Adding dog_excrement_bin

JamesKingdom:
I added a new preset that allows the use of dog excrement bin.
I also added the relative wiki page to pull a definition from.


bhousel:
Can you move the file:
`data/presets/presets/amenity/dog_excrement_bin.json`
to 
`data/presets/presets/amenity/waste/dog_excrement.json`

We have a naming convention to keep things organized.  The former would be for a tag like `amenity=dog_excrement_bin`, and the latter would be for some special kind of `amenity=waste`.

Other than that, looks good 👍

bhousel:
Looks good, thank you @JamesKingdom!

Issue #3919, opened 2017-03-21
Avoid style recalculation in breathe behavior


bhousel:
Currently d3 is doing a `getPropertyValue` to determine the original line width / circle radius in order to determine the animation parameters for the breathing.  We should just hardcode some reasonable values for vertices and lines to avoid doing this. 

![screenshot 2017-03-21 13 52 49](https://cloud.githubusercontent.com/assets/38784/24162379/b249b8b2-0e3d-11e7-8ac2-d1e052fc5d85.png)

bhousel:
A good read on this topic
https://medium.com/outsystems-experts/how-to-achieve-60-fps-animations-with-css3-db7b98610108

It applies more broadly to performance than just this issue.

kepta:
![image](https://user-images.githubusercontent.com/6966254/28819354-9afdef92-76cb-11e7-8fc4-7756a7c11b3d.png)
I got a chance to see forced reflow on chrome v60.

bhousel:
I'd really like to keep the animation.  Before we had it, we got a lot of feedback from people saying they couldn't see when things were selected.  (colorblind people especially need this)

I think this animation is pretty simple and we should be able to do it without triggering reflow.  Here are some more thoughts:

- opacity/transparency interpolation
  - browsers should have this optimized
  - can hardcode min and max values, no need to query style at each animation cycle
  - can possibly just interpolate the opacity of whatever layer all these shadow strokes are on, rather than interpolating the opacity of each shadow stroke
- vertices (circles)
  - we currently interpolate `radius`, but we might be more performant in browsers to interpolate scale of the circle
  - we query min and max radius values, as there are a few different sized vertices (endpoints or vertices with icon are larger)
  - if we switch to interpolating scale instead of radius, can hardcode min/max like 0.8/1.2 or something and avoid the style query
- nodes (poi markers)
  - all poi markers are the same, so we can hardcode min/max stroke width for this too.
- ways 
  - this is were we most need to query the stroke width and calculate a min/max range because
    - there are several different stroke widths of ways (think area edges, thin highways, thick highways, airport runways, etc)
    - toggling wireframe mode while things are selected will change the width too

- another thing that seems very slow is how we are adding the interpolator behavior to each thing selected.  Use the lasso to select 20 or more vertices and you will see it slow down a lot.  (though vertices are circles and there may be low hanging fruit to improve the perf of those, described above)

bhousel:
stale

PR #4177, opened 2017-07-24
try some placeholder values for breathe.js

kepta:
I tried to use some reasonable values for the fills and radius.

I guess we want a bit more fine grained placeholder values?

ref: https://github.com/openstreetmap/iD/issues/3919

### DIFF
--- a/modules/behavior/breathe.js
+++ b/modules/behavior/breathe.js
@@ -73,31 +73,29 @@ export function behaviorBreathe() {
 
 
     function calcAnimationParams(selection) {
-        selection
-            .call(reset)
-            .each(function(d) {
-                var s = d3.select(this),
-                    tag = s.node().tagName,
-                    p = {'from': {}, 'to': {}},
-                    opacity, width;
+        selection.call(reset).each(function(d) {
+            var s = d3.select(this),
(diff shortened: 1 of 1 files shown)


bhousel:
I tested with the Chrome 59 profiler to see if these changes make a difference but it's not really clear to me.  I think there is a benefit but I will need to do some more testing when I'm more awake.  Also, Chrome changed their profiler around 58 so I'm not super familiar with the new Performance tab.

One thing for sure, the max line width will need to be wider.  It is obscured on thick roads. 

The reason we used `s.style` before (to get the calculated style) was because:
- When the user changes a line preset, it may have a new width .  e.g. changing from a think line `highway=residential` to a thin line `highway=service`, or `power=line`
- When the user toggles wireframe everything has small width and needs to change.

bhousel:
Hey @kepta, last night Chrome 60 installed itself, so I took a look with the new profiler and I'm not sure whether querying the style is causing frame drops anymore.

Let's chat this week about performance stuff. There is a LOT in here that I don't understand and it would be great for us both to learn more about it.

kepta:
@bhousel sounds great.

bhousel:
closing as stale

PR #4183, opened 2017-07-26
Change reference field to be more specific

### LINKED ISSUE #4180: Give Reference field more specific labels
The Reference field, which is backed by the `ref` tag and appears in numerous presets, tends to be [misinterpreted as a place to indicate a source](https://taginfo.openstreetmap.org/tags/?key=ref&value=http%3A%2F%2Fglobus.tut.by), similar to the purpose of a “Reference” section in a Wikipedia article. As far as I can tell, none of the use cases for `ref` is actually called “a reference” per se in English.

iD should head off this confusion and promote appropriate usage of `ref` by naming the field more appropriately based on the preset. For example:

* On Road Route, Bus Route, Bicycle Route, etc., call it “Route Number”.
* On Motorway, Trunk Road, Primary Link, etc., call it “Road Number” – or “Route Number” also?
* On Motorway Junction, call it “Junction Number” or “Exit Number”.
* On Airport and Airfield, call it “Airport Code”.
* On Stop Position, call it “Stop Number”.
* On Platform, call it “Platform Number”.

For any miscellaneous uses, “Reference Code” would be a good alternative to “Reference”.

There’s already a variant of the Reference field with a placeholder reading “Hole number”; I think it would be better to label the field itself “Hole Number”.

JamesKingdom:
This changes the labels to fix #4180 
What's changed?

- Routes
- Roads
- Junctions
- Airports/airfields
- Stop position
- Platform
- Golf hole


boothym:
I'm wondering if ref is even needed for an airport - https://wiki.openstreetmap.org/wiki/Tag:aeroway=aerodrome says to use other tags for an airport code. Looking at airports in the UK the ref tag seems to just be the same as iata.

bhousel:
Thanks @JamesKingdom, looks good to me!   I just merged it.. 
(I'm also going to rename the golf_hole ref field to be like the other ones.)

PR #4187, opened 2017-07-28
Removing aeroway reference code

JamesKingdom:
This fixes the latest issue in #4180


bhousel:
Thanks for the quick fix @JamesKingdom!

PR #4192, opened 2017-07-31
Adding buffer_stop preset

JamesKingdom:
This adds a buffer_Stop preset, using a lovely new icon :+1:


bhousel:
Looks awesome, thanks @JamesKingdom.. 

I didn't know about these things until today, but around where I live they are mapped as `railway=halt`, which I guess is wrong. Looks like I have some fixing to do. 🚂

PR #4196, opened 2017-07-31
Adding some more railway presets

### LINKED ISSUE #4193: More Railway Presets
Following #4192, I took a look at https://wiki.openstreetmap.org/wiki/Key:railway, and there are a bunch of other common railway things that we don't have presets for.

We don't need everything, but I think these are easy to add:
* `railway=signal`  (I'll draw an icon - considered traffic light, but it should be different)
* `railway=switch` (`"icon": "poi-junction"`)
* `railway=derail` (`"icon": "roadblock"`)

@JamesKingdom want to add these (and whatever others you want to?)

JamesKingdom:
This is to, fix #4193 
So far I have added

- Railway Derailer
- Railway Milestone
- Railway Signal
- Railway Switch
- ~~Railway Wash~~ Train wash (6da0719f3a49d4173fe4769e2522a2edccd62885)

Todo:

- [x] Find an appropriate icon for milestone (triangle-stroked)
- [x] Draw railway-signal icon, ~~maybe using something similar to this:~~ Traffic light with two lights instead of 3
- [x] Update wash icon.
- [x] ~~Remove unneeded name from derail and other tags~~ Not possible
- [x] Add a placeholder that tells the user what to input into the milestone field

This could also change the car_wash icon, as currently it is just a car, and looks like a few other presets, this is also the same for railways. So I thought we could maybe add some brushes around the side like shown:
wash
All we need to do is change the car to a train, then we're done.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -3760,6 +3760,11 @@ en:
         name: Railway Crossing (Path)
         # 'terms: crossing,pedestrian crossing,railroad crossing,level crossing,grade crossing,path through railroad,train crossing'
         terms: '<translate with synonyms or related terms for ''Railway Crossing (Path)'', separated by commas>'
+      railway/derail:
+        # railway=derail
+        name: Railway Derailer
+        # 'terms: derailer'
+        terms: '<translate with synonyms or related terms for ''Railway Derailer'', separated by commas>'
       railway/disused:
         # railway=disused
         name: Disused Railway
@@ -3779,6 +3784,11 @@ en:
         name: Railway Crossing (Road)
         # 'terms: crossing,railroad crossing,level crossing,grade crossing,road through railroad,train crossing'
         terms: '<translate with synonyms or related terms for ''Railway Crossing (Road)'', separated by commas>'
+      railway/milestone:
+        # railway=milestone
+        name: Railway Milestone
+        # 'terms: milestone,marker'
+        terms: '<translate with synonyms or related terms for ''Railway Milestone'', separated by commas>'
       railway/monorail:
         # railway=monorail
         name: Monorail
@@ -3796,6 +3806,11 @@ en:
         # railway=rail
         name: Rail
         terms: '<translate with synonyms or related terms for ''Rail'', separated by commas>'
+      railway/signal:
+        # railway=signal
+        name: Railway Signal
+        # 'terms: signal,lights'
+        terms: '<translate with synonyms or related terms for ''Railway Signal'', separated by commas>'
       railway/station:
         # railway=station
         name: Railway Station
@@ -3811,6 +3826,16 @@ en:
         name: Subway Entrance
         # 'terms: metro,transit'
         terms: '<translate with synonyms or related terms for ''Subway Entrance'', separated by commas>'
+      railway/switch:
+        # railway=switch
+        name: Railway Switch
+        # 'terms: switch,points'
+        terms: '<translate with synonyms or related terms for ''Railway Switch'', separated by commas>'
+      railway/train_wash:
+        # railway=wash
+        name: Train Wash
+        # 'terms: wash,clean'
+        terms: '<translate with synonyms or related terms for ''Train Wash'', separated by commas>'
       railway/tram:
         # railway=tram
         name: Tram
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -730,6 +730,9 @@ en:
       maxweight:
         # maxweight=*
         label: Max Weight
+      milestone_position:
+        # 'railway:position=*'
+        label: Milestone Position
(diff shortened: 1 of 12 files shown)


manfredbrandl:
Maybe you could use this icon as a starting point, as caution is unique to railways:
image

bhousel:
Because there are so many different railway signals around the world, I think I'm just going to use the existing traffic signal icon, but with 2 lights instead of 3.   I don't want to overthink it - it just needs to look like a signal but not a `highway=traffic_signals`.

Nakaner:
Just my two cents as a railway mapper and OpenRailwayMap contributor:

@bhousel wrote:
> Because there are so many different railway signals around the world, I think I'm just going to use the existing traffic signal icon, but with 2 lights instead of 3. I don't want to overthink it - it just needs to look like a signal but not a highway=traffic_signals.

I agree with that. Icons should be simple because they are small. iD is no railway map.

Milestones are white signs with black numbers on them mounted on catenary masts or separate poles in many developed countries nowadays. But JOSM uses a milestone icon, too. Could you please add `railway:position=*` to the milestone preset? It is used to indicate the distance printed on the milestone/sign.

Using a roadblock as a marker is inappropriate from my point of view but I do not know a better icon. Derailers are [small devices](https://commons.wikimedia.org/wiki/Category:Derails).

bhousel:
Thanks for your feedback @Nakaner!





> Milestones are white signs with black numbers on them mounted on catenary masts or separate poles in many developed countries nowadays. But JOSM uses a milestone icon, too. Could you please add railway:position=* to the milestone preset? It is used to indicate the distance printed on the milestone/sign.

Sounds good, we can add this.


> Using a roadblock as a marker is inappropriate from my point of view but I do not know a better icon. Derailers are small devices.

Oh the Maki "roadblock" icon I'm thinking of looks like this - it's not a literal road block:
screenshot 2017-07-31 15 14 51

JamesKingdom:
> Could you please add railway:position=* to the milestone preset?

Added:
add screenshot

Nakaner:
@JamesKingdom Thank you for the quick response. Could the input field show a "only one decimal"? Railway mappers use two tags for the distance, `railway:position` for the rounded value which is printed in large digits on the sign and `railway:position:exact` which is the raw value (resolution 1 meter) which is either printed in small digits or is missing at all.

Adding two values might confuse the users. :-/

You could drop the name field because they are nameless and it invites to add `name=*` which are no names.

Nakaner:
@bhousel wrote:
> Oh the Maki "roadblock" icon I'm thinking of looks like this - it's not a literal road block:

The icon is ok. You could rotate the bar by 45° anti-clockwise if you need a distinction between road blcoks and derailers. If I think that I have a better icon, I will file a pull request. Even the OpenRailwayMap JOSM MapCSS style does not have a derail icon.

JamesKingdom:
> Could the input field show a "only one decimal"? 
You could drop the name field

Thanks for the recommendation, unfortunately we can't remove the name, or only show one decimal.

> Milestones are white signs with black numbers on them mounted on catenary masts or separate poles in many developed countries nowadays.

We are adding another icon to look more like a milestone, just to confirm what feature they are adding, as a triangle is quite generic and could stand for essentially anything.

Again, it is nice to hear ideas from someone with more experience in the subject, thank you!

bhousel:
This seems fine @JamesKingdom  - can you commit the above change that adds the Milestone Position field, and then you can merge it, and I'll add the icons.

Sorry @Nakaner, as @JamesKingdom said we can not add decimal point formatting to the Milestone field, or remove the Name field.  Both of these are reasonable suggestions though, so you can open up new issues for these things if it would help avoid problems with railway mapping.  

For now the best thing I can suggest is to make sure that the osm wiki has a really good description and image in the tag template - these things are displayed when the user hits the ℹ️ button next to the field.

Nakaner:
@bhousel wrote:
> Sorry @Nakaner, as @JamesKingdom said we can not add decimal point formatting to the Milestone field, or remove the Name field. Both of these are reasonable suggestions though, so you can open up new issues for these things if it would help avoid problems with railway mapping.

I think you misunderstood me (or I expressed my thoughts not clearly enough). I don't want that the input fields enforce a format. This would be out of scope here. But the text which is displayed in an empty input field could suggest a format, couldn't it?

`distance on the milestone rounded to one decimal (i.e. 123.4)` or `rounded to one decimal (i.e. 123.4)`

JamesKingdom:
> I don't want that the input fields enforce a format

Sorry for the misunderstanding, but yes we can certainly add a placeholder.
I shortened it down because it looked pretty messy with all that text in.
screen

bhousel:
awesome, thanks @JamesKingdom 👍

PR #4201, opened 2017-08-01
Don't force the name field onto every object

### LINKED ISSUE #4200: Name field should not be mandatory for presets
Names are important but there are lots of features in OSM which usually do not have a name. [According to @bhousel](https://github.com/openstreetmap/iD/pull/4196#issuecomment-319183856), iD displays a name input field for all presets and this cannot be turned off. This leads newbies to add names and being later told by experienced mappers that they should only add names which are names.

I would like to suggest to either drop this default and force the authors of the presets to add a name field explicitly if necessary. As an alternative, preset authors could disable the name field. The latter does not require a modification of most presets but is counter-intuitive for preset authors and has the smell of being legacy. :-)

Examples where a `name=*` tag is usually wrong:

* traffic signals
* street lights
* cable distribution cabinets

The world is not simple. There are lots of parking areas which do not have a name in reality but have a pseudo-name in OSM (`name=<supermarket> parking`). But there are parking areas which have a real name. Other examples are intersections (in some regions), trees (most don't have but some have).

JamesKingdom:
This fixes #4200

## How?
Currently all presets force a name field onto them.

- Each preset will need `field=name` added onto it, only a few will not.
- The ones like buffer stop, don't _normally_ have names, so it will be removed. This helps to resolve some confusion, as new mappers might put something in there that is not relevant. 

### But what if that object really has a name?
- The option will be available in the "Add field" box at the bottom, and work just like the other options, so that a name can be added if needed.
sc1

- This will then add back a name field:
sc2


modules/ui/preset.js
```diff
@@ -40,7 +40,7 @@ export function uiPreset(context) {
         field.show = show;
 
         field.shown = function() {
-            return field.id === 'name' || field.show || _.some(field.keys, function(key) { return !!tags[key]; });
+            return field.id === '' || field.show || _.some(field.keys, function(key) { return !!tags[key]; });
```
bhousel: You can just remove the `field.id === ''` clause.  No field will be like this.

bhousel:
This looks pretty good. I like the idea of not giving the `name` field any special treatment.

I fixed a small thing:  now it is possible for there to be no fields in the "all fields" section, and that was causing this empty div to look weird:

screenshot 2017-08-01 13 03 30

Added a rule so that div will now appear if there are no fields:

screenshot 2017-08-01 13 06 34

bhousel:
Merging now, so I can get a start on #3914 today.. Per chat, @JamesKingdom will add most of the `name` fields back and open a new PR.

PR #4210, opened 2017-08-02
Update field values for every preset 

### LINKED ISSUE #4200: Name field should not be mandatory for presets

JamesKingdom:
This fixes #4200 

cc @pnorman


data/presets/presets/amenity/_register_office.json
```diff
@@ -1,6 +1,7 @@
 {
     "icon": "town-hall",
     "fields": [
+	    "name",
```
bhousel: Can you fix the indenting on the ones where they look wrong?

JamesKingdom: Fixed in edf650e74f8c00201e401474809e3aa7e8baf02c

bhousel:
Thanks @JamesKingdom, I appreciate all your work on adding these 💪

PR #4214, opened 2017-08-03
Field refactor

bhousel:
This is a WIP for #3914 

It doesn't change how things look, but it will let us put a field anywhere, rather than having them only exist in the preset editor.  This will cleanup some technical debt, and unblock issues where we want to show a subfield inside a field.


bhousel:
This is done and merged 👍

PR #4217, opened 2017-08-04
Add custom references for runway and taxiway, improve searchability

JamesKingdom:
This PR does the following:

- Runway now has "Runway number"
- Taxiway now has "Taxiway Name"
- T-Bar lift would not appear if "tbar" was entered, this has been changed
- Aerialway station can now be drawn as an area and as a building


bhousel:
looks good, thanks @JamesKingdom 
(I'm changing "e.g" to "e.g." because I'm nitpicky.)

Issue #3241, opened 2016-07-10
POI presets are not recognized in interpolated address lines


virgilinojuca:
If you use lines for [interpolating house numbers](http://wiki.openstreetmap.org/wiki/Addresses#Using_interpolation), and there are POIs in it, they will not be recognized with presets. This happens either if they are in the middle of the line or on the end nodes. Not only they will not have an icon on the map, but they will also display a “Other” preset in the left sidebar if you click on it.

![Comparison of with and without interpolation line](http://i.imgur.com/bb5othY.jpg)

They are also not recognized if they are mapped on the edge of a building—a practice that I have noticed mainly in strip malls where I live. I don't know if this practice is correct in the first place though, so it may not be a problem there.

bhousel:
ohhh #3219 asked for this and I didn't see the need - but yes I agree it's important for address interpolation lines.

bhousel:
Thanks for this suggestion, I implemented it in 0d8fb87

We now treat entities on address interpolation lines as points, not vertices.. This affects both the rendering and the preset matching and preset suggestions:

<img width="206" alt="screenshot 2016-07-14 12 24 40" src="https://cloud.githubusercontent.com/assets/38784/16848324/48b8ff64-49c3-11e6-87c7-95b4c52e5dab.png">

virgilinojuca:
Great! Thanks for considering my suggestion.

Issue #1316, opened 2013-04-15
presets/styles for address interpolation


jfirebaugh:
- [ ] Line tagged with `addr:interpolation` should show 1-way markers.
- [ ] Selecting the line should show two address fields, one for each end.
- [x] Field for `addr:interpolation`.

bhousel:
wip at #4220

boothym:
Nice, I'd thought about suggesting this but never got round to it. Is this available to test anywhere? As it's not on http://preview.ideditor.com/master/

JamesKingdom:
@boothym 

> Is this available to test anywhere

I'll have it running on a [test server](http://james.glideservers.com:8081) for a while

> As it's not on http://preview.ideditor.com/master/

Only the master branch is there, this pull request hasn't been merged yet, when the PR is merged, it will update master and be available there.

boothym:
@JamesKingdom thought so, thanks for the test server. 

Looks good - only thing I that I can suggest is the ability to search for the address preset on vertexes...

PR #4220, opened 2017-08-07
add preset to address interpolation

willemarcel:
https://wiki.openstreetmap.org/wiki/Addresses#Using_interpolation

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -605,6 +605,16 @@ en:
           down: Down
           # incline=up
           up: Up
+      inclusion:
+        # 'addr:inclusion=*'
+        label: Accuracy Level
+        options:
+          # 'addr:inclusion=actual'
+          actual: actual
+          # 'addr:inclusion=estimated'
+          estimated: estimate
+          # 'addr:inclusion=potential'
+          potential: potential
       indoor:
         # indoor=*
         label: Indoor
@@ -634,6 +644,18 @@ en:
       internet_access/ssid:
         # 'internet_access:ssid=*'
         label: SSID (Network Name)
+      interpolation:
+        # 'addr:interpolation=*'
+        label: Interpolation Type
+        options:
+          # 'addr:interpolation=all'
+          all: all
+          # 'addr:interpolation=alphabetic'
+          alphabetic: alphabetic
+          # 'addr:interpolation=even'
+          even: even
+          # 'addr:interpolation=odd'
+          odd: odd
       kerb:
         # kerb=*
         label: Curb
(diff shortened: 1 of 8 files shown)


bhousel:
I'm still considering whether we want to support this.. Anyone have thoughts?

Reading here: https://wiki.openstreetmap.org/wiki/Addresses#Using_interpolation
it seems like it is easy for users to make mistakes.  We don't have a way yet of prompting them that they are supposed to put house numbers along the line, and the wiki reference does not link directly to that page of documentation.

This PR doesn't include any of the extras mentioned in #1316, and I feel like the trend in OSM has shifted since then to just mapping all the address points.

willemarcel:
@bhousel I changed the https://wiki.openstreetmap.org/w/index.php?title=Key:addr:interpolation link to redirect to the interpolation section of the `Key:addr` page.

My opinion is that, as this preset will not be listed in one of the main categories, probably the only way to someone find it is to search for 'interpolation'. So I believe who will use it knows what is doing.

In most Argentina cities the blocks are squares with 100m long in each face, so they use interpolation a lot. Currently if a user clicks in a interpolation line he won't have information about what it is. It just shows the label `Line`, so I think it'll be more educative to have this preset.

If you approve it, I suggest to change the label from `Interpolation` to `Address Interpolation`.

pnorman:
We should have *something* for interpolation so existing objects display correctly.

willemarcel:
@bhousel Have you made a decision on this? If you think users should not be encouraged to create interpolations, I could add `"searchable": false` to the preset.

bhousel:
Sorry for the delay @willemarcel & @pnorman - I agree we should add this.

I have a suspicion that what I did for #4353 and #3241 might conflict with an address interpolation preset.  I just need to set aside a few hours to check it out and really think through the issues.

ignaciolep:
Hi! In Argentina (and [many other places around the world](https://taginfo.openstreetmap.org/keys/addr:interpolation#map)) we have most of the addresses mapped with addr:interpolation. Unfortunately, iD shows them as a generic line.

As iD now has the autofix feature, it would be great to have a preset for addr:interpolation what would autoadd/autofix the addr:housenumber and addr:street tags in the nodes of the addr:interpolation lines.

quincylvania:
I added a version of this PR with an unsearchable "Address Interpolation" preset and an associated field. I'm leaving the initial issue open since that also covers rendering.

> As iD now has the autofix feature, it would be great to have a preset for addr:interpolation what would autoadd/autofix the addr:housenumber and addr:street tags in the nodes of the addr:interpolation lines.

@ignaciolep Interesting idea! It seems really difficult though.

Issue #3968, opened 2017-04-13
Write a changeset tag that says whether a user has completed the walkthrough


bhousel:
I've spent a lot of time on improving the walkthrough (#3921).  It might be interesting to know whether new users are actually doing the walkthrough or not.  

For example, if we see a new user making some interesting pattern of mistakes, and we know they _did_ complete the walkthrough, we can improve the walkthrough to address those mistakes.

If we see that they _didn't_ do the walkthrough, we can ask them to complete the walkthrough in a changeset comment when we review their edits.

Thoughts?

pnorman:
Do we reliably have this information? I've been asked several times to do the walkthrough. I suppose in the case of new users it might be valuable because it'd most likely be accurate.

bhousel:
We don't have this information now. Instead we just set a flag in the browser's local storage if the user has never used iD before. But:
* we could start storing things in the user's OSM profile (see #3002)
* we do know from the OSM API if a user has never saved an edit before

pnorman:
I'd go for #3002 and store walkthrough state server-side.

Issue #2834, opened 2015-11-19
Support for additional external changeset tags.


maxerickson:
A recent discussion on the talk mailing list raises the issue of separating project tracking information out of changeset comments. Starts here:

https://lists.openstreetmap.org/pipermail/talk/2015-November/thread.html#74992

A prerequisite for making this work for HOT would be support in id. I don't have a patch, but looking at #2311 I gather that deciding what to do would be a bigger problem than the implementation. It's also not entirely clear if there is really widespread buy-in for omitting hashtag like stuff from comments.

bhousel:
Yes I'm following the discussion.  I agree that this is a case where getting community consensus is more work than the actual implementation itself.  I'd happily add another field for tags next to the changeset comment field, and make it settable via the api, if that's how the discussion shakes out.

maxerickson:
There is also already a tracking bug here: https://github.com/hotosm/osm-tasking-manager2/issues/703

pnorman:
There are a number of common changeset tags like `import`, but those wouldn't normally be used within iD.

bhousel:
> I'd happily add another field for tags next to the changeset comment field,

Sorry @pnorman I just re-read what I wrote and I realize it's confusing.  I meant that I'd add another field for *hashtags*.

pnorman:
This issue is about changeset tags, not trying to encode information in the comment where we have structured alternatives.

bhousel:
> This issue is about changeset tags, not trying to encode information in the comment where we have structured alternatives.

Yes exactly - the proposal is to add a new changeset tag (TBD but named something like `hashtags`), if I understand it correctly.

maxerickson:
One way to implement `hashtags` would be to support adding arbitrary tags to the changeset, for example by checking parameters for the prefix `changeset_` and using the rest of that name for the key, so that `changeset_comment` would create a `comment` tag and `changeset_hashtags` would create an `hashtags` tag.

Imports would be a possible motivation for supporting arbitrary tags. I guess it might be nice to have QA tools identify themselves too (they aren't always quite a source). But those things would probably fit into a single additional key.

pnorman:
> Imports would be a possible motivation for supporting arbitrary tags. I guess it might be nice to have QA tools identify themselves too (they aren't always quite a source). But those things would probably fit into a single additional key.

Some imports have had multiple keys, `import=yes`, `source=...`, `website=<link to wiki page>`

Hedaja:
[...]

PR #4223, opened 2017-08-08
Add requesting review

kepta:
This adds review request.


pnorman:
It'd be good to sort the tagging out with the community, not on the iD issue tracker.

bhousel:
> It'd be good to sort the tagging out with the community, not on the iD issue tracker.

Why?  My feeling has always been that:
1. The iD issue tracker is part of the community.  It's open, and anybody who wants can follow what we do or comment or suggest changes.
2. Changeset tags are really more decided by the editor software than by anybody else.
3. We can use [any tags we like](https://wiki.openstreetmap.org/wiki/Any_tags_you_like).

I am really happy to change the tag, if anybody cares.  It doesn't have to be `review_requested=yes`.

iandees:
I realize I could check this out and look, but what are y'all thinking about for prompting the user to ask for review? A popup? A checkbox near the changeset save button?

bhousel:
This request review field is great and I could merge it right now, but instead I'm going to spend some time to refactor the commit pane a little bit - it is like 500 lines of code and getting difficult to manage as we add things to it.  

Managing a bunch of fields and the tags that they change is something that `uiEntityEditor` and `uiPresetEditor` already do well, so I'm going to use that as a model to split `uiCommit` into smaller modules.

@iandees, here's a screenshot showing current placement of the checkbox, along with some dummy fields that I stuck at the top.  The dummy fields are being rendered by the new `uiChangesetEditor` component.

screenshot 2017-08-11 17 29 02

bhousel:
Refactoring the commit pane by using the new `uiChangesetEditor` module is working nicely.  It now creates real `uiField` objects and supports a "more fields" section.  We can let users tag a `source` now.  Here are some things that I need to put back:

- [ ] dropdown for previous changesets
- [ ] placeholder texts
- [ ] google warning
- [ ] enable save button when comment exists
- [ ] double check the warnings and summary sections of `uiCommit`

and some scope creep..

- [ ] #2834 - add hashtags
- [ ] #4230 - <kbd>esc</kbd> should cancel save mode
- [ ] #3968 - few more changeset tags.. `first_edit`, `completed_walkthrough`, others?

bhousel:
Looks like this currently

save mode

iandees:
It would be great to think of a way to call out that "request review" checkbox, especially for new folks. Can it pulsate or be a different color or have an arrow with a friendly [globe](https://mapzen-assets.s3.amazonaws.com/images/introducing-tangram-work/globey.png) suggesting they give it a try?

bhousel:
> It would be great to think of a way to call out that "request review" checkbox, especially for new folks. Can it pulsate or be a different color or have an arrow with a friendly globe suggesting they give it a try?

I have thoughts on this request review checkbox and how it might be perceived by different kinds of users.

Not everybody _wants_ feedback on their edits.  For example, millennials may be more likely to want the feedback, and genx'ers may be more likely to not want the feedback. Majority users may be more comfortable with public edits, and minorities may not feel comfortable having their edits under additional scrutiny.

I think this feature could be very helpful and welcome to a lot of people - especially new users - but I don't want to put assumptions into the software, and so that's why we ask the user whether they want the review or not.

Also at this point we don't know how many people will check the box, or who will step up to do the reviews.  The answer is some combination of - OSMCha, whodidit, bots, rss feeds, dedicated validators, paid mapping teams, and volunteers.

I thought about just checking the box if it's a user's first edit, or having the preference be "sticky", but I don't want to promise something the OSM community can't deliver.  So for now, let's leave the box unchecked, and "opt-in" to review.  We can change this later and encourage more users to check the box, if after a few months we decide that the reviews are working well and helpful.

mvexel:
Agree with @bhousel -- until we know how many requests will actually be honored it's probably better to not call it out too much. If people don't want to review or use discouraging words to do so it would just serve to turn more people off than the other way around.

talllguy:
Here's an idea for frequently used sources as checks, streamlining taginfo a bit 
image

bhousel:
Hey @talllguy, one benefit of the recent refactoring work is that it makes it super easy to just reuse fields like the ones that iD already has other places.  We don't have a field that's an array-of-checkboxes, but we do have a multiselect field that turns the values into a semicolon separated list.

What do you think of this?:

sources

talllguy:
@bhousel I like to see this fields reused this way. This implementation looks cool! I'd use this.

Next steps would be using some an algorithm to analyze the changeset comment and other metadata to suggest source tags. ;)

bhousel:
FYI just pushed a working implementation of hashtags (see #2834).
This is just another semicolon-delimited multiselect field, and I added some extra code to detect hashtags embedded in the comment.

The idea here is to give mapathons, task managers, QA tools, a special field to store their hashtags, instead of (ab)using the comment field, which some people don't like.  Hashtags will also be settable through the API just like comments currently are.

hashtags

bhousel:
Just merged this - thanks for your patience!
Hope these new features will make it easier for other projects to analyze changesets.

PR #4224, opened 2017-08-09
Add jetty term to pier

boothym:
Looking through man_made=jetty, quite a number of them should be piers.

Think I messed up resetting github but hope this PR is ok for you (save me having to fully refork)


bhousel:
Thanks @boothym, this is fine - we can rebuild the presets.

PR #4226, opened 2017-08-10
Add terms to wetland

boothym:
I was retagging an old natural=marsh so searched for marsh in the presets, but didn't find anything. I've added the top five wetland=* values to help people find the wetland preset.

Maybe these could be made into proper presets in the future?

### DIFF
--- a/data/presets/presets/natural/wetland.json
+++ b/data/presets/presets/natural/wetland.json
@@ -10,6 +10,12 @@
     "tags": {
         "natural": "wetland"
     },
-    "terms": [],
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @boothym!  This is a good idea to help people find the wetland preset.

I don't think we need to make sub presets for all the different types, because there is already a `wetland=*` field (though I'm totally fine if somebody wants to do it, similar to how we have `water=*` sub presets: lake, pond, etc).

screenshot 2017-08-10 21 07 22

PR #3729, opened 2017-01-08
Add Turkish address scheme

mertemin:
This PR adds Turkish address scheme defined [here](http://postakodu.ptt.gov.tr/Dosyalar/adres.pdf). That is an official documentation, but only available in Turkish.

The scheme is defined as (in Turkish):
* Mahalle / Köy
* Bulvar/ Cadde / Sokak / Mevki / Site / Blok / Bina ismi / Dış Kapı No / Kat / İç Kapı No
* Posta kodu / Bucak (Semt) / İlçe / İl / Ülke 

That maps to English as:
* Neighborhood
* Street, House number
* Postal code, District, City

PR #4235, opened 2017-08-16
Adding unit field to USA

### LINKED ISSUE #3980: Add addr:unit to U.S. address field
Some U.S. mappers put unit/suite/room numbers in the `addr:street` tag because that’s one of two places in a written address where the unit number may go. (It may also be a suffix on the house number.) For example, [this POI](http://www.openstreetmap.org/node/4064072925) is tagged `addr:housenumber=7220` `addr:street=Telegraph Square Dr Suite V`, and [this POI](http://www.openstreetmap.org/node/2480216595) is tagged `addr:housenumber=5825` `addr:street=University Research Court, Suite 3500`. The U.S. address field in iD should have a small text box corresponding to the `addr:unit` tag to the right of the `addr:street` text box, to encourage more structured tagging.

JamesKingdom:
Closes #3980

### DIFF
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -61,7 +61,7 @@
             ]
         },
         {
-            "countryCodes": ["au"],
+            "countryCodes": ["au","ca"],
             "format": [
                 ["housenumber", "street", "unit"],
                 ["city", "province", "postcode"]
@@ -72,13 +72,6 @@
             "format": [
                 ["housenumber", "street", "unit"],
                 ["city", "state", "postcode"]
-            ]
-        },
-        {
-            "countryCodes": ["ca"],
-            "format": [
-                ["housenumber", "street","unit"],
-                ["city", "province", "postcode"]
             ]
         },
         {
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -2,7 +2,7 @@
     "dataAddressFormats": [
         {
             "format": [
-                ["housenumber", "street"],
+                ["housenumber", "street", "unit"],
                 ["city", "postcode"]
             ]
         },
@@ -10,7 +10,7 @@
             "countryCodes": ["gb"],
             "format": [
                 ["housename"],
-                ["housenumber", "street"],
+                ["housenumber", "street", "unit"],
                 ["city", "postcode"]
             ]
         },
@@ -18,7 +18,7 @@
             "countryCodes": ["ie"],
             "format": [
                 ["housename"],
-                ["housenumber", "street"],
+                ["housenumber", "street", "unit"],
                 ["city"],
                 ["postcode"]
             ]
@@ -31,21 +31,21 @@
                 "pt", "se", "si", "sk", "sm", "va"
             ],
             "format": [
-                ["street", "housenumber"],
+                ["unit","street", "housenumber"],
                 ["postcode", "city"]
             ]
         },
         {
             "countryCodes": ["fr", "lu", "mo"],
             "format": [
-                ["housenumber", "street"],
+                ["unit","housenumber", "street"],
                 ["postcode", "city"]
             ]
         },
         {
             "countryCodes": ["br"],
             "format": [
-                ["street"],
+                ["street","unit"],
                 ["housenumber", "suburb"],
                 ["city", "postcode"]
             ]
@@ -53,7 +53,7 @@
         {
             "countryCodes": ["vn"],
             "format": [
-                ["housenumber", "street"],
+                ["housenumber", "street","unit"],
                 ["subdistrict"],
                 ["district"],
                 ["city"],
@@ -77,7 +77,7 @@
         {
             "countryCodes": ["ca"],
             "format": [
-                ["housenumber", "street"],
+                ["housenumber", "street","unit"],
                 ["city", "province", "postcode"]
             ]
         },
@@ -86,7 +86,7 @@
             "format": [
                 ["postcode", "city", "district"],
                 ["place", "street"],
-                ["housenumber", "floor"]
+                ["housenumber", "floor","unit"]
             ]
         },
         {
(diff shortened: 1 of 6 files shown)


bhousel:
This looks good - why only USA?

data/address-formats.json
```diff
@@ -31,46 +31,53 @@
...
-                ["housenumber", "street"],
+                ["housenumber", "street","unit"],
                 ["subdistrict"],
                 ["district"],
                 ["city"],
                 ["province", "postcode"]
             ]
         },
         {
+            "countryCodes": ["au"],
```
bhousel: You can just combine this with `"ca"` (Canada), they're the same

bhousel:
Thanks @JamesKingdom this looks great 👍

data/address-formats.json
```diff
@@ -31,55 +31,55 @@
...
+                ["street","unit"],
                 ["housenumber", "suburb"],
                 ["city", "postcode"]
             ]
         },
         {
             "countryCodes": ["vn"],
             "format": [
-                ["housenumber", "street"],
+                ["housenumber", "street","unit"],
```
1ec5: I don’t think we should include `addr:unit` in the [Vietnam address format. There’s only [one (mistaken) use](http://overpass-turbo.eu/s/r2X) of this tag in all of Vietnam, reflecting the fact that units aren’t numbered in that country. Floor numbers are sometimes given in addresses, but generally the smallest level of addressing is the building number.

Addresses can be quite granular even without `addr:unit`, because the `addr:housenumber` and `addr:street` tags are both hierarchical. For example, this is the address of [building 8, alley 251, lane 48, Nguyễn Khang street](http://www.openstreetmap.org/way/351348266):

> Số 48/251/8 đường Nguyễn Khang
> Xã Yên Hòa
> Quận Cầu Giấy
> Hà Nội

```
addr:housenumber=48/251/8
addr:street=Nguyễn Khang
addr:subdistrict=Yên Hòa
addr:district=Cầu Giấy
addr:city=Hà Nội
```

`addr:housenumber` can be even more complex, for example [`7/14/12/3/23`](http://www.openstreetmap.org/node/5021351015). On the other hand, some mappers prefer to put all the alley and lane numbers in `addr:street` so the map doesn’t look so cluttered, for example [lane 19, Liễu Giai street](http://www.openstreetmap.org/node/4453205990).

More information about Vietnam’s address formats:

https://vi.wikipedia.org/wiki/Địa_chỉ
https://youbianku.com/files/upu/VNM.pdf
http://vforum.vn/diendan/showthread.php?91066-Thon-Xom-Ap-Xa-Phuong-Huyen-Quan-Tinh-Thanh-Pho-trong-tieng-Anh-la-gi-khi-viet-dia-chi-nha

/ref #2252

JamesKingdom: Alright, thanks for the information, I removed it in #4242

speedbiker:
Is it possible to leave this "addr:unit"-field away in the German version, especially in Austria? It's useless for adresses in Austria and Germany.
And then more space on the monitor is available for addr:street-field.

althio:
I think we should remove addr:unit in the `fr` address format. So PR #4667

But I must say... I think it was a bad move to [add `addr:unit`](https://github.com/openstreetmap/iD/commit/de6b4fb6a20b2a3438b4159192e2b54682fab2e5) by _default_ and in _all_ country formats.

bhousel:
> I think it was a bad move to add addr:unit by default and in all country formats.

Why?

althio:
It is beneficial for US, CA, AU and doubtlessly other countries around the world (I guess Malaysia, HK...)

But some countries already had an address format in iD, why change them?
The rollbacks show that it was wrong for:
- [VN](https://github.com/openstreetmap/iD/pull/4242)
- [CN](https://github.com/openstreetmap/iD/pull/4248)
- [BR](https://github.com/openstreetmap/iD/pull/4284)
- [AT, CH, DE](https://github.com/openstreetmap/iD/pull/4301)
- [PL](https://github.com/openstreetmap/iD/pull/4328)
- [FR, LU, MO](https://github.com/openstreetmap/iD/pull/4667)
- ping @rory for [IE](https://github.com/openstreetmap/iD/pull/2729)
- ping @david082321 for [TW](https://github.com/openstreetmap/iD/pull/3261)
- ping @mertemin for [TR](https://github.com/openstreetmap/iD/pull/3729)
- ping @Andygol for [UA](https://github.com/openstreetmap/iD/pull/3995)
- ping @boothym for [GB](https://github.com/openstreetmap/iD/pull/4667#issuecomment-356093439)
I have my doubts about NO and a few others.

I have my own bias about why addr:unit should not be in a standard default address format worldwide.
I feel it is suitable for some countries with dense cities and *HIGH* towers. It makes little sense worldwide, if I must guess.

speedbiker:
For HU and SI I also do not see any benefit.

rory:
I don't think `addr:unit` should be a default for IE.

I read the [wiki page for `addr:unit`](https://wiki.openstreetmap.org/wiki/Proposed_features/addr_keys_(2011-04)#addr:unit) and I'm still confused. We don't tend to have "multiple separate entrances". Irish addresses are all over the place, so it's not uncommon for it to be unclear where things are.

"Unit" is very rarely used in Irish addresses, mostly for separate addresses within a (light) industrial estate. So you might see "Mappy Ltd., Unit 5, Whatever Road, Ballytown Industrial Estate, Co. Whatever". Here "Unit" is `addr:housenumber`, and different from `addr:unit`. I'm unsure if there is >100 addresses in Ireland that would use `addr:unit`.

mertemin:
`addr:unit` is not needed for Turkish either.

bhousel:
Please submit PRs so that the address field looks however you want it to look.

PR #4236, opened 2017-08-17
add presets for theme park attractions

willemarcel:
attraction key reference: https://wiki.openstreetmap.org/wiki/Proposed_features/Key:attraction

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -1219,6 +1219,11 @@ en:
       source:
         # source=*
         label: Source
+      species:
+        # species=*
+        label: Scientific Name
+        # species field placeholder
+        placeholder: Scientific name of the species
       sport:
         # sport=*
         label: Sports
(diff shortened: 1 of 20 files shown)


data/presets/presets.json
```diff
@@ -3317,6 +3317,295 @@
...
+                "lion",
+                "tiger",
+                "bear"
+            ],
+            "tags": {
+                "attraction": "animal"
+            },
+            "name": "Animal"
+        },
+        "attraction/big_wheel": {
```
magol: Maybe height and/or Radius

data/presets/presets.json
```diff
@@ -3317,6 +3317,295 @@
...
+                "autoscooter"
+            ],
+            "tags": {
+                "attraction": "bumper_car"
+            },
+            "name": "Bumper Car"
+        },
+        "attraction/bungee_jumping": {
+            "icon": "pitch",
+            "fields": [
```
magol: Maybe height?

data/presets/presets.json
```diff
@@ -3317,6 +3317,295 @@
...
+                "river rapids ride"
+            ],
+            "tags": {
+                "attraction": "river_rafting"
+            },
+            "name": "River Rafting"
+        },
+        "attraction/roller_coaster": {
+            "icon": "amusement-park",
+            "fields": [
```
magol: Maybe some files from https://wiki.openstreetmap.org/wiki/Proposed_features/key:roller_coaster

At least a reference to https://rcdb.com

JamesKingdom: Would it also make sense to add `fee` onto these presets? Most parks charge per ride.

bhousel:
@willemarcel Can you please remove `species`?  per #3600, I'd prefer for all the sidebar fields in iD to be only the kinds of things that an untrained mapper could fill in.

Everything else looks great, thanks!  👍

data/presets/presets.json
willemarcel: added height!

data/presets/presets.json
willemarcel: added! thanks @magol !

willemarcel:
@bhousel done!

bhousel:
Thanks @willemarcel this looks great!.. merging..

bhousel:
BTW, if you are looking for another tag to do, you could do the same thing for playground equipment -
 `playground=*`
https://wiki.openstreetmap.org/wiki/Key:playground

(Someone was asking about this one [on reddit](https://www.reddit.com/r/openstreetmap/comments/6j5wvs/idea_playground_equipment/) recently.)

data/presets.yaml
```diff
@@ -2101,6 +2101,71 @@ en:
...
+        name: Bumper Car
+        # 'terms: theme park,dodgem cars,autoscooter'
+        terms: '<translate with synonyms or related terms for ''Bumper Car'', separated by commas>'
+      attraction/bungee_jumping:
+        # attraction=bungee_jumping
+        name: Bungee Jumping
+        # 'terms: theme park,bungy jumping,jumping platform'
+        terms: '<translate with synonyms or related terms for ''Bungee Jumping'', separated by commas>'
+      attraction/carousel:
+        # attraction=carousel
```
1ec5: Part of the tagging proposal is an [`attraction=swing_carousel`](https://taginfo.openstreetmap.org/tags/attraction=swing_carousel) tag for [swing rides](https://en.wikipedia.org/wiki/Swing_ride). That caught my attention because “swing ride” apparently has a well-known translation in Vietnamese, but ordinary horse carousels don’t have as obvious a translation.

PR #4248, opened 2017-08-19
Update Chinese address format

Stenive:
We have `addr:district` that works similar as `addr:unit` (in western style).
In Chinese language, the word 'unit' is translated to "单元"(a numbered entrance of a building), which is more specific than `addr:housenumber`, so there's no need to keep it.

See [https://en.wikipedia.org/wiki/Address_(geography)#China](https://en.wikipedia.org/wiki/Address_(geography)#China)

#3980 
#4235

### DIFF
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -122,8 +122,7 @@
             "countryCodes": ["cn"],
             "format": [
                 ["postcode", "province"],
-                ["city", "unit"],
-                ["district"],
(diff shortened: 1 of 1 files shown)


JamesKingdom:
Thanks for changing this, it looks like unit is only used a few times in china.
I have never used a Chinese phone code, so it is nice to hear from someone who knows about it.
Look good :+1:

PR #4255, opened 2017-08-21
Make office=physician non-searchable

M1dgard:
and add term "physician" to `amenity=doctors`.

`office=physician` is only documented on the wiki at the seemingly abandoned Healthcare 2.0 proposal. It makes no sense to have its preset searchable. The supported tag for physicians is `amenity=doctors`.


bhousel:
Thanks @M1dgard, looks good 👍

PR #4279, opened 2017-08-25
Move zoom to edit notice from sidebar to map container

leegenes:
* moved uiNotice from sidebar.js to init.js 
* made couple of small styling changes on the for the notice module to play well over the map

Hope this is all on the right track. All hints, tips, critiques are very welcome for this beginner.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -3627,18 +3627,17 @@ img.tile-removing {
 
 .notice {
     position: absolute;
-    top: 0;
-    bottom: 0;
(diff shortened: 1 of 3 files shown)


bhousel:
Thanks @leegenes!   Sorry it took me a while to get to this.
It looks excellent, I just changed it to use a dark style to better match the dark map UI elements, now that the button is not longer on the bright sidebar.

screenshot 2017-09-17 00 40 36

PR #4284, opened 2017-08-27
remove addr:unit from Brazil addresses field

willemarcel:
addr:unit is not common in Brazil addresses


JamesKingdom:
Thanks @willemarcel :+1:, it does seem this is true:

PR #4285, opened 2017-08-27
update select features documentation

willemarcel:
@bhousel you can change the text, if needed


bhousel:
Thanks @willemarcel good catch 👍

PR #4296, opened 2017-08-29
add more fields to Ferry Route preset

willemarcel:
add operator, duration and access fields. https://wiki.openstreetmap.org/wiki/Tag:route%3Dferry


bhousel:
Seems fine, thanks @willemarcel

Issue #4238, opened 2017-08-17
Reflect operation isn't completely invertible


bhousel:
Multiple reflect operations should just flip the shape back and forth in place, not move it around.  This is minor but we should figure out why.  There is probably some math error or rounding issue that causes the axis to shift a little bit.

![reflect](https://user-images.githubusercontent.com/38784/29423471-e2030d20-8349-11e7-96df-2996915687f7.gif)

bhousel:
The one midpoint that doesn't move is a good clue to what's going on... 🕵️‍♀️

bhousel:
closed in #4300

PR #4300, opened 2017-08-29
check final edge in poly hull in reflect action

leegenes:
Not all edges of the poly hull were being checked when determining hull's smallest enclosing rectangle

edit: referencing #4238


bhousel:
Awesome thanks @leegenes - I just tried this out and it works great.

Also congrats on your first contribution and welcome 🎊

PR #4301, opened 2017-08-30
remove addr:unit from address fields for Germany, Austria and Switzerland

manfredbrandl:
addr:unit removed from address fields for Germany, Austria and Switzerland


speedbiker:
Fully agree!

manfredbrandl:
@bhousel can you merge this?

Issue #4313, opened 2017-09-06
Phone number pattern for Benin and Ivory Coast


sommerluk:
Currently, there is no phone number pattern in iD for Benin and Ivory Coast.

The E.123 format (what OpenStreetMap wiki asks for) is:

- for Ivory Coast: +225 ab uvwxyz
- for Benin: +229 ab uvwxyz

“ab” is the area code.

But in real-world life the phone numbers are _always_ structured by groups of two digits:
- for Ivory Coast: +225 ab uv wx yz
- for Benin: +229 ab uv wx yz

I would propose something like
- for Ivory Coast: +225 20 12 34 56
- for Benin: +229 20 12 34 56

(20 is in both countries a landline code.)

bhousel:
Thanks for the info!  This is a great task for someone getting started to add those formats to [`phone-formats.json`](https://github.com/openstreetmap/iD/blob/master/data/phone-formats.json).

manfredbrandl:
Already solved in PR #4314 .

bhousel:
> Already solved in PR #4314 .

Thanks @manfredbrandl!  Sorry I didn't see that, first day back from vacation....

sommerluk:
Thanks!

PR #4314, opened 2017-09-06
Phone formats for Austria, Ivory Coast and Benin

manfredbrandl:
Resolves Issue #4313 and adds Austrian Format also.


sommerluk:
Thanks!

PR #4329, opened 2017-09-08
Adding healthcare presets

### LINKED ISSUE #3589: Presets for Healthcare
According to [this wiki page](https://wiki.openstreetmap.org/wiki/Key:healthcare) there are a lot of presets to add about healthcare:speciality or healthcare

JamesKingdom:
Still needs work, closes #3589


bhousel:
Thanks @JamesKingdom for starting this!  I cleaned up the `healthcare:speciality=*` field, and added a few more of the provider types found on the [OSM `healthcare=*` wiki page](https://wiki.openstreetmap.org/wiki/Key:healthcare) and just merged this.  👍

bhousel:
> Why not "Specialities"? Isn't the correct plural form for BrE?

@maraf24, the tag is `healthcare:speciality`, but we display "Specialities" on the screen and allow the user to pick multiple values.  (This field combines the multiple values together into a single value with semicolons).

screenshot 2017-11-14 08 30 04

maraf24:
Well, that's why I nit-picking: the spelling as seen on the screenshot is American - "Specialties", not "Specialities" :)

bhousel:
> Well, that's why I nit-picking: the spelling as seen on the screenshot is American - "Specialties", not "Specialities" :)

Oh ok - we have an `en-GB` localization of iD, so the best place to make the change is on Transifex:
https://www.transifex.com/openstreetmap/id-editor/translate/#en_GB/presets/102484336?q=specialt

I'll just change this now, since I want to make a bugfix release this week to fix a few OpenStreetCam and Mapillary things.  If you find any other places where iD doesn't look properly British, please go into Transifex and fix it, thanks!

PR #4332, opened 2017-09-09
add preset for hackerspace

willemarcel:
https://wiki.openstreetmap.org/wiki/Talk:Tag:leisure%3Dhackerspace

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -8271,7 +8271,9 @@
                 "point",
                 "area"
             ],
-            "terms": [],
+            "terms": [
+                "bowling center"
+            ],
             "tags": {
                 "leisure": "bowling_alley"
             },
@@ -8461,10 +8463,13 @@
             "fields": [
                 "name",
                 "address",
+                "building_area",
+                "opening_hours",
                 "website"
             ],
             "geometry": [
-                "point"
+                "point",
+                "area"
             ],
             "terms": [
                 "makerspace",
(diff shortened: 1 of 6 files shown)


data/presets/presets.json
```diff
@@ -8456,6 +8456,26 @@
...
         },
+        "leisure/hackerspace": {
+            "icon": "commercial",
+            "fields": [
+                "name",
+                "address",
+                "website"
+            ],
+            "geometry": [
+                "point"
```
bhousel: Thanks @willemarcel, you can `"area"` here after `"point"`.  I know the wiki page says point only, but it's wrong.  All leisure tags that can happen in a building can be mapped as area.  (in fact the summary `leisure=` page shows area also: https://wiki.openstreetmap.org/wiki/Key:leisure )

data/presets/presets.json
```diff
@@ -8456,6 +8456,26 @@
             },
             "name": "Golf Course"
         },
+        "leisure/hackerspace": {
+            "icon": "commercial",
+            "fields": [
+                "name",
+                "address",
+                "website"
```
bhousel: You can add `"building_area"` and `"opening_hours"` here.. Actually I'd probably just copy whatever other indoor leisure presets have, like [`bowling_alley.json`](https://github.com/openstreetmap/iD/blob/master/data/presets/presets/leisure/bowling_alley.json)

willemarcel:
@bhousel thanks for the tips. It's done!

PR #4350, opened 2017-09-14
Add download changes link

### LINKED ISSUE #4346: Add changeset download link to save screen
from https://github.com/openstreetmap/iD/issues/2739#issuecomment-325991845

> @bhousel is there any way to get the changeset data out of iD so it can be preserved and edited then uploaded (perhaps in JOSM)?

We already have this on the conflict resolution screen, so it should be easy enough to add to the general save screen in an unobtrusive way.  It's very rare that someone can't save, but they shouldn't lose their edits because of an iD bug, and might help us in debugging the issue.

bhousel:
(WIP: still need to check IE11 and fix download on conflict screen)
closes #4346

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -260,6 +260,7 @@ en:
     save: Upload
     cancel: Cancel
     changes: "{count} Changes"
+    download_changes: Download OsmChange file
     warnings: Warnings
     modified: Modified
     deleted: Deleted
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -260,7 +260,7 @@ en:
     save: Upload
     cancel: Cancel
     changes: "{count} Changes"
-    download_changes: Download OsmChange file
+    download_changes: Download osmChange file
     warnings: Warnings
     modified: Modified
     deleted: Deleted
@@ -464,7 +464,7 @@ en:
       keep_remote: Use theirs
       restore: Restore
       delete: Leave Deleted
-      download_changes: Or download your changes.
+      download_changes: Or download osmChange file
       done: "All conflicts resolved!"
       help: |
         Another user changed some of the same map features you changed.
--- a/modules/ui/commit_changes.js
+++ b/modules/ui/commit_changes.js
@@ -1,6 +1,11 @@
 import * as d3 from 'd3';
 import { t } from '../util/locale';
+import { JXON } from '../util/jxon';
+import { actionDiscardTags } from '../actions';
+import { osmChangeset } from '../osm';
 import { svgIcon } from '../svg';
+import { utilDetect } from '../util/detect';
+
 import {
     utilDisplayName,
     utilDisplayType,
@@ -9,10 +14,13 @@ import {
 
 
 export function uiCommitChanges(context) {
+    var detected = utilDetect();
+
 
     function commitChanges(selection) {
 
-        var summary = context.history().difference().summary();
+        var history = context.history(),
+            summary = history.difference().summary();
 
         var container = selection.selectAll('.modal-section.commit-section')
             .data([0]);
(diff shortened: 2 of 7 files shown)


modules/ui/commit_changes.js
```diff
@@ -85,6 +93,36 @@ export function uiCommitChanges(context) {
             .on('click', zoomToEntity);
 
 
+        // Download changeset link
+        var changeset = new osmChangeset({ id: 'CHANGEME' }),
```
pnorman: Generally for downloading you don't want to supply anything for changeset ID.

data/core.yaml
```diff
@@ -260,6 +260,7 @@ en:
     save: Upload
     cancel: Cancel
     changes: "{count} Changes"
+    download_changes: Download OsmChange file
```
pnorman: The standard spelling for OSC files is "osmChange"

modules/ui/commit_changes.js
```diff
@@ -85,6 +93,36 @@ export function uiCommitChanges(context) {
...
+            uri = 'data:text/xml,' + encodeURIComponent(data);
+
+        var downloadLink = container.selectAll('.download-changes')
+            .data([0]);
+
+        var enter = downloadLink.enter()
+            .append('a')
+            .attr('class', 'download-changes')
+            .attr('href', uri)                     // no IE11 ?
+            .attr('download', 'changes.osc')       // no IE11 ?
```
pnorman: The standard way is with `Content-Disposition: attachment; filename="changes.osc"` header for downloading content to a file, and I think that works for IE, but I'm not sure what you do when generating a file on the browser to save.

data/core.yaml
bhousel: Thanks, I'll change it.  I was going off of the [osm wiki](https://wiki.openstreetmap.org/wiki/OsmChange) - should we change it there?

modules/ui/commit_changes.js
bhousel: Yeah this is different in client side javascript, see [stack overflow](https://stackoverflow.com/questions/3975648/how-to-set-content-disposition-attachment-via-javascript).  That's why we are using the `download` attribute here, which is now finally supported in all modern browsers.  

I'm going to test today what happens in IE11, but I think we should be able to fallback to the previous behavior, which is opening a new tab with the `data://` uri contents in it (this trick no longer works after Chrome 59, so that's why I had to figure something out quickly - the existing download link on the conflict resolution screen no longer works).

modules/ui/commit_changes.js
bhousel: Can you check this for me?  I'm really not sure what other tools need.

It currently ends up generating something kind of like this, but we can remove the `changeset="CHANGEME"` stuff if needed:
```xml
  <create/>
  <modify>
    <node id="105439801" lon="-74.53666020842094" lat="40.663296528227306" version="2" changeset="CHANGEME"/>
    <node id="105439798" lon="-74.53607265426908" lat="40.66309431604051" version="2" changeset="CHANGEME"/>
  </modify>
  <delete if-unused="true">
    <node id="5106517631" lon="-74.5367492" lat="40.6633484" version="1" changeset="CHANGEME"/>
  </delete>
</osmChange>
```

bhousel:
> UI will need some thought to make sure that people don't try to download as a part of an attempting to save.

It looks like this currently - I think it's pushed far enough below the other stuff that it won't mislead anybody.

screenshot 2017-09-15 11 29 03

modules/ui/commit_changes.js
pnorman: Remove the changeset="CHANGEME" and it looks okay. I'm not 100% sure what programs support reading osmChange for uploads.

data/core.yaml
pnorman: Wiki page titles have to begin with a capital character.

modules/ui/commit_changes.js
bhousel: sounds good, thanks!

bhousel:
Sweet!  I finally got change file downloading working and tested in Chrome, Safari, Firefox, Edge, and IE11.
Hat tip to @keemor's comment [here](https://github.com/mholt/PapaParse/issues/175#issuecomment-201308792) for the solution 🎩

PR #4360, opened 2017-09-20
Changed uiModes Button to Flexbox

tridip1931:
@kepta Can you review this.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -74,10 +74,6 @@ body {
     display: flex;
 }
 
-[dir='rtl'] .limiter {
-    flex-direction: row-reverse;
-}
-
 .spinner {
     opacity: .5;
     float: right;
@@ -281,24 +277,6 @@ table.tags, table.tags td, table.tags th {
 .col10   { float:left; width:83.3333%; }
 .col11   { float:left; width:91.6666%; }
 .col12   { float:left; width:100.0000%; }
-
-
-/* Sizing
-------------------------------------------------------- */
-
-.w0 { width:04.1666%; }
-.w1 { width:08.3333%; min-width: 70px;}
-.w2 { width:16.6666%; }
-.w3 { width:25.0000%; }
-.w4 { width:33.3333%; }
-.w5 { width:41.6666%; }
-.w6 { width:50.0000%; }
-.w7 { width:58.3333%; }
-.w8 { width:66.6666%; }
-.w9 { width:75.0000%; }
-.w10 { width:83.3333%; }
-.w11 { width:91.6666%; }
-.w12 { width:100.0000%; }
 
 /* UI Lists
 ------------------------------------------------------- */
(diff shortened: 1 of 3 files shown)


css/80_app.css
```diff
@@ -277,6 +282,24 @@ table.tags, table.tags td, table.tags th {
...
+------------------------------------------------------- */
+
+.w0 { width:04.1666%; }
+.w1 { width:08.3333%; min-width: 70px;}
+.w2 { width:16.6666%; }
+.w3 { width:25.0000%; }
+.w4 { width:33.3333%; }
+.w5 { width:41.6666%; }
+.w6 { width:50.0000%; }
+.w7 { width:58.3333%; }
```
kepta: Do we need the `.wX` ? I might be wrong but can't we substitute this with `flex-grow` / `flex-shrink`?

tridip1931: I believe flex-grow/flex-shrink only defines how much space to take in a given estate. 
CSS grid helps a lot if we are using lots of widths
I thought we needed width minus the `display: block`  but right now I am not using them anywhere. We can delete it.  @kepta

bhousel: I agree we might not need this, but I think the idea is for places where you have several joined things, (like buttons) you want them to try to share equal width.

e.g. for OK/Cancel, they'd have 50% width each, even though those words have very different natural width.  

I've found I still need to set flex items with a width in order to get them to layout correctly.  `flexbox-grow` is about how to distribute extra width to contained items, and `flexbox-shrink` is about how to steal width from them (I think - I might be wrong too).

kepta: Yup, but I feel we can simulate the case of proportions `%` with flex.
Suppose we want 2 guys to share `50:50`, we can use simply use default flex.
if `75:25`, flex-grow:3 and other to 1.

@tridip1931  @bhousel Is there a situation you guys can think where `width%` can't be replaced ?

tridip1931: Yes @bhousel I also faced this issue while working in flexbox for keeping some widths, upon looking for answers in google, I found this article very helpful. 
https://css-tricks.com/things-ive-learned-css-grid-layout/

TL;DR We should not depend entirely on flexbox for layout purposes. 

How do you feel about this guys? @bhousel @kepta

kepta: @tridip1931 `css-grid` is amazing, but we can't use it yet because of very low [browser support](http://caniuse.com/css-grid).

tridip1931: Oh ya. I will delete those widths. Also will figure out the flex-grow/shrink if I come across. 
Do we use flex-grow anywhere in our code right now?

bhousel: > Do we use flex-grow anywhere in our code right now?

It's part of the shorthand `flex:` property.  The keyboard shortcuts screen is probably the place I've most recently used it:

https://github.com/openstreetmap/iD/blob/d9b3b7b577ce9af3bff3eb5f6220a4685dbd7688/css/80_app.css#L3419-L3432

tridip1931: Thanks @bhousel . Thats very helpful. 👍

bhousel:
closing as stale

Issue #4356, opened 2017-09-18
Flexbox the buttons


bhousel:
This is like #4349 in that sometimes we should be better at layout long things in narrow space.  The solution is usually to just use [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), which gives us more control over how and whether things are allowed to wrap.

Currently our buttons are just floated things with fixed width, but I think we can use flexbox for layout instead.  If the buttons themselves are also flexbox containers for the icon and text, I think we might be able to drop the text labels when there is not room for them, leaving only the icons.  (This is a bit more elegant than just dropping the labels with a media query).

This is at around 750px wide screen.  It's worse for languages that use longer words:
<img width="689" alt="screenshot 2017-09-18 09 36 11" src="https://user-images.githubusercontent.com/38784/30548762-023f7d8c-9c61-11e7-83f0-b4ec0da33ad0.png">

cc @kepta

tridip1931:
sure @bhousel.

tridip1931:
👋 @bhousel 
Created a pull request of the changes. 

I faced 1 issue while building CSS in dist. I had to run `npm run all` everytime for my CSS changes to go to `dist` iD.css. Is there any way I can cut on that?

bhousel:
> I faced 1 issue while building CSS in dist. I had to run npm run all everytime for my CSS changes to go to dist iD.css. Is there any way I can cut on that?

Not currently, unfortunately the live reload server only rebuilds the .js code bundle.  Even with that, I dont think it catches all the stuff under `/data` properly, like presets.

https://github.com/openstreetmap/iD/blob/ccf452acf6f2d97b8f72bc9e8efca3ea8e2f2b4d/development_server.js#L17-L21

I think there are [rollup plugins](https://github.com/rollup/rollup/wiki/Plugins) that can do this stuff, I just haven't had the time to look into it.

bhousel:
@tridip1931 Looks like a good start - can you see if you can get the buttons to be uniform width?  Also check RTL, looks like they're not centering correctly without the label (like there is some margin remaining in there).

tridip1931:
@bhousel Thanks for pointing out the inconsistencies.  Sorry for overlooking it. 

### LTR
<img width="787" alt="screenshot 2017-09-21 15 48 23" src="https://user-images.githubusercontent.com/7458983/30691301-f1789a84-9ee4-11e7-823c-3f3aa4ac7603.png">

### RTL
[...]

PR #4364, opened 2017-09-21
streamline the build process

kepta:
We are not handling CSS bundling in the dev environment. 
It is painful to start the build step again and again.

This PR adds the css concat under the build process.
I got no luck in using rollup css plugins, they seem to not work with caching.

Thanks to @tridip1931 ([ref](https://github.com/openstreetmap/iD/issues/4356#issuecomment-330815271)) for flagging this.


tridip1931:
This is awesome @kepta 💯

bhousel:
Thanks @kepta, looks great!

While you are in this file, can you also make sure it runs [`build.js`](https://github.com/openstreetmap/iD/blob/master/build.js) whenever the preset sources change?  I think that's the other thing missing from the live reload.

kepta:
@bhousel I thought about this, do you think you think it is worth adding the `build.js` considering we update it less frequently than css, js and the added wait time between incremental builds.

bhousel:
> do you think you think it is worth adding the build.js considering we update it less frequently than css, js and the added wait time between incremental builds.

but it would only do this if one of the files changed, right?  It shouldn't add build time always..

kepta:
oh cool, my bad.
Will add it

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-  - "4"
```
bhousel: we still need to test on and support 4 and 6

bhousel:
👎 to yarn

.travis.yml
kepta: @bhousel since it is just the dev side, is there something which is blocking for just going for node 8.
The reason I am pushing for node 8 and ditching 4, 6 is for being able to to use async await in the build process.

bhousel: I'm not comfortable forcing node 8 on anyone else who wants to contribute to iD.  It's important for this project keeping developer friction low.

Anyway, `npm run build` currently takes 5 seconds.  I don't think we need to optimize this.

kepta:
@bhousel had a couple of ideas while working on this PR. Would love to hear your opinion as some of them might be breaking or non-desirable changes.

1. If we could decouple the `*.js` files from the `iD/data` folder. This would help in keeping concerns separate. If the data changes we have to rebuild the `iD/modules`  _(I am guessing this one!, might need to verify)_, but if the `iD/data/*.js` changes we can skip compiling the data again.

2. The compiled `.json` [files](https://github.com/openstreetmap/iD/blob/master/build.js#L41) are sitting with the input json files, making it harder to reason. Keeping the output separate would also help in watching for file changes.

3. The [extra build](https://github.com/openstreetmap/iD/blob/master/package.json#L18) steps in `package.json` make the build process inconsistent. It would be great if we could bring them under the bracket of `build.js` file.

4.  I know we were using a nice `Makefile` which kept things under control. We are doing fine without a proper build tool for now. But I see a growing uneasiness in understanding the entire build process scattered here and there. Should we consider using a build tool like [gulp.js](https://gulpjs.com) ?

bhousel:
Since the css rebuilding was merged as 24fef3cba, let's close this PR and @kepta you can open another one for the follow-on enhancements..

@kepta I'm ok with all your suggestions.. You probably want to start out by adding gulp, then seeing if you can then simplify the build steps and package scripts..

Mostly I just want to preserve the following:

1. iD should be completely buildable with `npm run install` and `npm run all`.  No extra dependencies.
2. This should work on all reasonable versions of node, which is currently 4, 6, 8.

This helps keep development simple, especially so that we can quickly test with confidence on weird environments like Travis or Windows.

Issue #3948, opened 2017-04-05
Aid Treeshaking by importing only what is needed


shaula:
There are two big dependencies within iD, one is lodash, the other is d3. By using:
`import * as d3 from 'd3'` everything is imported. Within a modern setup using Rollup or Webpack2 this hinders Treeshaking to perform well. If only e.g. the dispatch function is used from d3 within an iD module, then its just fine to import:
`import dispatch from 'd3-dispatch';`
If this is done consequently, the build system is able to get rid of unused modules easily, saving tons of KBs. While d3 is already split into various modules, for lodash we would have to switch to https://www.npmjs.com/package/lodash-es
Any thoughts or blockers on this issue?

bhousel:
Yes, when we did #3257 and #3344, I had the idea that we would eventually replace those `import *` with better imports that just target the modules that are really needed.  Thanks for opening this issue!

There aren't any real blockers, but it would probably be good to wait until we release 2.2 and then clear out some of the PR backlog before changing lots of files.

bhousel:
I did some work on this which lives here:  https://github.com/openstreetmap/iD/pull/4372#issuecomment-332292559

Leaving this issue open because the bundle didn't actually shrink at all and there are other blockers now - but maybe we are closer!

PR #4371, opened 2017-09-24
Upgrade to rollup v0.50 [DO NOT MERGE]

bhousel:
This is just to upgrade to rollup v0.50

There is a rollup bug that seems to be affecting our d3 import.:  https://github.com/rollup/rollup/issues/1338 , https://github.com/rollup/rollup/issues/1411

So we can't actually take this upgrade until either the bug is fixed, or we stop using a `*` star import of d3 (this is #3948)

### DIFF
--- a/development_server.js
+++ b/development_server.js
@@ -48,10 +48,10 @@ function build() {
     console.time('Rebuilt');
 
     rollup.rollup({
-        entry: './modules/id.js',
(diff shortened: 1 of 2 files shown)


bhousel:
Looks like around v0.51 tests started passing again, so this older PR is superseded by #4513

PR #4372, opened 2017-09-24
Import only what is needed for lodash and d3

### LINKED ISSUE #3948: Aid Treeshaking by importing only what is needed

bhousel:
This is for converting our `*` star imports of lodash and d3 into more targeted named imports. 

re: #3948


bhousel:
Currently a work in progress.. I've only done actions and lodash so far.

bhousel:
Putting the pause on this for now.. It's a big several day long mess that resulted in:
 
1. bundle size did not shrink at all (maybe even got bigger)
2. tests fail now for mysterious reasons, but seems to be something with `d3.event`
3. require `lodash` for build scripts and `lodash-es` for the bundle (ugh)
4. treeshaking seems to be a lie
5. rollup bugs with name mangling?  (d3 exports `event`, we import `as d3_event`, rollup renames it back to `event` which conflicts with  window.event??)

Need a break from this.. 😞

bhousel:
I restored the global d3 for now so that the tests would be happy, and figured out how to get rollup to resolve the d3 dependencies on node4/npm2.  This returns things to a mergeable state.

There is still some ugly stuff here but I'm going to merge as-is and open follow-on issues for cleanup, since I don't want this huge bunch of work to go stale.

PR #4375, opened 2017-09-24
Adding playground presets

### LINKED ISSUE #4252: Add presets for playground
There are lots of playground presets [here](https://wiki.openstreetmap.org/wiki/Key:playground), that could be added

JamesKingdom:
Closes #4252 

### Presets Added

- Balance Beam
- Basket Spinner
- Basket Swing
- Climbing Frame
- Bouncy Cushion
- Horizontal Bar
- Springy Rocker
- Play Roundabout
- Sandpit
- Seesaw
- Slide
- Playground Structure
- Swing
- Zipwire

### Things to do

- ~~Add a new field for the `swing` preset that says whether it has a special wheelchair seat~~
- [x] Change text for the `baby` field on the `swing` preset 
I changed it from `Designed for babies` to `Baby seat`, but I'm still unsure if I like this or not
- [x] Add some more fields to the `playground` preset, such as operator and age restrictions

### DIFF
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1390,8 +1390,13 @@
         },
         "playground/max_age": {
             "key": "max_age",
-            "type": "text",
+            "type": "number",
             "label": "Maximum Age"
+        },
+        "playground/min_age": {
+            "key": "max_age",
+            "type": "number",
+            "label": "Minimum Age"
         },
         "population": {
             "key": "population",
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1383,6 +1383,11 @@
             "label": "Power Output",
             "placeholder": "500 MW, 1000 MW, 2000 MW..."
         },
+        "playground/baby": {
+            "key": "baby",
+            "type": "check",
+            "label": "Baby Seat"
+        },
         "population": {
             "key": "population",
             "type": "text",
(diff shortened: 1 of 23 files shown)


data/presets/presets/playground/slide.json
```diff
@@ -0,0 +1,11 @@
+{
+    "icon": "entrance",
```
bhousel: Haha not sure how I feel about this one..

data/presets/fields/playground/max_age.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "max_age",
+    "type": "text",
```
bhousel: `min_age` and `max_age` can both be fields and have type `number`.  This will add up/down buttons to the side.

data/presets/fields.json
```diff
@@ -1383,6 +1383,16 @@
             "label": "Power Output",
             "placeholder": "500 MW, 1000 MW, 2000 MW..."
         },
+        "playground/baby": {
+            "key": "baby",
+            "type": "check",
+            "label": "Baby Seat"
```
bhousel: I think this is fine 👍

bhousel:
What a fun PR! 

Per your todos:
- I think we don't need a special wheelchair swing field.. Users can tag with the existing universal `wheelchair` field if it makes sense.
- I think what you did with "Baby Seat" checkbox is good..

I added a few other comments.. I know picking icons is tough, but I think we should not use the "entrance" icon for the slide, even though it does look kind of like a slide.  Using the same icon for 2 very different concepts makes the map harder to use.

PR #4380, opened 2017-09-28
Streamline build steps

kepta:
This PR breaks build steps into three parts.
1. build_css.js
2. build_data.js
3. build_src.js

- All files return a promise, for easier handling.
- One might notice that the build_src time has increased, this is due to logging the time after the build.js has been written. (no performance regression, we were just not counting properly before).
- Also, the `build_data.js` is promisified, I am not confident if I broke something or not (@bhousel ?)


package.json
```diff
@@ -26,13 +26,14 @@
     "translations": "node data/update_locales"
   },
   "dependencies": {
+    "@mapbox/sexagesimal": "1.1.0",
+    "@mapbox/togeojson": "0.16.0",
+    "colors": "^1.1.2",
```
bhousel: colors looks like a devDependency

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: node_js
 node_js:
-  - "4"
```
bhousel: remember to leave node 4 and 6 in the travis matrix

kepta:
@bhousel the `build_data.js` has a bit of breaking changes to make it async, 
Hope I didn't overlook any potential bug.

bhousel:
@kepta I was thinking, can we just switch to [`fs-extra`](https://www.npmjs.com/package/fs-extra) to get promisified filesystem calls plus some other useful stuff?

bhousel:
This is good enough to merge now.. 👍

PR #4389, opened 2017-10-03
Update confectionery presets

### LINKED ISSUE #4388: Add "sweet shop" as a term for shop=confectionery and distinguish from craft
Can we add "sweet shop" as a term for shop=confectionery? It's what they're commonly called in the UK. 

Also, on a related note, when I search "sweet" in iD it comes up with "Confectionery" which is actually craft=confectionery. Can we change the name of this to make it more obvious that this is the craft tag and not the shop tag. Maybe craft=confectionery could be called "Confectionery Producer/Sweet Maker" or similar?

JamesKingdom:
This closes #4388

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -2530,9 +2530,9 @@ en:
         terms: '<translate with synonyms or related terms for ''Clockmaker'', separated by commas>'
       craft/confectionery:
         # craft=confectionery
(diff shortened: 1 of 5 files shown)


bhousel:
perfect, thanks @JamesKingdom 👍

PR #4402, opened 2017-10-06
Disable source switcher during walkthrough

### LINKED ISSUE #4401: Source Switcher should be disabled while walkthrough is active
I know that the source switcher is mostly for local iD installs for debugging, but clicking it while the walkthrough is active throws away the walkthrough graph.

pwelby:
Closes #4401 -- fixed as proposed in the issue's comments


bhousel:
Perfect, thanks @pwelby! 💪

Issue #4410, opened 2017-10-07
shop=trade , shop=tiles presets


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Tag:shop=trade
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dtiles

iblancasa:
Could you add more information about what is this issue about?

Thank you :)

bhousel:
Sure @iblancasa, the preset system is explained in [this README](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md).   The presets and fields are all defined in `.json` files under `data/presets/...`

We're looking for:
- [ ] **A new preset file `data/presets/presets/shop/tiles.json`** - it should look like any of the other files inside the `shop` folder.. You can just copy an existing file like `data/presets/presets/shop/tobacco.json` and change the "name" to say `"Tile Shop"` and "tags" to say `"shop": "tiles"`.
- [ ] **A new preset file `data/presets/presets/shop/trade.json`** - same as with tiles, copy an existing shop preset, change "name" ("Trade Shop") and "tags" (`"shop": "trade"`)
- [ ] **A new field file `data/presets/fields/trade.json`** - This will let people pick from a list of common values what trade the "Trade Shop" is for.  You can copy an existing file like `data/presets/fields/tourism.json`, and change the "key" to `"trade"` and label to `"Trade"`.
- [ ] To make the `trade` field appear on the trade preset, open `data/presets/presets/shop/trade.json` that you just created, and add `"trade"` to the "fields" array.  You can put it after `"name"`, since it is kind of important (the order in the array is the order they'll display on the screen).

bhousel:
Done in #4411, #4413

PR #4411, opened 2017-10-08
Added Tile Shop preset

TheGreenToaster:
Added Tile Shop presets as part of #4410


bhousel:
Thanks for your contribution, @TheGreenToaster this is great! 👍   

If you have the project checked out locally, as a final step you can also `npm run build` to update the generated files too, then include them in the PR.  It's no big deal to skip this step - I or someone can do it too.

PR #4415, opened 2017-10-08
Refine playground preset terminology

1ec5:
A couple improvements to playground presets:

* Renamed the “Springy Rocker” preset to “Spring Rider” and added “spring rocker” as a synonym. Both “[spring rider](https://en.wikipedia.org/wiki/Spring_rider)” and “spring rocker” appear to be more common than “springy rocker” in usage online.
* Added “merry-go-round” as a synonym for “Play Roundabout”.

### DIFF
--- a/data/presets/presets/playground/rocker.json
+++ b/data/presets/presets/playground/rocker.json
@@ -6,5 +6,9 @@
     "tags": {
         "playground": "springy"
     },
-    "name": "Springy Rocker"
+    "name": "Spring Rider",
+    "terms": [
+        "spring rocker",
+        "springy rocker"
+    ]
 }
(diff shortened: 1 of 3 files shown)


bhousel:
@1ec5 , you updated the generated `dist/locales/en.json` file - can you instead update the source .json files:
- `data/presets/presets/playground/rocker.json`
- `data/presets/presets/playground/roundabout.json`

then `npm run build`  (this will overwrite `en.json`)

1ec5:
Thanks, I updated those two JSON files. Running `npm run-script build` produced tons of changes, enough to crash FileMerge. 😬 So I only added the relevant changes to en.json.

bhousel:
Thanks @1ec5!

PR #4418, opened 2017-10-09
Add async storage for saving/getting large data [DO NOT MERGE]

### LINKED ISSUE #3239: Switch to async storage (i.e. indexdb instead of localstorage)

Semigradsky:
Fix #3239

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -9,6 +9,7 @@ import _isString from 'lodash-es/isString';
 import { dispatch as d3_dispatch } from 'd3-dispatch';
 import { json as d3_json } from 'd3-request';
 import { select as d3_select } from 'd3-selection';
+import localforage from 'localforage';
 
 import {
     t,
@@ -101,6 +102,24 @@ export function coreContext() {
             if (typeof console !== 'undefined') console.error('localStorage quota exceeded');
             /* eslint-enable no-console */
         }
+    };
+
+    context.asyncStorage = function(k, v) {
+        var action;
+
+        if (arguments.length === 1) {
+            action = localforage.getItem(k);
+        } else if (v === null) {
+            action = localforage.removeItem(k);
+        } else {
+            action = localforage.setItem(k, v);
+        }
+
+        return action.catch(function(err) {
+            /* eslint-disable no-console */
+            console.error('async storage error', err);
+            /* eslint-enable no-console */
+        });
     };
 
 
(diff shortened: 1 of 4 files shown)


modules/core/history.js
```diff
@@ -577,18 +577,27 @@ export function coreHistory(context) {
...
 
         // is iD not open in another window and it detects that
-        // there's a history stored in localStorage that's recoverable?
+        // there's a history stored in async storage that's recoverable?
         restorableChanges: function() {
-            return lock.locked() && !!context.storage(getKey('saved_history'));
+            var isLocked = lock.locked();
+
+            if (isLocked) {
+                return Promise.resolve(true);
```
bhousel: I'm pretty sure native Promises don't work in IE11.. Can you use localForage's callback API instead?

Semigradsky: Oh, my fault. Rewrited to callback API.

Semigradsky:
@bhousel ping

modules/core/context.js
```diff
@@ -431,7 +455,7 @@ export function coreContext() {
 
     // Debounce save, since it's a synchronous localStorage write,
     // and history changes can happen frequently (e.g. when dragging).
-    context.debouncedSave = _debounce(context.save, 350);
+    context.debouncedSave = _debounce(context.save, 100);
```
bhousel: Is there a reason why you changed this?  (If anything I think we could slow down the saves, not speed them up)

bhousel:
Thanks for being patient @Semigradsky - I think I'm going to move this new code into its own module, and also try to update the other places in iD that use storage to use this new store, and add some code to grab the existing localStorage values if they exist.  Hope to merge it today 👍

bhousel:
Hey @Semigradsky - I did some work on this yesterday but here's where I ended up:

- The switch from sync to async storage is a pretty big deal.  We can do it, (and there are real benefits) but it changes the flow of when stored variables can be available to the iD code
- IndexedDB is actually kind of slow.  I'm reading that get times around like 100-300ms are normal for largish objects.
- Some components in iD need access to the stored variables at startup (think like the splash screen and the history restore)  - however we don't have an ability to _delay_ startup until the stored variables are ready - this would be an API breaking change..
- So considering the current code and that it would be a breaking change, I want to put the pause on this PR until we start work on iD v3.

I don't have a timeline on when it would happen, but it is something we want eventually..

PR #4419, opened 2017-10-09
Treat punctuation as hashtag delimiters

### LINKED ISSUE #4398: Hashtag regex doesn't catch Unicode hashtags
After 9719a31c79844bfc4f39a791fcc687c70794898e, Unicode characters (including emoji ;-) in hashtags are no longer included (they're now treated as word separators):

```
> "#g1b".match(/#[\w-]+/g);
[ '#g1b' ]
> "#g1ü".match(/#[\w-]+/g);
[ '#g1' ]
```

I've been using `/(#[^\d][^#\s,;:]*)/g` as my current regex of choice (with the dubious assumption that hashtags shouldn't start with numbers).

mojodna:
Unicode ranges for punctuation are simpler than creating a Unicode-aware word class, so delimit on non-words.

Fixes #4398

### DIFF
--- a/modules/ui/commit.js
+++ b/modules/ui/commit.js
@@ -26,6 +26,10 @@ var readOnlyTags = [
     /^locale$/
 ];
 
+// treat all punctuation as hashtag delimiters
+// from https://stackoverflow.com/a/25575009
+var hashtagRegex = /(#[^\u2000-\u206F\u2E00-\u2E7F\s\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]*)/g;
+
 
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @mojodna!   Looks pretty good.. I merged it and then made 2 other little changes in 8f82caf55:
- I removed the `_`, so hashtags can have underscores in them
- I changed the quantifier from `*` to `+` so it would match 1 or more instead of 0 or more.  The original would detect a single lonely "#" character, even in cases where the hashtags string was empty.

mojodna:
Thanks. I was being spacey when I put that together; too much of a hurry

PR #4422, opened 2017-10-09
Allow -,_,+,& as hashtag characters

mojodna:
@bhousel your comment in #4419 made me realize that these characters also valid: `-`, `+`, `&` (and in use)


bhousel:
LGTM 👍

PR #4427, opened 2017-10-11
Added plot preset

humanforklift:
I have added a new presets JSON file for place=plot. 

I'm new to this, so let me know if I need to change anything.

### DIFF
--- a/data/presets/presets/place/plot.json
+++ b/data/presets/presets/place/plot.json
@@ -1,0 +1,17 @@
+{
+    "icon": "triangle-stroked",
+    "fields": [
+        "name"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "place": "plot"
+    },
+    "terms": [
+        "plot"
+    ],
+    "name": "Plot"
+}
--- a/data/presets/presets/place/plot.json
+++ b/data/presets/presets/place/plot.json
@@ -11,7 +11,10 @@
         "place": "plot"
     },
     "terms": [
-        "plot"
+        "tract",
+        "land",
+        "lot",
+        "parcel"
     ],
     "name": "Plot"
 }
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -41258,13 +41258,13 @@
             ],
             "suggestion": true
         },
-        "amenity/pharmacy/Adler-Apotheke": {
+        "amenity/pharmacy/Adler Apotheke": {
             "tags": {
-                "name": "Adler-Apotheke",
+                "name": "Adler Apotheke",
                 "healthcare": "pharmacy",
                 "amenity": "pharmacy"
             },
-            "name": "Adler-Apotheke",
+            "name": "Adler Apotheke",
             "icon": "pharmacy",
             "geometry": [
                 "point",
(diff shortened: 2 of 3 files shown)


data/presets/presets/place/plot.json
```diff
@@ -0,0 +1,17 @@
...
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "place": "plot"
+    },
+    "terms": [
+        "plot"
```
bhousel: You don't actually need to include the word "plot" as a search term - it will be included automatically because it is the tag value.

Instead, add synonyms here like `["tract", "parcel", "lot"]`

bhousel:
Thanks so much for your contribution!  The `plot.json` file looks great - I requested a minor change for the search terms array.

humanforklift:
Thanks for letting me know. I have updated the search terms array, so hopefully now it should contain synonyms as requested.

bhousel:
Thanks for your contribution @humanforklift , looks great 👍

Nakaner:
I don't think that adding a preset for `place=plot` is a good idea. As another user pointed out on IRC yesterday, mapping plots might lead the OpenStreetMap project into legal trouble with data protection regulation in the European Union and some other countries in Europe. Having data which falls under these regulation in the planet could hinder distribution and usage of the data in several countries.

By adding the preset to the "official" editor (it is embedded into the official web page of the project) people could have the impression that it is permitted [1] to map them. In addition, the presets asks for a name tag. What will the name of mapped parcels be? I fear that it will be the name of the owner.

I abstain from writing down more arguments against this tag because this bug tracker is not the tagging mailing list.

Therefore, I would like to ask all who are involved to reconsider their decision.

A technical comment: Why is the diff of this pull request so large? The pull request aims to add a single tagging preset but changes lots of other files. Is this necessary?


[1] I know the "you can tag what you want roule"

bhousel:
> By adding the preset to the "official" editor (it is embedded into the official web page of the project) people could have the impression that it is permitted [1] to map them.

There are 28k of them already mapped and it has an osm wiki page.  It's an established tag, so that's why we added a preset.  I try to avoid make decisions about whether something "belongs" in OSM or not.  If people want it mapped and they are using legal sources, that is good enough for me.

> In addition, the presets asks for a name tag. What will the name of mapped parcels be? I fear that it will be the name of the owner.

Presets like "Building" and "House" support a name field too.  It's also discouraged to put people's real names there, but nobody is calling to remove the building presets just because a few mappers might make that mistake.

I do think [adding rules to OSMCha](https://github.com/mapbox/osmcha-frontend/issues) like "Name added to Building" or "Name added to Plot" would be a good way to surface issues like this to local reviewers.

I don't actually know how the tag is used, but if `name` really doesn't make sense, we could remove it. 
 Also I think adding a `ref` field might make more sense.  I know this differs all around the world - where I live plots are usually referred to by Block/Lot numbers.

> I abstain from writing down more arguments against this tag because this bug tracker is not the tagging mailing list.

Feel free to discuss on the mailing list.. I'll see it, but I usually don't participate unless somebody tries to suggest a tag scheme that is technically difficult for us to support.  If the OSM community decides to remove all of these from the database and wiki, I'll happily remove the preset too.

> A technical comment: Why is the diff of this pull request so large? The pull request aims to add a single tagging preset but changes lots of other files. Is this necessary?

The user who submitted the pull request is a new contributor (thank you!!), and rebuild the generated presets file using outdated name-suggestion and editor-imagery dependencies.  This is a minor issue, so I just rebuilt them myself (2bb7b5a7d) using updated dependencies.

humanforklift:
Apologies if I caused any confusion - definitely still getting the hang of this. How can I avoid using outdated name-suggestion and editor-imagery dependencies in the future?

Issue #4430, opened 2017-10-12
"substance=" field to pipeline preset


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Key:substance

bhousel:
Sounds good to me 👍 

@TheGreenToaster - want to take care of this?
It would be similar to the steps you did for the trade shop in #4410..

- [ ] `git pull` and `npm install` to update your local copy of the repo
- [ ] copy a combo field like `data/presets/fields/site.json` to a new field `data/presets/fields/substance.json`, then change the "key" to say "substance" and the label to say "Substance"
- [ ] add "substance" to the "fields" array in `data/presets/presets/man_made/pipeline.json`
- [ ] `npm run build` to update the generated files

xmile1:
Hi @bhousel, I am new to this project, If its okay for me to work on this let me know, I have my dev env setup.

bhousel:
> Hi @bhousel, I am new to this project, If its okay for me to work on this let me know, I have my dev env setup.

Yes @xmile1 go ahead and work on it 👍   Thanks!

xmile1:
so to be clear, following those highlighted step are what I need right?

bhousel:
> so to be clear, following those highlighted step are what I need right?

Yes @xmile1 you got it... Check the [README under data/presets](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md) if you want some more context, but it is really just adding one new .json file and editing another existing .json file.

bhousel:
closed in #4432

PR #4432, opened 2017-10-12
feat(preset): add substance field to pipeline preset

xmile1:
- add substance field to pipeline preset
https://github.com/openstreetmap/iD/issues/4430


bhousel:
This is perfect, thanks for adding the field @xmile1 👍 


screenshot 2017-10-13 10 51 06

PR #4439, opened 2017-10-14
add the features api parameter which enables the list of features

### LINKED ISSUE #4393: Add API parameter to conditionally enable/disable feature filters
This would let referring sites to start iD with certain filters enabled or disabled e.g. only shows roads - or whatever they want the user to focus on editing.

see #2524 for a similar issue.

ferdibiflator:
Fixes #4393

`features` api parameter is added for enabling the features in the list (`,` is the splitter)

Also the hash is being updated when the user toggles features

### DIFF
--- a/API.md
+++ b/API.md
@@ -34,6 +34,10 @@ in the hash portion of the URL:
   _Example:_ `hashtags=%23hotosm-task-592,%23MissingMaps`
 * __`rtl=true`__ - Force iD into right-to-left mode (useful for testing).
 * __`walkthrough=true`__ - Start the walkthrough automatically
+* __`features`__ - Enables features in the list.<br/>
+  _Example:_ `features=water,service_roads,points,paths`<br/>
+  _Available features:_ `points` `traffic_roads` `service_roads` `paths` `buildings` `landuse`
+  `boundaries` `water` `rail` `power` `past_future` `others`
 
 ##### iD on openstreetmap.org (Rails Port)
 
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @ferdibiflator - this looks really good!   I'll check it out some more tomorrow, in the meantime can you also update the [API.md doc](https://github.com/openstreetmap/iD/blob/master/API.md#url-parameters) "url parameters" sections to include the new option?

ferdibiflator:
@bhousel I found a problem with `boundaries` feature. At the moment this one may not be enabled through the url parameter `features`.
Because the `boundaries` is disabled within the `/index.html`:
```javascript
id = iD.Context()
            .assetPath('dist/');

id.features().disable('boundaries');

id.ui()(document.getElementById('id-container'), function() {
```
I suggest to disable this one in accordance with some logic within the `modules/renderer/features.js`. But I don't know in which cases the `boundaries` should be disabled.

ferdibiflator:
Hello @bhousel. Any news?

bhousel:
> Any news?

This looks good, and I'll merge it any day now, thanks for being patient. 
I'm merging about 1 PR/day but for this one I would like to check out the branch locally and test.

bhousel:
Hey @ferdibiflator I tried it out and I think I'm going to change it so that the url querystring contains the hidden features, instead of the shown features.. It's just a little less wordy that way.  What you did is great, so I'm just going to merge as-is and make the change in a followup commit 👍

bhousel:
btw @ferdibiflator re  this:

> I found a problem with boundaries feature. At the moment this one may not be enabled through the url parameter features

I worked around this in 30ff68348 by changing the code in `index.html` to only disable boundaries if there is no `disable_features` list in the querystring.  This means that if somebody _really_ wants the boundaries enabled, they can add an empty param like `&disable_features=`.  

I mostly consider `index.html` as sample code that anyone is free to modify if they are reusing iD for other things.

PR #4440, opened 2017-10-14
push the gpx src (file name or url) to imagery_used

### LINKED ISSUE #4385: Don't write "Local GPX" in imagery_used changeset tag if that's not true
Some trace ids should be recorded in changesets
https://github.com/openstreetmap/openstreetmap-website/issues/1645#issuecomment-333304771

ferdibiflator:
Fixes #4385


bhousel:
Thanks @ferdibiflator, I just tested this and it works perfectly 👍 
I found another issue in testing, but it's a different bug and I'll ticket separately.

PR #4445, opened 2017-10-15
Fix #4443 Imagery only updated on first save

### LINKED ISSUE #4443: Imagery used is only updated the first time the user attempts to save
Because of how this code is structured below, the `imagery_used` changeset tag is getting set only the first time a user attempts to save.

This means if a user cancels the save and does some more editing, whatever other imagery sources they use won't be included in the tag.

https://github.com/openstreetmap/iD/blob/fd4ca2368d0b28bd9968ccec5adaaedd5c79004d/modules/ui/commit.js#L63-L85

The fix is really easy - just remove
`imagery_used: context.history().imageryUsed().join(';').substr(0, 255)`
...from where the `tags` get set up inside the `if` block.

Instead move it just after the `if` block so the value gets updated every time this code is entered.
After the tags are cloned, edit the `imagery_used` property, and update the changeset like below.
```js
tags = _clone(changeset.tags);
tags.imagery_used = context.history().imageryUsed().join(';').substr(0, 255);
changeset = changeset.update({ tags: tags });
```

NB: In case anyone is wondering, why bother with the clone/update stuff instead of just modifying `changeset.tags` directly - it's because we try to keep these osm objects immutable.

moshen:


bhousel:
Just tested this and it works great, thanks @moshen 🙇

Issue #4378, opened 2017-09-27
Don't depend on 2 lodashes


bhousel:
followup from https://github.com/openstreetmap/iD/pull/4372#issuecomment-332292559

We use `lodash` (CommonJS) in a few of our build scripts, but we now use `lodash-es` (ES6) in the main iD code.  It seems dumb to install lodash twice, so we have 2 options:

1. Eliminate use of lodash from the build scripts?  The build scripts need to run in node 4 and up and we don't use lodash that much.
2. Find a way to `require` the es6 `lodash-es` library from a node script.  [`@std/esm`](https://github.com/standard-things/esm) may offer some way forward but I couldn't get it to work after playing with it for an hour or so.  It's only released a few weeks ago, so we should wait and see if it improves.

Semigradsky:
3. Add `babel` (for running build scripts in node 4 and using cool es2015+ features) and `babel-plugin-lodash` (for eliminate use paths in imports).

@bhousel? I can help if you agree to this.

bhousel:
> Add babel (for running build scripts in node 4 and using cool es2015+ features) and babel-plugin-lodash (for eliminate use paths in imports).

I'd really prefer not to need to transpile the build scripts or introduce a babel dependency.  These scripts really don't use much lodash at all, so it should be possible to replace whatever they do with traditional ES5 features.

DzikowskiW:
@bhousel I worked on the issue today. I was able to set up `@std/esm` to work along with build scripts, which led to removing lodash dependencies from the build scripts. 

However, many of the tests that run in PhantomJS depend on the original lodash (`_.map`, `_.clone`, `_.isEqual`, `_.extend`, `_.reduce`, `_.find`,` _.intersection`). If you look at `test/index.html` there is a link to lodash. How do you want to handle that? Should we have a separate file with helper functions for tests, or should we try to rewrite the tests not to use any of these functions?

bhousel:
@DzikowskiW good question... Those functions seem like ones we could do without.. If you submit a PR to deal with the build scripts, I'll eliminate lodash from the tests.

bhousel:
> I'll eliminate lodash from the tests.

This was done!  

Going forward, no lodash in tests, plz thx.  All the tests should run in all of iD's supported browsers -
 you can test this by starting a local server and browsing to `localhost:8080/test/`

Automated testing currently runs in PhantomJS 2.1 - which is conveniently about the same level of oldness as IE11.  😭 

This means:
[...]

PR #4447, opened 2017-10-16
Lodash dependencies removed from build scripts

DzikowskiW:
1. @std/esm added to handle ES6 imports in older node version
2. Swapped lodash dependencies for lodash-es in build scripts (#4378)


bhousel:
Thanks @DzikowskiW!

PR #4456, opened 2017-10-20
Copyable content text

### LINKED ISSUE #4406: Content in info panels is not copyable
It would be nice to have current object id displayed somewhere and have a possibility to copy it.
The current workarround is  to open it in the openstreetmap.org and copy from the url, but that's not a comfortable thing.

ferdibiflator:
Fixes #4406

The decision allows users to copy any text nodes within the app with the copy hotkeys

### DIFF
--- a/modules/behavior/copy.js
+++ b/modules/behavior/copy.js
@@ -48,7 +48,7 @@ export function behaviorCopy(context) {
 
 
     function doCopy() {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @ferdibiflator, just tried this and it works great 👍

PR #4470, opened 2017-10-25
Add "Climbing hut" term to alpine hut

### LINKED ISSUE #4469: Add search term to tourism=alpine_hut
https://github.com/openstreetmap/iD/blob/8bcd44618869828fa5381b43f554f422ac3afcaa/data/presets/presets/tourism/alpine_hut.json

- In the UK, "climbing hut" is what we would describe tourism=alpine_hut as. Can we add "climbing hut" as a term?
- ~~The address section is greyed-out (see screenshot) in iD, I understand that some won't have postal addresses but many do, can we fix this?~~
alpine_hut

Edit: address issue just seemed to be a glitch - it's working now :)

JamesKingdom:
Closes #4469

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -4942,6 +4942,7 @@ en:
       tourism/alpine_hut:
         # tourism=alpine_hut
         name: Alpine Hut
+        # 'terms: climbing hut'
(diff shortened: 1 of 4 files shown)


bhousel:
Thanks @JamesKingdom !

PR #4475, opened 2017-10-26
Don't add unneeded underscores to source

### LINKED ISSUE #4474: Unneccesary underscores added to source
When writing a changeset comment and adding a source, iD will automatically add underscores to the source tag. I believe that this is actually an unneccesary feature; most editors (e.g.: JOSM) will apply the source tag without the underscores and the wiki says nothing about it. As this is bad for the readability, can we remove this feature?

source

source field

octagonal:
Resolves #4474

### DIFF
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1721,6 +1721,7 @@
             "icon": "source",
             "universal": true,
             "label": "Sources",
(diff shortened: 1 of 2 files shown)


bhousel:
Great, thank you @octagonal 👍

Issue #4476, opened 2017-10-26
preset for landuse=religious


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Tag:landuse=religious

abdeldjalil09:
merged in #4478

PR #4478, opened 2017-10-27
Add landuse=religious preset

YuliyaVeklicheva:
for #4476

### DIFF
--- a/data/presets/presets/landuse/religious.json
+++ b/data/presets/presets/landuse/religious.json
@@ -2,6 +2,9 @@
   "geometry": [
     "area"
   ],
+  "fields": [
+    "name"
+  ],
   "tags": {
     "landuse": "religious"
   },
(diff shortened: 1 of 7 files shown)


data/presets/presets/landuse/religious.json
```diff
@@ -0,0 +1,10 @@
+{
+  "geometry": [
+    "area"
+  ],
+  "tags": {
```
bhousel: This is great thank you!
Can you also add a name field into this file too?
```
    "fields": [
        "name"
    ],
```

bhousel: (and then run `npm run build` again)

YuliyaVeklicheva:
@bhousel updated

bhousel:
Great thank you @YuliyaVeklicheva !
The name shows up strangely because there is no icon, but that is a different bug with the label placement code that I will ticket separately.

PR #4484, opened 2017-10-28
Merge way nodes

### LINKED ISSUE #3774: Merge points
With multiple points selected, should be possible to merge them.

merge points

ferdibiflator:
Fixes #3774

There is an ability to merge way endpoints


bhousel:
This looks pretty good - I have some questions..
- It looks like it only merges end nodes.  I know that matches with the description in #3774, but I think it would be useful to merge any nodes (for example to merge one end node with one internal node to make a 'T' junction).
- What does it do with tagged nodes? - it should probably merge all the tags too.
- Can you add a test for this new action under `test/spec/actions/` ?

ferdibiflator:
@bhousel You are right. I didn't pay attention. I'll work on it.

ferdibiflator:
@bhousel can you check it pls?

ferdibiflator:
Thanks for feedback @Nakaner . Agree with you. I'll work on it

slhh:
Deleting such ways as @Nakaner has mentioned can also be a problem.

- Valueable features might get lost and the user wouln't always be aware, for example if turn lane was tagged on the two-node way.
- Some features/properties of such a way might need to be transfered to the merged node, e.g a highway=platform.
- Relations might be broken, e.g. when deleting `from` or `to` members of a turn restriction. `via` membership needs to be transfered to the merged node if no other via member is left. Multipolygons might become invalid because inner members get connected to outer members.

bhousel:
Just a quick update - the work I just did to close #4921 is relevant here.  We now have some pretty good checks to prevent node merges from breaking relations and turn restrictions.  I'm going to try to get this merged after releasing iD v2.8

bhousel:
Ok I rewrote some of the code to reuse `actionConnect` which already handles all of the cases in the comments above.   Merging!

PR #4492, opened 2017-10-31
Do not capitalise English words.

jaller94:
This PR tries to decapitalise four English words in the interface of iD. In my opinion their capitalisation is not justified as they are not directly referring to a feature name of the app.

E.g. "Zoom in to edit" where edit is the standard English word. If it was referring to the Edit feature it would have to be "Zoom in to use Edit".


bhousel:
Thank you @jaller94, I agree these read much better 👍

PR #4495, opened 2017-11-02
Avoid exporting all of d3.js just so we can use it in tests

darthgera123:
Commented out the bit and made the required addition of the script file


bhousel:
That's the first step - now the challenge is to get the tests to pass!

darthgera123:
just check with the latest commit please

bhousel:
> just check with the latest commit please

Travis still shows that the tests are failing.. If you run them locally in your dev environment, it will probably give you some messages about what the issues are that need to be fixed.

bhousel:
Closing here, as this isn't going anywhere.

PR #4500, opened 2017-11-04
add preset for music schools

tyrasd:
see https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dmusic_school


bhousel:
Thanks @tyrasd 👍

PR #4504, opened 2017-11-04
It is sculptor not sculpter

simonpoole:
I believe this is a fairly clear typo in the presets

### DIFF
--- a/data/presets/presets/craft/sculptor.json
+++ b/data/presets/presets/craft/sculptor.json
@@ -12,7 +12,7 @@
         "area"
     ],
     "tags": {
-        "craft": "sculpter"
+        "craft": "sculptor"
     },
-    "name": "Sculpter"
+    "name": "Sculptor"
 }


bhousel:
Thanks @simonpoole good catch!
I'm surprised it has been this way for so long.

Looks like there are 291 of `craft=sculpter` to be cleaned up..
https://taginfo.openstreetmap.org/search?q=craft%3Dsculpter

We should wait until the next iD release and then after the fix goes live, I'll change them.

abdeldjalil09:
and also these tags : "craft=sweep" it should be "chimney_sweeper"

simonpoole:
@abdeldjalil09 sweep is IMHO a legit short form

abdeldjalil09:
see https://wiki.openstreetmap.org/wiki/Tag:craft%3Dchimney_sweeper

bhousel:
> see https://wiki.openstreetmap.org/wiki/Tag:craft%3Dchimney_sweeper

Thanks @abdeldjalil09 I did not know that the tag had changed back in 2014 based on a brief discussion on the tagging@ list.  Please open a new issue.

simonpoole:
In any case it is a wonderful example of, even if completely well intentioned, you shouldn't change tags without really good reasons. And if you do, just changing the wiki is not enough.

abdeldjalil09:
>  Please open a new issue.
I'm going to do it.

PR #4507, opened 2017-11-06
add shop=agrarian preset and agrarian=* field

willemarcel:
https://wiki.openstreetmap.org/wiki/Tag:shop=agrarian

### DIFF
--- a/data/presets/presets/shop/agrarian.json
+++ b/data/presets/presets/shop/agrarian.json
@@ -2,7 +2,7 @@
     "icon": "shop",
     "fields": [
         "name",
-        "brand",
+        "operator",
         "agrarian",
         "address",
         "building_area",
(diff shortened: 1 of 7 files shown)


data/presets/presets/shop/agrarian.json
```diff
@@ -0,0 +1,27 @@
+{
+    "icon": "shop",
+    "fields": [
+        "name",
+        "brand",
```
bhousel: Can you remove `brand`, otherwise looks ok...

None of our other shop presets offer this field because we don't have a good way to suggest values for it.  Users would need to just know how to create a semicolon separated list of brands sold here.

willemarcel:
@bhousel I removed the `brand` field. Thank you!

PR #4514, opened 2017-11-08
corrected a spelling mistake

Lukas458:
corrected "relation": "Dies kann nicht getrennt werde, da es Mitglieder einer Relation verbindet." to "relation": "Dies kann nicht getrennt werden, da es Mitglieder einer Relation verbindet."


bhousel:
Hi, thanks for the report. Our translations are handled through [Transifex](https://www.transifex.com/openstreetmap/id-editor/dashboard/), so the way to fix this is to sign up for an account there and make the change.  The `de.json` file will actually be replaced by whatever Transifex has the next time we make a release of iD.

hikemaniac:
I fixed this on Transifex. 👍

bhousel:
> I fixed this on Transifex.

Thank you!

Issue #4529, opened 2017-11-14
Add more fields to "Telephone"


JamesKingdom:
https://wiki.openstreetmap.org/wiki/Tag:amenity=telephone lists some more fields that could be added

bhousel:
This was done!

PR #4535, opened 2017-11-16
added common fields for preset amenity=telephone

Vonter:
added common fields in telephone preset(for #4529)
https://wiki.openstreetmap.org/wiki/Tag:amenity=telephone


bhousel:
Thanks @Vonter this looks great 👍

PR #4539, opened 2017-11-18
Kiosk name update

tohaklim:
Wiki makes no mention of it being specifically a "news" kiosk, describing it instead as selling various things. Is this country-specific usage?


bhousel:
Thanks for the update, I agree  👍

PR #4468, opened 2017-10-24
New help texts

manfredbrandl:
I give it a fresh try.

PR #4557, opened 2017-11-22
Updates to help texts

### LINKED ISSUE #4018: Help should be rewritten and split into smaller chunks
The help texts in the application all need to be rewritten to relate to the changed UI with the new right click concept for editing selected features.

Also, I noticed some discrepancies between terminology used in the current help texts and the walkthrough. Here is the most obvious;

- "inspector" vs "feature editor"
- "tool menu" vs "edit menu"
- "edit operations" vs "(merge/move etc.) tool"

Would you accept a PR with patch to data/core.yaml? It looks like the authoritative place to edit by skimming the project structure and build.js.

bhousel:
This is just a cleaned up version of #4468 to make things easier to review
closes #4018
(thanks @1ec5 👍 )


data/core.yaml
```diff
@@ -552,284 +552,158 @@ en:
...
-      button next to the relation you want to remove it from.
-
-      You can create multipolygons with holes using the "Merge" tool. Draw two areas (inner
-      and outer), hold the Shift key and click on each of them to select them both, and then
-      press the 'C' shortcut key. Other option is to select both, and then right-click one
-      of the and click the "Merge" (+) button.
+    help:
+      title: Help
+      welcome: "Welcome to the iD editor for [OpenStreetMap](https://www.openstreetmap.org/). With this editor you can update OpenStreetMap right from your web browser."
+      open_data_h: "### Open Data"
```
1ec5: Perhaps as tail work, can we define the heading levels somewhere outside these strings? That would make it less error-prone to translate the help UI.

data/core.yaml
```diff
@@ -552,284 +552,158 @@ en:
...
+      open_data: "Edits that you make on this map will be visible to everyone who uses OpenStreetMap. Your edits can be based on personal knowledge, on-the-ground surveying, or imagery collected from aerial or street level photos. Copying from commercial sources, like Google Maps, [is strictly forbidden](https://www.openstreetmap.org/copyright)."
+      before_start_h: "### Before you start"
+      before_start: "You should be familiar with OpenStreetMap and this editor before you start editing. iD contains a builtin walkthrough to teach you the basics of editing OpenStreetMap. Click 'Start the Walkthrough' on this screen to take the tutorial - it takes only about 15 minutes."
```
1ec5: Since this is for the base American English localization, we should probably use double quotation marks consistently. The British English localization can use single quotation marks as usual.

bhousel:
> Translators are going to experience a bit of inconvenience updating their translations in Transifex this time around, but future updates will go a lot more smoothly than before. I think we should blast out a message to translators that they can take a look at the iD source code if they’re missing translation memory.

Yes!  As soon as this merges, I'll announce to everyone that we need help with translations.  I'm not anticipating a need to release a new version of iD for a little while (maybe mid December?), so that will give everyone time to work on it.

data/core.yaml
bhousel: Good suggestion, I removed the markdown from the titles, but left it here, and it feels messy. I'll think of a way.

bhousel:
Happy to say I just merged this 👍   This will make it so much easier to maintain and translate the help texts.

Thanks @manfredbrandl and @1ec5 for your help with this.. I will make some announcements to request translation help.

Issue #3041, opened 2016-03-19
Improve public_transport tagging


afmenez:
_highway=platform_ was obsoleted in 2011 (see http://wiki.openstreetmap.org/wiki/Tag:highway%3Dplatform ), but the ID editor still suggests it when adding a highway.
I suggest we keep the interface as it is, but change the tag added to public_transport=platform instead.

More details on the available option in http://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_stop .

d1g:
Not just public_transport=platform, but all 3-4 tags from vastly approved proposal: http://wiki.openstreetmap.org/wiki/Proposed_features/Public_Transport#Stop

IMO iD should treat older highway=platform/bus_stop/random tags as secondary.and public_transport=* family as primary.

Problem with highway=platform/bus_stop is that they are directly contradict [a highway=* tag definition: importance of road in a highway graph](https://wiki.openstreetmap.org/wiki/Proposed_features/Highway_key_voting_importance).

Relevant discussion: https://github.com/gravitystorm/openstreetmap-carto/issues/435#issuecomment-199094591

bhousel:
I'm confused by this issue.. iD does not have a preset for `highway=platform`.

We do have one preset for `railway=platform` and one for `public_transport=platform`.

d1g:
There separate paragraph you can use to quick search: http://wiki.openstreetmap.org/wiki/Proposed_features/Public_Transport#Compatibility_with_well_known_tags - it says about following mapping:

No 1 railway case
railway=platform <-> public_transport=platform

trivial info: transportation mode (bus=yes, tram=yes) is omitted because nobody would use buses/trams in railways

No 2 highway case: is slightly more complex (see first 2 lines in table)

bhousel:
@afmenez, @d1g, thanks for pointing me at the documentation for the `public_transport` tag. I'm ok with adding some of the new tags alongside the existing tags where it makes sense to do so, and adjusting the preset descriptions if it can help reduce mapper confusion, but I don't think it makes sense to remove any of the existing tags that are in widespread use.

I'll leave this issue open for now and let's try to make a list of actionable steps to improve the presets.

afmenez:
Would it be possible to use a new platform template and automatically add the new tags whenever a highway=platform or railway=platform is edited or inserted?

bhousel:
[...]

PR #4566, opened 2017-11-26
add public_transport tag to the bus_stop preset

nlehuby:
This PR add a field to the bus stop preset, to fill the public_transport tag
ptv2

Do I need to do something more to handle the translations ?

I did not manage to get the tests run, it gets stuck at the `iD.uiFieldWikipedia` doing nothing 😣

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -1102,6 +1102,14 @@ en:
       product:
         # product=*
         label: Products
+      public_transport_stop:
+        # public_transport=*
+        label: Stop type
+        options:
+          # public_transport=platform
+          platform: Where the passengers wait
+          # public_transport=stop_position
+          stop_position: Where the bus stops on the road
       railway:
         # railway=*
         label: Type
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1468,6 +1468,17 @@
             "type": "semiCombo",
             "label": "Products"
         },
+        "public_transport_stop": {
+            "key": "public_transport",
+            "type": "combo",
+            "label": "Stop type",
+            "strings": {
+                "options": {
+                    "platform": "Where the passengers wait",
+                    "stop_position": "Where the bus stops on the road"
+                }
+            }
+        },
         "railway": {
             "key": "railway",
             "type": "typeCombo",
--- a/data/presets/fields/public_transport_stop.json
+++ b/data/presets/fields/public_transport_stop.json
@@ -1,0 +1,11 @@
+{
+    "key": "public_transport",
+    "type": "combo",
+    "label": "Stop type",
+    "strings": {
+        "options": {
+            "platform": "Where the passengers wait",
+            "stop_position": "Where the bus stops on the road"
+        }
+    }
+}
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -6660,6 +6660,7 @@
                 "name",
                 "network",
                 "operator",
+                "public_transport_stop",
                 "bench",
                 "shelter"
             ],
--- a/data/presets/presets/highway/bus_stop.json
+++ b/data/presets/presets/highway/bus_stop.json
@@ -4,6 +4,7 @@
         "name",
         "network",
         "operator",
+        "public_transport_stop",
         "bench",
         "shelter"
     ],
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -2172,6 +2172,13 @@
                 "product": {
                     "label": "Products"
                 },
+                "public_transport_stop": {
+                    "label": "Stop type",
+                    "options": {
+                        "platform": "Where the passengers wait",
+                        "stop_position": "Where the bus stops on the road"
+                    }
+                },
                 "railway": {
                     "label": "Type"
                 },


bhousel:
My concern about this is that it would allow *any* of the `public_transport=` values to be tagged on *any* of the `highway=bus_stop`, and from reading the [bus stop wiki page](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_stop), it looks like there are some combinations that are accepted and some that are not.  

Maybe a better solution would be to split the bus stop preset into two presets - a vertex one that would automatically add `public_transport=stop_position` and another one for all other geometries that would automatically add `public_transport=platform`.  They could both be labelled "Bus Stop" so that the user doesn't need to care about which preset they are using, and the `public_transport` complexity can be hidden from them.

bhousel:
Just realized this relates to #3041 too

nlehuby:
Ok. I like very much your counterproposal, this would be even simpler for the user 👍 

How do we move on with this ? Do I need to convert this to an issue, so that we can have a proper discussion ? Can I change my current PR to do what you proposed ?

slhh:
>Maybe a better solution would be to split the bus stop preset into two presets - a vertex one that would automatically add public_transport=stop_position

I agree, but we need to limit this to vertexes of a road in order to allow an automatic separation from the platform one.

> and another one for all other geometries that would automatically add public_transport=platform.

I think the platform node can also be a vertex, e.g. of a footway, kerb, shelter etc.

>They could both be labelled "Bus Stop" so that the user doesn't need to care about which preset they are using, and the public_transport complexity can be hidden from them.

Hiding public transport complexity is quite desirable, but using the same preset name seems to add even more complexity because both presets have to be handled differently in view of role and position in the route relations. In addition a common name for both presets would prevent a smooth transition to other or mixed public transport modes.

bhousel:
Hmm, yeah @slhh good points.

What if we made the 2 presets "Bus Stop Platform" and "Bus Stop Position"... platform can be any geometry, and position must be vertex only.   For now, we can't enforce that the stop position vertex must be a road.

slhh:
Some of the approved public transport rules are a nightmare.
We do potentially have the following compliant cases of a bus stop feature:

1. A vertex on the road representing the whole bus_stop:
  - highway=bus_stop
  - public_transport=stop_position (optional)
  - bus=yes (optional)
2. A platform node:
  - highway=bus_stop
  - public_transport=platform (optional)
  - bus=yes (optional)
3. An optional stop position to be used in addition to the platform node of 2.:
  - public_transport=stop_position (mandatory)
  - bus=yes (mandatory)
4. A platform way or area (can also be node, but not prefered here):
  - public_transport=platform (mandatory)
  - bus=yes (mandatory)
5. A mandatory stop position to be used in addition to the platform of 4.:
  - highway=bus_stop
  - public_transport=stop_position (optional)
  - bus=yes (optional)

1. and 5. need to share a preset due to using the same tag, but assigning a preset name which is teaching the user to use it for case 1 and 5, but not for case 3 seems to be hard.
In general it seems to be impossible to assign presets to all these cases and teach a typical user to use the right one.
I think we need to drive the community to get things simplified.
I will discuss some approches to simplify in subsequent posts.

slhh:
First simplification approch:

The public transport proposal didn't deprecate highway=busstop, but it seems to be designed to do so later.

If we omit/deprecate highway=busstop, the optional public transport tags 
become mandatory and we can condense the 5 cases to 2 cases:

1. A vertex on the road:
  - public_transport=stop_position (mandatory)
  - bus=yes (mandatory)
2. A platform node, line, or area:
  - public_transport=platform (mandatory)
  - bus=yes (mandatory)

We could call the presets "Bus Stop Position" and "Bus Stop Platform" then.

Unfortunately, there is a drawback in omitting the highway=bus_stop tag. Rendering bus stops based on the public transport tags is quite hard, and currently not supported by the Carto style. We would likely need another change to public transport to simplify this rendering.

slhh:
Second simplification approch:

If we omit/deprecate highway=busstop for bus_stop with mapped platform, but keep it for the vertex representing the whole bus_stop, we can condense the 5 cases to 3 cases:

1. A vertex on the road representing the whole bus_stop:
  - highway=bus_stop
  - public_transport=stop_position (optional)
  - bus=yes (optional)
2. A platform node, line, or area:
  - public_transport=platform (mandatory)
  - bus=yes (mandatory)
3. An optional stop position to be used in addition to the platform of 2.:
  - public_transport=stop_position (mandatory)
  - bus=yes (mandatory)

We could call the presets "Bus Stop (simplified)" and "Bus Stop Platform", and "Bus Stop Position" then.

The rendering issue would be easy to be solved because highway=bus_stop and public_transport=platform plus bus=yes can both be rendered as a bus stop.

slhh:
Third simplification approch:

The highway=bus_stop tag is currently defined for nodes only. If we break with this rule, and allow it to be used for lines and areas too, we can always put it on the platform, where a platform is mapped.
There are already 5000 highway=bus_stop tag used on lines or areas in the database, and Carto does already render the tag for areas.

We can condense the 5 cases to 3 cases then:

1. A vertex on the road representing the whole bus_stop:
  - highway=bus_stop
  - public_transport=stop_position (optional)
  - bus=yes (optional)
2. A platform node, line, or area:
  - highway=bus_stop
  - public_transport=platform (optional)
  - bus=yes (optional)
3. An optional stop position to be used in addition to the platform of 2.:
    public_transport=stop_position (mandatory)
    bus=yes (mandatory)

We could call the presets "Bus Stop (simplified)" and "Bus Stop Platform", and "Bus Stop Position" then.

bhousel:
> What if we made the 2 presets "Bus Stop Platform" and "Bus Stop Position"... platform can be any geometry, and position must be vertex only. For now, we can't enforce that the stop position vertex must be a road.

I'm just going to do something like this ☝️   It seems easy and I think this matches up with @slhh's third comment.  Thanks for kicking off this discussion @nlehuby 👍

Issue #4559, opened 2017-11-23
Remove the default building=yes for marketplace preset


maraf24:
The current preset for marketplace defaults to building=yes. Careless or inexperienced mapper misses the feature to turn the building tag off which leads to creation of non-existing building in the market square, as can be seen [here](https://www.openstreetmap.org/way/460319699) 
Even [the wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dmarketplace) focuses on open-air marketplaces, with building=yes listed only as 'tag used in combination'.
Therefore I think it would be better to not add this tag automatically.

bhousel:
This is a great task for someone who wants to learn how the iD presets work.

They are all defined in `.json` files.. The contents of the `.json` file define which fields iD will display.

* open the file `data/presets/presets/marketplace.json`
* change the line from `building_area` to say `building`
https://github.com/openstreetmap/iD/blob/434bc5af960d23ebb8e32b35e26d507893d9d6e7/data/presets/presets/amenity/marketplace.json#L7
* `npm run build` 

The difference is: `building_area` is a field that defaults to yes if the user has drawn an area, and `building` is just a normal combo field that doesn't have a default value, but the user can still pick one if they choose to.

bhousel:
Done in #4568

PR #4568, opened 2017-11-27
Remove default building=yes for marketplace preset

tastrax:
Remove the default building=yes for marketplace preset #4559


bhousel:
Thanks again @tastrax! 👍 

If you can get a local dev environment set up, remember to also run `npm run build` to update the generated files.

Issue #4145, opened 2017-07-11
Switch Mapillary signs to use SVG icons instead of foreignobject


bhousel:
Sign detection icons on the map are appearing a little blurry.  I think they are being scaled now instead of being displayed at their native resolution.  This probably happened in #4050 when switching from traffico to the spritesheet.  We just have to determine the correct dimensions and change code in a few places to make them sharp again.

<img width="728" alt="screenshot 2017-07-11 11 35 28" src="https://user-images.githubusercontent.com/38784/28076560-1c1f01ec-662d-11e7-8b11-dbb748eba7a8.png">

bhousel:
done in #4148

bhousel:
Turns out that the svg source files themselves are in a different source repository:
https://github.com/mapillary/mapillary_sprite_source

We can just use these instead of the generated spritesheet.  This gives us:
* sharp icons at any resolution
* can stop using `<foreignObjects>` and just use our `svgIcon` class 💥 
* support for all browsers now

bhousel:
Switched the icons to svg!  They look so much better, and there are many more of them too.

<img width="1025" alt="screenshot 2018-06-07 01 14 46" src="https://user-images.githubusercontent.com/38784/41080158-1f0c8a62-69f3-11e8-80a8-7f3bae438174.png">

Issue #2785, opened 2015-09-21
Replace png assets with svg


bhousel:
These things are all available in svg and will look much better on retina.  Using svg would also eliminate the extra build step generating the png.

- [X] line preset icons -> move to `sprite.svg`
- [X] relation preset icons -> move to `sprite.svg`
- [X] Maki
- [ ] cursors?  check browser support first.
- [X] cleanup `/svg` folder

capturing convo w/@samanpwbb

bhousel:
Rolling most of this into #2811, will merge what is done in the next few days, and consider svg cursors another time.

sisskind:
@bhousel Having an issue getting the SVG icons to appear.  Have run the ```npm install``` command and make clean && make, but just getting blank icons.  Is there something I am missing?

bhousel:
We did have an error in the makefile that was causing the svg sprite to not get built unless `svg-sprite` was in your path.  Are you testing this with latest code from `master`?

You can also try `rm -rf node_modules` and the redo the `npm install` step.  This sometimes can clear things up, e.g. if you got stuck with an unlucky combination of dependency versions.

tmcw:
According to [this document](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_User_Interface/Using_URL_values_for_the_cursor_property), SVG cursors aren't supported in WebKit and [this testcase](http://bl.ocks.org/tmcw/3478294ed1ea794f552e6f0f011bd3f3) confirms that.

bhousel:
Thanks for checking!  We'll close this..

PR #4573, opened 2017-11-29
Optimised images.

grischard:
Saved 91 KB out of 762 KB. 20.5% per file on average (up to 56.8%).

For every file, [ImageOptim](https://github.com/ImageOptim/ImageOptim) compared the different tools and picked the smallest file that was produced by a combination of `advpng`, `pngout` and `zopfli` for PNGs, `svgo` for SVGs and `gifsicle` for GIFs.

No related issues, no code change, no functionality change, no visual differences in images (compression is lossless).


bhousel:
This seems ok, but a bunch of these images aren't used anymore and we should just remove them anyway.  Some thoughts:

* The 2x cursors don't seem to be used currently.  (we should fix that)
* Really the only things that matter for optimization are the assets in `dist/`
* Everything under `docs/` is very old and needs to go away or be redone
* The one `svg/iD-sprite.src.svg` file definitely does not need to be optimized, because it's an intermediate step generated file.  It's later used to generate a different file which contains `<symbol>` elements in `dist/img/iD-sprite.svg` as part of the build process
* We've had issues with `svgo` optimizing away our symbol spritesheet.  I opened an issue svg/svgo#743  - anyway just be careful using `svgo`, some of the rules are disabled on openstreetmap-website because of this bug.

Can you change this PR to be just the images under `dist/`

grischard:
Happy to! I'll try to get around to it today.

grischard:
Couldn't most of those images in dist/ be SVGs? Do we still have the vector source for them somewhere?

bhousel:
> Couldn't most of those images in dist/ be SVGs? Do we still have the vector source for them somewhere?

We already went through in #2785 and moved most of the image assets into the SVG spritesheet.

- We can't switch the cursors because not all browsers support SVG cursors (see #2785)
- I'm not sure whether we can switch the patterns or not.  They are small anyway, and there might be some browser compatibility issues (see comments around https://github.com/openstreetmap/iD/blob/master/modules/svg/areas.js#L11-L41), so it hasn't been worth it to me to investigate. 
- We can't switch animated objects like the loading spinners
- We can switch the traffic signs used by Mapillary to SVG, and this is being tracked on #4145 
- `arrow-icon.png` is not used and can be removed (I think)

I think that covers everything!

grischard:
https://github.com/yangshun/awesome-spinners has some cool non-gif spinners, and the current spinner isn't optimised for retina. Maybe I'll take that as a project for a hacking weekend :).

Could this be merged now as it is?

bhousel:
> We can't switch animated objects like the loading spinners

Another thought on this:  technically, we _can_ but it's not a drop in replacement, and probably not worth it to redo the image and wire up a css transform and sort out all the cross browser issues, just to have a  loading spinner that takes up a bit less space and looks a little nicer on retina displays.  If you want to ImageOptim them though, that's cool.

bhousel:
> https://github.com/yangshun/awesome-spinners has some cool non-gif spinners, and the current spinner isn't optimised for retina. Maybe I'll take that as a project for a hacking weekend :).

Cool, yeah if you really want to do this, I won't say no 👍 

Watch out for `<animateTransform>` which [isn't supported on IE](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/animateTransform)

bhousel:
I took a look and this seems fine, thanks @grischard 👍

Issue #2211, opened 2014-04-27
Sharpen the Background aerial imagery


adivik2000:
I sharpen the imagery on potlatch when the imagery is a bit hazy or the locality is densely populated - helps a lot when adding streets. Is this left out on iD on a deliberate note?

bhousel:
Would like to reopen this following discussions at SOTM-US with some HOT folks.  This would really be useful for tracing.

CSS filters offer a way do do this without needing to worry about CORS, and [support in modern browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/filter#Browser_compatibility) is pretty good.

Live example here: http://stackoverflow.com/questions/29599794/how-to-sharpen-an-image-in-css

For example, the below rule bumps up the contrast and brightness slightly on all loaded tiles, and this does make a noticeable improvement in some places.
```
img.tile-loaded {
  opacity: 1;
  -webkit-filter: contrast(120%) brightness(120%);
}
```

We could make a very minimal UI for this in the background pane similar to the opacity switcher.
Thoughts?

tmcw:
@mikelmaron Do you think this is a top priority for HOT, or a nice-to-have?

bhousel:
This was just merged!  (It only took about 4 years, haha)

Now iD can adjust brightness, contrast, saturation, and sharpness (implemented as an unsharp mask) using sliders on the background pane. 

Please check out #4575 for more info, and try it out at http://preview.ideditor.com/master/

(I just realized that while it should work on _most_ browsers, I need to check all of them).

![imagery adjustment](https://user-images.githubusercontent.com/38784/33491848-76a62dda-d689-11e7-988c-3e48d216c8cb.gif)

iandees:
Nice!

A feature suggestion: it would be neat if these settings could be passed in via URL params so that HOT managers could find good settings and send them to the browser directly.

bhousel:
> A feature suggestion: it would be neat if these settings could be passed in via URL params so that HOT managers could find good settings and send them to the browser directly.

Yes, some things that would be great include:
- pass in settings via url hash
- let users save them somehow
- maybe stand up a service like the imagery offset database that will adjust imagery for the user automatically

Not sure what's next, but I'm planning to let users try it out a bit before deciding..

Issue #4570, opened 2017-11-28
Using uiDisclosure messes up the url


bhousel:
This happens because the uiDisclosure is uses an `<a>` and we need to `preventDefault` the click on it.

![disclosure_bug](https://user-images.githubusercontent.com/38784/33302717-4425a0ee-d3cc-11e7-85f3-878989b60817.gif)

Issue #4553, opened 2017-11-21
Manually entered imagery offset is ignored


bhousel:
You're supposed to be able to type the imagery offset in meters into the `input` field in the center of the offset adjustment control, but looks like manually entered values are ignored.

<img width="200" alt="screenshot 2017-11-21 11 20 05" src="https://user-images.githubusercontent.com/38784/33083672-04721b4e-ceae-11e7-8420-ab8a29800113.png">

Issue #4546, opened 2017-11-20
Change "Where does this imagery come from?" link


bhousel:
The "Where does this imagery come from?" link is strange and confusing.  I think it would be good to give users a link to learn more about imagery.  But with the current design, the imagery overlay checkboxes are positioned as if they answer that question.

Q: "Where does this imagery come from?"
A: "Locator Overlay"
🙅‍♂️ 

<img width="288" alt="screenshot 2017-11-19 23 07 58" src="https://user-images.githubusercontent.com/38784/33001863-8c6de9a0-cd7e-11e7-96cc-d2324d9ed42d.png">

manfredbrandl:
Maybe "Where does this background come from?" directly below "Background"?

bhousel:
I made a few design changes so the pane content works better: 
- changed the sections to be expandable
- added more separation between sections
- reworded the link text


<img width="295" alt="screenshot 2018-01-03 15 19 28" src="https://user-images.githubusercontent.com/38784/34538240-aa2a96d0-f099-11e7-9e43-a23923b03fcc.png">

PR #4575, opened 2017-11-29
Enhancements to Background, Map Data, Help panes

bhousel:
I've been working on these for a few days, so I'm going to just roll a few semi-related things into one PR.

- [x] #4601
- [x] #4553 
- [x] #4570 
- [x] #4546 
- [x] #4635
- [x] #2211  😏


bhousel:
The .gif doesn't do this justice, but here's what I've been working on the past few days.  These sliders make it possible to improve any imagery. I'm pretty excited about it 😄  

This feature uses CSS filters to close #2211, something that's been in the back of my mind for a while.  It will work well in Chrome and Firefox and some partial to full support in Safari, Edge, and not sure about IE11.

imagery adjustment

This spun off from an effort to slightly improve the usability of the side panes.  The sliders replace the old "brightness buttons", which were a bit confusing to people (they did not look like buttons) and could only do steps from 25%, 50%, 70%, 100%.

boothym:
Talking of the background pane - is it possible to do something like in this image https://github.com/openstreetmap/iD/issues/3453#issuecomment-258199010 and hide all background > 20 years old in a "historic imagery" dropdown? Because the background list is now quite long for mappers in the UK.

bhousel:
> Talking of the background pane - is it possible to do something like in this image #3453 (comment) and hide all background > 20 years old in a "historic imagery" dropdown? Because the background list is now quite long for mappers in the UK.

This is a really good idea.. I don't want to extend the scope of the PR right now but I do want to think about letting people customize the background imagery list.. e.g. add more / or remove the ones that they never want to see.   I'll reopen and add some comment to #3453

bhousel:
This was merged!  As always, please test out the new Background pane on http://preview.ideditor.com/master/ - thanks 🙇

PR #4598, opened 2017-12-06
add a field preset for the brewery tag

nlehuby:
screenshot-2017-12-6 id


bhousel:
Nice, thanks!  🍻 
Would this also be useful to be added to `bar.json` and `biergarten.json` presets too?

nlehuby:
sure !

Issue #4455, opened 2017-10-19
Ways do not self-connect unless there is a node


neuhausr:
When creating features like a cul-de-sac, an iD user will create a line with a loop on the end. If the user connects the loop back to the line at a node, it will properly connect. But if the user tries to connect the loop back to the line at a place that isn't a node (for example, think of a p or d shape), it will not actually connect. That is, the user will then be able to move the end node without the line there also moving. 

I would expect the end node to connect back to the line whether there is a node at that location or not.

(I have noticed many of these cases when looking through the Self-intersecting Ways in OSM Inspector)

![image](https://user-images.githubusercontent.com/3976407/31776223-8c74c45a-b4b0-11e7-8466-c0a534a186fc.png)

bhousel:
I agree, there is a bug here preventing a new way from self connecting.

bhousel:
done in #4602

Issue #4271, opened 2017-08-25
Street name labeling obscuring node features


simonpoole:
See http://www.openstreetmap.org/edit?editor=id#map=21/47.40755/8.54858 there is actually a pedestrian crossing already there, but the node can't be seen (a slightly more prominent icon would address the issue too).

bhousel:
I agree, the label placement should avoid tagged nodes..

pnorman:
![image](https://user-images.githubusercontent.com/1190866/30236358-c3ba382c-94cc-11e7-9ec1-c1e29477f793.png)

![image](https://user-images.githubusercontent.com/1190866/30236369-d5d25cec-94cc-11e7-9081-7ef49e1bfb6d.png)

bhousel:
Relevant code is [here](https://github.com/openstreetmap/iD/blob/master/modules/svg/labels.js#L404) for anyone who wants to work on this..

Nitiquita:
Hi @simonpoole, can you clarify what is being obscured by the street name? Thanks!

bhousel:
Hi @Nitiquita this code is a bit advanced for a get started issue, but I'll try to break down what's happening.

This code in `map.js` draws all the map stuff - `drawLabels` is a render function that gets passed all the map data in an array called `entities`:
https://github.com/openstreetmap/iD/blob/7e98d7f9893068fb8f5a2404df328e9a42f6fa34/modules/renderer/map.js#L296-L302

Inside `drawLabels`, we have a spatial index called `rdrawn` (see the [rbush repo](https://github.com/mourner/rbush) for details) - it will store bounding boxes around where the text labels can go.  There are lots of calls to `tryInsert()` scattered throughout this file - this is how we add labels to the `rdrawn` index as long as they don't collide with anything else.
https://github.com/openstreetmap/iD/blob/7e98d7f9893068fb8f5a2404df328e9a42f6fa34/modules/svg/labels.js#L275-L278

One of the first things the `drawLabels` code does is it loops through all the `entities` to decide which ones get labels:
https://github.com/openstreetmap/iD/blob/7e98d7f9893068fb8f5a2404df328e9a42f6fa34/modules/svg/labels.js#L293-L316

☝️ My thought is that _inside_ the `if (geometry === vertex) { ... }` block,  you could add code like this to block out some space around "interesting" vertices, so that no labels will get drawn on top of there:
```js
if (entity.hasInterestingTags()) { 
  var coord = projection(entity.loc);
[...]

Issue #4013, opened 2017-05-05
prevent drawing invalid areas (spikes, self crossing)


Klumbumbus:
During work in maproulette challanges regarding buildings with spikes I noticed that there are alot buildings with the same error, e.g. https://overpass-api.de/achavi/?changeset=48431373
A possible scenario I can think of is that new users might do not know how to end the drawing of an area in iD resulting in such invalid areas. See possible scenario below:
![01](https://cloud.githubusercontent.com/assets/7115609/25754414/56e993a2-31bf-11e7-9c94-71b1c44964e5.gif)
The node order is AB

bhousel:
done in #4602

Issue #3610, opened 2016-11-23
Area tools allows creation of false inners


simonpoole:
Seems as if some users are using the area tool in a slightly unexpected way:

![fake_inner](https://cloud.githubusercontent.com/assets/1011860/20559494/09703f60-b175-11e6-9492-a6ff0d5fa5b1.png)

While not strictly illegal in the OSM data model, it is likely something that iD should catch in one way or the other (only the outer node is shared in the example). The question is, naturally, if iD should create the inner ring automatically or not. In any case joining nodes as in this example should not be possible.

pnorman:
Topologically, is it the same as

![image](https://cloud.githubusercontent.com/assets/1190866/20584002/6c02ef82-b1a3-11e6-975a-854a07e992a1.png)

It seems likely that they're trying to do this

![image](https://cloud.githubusercontent.com/assets/1190866/20584019/93fd6012-b1a3-11e6-9eec-ec7fa86431e7.png)

So maybe the issue is that the method to create areas with holes isn't discoverable?

note: the correct way to represent the latter is

![image](https://cloud.githubusercontent.com/assets/1190866/20584044/d7255f20-b1a3-11e6-8d6b-fa787069d793.png)

But this requires manually editing the tags.

1ec5:
Perhaps we could add a validator warning about self-intersecting areas, by checking whether a node ID is repeated within a way? There are some valid self-intersecting areas, such as the one that @pnorman points out. But I wonder if we can detect this case:

> only the outer node is shared in the example

Let’s call the repeated node A. In this example, a node adjacent to A is far from A but near another edge of the area. Sounds like something that could be cobbled together using [turf-point-to-line-distance](https://github.com/Turfjs/turf/tree/master/packages/turf-point-to-line-distance/), but I haven’t considered performance.

bhousel:
After I merge #4602, users won't be able to draw self intersecting polygons anymore.  

It will still be possible for someone really dedicated to manually construct a multipolygon like this out of ways and relations, but at least casual users will be discouraged from creating them with the area drawing mode.

That still doesn't really solve the problem of training users to do the right thing to create a hole.  So I'll leave this issue open until we have added some better help text and/or training in the walkthrough.

PR #4602, opened 2017-12-07
Support rendering `direction` on vertices (stop sign, traffic_signals, etc)

### LINKED ISSUE #3815: Adjust rendering of node/vertex with direction
Followup from #3115..

Some node/vertex features in OSM can have a direction.  stop signs, traffic signals, billboards.
The normal method to tag this is `direction=forward` or `direction=backward` (although, this being OSM, there are of course [alternative bizarro tag schemes](https://wiki.openstreetmap.org/wiki/Key:direction)).

Anyway, what we do with directional vertices currently is - the same as any other interestingly tagged vertex, just show it as a circle with a dot (or in some cases a maki icon):

screenshot 2017-02-03 10 35 28

But maybe we can do something else to indicate that there is a direction:

screenshot 2017-02-03 10 35 28 copy  screenshot 2017-02-03 10 53 21

bhousel:
- [x] #3003
- [x] #3815 
- [x] #4271 
- [x] #4455 
- [x] #4013 
- [x] #3610
- [x] update walkthrough
- [x] update presets
- [x] fix turn restrictions editor
(updated: scope creep)

Something I threw together today - I still might adjust the marker style a bit, and I'm not 100% sure if it handles all of the tag cases correctly.

#### `highway=stop` with `direction=forward` / `direction=backward`
screenshot 2017-12-07 16 42 35

#### This is what it does for `stop=all` or `direction=both`
screenshot 2017-12-07 16 43 04


iandees:
An arrow is kinda confusing to me. Is that the direction of travel or the direction the sign would be facing?

bhousel:
> An arrow is kinda confusing to me. Is that the direction of travel or the direction the sign would be facing?

Yeah, this is a hard problem.  Currently, the arrowhead faces whatever the direction tag says: "Forward" or "Backward" - so that's "direction of travel".  

Even those words are not great - I don't think an untrained mapper would really be able to understand this concept, or even why stop signs (and even moreso traffic signals) would need a direction.  The help text is not very helpful either.

screenshot 2017-12-07 19 02 22

I'd definitely welcome some ideas on how to make this easier.

bhousel:
Some other design thoughts.. whatever we do needs to:
- have good contrast on top of everything
- be distinct from both the oneway markers and the streetlevel photo view cones
- look good very close to or under the vertex (even when the vertex has a selection halo)

bhousel:
@iandees I do kind of like the viewfield a little better, what do you think?

screenshot 2017-12-08 11 58 32screenshot 2017-12-08 11 58 21

iandees:
Yea, that makes more sense I think. I was also going to suggest something like that but with a flat line perpendicular to the incoming way representing the top of the sign/signal.

bhousel:
So this functionality is turning out to be pretty cool.  I have directions rendered for a bunch of things now (street signs, cameras, etc).

The main issue remaining is that a point with directions disappears when dragged.

nodrag-directional

I know why - it's because vertices aren't displayed during a drag, which is also #3003.

bhousel:
This PR is pretty wild... Here's the overview:

- I wanted to show directions on vertices and put some view cones around them to do that (#3815). This is useful for mapping street signs and traffic signals.
- Then I wanted to put the view cones on standalone nodes too, to support surveillance cameras and stuff that is not on a way.
- To make that look good, the nodes needed to render like vertices (circles), not markers.  We already do this in wireframe mode, so I thought "no big deal".
- This worked ok, but then you couldn't drag the nodes around because we don't render vertices while dragging (#3003).
- So I made the vertices visible during dragging.  (also rotating, moving, etc)
- Now with vertices being rendered, having text drawn on top of them while the user dragged them around was a big issue, so I had to fix the label placement. (#4271).
- Rendering the vertices while dragging and drawing broke a bunch of other things.  I needed to get much more strict about which things can be targets and where the pointer events can go (only to certain vertices and segments along a way).
- So I added invisible touch targets to control which things can listen to pointer events, and a debugging mode to make them visible and see what's going on.  
- These touch targets are circles around the vertices and MultiLineString geojson segments along the ways, and the targets allowed me to again say where it is ok for a line to loop back on itself (#4455)
- If we can draw touch targets, we might as well also draw "nope" targets (which are parts along the line or vertices that the user specifically can't interact with.  
- This let me prevent the user from drawing self-intersecting polygons or lines that return back on themselves (#4013).
- Then I realized we didn't need a lot of complicated drawing code anymore, which used temporary nodes and segments to get around the problems of targeting the way that the user was drawing. (2be62ff is pretty wild)
- Then I realized that if we can draw targets for connecting and for not connecting, we can have the not-connecting targets do useful things like change the cursor and put the line back where it was.
- I made it transitionable, because it looks cool:  

nope2

Anyway that's what is new.  It will merge soon!
Some things that we can do later:
- Use the nope targets to prevent users from joining things that should not join (e.g. power lines with roads) or from dragging things where they should not go (e.g. a junction that zig-zags across several roads)
- Not having the temporary drawing objects anymore simplifies the draw code a lot!  This will make it easier to do square building drawing mode.  (yes one of my next priorities for early 2018)

iandees:
Great work, @bhousel! I've been following along on your commits and this has gotten wild. What a great way to end the year. Happy Holidays! 🍰 🎉

1ec5:
🎉

if you give a mouse a cookie

bhousel:
Happy New Year!   Time to merge this thing already!  🎉🎆🎊🍾 

It will show up in http://preview.ideditor.com/master shortly, and I encourage anyone who is interested to try it out there...

fidelis-assis:
Hi,

I like the new viewfield for direction but aren't forward and backward reversed?

bhousel:
> I like the new viewfield for direction but aren't forward and backward reversed?

I don't think they are reversed.  For example, a stop sign that affects forward travel will be pointed backwards so the cars can see it:

screenshot 2018-02-20 09 48 35

But if you found a situation that we are rendering incorrectly, please open an issue and me know!

fidelis-assis:
2018-02-20 11:49 GMT-03:00 Bryan Housel <notifications@github.com>:

> I like the new viewfield for direction but aren't forward and backward
> reversed?
>
> I don't think they are reversed. For example, a stop sign that affects
> forward travel will be pointed backwards so the cars can see it:
>
Thanks for your quick answer. Your explanation is clear but the value used
for direction and the viewfield shown confuse me. If the stop sign points
backwards relatively to the way direction and I'm mapping its direction
(where it faces to), why tag it with 'forward' as in this example and not
'backward'?

In your example, the viewfield looks backwards (I think), what is OK, after
all the stop sign faces backwards, but the value for direction is
'forward'. What am I missing?

-- Fidelis Assis

bhousel:
> If the stop sign points
> backwards relatively to the way direction and I'm mapping its direction
> (where it faces to), why tag it with 'forward' as in this example and not
> 'backward'?

Oh yeah, I left out an important detail in that screenshot above.  The `direction=forward/backward` tag works relative to the direction _that the way is drawn_, not the direction of the sign.  A funny thing about these tags - if you reverse the way, iD also needs to (and does) reverse the `direction=*` tag.

See:  
- https://wiki.openstreetmap.org/wiki/Key:direction#Forward_and_backward
- https://wiki.openstreetmap.org/wiki/Tag:highway=stop#Tagging_minor_road_stops

The main reason that I started rendering these viewfields is because it's nearly impossible (especially for casual mappers) to understand how the direction apply to a way.

fidelis-assis:
2018-02-20 14:32 GMT-03:00 Bryan Housel <notifications@github.com>:

> If the stop sign points
> backwards relatively to the way direction and I'm mapping its direction
> (where it faces to), why tag it with 'forward' as in this example and not
> 'backward'?
>
> Oh yeah, I left out an important detail in that screenshot above. The
> direction=forward/backward tag works relative to the direction *that the
> way is drawn*, not the direction of the sign.
>
OK, I understand that. Just to make myself clearer, let's consider a oneway
road with a stop sign (or a traffic calming - that was my initial
motivation, I was confused when checking a few tagged with forward). On a
oneway road, all stop signs and traffic calmings face backwards because the
drawing direction is the same as the traffic flow (perhaps a few old
exceptions). So, I was thinking they, if tagged with direction (unnecessary
for one-way roads anyway, also for two-way roads when affect both
directions), the value should be *backward*. But the viewfield in this case
shows the opposite direction.

In the example below, both traffic calmings were tagged with forward,
probably because the mapper wanted to align the viewfield correctly. Again,
in this case the direction isn't necessary.

Ex:
https://www.openstreetmap.org/edit?node=349338170#map=19/-23.02126/-43.44967

I extract speedcams for GPS from OSM. If 'forward' is the right value in
these cases I'll have to change how I calculate direction when it matters
(ex: traffic calming or speed sensor acting on a single direction on a
two-way road).

A funny thing about these tags - if you reverse the way, iD also needs to
> (and does) reverse the direction=* tag.
>
Nice. Does it also reverse if the value is in degrees?


> See:
>
>    - https://wiki.openstreetmap.org/wiki/Key:direction#
>    Forward_and_backward
>    - https://wiki.openstreetmap.org/wiki/Tag:highway=stop#
>    Tagging_minor_road_stops
>
> The main reason that I started rendering these viewfields is because it's
> nearly impossible (especially for casual mappers) to understand how the
> direction apply to a way.
>
Yes, I think it's a very helpful facility. Thanks for adding it!

-- Fidelis

PR #4620, opened 2017-12-17
Removed extra 'be' from en.json, en-GB.json, and core.yaml

SeanBarber:
Three files contained an extra be, before the word also. I have removed this typo.


bhousel:
Thanks @SeanBarber!  Yay first time contributor 🎉 

It's not a problem to include `en-GB.json` in this PR, but that file will be overwritten when I make a new release of iD.  So somebody will need to fix the `en-GB` translation on Transifex also (@JamesKingdom?).

JamesKingdom:
I've updated this on transifex. :+1:

PR #4621, opened 2017-12-17
Short changeset comment: Change style and add warning

### LINKED ISSUE #4613: focus to changeset comment not obvious (enough)
I finished my edits. 
When I click "Save" the map greys out and the left panel becomes "Upload to OpenStreetMap".
Everything looks pretty grey and washed out. Now I wonder, did my click on "Save" was enough? What happens now?
changesetcommentid

The cursor moved automatically to the box "Changeset Comment" but it is not obvious enough to my taste. The mention "(required)" is very discrete. 

Writing a changeset comment is the next action.
I would like the box "Changeset Comment" with some highlight (color, fill, soft blink, ...) as long as the text is empty.

changesetcommentidcolor

ref: The only similar issue I could find is old https://github.com/openstreetmap/iD/issues/588

althio:
Change style and add warning when comment is empty/short

resolves #4613

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -269,6 +269,7 @@ en:
     created: Created
     about_changeset_comments: About changeset comments
     about_changeset_comments_link: //wiki.openstreetmap.org/wiki/Good_changeset_comments
+    short_changeset_comment_warning: "Please provide a concise and meaningful description of your contributions."
     google_warning: "You mentioned Google in this comment: remember that copying from Google Maps is strictly forbidden."
     google_warning_link: https://www.openstreetmap.org/copyright
   contributors:
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -344,6 +344,7 @@
             "created": "Created",
             "about_changeset_comments": "About changeset comments",
             "about_changeset_comments_link": "//wiki.openstreetmap.org/wiki/Good_changeset_comments",
+            "short_changeset_comment_warning": "Please provide a concise and meaningful description of your contributions.",
             "google_warning": "You mentioned Google in this comment: remember that copying from Google Maps is strictly forbidden.",
             "google_warning_link": "https://www.openstreetmap.org/copyright"
         },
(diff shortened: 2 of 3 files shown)


althio:
I am happy with my current result and my testing so far.
I would appreciate your review, if result suits you or if code needs some work.

bhousel:
> am happy with my current result and my testing so far.
> I would appreciate your review, if result suits you or if code needs some work.

25 character comment threshold is way too high.. I sometimes write a comment like "traced some buildings" and I think that's fine.

bhousel:
#4613 is only about making it more obvious that the user needs to enter something.  I'm a bit annoyed that you tried to include in this PR a style that changes colors if the user doesn't type a comment that's "long enough".

bhousel:
I merged #4624 instead.. It's a simpler solution using only CSS.

althio:
Thanks @bhousel for the review and merging the other simpler fix.

I certainly got carried away and added more than strictly mentionned in the initial #4613.

I was trying to re-use the idea of the [changing color _save_ button ](https://github.com/openstreetmap/iD/pull/3105#issuecomment-220067219) as a nudge / soft warning to add a good comment. (Soft in the sense that the Upload button is enabled even with a short comment for instance)
I think the dynamic styling would be a good incentive to add just a few more character to describe more precisely the contributions.

@ iD team
If you agree that the idea has merit, let me know, I can open a new issue to discuss and/or a different PR.
Of course the thresholds can be easily modified if people think that 25 characters is too high or that the warning message should be displayed only for strictly empty messages.

bhousel:
> I certainly got carried away and added more than strictly mentionned in the initial #4613.

It's ok - I do appreciate the PR!

> If you agree that the idea has merit, let me know, I can open a new issue to discuss and/or a different PR.
> Of course the thresholds can be easily modified if people think that 25 characters is too high or that the warning message should be displayed only for strictly empty messages.

I really don't want to spend effort trying to force users to enter good changeset comments, especially not using comment length as a metric for quality.  A 3 character comment might be fine in Chinese.  This is just one of those things where:
- whatever we guess will be wrong for someone
- annoyed users will just enter garbage to pad their comment to whatever length we pick
- it takes time away from developing other features that matter more

PR #4624, opened 2017-12-20
Highlight the changeset editor when the comment is empty

### LINKED ISSUE #4613: focus to changeset comment not obvious (enough)

nnodot:
Change the style of the changeset editor when the comment is empty to explicitly indicate the user what he has to do.

This PR is a minimal change to fix #4613. It only modifies the CSS for the changeset editor.

I've just seen there is another PR for this issue (https://github.com/openstreetmap/iD/pull/4621) which implements some other features that might be helpfull. I still submit mine in case a simpler fix is preferable (also, it's my first one for iD and I chose a "good first issue").

(closes #4613)

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1834,6 +1834,28 @@ input[type=number] {
     text-align: center;
 }
 
+/* Changeset editor while comment text is empty */
+
+.form-field-comment:not(.present) #preset-input-comment {
+    border-color: rgb(230, 100, 100);
+}
+
+.form-field-comment:not(.present) .form-label {
+    border-color: rgb(230, 100, 100);
+    background: rgba(230, 100, 100, 0.2);
+}
+
+.form-field-comment:not(.present) .form-label {
+}
+
+.form-field-comment:not(.present) .form-label-button-wrap {
+    border-color: rgb(230, 100, 100);
+}
+
+.form-field-comment:not(.present) button {
+    border-color: rgb(230, 100, 100);
+}
+
 /* combobox dropdown */
 
 div.combobox {


bhousel:
This is perfect, thanks @nnodot 👍

althio:
Thanks @nnodot for the PR and @bhousel for merging this in.
I like the results with a simple CSS fix.

One question though:
What is the expected behaviour when the _changeset comment_ field is pre-filled by an external application?

nnodot:
@althio I was asking myself the same question. I guess one might want to have the same behavior as the empty comment one's (as in this PR)?

bhousel:
I did not test the behavior with a pre-filled comment, but I'm assuming it would work the same way (that's good!).  If the changeset comment field has some value, the field will be classed as `present` during the update selection. 

https://github.com/openstreetmap/iD/blob/baeff8f59c90b2583950c223c91f4c3074948a4a/modules/ui/field.js#L142-L144

PR #4630, opened 2017-12-26
Fixed click counter persisting

### LINKED ISSUE #4605: Click counter stays around if you interrupt that walkthrough step
If you get to the part of the walkthrough where you are asked to "left click 5 times" and then you skip ahead to a different section, the click counter doesn't go away.

screenshot 2017-12-08 14 13 48

ajlomagno:
Fixes #4605

The code that controls the left click and right click portions of the welcome walkthrough removes the click counter element upon successfully clicking the required 5 times then moving onto the next chapter of the walkthrough. It does not take into account the possibility of the user clicking onto a later portion of the walkthrough while the clicking portions still have not been completed, skipping the deletion of the element altogether.

This fix adds a line to curtain.js that deletes any leftover counter elements when constructing the new tooltip.

id-fix

### DIFF
--- a/modules/ui/curtain.js
+++ b/modules/ui/curtain.js
@@ -221,9 +221,6 @@ export function uiCurtain() {
         }
 
         curtain.cut(box, options.duration);
-        
-        //remove any leftover .counter elements 
-        tooltip.selectAll('.counter').remove();
 
         return tooltip;
     };
--- a/modules/ui/curtain.js
+++ b/modules/ui/curtain.js
@@ -221,6 +221,9 @@ export function uiCurtain() {
         }
 
         curtain.cut(box, options.duration);
+        
+        //remove any leftover .counter elements 
+        tooltip.selectAll('.counter').remove();
 
         return tooltip;
     };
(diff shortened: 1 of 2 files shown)


bhousel:
> This fix adds a line to curtain.js that deletes any leftover counter elements when constructing the new tooltip.

Great!  It looks ok - would it work if the removal code is moved to here instead?:
https://github.com/openstreetmap/iD/blob/8720f7e6d7796046d158b3bc4a45fa3d74a1eaa6/modules/ui/intro/welcome.js#L146-L148

Each chapter has an `exit` method that is called when the user leaves that chapter, so I think it makes more sense for cleanup code to be there.  It's not really the curtain's job to clean up stray DOM elements.

ajlomagno:
That sounds like a much better way to fix this. I originally was looking for a way to contain the cleanup code within that chapter but I seemed to have missed the exit method.

I'll make a new commit very soon and we can take another look at it. Thanks!

ajlomagno:
@bhousel I fixed up this pr with what you mentioned earlier on. I believe it's good to go now.

bhousel:
Thanks!  yay first-time contributor!

PR #4643, opened 2018-01-02
add preset for amenity=love_hotel

willemarcel:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlove_hotel


bhousel:
Thanks @willemarcel!

PR #4644, opened 2018-01-02
add number of nodes information to measurement panel

willemarcel:
As OSM has a limit of 2000 nodes in a way, some users asked me if this information could be available in iD, so I added it to the measurement panel.


bhousel:
I just tried it out and this is great!  Thanks again @willemarcel  👍

PR #4652, opened 2018-01-03
Update da.json

Hjart:
After seeing someone creating craft=sculptor tags for sculptures, I guess these are needed

### DIFF
--- a/dist/locales/da.json
+++ b/dist/locales/da.json
@@ -5011,15 +5011,15 @@
                 },
                 "tourism/artwork": {
                     "name": "Kunstværk",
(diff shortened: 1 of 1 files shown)


bhousel:
Hi, thanks for the report. Our translations are handled through [Transifex](https://www.transifex.com/openstreetmap/id-editor/dashboard/), so you should sign up for an account there to fix issues with the translations.  The contents of the `da.json` file will be replaced by whatever is in Transifex the next time we make a release of iD.

Hjart:
Thank you. I was already notified on IRC about Transifex and have made changes there

PR #4657, opened 2018-01-05
Presets: update Costco, Sam's Club to shop=warehouse

HubMiner:


bhousel:
Thanks for the PR, but this change needs to be made in the [name-suggestion-index](https://github.com/osmlab/name-suggestion-index) project. That is iD's source for the presets for chain stores.

I'd also drop the `building` and `access` tags for these.  `access=customers` is implied, or more for parking lots.  `building` really depends on whether a mapper wants to map it as a building or not (and I'd consider it more `building=retail` than `building=commercial` anyway).

1ec5:
By the way, [`shop=warehouse`](https://taginfo.openstreetmap.org/tags/shop=warehouse) appears to be little used at this point. Most Costco and Sam’s Club locations have been mapped as [`shop=wholesale`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dwholesale) or [`shop=department_store`](https://taginfo.openstreetmap.org/tags/shop=department_store).

HubMiner:
shop=warehouse : good catch, typo in subject.

HubMiner:
I was hoping to keep building=retail, to normalize future values, but I will defer to my seniors...   :)

Another topic, I am proposing to update name:
1) Correct name is "Costco Wholesale",
2) Any objections?

PR #4667, opened 2018-01-08
Restore address format for ["fr", "lu", "mo"]

althio:
addr:unit is not standard in addressing for France
in France : 800 addr:unit / 4 000 000 addr:housenumber = 0,0002 (0,02%)

This partly reverts https://github.com/openstreetmap/iD/pull/4235/commits/de6b4fb6a20b2a3438b4159192e2b54682fab2e5 in https://github.com/openstreetmap/iD/pull/4235

### DIFF
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -53,7 +53,7 @@
         {
             "countryCodes": ["fr", "lu", "mo"],
             "format": [
-                ["unit","housenumber", "street"],
+                ["housenumber", "street"],
                 ["postcode", "city"]
             ]
         },


bhousel:
Thanks @althio 👍

boothym:
@bhousel is there a reason why the unit tagged has been added to all countries' address formats? It seems to have been originally added just for the USA but then expanded everywhere and now some countries are having it removed.

Like France it's not used that often in the UK - only 2,828 uses: http://taginfo.openstreetmap.org.uk/keys/addr%3Aunit - and a number of them near me seem to be wrongly tagged, either as "units" (buildings) in a retail or business park, or other mistaggings.

bhousel:
> @bhousel is there a reason why the unit tagged has been added to all countries' address formats?

It seemed fine.  Don't most countries have some concept of building/floor/suite/etc?
From looking at [common address formats here](https://en.wikipedia.org/wiki/Address_(geography)), then search for "floor" or "apartment", it seems pretty common.

althio:
YMMV, about "floor" or "apartment", they do not seem common to me within a list of *62* countries.
And moreover, they might be needed for real life private adresses, but we map only buildings and entrances adresses in OSM where I live. 4 millions adresses is enough for us, thank you :smile: We won't map every private floor and apartment with its own address soon.

bhousel:
I'm really not interested in discussing address formats, and I just want people to be able to map their addresses however they want to.

If adding an `addr:unit` field to addresses is wrong in the default case, please submit a PR to remove it, and I'll merge it.

althio:
Sorry @bhousel I don't know if the default case would be better with or without `addr:unit`; I can only guess. And it would be difficult to know for sure, or to ask mappers for every country currently included in the default scheme :scream:

bhousel:
> I don't know if the default case would be better with or without addr:unit; I can only guess. And it would be difficult to know for sure, or to ask mappers for every country currently included in the default scheme

I'll just remove it.. I'm really tired of the complaining.

althio:
Do as you see fit but for the record:

My "complaining" was more about the change on existing formats for specific
countries. Sorry if I hurt you or if I took too much of your time on this.
I and all other contributors are simply trying to get the best out of iD
for each local situation.

Leaving aside specific countries, the default is more of a fuzzy case.

PR #4678, opened 2018-01-10
Increase GPX labels readability

### LINKED ISSUE #4617: Hard to read imported trace's points text
The following is a gpx trace with a custom background on iD:
A border would improve the text visibility of the POIs, also increasing the text size might help too.
image

nnodot:
Add a d3 text for the GPX label's halo and update styles.

It might have been better to follow halo handling as with other labels (roads...). I guess this means a full rewrite of the GPX labels handling. But as the issue was tagged "good first issue" I stick to this simple implementation.

(closes #4617)

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -202,6 +202,7 @@ text.pointlabel {
   dominant-baseline: auto;
 }
 
+text.gpxlabel-halo,
 .layer-labels-halo text {
     opacity: 0.7;
     stroke: #fff;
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @nnodot this is great!  I think I'm going to just change the halos to dark and leave everything else as-is..

##### light halos:
screenshot 2018-01-22 23 25 49

##### dark halos:
screenshot 2018-01-22 23 30 35

Issue #2836, opened 2015-11-20
Ask user to repair invalid values


GerdP:
If I got it right, version 1.8.0 will not so easily produce
these invalid tags now, see
https://github.com/openstreetmap/iD/pull/2809

Would it also be possible to ask the mapper to correct existing
invalid tags ?
Example:
If I change a highway with oneway="yes;no" or "surface=asphalt;gravel"
could iD ask me to correct  the wrong tags?
Something like a popup with the text
"If you are local, please try to correct (improve) the attribute xyz before saving"

bhousel:
Thanks for the suggestion, but I'd prefer for users who are interested in OSM data quality and cleanups to use dedicated tools like [Keepright](http://keepright.at/) or [Osmose](http://osmose.openstreetmap.fr/).  

While I don't know of any actual studies done, I suspect that there is a kind of [1% rule](https://en.wikipedia.org/wiki/1%25_rule_(Internet_culture)) in OSM participation.  E.g. for every 1000 casual mappers, we might get 100 who stick around as power mappers (creators), 10 who focus on cleanup (editors), and 1 who actually contributes code (creators).   These numbers are completely made up, but they *feel* about right to me.

iD's goal is to widen the mouth of this participation funnel, so we don't want to turn away the casual mappers by prompting them to cleanup tags (a task that, while important, relatively few OSM participants are interested in).  I would rather get them addicted to mapping, knowing that a handful of of them will convert into users that will take ownership of the map where they live and keep it clean using a dedicated QA tool. 

So here are some things that I would support in iD: 

1. I [mentioned before](https://github.com/openstreetmap/iD/issues/2358#issuecomment-116397055) that I think it is reasonable for iD to have a special "warning" style for UI fields that have a weird value in them. One common way I have seen this implemented in other systems would be if we support a validation regex in the `field.json` schema. For a lot of fields, it might just be a simple `/^;/` - so that if a field contains a weird value we could gently say "here's something that looks funny" rather than interrupting the user to ask them tagging questions that they can't answer.

2. Another idea I've mentioned before would be adding Keepright or Osmose issues as toggleable layers in the Map Data panel.  Both of these tools have APIs from where we could pull a list of QA issues in the current view and display markers for them.  By the way - both of those tools are very good at what they do, but not particularly user-friendly.  They are open source projects that need some love and there are many opportunities for someone who cares a lot about OSM data quality to step up and build a better tool for it.

I'll leave this issue open for discussion...

[...]

PR #4684, opened 2018-01-12
Add ref:isil=* for libraries

ltog:
I added the `ref:isil=*` tag to the library preset.

Its usage is documented here: https://wiki.openstreetmap.org/wiki/Key:ref:isil

### DIFF
--- a/data/presets/fields/ref/isil.json
+++ b/data/presets/fields/ref/isil.json
@@ -1,0 +1,5 @@
+{
+    "key": "ref:isil",
+    "type": "text",
+    "label": "International Standard Identifier for Libraries and Related Organisations"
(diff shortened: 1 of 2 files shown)


JamesKingdom:
I like this but can we change the name to something shorter?

ltog:
@JamesKingdom : I shortened the label.

The ISIL code has a somewhat fixed format that could be described using a regular expression. Is there a way to do this in order to perform a validation and/or prevent invalid input?

bhousel:
Thanks @ltog , this looks great 👍 
There isn't a way currently to validate the input with a regular expression, but that feature is being tracked on #2836

Issue #4565, opened 2017-11-26
Improve the editing of bus routes at bus stop


nlehuby:
Hi,

I would like to ease the action of telling that some bus route stops at a bus stop.
The UI should look like this :
![screenshot-2017-11-26 id](https://user-images.githubusercontent.com/919962/33239507-625568e4-d2a3-11e7-962c-24d4274a13c6.png)

* a semi combo field in the bus stop preset
* the field would be filled by existing bus route relations (using graph.parentRelations function)
* the autocomplete should propose the nearby bus route relations (maybe using the existing ui component ?)
* for the added bus routes, the bus_stop should be added as a member (using the relation addMember function)

What do you think ?

bhousel:
This would be really complicated to do.  I like the idea behind the UI, but almost all of the fields in iD (including the semicombos) are direct mapping between the UI and the tags, not relation memberships, so this would be a very different kind of field from how the semi combo fields works now.  

It would be almost like the turn restrictions field, creating and removing relation memberships.  So if we want to make a special route mapping field, we should think more about what's possible to do with a completely new kind of field, rather than repurposing the semi combo to do route memberships (which it really can't do).

PanierAvide:
@bhousel How much time do you think would be necessary to create this new kind of field ?

bhousel:
> How much time do you think would be necessary to create this new kind of field ?

I don't know, sorry!

slhh:
We need to address the real problems of adding a stop to a bus route here:
- place the stop at the correct position in the relation according to the itinerary.
- allow identification of the correct route relation. Variants (e.g. an express service) might share the same network, linenumber, and destination.

nlehuby:
Ordering the stops within the route is a whole different topic. When you are at the stop, or have some photo of the stop, you may not know how to place the stop within the route, you need to analyse the whole trace to find out.
By the way, iD already allows to add a bus_stop to a route, without handling this.

Yes, you are right, variants may share the same network, number and destination (in fact, they may even share all their tags)
It seems to me that we don't have so many variants with the same tags, and we may deal with it : dealing with 80% of the cases would be better than dealing with none, don't you think ?

PanierAvide:
+1 There are so many bus networks around, and only a few have bus lines with multiple variants/destinations. Think of all small networks having only less than 5 bus lines with very simple routes.

slhh:
@nlehuby 
>Ordering the stops within the route is a whole different topic.
[...]

PR #4694, opened 2018-01-14
Describe relation in tooltip in add membership list

1ec5:
In the dropdown for adding the current selection to a relation, each item now has a tooltip showing the same text as the item’s title, handy for situations where the title gets truncated:

route_master

This tooltip is similar to the one in the preset and tag editors. It addresses one of the pieces of feedback in https://github.com/openstreetmap/iD/issues/4565#issuecomment-357529846.


bhousel:
Good idea, thanks @1ec5  👍

Issue #4682, opened 2018-01-11
Remove building=yes from leisure=sports_centre


lakedistrictOSM:
These often include outdoor facilities such as pitches so adding the building tag is wrong. Wiki says to draw it around the grounds (https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dsports_centre)

bhousel:
Good catch, thanks for flagging

bhousel:
This was done 🎉

PR #4704, opened 2018-01-18
Remove building=yes from leisure=sports_centre

bencostamagna:
Removed building=yes from sport center preset as per this issue:
openstreetmap/iD#4682

### DIFF
--- a/data/presets/presets/leisure/sports_centre.json
+++ b/data/presets/presets/leisure/sports_centre.json
@@ -4,7 +4,6 @@
         "name",
         "sport",
         "address",
-        "building_area",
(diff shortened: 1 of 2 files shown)


data/presets/presets/leisure/sports_centre.json
```diff
@@ -4,7 +4,6 @@
         "name",
         "sport",
         "address",
-        "building_area",
```
bhousel: I think rather than removing it completely, it would be best to leave this as "building" so that users can choose whether they want it to be a building or not (but not have it _default_ to being a building).

PR #4705, opened 2018-01-18
Preset sportcenter

bencostamagna:
still about openstreetmap/iD#4682

I got a bit confused with my branches, so I had to make a new pull request.

I have taken your suggestions into account, the building tag has been added back without any default value.


bhousel:
Thanks @bencostamagna!  Looks great 👍

PR #4708, opened 2018-01-19
Update hindu.json

planemad:
- Removing uncommon hindi/sanskrit synonyms
- Add common English spellings of Hindu temples in various Indian languages.

cc @srividyacb

### DIFF
--- a/data/presets/presets/amenity/place_of_worship/hindu.json
+++ b/data/presets/presets/amenity/place_of_worship/hindu.json
@@ -12,9 +12,11 @@
         "area"
     ],
     "terms": [
-        "garbhargriha",
-        "mandu",
-        "puja",
+        "kovil",
+        "devasthana",
+        "mandir",
+        "kshetram",
+        "alayam",
         "shrine",
         "temple"
     ],


data/presets/presets/amenity/place_of_worship/hindu.json
```diff
@@ -12,9 +12,11 @@
...
     ],
     "terms": [
-        "garbhargriha",
-        "mandu",
-        "puja",
+        "kovil",
+        "devasthana",
+        "mandir",
+        "kshetram",
+        "alayam",
```
1ec5: Given the code below, is it the case that iD no longer weights matching terms by how close they are to the beginning of this list? (If so, that would explain the poor results I always get in Vietnamese, which includes a diacritic-less copy of every term at the end of the list.)

https://github.com/openstreetmap/iD/blob/015c8918f6ef1a6e2d8467a09d8b3b7359856729/modules/presets/collection.js#L110-L115

planemad:
Thank you @bhousel ❤️

data/presets/presets/amenity/place_of_worship/hindu.json
bhousel: @1ec5 The preset collection test suite has a good example of how the search priority is supposed to work: 
https://github.com/openstreetmap/iD/blob/1aebbc5bb38f94e865f80522ff73f2cde94f966e/test/spec/presets/collection.js#L101-L109

bhousel:
> Thank you @bhousel ❤️

Of course!  Thank you for the PR - I usually don't know what search terms people might want to use, so I often just guess.

data/presets/presets/amenity/place_of_worship/hindu.json
1ec5: I guess I was wondering whether it makes a difference for a localization to sort the `terms` from most relevant to least relevant. This test case contains no `terms` array with multiple elements. But I can look into that – thanks for the pointer.

Issue #4638, opened 2017-12-29
Add bus guideway preset


JamesKingdom:
![](https://i.glideservers.com/u/jam70ff04fb-6afd-4b08-9d9c-800033ff5dfa.png)
Bus guideways are rendered as blue lines, but currently have no preset in iD.
![](https://i.glideservers.com/u/jam5d47ffba-503d-44c4-8cd9-be410805e64a.png)

bhousel:
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_guideway

tordans:
For our convenience …
URL for Screenshot 1: http://www.openstreetmap.org/#map=19/51.11520/-0.16590
URL for Screenshot 2: http://www.openstreetmap.org/edit#map=19/51.11520/-0.16637

prattheev:
@JamesKingdom I'll take this!

bhousel:
@prattheev Do you need guidance to get started on this?

PR #4709, opened 2018-01-19
Added a bus guideway preset

bencostamagna:
There has been no activity for almost 3 weeks on issue #4638, so I picked it up.

I edited the CSS file to get the guideways to look as similar as possible to the picture posted in the issue.

For reference, here is how it looks now
bus_guideway

I tried to reuse existing CSS as much as possible and I added oneway to the list of tags suggested in the [wiki](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbus_guideway) to fit the example given.

### DIFF
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -215,6 +215,21 @@ path.casing.tag-residential {
     stroke:#444;
 }
 
+.preset-icon .icon.highway-bus_guideway {
+    color: #fff;
+    fill: #444;
+}
+path.stroke.tag-highway-bus_guideway,
+path.stroke.tag-bus_guideway {
+    stroke:#fff;
+    stroke-linecap: butt;
+    stroke-dasharray: 12, 16;
+}
+path.casing.tag-highway-bus_guideway,
+path.casing.tag-bus_guideway {
+    stroke:#66a3ff;
+}
+
 .preset-icon .icon.highway-unclassified {
     color: #dcd9b9;
     fill: #444;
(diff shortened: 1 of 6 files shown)


bhousel:
Looks great, thank you @bencostamagna .. I will try this out, and will probably change the colors a bit to make it look more like a road and less like a waterway.

bhousel:
I merged this!  Thanks again @bencostamagna ... 
I ended up making it just look like a service road.  I think this is more appropriate for iD, even though mapnik renders it as a dashed blue line.  I also added an icon for it.

screenshot 2018-01-19 10 41 13

bhousel:
Oh @bencostamagna there was one other thing I needed to change - I'm just letting you know in case you want to add more presets (I appreciate your work and hope you do!)

I needed to split the tags up and match _only_ the `highway=bus_guideway` in the `tags:` section, and put _all_ the tags into `addTags:` and `removeTags:` sections:

```js
{
    "icon": "highway-bus_guideway",
    "fields": [
        "name",
        "operator",
        "oneway"
    ],
    "geometry": [
        "line"
    ],
    "tags": {
        "highway": "bus_guideway"
    },
    "addTags": {
        "highway": "bus_guideway",
        "access": "no",
        "bus": "designated"
    },
    "removeTags": {
        "highway": "bus_guideway",
        "access": "no",
        "bus": "designated"
    },
    "terms": [],
    "name": "Bus Guideway"
}
```

This is kind of a weird thing about how tags and OSM work.  iD will recognize the preset based on whatever is in "tags", but it will use "addTags" and "removeTags" when it creates or changes the feature.  

The actual bus guideway example in #4638 was not getting recognized as such because it doesn't have `access=no` and `bus=designated`.  These are good tags to add, but iD needs to recognize the feature even if they are not present.  Hope this makes sense (please let me know if it doesn't!).

bencostamagna:
It does make sense, thanks for the explanation.

Issue #4702, opened 2018-01-18
Add barrier=kerb preset


higa4:
Important tag for mobility handicapped persons.
https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dkerb

tordans:
Could it be, that this change also triggered the new rendering of `kerb` on the edit-map?

It now looks like this:
[Example](https://www.openstreetmap.org/edit#map=19/52.47697/13.45030)
<img width="832" alt="bildschirmfoto 2018-08-10 um 18 26 00" src="https://user-images.githubusercontent.com/111561/43969656-2891b07a-9ccb-11e8-90b4-d3aa7ce93e07.png">

Inside the park there are real barrier=fence tags.
But along the street thats all barrier=kerb.

Which used to look like soft gray line but now is this really bold thing demanding a lot of attention.
It might just be something I need to get used to. But just in case: Is there a way to change the rendering to a softer style? Eg. a different line-width?

bhousel:
> But just in case: Is there a way to change the rendering to a softer style? Eg. a different line-width?

Yes, right now all the barriers just look like that.. 

We could eventually have different styles of rendering for different barriers.  It is great that people want to map this much detail, and tools like iD will need to deal with more data density.  

I've also thought about having certain features only render at much higher zoom (z20ish?).. Or maybe assigning a minZoom to each preset.  🤔

PR #4715, opened 2018-01-20
added support for kerb barrier preset

jay-manday:
Hi there,

I saw issue #4702 and this looked like a good place to get acquainted with the codebase and start contributing. I went ahead and wrote  kerb.json file as well, I was a little unsure of which icon to use so I went with 'wheelchair' from the [maki icon set](https://www.mapbox.com/maki-icons/). 

It might be a good idea to create icons to differentiate between wheelchair friendly and unfriendly barriers. Happy to draft up a proper issue/proposal for that as well n_n

Hope this helps!
-Jason

### DIFF
--- a/data/presets/presets/barrier/kerb.json
+++ b/data/presets/presets/barrier/kerb.json
@@ -1,0 +1,14 @@
+{
+    "icon": "wheelchair",
+    "fields": [
+        "access"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "barrier": "kerb"
+    },
+    "name": "Kerb"
+}
--- a/data/presets/presets/barrier/kerb.json
+++ b/data/presets/presets/barrier/kerb.json
@@ -1,10 +1,10 @@
 {
     "icon": "wheelchair",
     "fields": [
-        "access"
+        "tactile_paving",
+        "kerb"
     ],
     "geometry": [
-        "point",
         "vertex"
     ],
     "tags": {
--- a/data/presets/categories/barrier.json
+++ b/data/presets/categories/barrier.json
@@ -8,6 +8,7 @@
         "barrier/ditch",
         "barrier/gate",
         "barrier/hedge",
+        "barrier/kerb",
         "barrier"
     ]
 }


data/presets/presets/barrier/kerb.json
```diff
@@ -0,0 +1,14 @@
+{
+    "icon": "wheelchair",
+    "fields": [
+        "access"
+    ],
+    "geometry": [
+        "point",
```
bhousel: For this preset, change "point" to "vertex".. (Points are standalone nodes, and vertices are nodes that exist along a parent way)

data/presets/presets/barrier/kerb.json
```diff
@@ -0,0 +1,14 @@
+{
+    "icon": "wheelchair",
+    "fields": [
+        "access"
```
bhousel: I don't think "access" field makes sense here. 
You could add "kerb" and "tactile_paving" in its place, since we already have fields set up for those tags..

jay-manday: sure thing!

data/presets/presets/barrier/kerb.json
jay-manday: sounds good i'll make these tweaks 👍

bhousel:
Thanks, looks good!

jay-manday:
Thanks @bhousel!

Happy to help out, I'll keep an eye out for other issues to take on. Are there any projects or areas of interest you guys are particularly looking for help with?

bhousel:
Great @jay-manday - You can always dig into anything tagged with "bug" or "help wanted"..

Not sure whether you are in college, but if so you might be eligible for Google Summer of Code.  Later today I will add some more long term project ideas to GSoC ideas page here: https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2018/Project_Ideas

jay-manday:
Sounds good, I'll keep a look out!

Definitely, I see some really interesting projects there. Yup! I'm a senior studying comp sci and studio art. I'll keep my eye on projects there :)

Issue #2864, opened 2015-12-08
Persist wireframe area fill and feature filtering preferences in localstorage


bagage:
We can change these two options during a session but when the tab is closed, their value are lost. Same applies for last map layer used while the "custom map layer" URL is saved for future uses.

Is there a reason behind that or would it be possible to save user options so that they stay persistent through multiple tabs?

bhousel:
> Is there a reason behind that or would it be possible to save user options so that they stay persistent through multiple tabs?

No reason really.  Traditionally iD tried to avoid the idea of having user "preferences", but most feedback I hear from users lately is that they prefer to retain their preferences between editing sessions, so we do more of that now.

bagage:
Hey @bhousel, is there anything new around some user storage preferences since this issue? I'm using Osmose a lot which opens a new browser tap for each issue so I have to reconfigure my context everytime. I can see some similar issues like #2947 and #1756.
The alternative solution would be to manually modify lat/lon coordinates in URL bar and refresh tab but unfortunately iD does not support that. Should this be entered as another issue?

bhousel:
Hey @bagage, I was putting off working on the storage-of-preferences, hoping to get #4418 in first.  Since we aren't doing that, I think we could just do the simple thing of putting more preferences into localStorage.

bagage:
Thanks for the insight @bhousel :). I'll take a look at the local storage code and try to submit PR for this issue. Any documentation around and/or tips before getting started is good to hear if there is any?

bhousel:
- [ ] persist wireframe area fill
- [x] persist preferred feature filters

Thanks for the PR @bagage  - I'm just marking here that you did the feature filters part already..

PR #4720, opened 2018-01-21
Restore last imageries used and disable features from local storage

bagage:
This is part of #2864 but I'll split it in multiple PRs. 
I'd like to save/restore the previous background used too, because you can then use "ctrl+b" shortcut to quickly switch between imageries even in a new tab (In France most of us use imagery from BDOrtho IGN instead of Bing but also Cadastre is another great resource to have and we often switch between them).

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -386,6 +386,7 @@ export function rendererBackground(context) {
             background.baseLayerSource(
                 background.findSource(requested) ||
                 best ||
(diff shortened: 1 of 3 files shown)


bagage:
I pushed disabled features restoration on the same PR after all; there are 2 separate commits but both related to local storage so I guess it makes sense.

bhousel:
Thanks @bagage - I tried this out locally and it's perfect! 👍

PR #4724, opened 2018-01-22
Added a background for labels of gpx waypoints

bencostamagna:
I didn't notice the work in progress tag and the existing pull request on issue #4617 so I started working on it.

I have used a slightly different approach than the other contributor by adding a background to the text instead of a halo.

I am just putting it here in case you find it useful

Here is how it looks
gpxtags

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -260,5 +260,13 @@ path.gpx {
 }
 
 text.gpx {
+    fill: #444;
+}
+
+rect.gpxtag {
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @bencostamagna.. Your PR is good too, but #4678 was submitted first, so I accepted that one.. Sorry!

bencostamagna:
No worries, my fault for not checking

PR #4736, opened 2018-01-23
Added a preset for emergency/water_tank

bencostamagna:
I have been mapping some [water tanks](https://taginfo.openstreetmap.org/tags/emergency=water_tank) used by firefighters and I noticed iD doesn't have a preset for it, so here is one.

### DIFF
--- a/data/presets/presets/emergency/water_tank.json
+++ b/data/presets/presets/emergency/water_tank.json
@@ -1,5 +1,5 @@
 {
-    "icon": "poi-fire-hydrant",
+    "icon": "water",
     "fields": [
         "name",
         "operator",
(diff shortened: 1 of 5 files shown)


data/presets/presets/emergency/water_tank.json
```diff
@@ -0,0 +1,20 @@
...
+        "vertex"
+    ],
+    "terms": [
+        "cistern",
+        "reservoir"
+    ],
+    "tags": {
+        "emergency": "water_tank"
+    },
+    "name": "Water Tank"
```
bhousel: Can you change this to "Emergency Water Tank" ? 

I wouldn't want people to confuse it with just a normal water tank like: https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dstorage_tank

bencostamagna: Good point, I have changed the name accordingly and downgraded "water tank" to a search term.

Issue #4701, opened 2018-01-18
Add amenity=smoking_area preset


higa4:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dsmoking_area
Useful tag for smokers searching for outdoor smoking area located at road side, square, pedestrian deck.

PR #4737, opened 2018-01-23
Added smoking area preset

bencostamagna:
Issue #4701.

I have added a smoking area preset.
I haven't found any fitting icon in the maki set, so I have left the icon field blank. Let me know if using a placeholder such as circle or marker is preferred.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -2183,6 +2183,10 @@ en:
         name: Shower
         # 'terms: rain closet'
         terms: '<translate with synonyms or related terms for ''Shower'', separated by commas>'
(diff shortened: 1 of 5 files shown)


bhousel:
Looks good, thanks @bencostamagna ! 👍
It's ok for the preset to have no icon - iD will just use a generic one.

One other tip:  This is optional but you can write something like "(closes #4701)" in your commit message and GitHub will automatically close the issue when I merge it.  I usually just put that on a line by itself after the main commit message.. [more info](https://help.github.com/articles/closing-issues-using-keywords/)

PR #4738, opened 2018-01-23
Restore area fill from local storage including wireframe (Closes #2864)

### LINKED ISSUE #2864: Persist wireframe area fill and feature filtering preferences in localstorage

bagage:
Actually area fill was already partly saved in local storage, except for `wireframe` -- I changed logic a bit so that this preference can be saved too.


bhousel:
Thanks @bagage!  Looks good..

PR #4751, opened 2018-01-29
Warning on multiple deletions: added some details.

### LINKED ISSUE #4666: Totals on save warning can be confusing
Reported in IRC today, a user was confused when they saw this screen.  They didn't understand why it said they deleted 214 features, but only made 69 changes.  I think we should add some detail to list of what actually got deleted, so that the numbers make more sense.  

The user was also worried that they did a bad thing and almost quit completely.  Maybe we can reword this warning to make it a bit less scary. 

citbrjg

bencostamagna:
closes #4666

Modified the warning message when deleting many entities according to the suggestion in the discussion on the issue.

image


bhousel:
Looks good, thanks @bencostamagna !

PR #4756, opened 2018-01-30
Parking preset: Removed address from default fields

### LINKED ISSUE #4748: remove address-field from amenity=parking preset
I want to ask what do you think about removing the addr(ess)-field from the amenity=parking preset. The issue is that the addr-tagging-scheme is being provided explicitly for objects that do have a real address, so e. g. They have also a post box. Most parking areas do not have that. The addr-tagging-scheme is not to add a geographical reference like street reference, for that you have to use other schemes like ref:street or similar. The problem is, that for most parking areas people might be seduced to add a fictive address because of the preset. Of course, multi-storeys do have addresses sometimes, but surface parking areas do nearly never have and I'm very sure there are many more parking spaces without a real address than the opposite situation. And for parking buildings maybe it might be also worth to think about adding an own preset for parking=multi_storey, if you want to keep the address-field there(?)

bencostamagna:
closes #4748

I simply removed address from the list of default fields.

The original issue also mentions a preset for multilevel parking that would include the address field and the tag parking=multi-storey. I haven't pushed that change yet because it isn't 100% clear from your comment whether you also want that added. Tell me if I should.

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -2040,8 +2040,7 @@
                 "supervised",
                 "park_ride",
                 "surface",
-                "maxstay",
-                "address"
+                "maxstay"
             ],
             "geometry": [
                 "point",
--- a/data/presets/presets/amenity/parking.json
+++ b/data/presets/presets/amenity/parking.json
@@ -9,8 +9,7 @@
         "supervised",
         "park_ride",
         "surface",
-        "maxstay",
-        "address"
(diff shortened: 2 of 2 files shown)


bhousel:
Thanks @bencostamagna this is great!
I don't think we need a preset for `parking=multi-storey` because the existing Parking preset has that option in the "type" field.

Issue #839, opened 2013-02-21
Copy-paste tags


yvecai:
A control should allow copying tag from one feature to another. Controls in the tag editor could do the trick.

pauliuszaleckas:
+1 me too. I miss 'R' from Potlatch very much.

brycenesbitt:
"Turn into area" seems more iD-like, and meets mostly the same use case.

nighto:
@brycenesbitt no, it is a similar but not the same use case.

Let's say I want to add dozens of the same feature - bus stops in a street, for instance. They all have benches (so bench=yes) and are covered (so covered=yes). On Potlatch, JOSM, Vespucci etc. I could create the first node with all features I want, then create the following nodes and use the function to copy attributes (on Potlatch, press ```R```). On iD, I have to manually copy all attributes to the different nodes.

HolgerJeromin:
vespucci has a nice GUI for selective copy and paste of tags in its newest version. I already miss this in JOSM :-)
A checkmark before each tag and one on top for all.
http://imgsdown.1mobile.com/group3/M00/20/93/S34UR1Xp-aGAQbauAAGDUSXQ64Q549.png
having some tags selected i could copy or cut the tags with buttons and can paste them with a button into another object (when there is no checkmark).

nvkelso:
+1, especially helpful to correct tags.

slhh:
Copy tags is a quite useful operation, but there are some risks associated with it:
* The souce object might have additional tags, which should not be copied.
* The selection might contain objects of wrong type (like vertices) accidentially. This is especially an issue in case of area based selections like lasso selection.
* Conflicting values might be accidentially overwritten.

I have proposed #3028 as a substitue in order to reduce these risks by giving the user more contol making the user more aware of the performed operations.

jarlarntzen:
@slhh #3028 sounds good but I would also argue that a good start would be to replicate the existing "R"-shortcut key from Potlatch which simply repeats the tags of the previously selected object of the same type.

Point tags are repeated on points.
Line tags are repeated on lines.
Area tags are repeated on areas.

Conflicting tags are replaced, all others are kept. Nice for setting building=house where there is already a street address on each building.

Fast and nimble in use. Simple(r) to program.

slhh:
@jarlarntzen 
> "R"-shortcut key from Potlatch which simply repeats the tags of the previously selected object of the same type.

[...]

PR #4762, opened 2018-02-01
Copy/Paste tags between objects with Ctrl-C/Ctrl-V

magiraud:
Dear community, 

Thank you to all what you did with iD. Here is a proposal that solves parts of #839, being able to copy tags between objects:

1) It implements a simple `actionMergeTags(idsFrom, idsTo)` that iterates over the existing `entity.mergeTags()` (tags are actually *merged*) and then a `behaviorPasteTags `that pastes/merges all tags from previously copied entities in `context.copyIDs()` onto every selected entity. Thus: 
  - When several objects were copied, all their tags are copied - I am not sure that this is so useful, but we have to deal with these multiple copied entities.
  - When several objects are selected, the tags are pasted/merged onto all selected objects - making easy to copy a same set of tags to a bunch of objects. This partially answer to #1761. This may be dangerous for area-based selections, see https://github.com/openstreetmap/iD/issues/839#issuecomment-195622518, but we could assume that people doing such selections know what they are doing.

2) What should be the keybinding to trigger this behavior ? One solution much liked in #839 could be to have something like `Ctrl+shift+V`. Here we explore another possibility. In my favorite text editor, when something is selected, `Ctrl+V` replaces this selected thing with the copied content. This is not the case in iD, when one does not really understand what was the effect of having selected something when one later presses `Ctrl+V`. As suggested by @yvecai in https://github.com/openstreetmap/iD/issues/839#issuecomment-336423621, the proposal here is thus to use the same `Ctrl+V` to paste/merge tags onto the *existing* selected objects when we are in `modeSelect`. (Nothing is thus changed in `modeBrowse`, `Ctrl+V` still copies objects there.) Anyway, we have to think whether this is an acceptable idea, or whether `Ctrl+shift+V` as previously proposed would be better.

What this proposal does not solve is to partially copy tags. This would requires UI work such as drafted by https://github.com/openstreetmap/iD/issues/839#issuecomment-274321929. Anyway, with this proposal one can still duplicate an object, remove the unwanted tags, and copy/paste the remaining tags to the desired objects.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1042,8 +1042,8 @@ en:
         delete: "Delete selected features"
       commands:
         title: "Commands"
-        copy: "Copy selected features"
-        paste: "Paste copied features"
+        copy: "Copy selected features and tags"
+        paste: "Paste copied features or tags"
         undo: "Undo last action"
         redo: "Redo last action"
         save: "Save changes"
--- a/modules/actions/index.js
+++ b/modules/actions/index.js
@@ -20,6 +20,7 @@ export { actionJoin } from './join';
 export { actionMerge } from './merge';
 export { actionMergePolygon } from './merge_polygon';
 export { actionMergeRemoteChanges } from './merge_remote_changes';
+export { actionMergeTags } from './merge_tags';
 export { actionMove } from './move';
 export { actionMoveNode } from './move_node';
 export { actionNoop } from './noop';
--- a/modules/actions/merge_tags.js
+++ b/modules/actions/merge_tags.js
@@ -1,0 +1,23 @@
+export function actionMergeTags(idsFrom, idsTo) {
+
+    // Merge all tags from entities whose id is in idsFrom
+    // into every entity whose id is in idsTo
+
+    return function(graphFrom) {
+
+        var graphTo = graphFrom;
+
+        idsTo.forEach(function(idTo) {
+            var entityTo = graphTo.entity(idTo);
+
+            idsFrom.forEach(function(id) {
+                var entityFrom = graphTo.entity(id);
+                entityTo = entityTo.mergeTags(entityFrom.tags);
+            });
+
+            graphTo = graphTo.replace(entityTo);
+        });
+
+        return graphTo;
+    };
+}
--- a/modules/behavior/index.js
+++ b/modules/behavior/index.js
@@ -10,5 +10,6 @@ export { behaviorHover } from './hover';
 export { behaviorLasso } from './lasso';
 export { behaviorOperation } from './operation';
 export { behaviorPaste } from './paste';
+export { behaviorPasteTags } from './paste_tags';
 export { behaviorSelect } from './select';
 export { behaviorTail } from './tail';
--- a/modules/behavior/paste_tags.js
+++ b/modules/behavior/paste_tags.js
@@ -1,0 +1,39 @@
+import {
+    event as d3_event,
+    select as d3_select
+} from 'd3-selection';
+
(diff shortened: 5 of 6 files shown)


yvecai:
Shift-ctrl-V would have the benefit to be the same as JOSM, but your implementation also makes sense. 
UI for selecting the pasted tags would add additional steps, so it's counterproductive as the original request intent was to provide a fast way to map similar objects.

slhh:
A shortcut based copy/paste would be quite unsafe in iD. One reason are less experienced users, but the more important reason is iD's architecure. Tags are not shown in iD by default, and the "all fields" section does not cover all tags. The user might not be aware of some of the copied tags.
Therefore, iD needs to make the user aware, which tags are copied, either when copying or when pasting.

bhousel:
> A shortcut based copy/paste would be quite unsafe in iD. One reason are less experienced users, but the more important reason is iD's architecure. Tags are not shown in iD by default, and the "all fields" section does not cover all tags. The user might not be aware of some of the copied tags.
> Therefore, iD needs to make the user aware, which tags are copied, either when copying or when pasting.

Thanks @slhh, this is my thinking too.. This is why in #839 I was pushing for a solution that requires the user to specifically mark which tags they want copied.

magiraud:
Dear all,

On one side, I understand the points raisen by @slhh and @bhousel. We do not want that users not knowing internals of OSM doing harm to the database. In some case, we could make them more aware of what they are doing, but anyway we should be very careful. (As a side note, `All fields` can be misleading, `Fields` or even `Preset fields` could decribe more precisely what is there.) So yes, I agree that `Ctrl + V` could perhaps be too unsafe, some users would paste tags without knowing what they are doing. 

On the other side, as reminded by @yvecai :
> UI for selecting the pasted tags would add additional steps, so it's counterproductive as the original request intent was to provide a fast way to map similar objects.

And it is not only a matter of efficiency, but also of *quality* : such a feature helps to have a coherent state when using iD to create or to update some entities that should have the same tags. Manually assigning tags to several objects is tedious and error-prone. It seems that several people commenting at #839 (at least @nighto @sfkeller @ghost07tula @Dulahey) were also looking for something simple and efficient.

Could it be possible to find something that could be acceptable for everyone ? What if we go back to the original idea of using `Ctrl + Shift  + V` to trigger this "paste tags" behavior ? As this keystroke is not so common (and it furthers requires to be in `modeSelect`), it seems quite unlikely that beginner users will use that, and it will still allow advanced users to use it and to be very efficient with that.

sfkeller:
@magiraud: +1. I can live with Ctrl-Shift-V and I'd like to support any progress on the Copy-paste tags issue like with this patch. 

I just want to make all aware of the specification https://github.com/openstreetmap/iD/issues/839#issuecomment-361880543 I proposed which seems to me not much more to implement but with a slightly more holistic approach:
* GUI: Adding icons for "Cut | Copy | Paste" to the "icon bar"
* GUI/Copy: Showing check boxes of source tags for selecting
* Copy: Adding a parser to allow copying from multiline text from outside iD (like .g. from OSM notes)
* Paste: Adding values to same tag key, and ignoring duplicate tags.

bhousel:
> Currently this pull request does not implement what Bryan pushed for, namely requiring the user to specifically mark which tags they want copied.
> @bhousel Anything more missing for you to accept this pull request?

You said it yourself.. It doesn't implement what I asked for.  It also overloads Paste to do a second function which is not at all intuitive or like pasting.

I'm sorry, but I probably should have closed this back in February.  If you're looking for a really efficient way of pasting tags onto features, you can use JOSM or level0 until someone can afford the time to make this feature beginner-friendly in iD.

PR #4633, opened 2017-12-27
Create features from geoservices

jgravois:
demoed at [SOTMUS US 2017](https://2017.stateofthemap.us/program/adding-arcgis-data-to-id). resolves #4164.

Fetch service metadata and confirm license is valid:
add_new_service

Re-add a service with a license that was previously confirmed:
add_existing_service

Map service attributes to OSM tags and import the features in view:
map_attributes2

Review and approve individual features:
review_and_approve_features5

we can continue to use the fork [issue tracker](https://github.com/mapmeld/iD/issues/
) if the discussion here gets difficult to follow.

cc/ @slibby @mapmeld (primary dev)

PR #4767, opened 2018-02-02
query the appropriate metadata for esri clarity

### LINKED ISSUE #4766: Esri World Imagery and World Imagery Clarity are showing same metadata
We are supporting the both Esri World Imagery and Esri World Imagery (Clarity), but the "clarity" source is still querying the same metadata server, so we don't actually see the true values for imagery vintage, etc.
cc @jgravois 

Code to fix is here:
https://github.com/openstreetmap/iD/blob/db0858f7d2f3ceca98a5bfa42422f422791f182f/modules/renderer/background_source.js#L292-L293

esri_metadata

jgravois:
resolves #4766 

reference: https://www.arcgis.com/home/item.html?id=298f03f7f02c465a87bcf817c3b8859b

### DIFF
--- a/modules/renderer/background_source.js
+++ b/modules/renderer/background_source.js
@@ -276,6 +276,9 @@ rendererBackgroundSource.Esri = function(data) {
         if (inflight[tileId]) return;
 
         switch (true) {
+            case (zoom >= 20 && esri.id === 'EsriWorldImageryClarity'):
+                metadataLayer = 4;
+                break;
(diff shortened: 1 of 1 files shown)


bhousel:
Wow thanks for the quick fix @jgravois 👍

jgravois:
no worries!

i'm just trying to make sure you have as much time as possible to review that mammoth #4633 😉

Issue #2622, opened 2015-05-01
Support for only_straight_on turning restrictions


socketpair:
Today, we can easily add turn restrictions named no_\* but only_\* can be added only by hand as creating restriction and then fixing its type.

It will be nice if these restriction tyoes can be switched like second clicking on that. Note that shift click is not acceptable for ipad.

bhousel:
There's some discussion on #224, @jfirebaugh knows more what's involved here.

afmenez:
That issue is closed already.  i liked the suggestion to cycle red/green/blue on each click.  It is not mentioned there, but maybe there should be an "optimization" phase were redundant restrictions are eliminated (if I have an only_forward I don't need a _no_left and no_right restrictions).

jfirebaugh:
No plans to add a UI for this; I'm happy with the current balance between complexity and functionality.

bhousel:
Reopening this.. We'll probably add it, but will need to be careful to preserve the previous state of the intersection so we can cycle through `only_` states without destroying all the `no_` restrictions.  A lot is moving on turn restriction tagging to support #2346, which will land soon.

bhousel:
This was done in #2346!

Issue #2346, opened 2014-09-02
Support turn restrictions via way


bhousel:
Eventually.. Add support for `no_u_turn` via a way (or multiple ways?).
e.g. `no_u_turn` via the short section joining the parallel ways of a dual carriageway.

bhousel:
Update on this, I've built a version of this and it is working pretty well.. Will merge soon.

bhousel:
This was done in #4768!

Issue #3136, opened 2016-05-26
Turn Restrictions field is too big and obtrusive


martenjj:
It's very good for editing turn restrictions.  But, for most of the time when it appears, it is not relevant and takes up a lot of vertical space in the left panel.  In particular, it pushes the "All tags" and "All relations" areas downwards which means that scrolling is needed to see them.  If these two "All" lists were more visible then it would help with being able to check them before a merge or delete - the source of many problems.

I can think of two ways to achieve this:

1) Make the "Turn Restrictions" field expandable, like the "> All tags (N)" as it is currently displayed, and collapse it by default - but show the number of restrictions in the brackets to indicate that some exist.

2) Make it an optional field which can be added from the "Add field" combo, in the same way as "Description" and the others.

bhousel:
This is lower priority, but I kind of agree with the idea of displaying some summary information ("<count> turn restrictions") and letting the user expand for more detailed tagging... 

My other issue with this field is that on hover it doesn't show any useful information.  You need to select the vertex and click on ways to see what if any turn restrictions have been added.

matkoniecz:
> Make it an optional field which can be added from the "Add field" combo, in the same way as "Description" and the others.

that would severely limit its discoverability

Issue #597, opened 2013-02-01
Better i18n plural support


jfirebaugh:
Multiselect produces translations like "Deleted {n} objects", which we're currently translating without real plural form support.

althio:
I would like to see this re-assessed, and maybe flagged "help wanted" from someone fluent in internationalization stuff.

bhousel:
Sure @althio we can reopen and ask for help.  I agree that it's not critical, but if somebody wants to work on it, I'd accept a PR for it.

In some places we support pluralization just by having multiple translation strings:
https://github.com/openstreetmap/iD/blob/c8e89b7a0542a6ebf1c07f7d1cccbb01e7a52152/data/core.yaml#L80-L103

althio:
Thanks @bhousel 

> In some places we support pluralization just by having multiple translation strings:

There are shortcomings with this simple approach and maybe a systematic approach would be better (but more complex).
From what I read, plural rules are different depending on languages and lots of work and thoughts already on the subject.
https://stackoverflow.com/questions/14326653/java-internationalization-i18n-with-proper-plurals#14327683
http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html

bhousel:
> There are shortcomings with this simple approach and maybe a systematic approach would be better (but more complex).

Thats fine, but I'd strongly prefer "simple" over "perfect"..  If we need to redo all of our translations and the code to produce them to get to "perfect", lets not do that.

1ec5:
Some of the supported languages pluralize according to more complex rules that couldn’t be accommodated by adding a few extra strings for translation. For example, some languages have a different plural form for numbers ending in 1. I think it would be a reasonable goal to support each of iD’s languages as a first-class citizen, but doing so by adding strings would unnecessarily increase the workload for all translators. For example, a Spanish translator shouldn’t have to translate redundant strings because Russian has a dual form.

The good news is that Transifex supports alternatives to the YAML format that explicitly handle pluralization. (I’m familiar with .stringsdict from iOS development, but there are others.) We could supplement the existing YAML file with another file containing a few override strings for which pluralization matters. For languages like Chinese that have no plural forms, there would only be one set of strings to translate. Back here in this repository, iD could depend on one of several libraries that support the CLDR plural format.

1ec5:
> The good news is that Transifex supports alternatives to the YAML format that explicitly handle pluralization.

[...]

PR #4768, opened 2018-02-03
Add support for complex intersection and via way restrictions

bhousel:
This PR adds support for creating turn restrictions in more complex intersections.  
- [x] #2346
- [x] #2622 

### Before
- You click on a junction node
- The turn restriction editor shows you all the ways connected to that junction node
- You select a _from_ way, and can add restrictions _to_ another way in that intersection
- Turn restrictions _via node_ only

### After
- You click on a junction node
- The turn restriction editor shows you all ways and important junction nodes within 20 meters
- You select a from way, and can add restrictions _to_ any other way
- Turn restrictions _via node_ and _via way_ are both supported automatically
- (If there are no other nearby junctions found, the turn restriction editor works just as before)

This is work that I did a few months ago, and has been used internally at Mapbox by our data team.  I'm confident that it is stable enough to merge into the main iD that everyone uses.  🎉 

Would love to hear people's feedback on ways to make this more useful or easier to understand for casual mappers.

advanced_intersection

(The text that flashes while I'm using the editor is for debugging - I'll make this friendlier!)

↩️ ⬆️ ⬇️ 🔀


manfredbrandl:
I would like to try this to give you feedback. Is it possible to try this at http://preview.ideditor.com/newturnrestrictions/ ?

bhousel:
@manfredbrandl - I put a copy of this branch here for testing:
http://sonny.7thposition.com/osm/iD/

manfredbrandl:
@bhousel thanks, i tried and had no problems with useability at all.

manfredbrandl:
I tried somewhere else and found a bug. i cannot select the "Maria-Theresei-Allee" (grey street connecting from south-west) in the turn restrictions editor:
image
When i move the mouse over it "w4014543" is displayed in the bug area and the street is surrounded by the pink border (not flashing), click on the street does not change anything.
See http://sonny.7thposition.com/osm/iD/#background=basemap.at-orthofoto&disable_features=boundaries&map=20.65/47.07766/15.44361

bhousel:
> When i move the mouse over it "w4014543" is displayed in the bug area and the street is surrounded by the pink border (not flashing), click on the street does not change anything.

Thanks @manfredbrandl that does look like a bug!

bhousel:
By the way @manfredbrandl thanks for posting this intersection..
It's a good example of one where it doesn't fit very well into the turn restriction editor view.  This is partly because it fills the 20 meter limit of what we will show, and has very short segments, like the service and unclassified roads.

I made some tweaks to extend the leaf nodes out far beyond the viewport edge, and to better calculate a zoom and center that includes all of the important parts of the intersection..

screenshot 2018-02-07 08 29 22

manfredbrandl:
@bhousel it was a pleasure for me to help you. Could you provide the new code on the branch so that i can test it again?

bhousel:
> Could you provide the new code on the branch so that i can test it again?

Just updated  http://sonny.7thposition.com/osm/iD/  to ea4ac80fe

slhh:
The following screenshot seems to show a bug:
When I click on the upper red icon it toggles both red icons to green, and with the next click both back to green.
When I click on the lower red icon it toggles both red icons to green, but the next click toggles only the lower one to red. Both icons can be toggled separately, but as soon as the upper icon is clicked while both are green, they get coupled again.

I would expect both icons to toggle always seperately.

grafik
grafik
Selected node was n26415364, but the yellow line (secondary) was temporarily added for testing.

slhh:
> Would love to hear people's feedback on ways to make this more useful or easier to understand for casual mappers.

Here are some ideas:

- Place a start arrow on the starting way pointing directly to the connecting node.

- When hovering an icon, highlight all ways belonging to the restriction  and/or animate a driving vehicle. 

- Do only show the first level of restriction icons. Split the icon into the circular part and the arrow part. Use the circular part to toggle the restriction, and use the arrow part to unhide the next level. In the next connecting node is out of view, zoom the restriction editor accordingly. Make the icons (maybe arrow part only) yellow, if some existing restrictions on further levels do exist, which are hidden.

- The restriction editor doesn't really fit into the inspector of a connecting vertex anymore. Showing properties of a selected node above and below, while editing restrictions which aren't related to the node, can be confusing.

- Show the full list of generated relations below the restrictions editor.

manfredbrandl:
> use the arrow part to unhide the next level
I would prefer to see all turn restriction with the selected from way without having to unhide several via ways in the next level. At https://wiki.openstreetmap.org/wiki/Relation:restriction there is an example with three via ways:
image

slhh:
I have found another bug. A very short (e.g. 5 m) deadend way is stretched to the frame in the restriction editor. This looks quite strange.

> The turn restriction editor shows you all ways and important junction nodes within 20 meters

Especially for large junctions, which are the typical ones requiring complex turn restrictions, 20m seem to be insufficient. Here are some ideas to address this:

- Analyse existing restrictions to select the default scaling
- Click on the arrow part of the icons to extend as I mentiond in my previous comment.
- Use small arrows where the way intersects the frame to extend on click to the next connection node. The arrows can be hidden if there isn't any connection node within resonable distance.

> You select a from way, and can add restrictions to any other way
> Turn restrictions via node and via way are both supported automatically

Via nodes and ways aren't redundant, and they can't be added automatically in the general case.
They have to be members for good reason, a fixed from-to pair can have different paths due to using different via members.
Unhiding selectively level by level, as mentioned in my previous comment, would give the user control which path to use.

manfredbrandl:
> My first thought was to show all existing restrictions by default, but I realized that there could be conflicts due to multiple possible via paths.

Thats a point. Maybe we can have a conflict-checker and initially support editing junctions without conflicts only. iD is easy to use and intentionally simple. It lets you do the most basic tasks while not breaking other people's data. We can support junctions with conflicts later.

bhousel:
(Trying to reply quickly to most of the feedback - sorry if I miss some)

from @slhh: 

> Place a start arrow on the starting way pointing directly to the connecting node.

will consider, but it seems like clutter.

> When hovering an icon, highlight all ways belonging to the restriction and/or animate a driving vehicle.

👍 I've added this now, it highlights all ways along the path in the "related" way style.  I might adjust it, since this is not very visible

> Do only show the first level of restriction icons. 

I've added a slider that will let users expand the search (see below)

> The restriction editor doesn't really fit into the inspector of a connecting vertex anymore. Showing properties of a selected node above and below, while editing restrictions which aren't related to the node, can be confusing.

Agree, but this is #3136 and I don't have time to do it now.

> Show the full list of generated relations below the restrictions editor.

Probably wont at this time - for some very complex intersections (like an SPUI) it can be several dozen.  But if we do #3136, this would be replaced by a summary view that might list them.

> I have found another bug. A very short (e.g. 5 m) deadend way is stretched to the frame in the restriction editor. This looks quite strange.

Yeah this is intentional - I do need to show the user something here and it needs to be long enough to be clickable even with an arrow stuck on top of it, but I think I don't need to extend it quite so far.  I'll try a shorter amount.  (can you link to the intersection?)

> Especially for large junctions, which are the typical ones requiring complex turn restrictions, 20m seem to be insufficient. Here are some ideas to address this:

Yes, I noticed this too (testing with the intersection at `map=20.00/39.89412/-75.09134`) -  I've extended it to 30m for now.  Might use the detail slider mentioned below to adjust the search distance.

> Analyse existing restrictions to select the default scaling

Yep, it does this now (the viewer adjusts the zoom to better fit all the key vertices).

> Click on the arrow part of the icons to extend as I mentiond in my previous comment.

Part of the main design goals for this editor is to continue to allow mappers to create restrictions with one click.  Our data team at Mapbox does a lot of TR mapping and this is one of their important workflows.  But I'll think about whether we can use clicks on different parts of the arrows to do different things.

> Unhiding selectively level by level, as mentioned in my previous comment, would give the user control which path to use.

Yeah intersections with cycles in them and multiple via-ways is a big issue.  I think you'll like the slider I made. It's still rough but I like the idea of giving users control of how much detail they want to see (and defaulting them to "not much")

max detail via way


BTW - I'm using this intersection as my "worst case scenerio" `20.00/39.89071/-75.08993`.  I don't think it is modeled very well, and I would do it differently, but it is good for testing, so please don't edit it!

BTW2 - I also added the ability to tag "only" restrictions, which makes the above intersection _much_ easier to reason about.

> I think the example can be reduced to using one via way only, but in general we need more than two via members.

Yes I would also model that u-turn as a single via way.  The code could handle any number, but I really don't want to go above 2 because it would encourage people to build unnecessarily complicated things.

slhh:
>>    Place a start arrow on the starting way pointing directly to the connecting node.

>will consider, but it seems like clutter.

I was definitely missing this, but highlighting all ways along the path seems to make it dispensable now. Maybe, we will need the arrows to control or indicate something later.

>>    Do only show the first level of restriction icons.

>I've added a slider that will let users expand the search (see below)

It doesn´t expand selectively, and therefore, it doesn't help to control the via path.

>>    The restriction editor doesn't really fit into the inspector of a connecting vertex anymore. Showing properties of a selected node above and below, while editing restrictions which aren't related to the node, can be confusing.

>Agree, but this is #3136 and I don't have time to do it now.

#3136 doesn't seem to address the logical misfit.
We could also avoid avoid the logical misfit by showing restrictions only, which are touching the selected node. This  might be a benefit anyway, because it gives the user a limited control of the via path and reduces conflicts.

>>    Show the full list of generated relations below the restrictions editor.

>Probably wont at this time - for some very complex intersections (like an SPUI) it can be several dozen. But if we do #3136, this would be replaced by a summary view that might list them.

Agree, it doesn't make sense at the current position of the restriction editor.
Maybe, we can address the main usecases differently:

Make the relation accessible:
- A right click on the restriction icon should link to the relation. This could be indirectly though a context menu listing multiple conflicting (in view of using same from-to pair) restrictions.

Get an overview of existing relations:
- Make the icons, which are representing potential restrictions only or sideeffects of restrictions only, smaller or lighter. This means the red ones which are generated by "ONLY" restrictions, and all green ones.
- Add a checkbox to display existing ones only, maybe next to the slider.

>>    I have found another bug. A very short (e.g. 5 m) deadend way is stretched to the frame in the restriction editor. This looks quite strange.

>Yeah this is intentional - I do need to show the user something here and it needs to be long enough to be clickable even with an arrow stuck on top of it, but I think I don't need to extend it quite so far. I'll try a shorter amount. (can you link to the intersection?)

I don't think the extension is really required. There can also be non dead ended short ways in the junction.
You can't extend there, and these do need a solution to be clickable too. Currently it doesn't seem to be an issue anyway, because the user can click in free space to remove all icons.

> can you link to the intersection?

No, it wasn't a real one. I drew a short dead ended way to the inside of the junction for testing.
It was extended to the frame across other ways. You should definitely stop extending the way before intersecting another way.

>>    Click on the arrow part of the icons to extend as I mentiond in my previous comment.

>Part of the main design goals for this editor is to continue to allow mappers to create restrictions with one click. Our data team at Mapbox does a lot of TR mapping and this is one of their important workflows. But I'll think about whether we can use clicks on different parts of the arrows to do different things.

The frequent simple restrictions would still one click only, but the rare complex ones would require more clicks. This doesn't seem to degrade efficiency significantly.

My approach of unhiding selectively level by level along a path can also be combined with your slider:
Let the slider define the default display, and let the user expand further if required.
Maybe, the slider setting can be automatically reduced to avoid any display conflicts due to different via paths.

    Unhiding selectively level by level, as mentioned in my previous comment, would give the user control which path to use.

>I think you'll like the slider I made. It's still rough but I like the idea of giving users control of how much detail they want to see (and defaulting them to "not much")
 Hiding existing restrictions totally doesn't seem to be good. We need to indicate that there is something hidden. This can either be a warning icon at the slider, or more a more specific indicator at the arrow part of the restriction icons of the last displayed level.

slhh:
Unbranched consecutive ways should be considered as if they were merged in view of selecting, placing restriction icons, and via count related to the slider.

Another method to address the multiple via paths issue:
Use e.g. a yellow restriction icon to indicate the conflicting state.
Hovering the icon should highlight the full set of via paths, and instruct the user to shift-select via segments, which should hide any restriction icons of via paths not containing all shift-selected ways.
As long as the conflict persists, toggling the restriction icon should be prevented.
The slider limit should be overidden to allow at least the shortest via path containing all shift-selected ways. Shift selecting a way touching the frame indicates that the next connecting node is part of the via path, the displayed region needs to be adjusted accordingly. Shift-clicks on restriction icons should always target the underlaying way to make short ways selectable even if hidden by the icon. Clicking in free space first isn't an option here, because the other selected ways would become unselected.

I'm not sure wheather cyclic via paths are valid, if yes we should ignore them as long as the shorter variant without cycle isn't filtered out. If we were considering the cyclic variant as a conflict to the shorter one, no shift selection fitting the shorter one only would be possible.

data/core.yaml
```diff
@@ -219,14 +219,41 @@ en:
...
         create: Added a turn restriction
         delete: Deleted a turn restriction
+  restriction:
+    controls:
+      distance: Distance
+      distance_up_to: "Up to {distance}"
+      via: Via
+      via_node_only: "Node only"
+      via_up_to_one: "Up to 1 way"
+      via_up_to_multiple: "Up to {num} ways"
```
1ec5: This is another place where non-anglocentric plural support would be desirable: #597.

data/core.yaml
```diff
@@ -219,14 +219,41 @@ en:
...
+      allowed: Allowed
+      indirect: "(indirect)"
+      turn:
+        left_turn: Left Turn
+        right_turn: Right Turn
+        u_turn: U-Turn
+        straight_on: Straight On
+      no_turn_string: "{no} {turn} {indirect}"
+      only_turn_string: "{only} {turn} {indirect}"
+      allowed_turn_string: "{turn} {allowed} {indirect}"
```
1ec5: In some languages, it may be difficult to translate these `turn` strings independently of the overall sentence. A more sound approach would be to multiply out these strings by all the possible combinations, e.g., “ONLY Left Turn”. In the meantime, languages that have problems with these strings may be able to use the imperative form, e.g., “ONLY Turn Left”.

data/core.yaml
```diff
@@ -219,14 +219,41 @@ en:
...
+      from: FROM
+      via: VIA
+      to: TO
+      from_name: "{from} {fromName}"
+      from_name_to_name: "{from} {fromName} {to} {toName}"
```
1ec5: Languages with declension rules won’t be able to translate this well, but I guess the fact that the prepositions are **bold** makes it feel more like a form and less like a coherent sentence anyways.

data/core.yaml
bhousel: Ok, makes sense.. The slider only goes up to 2 via ways, and I'm not even sure whether it's worth making it any more extensible than that, so I'm just going to hardcode the string as "Up to 2 ways" to make things easy on the translators.

data/core.yaml
bhousel: Good point.. I'll just enumerate all of these (and leave the `{indirect}` as a token replacement)

data/core.yaml
bhousel: Yeah this is extra tricky because the `{fromName}` and `{toName}` might not actually be proper named roads, but rather strings like "Residential Road" or "Service Road" which could follow different rules depending on the modifier.  We don't support different conjugation of the preset name strings, so I'll punt on this one for now.  (But yeah, hoping the way we present it in the UI makes it seem more like a filled in form than a grammatically incorrect sentence).

bhousel:
Thanks everyone who tested this out and provided feedback!
I haven't commented much but I've been working pretty steadily on it for a few weeks, and it's ready to merge.

The turn restriction editor has learned a few new tricks:

* **Hovering to preview restrictions** - Hover over a FROM way to preview all the turn restrictions along allowed paths  
tr_preview

* **More color hinting** - thanks @rasagy for the suggestion!

* **Sliders for Max Distance and Max Via Ways** - I realized from testing this that there is no universal best "max distance" value for how far to search from the initial selected vertex to find more connected ways.  Some parts of the world have very compact roads, while others have their dual carriageways almost 50 meters apart.   

* **Support for only_ restrictions** - Now you can click toggle to add `only_` turn restrictions.  Creating an `only_` turn restriction will remove all other turn restrictions from the FROM way, but they will be remembered so that if you click again to remove the `only_` restriction, any previous `no_` restrictions will be restored.
tr_toggle

*  **Field Help** - This field is unusually complicated for iD.  I struggled a lot to make it simple, and ended up building a special help component for the field.  This can be used for other fields too, but currently only works for this one.  We've considered (and rejected) the idea before for special help content on unusually confusing fields, but this was more in the context of offering tag guidance (like explaining concepts like "tracktype" or "smoothness").
tr_help

Issue #2898, opened 2016-01-07
Support `addr:place` in the Address preset


johanneshoehn:
Some villages don't have street names, but use the place name instead of a street name in addresses.
This is very common in my region. For instance the village Hohlweiler belongs to the city Scheinfeld and has no street names. Addresses are of the scheme: (name), Hohlweiler (housenumber), 91443 Scheinfeld.

The [OSM wiki recommends](http://wiki.openstreetmap.org/wiki/Key:addr:place) using `addr:place` instead of `addr:street` in those cases. iD only supports `addr:street`. It would be nice to be able to select wether to input `addr:place` or `addr:street` and even better if there was some kind of automatic detection that the name matches up with the place name.

stephan75:
That whenever this feature request should be implemented, please have a validation that addr:place can only be added to an object when addr:street is NOT used. Because according to the OSM wiki definition these two keys should not co-exist on an object!

Natsuyasumi:
I guess it could be solved by adding a tickbox which would _swap_ `addr:street` with `addr:place`.

RicoElectrico:
I second this as very needed (as much as 6M uses of addr:place worldwide). There are many confused newbies, here's one: http://www.openstreetmap.org/changeset/47780046 . They always struggle with it, and it's not even their fault.

@bhousel If you need help on figuring out how to do this properly UX-wise, I think you may ask out on the mailing list. I don't know about other countries, but in our case it would suffice to have "place with street names"/"place without street names" option, or a tickbox with "address without a street name" which as mentioned would change "street" to "place".

bhousel:
> I don't know about other countries, but in our case it would suffice to have "place with street names"/"place without street names" option, or a tickbox with "address without a street name" which as mentioned would change "street" to "place".

Yes, I am ok with sticking a radio button somewhere at the top of the address block in the situation where a country has multiple alternate address formats.  This is extra complicated though because iD would need to:

* Somehow add support for alternate per-country address formats to the address.json field schema (fine)
* Let the user tick the radio button to change the address field between alternate formats (easy)
* Detect how the radio button should be ticked when parsing existing address data (harder)

Natsuyasumi:
So, I think:
`addr:place == * && addr:street != *` --> Place without street names
`addr:place != * && addr:street == *` --> Place with street names
`addr:place != * && addr:street != *` (or simply no tags) --> Place with street names, beacause this kind of places are more common.


[...]

PR #4784, opened 2018-02-09
Polish Address Format fix

MKuranowski:
The current Polish address fields format includes `addr:place` alongside `addr:street`, which could make it prone to inexperienced users to enter both fileds - and this is, according to [OSM Wiki](https://wiki.openstreetmap.org/wiki/Key:addr:place), an error.

This PR simply removes the `addr:place` format. Main discussion on implementing `addr:place` is held under #2898.

Also, the Polish placeholders for `addr:place` and `addr:city` are the same ("Miejscowość"), which only makes more confusion.

### DIFF
--- a/data/address-formats.json
+++ b/data/address-formats.json
@@ -24,7 +24,7 @@
             ]
         },
         {
-            "countryCodes": ["at", "ch", "de", "si"],
+            "countryCodes": ["at", "ch", "de", "si", "pl"],
             "format": [
                 ["street", "housenumber"],
                 ["postcode", "city"]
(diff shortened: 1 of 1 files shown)


bhousel:
Looks good, thanks @MKuranowski !

maraf24:
@MKuranowski please revert this change.

bhousel:
Sorry @maraf24 but @MKuranowski and @Slawek234 raise some good points..

Supporting multiple alternate per-country address formats per #2898 is the ideal solution we want to get to - not just putting all the fields on one form and expecting the user to know which ones to use.

PR #4808, opened 2018-02-16
reduced search priority for boat builder

bencostamagna:
Issue:
When drawing an area and typing "build" in the search bar, "boat builder" is the first result and "building" is only second. This doen't sound like something a user would expect and is a bit annoying when drawing buildings.

Solution:
I changed the match Score of the boat builder preset, "build" now returns building as a first result while "builde" yields boat builder.


bhousel:
LGTM, thank you!

PR #4809, opened 2018-02-16
do not stop mapillary autoplaying if _mlyViewer is not initialized (closes #4804) 

### LINKED ISSUE #4804: Mapillary photo overlay can't get disabled
Hello

After enabling Mapillary photo overlay and later disabling them, the overlay doesn't disappear:

untitled

There is no problem with OpenStreetCam overlay, though.

As this misbehaviour is new, I suspect it has been introduced with the last update of iD.

Browser used: Safari 11.0.3 on macOS 10.12.6

fritruc:
If we don't open the image viewer, _mlyViewer is not initialized and when we toggle off the layer, an exception is throw because _mlyViewer is not valid.

### DIFF
--- a/modules/services/mapillary.js
+++ b/modules/services/mapillary.js
@@ -491,7 +491,7 @@ export default {
     hideViewer: function() {
         _mlySelectedImage = null;
 
-        if (!_mlyFallback) {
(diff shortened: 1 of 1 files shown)


bhousel:
Good catch!  thanks for the fix 👍

PR #4814, opened 2018-02-17
Load WMS tiles if EPSG:3857 (or equivalent) is supported by source

### LINKED ISSUE #1141: WMS support

tyrasd:
partially fixes #1141.


bhousel:
This looks great @tyrasd - is it still a WIP?  

It looks like you are planning to replace the `'{proj}'` with whichever supported projection `editor-layer-index` contains (currently hardcoded to 'EPSG:3857', which is the most common anyway).

tyrasd:
> is it still a WIP? It looks like you are planning to replace the '{proj}' with whichever supported projection editor-layer-index contains (currently hardcoded to 'EPSG:3857', which is the most common anyway).

Yeah, I'd like to still fix that hardcoded EPSG code [there](https://github.com/openstreetmap/iD/pull/4814/files#diff-5029207b7599da40f52ba1ddec797e50R116) (there might be some obscure services just supporting for example `EPSG:900913` but not `EPSG:3857`).

Also, before we merge/release this, it would be good if we checked the respective WMS layers in ELI and add missing information like geographic extent – I think I saw quite a few of these claiming to be world-wide (`"geometry": null`) which actually aren't.

tyrasd:
I'd say this can be merged now.

bhousel:
Awesome @tyrasd - this is going to open up lots more imagery available for use in iD.. thank you!
I'm going to merge and add some test for the token replacement..

Issue #4819, opened 2018-02-19
Make 'Add field'-dropdown higher


manfredbrandl:
Having more (seven/eight) items in the dropdown would improve useability:
![image](https://user-images.githubusercontent.com/5360273/36385294-734cf3ec-1592-11e8-8a94-fac3e11d4c60.png)

bhousel:
agree

bhousel:
Done in #4831 !

PR #4831, opened 2018-02-23
 Issue #4819 : Increase max height of combobox, so they can show more…

SteevenBosse:
#4819  item : 

Nothing terrible, I changed max-height css property of combo box to 245px. It increase the maximum shown items by 4.  (Now 8 items)

dropdown max


bhousel:
Thank you @SteevenBosse !   I tried it out and I agree a taller list works much better, thank you for this 👍

Issue #3744, opened 2017-01-11
Type checking


bhousel:
Lets look into using [Flow](https://flowtype.org/) to add some type safety to our code.  We can start out with the `geo` and `actions` modules and expand to other modules.  This will almost certainly catch bugs and improve development experience.

cc @kepta @jfirebaugh @tmcw for thoughts?

HolgerJeromin:
Flow is very good in type guessing but sometimes you have to help him. 
If you don't want to change the build pipeline Flow can use a "comment" type style where you can add type annotations into special comments.

bhousel:
> Flow is very good in type guessing but sometimes you have to help him.
> If you don't want to change the build pipeline Flow can use a "comment" type style where you can add type annotations into special comments.

Since we already use rollup for building, we should be able to just use a plugin to remove the flow annotations, if I understand it correctly. So thankfully any change to our build process should be minimal.

Psigio:
Hi - I've created a Proof Of Concept for this (PR at: https://github.com/openstreetmap/iD/pull/4837).  I've only added the annotations to a couple of files and updated the build process to include flow but hopefully it should be enough to demonstrate the process.  If you are happy with this approach, the annotations could gradually be added to the rest of the application.

bhousel:
Thanks @Psigio for getting this started!

We can gradually add Flow annotations to the entire codebase.  This is a great project for anyone who wants to contribute and learn more about the internals of iD (beginners are very welcome!) .

For anyone wants to help out with this, please tag yourself in a followup comment and claim a folder. 🤗 

The work to do is: 
1. Read up on Flow's [type annotations](https://flow.org/en/docs/types/).  
2. For each source file, 
  a. add the `// @flow` flag to the top of the file
  b. add Flow annotations to the functions in the file
3. You can do `npm run flow` to check the code for any issues.  Fix the issues if you want to (or leave them if you're not sure).
4. You can do `npm run flow:coverage` to see a report about well we're doing. 
5. Commit your work, using a commit message like:
    ```
    Added Flow annotations to functions in the `xxxx` submodule
    (re: #3744)
    ```
6. Submit a Pull Request when you are finished your folder (or a few folders, if you want to do several).

✋ Don't be afraid to ask for help if you're not sure about something - I've never used Flow before so I am also a beginner.  

TODO:

* [ ] `actions/`
* [ ] `behavior/`
* [ ] `core/`
* [ ] `geo/`
* [ ] `lib/`
[...]

PR #4837, opened 2018-02-25
Start adding flow annotations, plus updates to build and linting processes

Psigio:
This is an initial Proof Of Concept for using flow for https://github.com/openstreetmap/iD/issues/3744.  I've added some annotations to a couple of utils files and updated the build process, including using a rollup plugin to remove the annotations during the build.  The build now runs flow as part of the "run all" script and will report any errors found by flow which will prevent the build continuing.

I've also updated the eslint config to be compatible with the flow annotations, which required use of the eslint-babel parser (see https://github.com/eslint/eslint/issues/2157).  I've used the default config for the https://github.com/gajus/eslint-plugin-flowtype plugin.

If this PR looks ok and is along the lines you were thinking, flow annotations could be added gradually to the rest of the codebase.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,15 +5,20 @@
         "es6": true
     },
     "extends": [
-        "eslint:recommended"
+        "eslint:recommended",
+        "plugin:flowtype/recommended"
     ],
     "globals": {
         "d3": false
     },
+    "parser": "babel-eslint",
     "parserOptions": {
         "ecmaVersion": 6,
         "sourceType": "module"
     },
+    "plugins": [
+        "flowtype"
+    ],
     "rules": {
         "dot-notation": "error",
         "eqeqeq": ["error", "smart"],
@@ -67,5 +72,10 @@
         "space-unary-ops": "error",
         "wrap-regex": "off",
         "quotes": ["error", "single"]
+    },
+    "settings": {
+        "flowtype": {
+            "onlyFilesWithFlowAnnotation": true
+        }
     }
 }
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,11 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[lints]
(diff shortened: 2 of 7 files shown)


bhousel:
Hey @Psigio just wanted to give my feedback that I really like this!  I chatted with @kepta a bit yesterday and we are 👍 with adding flow annotations to the entire iD codebase gradually.

I would merge this right now, but I'm releasing a new version of iD today.  So I'll probably take a closer look over the weekend.  Thanks!

bhousel:
Looks good, I just merged this.. 

I also added [flow-coverage-report](https://github.com/rpl/flow-coverage-report), so we can check on our progress with `npm run flow:coverage`

screenshot 2018-03-03 23 29 12

41%! 📈

Psigio:
I'm not sure that percentage looks right for only three files with Flow annotations :smiley: - still it will be interesting to see how this value changes as more annotations are added. What's the best approach for adding more annotations? Will you keep the original issue open to track pull requests where the commits have only Flow annotations (rather than those with code changes)?

HolgerJeromin:
Flow is very good in guessing the type.
If your code has four 
var foo=true;
and six complex variables you get 40% coverage

bhousel:
> still it will be interesting to see how this value changes as more annotations are added. What's the best approach for adding more annotations? Will you keep the original issue open to track pull requests where the commits have only Flow annotations (rather than those with code changes)?

Sure, that seems like a good way to proceed.. I'll make a list of all the modules and we can check them off as people add the Flow annotations..

Also, yeah as @HolgerJeromin mentioned - the coverage statistics are kind of funny, but makes some sense.  For each file, there is a percentage of lines that are just trivially obvious.  So if we were to just do nothing but add the `// @flow` to the top of the file, it would start out with some base percentage complete already.

screenshot 2018-03-04 17 38 28

Issue #4843, opened 2018-03-02
Support non web mercator WMS


tyrasd:
follows #1141:

> JOSM fakes EPSG:4326 and EPSG:3875/900913 interoperability by stretching a 4326 image a little bit vertically in-code. […] maybe iD could do a similar thing?

see also https://github.com/mapbox/whoots-js/issues/19

andrewharvey:
is this all done now that #4858 was merged or is there still more to do?

tyrasd:
Well, we could try to add support for other non web mercator projections (apart from plate carrée = EPSG:4326). Not sure if it would be worth the effort, though.

andrewharvey:
> Well, we could try to add support for other non web mercator projections (apart from plate carrée = EPSG:4326). Not sure if it would be worth the effort, though.

Fair point!

c933103:
Is it possible for iD editor to support WMS in polar stereographic projection?

PR #4858, opened 2018-03-07
add support for 4326 WMS layers

tyrasd:
for #4843, follows #4814 – adds support for WMS'es supporting EPSG:4326 (WGS 84).

Are there any alternative codes for this projection around? I've seen `CRS:84` in some places ([examle](https://map.bern.ch/wms/OpenData/proxy.php?request=GetCapabilities)), but I'm not actually sure where that one actually originates from…

Also, it would be nice to do a similar analysis as in https://github.com/osmlab/editor-layer-index/issues/425 for EPSG:4326 as well.

### DIFF
--- a/data/update_imagery.js
+++ b/data/update_imagery.js
@@ -35,6 +35,7 @@ var blacklist = {
 };
 var supportedWMSProjections = [
     'EPSG:3857',
+    'EPSG:4326',
     'EPSG:900913', // EPSG:3857 alternatives codes
     'EPSG:3587',
     'EPSG:54004',
--- a/modules/renderer/background_source.js
+++ b/modules/renderer/background_source.js
@@ -32,23 +32,6 @@ function vintageRange(vintage) {
     return s;
 }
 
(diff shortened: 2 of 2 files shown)


bhousel:
> Are there any alternative codes for this projection around? I've seen CRS:84 in some places (examle), but I'm not actually sure where that one actually originates from…

Interesting - that source also supports EPSG:4326.  It looks like it's the same as CRS:84 but with x and y switched 🤔

Catclaws12:
> See #4843, follows #4814 – adds support for WMS'es supporting EPSG:4326 (WGS 84).
> 
> 
> 
> Are there any alternative codes for this projection around? I've seen `CRS:84` in some places ([examle](https://map.bern.ch/wms/OpenData/proxy.php?request=GetCapabilities)), but I'm not actually sure where that one actually originates from…
> 
> 
> 
> Also, it would be nice to do a similar analysis as in https://github.com/osmlab/editor-layer-index/issues/425 for EPSG:4326 as well.

- - [ ]

Issue #4655, opened 2018-01-04
Self intersection prevention is too aggressive


bhousel:
It seemed like a good idea, but we should disable it when drawing new features

![self-intersecting-bad](https://user-images.githubusercontent.com/38784/34584195-d1d494e0-f168-11e7-8c10-d3008ec359c3.gif)

slhh:
I don't think we need to disable it, but a slight modification seems to be required:

- For a target placing a new node or connecting to an existing one exclude the closing edge from the intersection check.
- For a target finishing the area keep the full intersection check

bhousel:
Good suggestion @slhh, I'll try that 👍

PR #4860, opened 2018-03-08
Can Esc/finish drawing an Area when cursor is in invalid position

### LINKED ISSUE #4845: Escape does not always exit area creation, depending on mouse position
Hitting Escape to end area creation does not always work.

When I am creating a new area in iD, I will create nodes clicking with the mouse, and then hit "escape" on the keyboard once all the nodes are made. When I hit escape to finish area creation, I do not care where the mouse is pointing, since it should have no relevance, so the mouse is often in a meaningless location.

In a recent update of iD, a feature was added to detect invalid area geometries, like this: https://thuejk.dk/id_escape_bug.png . Note the "denied" sign. This seems to have had the obviously unintended consequence of also disabling the escape key functionality, if the mouse is hovering in an invalid spot. Note that this does not mean an invalid node existing, just the mouse hovering the wrong place.

jguthrie100:
Fixes #4845 

Can press Esc/Return when cursor is in invalid position and the drawing will still complete

An exception is made for when the completed Area would be invalid (i.e. when a temporarily invalid leading edge has been made as per #4655 , which cannot be accepted as a valid edge when closing the Area)

### DIFF
--- a/modules/behavior/draw_way.js
+++ b/modules/behavior/draw_way.js
@@ -91,16 +91,16 @@ export function behaviorDrawWay(context, wayId, index, mode, startGraph) {
 
         context.replace(actionMoveNode(end.id, loc));
         end = context.entity(end.id);
-        checkGeometry(origWay.isClosed());    // skipLast = true when drawing areas
+        checkGeometry(false);
     }
 
 
     // Check whether this edit causes the geometry to break.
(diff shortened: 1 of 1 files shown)


bhousel:
Hey @jguthrie100, I just tested it out and it works great - thanks so much for the PR 🙇

Issue #4522, opened 2017-11-11
Add field for advisory speed limit to link presets


1ec5:
Advisory speed limits are typically posted on freeway ramps and tight curves on surface roads. They can coexist with legal speed limits, but at least in the U.S., freeway ramps have only advisory speed limits, not legal speed limits. [`maxspeed:advisory`](https://wiki.openstreetmap.org/wiki/Key:maxspeed:advisory) indicates the advisory speed limit of a way. The link way presets could have an “Advisory Speed Limit” field in addition to (or instead of?) the existing “Speed Limit” field. I don’t think it’s as important to add the field to the rest of the Highway presets, since mapping advisory speed limits on curves is another level of micromapping beyond mapping them on ramps.

MaciejWWojcik:
@1ec5 Hi, can you provide some example of this maxspeed:advisory? I'm new here and don't know where should I start.

1ec5:
[This way](https://www.openstreetmap.org/way/19051632) is tagged `highway=motorway_link` `maxspeed:advisory=30 mph` (and not tagged `maxspeed`):

[![openstreetcam](https://user-images.githubusercontent.com/1231218/37245832-0a67ec72-2453-11e8-8a3f-3619da9c6362.png)](https://openstreetcam.org/details/1121191/264)

[This way](https://www.openstreetmap.org/way/528915276) is tagged `highway=trunk` `maxspeed=45 mph` `maxspeed:advisory=25 mph`. You can legally go 45&nbsp;mph, but that would be dangerous around this curve.

bhousel:
This was done in #4870

PR #4870, opened 2018-03-10
Advisory speed limit added 

umarpreet1:
i created advisory speed limit field and added it to preset highway=road and highway=motorway #4522

### DIFF
--- a/data/presets/presets/highway/motorway.json
+++ b/data/presets/presets/highway/motorway.json
@@ -5,7 +5,6 @@
         "ref_road_number",
         "oneway_yes",
         "maxspeed",
-        "maxspeed_advisory",
         "lanes",
         "surface",
         "structure",
--- a/data/presets/presets/highway/motorway.json
+++ b/data/presets/presets/highway/motorway.json
@@ -5,6 +5,7 @@
         "ref_road_number",
         "oneway_yes",
         "maxspeed",
+        "maxspeed_advisory",
         "lanes",
         "surface",
         "structure",
--- a/data/presets/presets/highway/motorway_link.json
+++ b/data/presets/presets/highway/motorway_link.json
@@ -4,7 +4,7 @@
         "name",
         "ref_road_number",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
         "lanes",
         "surface",
         "structure",
--- a/data/presets/presets/highway/primary_link.json
+++ b/data/presets/presets/highway/primary_link.json
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
         "lanes",
         "surface",
         "maxheight",
--- a/data/presets/presets/highway/secondary_link.json
+++ b/data/presets/presets/highway/secondary_link.json
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
         "lanes",
         "surface",
         "structure",
--- a/data/presets/presets/highway/tertiary_link.json
+++ b/data/presets/presets/highway/tertiary_link.json
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
         "lanes",
         "surface",
         "structure",
--- a/data/presets/presets/highway/trunk_link.json
+++ b/data/presets/presets/highway/trunk_link.json
@@ -4,7 +4,7 @@
         "name",
         "ref_road_number",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
         "lanes",
         "surface",
         "structure",
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1139,6 +1139,12 @@
             "label": "Speed Limit",
             "placeholder": "40, 50, 60..."
         },
+        "maxspeed_advisory":{
+            "key": "maxspeed_advisory",
+            "type": "maxspeed",
+            "label": "Advisory Speed Limit",
+            "placeholder": "40, 50, 60..."
+        },
         "maxstay": {
             "key": "maxstay",
             "type": "combo",
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -1139,8 +1139,8 @@
             "label": "Speed Limit",
             "placeholder": "40, 50, 60..."
         },
-        "maxspeed_advisory":{
-            "key": "maxspeed_advisory",
+        "maxspeed/advisory":{
+            "key": "maxspeed:advisory",
             "type": "maxspeed",
             "label": "Advisory Speed Limit",
             "placeholder": "40, 50, 60..."
--- a/data/presets/fields/advisory.json
+++ b/data/presets/fields/advisory.json
@@ -1,6 +1,0 @@
-{
-    "key": "maxspeed:advisory",
-    "type": "maxspeed",
-    "label": "Advisory Speed Limit",
-    "placeholder": "40, 50, 60..."
-}
--- a/data/presets/fields/advisory.json
+++ b/data/presets/fields/advisory.json
@@ -1,5 +1,5 @@
 {
-    "key": "maxspeed_advisory",
+    "key": "maxspeed:advisory",
     "type": "maxspeed",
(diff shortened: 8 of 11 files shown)


data/presets/presets/highway/motorway.json
```diff
@@ -5,6 +5,7 @@
         "ref_road_number",
         "oneway_yes",
         "maxspeed",
+        "maxspeed_advisory",
```
1ec5: Normally advisory speed limits appear on `motorway_link` and other `*_link` roads. It can happen on non-`link`s too, but typically only at curves.

umarpreet1: Ok so i have to add this field to all *link roads or something else that i should do.

1ec5: Yes, it would make sense to add this field to all the `*_link` presets. Maybe even remove `maxspeed` in favor of `maxspeed_advisory` on those presets.

The non-`link` presets should be consistent with each other. So if `motorway` has it, then the other `highway` presets like `residential` and `tertiary` should too. For these presets, an extra “Advisory Speed Limit” field could be confusing if it appears by default, since most roads don’t have advisory speed limits. I wonder if we should leave it off of these non-`link` presets until #4871 is implemented.

bhousel:
This is a great start @umarpreet1 - some quick feedback:

- To echo what @1ec5 said - I agree these should only be on `_link` way presets for now.

- We have a naming convention for colon-namespaced fields.  Can you:
  - move the field definition file to  `data/presets/fields/maxspeed/advisory.json`
  - inside that file, `"key": "maxspeed:advisory"`
  - when referring to the field in a preset, it should be called `"maxspeed/advisory"`

slhh:
The key maxspeed:advisory is a rarely used key (12500 uses) and seems to be applicable mainly in the US and a few European countries only: https://taginfo.openstreetmap.org/keys/maxspeed%3Aadvisory#map
For comparision, the key maxspeed is used approximately 9 millon times.

1ec5:
To be sure, I wasn’t suggesting that `maxspeed` be removed from _all_ the `highway` presets, just the `*_link` ones. You’re right that `maxspeed:advisory` is much rarer than `maxspeed`. After all, advisory speed limits are rarer than legal speed limits in reality:

* [100,502 ways](https://taginfo.openstreetmap.org/tags/highway=motorway_link#combinations) are tagged `highway=motorway_link` and `maxspeed`.
* [8,588 ways](https://taginfo.openstreetmap.org/tags/highway=motorway_link#combinations) are tagged `highway=motorway_link` and `maxspeed:advisory`.

Specifically in the U.S., where legal speed limits are virtually never posted on freeway ramps:

* [11,439 ways](http://overpass-turbo.eu/s/wTg) are incorrectly tagged `highway=motorway_link` and `maxspeed`.
* [6,977 ways](http://overpass-turbo.eu/s/wTh) are correctly tagged `highway=motorway_link` and `maxspeed:advisory`.

One reason I opened #4522 is that mappers unaware of the `maxspeed:advisory` tag end up putting advisory speed limits in `maxspeed`, even when it’s semantically incorrect. There is some risk that presenting both fields alongside each other could confuse some mappers, which is why I raised the possibility of removing `maxspeed` from `*_link` presets only. But there’s a help button in case anyone is wondering what an advisory speed limit is.

In any case, I don’t want to derail this PR with this relatively minor question. @bhousel’s suggestions in https://github.com/openstreetmap/iD/pull/4870#issuecomment-372072423 would leave `maxspeed` in place in all the presets, which is fine for now.

umarpreet1:
i have changed the following in as mentioned by you:-
1.changed the naming convention and file convention as stated by you.
2.removed previous advisory speed limit from motorway and road.
3. replaced max speed limit by advisory speed limit in all the *link ways in highways preset.

slhh:
@1ec5 

> To be sure, I wasn’t suggesting that maxspeed be removed from all the highway presets, just the *_link ones.

Removing maxspeed from link presets would generate the inverse problem in other countries.
E.g. in Germany, limits on ramps are always legal maxspeed limits, a traffic sign to post advisory limits doesn't exist.
Even in the U.S., you might need maxspeed on ramps. Doesn't the general legal speed limit of the freeway persist, where an advisory limit is posted?

althio:
> > @1ec5 Maybe even remove maxspeed in favor of maxspeed_advisory on those presets
>> ...
> > To be sure, I wasn’t suggesting that maxspeed be removed from all the highway presets, just the *_link ones.

> @slhh Removing maxspeed from link presets would generate the inverse problem in other countries.

I agree with @slhh 
Add maxspeed_advisory on *_link if it feels better for some countries, but do not remove plain maxspeed on *_link, just propose both.

1ec5:
> Even in the U.S., you might need maxspeed on ramps. Doesn't the general legal speed limit of the freeway persist, where an advisory limit is posted?

As with anything in the U.S., the answer varies by state. In some states, [no legal speed limit](https://law.stackexchange.com/a/1348) applies to on- and off-ramps. In others, [the freeway’s legal speed limit technically applies](http://pugetsoundblogs.com/roadwarrior/2012/04/24/speed-limit-rule-the-same-for-off-and-on-ramps/). But either way, driving over the advisory speed limit could subject the driver to a citation for dangerous driving according to the [basic rule](https://en.wikipedia.org/wiki/Speed_limit#Reasonable_speed), which makes the legal speed limit a moot point.

>> Removing maxspeed from link presets would generate the inverse problem in other countries.
> 
> Add maxspeed_advisory on *_link if it feels better for some countries, but do not remove plain maxspeed on *_link, just propose both.

Agreed. It was just a suggestion, and it’s clear from this discussion that `maxspeed` shouldn’t be removed until #4871 is implemented.

(Thanks for putting up with this back-and-forth, @umarpreet1!)

data/presets/presets/highway/motorway_link.json
```diff
@@ -4,7 +4,7 @@
         "name",
         "ref_road_number",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
```
Nakaner: `maxspeed=*` is a valid tag on `higway=motorway_link` in some countries and should not be removed.

data/presets/presets/highway/primary_link.json
```diff
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
```
Nakaner: maxspeed=* is a valid tag on higway=primary_link in some countries and should not be removed.

data/presets/presets/highway/secondary_link.json
```diff
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
```
Nakaner: maxspeed=* is a valid tag on higway=secondary_link in some countries and should not be removed. `highway=secondary_link` in settlements where the normal urban speed limit (e.g. 50 kph in many European countries) applies should get `maxspeed=50` or `maxspeed=<country_code>:urban`.

data/presets/presets/highway/tertiary_link.json
```diff
@@ -3,7 +3,7 @@
     "fields": [
         "name",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
```
Nakaner: maxspeed=* is a valid tag on higway=tertiary_link in some countries and should not be removed. `highway=secondary_link` in settlements where the normal urban speed limit (e.g. 50 kph in many European countries) applies should get `maxspeed=50` or `maxspeed=<country_code>:urban`.

data/presets/presets/highway/trunk_link.json
```diff
@@ -4,7 +4,7 @@
         "name",
         "ref_road_number",
         "oneway",
-        "maxspeed",
+        "maxspeed/advisory",
```
Nakaner: maxspeed=* is a valid tag on higway=trunk_link in some countries and should not be removed.

bhousel:
Just catching up on the thread... 

I'm apologize - what I wrote above in my previous comment was confusing.  I meant to say that link roads should continue to have the `maxspeed` field, but they should also be the only roads that have the `maxspeed:advisory` field.

I'll cleanup and merge - thanks for your review @Nakaner @slhh @althio & @1ec5

Issue #4855, opened 2018-03-07
man_made=observatory


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dobservatory

Issue #4826, opened 2018-02-20
Add preset for building=farm_auxiliary and rename preset for building=farm


lakedistrictOSM:
`building=farm_auxiliary`
- https://wiki.openstreetmap.org/wiki/Tag:building=farm_auxiliary
- a non-residential farm building
- name: Farm Building

`building=farm`
- https://wiki.openstreetmap.org/wiki/Tag:building_farm
- a residential building on a farm
- name: Farm House

MaciejWWojcik:
Hello, sorry for that stupid questions - I'm new and want to learn and help. 

I want to make this issue, so I writed my code in presets.yaml and then start npm test, because of contributing.md

There I have two problems:
1. My changes in yaml disappeared 
2. Test stuck at  iD.uiFieldWikipedia 2)"before each" hook

What I'm misssing?

bhousel:
Hey @MaciejWWojcik thanks!   
We have a special readme just for working with presets and fields:
https://github.com/openstreetmap/iD/blob/master/data/presets/README.md

There is a lot of info in there, but the tl;dr is 
* presets and fields are all defined in the .json files
* run `npm run build` to rebuild the generated files (you found that the .yaml is one)

So for this request, the steps to do are:
1. copy `/data/presets/presets/building/farm.json` to `/data/presets/presets/building/farm_auxiliary.json`
2. edit `farm_auxiliary.json` and change the "tags" to contain `"building": "farm_auxiliary"`
3. edit `farm.json` and change the "name" from "Farm Building" to "Farm House"
4. `npm run build`

You can test whether it worked by running `npm run start` and using iD on localhost.  You should be able to find both presets "Farm House" (which will add the tag `building=farm`) and "Farm Building" (which will add the tag `building=farm_auxiliary`).

I'm not sure what is causing the other test failure, but it sounds probably unrelated.

MaciejWWojcik:
Okay, thanks a lot! I will work on this ;)

MaciejWWojcik:
#4839

PR #4889, opened 2018-03-14
Added preset for man_made=observatory.

vershwal:
In reference to #4855.
@bhousel Please let me know if I have missed anything.

### DIFF
--- a/data/presets/presets/man_made/observatory.json
+++ b/data/presets/presets/man_made/observatory.json
@@ -3,10 +3,9 @@
         "point",
         "area"
     ],
-    "type": [
+    "terms": [
         "astronomical",
-        "meteorological",
-        "espionage"
+        "meteorological"
     ],
     "tags": {
         "man_made": "observatory"
--- a/data/presets/presets/man_made/observatory.json
+++ b/data/presets/presets/man_made/observatory.json
@@ -1,0 +1,15 @@
+{
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "type": [
+        "astronomical",
+        "meteorological",
+        "espionage"
+    ],
+    "tags": {
+        "man_made": "observatory"
+    },
+    "name": "Observatory"
+}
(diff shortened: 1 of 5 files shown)


vershwal:
The tests are failing. Please suggest how can I fix the tests.

Xavier-J-Ortiz:
@vershwal Hello!

I believe this document could help you a bit more under the https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#adding-or-refining-presets section of the main CONTRIBUTING.md .

It's kind of easy to overlook. But there's a presets readme in that section that should clear things up.

As a final reference, check out this previous closed PR that had a similar issue which I believe that you're running into: https://github.com/openstreetmap/iD/issues/4826#issuecomment-369249797

data/presets/presets/man_made/observatory.json
```diff
@@ -0,0 +1,15 @@
+{
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "type": [
```
bhousel: This should be "terms"

data/presets/presets/man_made/observatory.json
```diff
@@ -0,0 +1,15 @@
...
+        "astronomical",
+        "meteorological",
+        "espionage"
```
bhousel: I'd remove this word "espionage", it's not really a term related to an [observatory](https://en.wikipedia.org/wiki/Observatory).

data/presets/presets/man_made/observatory.json
bhousel: By the way this is what is causing the test failures.   "type" doesn't fit here and so the build process will raise an error when it tries to validate the `observatory.json` file against the [preset schema](https://github.com/openstreetmap/iD/blob/master/data/presets/schema/preset.json).

vershwal:
Hey @bhousel !
Thanks for the review. Seems like it's working fine now.

bhousel:
Thanks!

PR #4905, opened 2018-03-16
Added preset height for buildings.

### LINKED ISSUE #2455: buildings should suggest height as well as levels
iD currently has a levels preset for buildings, but many buildings such as warehouses. supermarkets etc. have 20+ foot ceilings despite being only 1 level. For industrial buildings such as a silo the level count is nonsensical while the height is definitely meaningful. As 3-D rendering gains importance the actual building height also becomes more important.

vershwal:
Fixes #2455.

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -4197,6 +4197,7 @@
                 "name",
                 "building",
                 "levels",
+                "height",
                 "address"
             ],
             "geometry": [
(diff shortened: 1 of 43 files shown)


bhousel:
Hey @vershwal - this is good!

You can add it a lot more places too.. Use your code editor to search for the "levels" field and you'll find a lot of specific-kind-of-building presets under `data/presets/presets/building/* . Each of these can have a "height" field added exactly the same way.

vershwal:
Thanks @bhousel!!
I have added height field to all the buildings that have levels. Please check.

bhousel:
Thanks again @vershwal , looks great!

Issue #2300, opened 2014-07-22
Add support for “brand=*” for “amenity=fuel”


sommerluk:
It could be useful to have a text field for “brand=*” if working on an “amenity=fuel” object.

(JOSM has “brand=_” even before “name=_” in its template.)

afmenez:
I have to disagree, these 3 tags have different meanings.

Here in my region, for example, there's the "Buffon Network", which owns lots of stations with different brands (http://www.buffon.com.br/postos).  To correctly describe each station (like the B62 in their list), I would put the station name in the "name" tag (like "Posto Buffon Getúlio Vargas"), then put "Buffon" in the "operator" tag, and finally Shell in the "brand" tag.  I can then import the map to my GPS and see the brand icon in the map, and see "Shell: Posto Buffon Getúlio Vargas" if I search for nearby gas stations.  The "operator" tag is ignored in the conversion.  If I only add the brand to the name tag and then search for nearby stations, I'll get a lot of "Shell" stations with no way to know which is which.

My opinion is that the "brand" tag is way more important than the "operator" tag, and should be shown by default.

On a side note, there's the same exact problem for car dealerships.

brycenesbitt:
A more clear case is a marine fuel station ( https://github.com/openstreetmap/iD/issues/2589 ) where the operator is probably the marina, but the fuel probably comes from a major like "BP".  Unfortunately you might have multiple brands of fuel at one fuel dock.  These fuel docks are typically not accessible to cars, by the way.

naoliv:
Related with https://github.com/osmlab/name-suggestion-index/issues/29, in the future iD should stop to suggest some values for `name` and suggest `brand` instead.

d1g:
>  be encouraging contributors to simply use the "Name" field for things like "McDonalds", "Shell", etc

@jfirebaugh, I'm late to comment on it; but there clear distinction in "brand"/"name(s)" in shop=car

It was stated at wiki (Tag:shop=car) in quite clear manner:

`    name=John Doe One City + brand=Citroën;Ford;Honda;Mercedes-Benz`

@naoliv is absolutely correct that brands and names are different; but yes, in practice they often messed up or owners don't care to create names and simply use brand as a name.

Similar story with shops. Some shops have name but sell multiple so called "brands" with different name.

one company or manufacture can use multiple "brands"
brand is not company name ("Pepsico") https://en.wikipedia.org/wiki/PepsiCo#Products_and_brands
[...]

PR #4906, opened 2018-03-16
Add support for “brand=*” for “amenity=fuel”.

vershwal:
Related #2300.


bhousel:
Thanks @vershwal - this is fine, and I agree we should just change this and close out the issue.

vershwal:
Thanks @bhousel.

PR #4909, opened 2018-03-17
first addition of Flow annotations in 

Xavier-J-Ortiz:
`presets`. 

Refs: https://github.com/openstreetmap/iD/issues/3744

### DIFF
--- a/modules/presets/category.js
+++ b/modules/presets/category.js
@@ -1,9 +1,17 @@
+// @flow
+
 import _clone from 'lodash-es/clone';
 import { t } from '../util/locale';
 import { presetCollection } from './collection';
 
+type categoryObject = { 
+  geometry: string, 
+  icon: string, 
+  name: string, 
+  members: [string]
+};
 
-export function presetCategory(id, category, all) {
+export function presetCategory(id: string, category: categoryObject, all: () => mixed): categoryObject {
     category = _clone(category);
 
     category.id = id;
@@ -14,22 +22,22 @@ export function presetCategory(id, category, all) {
     }));
 
 
-    category.matchGeometry = function(geometry) {
+    category.matchGeometry = function(geometry: string): boolean {
         return category.geometry.indexOf(geometry) >= 0;
     };
 
 
-    category.matchScore = function() {
+    category.matchScore = function(): number {
         return -1;
     };
 
 
-    category.name = function() {
+    category.name = function(): string {
         return t('presets.categories.' + id + '.name', {'default': id});
     };
 
 
-    category.terms = function() {
+    category.terms = function(): [] {
         return [];
     };
 


Xavier-J-Ortiz:
@bhousel So, I've been taking a look at this today, and am a bit confused as to where to find some reference for the types.

I traced back some things from the spec files `test/spec/presets/category.js` for `modules/presets/category.js`however some of the types are unclear. For example: https://github.com/Xavier-J-Ortiz/iD/blob/cbe7e04e7e592c0e31a43064fda181cf804e685d/modules/presets/category.js#L14 I added `mixed => ()` because the `all` parameter in `presetCategory` is a function within it's spec test. However, this doesn't feel right and my gut says I'm not setting the type correctly here.

Also, `type categoryObject`is a type I created also based off of the `presetCategory`'s spec test: https://github.com/Xavier-J-Ortiz/iD/blob/cbe7e04e7e592c0e31a43064fda181cf804e685d/test/spec/presets/category.js#L5.

Is there a document, or a reference, that can show how these parameters are setup?

...or, if this should be traced, is there a place that I can get started in sleuthing my way to where these objects are built to have an idea of how to build them? I wouldn't know where the trail would begin. Thanks!

bhousel:
Thanks @Xavier-J-Ortiz !  Yeah this is a learning process for me too.

> I added `mixed => ()` because the `all` parameter in `presetCategory` is a function within it's spec test. However, this doesn't feel right and my gut says I'm not setting the type correctly here.

For a first pass, this is probably fine, and I would do the same..  Once we learn better what we're doing, we can add more type specificity if it makes sense.  

> Also, `type categoryObject` is a type I created also based off of the presetCategory's spec test

When I did the `geo` module, I ran into similar issues, and I ended up just creating types that matched the things that the Object needed to have.

Another thing that seems useful is:  I added common types into the `index.js` and export them:
https://github.com/openstreetmap/iD/blob/866aa650acb247f4ae17c5cacb887c4fb5247571/modules/geo/index.js#L1-L4

Then import them at the top of the files that use them.

https://github.com/openstreetmap/iD/blob/866aa650acb247f4ae17c5cacb887c4fb5247571/modules/geo/raw_mercator.js#L1-L2
https://github.com/openstreetmap/iD/blob/866aa650acb247f4ae17c5cacb887c4fb5247571/modules/geo/geom.js#L1-L4


Let me know if when you think it's ready to merge.. For this initial pass I'm really not aiming for 100% coverage.

Xavier-J-Ortiz:
@bhousel - Thanks for sharing.

I'll work on this tonight and tomorrow. Should be done by then.

bhousel:
Hey @Xavier-J-Ortiz thanks for working on this..
Per discussion on https://github.com/openstreetmap/iD/issues/3744#issuecomment-374950519  I'm going to close and remove the flow stuff that we've added.

Thanks for understanding!

Xavier-J-Ortiz:
@bhousel Yea! Not a problem!

Issue #4599, opened 2017-12-07
Add additional "X" to close side menus


jidanni:
The Background Settings, Map Data, and Help side menus all need an additional "X" to
close them placed in their upper left corners: ![d](https://user-images.githubusercontent.com/1230959/33720519-2a4efd70-db9f-11e7-8364-b7a7af3aa015.jpg)

Else the user thinks he must click the same spot, now blue, on the
rightmost side of the screen, that he used to open them, to close them.

That spot is far away from where he intends to continue working.

Yes it is true he can click outside of the menus to close them, but he
does not know that.

I think this second way to close the menus does not violate
https://www.nngroup.com/articles/duplicate-links/
too badly.

bhousel:
> Else the user thinks he must click the same spot, now blue, on the
> rightmost side of the screen, that he used to open them, to close them.

Yes, that is how to close them..  But I'm open to the idea of adding an 'X', since we use this design element everywhere else.

If I move the side toolbar down slightly, the 'X' could be in the upper right of the pane, which is how we position all the other close 'X's in iD - would this be ok?

jidanni:
Sure... however the user's next stop with his mouse is definitely to the left, not the right. So deep down he will be "cussing" the designers for, after he has chosen a radio button etc. from the menu, making him have to go back right to close the menu.

bhousel:
> So deep down he will be "cussing" the designers for, after he has chosen a radio button etc. from the menu, making him have to go back right to close the menu.

I'm ok with this.  Anyone who wants to work quickly can just toggle the menu with <kbd>B</kbd>.  We should probably bind <kbd>esc</kbd> to close the menu too.  Mousing will always be slower, no matter where we put the 'X'.

jidanni:
ESC is good. B should be called b if you ask me, else what do you call real B. OK never mind that. Mainly his hand is already on the mouse, and the next thing he wants to do also involves the mouse. So he shouldn't be expected to have to "fire up" his other hand to hit ESC. But anyway by this time he probably has learned that just clicking the map will close the menu. But he fears side effects. Anyway, he wishes that clicking the fat white left margin of the menu would close it.

slhh:
The user will likely become aware that a click on the map is sufficient, but adding the 'X' makes this more unlikely to discover.
Adding the X in its natural position (upper right corner) wouldn't be an benefit anyway.

[...]

PR #4913, opened 2018-03-18
Added button to close Map controls.

vershwal:
Ref. #4599
I have added a button for closing the Background Settings, Map Data, and Help side menus. This is how it looks.

screen shot 2018-03-18 at 5 28 10 pm

screen shot 2018-03-18 at 5 28 30 pm

@bhousel @jidanni I have reduced the padding as discussed in the issue, but I think the frontend part can be improved. What do you suggest?

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2727,12 +2727,12 @@ div.full-screen > button:hover {
     position: fixed;
     top: 60px;
     bottom: 30px;
-    padding: 20px 50px 20px 20px;
+    padding: 5px 35px 5px 5px;
     right: 0;
     overflow: auto;
 }
 [dir='rtl'] .map-overlay.content {
-    padding: 20px 20px 20px 50px;
+    padding: 5px 5px 5px 35px;
     left: 0;
     right: auto !important;
 }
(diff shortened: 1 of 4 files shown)


bhousel:
> I have reduced the padding as discussed in the issue, but I think the frontend part can be improved. What do you suggest?

Thanks @vershwal - we don't need to build exactly what the user requests.  What about this design do you think we should improve?

vershwal:
@bhousel I have changed the padding because it was discussed but I think it looked better earlier.
So I wanted your suggestions about whether we should reduce it or let it be the same.

vershwal:
Hey @bhouse, I have changed the padding(what I think is suitable) in commit 64b90a6.

bhousel:
Thanks @vershwal - I'm going to check this PR out today and try to get it merged.. I agree with keeping the padding as-is, and I will try to move the close button to be more consistent with other close buttons within iD  (top right when LTR, top left when RTL).

bhousel:
I ended up moving the map controls down a little bit, and changed the panes to have a fixed header section with title and close button to the right.   I also made Background and Map Data panes a little wider - they were starting to feel cramped.

Try it out in http://preview.ideditor.com/master/

screenshot 2018-03-27 20 59 17

vershwal:
Yeah!! It looks great.

Issue #4917, opened 2018-03-19
Handling of allotments=plot


eehpcm:
The editor currently doesn't seem to handle allotments=plot correctly (or I'm misunderstanding something, as usual).

According to the wiki [allotment=plot](https://wiki.openstreetmap.org/wiki/Tag:allotments=plot) is applied to an area.  Which I did.  But it seems that id is not treating it as a way with an implicit area=yes tag.

Why do I think that?  Because when I give the plot a name=* it is shown in the editor as a label on a way, not a label within an area.  This can be very confusing as allotments often have shared borders, so it's hard to determine which one the name applies to.  I kluge around this by adding an explicit area=yes just to get the name to show sensibly.  If tagging for the renderer is a venial sin then tagging for the editor is probably a mortal sin. :)

Should I be using name=* on a plot?  There's nothing to say I shouldn't and name applies to all objects.  True, the wiki on allotments=plot suggests using ref=whatever, but that doesn't get rendered.  Also, [the original proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Tag:allotments%3Dplot) gives [this example](https://www.openstreetmap.org/way/237662256) by the proposal's author where the plot in question has a name=* not a ref=*.

Is it worth bothering since it doesn't get rendered anyway?  In [this](https://github.com/gravitystorm/openstreetmap-carto/issues/432) github issue, kocio-pl appears, in Oct 2017, to have taken cognizance of a desire to display plots, so it might appear in the near future.  Even if it never makes it into OSM's renderer, it's still useful to have such detail in the db and therefore it's nice if the editor handles it well.

If you want to check this on an allotment with more plots and the area=yes kluge, look [here](https://www.openstreetmap.org/way/552820707)

bhousel:
> The editor currently doesn't seem to handle allotments=plot correctly (or I'm misunderstanding something, as usual).  According to the wiki allotment=plot is applied to an area. Which I did. But it seems that id is not treating it as a way with an implicit area=yes tag.

You're correct, we don't have a preset for `allotment=plot` yet.  We can add one for it.  After we add the preset, iD will know that it can be an area feature, and won't require the `area=yes` tag.

> Should I be using name=* on a plot? There's nothing to say I shouldn't and name applies to all objects. True, the wiki on allotments=plot suggests using ref=whatever, but that doesn't get rendered.

We can add fields for both `name` and `ref`.

> Is it worth bothering since it doesn't get rendered anyway?

[...]

PR #4920, opened 2018-03-20
Added preset for allotments=plot.

vershwal:
This PR adds preset for allotments=plot.
Related #4917.

### DIFF
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -535,6 +535,23 @@
             },
             "name": "Airport Terminal"
         },
+        "allotments/plot": {
+            "geometry": [
+                "area"
+            ],
+            "fields": [
+                "name",
+                "ref"
+            ],
+            "tags": {
+                "allotments": "plot"
+            },
+            "reference": {
+                "key": "allotments",
+                "value": "plot"
+            },
+            "name": "Community Garden Plot"
+        },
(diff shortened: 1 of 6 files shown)


bhousel:
Thanks @vershwal!
From reading [the wiki page](https://wiki.openstreetmap.org/wiki/Tag:allotments=plot), I think this preset is supposed to be its own feature, not combined with `landuse=allotments`.

so move:
`data/presets/presets/landuse/allotments/plot.json` 
to
`data/presets/presets/allotments/plot.json`

..and remove all references to landuse in that file.  It should just set the `allotments=plot` tag by itself.

data/presets/presets.json
```diff
@@ -8363,6 +8363,31 @@
...
+            },
+            "removeTags": {
+                "landuse": "allotments",
+                "allotments": "plot"
+            },
+            "reference": {
+                "key": "allotments",
+                "value": "plot"
+            },
+            "name": "Plot"
```
bhousel: Let's change the name to "Community Garden Plot".. This is how it would be called in `en-US`, which we use as the source language.

vershwal:
Hey @bhousel!!
Thanks for reviewing. I have made the required changes. Please check.

bhousel:
Perfect @vershwal - thanks again!

Issue #4923, opened 2018-03-21
natural=mud preset


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Tag:natural=mud

bhousel:
Whoever wants to work on this can just:

- copy `data/presets/presets/natural/sand.json` to `data/presets/presets/natural/mud.json`
- Edit `mud.json` contents to have appropriate tags, name, terms

Xavier-J-Ortiz:
I'll take a stab at this.

PR #4926, opened 2018-03-22
added mud to presets natural

Xavier-J-Ortiz:
Refs: https://github.com/openstreetmap/iD/issues/4923

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -4101,6 +4101,11 @@ en:
         # natural=heath
         name: Heath
         terms: '<translate with synonyms or related terms for ''Heath'', separated by commas>'
+      natural/mud:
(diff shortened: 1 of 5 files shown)


Xavier-J-Ortiz:
Not 100% sure about terms here, but I left it as `wetlands`. Mainly because I'm following the fact that in `sand.json` the value used in terms is `desert`, which is pointed out here: https://wiki.openstreetmap.org/wiki/Tag:natural=sand

Found the similar term in mud and adopted this.

bhousel:
Looks good, thanks @Xavier-J-Ortiz ...
Adding `wetland` in the terms is fine.. These are just words that users can search for to find the preset.

Issue #3630, opened 2016-12-01
Add communication: multiCombo


boothym:
For man_made=mast and man_made=tower there should be a multiCombo for the communication:*=yes tag, e.g.:

communication:radio:yes
communication:television:yes
communication:mobile_phone:yes
communication:microwave:yes

Also the name for man_made=mast should be changed to just "Mast".

bhousel:
> Also the name for man_made=mast should be changed to just "Mast".

I don't have any strong preference what it's called - maybe it was originally called "Radio Mast" to differentiate it from masts on boats?  The word "mast" also sometimes describes the riser where electrical service enters a building.

boothym:
Radio Mast was added here: https://github.com/openstreetmap/iD/pull/2613 - no mention of boats. Just seems weird to have radio in the name when a mast might not be used as a radio transmitter.  

> The word "mast" also sometimes describes the riser where electrical service enters a building.
No idea what that is, you'll need to show me a photo! :)

All of the `man_made=mast` terms should also be copied over to `man_made=tower` as the latter doesn't have any defined.

abdeldjalil09:
> Let's just leave it as "Radio Mast"

Masts are not always used for communication, this title is appropriate for man_made=mast + tower:type=communication tag

slhh:
>>     Let's just leave it as "Radio Mast"
> 
> Masts are not always used for communication, this title is appropriate for man_made=mast + tower:type=communication tag
> 
Unfortunately, it fits even better to tower:type=radio which is less frequently used.

Some other "tower:type", which are used in combination with man_made=mast, are lighting, siren, lightning_protection, and radar.

The key tower:type is defined in the wiki to allow user-defined values, and many man_made=mast don't have any specified tower:type.

Maybe we can call man_made=mast "Mast (Radio etc.)", "Mast (e.g. Radio Mast)", or "Mast (small tower)".

bhousel:
Ok, we can change it to just say "Mast".  I really don't care.

1ec5:
Maybe add “radio mast” as a synonym for this preset?

PR #4938, opened 2018-03-25
Create antenna.json

obama:
see https://wiki.openstreetmap.org/wiki/Proposed_features/Telecommunications_tower

### DIFF
--- a/data/presets/presets/man_made/antenna.json
+++ b/data/presets/presets/man_made/antenna.json
@@ -1,0 +1,17 @@
+{
+    "icon": "poi-mast",
+    "fields": [
+        "height"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "terms": [
+        "antenna",
+        "radio mast",
+    ],
+    "tags": {
+        "man_made": "antenna"
+    },
+    "name": "Antenna"
+}


bhousel:
Thanks @obama 👍

1ec5:
#2613 added a preset that could also be called “antenna” (and has been translated as such in at least one language). My understanding is that a standard guyed radio antenna would be tagged `man_made=mast`, as supported by [openstreetmap-carto](https://github.com/gravitystorm/openstreetmap-carto/blob/94fe461654bd1424e775a23207eed6ffb4aa2203/amenity-points.mss#L371-L392). Back when I added the preset, that would’ve been consistent with [this wiki page](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/1181155#First_question:_Is_it_a_mast_or_a_tower.3F), although the page has since been pared back; a separate page now describes the much rarer [`man_made=antenna`](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dantenna) tag and equivocates as to which tag should be used. So now I’m confused, and I can’t imagine the situation would be any better for the user. 😕

bhousel:
> My understanding is that a standard guyed radio antenna would be tagged man_made=mast, as supported by openstreetmap-carto. Back when I added the preset, that would’ve been consistent with this wiki page, although the page has since been pared back; a separate page now describes the much rarer man_made=antenna tag and equivocates as to which tag should be used. So now I’m confused, and I can’t imagine the situation would be any better for the user.

Hey @1ec5 - yes, this situation is pretty confusing.   Also remember in #3630 that we changed them from "Radio Mast" to simply "Mast".

I was assuming that Mast/Tower is the structure, and Antenna is the thing that attaches to it?  Many people will only tag the Mast/Tower, as its the most visible thing. 

@obama Can you discuss more your motivation for wanting this preset?

For reference:
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dantenna
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dtower
https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dmast

obama:
i thought it was OK to add, but if it confuses people than leave it out.

PR #4940, opened 2018-03-25
Add fields to Taginfo file

### LINKED ISSUE #3598: Taginfo project file should include icons, names
The [Taginfo project file](https://github.com/openstreetmap/iD/blob/master/data/taginfo.json) should include the icons and display names for each preset, as [JOSM’s](https://taginfo.openstreetmap.org/projects/josm_main_presets#tags) does.

### LINKED ISSUE #4937: Add fields to Taginfo file as well
Currently only preset tags are included in iDs taginfo.json file. It think that it would be a great improvement to the Taginfo documentation if we also add field tags and keys to the file.

mmd-osm:
Gave it a first try.. not quite sure if I covered all combinations in the original json files. I should probably remove those duplicates in `taginfo` as well.

Things like "fields/gender.json" still look pretty strange in the taginfo result. There seem to be some cases, where we don't want a cross product of those string.options values and the keys array.

Supposed to eventually
* fix #4937 
* fix #3598

### DIFF
--- a/build_data.js
+++ b/build_data.js
@@ -388,11 +388,11 @@ function generateTaginfo(presets, fields) {
          return;
        }
 
-       var isNewDescription = currentTaginfoEntries[0].description.split(' ,')
+       var isNewDescription = currentTaginfoEntries[0].description.split(', ')
                            .indexOf(tag.description) === -1;
 
        if (isNewDescription) {
-         currentTaginfoEntries[0].description += ", " + tag.description;
+         currentTaginfoEntries[0].description += ', ' + tag.description;
        }
     }
 
--- a/build_data.js
+++ b/build_data.js
@@ -160,14 +160,6 @@ function generateFields(tstrings) {
             }
         }
 
-        if (field.geometry) {
-          t.geometry = field.geometry;
-        }
-
-        if (field.icon) {
-          t.icon = field.icon;
-        }
-
         fields[id] = field;
     });
     return fields;
@@ -304,12 +296,6 @@ function generateTranslations(fields, presets, tstrings) {
         } else {
             delete preset.terms;
         }
-    });
-
-    // remove fields we're not interested in for translation
-    _forEach(translations.fields, function(field) {
-        delete field.icon;
-        delete field.geometry;
     });
 
     return translations;
--- a/build_data.js
+++ b/build_data.js
@@ -14,7 +14,7 @@ const path = require('path');
 const shell = require('shelljs');
 const YAML = require('js-yaml');
 const colors = require('colors/safe');
-const maki = require ('@mapbox/maki');
+const maki = require('@mapbox/maki');
 
 const fieldSchema = require('./data/presets/schema/field.json');
 const presetSchema = require('./data/presets/schema/preset.json');
@@ -349,8 +349,9 @@ function generateTaginfo(presets, fields) {
             tag.value = preset.tags[last];
         }
 
-        if (preset.name)
+        if (preset.name) {
           tag.description = [ preset.name ];
+        }
 
         if (preset.geometry) {
           setObjectType(tag, preset);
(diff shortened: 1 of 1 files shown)


bhousel:
> Gave it a first try.. not quite sure if I covered all combinations in the original json files. I should probably remove those duplicates in taginfo as well.
> 
> Things like "fields/gender.json" still look pretty strange in the taginfo result. There seem to be some cases, where we don't want a cross product of those string.options values and the keys array.

Thanks @mmd-osm - I took a look and this seems right to me.  I think it's pretty good.

What stands out to me is that we do have a lot of duplicates for things like:
```js
        {
            "key": "ref"
        },
        {
            "key": "ref"
        },
        {
            "key": "ref"
        },
        ...
```

So maybe we should start adding `"description": ""` to each of these (for both presets and fields).. The description could say which presets and fields use the tag, and we could coalesce duplicate key/value pairs together like:

```js
        {
            "key": "ref",
            "description": "Gate Number, Hole Number, Highway Number, Runway Number..."
        },
```

mmd-osm:
I think we're slowly getting there. I added descriptions for presets and fields now (where available) and also coalesced them like in the example above:

```JSON
        {
            "key": "ref",
            "description": "Gate Number, Hole Number, Junction Number, Platform Number, Road Number, Route Number, Runway Number, Stop Number, Taxiway Name, Reference Code"
        },
```

bhousel:
Hey that's pretty good @mmd-osm !

What else is left?   
Are you thinking to add support for `"object_types": ["node", "way", "area", "relation"]` ?

mmd-osm:
> Are you thinking to add support for "object_types

I stil have to figure out how the iD <--> taginfo mapping looks like. point, line and relation seem quite obvious, the others aren't (yet). 

```
          const mapping = { "point"    : "node",
                            "vertex"   : "node",
                            "line"     : "way",
                            "relation" : "relation",
                            "area"     : "area" };
```

Then there's some strange side effect issue when adding geometry to generateField. It ends up in data/presets.yaml as well. 

```
        if (field.geometry) {
          t.geometry = field.geometry;
        }
```

Icons could be interesting as well. Can I assume that maki is used all over the place? Then this would be quite straightforward:

```
        if (preset.icon) {
          tag.icon_url = 'https://raw.githubusercontent.com/mapbox/maki/master/icons/' + preset.icon + '-11.svg';
        }
```

bhousel:
> const mapping = { "point"    : "node",
>                             "vertex"   : "node",
>                             "line"     : "way",
>                             "relation" : "relation",
>                             "area"     : "area" };

Yes, looks good ..

> Then there's some strange side effect issue when adding geometry to generateField. It ends up in data/presets.yaml as well.

Yeah it might not be straightforward which the fields can be used on which geometries.  You could try to build an index of field -> geometry when you build the preset list (presets have geometry). 

> Icons could be interesting as well. Can I assume that maki is used all over the place, which would be quite straightforward:

Maki isn't quite used everywhere, but you can use code like this
https://github.com/openstreetmap/iD/blob/523fb7f370c6d2282fbc86845c408563d11fc9ef/modules/svg/labels.js#L221-L222

Because maki is a dependency, you can just require it and get this array of all the icon names, then use that to build an `isMaki()` function:
```js
> const maki = require ('@mapbox/maki');
undefined
> var dataFeatureIcons = maki.layouts.all.all
[ 'aerialway',
  'airfield',
  'airport',
...
```
hope this helps!

mmd-osm:
Thanks again for all the hints. I have to admit I was bit lazy to set up a separate field -> geometry index, and instead decided to removed geometry and icon towards the end of `generateTranslations `. As it is a bit of a hack, I added a small comment there. Hope that's ok.

bhousel:
> Thanks again for all the hints.

Thank you, for working on it!   Looks great, I just merged it 👍

PR #4949, opened 2018-03-27
Add field for industrial=*

hikemaniac:
This PR adds a field for industrial=*. It is mainly used for landuse=industrial to describe the type of industry. The key has a [wiki page](https://wiki.openstreetmap.org/wiki/Key:industrial) and [over 23k uses](https://taginfo.openstreetmap.org/keys/?key=industrial#overview) at the moment. Is it okay to add this field?

### DIFF
--- a/data/presets/fields.json
+++ b/data/presets/fields.json
@@ -951,7 +951,7 @@
             "type": "check",
             "label": "Indoor"
         },
-        "indoor": {
+        "industrial": {
             "key": "industrial",
             "type": "combo",
             "label": "Type"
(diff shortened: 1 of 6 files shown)


data/presets/fields.json
```diff
@@ -951,6 +951,11 @@
             "type": "check",
             "label": "Indoor"
         },
+        "indoor": {
```
bhousel: I think this is a mistake, should be "industrial"..
(I don't usually edit the `fields.json` file, it's autogenerated when doing a local build.  So even if we did merge it this way, it would get fixed automatically anyway)

bhousel:
Sure, thank you!
I'm surprised we don't have this field already.

hikemaniac:
@bhousel Oh dear, I have no idea how that happened! I just fixed it anyway.

Issue #4961, opened 2018-04-03
add billboard and column advertising preset?


Lukas458:
Hello,
I think it could be a good idea to add presets for advertising=billboard and advertising=column, isn't it? Both are in very common use.
And some further: In the preset for amenity=clock, there is the key faces=* missing.

bhousel:
Thanks for the suggestions!   iD already has a preset for `advertising=billboard`

We can:
- [ ] add preset for [advertising=column](https://wiki.openstreetmap.org/wiki/Tag:advertising%3Dcolumn)
- [ ] add [faces](https://wiki.openstreetmap.org/wiki/Key:faces) field to `amenity=clock`

Xavier-J-Ortiz:
@bhousel I'll be happy to take this issue and get started on this soon.

Lukas458:
Thank you! :)

bhousel:
This was done..

PR #4963, opened 2018-04-04
added column in presets/presets/advertising

Xavier-J-Ortiz:
I copied `billboard.json` and used this as the base for `column.json`.

I removed the `line` field, since this doesn't make sense for a column. Debating `point`, since a column would have a not-ignorable "width" as well, giving it a shape when looked from above. I left `point` in there nonetheless. It could be used for a thin column, like a tube or pole, though I doubt there are many poles used by advertisers.

Refs:https://github.com/openstreetmap/iD/issues/4961

### DIFF
--- a/data/presets/presets/advertising/column.json
+++ b/data/presets/presets/advertising/column.json
@@ -1,0 +1,14 @@
+{
+    "fields": [
+       "direction",
+       "lit"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "advertising": "column"
+    },
+    "name": "Column"
+}
--- a/data/presets/presets/advertising/column.json
+++ b/data/presets/presets/advertising/column.json
@@ -8,7 +8,7 @@
         "area"
     ],
     "tags": {
-        "advertising": "advertising_column"
+        "advertising": "column"
     },
     "name": "Advertising Column"
 }
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -1684,14 +1684,14 @@ en:
         # 'addr:*=*'
         name: Address
         terms: '<translate with synonyms or related terms for ''Address'', separated by commas>'
-      advertising/advertising_column:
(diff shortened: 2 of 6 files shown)


bhousel:
> I removed the line field, since this doesn't make sense for a column. Debating point, since a column would have a not-ignorable "width" as well, giving it a shape when looked from above. I left point in there nonetheless. It could be used for a thin column, like a tube or pole, though I doubt there are many poles used by advertisers.

[osm wiki for `advertising=column`](https://wiki.openstreetmap.org/wiki/Tag:advertising%3Dcolumn) suggests "point" and "area" which makes sense to me. With high enough resolution imagery, someone could draw a circular area where the column is.

"vertex" is really for points that must fall along a line.  It's usually used for presets that we want to suggest along a road (like traffic signals or signs - I guess it sort of makes sense for billboard, but maybe not for column)

data/presets/presets/advertising/column.json
```diff
@@ -0,0 +1,14 @@
...
+       "lit"
+    ],
+    "geometry": [
+        "point",
+        "vertex"
+    ],
+    "tags": {
+        "advertising": "column"
+    },
+    "name": "Column"
```
bhousel: Can you call it an "Advertising Column" - otherwise people will probably misuse the preset for architectural stuff.

Xavier-J-Ortiz:
Thanks for clarifying that.

I removed `vertex`, and added `area`. And renamed the structure from `column` to `advertising_column`.

bhousel:
> And renamed the structure from column to advertising_column.

Oh sorry I wasn't clear..
* The file should still be `advertising/column.json`
* The tag should still be `advertising=column`
* The only thing I wanted you to change was the display name of the feature:
  `"name": "Advertising Column"`

Xavier-J-Ortiz:
Just pushed the changes.

bhousel:
Thanks @Xavier-J-Ortiz - I just merged it!

PR #4672, opened 2018-01-09
Localize numbers, units in scale, info panels

1ec5:
Numbers and units appear unlocalized throughout iD. This PR begins to localize some of them. Here are some before-and-afters:

scale before scale after  
**Scale bar:**

* Depending on the browser language, a period will group the thousands digits and a comma will serve as the decimal separator.
* In Chinese, the unit symbol would be translated as 公里.

search coordinate before search coordinate after  
**Coordinate in search results:**

* Coordinates appear in the familiar degree-minute-second format to clarify that latitudes are displayed before longitudes.

measurement area metric before measurement area metric after  
**Measurement panel**, seen here in metric with a closed area selected:

* “Closed” will appear after “area” or “line” in some locales.
* The English localization abbreviates square feet and square miles as “sq ft” and “sq mi”, respectively, instead of “ft²” and “mi²”.
* Lengths and areas are rounded to four significant figures instead of a fixed decimal place.

location before location after  
**Location panel:**

* Coordinates appear in both degree-minute-second and decimal format, for better precision and easier copy-pasting.

measurement area persian&nbsp;
_Measurement panel partially localized into Persian_

Fixes #3597.

PR #4964, opened 2018-04-04
Pluralize localizable strings

### LINKED ISSUE #597: Better i18n plural support

1ec5:
UI strings that include counts now use the appropriate plural form in English and (once translations are updated) in other languages. When looking up a translation and one of the tokens is numeric, look one level deeper in the translation file for the CLDR plural form corresponding to the numeric token.

singular plural

In order to support this feature for all the languages that iD supports, with all their varying pluralization rules, this change introduces dependencies on:

* [cldr-core](https://github.com/unicode-cldr/cldr-core/), which supplies plural rules for a _number_ of languages (part of [cldr-data](https://github.com/rxaviers/cldr-data-npm/))
* [CLDRPluralRuleParser](https://github.com/santhoshtr/CLDRPluralRuleParser/), which parses a plural rule, determining whether it matches the given number (used by MediaWiki)

This works locally using `npm start`, but I’d like someone to double-check that I’m including these dependencies correctly. (All we need is a single JSON file from cldr-core, not the entire package.)

Keying off numbers in token replacements means any number formatting that occurs before calling `t()` will defeat the pluralization logic therein. Unlike the unit formatting that occurs in `t()`’s callers (as of #4672), number formatting needs to happen inside `t()`.

More to come:

* [x] Depend on cldr-core and CLDRPluralRuleParser
* [x] Look up pluralized translations in `t()`
* [x] Pluralize some strings in core.yaml
* [ ] Move number formatting into `t()`
* [ ] Combine manually pluralized strings like `restriction.controls.via_up_to_one`/`restriction.controls.via_up_to_two`
* [ ] Break up `validations.many_deletions` into strings with one pluralized word each
* [x] Mention plural format in contributing documentation
* [ ] Update changelog

Fixes #597.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -252,6 +252,17 @@ to iD, only display them in the interface through the `t()` function.
 Then, add the new string to `data/core.yaml`. The translation system, Transifex,
 will automatically detect the change.
 
+If you need to insert something into the string programmatically, insert a
+`{token}` into the string in the YAML file. If the token is a number that can
+affect how surrounding words are pluralized, define the singular and plural
+forms separately like this:
+
+```yaml
+    changes:
+      one: "1 Change"
+      other: "{count} Changes"
+```
+
 If you are updating an existing string, update it in `data/core/yaml` and run
 `npm run build` to generate the `en.json` file automatically, then commit both
 modified files.
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -92,7 +92,9 @@ en:
         line: Deleted a line.
         area: Deleted an area.
         relation: Deleted a relation.
-        multiple: "Deleted {n} features."
+        multiple:
+          one: "Deleted a feature."
+          other: "Deleted {n} features."
       too_large:
         single: This feature can't be deleted because not enough of it is currently visible.
         multiple: These features can't be deleted because not enough of them are currently visible.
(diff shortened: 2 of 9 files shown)


bhousel:
> This works locally using npm start, but I’d like someone to double-check that I’m including these dependencies correctly. (All we need is a single JSON file from cldr-core, not the entire package.)

Looks great!  Rollup will include only the single JSON file 👍 

> Keying off numbers in token replacements means any number formatting that occurs before calling t() will defeat the pluralization logic therein. Unlike the unit formatting that occurs in t()’s callers (as of #4672), number formatting needs to happen inside t().

Ok, that makes sense.  I don't know how often in the code we pass stringified numbers to `t()` but it is probably not too much work to change these.

magol:
Any work on this PR?

1ec5:
Sorry, I don’t know when I’ll get a chance to get back to this PR, but it’s certainly something I’d like to get in eventually. (At some point, browser support for `Intl.PluralRules` would improve enough that we could radically simplify this PR.)

quincylvania:
To be clear, this was closed automatically when I renamed the `master` branch, but it was two years old.

I'd like to spend some time soon to get pluralization right as we work on expanding iD's global usability (see #7713).

Issue #4968, opened 2018-04-06
Cannot add restrictions in complex multiway junction


srividyacb:
[Area in OpenStreetMap](https://www.openstreetmap.org/edit?editor=id#map=20/37.76421/-122.43304)

![image](https://user-images.githubusercontent.com/3423533/38406672-c93c01e2-3993-11e8-8d98-9e1e30e002ea.png)
_Red box in above image showing overlapping restriction arrow in TR editor_

In the same junction cannot add `only` turn-restrcition for highway going straight. The restriction arrow is not going blue even after multiple clicks. 

![tr-issue](https://user-images.githubusercontent.com/3423533/38406734-1be76026-3994-11e8-85ec-4a61c8f2dc3a.gif)

cc @bhousel

tyrasd:
confirming. in the second case, the following exception is thrown:

```
graph.js:48 Uncaught Error: entity r-1 not found
    at coreGraph$$1.entity (graph.js:48)
    at action (delete_relation.js:18)
    at Array.<anonymous> (unrestrict_turn.js:11)
    at _act (history.js:59)
    at _perform (history.js:74)
    at Object.perform (history.js:168)
    at Object.perform (context.js:449)
    at SVGSVGElement.click (restrictions.js:405)
    at SVGSVGElement.<anonymous> (on.js:27)
```

tyrasd:
> adding a no-u-turn restriction at either end of the selected from way causes the no-u-turn icons to be activated at both ends of the way (although only one turn restriction relation is created)

this is actually a regression introduced by 0d262e5bb5572e1ad740a6a97777c54fb0e1da01 for #4928. :thinking:

tyrasd:
no, unfortunately the fix from #4928 doesn't work in this situation (I've originally tested this with a locally checked out master branch, but [http://preview.ideditor.com/master](http://preview.ideditor.com/master/#background=Bing&disable_features=boundaries&id=n581047109&map=20.95/37.76422/-122.43307) shows the same behavior), and introduced the above mentioned "no-u-turn reflection" regression.

bhousel:
> no, unfortunately the fix from #4928 doesn't work in this situation (I've originally tested this with a locally checked out master branch, but http://preview.ideditor.com/master shows the same behavior), and introduced the above mentioned "no-u-turn reflection" regression.

Ah good, thanks for checking 👍

Issue #4928, opened 2018-03-22
Can only Toggle no_straight_on Turn Restriction Once


zlavergne:
I ran into an issue where I can only toggle a turn restriction once in the new Restriction editor. I discovered the bug because I accidentally hit the `no_straight` but I couldn't click through to reset it. 

![turn_restriction](https://user-images.githubusercontent.com/9849118/37787983-e5196082-2dc5-11e8-96d3-4a6a6b54e686.gif)

I cannot undo the `no_straight` on any of the ways I could add it to. 

[Here's the way we're looking at.](https://www.openstreetmap.org/way/28598125#map=19/-34.34638/-70.84229)

bhousel:
Hey @zlavergne which Browser/OS did you see this on?
I can't seem to reproduce on MacOS with Chrome 65.0.3325.162

zlavergne:
I'm on MacOS and Chrome 65.0.3325.181 
I'm getting this error from the console: 
```
id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:119 Uncaught Error: entity r-1 not found
    at coreGraph$$1.entity (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:119)
    at id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:21
    at Array.<anonymous> (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:22)
    at n (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:20)
    at t (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:20)
    at Object.perform (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:20)
    at Object.perform (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:21)
    at SVGSVGElement.a (id-18ccfbbf4674e7fcefc636d54df5ee8f3363dadd80070ea544340f99000d0bb7.js:18)
[...]

PR #4969, opened 2018-04-06
only drop restrictions from FROM towards VIA when adding new "only" restriction

tyrasd:
partially addresses bugs in #4968, see also #4928

* iD doesn't crash anymore in complex situations (short FROM ways where both ends connect to a TO way)
* adding a only-restriction at one end of a short FROM doesn't delete restrictions on the other end of the same FROM anymore
* fixes regression where adding a no-u-turn restriction at either end of the a FROM way causes no-u-turn icons to be shown at both ends of the way (by reverting 0d262e5)


modules/osm/intersection.js
```diff
@@ -410,18 +410,8 @@ export function osmIntersection(graph, startVertexId, maxDistance) {
...
-                                    if (currPath[k][0] === 'w') pathVias.push(currPath[k]);
-                                }
-                                var restrictionVias = [];
-                                for (k = 0; k < v.length; k++) {
-                                    if (v[k].type === 'way') restrictionVias.push(v[k].id);
-                                }
-                                var diff = _difference(pathVias, restrictionVias);
-                                matchesViaTo = !diff.length;
+                            } else if (_every(v, function(via) { return currPath.indexOf(via.id) !== -1; })) {
+                                matchesViaTo = true;    // match all VIA ways
```
bhousel: I have to check this part.. I think the reason I changed this was because `v` can be a via node that happens to match somewhere else along the path.   (adding `&& via.type ==='way'` might be ok)

So the stricter test is that we want to match via ways and that they should all exactly match the via ways that the current path follows.

tyrasd: ok, I see the purpose now. but it had a bug for the following case: `currPath = ["w?", "n1"]` and `v = [{type:"node", id: "n2"}]`.

I've replaced the check with a more compact implementation in ae9e9fbdd.

modules/osm/intersection.js
```diff
@@ -410,18 +410,9 @@ export function osmIntersection(graph, startVertexId, maxDistance) {
...
-                                    if (currPath[k][0] === 'w') pathVias.push(currPath[k]);
-                                }
-                                var restrictionVias = [];
-                                for (k = 0; k < v.length; k++) {
-                                    if (v[k].type === 'way') restrictionVias.push(v[k].id);
-                                }
-                                var diff = _difference(pathVias, restrictionVias);
-                                matchesViaTo = !diff.length;
+                            } else if (v.length === currPath.length / 2 - 1 &&
+                                _every(v, function(via, index) { return via.type === 'way' && currPath[(index+1)*2] === via.id; })) {
```
bhousel: I still don't completely understand this code...
It seems like the `_every` version of the code requires that the order of the multiple via ways must match the order in the path according to their index.  However the vias in `v` will not necessarily be ordered correctly.  They'll be ordered however they are stored in the Relation member array.

That's probably why I switched to gathering the restriction vias and the from vias into two arrays and using `_difference` to see if they have the same elements.

modules/ui/fields/restrictions.js
```diff
@@ -359,13 +359,13 @@ export function uiFieldRestrictions(field, context) {
...
 
-                    // Adding an ONLY restriction should destroy all other direct restrictions from the FROM.
+                    // Adding an ONLY restriction should destroy all other direct restrictions from the FROM towards the VIA.
                     // We will remember them in _oldTurns, and restore them if the user clicks again.
                     turns = _intersection.turns(_fromWayID, 2);
                     extraActions = [];
                     _oldTurns = [];
                     for (i = 0; i < turns.length; i++) {
-                        if (turns[i].direct) {
+                        if (turns[i].direct && turns[i].path[1] === datum.path[1]) {
```
bhousel: For what it's worth, this part of the code is what we need to fix the bug and looks great 😄

tyrasd:
> duplicate all of these tests, but on a graph with loops

yeah, I think so as well

modules/osm/intersection.js
tyrasd: > They'll be ordered however they are stored in the Relation member array.

oh. I thought the vias should be ordered in the osm relations… but I see that the [osm-wiki](https://wiki.openstreetmap.org/wiki/Relation:restriction) apparently doesn't specify the order… :thinking: could there be cases where in a really really complex intersection the order of traversed via-ways is also important?

bhousel: > could there be cases where in a really really complex intersection the order of traversed via-ways is also important?

It doesn't really have to be very complicated - but if there are 2 via ways, I think the algorithm needs to match the restriction regardless of what order the vias appear in the relation.
 
screenshot 2018-04-10 09 44 48

bhousel: > does d2d464d look better?

I *think* so.. 🤔   The only situations I can think of that would break it would be invalid relations (like multiple via nodes, or mixed via node and way) and we dont need to worry about that..

bhousel: > oh. I thought the vias should be ordered in the osm relations… but I see that the osm-wiki apparently doesn't specify the order…

btw, I just thought more about this - another good reason for not having the via order matter: If you split a via way, the split code would need to intelligently reorder the vias (super complicated - see the issues we're having with PTv2 routes), rather than how it currently works of just making sure both split survivors end up as members.

bhousel:
I just pushed a failing test for the situation which was worrying me.. This is also #4928 - which is the code that causes the relation to be recognized twice (and then attempting to delete it twice causes the crash).

bhousel:
Whew 😅 this is working now..
I tested a bunch with the intersections from #4968 and #4928.
It feels really good to have added test cases for these .. Thanks @tyrasd for your help with this!

Issue #4986, opened 2018-04-13
attraction=maze preset


abdeldjalil09:
https://wiki.openstreetmap.org/wiki/Proposed_features/Key:attraction

PR #4987, opened 2018-04-13
Added attraction maze

sulfo:
Ref: #4986


bhousel:
Looks great, thanks @sulfo 👍

Issue #4951, opened 2018-03-28
Deleting a restriction creating a new restriction in the TR window


abhisheksaikia:
While trying to delete a wrongly added restriction (no_u_turn) in iD, through the TR window, ended up creating a new restriction (only_u_turn) in addition to deleting the old one without any visible warnings.

![trbug4](https://user-images.githubusercontent.com/13744156/38026888-72fa0c32-32ab-11e8-9345-d363cd22a13e.gif)

osm Change file:

<details> 

```
<?xml version="1.0" encoding="UTF-8"?>
<osmChange version="0.6" generator="iD">
   <create>
      <relation id="-8" version="0">
         <member type="way" role="from" ref="417310791" />
         <member type="way" role="via" ref="573973104" />
         <member type="way" role="to" ref="573973107" />
         <tag k="type" v="restriction" />
         <tag k="restriction" v="only_u_turn" />
      </relation>
   </create>
   <modify />
   <delete if-unused="true">
      <relation id="8155829" version="1">
         <member type="way" role="from" ref="417310791" />
         <member type="way" role="via" ref="573973104" />
         <member type="way" role="to" ref="573973107" />
         <tag k="restriction" v="no_u_turn" />
         <tag k="type" v="restriction" />
      </relation>
   </delete>
</osmChange>
```
</details>

srividyacb:
@bhousel 

[Area in OpenStreetMap](https://www.openstreetmap.org/edit?editor=id&relation=7517143#map=20/37.79673/-122.42363)

- click the [way](https://www.openstreetmap.org/way/527829981) in iD 
- Delete `no_left` and `no_u` turn-restriction
- While deleting `no_left` - first click it changes to `only_left` (blue) and second click deletes the TR.
- While deleting `no_u` - the first click (in single action) both `only_u` and deletion is happening (no visible change in colors). 

(_in gif you can notice, deletion of no_u doesn't change to blue but directly deletes but doesn't give indication of TR creation_)


Follow the steps as per to reproduce the problem

![new-relation](https://user-images.githubusercontent.com/3423533/38091353-f3603978-3382-11e8-98f1-77445c4a9f73.gif)

bhousel:
Awesome thanks @srividyacb and @abhisheksaikia for the detailed report 👍

PR #5003, opened 2018-04-22
CHANGELOG: minor fixes

waldyrious:
- add missing link
- format list
- clarify an entry

Note: the change on the description of a buggy behavior was based on from reading  #4951 and 4041ab48f, but I might have misread something. Please confirm that it is correct.


bhousel:
Thanks for catching these!  👍

PR #5020, opened 2018-05-05
Added dance style field to dance hall preset

hikemaniac:
This PR adds the dance style field to the dance hall preset. 
Most dance halls have a certain range of dances that are usually exercised there. So I think it would make sense to also add the field to the dance hall preset and not only the dance school preset.

Because of the three rebase commits, that are also included in this PR, could you please do a squash merge on GitHub when merging this pull request?


bhousel:
Looks good, thanks @hikemaniac 👍

PR #5025, opened 2018-05-07
Fixing a bug in displaying read-only tags for entities

guyarad:
The normal flow of current *iD editor* code does not display read-only tags for entities. However, for my company's needs we needed to have that ability. Once we created read-only tags list (more accurately, a RegExp list), we got the exception shown below.

The root-cause is wrong invocation of `isReadOnly` from `bindTypehead`.
The resolution is rather simple and demonstrated below...

exception screenshot

### DIFF
--- a/modules/ui/raw_tag_editor.js
+++ b/modules/ui/raw_tag_editor.js
@@ -207,7 +207,8 @@ export function uiRawTagEditor(context) {
 
 
         function bindTypeahead(key, value) {
-            if (isReadOnly({ key: key })) return;
+            if (isReadOnly(key.datum())) return;
+
             var geometry = context.geometry(_entityID);
 
             key.call(d3_combobox()


modules/ui/raw_tag_editor.js
```diff
@@ -207,7 +207,8 @@ export function uiRawTagEditor(context) {
 
 
         function bindTypeahead(key, value) {
-            if (isReadOnly({ key: key })) return;
+            if (isReadOnly(key.datum())) return;
```
guyarad: are there any tests we can add for future modifications?

guyarad: Unfortunately I'm not familiar enough with JS nor with D3 so I won't be helping here. If and when you guys add some testing I will be glad to assist adding tests for that specifically.

PR #5026, opened 2018-05-07
add preset for grandstands

tyrasd:
see https://wiki.openstreetmap.org/wiki/Tag:building%3Dgrandstand


bhousel:
Thanks @tyrasd 👍

Issue #4327, opened 2017-09-08
Allow some kind of "over-zoom" on ESRI World Imagery


mboeringa:
While a nice addition to iD and other editors, the ESRI World Imagery seems not to be available in some of the highest zoom levels possible with iD. If you zoom in to far in a certain region, you end up with grey tiles with an error text "Map data not yet available" visible in the editor.

This somewhat limits the usefulness of this imagery layer, compared to e.g. Bing, where I have never witnessed this behaviour.

Actually, it seems to depend on the specific spot as to what zoomlevel you still get imagery, e.g. I have seen the imagery being replaced by the warning at Z19 or 20 in different locations.

I have no idea if it is possible to detect this based on the imagery service, but it would be kind of nice if iD allowed for some form of "over-zoom" on the ESRI layer, never showing the warning, but instead showing enlarged / interpolated (but blurred) tiles of the last available zoom levels. 

Of course, the quality will suffer, but at least you would still have some guidance when zooming and trying to make some edit.

![afbeelding](https://user-images.githubusercontent.com/7635726/30225378-8f9b56f6-94d2-11e7-9456-63c221f25fa9.png)

bhousel:
I don't *think* there is much we can do with this, as the server returns a valid response for those zooms, and iD can't really tell whether the tile has imagery in it or not.  If we received an error response, like a 404 or something, iD would lookup to a lower zoom and overzoom it.

bhousel:
cc @jgravois for ideas - we might be able to use their metadata service (see #4280) to know which zooms will return imagery, but that's a bunch of extra lookups I'd rather avoid.

jgravois:
thanks for pointing out the issue @mboeringa!

We could have locked the Esri Imagery at zoom level 19 and overzoomed everything beyond that (like the DigitialGlobe and Mapbox layers) but I made a concious decision to allow requests all the way to `22` because we do serve higher resolution stuff in *some* areas.

The downside of the current implementation is, as you noted, that imagery is lost entirely as soon as the highest native zoom is exceeded. it would be a substantial improvement to overzoom *selectively*.

We have a couple options here:

1. from @bhousel's reply, it sounds like tacking on [`?blankTile=false`](https://server.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/tile/22/1674571/731704?blankTile=false) *might* be sufficient all on its own to leverage logic already present in iD <sup>1</sup>

[...]

PR #5029, opened 2018-05-07
use a tilemap for esri imagery instead of tacking on blankTile=false

jgravois:
greetings! :wave:

currently OSM editors are requesting roughly 35 million Esri imagery tiles a month 🎉 and making roughly the same number of requests for tiles that don't exist 🚫.

this PR still needs work, but it introduces support for the 'tilemap' option i described in https://github.com/openstreetmap/iD/issues/4327#issuecomment-328198775

the gist of it is that we can take advantage of a service that indicates whether or not tiles exist in a particular area of interest.

by asking whether tiles are present at zoom level 20 in the current extent of the map as soon as the edit session is instantiated, we'd forgo a _lot_ of unnecessary requests for tiles.

### DIFF
--- a/modules/renderer/background_source.js
+++ b/modules/renderer/background_source.js
@@ -296,6 +296,42 @@ rendererBackgroundSource.Esri = function(data) {
         cache = {},
         inflight = {};
 
+    // use a tilemap service to set maximum zoom for esri tiles dynamically
+    // https://developers.arcgis.com/documentation/tiled-elevation-service/
+    esri.fetchTilemap = function(center) {
+        // tiles are available globally to zoom level 19, afterward they may or may not be present
+        var z = 20;
+
+        // first generate a random url using the template
+        var dummyUrl = esri.url([1,2,3]);
+
+        // calculate url z/y/x from the lat/long of the center of the map
+        var x = (Math.floor((center[0] + 180) / 360 * Math.pow(2, z)));
+        var y = (Math.floor((1 - Math.log(Math.tan(center[1] * Math.PI / 180) + 1 / Math.cos(center[1] * Math.PI / 180)) / Math.PI) / 2 * Math.pow(2, z)));
+
+        // fetch an 8x8 grid because responses to leverage cache
+        var tilemapUrl = dummyUrl.replace(/tile\/[0-9]+\/[0-9]+\/[0-9]+/, 'tilemap') + '/' + z + '/' + y + ' /' + '/8/8';
+
+        // make the request and introspect the response from the tilemap server
+        d3_json(tilemapUrl, function (err, tilemap) {
+            if (err || !tilemap) return;
+
+            var tiles = true;
+            for (var i=0; i<tilemap.data.length; i++) {
+                // 0 means an individual tile in the grid doesn't exist
+                if (!tilemap.data[i]) {
+                    tiles = false;
+                    break;
+                }
+            }
+
+            // if any tiles are missing at level 20 we restrict maxZoom to 19
+            if (!tiles) {
+                esri.scaleExtent[1] = 19;
+            }
+        });
+    };
+
     esri.getMetadata = function(center, tileCoord, callback) {
         var tileId = tileCoord.slice(0, 3).join('/'),
             zoom = Math.min(tileCoord[2], esri.scaleExtent[1]),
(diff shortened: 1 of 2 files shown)


modules/renderer/background_source.js
```diff
@@ -409,6 +405,37 @@ rendererBackgroundSource.Esri = function(data) {
...
+            // calculate url z/y/x from the lat/long of the center of the map
+            const urlX = (Math.floor((center[0] + 180) / 360 * Math.pow(2, urlZ)));
+            const urlY = (Math.floor((1 - Math.log(Math.tan(center[1] * Math.PI / 180) + 1 / Math.cos(center[1] * Math.PI / 180)) / Math.PI) / 2 * Math.pow(2, urlZ)));
+
+            // we fetch an 8x8 grid because they cover a normal extent and responses are cached
+            const tilemapUrl = tileCoord[3].replace(/tile\/[0-9]+\/[0-9]+\/[0-9]+/, 'tilemap') + `/${urlZ}/${urlY}/${urlX}/8/8`;
+
+            // make the request and introspect the response from the tilemap server
+            fetch(tilemapUrl)
+            .then(response => response.json())
```
mmd-osm: a bit of error handling might be good here... (url unavailable, cannot parse result, etc.)

bhousel: The bigger issue here is that we'll have to polyfill fetch / promises for IE11..
(But we need to do this anyway for D3 v5 / #4929)

jgravois: i don't have my heart set on using `fetch`, i just didn't know what you were using for AJAX requests to a CORS enabled web server because the only other import in the file previous was for making JSONP requests.

i've pushed up another commit which uses `d3_json` and aborts if an error is encountered. at this point i just need help triggering the function when editing is initially activated instead of waiting until the metadata context menu is opened.

bhousel:
Cool! Thanks @jgravois  - what is missing from this now?  It probably would work if you switch the `const`/`let` back to old-school `var`.. (again strict ES5 for IE11/Phantom 🙄)

modules/renderer/background_source.js
```diff
@@ -312,6 +307,9 @@ rendererBackgroundSource.Esri = function(data) {
 
         if (inflight[tileId]) return;
 
+        // instead of calling fetchTilemap when the metadata window is open, we should call it when editing is first activated
+        fetchTilemap(center, esri);
```
jgravois: > switch the `const/let` back to old-school `var`

👍 

> what is missing from this now?

right now `fetchTilemap()` is only called when the background metadata window is opened w/ <kbd>⌘ + Shift + B</kbd>.

i'm not sure how/where, but the method needs to be called as soon as users cross the zoom threshold to put `iD` into 'edit mode' (when Esri imagery is being used in the background anyway).

jgravois:
- [x] ~~`const`~~ > `var`
- [x] found existing hook trigger requests to check the esri tilemap.

this is ready for review.

bhousel:
Thanks @jgravois!  I tried this out and ran into a few issues:

1. It looks like the tilemap url isn't put together correctly:

screenshot 2018-06-15 09 21 11

I also get a eslint warning around [this code](https://github.com/openstreetmap/iD/pull/5029/files?utf8=%E2%9C%93&diff=unified#diff-5029207b7599da40f52ba1ddec797e50R309)
`  309:13  warning  'x' is assigned a value but never used  no-unused-vars`

Would it make more sense to calculate `z/x/y` first and then call `esri.url`, and then append a /tilemap?  Rather than calling `esri.url` first, then calculating `x/y` and using regex to replace parts of the url?  (just guessing)


2. I think `esri.fetchTilemap` might need to use `jsonpRequest` instead of `d3_json` (similar to how the `esri.getMetadata` function works).  
I'm getting blocked by CORS.

screenshot 2018-06-15 09 27 58

bhousel:
Looks better now @jgravois  - should there be code that puts the max zoom back to a higher number if the tileMap supports it?  AFAICT it will just stay at [0,19] forever if it gets a negative result from the tilemap query..

jgravois:
my thought was that once a user is zoomed in tight enough to trigger a tilemap request and encounters missing tiles it'd be appropriate to downsample from there on out.

that said, I didn't really consider the scenario where they zoom back out beyond 16 and move somewhere entirely different.

I'm AFK for a couple days but happy to take another look when I get back.

bhousel:
> my thought was that once a user is zoomed in tight enough to trigger a tilemap request and encounters missing tiles it'd be appropriate to downsample from there on out.
> 
> that said, I didn't really consider the scenario where they zoom back out beyond 16 and move somewhere entirely different.

Hmmm - I tested this by accident today by opening the iD walkthrough, which uses the Esri Clarity layer around Three Rivers MI, and it put the maxZoom to 19.  

I'll see if I can find a way to have iD check the tilemap sometimes.

jgravois:
awesome @bhousel. thank you :clap:!

PR #5031, opened 2018-05-10
Create Bleachers

rivermont:
[OSM Wiki page](https://wiki.openstreetmap.org/wiki/Tag:leisure=bleachers)

### DIFF
--- a/data/presets/presets/leisure/bleachers.json
+++ b/data/presets/presets/leisure/bleachers.json
@@ -1,0 +1,17 @@
+{
+    "icon": "picnic-site",
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "leisure": "bleachers"
+    },
+    "terms": [
+        "stand",
+        "stands",
+        "bench",
+        "benches",
+        "bleachers"
+    ],
+    "name": "Bleachers"
+}


bhousel:
hmm, I looked at https://wiki.openstreetmap.org/wiki/Tag:building%3Dgrandstand and it seems that `leisure=bleachers` is now discouraged?  I think the existing `building=grandstand` preset is enough, right?

rivermont:
My understanding from reading the wikis is that Grandstands are larger buildings (permanent) with seating, whereas Bleachers are like [these](https://www.openstreetmap.org/edit#map=21/35.92705/-79.10566), small versions of Grandstands that could be portable and have no roofs.

bhousel:
> My understanding from reading the wikis (and informal knowledge) is that Grandstands are larger buildings (permanent) with seating, whereas Bleachers are like these, small versions of Grandstands that could be portable and have no roofs.

Yes, that's what I thought too, but someone added this to the `building=grandstand` page:

screenshot 2018-05-13 13 02 58

1ec5:
There’s more discussion on [this talk page](https://wiki.openstreetmap.org/wiki/Talk:Tag:leisure%3Dbleachers). Note that the distinction between bleachers and grandstands can be a bit blurred: at least in the U.S., a soccer field or tennis court might have small, movable bleachers, while an American football field might have massive (still uncovered) stands. Those stands can be as large as the grandstand at a horse racing track, which would be covered.

bhousel:
I'm just going to merge this anyway.  There's a wiki page and over 1600 uses, and plenty of people can tell the difference between a grandstand and a bleachers.  Maybe eventually someone will take the scary warnings off the wiki pages.

PR #5054, opened 2018-05-27
Add preset for motorcycle_repair

demonshreder:
The tag already exists - https://wiki.openstreetmap.org/wiki/Tag:shop%3Dmotorcycle_repair

This could help people tag the POI right.


bhousel:
Thanks @demonshreder 👍

PR #5068, opened 2018-06-07
Render pier like railway platform

### LINKED ISSUE #5061: Support rendering of man_made=pier
Currently piers are just displayed as lines, but should have their own style.

This is also the same for areas :

JamesKingdom:
This renders piers.
Now they look like this: 
Closes #5061


bhousel:
Cool thanks @JamesKingdom looks good 👍

Issue #5024, opened 2018-05-07
Smart handling of `key=value` when pasting tags


tordans:
This is a feature request.
The goals is to make it easier to paste tags from the wiki.
The iD editor could behave smarter in this situation with a very nice UX touch.

### Workflow

1. I tag something, eg `leisure=garden` from the search bar
2. I click [on the wiki-link](https://wiki.openstreetmap.org/wiki/DE:Tag:leisure%3Dgarden) to learn how to tag details that are not part of the presets
3. [I learn about](https://wiki.openstreetmap.org/wiki/Key:garden:type) `garden:type` and copy the tag `garden:type=castle` from the wiki page
4. I go back to iD editor and paste `garden:type=castle` into a new tag on "all tags".

**Current behavior:**

The copied text shows up as key. I have to manually split it and remove the "=".

**Expected / nicer behavior:**

1. iD Editor recognizes that I copied a key-value-pair and splits the pair at the "="-sign into key and value.
2. It then add a new blank tag pair below and thereby completes the tagging and the (i) and such show up next to my tag

bhousel:
I like this suggestion 👍

AndreasHae:
I'm going to work on this

tordans:
A quick thank-you-feedback @AndreasHae: I used it the other day and it was awesome ❤️. Thanks for adding this.

PR #5070, opened 2018-06-09
Handle key-value pairs pasted as tags

AndreasHae:
Implements the behaviour requested in #5024

When a user pastes a string like 'key=value' into the raw tag editor, it gets split into key and value.

### DIFF
--- a/modules/ui/raw_tag_editor.js
+++ b/modules/ui/raw_tag_editor.js
@@ -1,4 +1,5 @@
 import _map from 'lodash-es/map';
+import _includes from 'lodash-es/includes';
 
 import { ascending as d3_ascending } from 'd3-array';
 import { dispatch as d3_dispatch } from 'd3-dispatch';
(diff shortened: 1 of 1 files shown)


tordans:
@AndreasHae thanks for this!
I wonder …
- will this also "submit" the tag or just pre-fill the fields. In later case, the user still has to press enter? Advantage auf auto-submitting is, that the UI shows more details like the (i) icon next to the key-value-pair.
- where will the cursor/focus be after splitting? Eg at the end of the value-field?
- what happens if the copied text has new linkes in it?, eg
> key1=value1
> key2=value2

(I don't have a dev-environment to test this myself, so I need to ask this way…)

AndreasHae:
@tordans 
The tag gets split as soon as the input field loses focus. I am not quite sure what you mean by the other two points, but hopefully this GIF explains the functionality better:
id_key_value_split

bhousel:
Thanks @AndreasHae!  I just tried it out and it works great.  Hope people find it useful 👍

Issue #3742, opened 2017-01-11
Vector data


bhousel:
We will need to support vector tiled data for #3152 anyway, but there are other vector sources (e.g. QA layers) that would be great to bring into iD.

pnorman:
> We will need to support vector tiled data for #3152 anyway, but there are other vector sources (e.g. QA layers) that would be great to bring into iD.

There's lots of types of vector layers, what are we looking at supporting?

bhousel:
> There's lots of types of vector layers, what are we looking at supporting?

Not sure yet.  It might be worth having something like (or built into) `editor-layer-index` but for vector layers that are useful for editing.  This ticket is just to kick around ideas, and also to track that we need something like [vector-tile-js](https://github.com/mapbox/vector-tile-js).

slibby:
#4164 provides a starting point for bringing in vector QA layers from ArcGIS-based services, if that is of interest. Not vector tiles, but along the same lines and I think the implementation (in the Map Data expando-tab) could be similar. #3452 also mentions this and might be one way that #4164 could be helpful to start (read-only QA layers from ArcGIS sources).

bhousel:
CC @kriscarle for visibility - I proposed this as a GSoC project, and would be great to get this support into iD sometime soon.

mmd-osm:
Just cross linking: https://www.openstreetmap.org/user/Princi%20Vershwal/diary/43802

vershwal:
The first step towards supporting vector tiled data is that a user can add a URL and have the vector tiles show up on the screen. It will look something like this :

<img width="992" alt="screen_shot_2018-05-23_at_6 34 08_pm" src="https://user-images.githubusercontent.com/17474532/41250877-d7ccd058-6dd5-11e8-80b8-e107400775a7.png">


For this, I have added another layer 'mvt'. Currently, it will support .pbf(Protocolbuffer Binary Format) and .mvt(Mapbox Vector Tiles) format. These files contain data in Protocolbuffer binary format which is converted to geojson using mapbox's [vector-tile-js](https://github.com/mapbox/vector-tile-js) and this data is passed directly to the D3 draw functions to show up the tiles on the screen.
All the code related to this will be [here](https://github.com/openstreetmap/iD/pull/5072).

vershwal:
For some details: https://www.openstreetmap.org/user/Princi%20Vershwal/diary/44151

bhousel:
This happened 🎉

[...]

PR #5072, opened 2018-06-11
 Vector tile drawing for .mvt and .pbf format

vershwal:
Related #3742.


mmd-osm:
I don't know if this is a good place to report some issues. Testing a few mvt + geojson example file in https://github.com/mapbox/mapnik-vector-tile/tree/master/examples/data produced some mixed results:

14_8716_8015.vector.mvt: takes several minutes to load, feels quite slow, browser becomes somewhat unresponsive.
14_2620_6331-barrier_line.geojson: returns the error message on javascript console, otherwise no feedback to user.

```
VM26:1 Uncaught SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
    at parseSaveAndZoom (mvt.js:180)
    at FileReader.<anonymous> (mvt.js:236)
```
This makes me wonder a bit, if those files would be ok for testing.

mmd-osm:
> Do you think this layer should also draw geojson along with mvt?

Not sure about this one, maybe you should discuss this with your mentors. In any case, if a user uploads some unknown format or broken file, some kind of error message would be incredibly helpful to find out what's going on.

> if the code works with some small tiles then we can start working on the performance part next

Ok, sounds fine to me.

vershwal:
> if a user uploads some unknown format or broken file, some kind of error message would be incredibly helpful to find out what's going on.

Ya, this sounds good! Will do that. Thanks.

bhousel:
Thanks @vershwal this is a great start!

Quick update on this - I changed a few things below and merged the code:
- I replaced `d3_fetch` (a newer API which isn't supported on IE11)  with `d3_request` (traditional XHR)
- There are still some todos in the code
- I commented out the item from the map data panel.

Next action for me:
- We really should just combine the "gpx" and "mvt" layers into a single module which handles all overlay data.  They each currently register a drag-and-drop handler and these handlers are conflicting with each other.  Also they have misleading names.  And it's just redundant and confusing to have two things.  So I'm going to combine them 😅 

Next action for @vershwal : 
So the code was merged but is commented out and not ready yet. I didn't want your PR to diverge too much from `master`.  The `vershwal:check` branch was not really current when you started the work, so that's why the merge above did some weird things to the PR.  Going forward I'd like to work more disciplined like this:
- set main iD repo as your origin remote (you are a collaborator so this is ok)
`git remote set-url origin git@github.com:openstreetmap/iD.git`
- pull from `master` frequently, but avoid developing on that branch or pushing to it
`git checkout master && git pull origin master`
- create feature branches off of `master` before working on something
`git checkout -b new_thing`
- try to check in a little bit of work every day, and push to your feature branch.
`git push origin new_thing`

We can chat on Slack more if you have questions!

PR #5073, opened 2018-06-12
Update en.json: +Belgium maps names

tbertels:


bhousel:
Hey @tbertels - these properties are autogenerated from the [editor-layer-index project](https://github.com/osmlab/editor-layer-index).
If you want these names to be available for translation, the way to make the change is to find the source file in editor-layer-index, then add a property to the file like `"i18n": true`

(For example [see the Bing source file](https://github.com/osmlab/editor-layer-index/blob/e21aeafbc08296237e3811940158cc991267a35f/sources/world/Bing.geojson?short_path=9f1c427), which has the `"i18n": true` property set.)

tbertels:
Will do, thanks!

Issue #5105, opened 2018-06-22
Remove user mocks in several /services specs


thomas-hervey:
As noted by @bhousel, several of the test specs for /services (e.g., mapillary, openstreetcam) have [user mocks](https://github.com/openstreetmap/iD/blob/2786974210f43e4ef62f0b23c4906f72caba4867/test/spec/services/openstreetcam.js#L3-L5) that are no longer required. They should be removed since they don't serve a purpose.

PR #5106, opened 2018-06-22
removed: user mocks from image service test specs

thomas-hervey:
This should clean up #5105, however it seems that there is a localization file that was updated on build. This may need to be removed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ land.html
 /css/img
 /test/css
 /test/img
+
+/.vscode/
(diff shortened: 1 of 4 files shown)


bhousel:
Thanks @thomas-hervey this looks good!  I can rebuild the localization file if needed..

thomas-hervey:
@bhousel if you could rebuild the localization that would be good. After attempting to rebuild, I am not certain why anything was new.

bhousel:
I ended up not needing to rebuild it..
This just happened because we had different versions of the `editor_layer_index` dependency.  This is possible because it's not a proper dependency with a version number, so npm won't necessarily update it when you do an `npm install`

PR #5107, opened 2018-06-23
Notes

thomas-hervey:
Added initial display functionality of osm notes. There are several small bugs to fix including repeatedly appending svg tags within notes. Next, I will focus on hover triggers that will populate the sidebar with details related to a hovered note.


modules/services/notes.js
```diff
@@ -0,0 +1,290 @@
...
+import {
+    osmNote,
+} from '../osm';
+
+import {
+    utilRebind,
+    utilIdleWorker
+} from '../util';
+
+var urlroot = 'https://api.openstreetmap.org',
```
mmd-osm: This shouldn't really be a hardcoded URL endpoint, right?

thomas-hervey: thanks @mmd-osm !
It is hard coded in the other services including [osm.js](https://github.com/openstreetmap/iD/blob/7716532061d4e359621201ebf9b6aff709833f1a/modules/services/osm.js#L29). The remainder of the url includes the [notes portion](https://github.com/openstreetmap/iD/blob/1878397387b15c8c822c105fcfc8c74a37ab6185/modules/services/notes.js#L273) and then a [bounding box](https://github.com/openstreetmap/iD/blob/1878397387b15c8c822c105fcfc8c74a37ab6185/modules/services/notes.js#L225) once it is determined. What would you suggest?

mmd-osm: I was just wondering if you could run the Notes feature against a local database only (which is convenient for testing). As I have the osm.js part running in such a way, I'm assuming that there's some override mechanism in place. Looking very briefly at your implementation I couldn't find something similar for notes. Could be I simply missed it.

thomas-hervey: @mmd-osm no I think you're right. There isn't anything in place to point to a local server. I will address this and check the incoming options to see if a different path is provided. Thoughts @bhousel ?

bhousel: Hmm yeah @mmd-osm and @thomas-hervey - it might make sense to refactor this notes service into `services/osm.js`, which already is setup to switch between dev and live servers.

For now, let's just test it how it is..

bhousel:
This is awesome 👏 
Nice work @thomas-hervey, love where this is headed.

screenshot 2018-06-29 21 59 08

mmd-osm:
I bit of feedback from local testing: sometimes, a notes has quite a number of comments and you need to scroll through the list on the left hand side to see all details. Currently, I don't see a way to focus a specific note and easily browse through the list. Am I missing something here and this is by chance already implemented?

peek 2018-06-30 09-38

bhousel:
> I bit of feedback from local testing: sometimes, a note has quite a number of comments and you want to scroll through the list on the left hand side to see all details. Currently, I don't see a way to focus a specific note and easily browse through that list.
> Am I missing something here and this is by chance already implemented?

Yep, this isn't implemented yet - coming soon!

thomas-hervey:
Great work @bhousel. These changes are really helpful. On occasion, it seems that 'loading' isn't getting called still because the spinner still remains. I'll dig into this. @mmd-osm, click handling for notes is in the works right now.

bhousel:
> On occasion, it seems that 'loading' isn't getting called still because the spinner still remains. I'll dig into this.

I've noticed this too. My guess is that it happens when a user scrolls the map to a new location (which dispatches a `loading` event to start the spinner), then they scroll back to a part of the map where we've already loaded the data.  This will abort the inflight requests, then `tiles.forEach` will do nothing because the viewed tiles are already cached (so the `loaded` event never fires to stop the spinner).

bhousel:
Just pushed a change that should prevent the spinner from getting stuck spinning.

By the way, that spinner is only for OSM data tile requests.. It shouldn't do anything for other kinds of background requests (like notes, streetlevel data, whatever else we add, etc).

bhousel:
This is pretty good, and I think we'll be able to merge in the next few days..  Here are a few things I'd like to see:

- [ ]  I think we need to move "ownership" of the sidebar back to `behavior/hover.js`.  I've noticed that having `svg/notes.js` manage the sidebar means that if a user has an entity selected, and they happen to hover over a note, it can remove the sidebar content for the selected entity.  So easier to just have all of this sidebar stuff in one place
- [ ] Cleanup the sidebar styling
- [ ] Friendlier note date and user names.  Currently it displays something like:  "Comment from bhousel on 2017-11-06 08:59:55 UTC"
- [ ] Add an input field for a user to add a comment
- [ ] Add "Comment" and "Resolve" buttons

mmd-osm:
> click handling for notes is in the works right now.

Great, seems to be implemented now. I have to say that I'm not a big fan of the "auto center" feature, which I find a bit unexpected. Let's say, I zoomed in on my home area and want to go through all the notes, say, from the bottom left to the top right. Now every time I click on a note, the map gets centered on that note and I lost track on where I was before. Can we turn this auto center thing off please?

tordans:
Thanks for this feature! Where would you like to collect feedback – Ideas, issues, …? Separate issues in iD? Or here? Or somewhere external like a [reply on Twitter](https://twitter.com/bhousel/status/1019021979117072386?s=21)?

bhousel:
> Thanks for this feature! Where would you like to collect feedback – Ideas, issues, …? Separate issues in iD? Or here? Or somewhere external like a reply on Twitter?

Separate issues would be perfect, and we already have a few.  I read the replies on Twitter too because I know not everyone wants to use GitHub.

PR #5113, opened 2018-06-25
Add field `heights` to tag for `barrier=retaining_wall`

tordans:
The `retaining_wall` behaves the same as a [`hedge`]( https://github.com/openstreetmap/iD/blob/mast/data/presets/presets/barrier/hedge.json), then.

Wiki-Page: https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dretaining_wall

Do you need anything else for this PR?


bhousel:
Thanks @tordans  - this looks good to me 👍

PR #5116, opened 2018-06-29
use 22 as maxZoom for Esri imagery (selectively)

jgravois:
we actually serve the goods at 5cm in select locations!

new zealand

### DIFF
--- a/modules/renderer/background_source.js
+++ b/modules/renderer/background_source.js
@@ -338,7 +338,7 @@ rendererBackgroundSource.Esri = function(data) {
             }
 
             // if any tiles are missing at level 20 we restrict maxZoom to 19
-            esri.scaleExtent[1] = (hasTiles ? 20 : 19);
+            esri.scaleExtent[1] = (hasTiles ? 22 : 19);
         });
     };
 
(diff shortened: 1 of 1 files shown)


mmd-osm:
`esri.fetchTilemap` does all its analysis for z = 20. How do you know that you can switch from a successful  z = 20 evaluation to level 22? This seems a bit non-obvious.

jgravois:
that's a good point @mmd-osm.

in _most_ locations Esri serves up composited imagery from commercial providers and it is appropriate to set `19` as the native max zoom level. the tweak we incorporated in #5029 only checks the tilemap at zoom level `20` to provide a quick confirmation whether or not _any_ superior imagery is accessible.

i did this because i felt it would be sufficient to resolve the vast majority of the millions of 404s that this application generates knowing that iD can safely continue to downsample if/when tiles are _not_ fetched successfully in locations where the tilemap indicates that tiles are present at level `20` and users continue to zoom in closer only to find that they are _not_ present at `21` (or `22`).

that said, until you mentioned it, i forgot that this will require that we continue to catch 404's in order to not see `Map data not yet available` in those situations, so i've pushed up another commit to make that correction.

you can find a good example of a location with data at `20`, but no further here:

https://openstreetmap.org/edit?editor=id&#map=&map=20.00/34.09225/-117.55926

bhousel:
LGTM! 👍  Thanks @jgravois

PR #5120, opened 2018-07-03
Add capacity to playground:swing-preset

tordans:
The capacity defines, if there is one swing on the structure or more. A lot of playgrounds have one structure with two swings on it.

Wiki Capacity https://wiki.openstreetmap.org/wiki/Key:capacity
Wiki Playground https://wiki.openstreetmap.org/wiki/Key:playground (which does not reference capacity)


bhousel:
Thanks @tordans 👍

PR #5122, opened 2018-07-04
Some gpx tests.

vershwal:
Hey @bhousel!! I have added some tests in the gpx.js Please check.


bhousel:
Thanks!   I realized what was going on with the failing tests too.. The tests use a dummy `projection`, and it was pointing at 0,0, not at where the test point was.  So nothing was being rendered.

I fixed the projection and also made a few other changes to test that the path and texts actually contain some data we are expecting them to (see de1239665 for details) and then merged..

vershwal:
Thanks a lot, @bhousel !!

Issue #5040, opened 2018-05-16
Avoid jsonp


tomhughes:
Is there a reason why the wikipedia and wikidata API access uses jsonp rather than just fetching the JSON data with an XHR request?

The reason I ask is that I'm hoping to enable Content-Security-Policy enforcement for www.openstreetmap.org soon and as it stands I've had to enable script evaluation for *.wikipedia.org and www.wikidata.org but it's not clear that is really necessary.

bhousel:
Hmm, I don't think there is a reason - @1ec5 can we just use d3.json to fetch these?

mmd-osm:
Looks like there's some prior discussion on this topic: https://github.com/openstreetmap/iD/issues/2680#issuecomment-122054509

pnorman:
We should get CORS enabled on the specific read-only endpoints we're using, or come to the conclusion that WMF doesn't want their services used by external sites and do something from data dumps.

1ec5:
> can we just use d3.json to fetch these?

Sure, sounds good to me. (For the Wikidata field, I just followed the example of the preexisting Wikipedia field, which also used JSONP.)

> We should get CORS enabled on the specific read-only endpoints we're using, or come to the conclusion that WMF doesn't want their services used by external sites and do something from data dumps.

The [`query`](https://en.wikipedia.org/w/api.php?action=help&modules=query) and [`wbgetentities`](https://www.wikidata.org/w/api.php?action=help&modules=wbgetentities) actions are officially documented parts of MediaWiki’s public API, so I’m pretty sure the intention was for them to be used by external sites. (`query` has been used by bots and desktop editors for well over a decade.) As far as I can tell from Phabricator tickets, Wikimedia does enable CORS for requests coming from specific sites, but iD can be hosted literally anywhere.

bhousel:
Hmm @tomhughes I wonder if this would be an issue for #5050 - it also seems to use JSONP.

tomhughes:
Well that will certainly need a new CSP rule, which is fine. Allowing XHR is preferable to to JSONP though if it's possible.

bhousel:
There are a bunch of other uses of JSONP in the iD codebase that we should try to replace with XHR requests if possible..  I'm updating this issue to list them here:

- [ ] Wikipedia
- [ ] Wikidata
- [ ] Esri Imagery Metadata
- [ ] Bing Streetside

Abbe98:
[...]

PR #5123, opened 2018-07-04
Replace JSONP with XHR where possible

tomhughes:
This addresses #5040 by replacing JSONP with XHR where we can, namely:

* Bing imagery metadata access
* Esri imagery metadata access
* Wikipedia API access
* Wikidata API access

The only thing still using JSONP after this is Streetside because that doesn't currently have CORS support as far as I know.


bhousel:
Thanks @tomhughes for this!  Just tried it out and it works great 👍

Issue #4921, opened 2018-03-21
Prevent node drags that can damage relations


abhisheksaikia:
When remodelling roads in iD, nodes are dragged and small segments may be deleted. This causes the relations in that segment to be modified/deleted and the mapper uploads it without knowledge. A warning box saying a relation has been affected will prevent relation breakage or unknowingly adding or modifying a relation. Particularly helpful in case of turn restrictions

An example where a node was dragged and the `no_u_turn` got affected:`from` role was deleted without any warning to the mapper 👇 

![id feature 1](https://user-images.githubusercontent.com/13744156/37697480-1d3652e0-2d03-11e8-8eee-2f1ac5dea7ac.gif)

1ec5:
Part of the problem is that relation membership isn’t very visible – not on the map (#2946), and way below the fold in the feature editor. So even a mapper who’s familiar with turn restriction relations may be prone to accidentally deleting them.

bhousel:
I hope that's just an example edit and not something that people are actually doing.  That node drag probably moved the crosswalk tag onto the traffic signals node also.

We do have some situations where an action will be disabled if it will affect a relation, and we should add node drags to this check also.

bhousel:
I hacked together something today that should prevent this and warn the user:

![relation_conflict](https://user-images.githubusercontent.com/38784/38705758-aa85b092-3e78-11e8-867d-fcdde3c506ea.gif)

bhousel:
The change I pushed does fix the issue but also prevent certain harmless drags, so I'm reopening this improve it.

slhh:
@bhousel 
> Ok this is really fixed this time..
At least not without generating new issues. I think it will turn out to be much more complicated than you have expected again.

I had a short look at the test cases, and realized that it does still prevent certain harmless drags, especially ones fixing a broken relation. I have been able to verify this using the master deployment. Overiding the prevention using the Alt key doesn't really work, because the Alt key prevents connecting by itself. This makes the situation even worse.

Connecting members of the same relation with different roles is not invalid in general, and even in case of turn restrictions this can be quite valid. In addition the checks need to be relation type specific.

I can offer to work on this issue, but it will take some time. When do you intend to release the next version of iD?
I would likely combine this with my work already done for #4769 and #4771.

slhh:
@bhousel
>So if you have a specific case that the current code gets wrong, let me know!

1. nodes A,B,C,D,E
    way A-B: from member
    way B-C: via member
[...]

Issue #4320, opened 2017-09-07
Possibility to detach a node from a way


yaugenka:
It's not a killing feature but let me post it anyway in case somebody finds it helpful as well.

The only usecase I have for this is correcting someone's mistake of a named node attached to a way..
E.g. on this screenshot you see a place node attached to a way and there are no means of correcting this in iD other than making a copy of this node and clearing the tags in the old one, but in this case you looe the history.
![untitled2](https://user-images.githubusercontent.com/14937001/30189499-ed549136-943c-11e7-9fbb-7a08e1b9bb0a.png)

It would be nice  to make it possible to deattached a named node from a line in such a way that a new node is created in the place of the original one and the latter is preserved.

bhousel:
Agree, I actually wanted this feature earlier today too.

bhousel:
See #2511 for a convoluted workaround.  I closed that issue because I don't think shaking the node is a good way to detach it, but I'm open to the idea of some other operation to do it.

jjiglesiasg:
Agree as well, very often I found this cases and correction take much time...

jjiglesiasg:
Any news about this issue?

jjiglesiasg:
Any way to add a Disconnect feature for nodes? in #2511 you show the "current" ways to do the fix, but it take a lot of time and worse if the "way" is a part of a "relation"...

bhousel:
I'd do it like this:
- make `actionDetatch` which does the necessary changes to the graph (and add tests for it too)
- make `operationDetatch` which wraps `actionDetatch` into a nice menu command (and pick a key shortcut that will make everyone happy)
- add an icon for the menu too

slhh:
The usecase is relevant, but I don't like the suggested operation because it is generating coincident nodes.  These are hardly visible, and iD is quite bad in handling coincidence. And also the whole change of the operation would be hardly visible.
Using the operation accidentially can break OSM data, for example when detaching a barrier from the highway.

What we can do instead:
1. Don't let the detach operation replace the vertex. This have its own usecases, especially if it can be used selectively in view of the way or even the position in the way to detach from.
Unfortuneately, this doesn't support the special case of the original usecase well, where a tagged node is a connecting vertex.
2. We can implement the operation as originally proposed, but let it switch to move mode with the tagged node selected afterwards. This does unlikely result in coincident nodes, and the user gets aware of the operation because the node hangs on the mousepointer. Maybe, we can call that operation an "detaching move" or "move POI".

[...]

PR #5127, opened 2018-07-06
Add Detach Node operation #4320

Psigio:
Hi - I've had an attempt at implementing the Detach Node operation as described in #4320.  I've added the new operation, action & some unit tests for the action.  It will only operate on nodes that are part of at least one way and have at least one tag.  After detaching the node (and replacing with a duplicate without any tags in the way), it switches to move mode immediately so the user can move the detached node.

I'm not sure about the move mode switch, it's not perfect as the mouse pointer is offset from the node based upon the position in the context menu where the Detach Node operation is rendered.

Let me know if you want any changes made!

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -233,7 +233,8 @@ en:
       title: Detach
       key: T
       description: Detach this node from these lines/areas.
-      annotation: Detached a node from owning lines/areas.
+      annotation: Detached a node from owning lines/areas.      
+      via_restriction: "This can't be detached because it would damage a turn restriction."
   restriction:
     controls:
       distance: Distance
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -229,6 +229,11 @@ en:
       annotation:
         create: Added a turn restriction
         delete: Deleted a turn restriction
+    detachNode:
+      title: Detach
+      key: D
+      description: Detach this node from these lines/areas.
+      annotation: Detached a node from owning lines/areas.
   restriction:
     controls:
       distance: Distance
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -231,7 +231,7 @@ en:
         delete: Deleted a turn restriction
     detachNode:
       title: Detach
-      key: D
+      key: T
       description: Detach this node from these lines/areas.
       annotation: Detached a node from owning lines/areas.
   restriction:
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -302,7 +302,8 @@
                 "title": "Detach",
                 "key": "T",
                 "description": "Detach this node from these lines/areas.",
-                "annotation": "Detached a node from owning lines/areas."
+                "annotation": "Detached a node from owning lines/areas.",
+                "via_restriction": "This can't be detached because it would damage a turn restriction."
             }
         },
         "restriction": {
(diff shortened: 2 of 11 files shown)


mmd-osm:
I tested this patch, and I believe it causes serious side effects, if the node in question is part of a relation. Let's say, I'm detaching a via node of a turn restriction. This pull request creates a new node and replaces all occurrences in the involved ways, but it fails to update the relations. This will just break the turn restriction.

In JOSM, this detaching operation is quite involved. Maybe you could take a look how it is implemented there to catch all possible use cases and how to deal with them.

bildschirmfoto von 2018-07-08 14-14-18

Psigio:
Thanks for the information @mmd-osm, I hadn't considered relations membership! I've taken a quick look at the JOSM action, and I'll ask a question back on the original issue as I'm not sure if there are already any examples of actions in ID that require additional user input.

bhousel:
> Thanks for the information @mmd-osm, I hadn't considered relations membership! I've taken a quick look at the JOSM action, and I'll ask a question back on the original issue as I'm not sure if there are already any examples of actions in ID that require additional user input.

I recently added a lot of this logic to connect.js to close #4921 
I think the logic needed to disable the Detatch operation to prevent relation breakage might be similar (or simpler, hopefully)

https://github.com/openstreetmap/iD/blob/de123966541ad8301dbc3c7fc3161656eac02d26/modules/actions/connect.js#L60-L211

Psigio:
Thanks @bhousel - I've done a few experiments using the logic you added in the actions/connect file (I've moved it out to a new util file so I can re-use it without repeating it), and it seems ok.  My knowledge of turn restrictions is a bit basic, so I'll do a bit more reading up to see if I'm covering all the cases or whether a modified version of the logic is necessary for the detach node operation.

Psigio:
I've done a bit more reading on restrictions at https://wiki.openstreetmap.org/wiki/Relation:restriction and I think that the code can be simplified from the connect.js logic.  I've got an experimental branch which I'll work on for a bit to check it manages to prevent breaking the turn restrictions and once I've got some unit tests, I'll update the pull request.

bhousel:
cool thanks @Psigio !  I forgot to comment yesterday - I'm not sure whether it makes sense to split out the logic into a new util file.  I think it might depend on what the action is trying to do.  But if you can find a way to make it generic , go for it!

Psigio:
Hi,

I've done a bit more on this. The operation will now prevent the detach operation if the target node is a via or location_hint role within a restriction relationship. It won't prevent a node within a via way being detached as this shouldn't affect the restriction.

The action now also deals with other relationships too, it will swap the membership of the detached node to the new replacement one.

Hopefully this is a bit better now, happy to make any other changes that are necessary!

bhousel:
Thanks again @Psigio - I just merged this!

I made a bunch of cleanups in 119792fd2, but all minor stuff:
    - change keyboard shortcut to 'E' to not conflict with anything
    - move `disabled` check from operation into action and simplify
    - use `actionMoveNode` to place the detached node at the mouse cursor
    - disable the operation if the node is connected to hidden features
    - lots of code simplification
    - make the icon more centered

detach

Another cool thing with using `actionMoveNode` to put the detached node at the mouse cursor, is that this action can be performed in a transitioned way.  It feels magic to be able to do this just by pressing 'E' ✨:

detach2

PR #5130, opened 2018-07-07
Attempt to improve entrance searchability

tohaklim:
I am currently unable to bring-up entrance by typing in "entrance" on a vertex, this might help?

### DIFF
--- a/data/presets/presets/entrance.json
+++ b/data/presets/presets/entrance.json
@@ -3,6 +3,11 @@
     "geometry": [
         "vertex"
     ],
+    "terms": [
+        "entrance",
+        "exit",
+        "door"
+    ],
     "tags": {
         "entrance": "*"
     },


JamesKingdom:
Just typing `ent` is enough to bring it up for me.
iD already matches the preset name without having it in `terms`, are you sure it was on a vertex?

tohaklim:
I am guessing this has something to do with my browser locale not being English (and hence iD not being in English either). For some tags, just typing tag or key values works fine, but not for entrance

bhousel:
Oh interesting, typing 'ent' _should_ match this rule because it is the leading word in the tag key, but doesn't because of the `*`.   I don't know why it works this way, but this PR to add the terms should help.

https://github.com/openstreetmap/iD/blob/416ae66f7121c6293ad3d196ac90782a0f9d9d34/modules/presets/collection.js#L91-L93

tohaklim:
For some reason that didn’t fix the issue. Not a deal breaker so I’ll leave this for now, might look into it later

PR #5133, opened 2018-07-10
added level field to shop/kiosk

### LINKED ISSUE #5131: Default building:levels value for kiosks/others
Just tried to edit the preset and failed miserably, so opening an issue to see whether this is something worthwhile.
While kiosks are usually just nodes with `shop=kiosk`, some also have an outline with `building=yes/kiosk` (rarely).
Since I am yet to see a 2 level kiosk, would that be correct to default `shop=kiosk` to `building:levels=1`, despite not all of them having an outline? Could be only applicable to those that are an area (e.g. `building=kiosk`, or a trickier case `building=yes` + `shop=kiosk`, which is what I attempted to catch by setting levels on a node?
Other buildings I could see having default levels value are `building=service,bungalow,static_caravan,cabin,carport,cowshed,garage,garages,shed,sty`
Comments?

vershwal:
Fixes  #5131


bhousel:
Thanks @vershwal 👍

Issue #4526, opened 2017-11-14
Street-level imagery permalink does not update after switching images


1ec5:
As of v2.5.0, the link at the bottom-right corner of the Mapillary or OpenStreetCam imagery panel no longer updates as you switch from one image to another. mapillary.com always links to the first Mapillary image you view, even after you switch tracks; the same goes for openstreetcam.org.

/cc @bhousel

bhousel:
> As of v2.5.0, the link at the bottom-right corner of the Mapillary or OpenStreetCam imagery panel no longer updates as you switch from one image to another. mapillary.com always links to the first Mapillary image you view, even after you switch tracks; the same goes for openstreetcam.org.

Thanks for reporting @1ec5 - I do see this issue for the mapillary.com sitelink, but not the openstreetcam.org sitelink.  

Can you double check the openstreetcam case?  If so, can you let me know a location where you see openstreetcam traces doing this?

1ec5:
> Can you double check the openstreetcam case? If so, can you let me know a location where you see openstreetcam traces doing this?

I thought I was seeing this last night, but it isn’t happening now, so I may’ve been imagining things.

bhousel:
> Can you double check the openstreetcam case? If so, can you let me know a location where you see openstreetcam traces doing this?

This issue got autoclosed by the commit that fixed the Mapillary attribution.  I think OpenStreetCam is ok, but let me know if it's not and I need to reopen.

(can check http://preview.ideditor.com/master/  to confirm fix for Mapillary)

bhousel:
> I thought I was seeing this last night, but it isn’t happening now, so I may’ve been imagining things.

Thanks for checking!

PR #5137, opened 2018-07-12
Show MJS photo attribution

kratico:


bhousel:
Hmm we had a reason for this in #4526 which was closed by 1dd437a1f2
I think the reason was because at the time the built-in viewer attribution did not display the capture date, and we weren't able to reliably update it with d3.js.

Do we lose anything by switching back to the built-in viewer attribution?  (I'm ok with switching back but I would want to know the difference).

kratico:
The built in viewer displays a compact attribution if the viewer size is less than 640px.
With #5138 the viewer can be resized and a non-compact attribution is displayed.

MJS non-compact attribution
image

MJS compact attribution
image

Current attribution overlaps with MJS built in attribution
image

I think that the UX is better with the built in attribution, and the only difference it that MJS built in attribution does not show the time.

bhousel:
Nice thanks for screenshots, I agree the built in attribution is an improvement 👍

PR #5142, opened 2018-07-13
Add passenger information display to bus and tram presets

AndreasHae:
The [passenger information display](https://wiki.openstreetmap.org/wiki/Key:passenger_information_display) key was missing from bus and tram platform presets, so I added it. It could also be added to other transit types such as railways, but according to the wiki it does not need to be mentioned, so I left it out for now.


AndreasHae:
Oops, seems like I pushed some commits that do not belong here...

bhousel:
Thanks @AndreasHae  👍

PR #5145, opened 2018-07-14
Update de.json "K_I_nderskilift", the I is an i

tordans:
Fix typo

### DIFF
--- a/dist/locales/de.json
+++ b/dist/locales/de.json
@@ -2867,7 +2867,7 @@
                     "terms": "Materialseilbahn, Lastenseilbahn"
                 },
                 "aerialway/magic_carpet": {
-                    "name": "KInderskilift",
+                    "name": "Kinderskilift",
                     "terms": "Kinderskilift, Zauberteppichlift, Babylift"
                 },
                 "aerialway/mixed_lift": {


mmd-osm:
You need to fix translations in Transifex rather than creating a pull request for it here: https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#translating

tordans:
Thanks for the hint @mmd-osm.

I changed it at https://www.transifex.com/openstreetmap/id-editor/translate/#de/presets/19484573?q=key%3Amagic_carpet.

However, the translation does not show in http://preview.ideditor.com/master/ (its been a few hours).

Do I need to do something else?

mmd-osm:
> Do I need to do something else?

No, that's all. @bhousel will run `npm run translations` once in a while to pull in new translations from Transifex.

tordans:
Great, thanks.

I guess the wiki is out of date, right? It says

> You can check your translations in the master deployment (http://preview.ideditor.com/master/), which is updated every 10 minutes with the latest code and translation strings.

Should I create a PR to remove it? What would be the new recommended step for checking? (Maybe no step is OK as well.)

mmd-osm:
Ah, right, that's this preview site. It will indeed get updated much quicker. So just leave the wiki as is for now.

bhousel:
> I guess the wiki is out of date, right? It says
> 
> You can check your translations in the master deployment (http://preview.ideditor.com/master/), which is updated every 10 minutes with the latest code and translation strings.
> 
> Should I create a PR to remove it? What would be the new recommended step for checking? (Maybe no step is OK as well.)

Sorry about that!  I disabled it a few weeks ago because the transifex API was down and I was getting a lot of emails.  I just enabled the sync again and the updated translations should be appearing on the preview site again.

tordans:
> I just enabled the sync again and the updated translations should be appearing on the preview site again.

Perfekt, I see the change now.

Issue #5104, opened 2018-06-22
Refactor /services `loadTiles`


thomas-hervey:
Several services, including the image viewers like openstreetcam, mapillary, and streetside (along with notes and osm) require loading tiles. Besides checking for nullIsland (as the image viewer ones do), all of these files load tiles in a very similar way. See [openstreetcam](https://github.com/openstreetmap/iD/blob/2786974210f43e4ef62f0b23c4906f72caba4867/modules/services/openstreetcam.js#L107-L118) for example.

This could probably be refactored to a utility.

bhousel:
Yes!  Strong 👍 on this one..
It's been something I've wanted to do for a while.

We do kind of have a utility already in [`d3.geo.tile.js`](https://github.com/openstreetmap/iD/blob/master/modules/lib/d3.geo.tile.js), so it might make sense to:
1. move a lot of the common code into there
2. while also adding the capability to use variable tile sizes (it's currently hardcoded to 256px)

For historical context, d3.geo.tile is an older library that we pulled into iD and changed a little bit.  The version we use works fine for our use.  An old version of it lives [here](https://github.com/d3/d3-plugins/tree/master/geo/tile) and hasn't been updated in years.  That d3 plugins repo is not maintained anymore. 

I think the best way forward might be to move our internal `lib/d3.geo.tile.js` module into a `/util/utilTile.js` module with a new name so it wouldn't be confused with the old D3 v3 plugin, then give it the superpowers listed above.

thomas-hervey:
Several changes made in master after this merge including https://github.com/openstreetmap/iD/commit/e4d829ec06ccc9d73a8e45b8c627f95ca8fb0395 https://github.com/openstreetmap/iD/commit/d1fe81b9053d47e85dc0dfd687679f6e59afe95d https://github.com/openstreetmap/iD/commit/ff64455370b9a49679145d0b3e83c014e5d940c0 https://github.com/openstreetmap/iD/commit/02713e48fbeaa95fe3b82e4eae6c5c994f0cc262

bhousel:
I'm still working on a few more cleanups, and supporting sizes other than 256px.

PR #5148, opened 2018-07-15
moved getTiles and filtering to utils

thomas-hervey:
As noted in #5104 several services load tiles and rely on a modified version of d3.geo.tile. This reduces redundancy and refactors loading tiles to a new utility. The [notes feature](https://github.com/openstreetmap/iD/blob/0d8792d4dec4de2aae3f6984a92d68e5330badb6/modules/services/osm.js#L572) also refactors loading tiles and will need checking once merged.


bhousel:
I tried this out and activated the notes layer, but it looks like it's firing off a bunch more requests than it is supposed to which slows down the performance.. @thomas-hervey, can you poke around with the debugger and see what it's doing?

thomas-hervey:
@bhousel it seems that there are too many tiles being loaded and therefore too many requests. Debugging now.

bhousel:
Thanks @thomas-hervey  👍

PR #5149, opened 2018-07-15
Some mvt tests.

vershwal:
Hey @bhousel!!
Here are some tests for #5072. Please check.


test/spec/svg/mvt.js
```diff
@@ -0,0 +1,98 @@
...
+        surface.call(render);
+
+        var path = surface.selectAll('path.mvt');
+        expect(path.nodes().length).to.eql(1);
+        expect(path.attr('d')).to.match(/^M.*z$/);
+    });
+
+    describe('#url', function() {
+        it('handles pbf url', function () {
+            var url = 'https://api.mapbox.com/v4/mapbox.mapbox-streets-v6/9/150/194.vector.pbf?access_token='
```
bhousel: We should avoid linking to a mapbox streets pbf here.
Can you try to use [sinon](http://sinonjs.org/releases/v6.1.3/
) to fake a response?

Here is an example in the OpenStreetCam test:
https://github.com/openstreetmap/iD/blob/master/test/spec/services/openstreetcam.js#L54-L105

I think it would be ok to store .pbf test fixture in the `test/data` folder, and use [`fs.readFile`](https://nodejs.org/dist/latest-v6.x/docs/api/fs.html) to load the data .

bhousel:
Cool, thanks for fixing these to avoid the network access!

PR #5152, opened 2018-07-17
Update Mapillary color scheme

kratico:
This PR includes
- Update lime-green to Mapillary green
- Add a different layer style for selected and hovered sequences

Before
image

After
image

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -141,8 +141,17 @@
 .layer-mapillary-images .viewfield-group * {
     fill: #39AF64;
 }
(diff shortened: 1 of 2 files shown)


bhousel:
Ahh sorry, I'd rather not do this.. The colors were chosen so that they would contrast the other two streetside imagery providers and all the other stuff that appears on the map.

I think it's more important for mappers to be able to distinguish Bing Streetside, Mapillary, and OpenStreetCam traces, even if that means we don't strictly use the preferred brand colors of each one.

streetlevel

PR #5155, opened 2018-07-17
fix browser-specific Date() issues in notes

### LINKED ISSUE #5154: Notes showing with date = "null"
This is from https://github.com/openstreetmap/iD/issues/5153

@manfredbrandl is using German locale.

1c9e8552-fd24-4017-b976-4d5fdf836847

thomas-hervey:
fixes #5154 by replacing "-" with "/" in note comment Date strings like "2018-07-05 20:27:09 UTC"

### DIFF
--- a/modules/ui/note_comments.js
+++ b/modules/ui/note_comments.js
@@ -99,6 +99,7 @@ export function uiNoteComments() {
         if (!s) return null;
         var detected = utilDetect();
         var options = { day: 'numeric', month: 'short', year: 'numeric' };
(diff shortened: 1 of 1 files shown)


bhousel:
Looks good, thanks for the quick fix @thomas-hervey 👍 

We do have a handful of other fixes like this scattered around the iD codebase, so it might be worth considering adding a library to work around the quirks of JavaScript date string parsing.. @kamicut suggested [date-fns](https://github.com/date-fns/date-fns) for this.  Feel free to evaluate it and send another PR if it's something you want to try!

PR #5161, opened 2018-07-19
Update map directional indicator based off of current bearing in MapillaryJS viewer

kratico:
This PR also leaves visible the directional indicator only for the current image

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -177,6 +177,13 @@
 .layer-mapillary-images .viewfield-group * {
     fill: #55ff22;
 }
+.layer-mapillary-images .viewfield-group .viewfield {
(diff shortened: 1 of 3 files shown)


bhousel:
> This PR also leaves visible the directional indicator only for the current image

Thanks, I like the bearing indicator for the panoramic images!  But I ended up restoring the display of viewfields for the non-selected images.  

I tried using the Mapillary layer without them, and I thought it was just too frustrating not knowing which direction an image would be pointed until I actually clicked on it.  I get that the viewfields are kind of visual clutter, but 1. they are only shown at z18 and higher, and 2. the solution to visual clutter in iD can be "just zoom in more until it's not a problem".

PR #5162, opened 2018-07-19
Add new note

thomas-hervey:
Adds a new note feature, including `new_note` and `drag_note` modes.


bhousel:
Thanks again @thomas-hervey - I just merged this!  I took out the categories for now, but I want to put them back in (#5171) after we think about it more.

PR #5180, opened 2018-07-25
Use Web Workers(wip)

kepta:
This brings in support for web workers in iD. The goal of this PR is to offload some work from the main iD process to a thread.  As a first attempt, I am trying to only parse the node, ways, relations in the web worker and then sending a simple JSON back to the main thread. 

There are a bunch of things that need to be fixed, for eg tests, renaming `idly-faster-osm-parser` -> `osm-bizly`, remove es6 code, get the changes in osm-auth merged, etc. 

@bhousel what do you think about the overall changes/code flow?

### DIFF
--- a/build_src.js
+++ b/build_src.js
@@ -15,6 +15,8 @@ module.exports = function buildSrc() {
 
         // Start clean
         unlink('dist/iD.js');
+        unlink('dist/iD-worker.js');
+        unlink('dist/iD-worker.js.map');
         unlink('dist/iD.js.map');
 
         console.log('building src');
@@ -22,7 +24,7 @@ module.exports = function buildSrc() {
 
         building = true;
 
-        return rollup
+       const main =  rollup
             .rollup({
                 input: './modules/id.js',
                 plugins: [
@@ -48,12 +50,34 @@ module.exports = function buildSrc() {
                     sourcemap: true,
                     strict: false
                 });
+            });
+        const worker = rollup
+            .rollup({
+                input: './modules/worker.js',
+                plugins: [
+                    nodeResolve({
+                        module: true,
+                        main: true,
+                        browser: false
+                    }),
+                    commonjs(),
+                    json()
+                ],
             })
-            .then(function () {
+            .then(function(bundle) {
+                return bundle.write({
+                    format: 'iife',
+                    file: 'dist/iD-worker.js',
+                    sourcemap: true,
+                    strict: false
+                });
+            });
+        return Promise.all([main, worker])
+            .then(function() {
                 building = false;
                 console.timeEnd(colors.green('src built'));
             })
-            .catch(function (err) {
+            .catch(function(err) {
                 building = false;
                 console.error(err);
                 process.exit(1);
--- a/modules/util/worker_handler.js
+++ b/modules/util/worker_handler.js
@@ -1,0 +1,21 @@
+import { createWorker } from './walkie_talkie';
+
+var workerInstance = createWorker('dist/iD-worker.js');
+
+export function parseBboxEntities(url, done) {
+    workerInstance.sendMessage(
+        { type: 'PARSE_BBOX_ENTITIES', url: url },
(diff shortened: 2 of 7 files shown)


build_src.js
```diff
@@ -45,12 +47,34 @@ module.exports = function buildSrc() {
...
+                    }),
+                    commonjs(),
+                    json()
+                ],
             })
-            .then(function () {
+            .then(function(bundle) {
+                return bundle.write({
+                    format: 'iife',
+                    file: 'dist/iD-worker.js',
```
bhousel: It's fine to generate a worker.js, but we really should leave `dist/iD.js` alone

modules/util/worker_handler.js
```diff
@@ -0,0 +1,21 @@
...
+export function parseBboxEntities(url, done) {
+    workerInstance.sendMessage(
+        { type: 'PARSE_BBOX_ENTITIES', url: url },
+        function(error, reply) {
+            done(error, reply);
+        }
+    );
+}
+
+export function authdParseBboxEntities(xhrParams, done) {
```
bhousel: I don't understand why there are separate auth'd and non auth'd parsers.  At the point that the worker gets involved haven't they already been downloaded?

build_src.js
kepta: Do you mean not putting `iD-worker.js` to the `dist` folder? or renaming it `worker.js`?

modules/util/worker_handler.js
kepta: The worker code makes the network request to OSM. There are two possible network requests that iD wants to make, one with authentication headers and the other without it. Since the worker makes the network request and not the main iD, I made two separate worker functions to do the same.

bhousel: Yea but I thought you said you were going to start off with a version that doesn't do the request, and only does the tile parsing, since that's easier.

kepta: Ya, I did intend to start off with that but then the passing back and forth between worker and main is too slow. I think letting the Webworker make the network request is the best possible way we can offload the entire task and with the current approach I think it is working quite fine.

mmd-osm:
Parsing XML line by line feels like a bit of a hack, and not very sustainable for the future. To make parsing easier, we're looking into ways to enable JSON output, and getting rid of XML parsing altogether.

Watch this one for progress: https://github.com/zerebubuth/openstreetmap-cgimap/pull/153

mmd-osm:
You can try out some JSON output on the dev server for now: https://upload.apis.dev.openstreetmap.org/api/0.6/map.json?bbox=-80.28218507766725,25.773537461179274,-80.24941921234132,25.782676914766654

kepta:
@mmd-osm is there an expected timeline for _JSON output_ to land to production stack?

mmd-osm:
@kepta : https://github.com/openstreetmap/operations/issues/244 has the latest status.

The cgimap PPA package is ready for deployment. However, there are some concerns that the Rails port doesn't support JSON, and this would complicate future development. Hopefully this will get resolved very soon.

bhousel:
> However, there are some concerns that the Rails port doesn't support JSON, and this would complicate future development. Hopefully this will get resolved very soon.

AFAIK doing this in Rails is more-or-less automatic.
https://www.leighhalliday.com/responding-with-json-in-rails
https://stackoverflow.com/questions/2088280/in-rails-how-do-you-render-json-using-a-view

bhousel:
Anyway we don't need to wait for the OSM JSON API to be ready to set up a webworker background task queue.  I'd use it immediately for fetching lots of things.. (pretty much anything under `modules/services` would benefit).

mmd-osm:
> Anyway we don't need to wait for the OSM JSON API to be ready

In case it wasn't obvious:  it **is** ready on the dev instance, and we'd very much appreciate any sorts of feedback **before** JSON output hits production. So please do give it a try while it's still on dev!

PR #5196, opened 2018-08-01
Fix shellbang

district10:
fixed missing `!` in shellbang. You can check which shell is using with `readlink /proc/$$/exe` in the scirpt. (On my Ubuntu:16.04, it's using `/bin/dash` right now)


bhousel:
Thanks!  Weird that this line was working without it:
https://github.com/openstreetmap/iD/blob/master/.travis.yml#L10

PR #5198, opened 2018-08-03
Allow Float values in height input box

### LINKED ISSUE #5184: Editor: Height Tag Forced to Integer Value
When specifying the height for a building as a decimal (eg. 4.7), the input field converted my input to an integer. Looking at the [wiki for Key:height](https://wiki.openstreetmap.org/wiki/Key:height), decimal values should be permitted.

jguthrie100:
Fixes #5184 
Seems to be as simple as switching the parseInt to parseFloat when parsing input values.

Only issue is that it allows float values for all tags set using the input box (i.e. 'motorway lanes', which should presumeably just be an integer! Not sure how to avoid this though, without doing something like `type=integer` or `type=float` in the tag presets json files

### DIFF
--- a/modules/ui/fields/input.js
+++ b/modules/ui/fields/input.js
@@ -92,7 +92,7 @@ export function uiFieldText(field, context) {
 
     // parse as a number
     function parsed(val) {
-        return parseInt(val || 0, 10) || 0;
+        return parseFloat(val || 0, 10) || 0;
     }
 
     // clamp number to min/max


bhousel:
Thanks @jguthrie100 !

> Only issue is that it allows float values for all tags set using the input box (i.e. 'motorway lanes', which should presumeably just be an integer! Not sure how to avoid this though, without doing something like type=integer or type=float in the tag presets json files

Yeah, I thought about adding a dedicated `integer` field for fields that should be one, but I worry that many of the tags that we think of as integers also need to support text entry like `no` or `yes`, because this is an established OpenStreetMap convention.

PR #5201, opened 2018-08-03
Keep right QA

thomas-hervey:
After the recent introduction of notes in iD, I've been discussing with @kamicut and @bhousel about the future of QA tools. A new `QA` section under "Map Data" could be used to display QA issues, such as [KeepRight](https://www.keepright.at/), a data consistency check that automatically identifies issues.

screen shot 2018-08-03 at 5 00 27 pm


bhousel:
So excited to try this out!! @thomas-hervey I know you've been working on it the past few days, but it looks like tests are passing again. Let me know if it's ready!

thomas-hervey:
@bhousel, go ahead and test it out. I will be pushing small updates, but at this point it's ready for checking and cleaning.

thomas-hervey:
@magol thanks for your interest. I have been on the road so my work has been intermittent over the last weeks. I am going to work on it a bit more today and I hope to get feedback from @bhousel. However, feel free to checkout the branch and play around with it.

tordans:
Hi @thomas-hervey can you give me a hand testing this? I cannot find the new feature in my dev editor.
What I did:
1. follow https://github.com/openstreetmap/iD#installation
2. checkout `keep-right_QA`
3. run the rpm commands from 1 again
4. open http://localhost:8080/#background=Berlin-2018&disable_features=boundaries&map=17.00/52.51669/13.39015

How would I see the new feature? I dont find it in the panels (eg next to the nodes?) or on the map itself.

Thanks

thomas-hervey:
@tordans, I reproduced your steps and everything should be working. To turn on the layer, look under Quality Assurance at the bottom of the Map Data menu (hotkey F). There is a toggle for KeepRight. 

At this point, there are several outstanding tasks including tile optimization, more complex translations, custom settings (in the works), and more. Let me know if you still have issues.

SilentSpike:
I see there used to be an issue with the error description parsing on error type 411 (http error) due to the URL being collected together with the surrounding tag data (`href="URL">URL</a>)`). Which would result in:
```
The URL (<a target="_blank" href="\href=URL>URL</a>)\">{var1}</a>) cannot be opened (HTTP status code 503)
```

I was looking into this until I realised some of the newer commits resolved the issue, but it got me thinking about the way you're parsing the descriptions in `parseErrorDescriptions`. Is there any reason you opted not to just extract the details you're looking for by capturing them as regex groups? It would save all the hassle of iterating over the split strings as well as prevent having to add special cases for the descriptions that don't have details split neatly between space characters (like type 220).

The strings you have in `errorSchema.json` are effectively patterns of a different syntax, in a way you're reinventing the wheel there 😄

thomas-hervey:
@SilentSpike thanks for your input. Hmm I see what you're saying. No I didn't do this for a particular reason except that I figured splitting would handle most cases, and regex would be used for identifying IDs. But yes at this point something more consistent like regex groups could help.  A PR would be great. I'm only able to spend part time working on this atm.

SilentSpike:
Minor bug I've found while testing. [This error](https://www.keepright.at/report_map.php?schema=86&error=105286846) (type 70) seems to show up as the wrong error type (I've had 41 and 50) in the iD layer.

thomas-hervey:
@tordans, thanks for your feedback. Looking into your list of errors now. Hopefully some are handled by @SilentSpike's refactoring ^.

SilentSpike:
An update on my previous comment: It should be fixed in my PR (I found that what the wiki lists as error 70 is actually error 74 when inspecting the GeoJson).

modules/util/keepRight/errorSchema.json
```diff
@@ -217,7 +217,7 @@
             },
             "_220": {
                 "title": "misspelled tags",
-                "description": "This (node|way|relation) is tagged '([\\w:]+)=(.+)' where &quot;(\\2|\\3)&quot; looks like &quot;([\\w\\s]+)&quot;",
+                "description": "This (node|way|relation) is tagged '([\\w]+)(:([\\w]+))?=(.+)' where &quot;(\\2|\\3|\\4|\\5)&quot; looks like &quot;([\\w\\s]+)&quot;",
```
SilentSpike: Hmm, this is an interesting case. I have a feeling this won't work for all cases because the pattern will now only match keys with 0 or 1 namespace (which doesn't handle the likes of `turn:lanes:backward`).

I see that the bit in quotes can be any substring of the key though (I wonder if that also applied to the value). It might be easiest to revert the key group and then just use a `.+` in the group within the quotes.

matkoniecz:
From what I see at https://github.com/keepright/keepright this tools seems to be not maintained anymore

 -  last commit in February 2017 https://github.com/keepright/keepright/commits/master
 - unreviewed PR - https://github.com/keepright/keepright/pull/91
 - known false positives waiting for fix (for example https://github.com/keepright/keepright/issues/60 )

I am not sure whatever exposing iD users to this tool is a good idea.

Maybe Osmose would be a better idea?

matkoniecz:
>  I have a feeling that if iD links to it, people will take more of an interest in maintaining and keeping the server going.

I have not considered this.

bhousel:
Just rebased this work onto the latest master, so we can get this merged soon!

tordans:
@thomas-hervey would now be a good time for me to update https://github.com/openstreetmap/iD/pull/5201#issuecomment-417834268 with a second impression feedback? – I am just checking if you think its worth the effort (for me) and helpful (for you/here).

SilentSpike:
Tried using github's built in conflict resolver (as my local repo is in a very dirty WIP state) for the first time on #5298 which seems to have done something odd to the commit history on that branch. Just going to close that and comment here instead:

Error type 73 isn't limited to `tracktype` which results in the incorrect string often being displayed for errors of this type, valid fix is to capture the group `([\w:]+)` in place of word `tracktype` in its [error schema](https://github.com/openstreetmap/iD/blob/ca2d4e2c34475e3c5d78ff419317425eeaba9e69/modules/util/keepRight/errorSchema.json#L50). Remember to update the corresponding [localised description](https://github.com/openstreetmap/iD/blob/ca2d4e2c34475e3c5d78ff419317425eeaba9e69/dist/locales/en.json#L1223) too.

thomas-hervey:
@tordans that would be great! However I think that a good number of those issues haven't been addressed. What would probably be better would be to re-catalog those issues (and I can help too) once this branch has been merged. @bhousel and I have been thinking that we will likely try to merge this soon and then open another PR to address the small issues.

bhousel:
> What would probably be better would be to re-catalog those issues (and I can help too) once this branch has been merged

I agree - lets get this merged first as its own layer and data service and then we will consider feedback.  This week a group of us is working on a more general validation flow, and a new dedicated issue pane, and the KeepRight detections can be just one set of validation issues that we think might fit into it for mappers to act on.

thomas-hervey:
@nyurik I agree that limiting validation data is a good approach. However, someone who is more familiar with data items should comment. I think this should be revisited once this branch is merged and bugs are addressed but before integrating other QA tools.

data/keepRight.json
```diff
@@ -0,0 +1,393 @@
...
+    },
+    "72": {
+      "title": "",
+      "severity": "error",
+      "description": "This node is not member of any way and does not have any tags"
+    },
+    "73": {
+      "title": "",
+      "severity": "error",
+      "description": "This way has a tracktype tag but no highway tag"
```
SilentSpike: Example of where this error requires regex:

https://www.keepright.at/report_map.php?schema=86&error=93785774

bhousel: Thanks good catch @SilentSpike !

from the source https://github.com/keepright/keepright/blob/dc7a8ada2c87f7cf68d9a1a546cebb53a1f0c281/checks/0070_missing_tags.php#L97-L108

SilentSpike: Oh nice, I hadn't even realised source was available for keepright 😄

bhousel:
### I just merged this!  🎉

Thanks @thomas-hervey for working on this, and I'm looking forward to more integration with other Q/A tools.  It will be great to see mappers using this to improve OpenStreetMap!

Thanks also @tordans and @SilentSpike for the detailed feedback on this thread.  I've spent the past few weeks improving the user experience and was able to incorporate a lot of your feedback into the latest code.  There are also some other good ideas in there that I'll open as follow up issues.

For the curious, you can try it out now in http://preview.ideditor.com/master by enabling the layer on the Map data pane.

screenshot 2019-01-04 22 38 04

mvl22:
Great news! - thanks :)

Looking forward to seeing this in live.

Bibi56:
> From what I see at https://github.com/keepright/keepright this tools seems to be not maintained anymore
(...)
> I am not sure whatever exposing iD users to this tool is a good idea.
> 
> Maybe Osmose would be a better idea?

By selecting Osmose and KeepRight in iD, I got false positive from KeepRight ([name and name:XX](https://github.com/keepright/keepright#106)).
So Osmose is definitively a better option: here KeepRight lower the quality of the database. A QA tool should do the opposite^^.

jjiglesiasg:
In my experience, both are very useful, why to remove one of them?

Issue #4977, opened 2018-04-09
Custom WMS imagery


NKAmapper:
Now that WMS support is implemented, it would be great to be able to use WMS in the custom background menu as well.

bhousel:
Tagging @vershwal, who is working on replacing the alert box with a proper dialog per #4806

jidanni:
Currently all the user can find is
```
Custom Background Settings

Enter a tile URL template. Valid tokens are:
   {zoom} or {z}, {x}, {y} for Z/X/Y tile scheme... etc.
```
In https://www.alpinequest.net/en/help/v2/maps/on-demand-create#wms_servers they use a elegant $bbox.

1ec5:
A minimal implementation of WMS background imagery support was added in 2a28a03ca13f68f0d94ae0ca7aedb2b9b4530a3d as part of #7428, which was reverted for unrelated reasons. Reopening this issue to track either restoring 2a28a03ca13f68f0d94ae0ca7aedb2b9b4530a3d or finishing #6915.

Issue #4806, opened 2018-02-15
Replace alert box with modal ui


jidanni:
Allow copying the example (switch...) in
![1](https://user-images.githubusercontent.com/1230959/36260649-0450b1ba-129d-11e8-9614-d3c2d74e1e6c.jpg)
in fact allow copying the whole thing.
As it is currently, the user can't even do Inspect Element on it.

bhousel:
All the normal window stuff like select/copy/paste works fine here for me (in chrome 64)... What else have you tried?   Are you sure this is a real issue?

![copy alert text](https://user-images.githubusercontent.com/38784/36260978-ea2cbd90-1230-11e8-9831-62212da32a0a.gif)

jidanni:
I can do like you in Firefox.
I can copy the text mentioned in
https://bugs.chromium.org/p/chromium/issues/detail?id=5879


But, with

Package: chromium
Version: 64.0.3282.119-2
Kernel: Linux 4.14.0-3-amd64 (SMP w/2 CPU cores)

the pointer remains an arrow over the entire box.
There is no right context menu.

bhousel:
It sounds like a browser bug outside of our control. 

If we swap out the `window.prompt` for a custom modal, as described here, https://github.com/openstreetmap/iD/issues/4802#issuecomment-365741541 then we wouldn't be relying on the browser behavior and this issue would go away.

jidanni:
OK submitted https://bugs.chromium.org/p/chromium/issues/detail?id=812644

jidanni:
In https://bugs.chromium.org/p/chromium/issues/detail?id=812644#c6
they say
> Could you provide HTML+CSS for alter box?
> It seems user-select CSS property prevents to select contents in alter box
> except for text box.

bhousel:
> In https://bugs.chromium.org/p/chromium/issues/detail?id=812644#c6
> they say

You can point them at http://preview.ideditor.com/master as an example..

`user-select` should not affect the selectability of text in a `window.prompt`, and I'm using Chrome and I don't see this issue, so it might just be Chrome on Linux that has the issue.

bhousel:
Assigned to @vershwal

bhousel:
This was done in #5207!

PR #5207, opened 2018-08-06
Replacing alert box with modal ui

vershwal:
Related #4806 #4977.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4602,3 +4602,24 @@ li.hide + li.version .badge .tooltip .tooltip-arrow {
     width: 100px;
     color: #7092ff;
 }
+
+.button-ok {
+    background-color: #6f92ff; /* Green */
+    border: none;
+    color: white;
+    padding: 15px 15px;
+    text-align: center;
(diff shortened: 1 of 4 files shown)


bhousel:
Hey @vershwal this is good!  I made some changes and merged it..

Here are some of the things I changed:
- I moved the whole thing into a new subfolder `modules/ui/settings`.  What you built was really good and I realized that we are going to add a bunch more modal settings screens just like this one (for example to customize the vector tile and other data layers).  So this gives us a place to organize them.
- There is another special kind of `uiModal` called `uiConfirm`, which already has an OK button, so I used that one.  It makes the code a bit simpler..
- Dispatch a change event instead of trying to call a function back in `uiBackground`
- Add `utilNoAuto` to the textarea (this just prevents autocomplete and other annoying behaviors especially in Safari)

vershwal:
Thanks, @bhousel!! Just saw the UI, looks great!!
There were many new things involved in this issue...Enjoyed doing this!!! If you find any other similar issue please tag me.

PR #5222, opened 2018-08-13
remove ?blankTile=false (and space) from Esri tilemap urls to fix 404

jgravois:
calls to `tilemap` currently fail because i forgot to trim an invalid url parameter when they are constructed.

screenshot 2018-08-13 09 28 26

should be: https://server.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/tilemap/20/550340/621736/8/8


bhousel:
Thanks @jgravois !

PR #5243, opened 2018-08-21
Vector Tiles pt 2

bhousel:
Still a bit of cleanup to do, but this PR merges: 
* the layer formerly known as "gpx" (really gpx / kml / geojson) with 
* the layer formerly known as "mvt" (really pbf / mvt but without the ability to place spatially in a z/x/y tile coordinate).

These features now live on a single layer called "Custom Map Data"  (yes we will support multiple custom layers soon) and it has a settings screen (like #5207 screen added for custom background imagery).   Users can choose a local file, enter a url to a remote file or a url template for a vector tile source:

screenshot 2018-08-21 18 52 26

It currently converts the vector tile data to geojson on the fly and renders just like the old gpx layer did (better styling coming soon too)!

screenshot 2018-08-21 18 34 50

screenshot 2018-08-21 18 36 26

HUGE thanks to @vershwal for her Google Summer of Code work that made this possible and @geohacker for being an awesome mentor 🙌


bhousel:
This is ready to merge!   I've added the ability to hover over data and select it.  Internally the data is stored as GeoJSON, so we show whatever is in `feature.properties` in a readonly raw tag editor.  

I'm still considering this feature somewhat "beta", but I'll release a new version of iD in the next day or so that allows users to play around with it, and will include a few example vector tile layers which can help mappers participating in the [Detroit Mapping Challenge](https://www.openstreetmap.us/2018/07/detroit-mapping-challenge-sotmus2018/).

data inspector2

Issue #5239, opened 2018-08-19
public outdoor wifi hotspot


jidanni:
iD doesn't help one put a https://wiki.openstreetmap.org/wiki/Tag:internet_access=wlan on the map.

JamesKingdom:
This was done in #5251

PR #5251, opened 2018-08-26
Add preset for wi-fi hotspots

simonbilskyrollins:
Fixes issue #5239 see also https://wiki.openstreetmap.org/wiki/Tag:internet_access=wlan


vershwal:
Hey @simonbilskyrollins. Looks good, but you need to run npm run build before pushing any code, this command will build your project and also notify if there is any error. In your case after adding data/presets/presets/internet_access/wlan.json if you run npm run build it will update the following four files :
data/presets.yaml
data/presets/presets.json
data/taginfo.json
dist/locales/en.json

I would suggest you to build the code and push the changes again.
Thanks!!

simonbilskyrollins:
Thanks for the feedback @vershwal! I just built and pushed the changes, so I think it should be good to go now.

JamesKingdom:
Thanks @simonbilskyrollins! 👍

PR #5269, opened 2018-08-31
Fields and default layer for man_made:bridge

LaszloEr:
- Added name, bridge type, layer and maxweight fields
- default layer should be set to 1 (like it is for highway=bridge)

### DIFF
--- a/data/presets/presets/man_made/bridge.json
+++ b/data/presets/presets/man_made/bridge.json
@@ -1,9 +1,16 @@
 {
+  "fields": [
+    "name"
+    "bridge"
+    "layer"
+    "maxweight"
+  ],
   "geometry": [
     "area"
   ],
   "tags": {
     "man_made": "bridge"
+    "layer": "1"
   },
   "name": "Bridge"
 }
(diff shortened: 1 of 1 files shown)


lakedistrictOSM:
Thanks for this PR, this is a much needed improvement.

vershwal:
Great @LaszloEr!! Seems like some changes are required in the code. After you do the changes suggested by @SilentSpike do not forget to build the project before you push the changes. You can do it by running the command `npm run build`
Thanks!!

LaszloEr:
I added the missing commas, this is what it looks like:
bridge

Nakaner:
I don't think that adding a default `layer=1` is a good idea. It bears the risk that users who don't care for the `layer` tag add a bridge which has a different layer than the road on it.

It is easier for both data consumers and mappers to find objects without a certain tag than objects where a tag which was set by default is wrong.

LaszloEr:
Good point. The reason why I wanted to add it as the default is because setting the bridge attribute on roads also adds it there automatically, probably because it's the correct for 99% of all bridges.

bhousel:
Thanks @LaszloEr - this seems ok to me 👍

PR #5270, opened 2018-08-31
Update column.json

tordans:
An ad-column is a round thing the way I know it and the way it is shown in the wiki (https://wiki.openstreetmap.org/wiki/Tag:advertising%3Dcolumn). So the direction key as preset does not make sense. There seems to be no relevant usage of this combination as well (https://taginfo.openstreetmap.org/tags/advertising=column#combinations).


vershwal:
Hey @tordans, you should build the project before you push any changes this will also notify if there is any error. You can do it by running the command ` npm run build`. In your case once you change the file data/presets/presets/advertising/column.json and run ` npm run build` it will update the changes to data/presets/presets.json

I would suggest you to build the code and push the changes again.
Thanks!!

tordans:
@vershwal thanks, I did so just now.

bhousel:
Thanks @tordans 👍

PR #5273, opened 2018-08-31
Changing highway colors for iD v3.0

KylieStarzz:
Basically, iD has been around for some time and it can only be more beneficial if we update it to v3.0.

Firstly, I've suggested that the in-editor highway colors be changed for iD 3.0, as the reason is that the highways don't look similar to the ones on OpenGeoFiction's style by Histor.


bhousel:
Thanks but I’m pretty happy with the colors as they currently are.  The main goal of iD is to support mappers on openstreetmap.org, so conflicts with the style of OpenGeoFiction are not really a good reason to change things. 

I’m very happy to see other sites fork iD and make it their own!

PR #5275, opened 2018-08-31
[KeepRight] Use regex group capture to extract error details

SilentSpike:
As discussed [here](https://github.com/openstreetmap/iD/pull/5201#issuecomment-417714858), this PR improves the extraction of details from error descriptions by directly using regex group capture instead of splitting the strings and using a lot of logic to figure out what's what (effectively re-implementing the same group capture behaviour).

Advantages of this approach:
- Can precisely extract what we want, no relying on space characters to split the string nicely.
  - No need to handle edge cases in the code like this:
https://github.com/openstreetmap/iD/blob/74e06d47472af3c32c8b8012f5b89b5f1af43763/modules/util/keepRight/keepRight_error.js#L66-L73
- Less looping and calls to regex `.test()/.match()` methods.
- We're being more explicit about what we expect in the error strings (aka `\d+` shows that we expect a number whereas `{$1}` didn't). This is a minor maintainability improvement.
- Can easily handle errors that have multiple possible string templates (see types 312 and 282).

Side effect of this approach:

- I've changed the way that IDs are identified. The code to identify IDs and their type (node/way/relation) was a bit clumsy (and wouldn't correctly handle some cases) especially with some inconsistencies in the error messages it had to account for. It became a bit more so once the code was no longer looping over every word. So instead I've added an extra optional `"IDs"`  key to the `errorSchema.json` which provides an array of strings where the indices correspond to the regex groups ([see here](https://github.com/SilentSpike/iD/blob/eb78395ee8132114d773f80f4b0c2497b91f1081/modules/util/keepRight/errorSchema.json#L82-L85)). This let's us explicitly tell the code which groups capture an ID and what type of ID they are (`"n"/"w"/"r"`).
  - It seems there are a couple of cases where strings can have an arbitrary list of node/way/relation IDs depending on the geometry involved in the error. For getting the IDs in these cases I'm capturing the whole list as a group and parsing those with unique code (see error types 211, 231 and 294).
- To make life easier I've added a `"regex"` key to the `errorSchema.json` so that errors where the message is fixed aren't parsed for details at all (saves us having to escape any special regex characters and prevents needlessly matching a fixed string for many errors). However, it might actually make more sense to just remove these messages all together from the file since we don't need that data for anything.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -695,6 +695,15 @@ en:
           _72:
             title: 'node without tags'
             description: 'This node is not member of any way and doesn''t have any tags.'
+          _73:
+            title: 'tracktype with no highway tag'
+            description: 'This way has a "tracktype" tag but no "highway" tag.'
+          _74:
+            title: 'empty tag'
+            description: 'This {var1} has an empty tag: "{var2}".'
+          _75:
+            title: 'name without tags'
+            description: 'This {var1} has a name ("{var2}") but no other tags.'
           _90:
             title: 'motorways without ref'
             description: 'This way is tagged as motorway and therefore needs a ref nat_ref or int_ref tag.'
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -781,7 +781,7 @@ en:
             description: 'These errors contain self intersecting ways.'
           _211:
             title: ''
-            description: 'This way contains more than one node multiple times. Nodes are {var1}, {var2}. This may or may not be an error.'
+            description: 'This way contains more than one node multiple times. Nodes are {var1}. This may or may not be an error.'
           _212:
             title: ''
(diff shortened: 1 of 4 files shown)


SilentSpike:
It seems there are a couple of edge cases all related to strings which can have an arbitrary list of nodes/ways/relations depending on the geometry involved in the error.

At the moment I'm capturing the first item in the list as it's own group (because we know there must be at least one item) and then the rest of the list as another group. This means that the first ID becomes a link but the rest remain as text.

I've marked these cases in `errorSchema.json` with `"TODO"`.

thomas-hervey:
Thank you very much @SilentSpike for the great refactor. I've reviewed these changes and I am fine merging them so long as there aren't parsing conflicts between the templates and the KeepRight.at server responses.

Issue #5049, opened 2018-05-22
Replace or merge iD data at runtime


bhousel:
related to https://github.com/osmlab/osm-planning/issues/2

It would be useful to be able to replace the `imagery` and `presets` based on a config file passed in via the url params.

Ideally we would support both "replace" and "merge" for these. 

http://example.com/#imagery=whatever.json
http://example.com/#mergeImagery=whatever.json

http://example.com/#presets=whatever.json
http://example.com/#mergePresets=whatever.json

cc @kriscarle @geohacker

Issue #2589, opened 2015-04-16
Request to support marine focused "fuel" tag


brycenesbitt:
The iD editor presently supports tagging of amenity=fuel, calling it a gas station in English.

This is a request to support tagging of a marine fuel facility.  These are not interchangeable features.  There are two candidate tags for this:

seamark:small_craft_facility:category=fuel_station
waterway=fuel

Note that separate amenity=fuel and waterway=fuel may exist within the same leisure=marina.  So proximity is not enough to determine the facility type.

bhousel:
> seamark:small_craft_facility:category=fuel_station
318 uses in taginfo

> waterway=fuel
166 uses in taginfo

so it seems that the seamark tag is the better one, yes?

brycenesbitt:
The seamark:small_craft_facility:category tag takes a semicolon separated list, and is
not processed by anything other than OpenSeaMap.  The tags have long diverged from OpenStreetMap.
So while the numbers are higher, it's not clear if the tag is actually better.
A pragmatic solution is to tag both, not trying to predict a winner, but helping support the mapping of the feature.

pnorman:
Any idea how often `amenity=fuel` occurs on water?

brycenesbitt:
@pnorman I've seen this tagging, but don't know how to readily make a query to count.

Note that marine fuel stations are invariably on a floating dock, and thus for new mapping can reasonably be tagged on a node of a man_made=pier.

bhousel:
> The seamark:small_craft_facility:category tag takes a semicolon separated list, and is
> not processed by anything other than OpenSeaMap. The tags have long diverged from OpenStreetMap.

Oh, I see.  Yeah the iD presets are not able to adjust specific values inside a semicolon separated list and that not something we will support.

`waterway=fuel` is my preference also..  Looks like you added a wiki stub for it recently?

brycenesbitt:
Yes, I added the stub.  It's in use, it makes sense, and it sets a nice pattern (amenity=fuel/waterway=fuel,  amenity=sanitary_dump_station/waterway=sanitary_dump_station).  This is a case of structuring data well for later rendering also.

pnorman:
I'm downloading the data, I suspect water and land-based fuel stations are tagged the same.

brycenesbitt:
@pnorman I'm interested in your results.

[...]

PR #5286, opened 2018-09-03
Adds preset for seamark:type=buoy_lateral

### LINKED ISSUE #5285: Create preset for seamark:type=buoy_lateral
Lateral [buoys](https://wiki.openstreetmap.org/wiki/Seamarks/Buoys) are anchored, floating makers in waterways that line the edges of channels. The [`seamark:type=buoy_lateral`](https://wiki.openstreetmap.org/wiki/Tag%3Aseamark%3Atype%3Dbuoy_lateral) tag has been used [over 42,000 times](https://taginfo.openstreetmap.org/tags/seamark%3Atype=buoy_lateral#overview), yet it has no preset in iD.

- Name: Lateral Buoy
- Tags: `seamark:type=buoy_lateral`
- Allowed geometry: points
- Common fields: Name, Ref, Operator
    - Note: seamarks mostly use the [`seamark:name`](https://wiki.openstreetmap.org/wiki/Key:seamark:name) key rather than `name` or `ref`
- Speciality fields. These are critical since a buoy's appearance determines its meaning. Each can be edited with dropdown selectors.
    - Color (`seamark:buoy_lateral:colour`)
    - Category (`seamark:buoy_lateral:category`)
    - Shape (`seamark:buoy_lateral:shape`)
    - System (`seamark:buoy_lateral:system`)
- Icon: buoy

quincylvania:
Closes #5285 by adding a Lateral Buoy preset and four related fields: buoy color, category, shape, and system.


bhousel:
Ohhh [seamark](https://wiki.openstreetmap.org/wiki/Seamark_tagging_schema)..  😭 
I've avoided adding any of these kinds of tags to iD as presets.

The reasons are:
* The practice of embedding `object`, `attribute`, `#` in the key is not always compatible with how iD can setup presets and fields.  For example, specifying [units]( https://wiki.openstreetmap.org/wiki/Seamarks/General_Attributes#Height.2Flength_units_.28HUNITS.29) or sector number of a [sectored light](https://wiki.openstreetmap.org/wiki/Seamarks/Sectored_and_Directional_Lights) (like `seamark:light:1:colour=green`) in the key is a real problem for how iD works.  I actually think we could support most of the basic `seamark:<object>` stuff, but it gets really scary fast.
* I don't think these are the sort of preset an untrained mapper could use.  Most people know what a "buoy" is, but definitely not whether it is "lateral", or how to recognize any of the categories or iala systems.  I think you'd find that people would just add a node for anything that looks like a buoy and the preset would be very likely to be misused.

Are you sure you really want to add this into the main iD for everyone?  I'm inclined to say no, but I could definitely see a "basic seamark preset pack" being a thing once we implement #5049 and people can add their own custom presets to iD at runtime.

quincylvania:
Hi Bryan, thanks for the detailed response. I was thinking there must be a reason that no seamark presets have been added yet. I agree with your points. I'm not set on this addition whatsoever—I'm not even a heavy seamark user myself—but here are some additional thoughts you may consider:

- Seamarks represent a sizable percentage of OSM features. iD shouldn't be all things to all people, but it currently excludes the needs of seamark mappers altogether. I admit the tagging schema is unfortunately complex, though like you said, supporting basic `seamark:<object>` style tagging would be doable and should be considered.
- A "seamark preset pack" sounds like it could be an excellent idea. However, I would still find it amiss to not include _any_ support for so numerous a set of tags out of the box.
- I wonder how often most casual mappers run across nautical features at all. I would assume that most stick to land and wouldn't have much opportunity to mistag seamarks, but I don't have proof of this.
- Presets would help casual mappers understand existing seamarks that they run across, as well as easily follow the links to the seamark documentation. Without presets, encountered seamarks just look confusing and possibly incorrect. If mistagging is truly an issue, a read-only approach by excluding seamark presets from search might be worthwhile.
- As for specifically mistagging the type of buoy, presets such as "Red Buoy" and "Green Buoy" could be added. The color is a sufficient start for identifying the type of buoy, is observable in imagery, and doesn't require the user know the difference between lateral, ordinal, and other buoy types.

Since trepidation remains on the matter, I suggest further discussion with the community before proceeding. I may try implementing more basic seamarks to see how far iD can get with the schema. Ideally, some sort of user testing might tell us how seamark presets would actually be received, but I'm new around here and I'm not sure if that's possible.

Thanks for taking the time to read and consider this ✌️

bhousel:
Thanks for understanding @quincylvania 

The surface area of seamark tagging is huge:
https://wiki.openstreetmap.org/wiki/Seamarks/Seamark_Objects

I'm ok with adding a few of the most common things that casual mappers might come across and understand how to map.  Buoys seem like a good start.

Can you share a location on OpenStreetMap that you think is pretty well mapped seamark-wise?  Most mappers (including me) are pretty unfamiliar with it.  As far as I can tell, it's only ever come up for discussion once before on the iD issue tracker (#2589)

quincylvania:
This PR has been superseded by #5297.

PR #5293, opened 2018-09-05
Creates Lean-To and Gazebo presets for the shelter_type tag

### LINKED ISSUE #5292: Create Lean-To and Gazebo shelter_type presets
`lean_to` and `gazebo` are both common values of [`shelter_type`](https://wiki.openstreetmap.org/wiki/Key%3Ashelter_type) (5000 and 1100 instances, respectively) and are typical features that a mapper may identify and search for. I propose adding these as presets in the same vein as the Pavilion preset.

quincylvania:
Closes #5292.

Also creates a Fireplace checkmark field and adds it to the Lean-To and Wilderness Hut presets.


bhousel:
Thanks @quincylvania !  Looks pretty good.. 

Just 2 nitpicky things:
* When there are a bunch of "nested" presets like this, we usually put them under another folder.. (see like `amenity/vending_machine/*`).  Can you move `lean-to.json`, `gazebo.json`, and `pavilion.json` under a new folder `amenity/shelter/`

* Can a lean-to really have a fireplace?  The [wiki page for it](https://wiki.openstreetmap.org/wiki/Tag:shelter_type%3Dlean_to) specifically says it doesn't, and I wouldn't want people to complain that there is a checkbox field for it.  (However it seems totally fine to add it to wilderness hut, thanks!)

quincylvania:
Hi Bryan, thanks for the notes. I agree with your points and will make the changes. Some explanation:

- The pre-existing Pavilion preset wasn't nested so I didn't nest the rest.
- I added the Fireplace field to Lean-To because I [noticed some lean-tos that use it](https://overpass-turbo.eu/s/BFI). However, these may be tagging mistakes since, as you said, lean-tos generally don't have indoor fireplaces. Those examples may be mislabeled cabins or wilderness huts, or more likely lean-tos near external firepits where the firepits should be mapped separately.

bhousel:
Awesome, thanks again @quincylvania  👍

PR #5298, opened 2018-09-06
[KeepRight] Fix error type 73

SilentSpike:
Looks like this one isn't specific to tracktype


SilentSpike:
Realised I accidentally closed this while cleaning up branches on my fork, believe the issue is still present

PR #5301, opened 2018-09-08
Adds Location and Length fields to the Swimming Pool preset

### LINKED ISSUE #5300: Add location field to swimming_pool
As per the wiki https://wiki.openstreetmap.org/wiki/Tag:leisure=swimming_pool

Perhaps also length as well?

quincylvania:
Closes #5300.

This implementation uses `location=roof` over `location=rooftop` since the [`location` page](https://wiki.openstreetmap.org/wiki/Key%3Alocation) indicates they're likely synonyms and `roof` has far more usage. "Rooftop" is used for the label, however, to make all the labels adjectives.


bhousel:
Looks great, thanks @quincylvania 👍

PR #5304, opened 2018-09-10
Adds navigation of the preset list using the arrow keys

### LINKED ISSUE #4324: Navigate presets with ↓↑
When typing a preset name, you can press Enter to accept the top one. It would be convenient to be able to select a different one with the arrow keys.

screenshot_2017-09-08_18-16-16

Pressing ↓ in this case should highlight Huishoudelijkeapparatuurwinkel, pressing again should go to the one below...

quincylvania:
Closes #4324. 

- The up and down arrows move the focus to the previous and next list items
- The right and left arrows open and close category items
- The search field is treated as the top item in the list and can be navigated to and from

This is my first JavaScript contribution so feedback and critique are especially welcome.


bhousel:
Nice @quincylvania the code looks really great.  I really like how you handled expanding/contracting the category menus, and included the search at the top.

The only piece of feedback is that I'd really like for the styling to work the same as if the user is using the mouse.  It looks like with keyboard navigation, only the icon part is shaded, where with the mouse, the entire list item is shaded.  Can you change this and then merge it?

keyboard nav preset

PR #5310, opened 2018-09-15
Adds Sculpture, Statue, and Mural presets

quincylvania:
Also adds a universal Material field. This is used for the Sculpture and Statue presets, but is made universal since the [`material=*`](https://wiki.openstreetmap.org/wiki/Key:material) tag has been used over[ 836,000 times](https://taginfo.openstreetmap.org/keys/material#overview) on a variety of features.


bhousel:
Hey @quincylvania this looks pretty good.. Can you do one more thing:

When a preset uses multiple tags, we can add a `"reference"` property to control what osm wiki page will get fetched when the user hits the ℹ️ button.

For example, this is from a vending machine preset:

https://github.com/openstreetmap/iD/blob/1bf77fa1c73e9ee749971b5af73095010e16c282/data/presets/presets/amenity/vending_machine/public_transport_tickets.json#L20-L27

This means "use the `vending=public_transport_tickets` page, instead of the `amenity=vending_machine` page".

Can you add this property for all the new artwork presets too?  Thanks!

quincylvania:
Oh _that's_ what the `reference` property is for. Thanks for the explanation and example. Updated in the last commit.

PR #5330, opened 2018-09-26
Changes the preset name "Tree row" to "Tree Row"

quincylvania:
This aligns the preset's name capitalization with other multi-word preset names.


bhousel:
Good catch 👍

PR #5340, opened 2018-09-29
Adds the "P" icon to the parking space preset

quincylvania:


bhousel:
Oh I don't think this is a good idea to use the 🅿️icon.  I have seen issues with new users picking "Parking Space" when they really should pick "Car Parking" (parking lot).  It really is kind of confusing, especially for people who aren't native English speakers.. 

What we should probably do is:
- Rename "Car Parking" preset to "Parking Lot"

I had some other thoughts about Parking Space too.. The proposal for them is pretty confusing:  https://wiki.openstreetmap.org/wiki/Proposed_features/parking

I kind of think we should leave them alone for now.  If we can figure out an easy way to tag what kind of parking space it is, that would be nice, but the `access:whatever` tag is unfortunate.  I was hoping we could just show a dropdown with options like "disabled, motorcycle, auto, truck..." but I don't think we can.

quincylvania:
Okay! I was wondering if leaving the icon off was an oversight or intentional. I'm fine not adding it.

1ec5:
> Oh I don't think this is a good idea to use the 🅿️icon.  I have seen issues with new users picking "Parking Space" when they really should pick "Car Parking" (parking lot).  It really is kind of confusing, especially for people who aren't native English speakers..

I opened #5341 about the 🅿️ icon being used for parking lots.

PR #5343, opened 2018-09-29
changelog: fix typo

tordans:
I guess https://github.com/openstreetmap/iD/search?q=embarkment&type=Code is a typo, since the preset is https://github.com/openstreetmap/iD/blob/master/data/presets/presets/embankment.json and the wiki page as well https://wiki.openstreetmap.org/wiki/DE:Key:embankment


bhousel:
Yes you're correct - good catch 👍

PR #5355, opened 2018-09-30
Add Picnic Shelter preset

programistka:


bhousel:
Thanks again @programistka - looks good 👍

Issue #5334, opened 2018-09-27
Add design tag values preset to high-voltage tower 


Adamant36:
Hello. Currently the design tag and its values are not available as presets when someone is adding a high-voltage tower to a power line. Is there any way it can be added?

bhousel:
Good suggestion @Adamant36 - This is a great task for anyone who wants to get involved and learn how to add a preset field to OpenStreetMap's iD editor.  I've tagged it as Hacktoberfest, so wait until October 1 to get credit.

For reference here is the [OSM wiki for high voltage power towers](https://wiki.openstreetmap.org/wiki/Tag:power=tower).

That page does suggest a `design` tag with several choices.  Because there are so many, I think we should use a `combo` field configured to pick from the most common choices that mappers have added to OpenStreetMap.  We can use taginfo to see [how the `design` key is currently used](https://taginfo.openstreetmap.org/keys/design#values).

The presets and fields in iD are all defined in .json files under `data/presets/*` folder.  There is also a [README.md](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md) there to explain how to format the .json files.

To complete this task:
1. create a file `data/presets/fields/design.json` that contains this:
```js
{
    "key": "design",
    "type": "combo",
    "label": "Design"
}
```
2. Edit the file `data/presets/presets/power/tower.json` and add `"design"` to the `"fields"` array.  The order of elements here is the order they will appear on the screen, so add it to the beginning.
3. run `npm run build` - this will rebuild the generated files
4. run `npm run start` and test your change locally.  If you draw a line and tag it as a "Power Line", then add a vertex along that line and tag it as a "High Voltage Tower", you should see your new "Design" field offered as a choice here in the sidebar:

<img width="200" alt="screenshot 2018-09-27 11 13 36" src="https://user-images.githubusercontent.com/38784/46155990-859da300-c246-11e8-8f4e-9bc77793ed39.png">

If all looks good, send a Pull Request.  Thanks in advance!

boothym:
Good idea, maybe tower:type as well - though that also contains values for man_made=tower. Unless only the values here can be shown in iD? https://wiki.openstreetmap.org/wiki/Tag:power=tower#Tower_type

What about changing the name to "High-Voltage Power Tower" so it shows up in a search for power?

And is there a tower icon that could be used instead of the dot?

bhousel:
> What about changing the name to "High-Voltage Power Tower" so it shows up in a search for power?

[...]

PR #5267, opened 2018-08-30
Add preset for shop=sewing

SilentSpike:
Fixes #5244

PR #5356, opened 2018-10-01
Issue #5334

hchho:
- Add power term to tower
- Add design to fields for tower
- Add design tag values preset to high-voltage tower

### DIFF
--- a/data/presets/fields/design.json
+++ b/data/presets/fields/design.json
@@ -1,0 +1,5 @@
+{
+    "key": "design",
+    "type": "combo",
+    "label": "Design"
+}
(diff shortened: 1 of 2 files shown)


quincylvania:
@hchho Thanks for the PR! A few things:

- Before we can merge the changes, you need to run `npm run build`, verify there are no errors, and then commit the generated files. See [this comment](https://github.com/openstreetmap/iD/pull/5267#issuecomment-417607117) for more info.
- I'd recommend giving the PR a descriptive title so people know what it does at a glance, and then say something like "closes #5334" or "resolves #5334" in the comment. This will link the PR and the issue and let GitHub automatically close the issue when we merge.

hchho:
Got it. Thanks for the feedback! I've updated my pull request with your suggestions.

Issue #5346, opened 2018-09-29
Add Softball Field preset


quincylvania:
A [softball field](https://duckduckgo.com/?q=softball+field&iax=images&ia=images) has different dimensions than a baseball field, and often has a dirt infield. Softball fields are common in the U.S., especially around schools and universities.

Note: there is currently a page for a [`baseball=softball`](https://wiki.openstreetmap.org/wiki/Tag:baseball%3Dsoftball) tag. This proposal does not appear to be widely accepted and has fewer uses than [`sport=softball`](https://taginfo.openstreetmap.org/tags/sport=softball), so I'm ignoring it for now.

- Path: `leisure/pitch/softball`
- Name: Softball Field
- Geometry: `point`, `area`
- Fields: `name`, `lit`, possibly others
- Tags: `"leisure"="pitch"`, `"sport"="softball"`
- Reference: `"key": "sport"`, `"value": "softball"`
- Terms: `softball diamond`
- Icon: `maki-baseball`

s-Hale:
Added :)

Issue #5345, opened 2018-09-29
Rename "Baseball Diamond" preset to "Baseball Field"


quincylvania:
While the term "baseball diamond" is sometimes used to refer to the entire play area in baseball, it more precisely refers to just the area within the four bases (which form a diamond). The [recommended way to map baseball pitches](https://wiki.openstreetmap.org/wiki/Tag:sport=baseball#How_to_Map_Baseball_Fields) in OSM is to cover the entire field, not just the infield. The English name of the `leisure/pitch/baseball` preset should be changed to "Baseball Field" to better indicate what should be mapped.

`baseball diamond` should be added under `terms` so it will still be searchable.

Spotts9:
Wouldn't mind taking this one on.  I've forked and started looking at it

quincylvania:
@Spotts9 Go for it!

PR #5357, opened 2018-10-01
Changed Baseball Diamond to Baseball Field, left Diamond in the searc…

s-Hale:
…h terms, added field to serch terms


bhousel:
closes #5345 
closes #5346

PR #5358, opened 2018-10-01
Add Transit Shelter preset

### LINKED ISSUE #5348: Add Transit Shelter preset
`public_transit` is the most common value of [`shelter_type`](https://wiki.openstreetmap.org/wiki/Key:shelter_type), with over 66,000 uses. It would be useful as a preset, especially for possible use with public transport stop area relations.

- Path: `amenity/shelter/public_transport`
- Name: Transit Shelter
- Geometry: `point`, `area`
- Fields:`name`, `shelter_type`, `building_area`, `bench`, `lit`
- Tags: `"amenity"="shelter"`, `"shelter_type"="public_transport"`
- Reference: `"key": "shelter_type"`, `"value": "public_transport"`
- Terms: `bus stop`, `metro stop`, `waiting`
- Icon: `maki-shelter`

enighter:
Closes #5348


data/presets/presets/amenity/shelter/public_transport.json
```diff
@@ -0,0 +1,28 @@
...
+        "lit"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "terms": [
+        "bus stop",
+        "metro stop",
+        "waiting"
```
Nakaner: A bus stop is something different than a shelter and tagged differently (a preset exists already). I think that this will let users use the wrong preset to add bus stops.

PR #5362, opened 2018-10-01
Flow actions

mattbhenley:


bhousel:
Thanks, but per https://github.com/openstreetmap/iD/issues/3744#issuecomment-374950519 I've decided not to do flow annotations for now.

PR #5372, opened 2018-10-06
Adds the Solar Panel preset

### LINKED ISSUE #5371: Add Solar Panel preset
```JSON
"tags": {
        "power": "generator",
        "generator:source": "solar",
        "generator:method": "photovoltaic",
        "generator:type": "solar_photovoltaic_panel"
}
```

quincylvania:
Closes #5371. Also:

- Changes the Power Output field type from "text" to "typeCombo"
- Renames the source_nuclear and source_wind presets to source/nuclear and source/wind
- Adds point geometry to the power/tower preset


data/presets/fields/generator/output/electricity.json
```diff
@@ -1,6 +1,6 @@
 {
     "key": "generator:output:electricity",
-    "type": "text",
+    "type": "typeCombo",
```
bhousel: I feel like this field might have been `text` to allow capital letters like "MW" in the value units. 
Can you test to make sure that taginfo will still return results with capital letters?  

If taginfo is preferring lowercase values, you might need to add this field to the uppercase regex here: https://github.com/openstreetmap/iD/blob/master/modules/services/taginfo.js#L263-L267

quincylvania: It looks like `:output` is already in the regex, so there's no problem with capitalization.

However, I noticed that the field replaces spaces with underscores, which is undesirable for this key. I'll remove the type change from this PR and make a new issue.

PR #5376, opened 2018-10-07
Fix nordic type field

yvecai:


bhousel:
Thanks @yvecai good catch

PR #5383, opened 2018-10-07
Fix spelling

JamesKingdom:


quincylvania:
Oops! That's what I get for not making a PR.

Issue #5367, opened 2018-10-04
Add fields to amenity=bench preset


Adamant36:
As it is, the only field that the bench preset has is for backrests. It would be cool if it also had fields for the material, seats, and colour tags. Since they are all used on benches a lot. Especially material.

manfredbrandl:
I support this, because taginfo tells me that usage of fields for material, seats and colour is high (nearly 400K to above 100k) of total over 950K benches.

dryo:
Good idea! I'd also like this for leisure=picnic_table, especially if it has a roof or not (covered = yes/no)

Adamant36:
@dryo, good call. I was actually thinking the same thing for picnic tables also. You should do a new issue for it.

PR #5387, opened 2018-10-08
Add fields to bench preset and create seats field

simonbilskyrollins:
Fixes issue #5367


quincylvania:
@simonbilskyrollins Thanks! Could you add `"minValue": 0` to the `seats` field? Nobody wants negative seats 😆

Also, consider adding some placeholder like `"placeholder": "2, 4, 6..."` to the field to indicate that it's numeric.

simonbilskyrollins:
@quincylvania good point! I've added both of those things to the PR.

bhousel:
@simonbilskyrollins - thanks for your PR! 👍   
@quincylvania - good catch on the "negative seats" 😆

Issue #5374, opened 2018-10-06
Mapillary sign overlay not showing up


tohaklim:
Judging by the lack of issues on here, I am the only one that can't bring up Mapillary traffic sign overlay?
The mapillary image cone overlay plugin works. 
Tried Chrome 60 and Edge

bhousel:
@cbeddow fixed this with #5395, but there is a follow-on issue to fix.
The signs layer is working again, but the detections overlaid on each image are not working yet.

PR #5395, opened 2018-10-10
fix: traffic signs to new endpoint

cbeddow:
iD was using the old `v3/objects` API and needs to be `v3/map_features` with the parameter `layers=trafficsigns` set. Also changed `v3/detections` to `v3/image_detections` API although I think this is not used right now.


bhousel:
Thanks @cbeddow - I restored the newlines back from CRLF -> CR, and fixed the tests.  

There is another issue with the api call which fetches the detections within the current image (this is to draw the boxes over the detected signs), so I'll leave #5374 open for now.

bhousel:
This was actually merged - not sure why GitHub missed it

PR #5397, opened 2018-10-11
No differentiation of piste type field is needed

yvecai:
Limiting the choice to connection, playground or snowpark when in the piste:type=downhill or nordic preset is probably a bad good idea: if it's a downhill, it's not a snowpark.
Also after testing it a little, it's disturbing to see the piste field change when you change the piste type.


bhousel:
Thanks ⛷

PR #5396, opened 2018-10-11
Adds a button that lets users manually download individual relation members

quincylvania:
This is a complete version of the very stale #2284. Differences:
- Only individual members can be downloaded, one at a time.
- The map now zooms to relation members selected from the sidebar if they're not visible. This is especially needed since manually downloaded members are often offscreen, but it useful anytime a member is offscreen.
- A download icon is provided.
- Tooltip text is added.

**Hover**
screen shot 2018-10-10 at 8 44 19 pm
**Downloading**
screen shot 2018-10-10 at 8 44 21 pm
**Downloaded**
screen shot 2018-10-10 at 8 44 23 pm
**Selected (zooms to member)**
screen shot 2018-10-10 at 8 47 43 pm

PR #5399, opened 2018-10-12
Adds the member geometry type to the label of non-downloaded relation members

quincylvania:
With #5396 merged, users can now manually download individual members. The issue is that that it is difficult to tell what the member is before it is downloaded. While the member's role is shown in the input field and its index is indicated by its list order, its geometry type is available but hidden. This PR adds the type (node, way, or relation) to the non-downloaded member's label. This should make it slightly easier for users to determine what members to try downloading, as well as surface potential mistakes such as `node` members with `north` or `south` roles.

A potential drawback is that novice users may be confused by the node/way terminology since iD hides it behind points, lines, and areas, but I don't think that novice users will be editing complex relations anyway.

screen shot 2018-10-11 at 7 36 21 pm

### DIFF
--- a/modules/ui/raw_member_editor.js
+++ b/modules/ui/raw_member_editor.js
@@ -141,7 +141,14 @@ export function uiRawMemberEditor(context) {
 
                     } else {
                         var incompleteLabel = d3_select(this).append('label')
-                            .attr('class', 'form-label')
+                            .attr('class', 'form-label');
+                        
+                        incompleteLabel.append('span')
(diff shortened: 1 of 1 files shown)


bhousel:
> A potential drawback is that novice users may be confused by the node/way terminology since iD hides it behind points, lines, and areas, but I don't think that novice users will be editing complex relations anyway.

I'm ok with this - we use the Node/Way terminology in a few other places that are more intended for advanced users.

Issue #5398, opened 2018-10-11
Are preset tags exclusives?


yvecai:
Say I draw a line, uses the presets to tag it waterway=river.
Then I use the preset to tag it boundary=administrative, admin_level=2.
Then it's no longer a river ?

Or did I don't do it properly ?

bhousel:
> Say I draw a line, uses the presets to tag it waterway=river.
> Then I use the preset to tag it boundary=administrative, admin_level=2.
> Then it's no longer a river ?

Yes, a feature can only really match one preset in iD.

yvecai:
Wow, then the presets are a bit useless, isn't it?
Is this intended or just hard to fix?

yvecai:
Hmm, you're taking this too literally, even this page mentions exceptions.
My issue at hand: What if a road is a ski piste in winter?

bhousel:
> My issue at hand: What if a road is a ski piste in winter?

My recommendation is to just draw an extra line for the ski piste, next to the road.  It's easier for people to understand and easier for piste-map-makers to consume the data that way.

yvecai:
Huh, what? 8-)

yvecai:
A good part of the ski pistes mapped are also highways (37%, or 61 789 elements). This mapping pattern is quite common, and your suggestion doesn't make a lot of sense. I won't even comment on duplicate, offset geometries ;-). I guess one can argue that being a piste in winter is a property of the underlying highway, and as a data consumer, I'm please to know that.

But in this state of things, this means that having pistes presets in iD is not a good idea, given that the contributor using them risk to erase highway without notice.
However, from a UI point of view, I think that using preset to map piste is very good.

What would be the way to go forward ?

yvecai:
##1408

yvecai:
Apparently, using the highway field like in #1408 preserve highway=**yes**.
Like in:
```
{
    "icon": "maki-skiing",
    "fields": [
        "name",
        "piste/type",
        "piste/difficulty_nordic",
        "piste/grooming_nordic",
        "oneway",
        "lit",
        "highway"
    ],
```
Maybe it should preserve highway=whatever.

yvecai:
A solution could arise from the mappers or from iD before these preset can be used safely. 
See https://taginfo.openstreetmap.org/keys/piste%3Atype#combinations for reference.

yvecai:
The change to avoid removing existing tags is easy.

[...]

PR #5401, opened 2018-10-12
Remove piste presets for now, as they are likely to delete highways

yvecai:
See #5398 
It seems safer to remove these presets for the time being, contributors being use to tag smaller highway=* with piste:type=* tags.


bhousel:
> It seems safer to remove these presets for the time being, contributors being use to tag smaller highway=* with piste:type=* tags.

Really? If you remove them then even standalone pistes will not be recognized as pistes - this is probably not what you want.

yvecai:
That's 4 chance out of 10 to delete a track, 60% for piste:type=nordic which is the majority of pistes mapped. 
No, I don't feel like to take the chance.

I hope it's not forever though, I may dig more into the code and also take some advice in the community.

bhousel:
Look, I'm not going to do this.  It would remove even the downhill ski runs.  iD has had piste presets for like 5 years and people seem to be able to figure them out.  We render things so people don't get confused, and local communities are mostly good at watching for unintentional breakage.

Similarly, we aren't removing the tram presets just because we are worried someone might change a street into a tram.  Nobody does that.

yvecai:
Bryan, I've figured this issue while writing a short tutorial to map with iD some nordic pistes the way it's done for maybe 10 years.
Look how it's easy too ... wait, where the track? 


My advice is to think of it for a while and not to hurry. I guess downhill pistes are less subject to follow highways, so the preset could be limited to downhill if this please you.


But we need to figure it out properly in the end.

I say this while I'm a nordic skier, and data consumer at opensnowmap.

PR #5402, opened 2018-10-14
Adds highlighting of relation members in the map when hovering on their list item

quincylvania:
This lets users see what feature the member refers to before selecting it, reducing the need for guessing.

ezgif-1-ae0af1a4c3b9

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -265,6 +265,18 @@ text.point {
     opacity: 0.8;
 }
 
+/* Highlighting */
+
+g.point.highlighted .shadow,
+path.shadow.highlighted {
+    stroke-opacity: 0.95;
+    stroke: #7092ff;
+}
+g.vertex.highlighted .shadow {
+    stroke-width: 7;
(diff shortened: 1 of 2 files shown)


bhousel:
Nice, looks good to me..
I wonder if it might even be better if the shadow was a little more obvious, like how we style the shadows in the turn restrictions editor.  Are there situations where the selected feature is pulsing and so you can't see the highlight on the relation member?

also +1 for screenshot of Philly

quincylvania:
Now that you mention it, if the relation is a multipolygon then all the lines are pulsing and you couldn't see the member highlighting. I'll try adding a more prominent style that works in both situations.

I'm originally from Philly!

quincylvania:
I updated the hover styling to use the link color and be more opaque. It now also works with multipolygons.

screen shot 2018-10-13 at 11 05 35 pm
screen shot 2018-10-13 at 10 57 12 pm

bhousel:
This looks pretty great!  Thanks @quincylvania 🎉

PR #5404, opened 2018-10-15
Adds hover highlighting to the selected features list

quincylvania:
The behavior is the same as the hover highlighting in the relation members list, implemented in #5402.

screen shot 2018-10-14 at 6 29 43 pm


bhousel:
Looks good, thanks! 👍

Issue #5408, opened 2018-10-17
Add To and From fields to Route presets


quincylvania:
`to` and `from` are common keys for [routes](https://wiki.openstreetmap.org/w/index.php?oldid=625726#Route). Text fields should be added for these keys and the fields should be added to all Route presets, such as Train Route and Bus Route.

PR #5410, opened 2018-10-18
Add fields for To and From and add fields to all route presets.

castriganoj:
I took a stab at this as a first time iD contributor. Please review and let me know if anything if changes are needed. 

Thanks for participating in Hacktoberfest. 

Best, 
Jim

#5408


JamesKingdom:
@manfredbrandl The changes about Helsingborg and Stockholm don't need to be removed. it's just the imagery information updating when iD has been rebuilt. 
You can see in other PR's such as #5397, that it's just left in.

data/presets/fields/to.json
```diff
@@ -0,0 +1,5 @@
+{
+    "key": "to ",
```
bhousel:
```suggestion
    "key": "to",
```

bhousel: This has an extra space

bhousel:
> The changes about Helsingborg and Stockholm don't need to be removed. it's just the imagery information updating when iD has been rebuilt.

Agree - this is just different people having different versions of ELI, which we can't help.  It all sorts itself out when I make a release.

bhousel:
Thanks @castriganoj 👍

PR #5413, opened 2018-10-19
Adding a new point on a way will add a vertex, not a standalone point

### LINKED ISSUE #5409: Adding a new point on a way should add a vertex, not a standalone point
I'm not sure if the search index is automatically populated from existing data or if there's a hardcoded list somewhere, so feel free to close this if it makes no sense. However I recently, when demo'ing iD, wanted to add a gate, entered "gate" in the search field, and all it came up with was an airport gate. I had to enter "barrier" which is a bit un-intuitive to those without OSM exposure, and then I could select "gate" in the drop down. Could "gate" be added to the first-class search index?

quincylvania:
Closes #5409.


bhousel:
Nice, thanks @quincylvania 👍

PR #5422, opened 2018-10-21
Style unpaved Taxiway

### LINKED ISSUE #5419: Style unpaved Taxiway similarly to unpaved highways
Some rural airports have unpaved taxiways. Their styling should indicate this.

Example: https://www.openstreetmap.org/edit#map=18/43.34154/-114.79322

screen shot 2018-10-19 at 4 25 19 pm

Note: unpaved runways could also benefit from similar styling, but that change would be more complicated and is not part of this issue.

thefifthisa:
Fixes #5419

### DIFF
--- a/modules/svg/tag_classes.js
+++ b/modules/svg/tag_classes.js
@@ -100,6 +100,13 @@ export function svgTagClasses() {
                 }
             }
 
+            // Style unpaved taxiways
+            if (primary === 'aeroway') {
+                if (t.aeroway === 'taxiway' && t.surface === 'unpaved') {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @thefifthisa !
What you did would work ok, but I'm wondering if it would be easier to just modify the existing `highway` block like:

``` if (primary === 'highway' || primary === 'aeroway') { ```

Because there is some other logic inside that block that would be useful to catch other kinds of unpaved surfaces, like `surface=dirt`.

bhousel:
Merging as-is.. I'll make the `if` change in a followon commit.

PR #5423, opened 2018-10-22
Shapefile custom data

superDoss:
Hey guys, for a long time now i am using ID editor, and would like to add a feature that in my opinion is missing.

Really not big of a deal, just added support for [shp](https://en.wikipedia.org/wiki/Shapefile) files.

one problem though  didn't find a way to test it as i need to read .shp file, and didn't know how to read the files in the testing framework.


modules/svg/data.js
```diff
@@ -18,6 +18,7 @@ import {
 
 import stringify from 'fast-json-stable-stringify';
 import toGeoJSON from '@mapbox/togeojson';
+import {open as shapfileOpen} from 'shapefile';
```
bhousel: Can you fix the spelling here?

modules/svg/data.js
```diff
@@ -326,6 +327,23 @@ export function svgData(projection, context, dispatch) {
         return (new DOMParser()).parseFromString(textdata, 'text/xml');
     }
 
+    function shapeToGeoJson(arrayBufferData,callback){
+        var gj = { type:'FeatureCollection' , features:[]};
+        shapfileOpen(arrayBufferData)
```
bhousel: misspelling here too

modules/svg/data.js
```diff
@@ -345,15 +363,26 @@ export function svgData(projection, context, dispatch) {
...
+                _geojson = ensureIDs(gj);
+                _src = extension + ' data file';
+                drawData.fitZoom();
+            }
+    
+            dispatch.call('change');
         }
 
-        dispatch.call('change');
+        _callback();
```
bhousel: Looks like `_callback()` is always synchronously called here, even in the async case where `shapeToGeoJson` calls it later with actual data. 

Can you rename `_callback` to `done`, and then move this lone call after the `gj=JSON.parse(data)`.  I just think it would be clearer what the code does if the "finalizer" is called only once.

bhousel:
Thanks @superDoss !  This is pretty great.. I do get asked about supporting shapefiles sometimes.

A few other things I notice are that we will _probably_ need to polyfill a few things for IE and maybe Safari support.  (see https://github.com/mbostock/shapefile/blob/master/README.md)
> In-browser parsing of dBASE table files requires TextDecoder, part of the Encoding living standard, which is not supported in IE or Safari as of September, 2016. See text-encoding for a browser polyfill.

Also, I don't know how common it is for shapefiles to be in a crs other than EPSG:4326.  I'm not sure whether we would want to let people know that iD won't do any reprojection on these.  We also don't support the `crs` property in GeoJSON, so it might not be a big deal.

Nakaner:
Are the geometries just shown as a background layer or are they converted into OSM objects to be uploaded?

If the latter is possible, mentioning the [Import Guidelines](https://wiki.openstreetmap.org/wiki/Import/Guidelines) might be a good idea to avoid unnecessary reverts of buggy imports (not converting attributes into proper tags etc.).

bhousel:
> Are the geometries just shown as a background layer or are they converted into OSM objects to be uploaded?

Currently just shown

> If the latter is possible, mentioning the [Import Guidelines](https://wiki.openstreetmap.org/wiki/Import/Guidelines) might be a good idea to avoid unnecessary reverts of buggy imports (not converting attributes into proper tags etc.).

yep that's an important part of #4633 , which will be coming soon

modules/svg/data.js
superDoss: fixed

modules/svg/data.js
superDoss: changed

superDoss:
@bhousel i've fixed all of the things you have noted.

i'm still don't know how to test this as i need to read a shape file, and didn't find a way to do this in phantomjs

superDoss:
> A few other things I notice are that we will _probably_ need to polyfill a few things for IE and maybe Safari support. (see https://github.com/mbostock/shapefile/blob/master/README.md)
> 
> > In-browser parsing of dBASE table files requires TextDecoder, part of the Encoding living standard, which is not supported in IE or Safari as of September, 2016. See text-encoding for a browser polyfill.

About that, because of shape file is build from several files (shp,dbf,prj..) didn't figure out the best way to enforce uploading multiple files that are associated, or how to let the user know he needs to upload several files.

So for now there is no problem with dbf encoding as it is not supported and not being parsed.

quincylvania:
> didn't figure out the best way to enforce uploading multiple files that are associated

One potential workaround for this is to allow users to upload zip files containing the component files of the shapefile. An important drawback of this is the need to validate the contents and handle many potential errors.

superDoss:
@bhousel i've just added a working test but for some unknown reason travis have faild to test on node 6, hope you can check it and merge if everything is ok

PR #5429, opened 2018-10-25
Adds hover-highlighting for relations in the raw membership list

quincylvania:
Also:

- Moves hover-highlighting behavior to its own function
- Hover-highlighting now correctly highlights all members of the target relation

Completes task 1 of #2946. Continuation of #5402 and #5404.

screen shot 2018-10-24 at 10 11 10 pm

@bhousel Adding recursive highlighting for relations meant that the code snippet became too large to keep repeating everywhere, so I moved the highlighting logic to its own file/function. I'd like guidance on the naming and location of this file.


bhousel:
> @bhousel Adding recursive highlighting for relations meant that the code snippet became too large to keep repeating everywhere, so I moved the highlighting logic to its own file/function. I'd like guidance on the naming and location of this file.

I'd put the highlighting code in https://github.com/openstreetmap/iD/blob/master/modules/util/util.js
or in its own file under `util`.  (maybe reusing some of the code in `util.js` is there already).

The convention we follow is if the function is named `utilSomethingSomething`, the file should be `util/something_something.js`

👉  Also, relations are tricky and can be circular (a relation can contain itself or an ancestor as a child), so it's really best to avoid recursing.  e.g. `utilGetAllNodes` uses a `seen` variable to avoid infinite looping:  https://github.com/openstreetmap/iD/blob/master/modules/util/util.js#L44

quincylvania:
@bhousel Thanks for the tip on circular relations! I hadn't considered those were possible. I made some changes, let me know if you have further suggestions. Otherwise it's ready.

PR #5442, opened 2018-10-27
Makes the styling of paths with location=underwater the same as location=underground

quincylvania:
[location=underwater](https://taginfo.openstreetmap.org/tags/location=underwater#overview) is a lightly-used yet useful tag that could benefit from styling indicating that the features are not visible in imagery.

Example showing both underwater and overhead `power` features with the new styling:

screen shot 2018-10-27 at 10 53 40 am


bhousel:
Thanks! 👍

PR #5443, opened 2018-10-28
Resizable Sidebar

### LINKED ISSUE #3447: Resizeable / Responsive sidebar
for many tags the space is not sufficient to show the content (in the all tags view). See screenshot: 
screen shot 2016-09-23 at 12 30 18

Why not allow users to resize the left column?

quincylvania:
**Changes**
- The sidebar can be now be resized by dragging the divider (closes #3447)
- Dragging the sidebar below a minimum size collapses it
- A button in the top toolbar lets you uncollapse the sidebar, complete with icon and tooltip
- Double-clicking the divider toggles the sidebar collapse
- Collapse-toggling is animated
- Selecting a feature or ending an add-feature mode auto-opens the sidebar
- Photoviewer UI has been moved from init.js to photoviewer.js
- The photoviewer now has a slight margin to set it away from the sidebar
- The undo/redo, save, and progress indicator controls are now floated to the right
- The save count has been redesigned to visually integrate with the save button
- When the sidebar is collapsed, the mode buttons are centered
- Full right-to-left layout support for all of the above

**Uncollapsed**
screen shot 2018-10-27 at 5 16 36 pm

**Collapsed**
screen shot 2018-10-27 at 5 16 52 pm

**Resize and auto-uncollapse demo**
full screen demo 1

**Note**
As you can see above, making the sidebar resizable took me somewhat beyond the scope of the original issue. Namely, I thought we could make better use of space in the top bar now that it can take up the full width of the window, so I moved some buttons to the right side. This caused the save count to be cut off, so I redesigned it to be incorporated within the button rather than outside it. I'm open to exploring other solutions if these changes are controversial.

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -1,11 +1,16 @@
 /* photo viewer div */
 #photoviewer {
     position: absolute;
-    bottom: 30px;
+    bottom: 40px;
+    left: 10px;
     width: 330px;
     height: 250px;
     padding: 5px;
     background-color: #fff;
+}
+[dir='rtl'] #photoviewer {
+    left: auto;
+    right: 10px;
 }
 
 @media screen and (min-width: 1600px) {
@@ -345,4 +350,3 @@ label.streetside-hires {
     background: rgba(0,0,0,0.85);
     color: #fff;
 }
-
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -83,17 +83,9 @@ body {
     margin-right: 10px;
 }
 
-.limiter {
-    position: relative;
-    max-width: 1200px;
-}
-
 .spinner {
     opacity: .5;
-    float: right;
-}
-[dir='rtl'] .spinner {
-    float: left;
+    display: inline-block;
 }
 
 .spinner img {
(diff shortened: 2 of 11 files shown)


bhousel:
This looks great @quincylvania 🎉 
I love that you took the time to do separate LTR and RTL icons.

It looks like the loading spinner might be gone?  
Aside from that I didn't see any major issues with the code, so let's just merge as-is and we can tweak it if we find any issues.

quincylvania:
@bhousel Thanks!! The spinner _shouldn't_ be gone. I moved it to the left of the undo/redo buttons. I did add `display: none` to the fullscreen button so it won't take up space in the layout and since it wasn't shown anyway.

FYI: The reason I added the `wip` tag was that I wanted to give a little more attention to the top bar buttons, likely implementing #4356, but you're right that can be a future change.

slhh:
I got the following issue with the current Firefox on Windows 10:
When making the sidebar very wide, iD gets into a state where undo and save buttons are out of view, even after making the sidebar smaller again. When the sidebar is collapsed, the buttons are fully visible. When the sidebare is very small, the buttons are partially visible.
This state is somewhat permanent. Restarting iD wasn't sufficient. I had to clear the browser cache to clear the error state.

I wasn't able to reproduce the error with Edge.

Btw, the delete and info buttons in the all tags section should not scale with the sidebar width to provide more space for tag values.

Issue #5415, opened 2018-10-19
Minimap enhancement


slhh:
- [ ] Show selected features in the minimap too (https://github.com/openstreetmap/iD/issues/2962#issuecomment-180839678) to support #4128.
- [ ] Show features which are hovered in sidebar lists (e.g. all members, multi-selection) in the minimap. This would help to identify list items (related to #5407)
- [ ] Make the background of the minimap selectable. When working with aerial imagery as background of the main map using the standard OSM map in the minimap might give better navigation support.
We do not need to duplicate the full background selection options of the main map. The selection between using the standard OSM map, the same background as the main map, and maybe a a few more seems to be sufficient.
- [ ] Make the minimap resizeable

TAQ2:
Hello I'm new to this project and would like to contribute

> Make the minimap resizeable

Please could you give me a hint as to where the code for this is?

PR #5446, opened 2018-10-28
Add selected features to minimap

jguthrie100:
This is a work in progress for #5415 

It simply copies any selected features from the main map and replicates them on the minimap. The styles etc are all maintained (which will require some editing most likely as it can look a bit cumbersome, but that can all be fixed)

Multiple features can be selected and they'll all be replicated on the minimap without an issue.

Todo:
 - Modify styles so they are less intrusive
 - Add breathing (i.e. red pulsing on selected features) to minimap features (?)
 - Improve styling of Point markers (currently a plain red dot - maybe better if its a small black dot with breathe effect)
 - Improve relation/area selections (either a big red blob or a thin line atm)

Note that this doesn't fix the selection dropping when the selection is dragged off the screen. I'm looking into sorting that, but its not too easy!

image

### DIFF
--- a/css/55_cursors.css
+++ b/css/55_cursors.css
@@ -6,6 +6,7 @@
 }
 
 .map-in-map,
+.map-in-map #map-in-map-features .map-in-map-selection, /* extra specific to over-ride later assignments */
 #map {
     cursor: auto; /* Opera */
     cursor: url(img/cursor-grab.png) 9 9, auto; /* FF */
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -3259,6 +3259,17 @@ img.tile-debug {
     stroke-width: 5;
 }
 
+.map-in-map-selection {
+    stroke: #f60;
+    fill: #f90;
+    stroke-width: 3;
+}
+
+.map-in-map-selection.line {
+    fill: none;
+    stroke-width: 5;
+}
+
 
 /* Debug Data
 ------------------------------------------------------- */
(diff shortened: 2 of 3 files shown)


jguthrie100:
It might be worth gathering some suggestions for styling too, and whether the entire feature should be drawn, or whether it should just be a dot at the centre point of the feature for example.

slhh:
I wouldn't use any halo or even breathing. 
The entire feature should be drawn, a centerpoint can be quite confusing. 
The width of the yellow line on the minimap seems to be fine.
Selected features can easily get overlapping on the minimap due to a different zoom levels. Therefore, do not use feature type specific styles.
I'd use four colors/styles:
1. Members of selected relations
2. Selected ways
3. Selected nodes
4. Selected features, which are hovered on the main map or on a sidebar list.

bhousel:
Yes, along the lines of what @slhh said - I would avoid extensive map styling here too.  A simple stroke is enough to show it on the minimap.

I also noticed you're doing a lot of code to generate the GeoJSON that gets displayed.  Each of the entity types `node`, `way`, and `relation` has a `.asGeoJSON()` function.  So could you just do something like:

```js
var resolver = context.graph();
var features = context.selectedIDs()
    .map(context.entity)
    .map(function(entity) { return entity.asGeoJSON(resolver); });

var gj = { 
    type: 'FeatureCollection',
    features: features
};
...
```

jguthrie100:
I've made some changes to the styling and taken out a lot of redundant code. I think the failing tests are completely unrelated to my changes, but not sure..

image

Again, could do with advice on styling/colour/stroke-width choices etc

quincylvania:
> I think the failing tests are completely unrelated to my changes, but not sure..

@jguthrie100 We've been seeing some server-side tests fail lately due to timeouts, so as long as the local tests pass on http://localhost:8080/test/ then don't worry about it.

PR #5447, opened 2018-10-28
Added preset for attraction=summer_toboggan

hikemaniac:
This PR adds a preset and rendering for [attraction=summer_toboggan](https://wiki.openstreetmap.org/wiki/Tag:attraction%3Dsummer_toboggan). However I'm currently not able to use npm and I therefore just handwrote all the files, that would normally be built. I hope that this is okay.


data/presets/presets/attraction/summer_toboggan.json
```diff
@@ -0,0 +1,18 @@
...
+    ],
+    "geometry": [
+        "line"
+    ],
+    "terms": [
+        "alpine slide",
+        "mountain coaster"
+    ],
+    "tags": {
+        "attraction": "summer_toboggan"
```
bhousel: Reading on that wiki page ... does it also need `leisure=track` here?

hikemaniac: I'm not sure about that. The wiki indeed suggests it but a quick [overpass turbo query](http://overpass-turbo.eu/s/Dd0) shows that 460 of 715 ways do not have the leisure key present at all (and this doesn't neccesarily mean, that the remaining 255 objects leisure tag is `leisure=track`).

bhousel: ok, yeah seems fine to me to leave it off.  [`attraction=water_slide`](https://wiki.openstreetmap.org/wiki/Tag:attraction%3Dwater_slide) seems like almost the same thing, and doesn't say anything about adding extra tags to it.

Issue #4128, opened 2017-07-05
Avoid deselecting feature that goes out of view


1ec5:
iD automatically deselects a feature when it goes out of view. (As far as I can tell, this has always been the case.) iD should avoid deselecting features for this reason, or it should at least be more lenient. For example, it’s currently very difficult to relate a place POI to a boundary relation unless the place is very small in area. (By analogy, I find it annoying to use a text editor that insists on moving the insertion point when I scroll up and down.)

On the other hand, it could look weird if the user had accidentally panned the feature out of view. The tag editor would be open with no clear target. Perhaps there could be an “off-screen” indicator, sort of like those “distance to” arrows at the edges of paper maps, plus a keyboard shortcut to jump to the selection. Or maybe #3622 could be fixed instead. 😉

/ref https://github.com/openstreetmap/iD/issues/3622#issuecomment-263321927

HolgerJeromin:
Same with zooming out

BjornRasmussen:
I really don't see this as an issue, since the multiselected objects don't get deselected when you move the map, and that is when you would really need this feature.

1ec5:
The case where I really need this feature is when I’m piecing together a route relation where the nearest existing member is clear across town. (Most recently, this happened as I pieced together a bike route relation from spotty signage.) One of #3622, #3487, or this issue would make my workaround unnecessary:

1. Select a new bus stop, and open the Measurement panel. Jot down the coordinates.
1. In a separate tab, open osm.org’s page for the bus route relation or one of its existing members.
1. Select the existing member by searching for e.g. `w123` or `n123`.
1. With the existing member in view, draw a new line and add it to the relation without giving it any tags.
1. Begin appending to the line.
1. Without ending the line, search for the coordinate you jotted down in the first step.
1. Extend the tagless line to near the new bus stop.
1. Add the bus stop to the route relation.
1. Delete the tagless line.

quincylvania:
@1ec5 How do you reproduce this? I tried clicking on a feature then panning far away and it stayed selected. Safari and Chrome.

Nakaner:
I haven't digged through the history in this bug tracker but deseselecting features when going out of view could preserve us from people accidentally deleting or modifying objects.

jguthrie100:
Might be worth popping up a warning if something out of view will get deleted or edited

1ec5:
> How do you reproduce this? I tried clicking on a feature then panning far away and it stayed selected. Safari and Chrome.

[...]

PR #5455, opened 2018-11-02
Flexbox buttons in top bar

### LINKED ISSUE #4356: Flexbox the buttons

bhousel:
closes #4356

- Makes the "Sidebar" toggle button permanent but removes the label
- Did some things to the "Save" button to make it the same width whether there
  is a count or not (prevents the buttons from jumping when pressing undo/redo)
- Removes a lot of the floated col rules that aren't used much anymore

flexbox buttons

I'd like to figure out collapsing away the labels and leaving only icons on very narrow screens.  
(I know I can use a media query for this)

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -569,15 +569,17 @@ button.secondary-action:hover {
     display: flex;
     flex: 0 1 auto;
     flex-flow: row nowrap;
+    width: 100%;
 }
 .tool-group.leading-area {
+    flex-shrink: 2;
     justify-content: flex-start;
 }
 .tool-group.center-area {
     justify-content: center;
 }
 .tool-group.trailing-area {
-    justify-content: flex-end;
+    justify-content: flex-start;
 }
 
 .tool-group > div {
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -16,7 +16,6 @@ body {
         sans-serif;
     margin: 0;
     padding: 0;
-    min-width: 768px;
     color: #333;
     overflow: hidden;
     -ms-user-select: none;
@@ -32,7 +31,6 @@ body {
 .id-container {
     height: 100%;
     width: 100%;
-    min-width: 768px;
 }
 
 #content {
@@ -606,6 +604,19 @@ button.add-note svg.icon {
     transform: scaleX(-1);
     filter: FlipH;
     -ms-filter: "FlipH";
+}
+
+
+#bar.narrow .tool-group {
+    width: unset;
+}
+#bar.narrow .spinner,
+#bar.narrow button .label {
+    display: none;
+}
+#bar.narrow button .count {
+    border-left-width: 0;
+    border-right-width: 0;
 }
 
 
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4585,19 +4585,6 @@ li.hide + li.version .badge .tooltip .tooltip-arrow {
 }
 
 
-/* Media Queries
-------------------------------------------------------- */
-@media screen and (max-width: 1200px) {
-    .user-list { display: none !important; }
-}
-@media screen and (max-width: 1000px) {
-    #userLink { display: none !important; }
-}
-@media screen and (max-width: 900px) {
-    #scale-block { display: none !important; }
-}
-
-
 /* Scrollbars
  ----------------------------------------------------- */
 ::-webkit-scrollbar {
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -61,6 +61,7 @@ body {
     #content > #bar {
         width: 100vw;
     }
+/*
     #content.inactive > #bar > .spacer.col4 {
         width: 0px;
     }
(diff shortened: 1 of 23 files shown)


bhousel:
Flexbox is also nice because most things Just Work when right-to-left. 

flexbox buttons rtl

quincylvania:
Good call on removing the sidebar label and adding a shortcut! A couple notes:

- I assume the constant-width Save button works with longer localized text?
- When condensed, it looks like the margin between the sidebar button and the mode buttons is smaller than the margin between the other buttons.
- We might consider moving the spinner out of the top bar and recovering that space. The spinner is the only non-button up there anyway.

slhh:
I think a button to collapse the sidebar would be more intuitive if it were placed within the sidebar.

Editing with collapsed sidebar is somewhat dangerous because users might not be aware which feature they are really editing. To mitigate the risk, when the sidebar is collapsed, place the feature type name and icon of the selected featureas a button in the topbar (in the column where the sidebar was). This can be the button to uncollapse the sidebar. Having to click on the feature type to open the sidebar to edit the feature properties seems to be intuitive.

When multiple features are selected, the uncollapsing button might contain a selection summary like "10 nodes, 3 way, 2 relations" (but with words replaced by icons).

quincylvania:
@slhh When I was implementing the sidebar collapse I tried making the "open sidebar" button preview the selected feature, much like you describe, but it turned out to be a larger design challenge than expected. It might be worth pursuing in the future.

For now, the sidebar automatically opens whenever the user selects a feature, meaning the only way they can edit a feature with the sidebar closed is if they close it _after_ they select it. In my estimation, this makes accidental edits unlikely.

bhousel:
btw I changed the sidebar resizer a bit today.. It keeps the map still now, which I find a lot easier to use. 

keep map still

quincylvania:
@bhousel That's a great improvement! It had been visually unwieldy.

bhousel:
Although now this can happen 🤔 

sidebar hides point

bhousel:
sorry @quincylvania I just realized I forgot to answer your questions!

> * I assume the constant-width Save button works with longer localized text?

Yes!.. It's actually tricky.  The count span has a larger min-width to hold up to about a 3 digit number, but the label can grow if the word is longer.  When the save count goes away, the label jumps to the right slightly.. It's opaque so hard to see, but that's ok.  The important thing is that the undo/redo buttons don't jump!

save button

> * When condensed, it looks like the margin between the sidebar button and the mode buttons is smaller than the margin between the other buttons.

Good noticing!  I think it's because the sidebar button is just sitting by itself, and the other groupings have another div with padding there.  I will fix it 👍 

> * We might consider moving the spinner out of the top bar and recovering that space. The spinner is the only non-button up there anyway.

Yes, now with things more responsive, it looks a little weird to reserve space for it when it's not there, and it's not something the user can interact with.  Maybe we should put it at the bottom near the status-y things.

bhousel:
I was so annoyed at the save button changing its width and moving the undo/redo buttons that I recorded an image of that yesterday too 😆 

jumpy save

jguthrie100:
> Although now this can happen 
> 
> sidebar hides point

This would be fixed by #4128 (or you could pan only if the feature falls under the sidebar)

bhousel:
> (or you could pan only if the feature falls under the sidebar)

I'll see how this option feels.. It sounds like a good idea.

quincylvania:
> I was so annoyed at the save button changing its width and moving the undo/redo buttons that I recorded an image of that yesterday too 😆

Oh that is annoying. I didn't notice it during my implementation. Thanks for the fix!

> Yes, now with things more responsive, it looks a little weird to reserve space for it when it's not there, and it's not something the user can interact with. Maybe we should put it at the bottom near the status-y things.

Sounds like a good solution to me.

bhousel:
> Oh that is annoying. I didn't notice it during my implementation. Thanks for the fix!

It wasn't in your implementation - it only started once I switched them to flexbox and the buttons could resize more.

bhousel:
I've added a function to check the toolbars for overflow, and drop elements that we don't need on narrow screens (like the button labels and loading spinner).. 

responsive top bar

bhousel:
I just added some code to address https://github.com/openstreetmap/iD/pull/5455#issuecomment-435557142 - so the sidebar will not obscure selected things when it expands.  This works for OSM stuff (entities and notes) as well as external data.

sidebar avoid entity

This is good to merge now!

PR #5459, opened 2018-11-03
Use the green stroke color to render barrier=hedge

quincylvania:
This small change differentiates hedges from other barrier features.

screen shot 2018-11-02 at 7 37 31 pm


bhousel:
Nice!  Feel free to just merge stuff like this 👍

quincylvania:
@bhousel Thanks! I'm still feeling things out so sometimes even on small things I think, "maybe there's a good reason against this change I haven't thought of, better get a second opinion." But of course small things are easy enough to revert.

PR #5461, opened 2018-11-03
tag values with wiki page may be documented as bad

### LINKED ISSUE #5460: Stop treating all tag values with OSM Wiki page as recommended
tl;dr: [OSM Wiki page describing highway=unsurfaced as an outdated tagging scheme](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dunsurfaced) is causing iD to propose this tag to mappers. This is undesirable.

highway unsurfaced in id

-----

There are some tags that were used in past but nowadays are almost unused and should not be used ever again. Part of handling this situation is to create page on OSM wiki describing why tag is problematic and to mention possible alternatives.

It leads to one unfortunate side effect. As I undestand, in raw tag editing mode, iD proposes tags that are either with some minimal usage or have their own OSM wiki page.

https://github.com/openstreetmap/iD/blob/master/modules/services/taginfo.js#L93

So tags like

https://wiki.openstreetmap.org/wiki/Tag:noexit%3Dno
https://wiki.openstreetmap.org/wiki/Tag:natural%3Dland
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dunsurfaced
https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dwood
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Deducation
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dabandoned
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dproposed
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddisused
https://wiki.openstreetmap.org/wiki/Tag:highway%3Dbyway
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dhotel
https://wiki.openstreetmap.org/wiki/Tag:natural%3Dmoor

and other will be proposed by iD for use even after their usage reaches 0.

Is there any situation where almost unused tag should be promoted just because it has its own OSM wiki page? I think that only completely failed proposals will fail to get `>0.0%` usage, and promoting them is not necessary.

I propose to stop considering "has wiki page" as sufficient reason to suggest tags and rely just on Taginfo usage.

It is related to #4508 that proposed much stronger and more controversial change.

matkoniecz:
fixes #5460 

warning: this proposed code change is not tested

### DIFF
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -90,7 +90,7 @@ function filterValues(allowUpperCase) {
     return function(d) {
         if (d.value.match(/[;,]/) !== null) return false;  // exclude some punctuation
         if (!allowUpperCase && d.value.match(/[A-Z*]/) !== null) return false;  // exclude uppercase letters
-        return parseFloat(d.fraction) > 0.0 || d.in_wiki;
+        return parseFloat(d.fraction) > 0.0;
     };
 }
 
--- a/test/spec/services/taginfo.js
+++ b/test/spec/services/taginfo.js
@@ -85,7 +85,7 @@ describe('iD.serviceTaginfo', function() {
             );
         });
 
-        it('includes unpopular keys with a wiki page', function() {
+        it('does not include unpopular keys with a wiki page', function() {
             var callback = sinon.spy();
             taginfo.keys({query: 'amen'}, callback);
 
@@ -96,10 +96,7 @@ describe('iD.serviceTaginfo', function() {
             );
             server.respond();
 
-            expect(callback).to.have.been.calledWith(null, [
-                {'title':'amenity', 'value':'amenity'},
-                {'title':'amenityother', 'value':'amenityother'}
-            ]);
+            expect(callback).to.have.been.calledWith(null, [{'title':'amenity', 'value':'amenity'}]);
         });
 
         it('sorts keys with \':\' below keys without \':\'', function() {
@@ -220,7 +217,7 @@ describe('iD.serviceTaginfo', function() {
             expect(callback).to.have.been.calledWith(null, []);
         });
 
-        it('includes unpopular values with a wiki page', function() {
+        it('does not include unpopular values with a wiki page', function() {
             var callback = sinon.spy();
             taginfo.values({key: 'amenity', query: 'par'}, callback);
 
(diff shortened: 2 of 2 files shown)


matkoniecz:
I attempted to invert tests covering inverted behaviour.

matkoniecz:
It is still failing with

```
  1) iD.serviceTaginfo
       #keys
         does not include unpopular keys with a wiki page:
     expected spy to have been called with arguments null, [{ title: "amenity", value: "amenity" }]
null
[{ title: "amenity", value: "amenity" }, { title: "amenityother", value: "amenityother" }] [{ title: "amenity", value: "amenity" }] 
  AssertionError@file:///home/mateusz/Desktop/tmp/iD/node_modules/chai/chai.js:9449:24
  assert@file:///home/mateusz/Desktop/tmp/iD/node_modules/chai/chai.js:239:31
  file:///home/mateusz/Desktop/tmp/iD/node_modules/sinon-chai/lib/sinon-chai.js:105:24
  methodWrapper@file:///home/mateusz/Desktop/tmp/iD/node_modules/chai/chai.js:7824:30
  file:///home/mateusz/Desktop/tmp/iD/test/spec/services/taginfo.js:99:53
```

I admit that I am not sure what this test is really reporting.

matkoniecz:
Note to self: it is described in https://github.com/openstreetmap/iD#building-id

Note to self: `Open http://localhost:8080/ in a web browser` is missing in https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#building--installing (and ` test runner page http://localhost:8080/test` seems to be missing in both) - fixing that may be a good PR.

bhousel:
I cleaned up the tests in a29af100d

What I did was:
- remove the test for unpopular _values_ with a wiki page
- restore the original test for unpopular _keys_ with a wiki page

This matches what the code does now, and I think the intent of #5460 

The code doesn't care about wiki pages for _values_ anymore, so there is no need to test that behavior. The code continues to give a boost to _keys_ with wiki pages, so we do test that.

matkoniecz:
Thanks for merging and fixing tests! Hopefully my next PR will go even better :)

bhousel:
> Thanks for merging and fixing tests! Hopefully my next PR will go even better :)

Thanks!  This PR was great, no worries! 👍

matkoniecz:
got reverted in https://github.com/openstreetmap/iD/commit/a2cacaaf2456de4c8443a7b5dab0ac3393be5deb

fixed again https://github.com/openstreetmap/iD/pull/11794

PR #5468, opened 2018-11-06
Expanded undo messages for node-dragging scenarios

### LINKED ISSUE #1252: dragging a point to line says "connected a way to another"
have a line and a tagged point. drag point to a vertex/node of the line - undo history says "connected a way to another".
dragging it to a segment of the line says "connected a way to line" (and probably similar for area :) )

quincylvania:
Adds undo messages for connecting:
- Points to points, vertices, lines, and areas
- Vertices to sibling and adjacent vertices (nodes with the same parent way)

Closes #1252.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -118,10 +118,18 @@ en:
       annotation: Removed a member from a relation.
     connect:
       annotation:
-        point: Connected a way to a point.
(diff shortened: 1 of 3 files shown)


bhousel:
LGTM 👍

Issue #5469, opened 2018-11-06
Add preset for leisure=fishing


bhousel:
I'm kind of surprised we don't have a preset for a "Fishing Spot" yet 🎣 
https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfishing

quincylvania:
Recommendations
- Name: Fishing Spot
- Geometry: `point`, `area`
- Fields: `name`, `access_simple`, …
- Terms: `angler`, …



@bhousel You beat me to it!

bhousel:
> @bhousel You beat me to it!

Yeah I added this issue reel quick

iandees:
omg.

>

xmile1:
> I'm kind of surprised we don't have a preset for a "Fishing Spot" yet 🎣
> https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfishing

Hi, I can take this up, Its similar to [this](https://github.com/openstreetmap/iD/pull/4432/files) before I believe that i did before i believe.

bhousel:
sounds good @xmile1 !    you can copy an existing file like `data/presets/presets/leisure/bird_hide.json` -> `data/presets/presets/leisure/fishing.json` and change the contents.

PR #5476, opened 2018-11-08
Add preset for leisure=fishing

xmile1:
This PR creates a preset for leisure=fishing
https://github.com/openstreetmap/iD/issues/5469

screen shot 2018-11-09 at 12 35 06 am


data/presets/presets/leisure/fishing.json
```diff
@@ -0,0 +1,18 @@
+{
+    "icon": "aquarium",
```
xmile1: I am not certain this is the best icon here, perhaps a new icon needs to be created.

data/presets/presets/leisure/fishing.json
```diff
@@ -0,0 +1,18 @@
+{
+    "icon": "maki-aquarium",
```
xmile1: I am not certain this is the best icon here, perhaps a new icon needs to be created.

bhousel: I agree - you can use [`fas-fish`](https://fontawesome.com/icons/fish?style=solid) for this..

bhousel:
Looks great, thanks @xmile1 👍

PR #5479, opened 2018-11-10
Layer adjustments

### LINKED ISSUE #5257: Lines from custom data files block clicking ways underneath them
Im tracing trails from gpx files. When a node from an existing way is under a line from a gpx file it is not possible to select the node when making a new way.

This is a regression in 2.11.0

bhousel:
This adjusts the stacking order so that the interactive OSM objects take precedence over the interactive non-OSM objects.

fixes #5257

Still need to test a bit more..

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -66,7 +66,6 @@
 
 
 /* points & notes */
-
 g.note .stroke {
     stroke: #222;
     stroke-width: 1;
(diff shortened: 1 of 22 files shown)


bhousel:
This is good enough for now.  

I did also attempt to remove the supersurface and instead set separate transforms for each layer, but doing it that way ended up being worse for performance.  

If each layer could have a separate transform, they could render during async idle callback, instead of all together on the main thread.  

I dug a bit into performance and there are big issues with style recalculation and repainting that I have no idea how to fix. 😞

screenshot 2018-11-12 15 01 34

iandees:
Maybe it's ... 

image

PR #5480, opened 2018-11-10
Updates presets to use new icons from Maki 5.0.0

quincylvania:
Includes usage bridge, waterfall, shoe, watch, casino, and car rental icons and others


bhousel:
Looks good!
Also I think there are a few other new icons that didn't make it into the CHANGELOG.
https://github.com/mapbox/maki/commit/e71bcd950d719474e3b4720810fd5dd3d45744d9

What do you think about making presets for 
- Seafood Restaurant
- Noodle Restaurant
- Pizza Restaurant
- others?

These would just be sub-presets under restaurant that also set the `cuisine` field.

We don't have to be limited to what icons are available.  Maybe we could do a few of [the top cuisine values from taginfo](https://taginfo.openstreetmap.org/keys/cuisine#values).

quincylvania:
@bhousel Thanks for the tip on the other presets. I hadn't looked for additions outside the changelog!

I added presets for the common cuisines. There is room for confusion since `cuisine` is often used on both `amenity=fast_food` and `amenity=restaurant`. To try to disambiguate, the names of the `fast_food` ones use "Shop" instead of "Restaurant", like "Burger Shop". `pizza` is the only `cuisine` for which I made two presets since [both are well used](https://taginfo.openstreetmap.org/tags/cuisine=pizza#combinations) and I think the difference is commonly understood, at least in the U.S.

Feel free to make changes and merge if it looks good.

Issue #5452, opened 2018-10-31
Implement new tagging scheme for fire hydrants


momu:
There was a [successful proposal for improving the hydrants tagging][0]. [According to the wiki][1] there are some changes regarding `emergency=fire_hydrant`. E.g. `fire_hydrant:type=pond` is deprecated, `fire_hydrant:type=pipe` is new.

Can someone add it?

[0]: https://wiki.openstreetmap.org/wiki/Proposed_features/Fire_Hydrant_Extensions_(part_2)
[1]: https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_hydrant

momu:
I just started working on it: 

https://github.com/momu/iD/tree/fire-hydrants

If anybody is faster feel free to join.

bhousel:
This happened

PR #5488, opened 2018-11-14
Fire hydrants

momu:
Solves #5452.

Implements current [emergency=fire_hydrant tagging scheme][0]. Current scheme is based on [approved Fire Hydrant Extensions (part 2) proposal][1].

[0]: https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_hydrant
[1]: https://wiki.openstreetmap.org/wiki/Proposed_features/Fire_Hydrant_Extensions_(part_2)


bhousel:
Whoa this is a lot of stuff!   Can we trim it down to just the 5 or so most important and easiest fields that an untrained mapper could fill in?   We don't really add fields to iD that require specialized knowledge or training for someone to contribute.  (But things like where the hydrant is and the color of the cap seems fine)

fire hydrants

momu:
> Can we trim it down to just the 5 or so most important and easiest fields that an untrained mapper could fill in? 
Sure,I will have a look which parts are useful and easy to map.
Is it enough to remove the unnecessary from the preset (fire_hydrant.json) or do I have to remove the 
fields json-files also?

bhousel:
> do I have to remove the fields json-files also?

yes, please remove all the fields that aren't being shown anywhere.

PR #5491, opened 2018-11-16
Forest and wetland sub-type fill patterns

RudyTheDev:
This adds pattern fill images for forest and wetland sub-types (similar to the main map) to encourage tagging them. Specifically, it detects most common combinations like `landuse=forest; leaf_type=broadleaved` or `natural=wetland; wetland=marsh`.

Forest/wood:
forest patterns

Wetland:
wetland patterns

This changes how patterns are defined (since I need to match custom tag-values). This also stops incorrect tag/value pairs, such as `landuse=sand` or `amenity=wood` from being matched.


bhousel:
> This changes how patterns are defined (since I need to match custom tag-values). This also stops incorrect tag/value pairs, such as `landuse=sand` or `amenity=wood` from being matched.

Thanks, looks good @RudyTheDev 👍 

It's getting a bit complicated but that's ok for now.  Do you think it makes sense to eventually move the fill patterns into the preset definitions instead of having all the rules defined in `areas.js`?  I don't have a preference, just wondering what you think.

RudyTheDev:
@bhousel: My natural coding impulse is also to have these in presets. But I've done very little work in JavaScript so far. I might do a PR for this when I figure it out. (Having edited in iD for some 4 years, I definitely have a long wishlist, but not yet the knowledge of the project and JS.)

I'm not sure what sort of direction is best for fill patterns. For starters, I just wanted the functionality to add new ones. I'd like to add more, because few things motivate mappers more than pretty feedback. At the same time, prominent additions could mess up the visual language. I'll probably come up with a bunch of less-than-uncontroversial ideas and make a (proposal) issue for feedback.

PR #5496, opened 2018-11-17
Fire hydrants shortened

momu:
Solves #5452.

Implements current [emergency=fire_hydrant tagging scheme][0]. Current scheme is based on [approved Fire Hydrant Extensions (part 2) proposal][1].

As a result of discussion in #5488 only the most important basic  tags (easy to be mapped) are added:
  
* fire_hydrant:type
* fire_hydrant/pressure
 * fire_hydrant/diameter
* water_source
* survey/date
* fire_hydrant/position
* water_volume
* colour
* ref
* name

[0]: https://wiki.openstreetmap.org/wiki/Tag:emergency%3Dfire_hydrant
[1]: https://wiki.openstreetmap.org/wiki/Proposed_features/Fire_Hydrant_Extensions_(part_2)


quincylvania:
Hi @momu! Looks like this is coming along well. A couple notes:

- It looks like `survey:date` represents the last time a mapper surveyed the feature. Since it isn't specifically important to fire hydrants, I wouldn't recommend showing it by default. However, it'd be great if you could add it as a universal field, since then people could find it in the "Add Field" section and add it to any feature, including hydrants!
- I'm not sure `name` is necessary by default (I know I couldn't name a single fire hydrant in my area 😄) so maybe remove that from the default list and then mappers can find it with "Add Field" as necessary.

bhousel:
thanks @momu  - I simplified the labels a bit to better match the other fields in iD, and added `couplings` back (which casual mappers can see and understand)..

momu:
Thanks for merging @bhousel.
I agree with @quincylvania regarding the `name` and `survey:date` considerations. 
If time permits I would change it as he mentioned it. What do you think about it?

bhousel:
> If time permits I would change it as he mentioned it. What do you think about it?

I already removed `name` and left in `survey:date`.  I'm ok with it 👍

PR #5502, opened 2018-11-19
Redraw restrictions panel when dragging sidebar

### LINKED ISSUE #5474: Turn restriction editor should redraw when the sidebar resizes
turn restriction resize

related #4844

jguthrie100:
Fixes #5474 

I've put together a bit of a hacky fix for the restrictions editor not scaling properly with the sidebar being dragged.

Ideally, I'd call the `redraw()` method direction from the sidebar's drag handler, but can't figure out a way to get access to it efficiently, so instead I just add a `dragging` class when the drag starts and remove it when it stops.

The restrictions.js file then just sets a `setInterval` that checks whether the `dragged` class is present or not, and then redraws the panel every 50ms if it is.

The redraw interval can be changed easily enough (is 50ms too frequent?), but when dragging, the street graphics are chopped off so close to the border of the svg, that when dragging, the chopped-off street edge is still visible for a few ms until the panel redraws.

The only way I can think of getting around that is to render the junction so that the streets are initially drawn much further out past the svg border, meaning when the svg resizes, there is much more 'buffer' space before the street-end is reached. (if that makes sense?)

### DIFF
--- a/modules/ui/fields/restrictions.js
+++ b/modules/ui/fields/restrictions.js
@@ -66,6 +66,8 @@ export function uiFieldRestrictions(field, context) {
     var _intersection;
     var _fromWayID;
 
+    var _redrawHandler;
+
 
     function restrictions(selection) {
         _parent = selection;
@@ -301,6 +303,14 @@ export function uiFieldRestrictions(field, context) {
                 });
         }
 
+        if (!_redrawHandler) {
+            _redrawHandler = setInterval(function() {
+                if (d3_select('#sidebar-resizer').classed('dragging')) {
+                    utilSetDimensions(_container, null);
+                    redraw();
+                }
+            }, 50);
+        }
 
         // This can happen if we've lowered the detail while a FROM way
         // is selected, and that way is no longer part of the intersection.
@@ -661,6 +671,8 @@ export function uiFieldRestrictions(field, context) {
 
         d3_select(window)
             .on('resize.restrictions', null);
+
+        clearInterval(_redrawHandler);
     };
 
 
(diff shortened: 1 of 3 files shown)


jguthrie100:
I've got a laggy computer, but shows the general idea..

image

quincylvania:
Hi @jguthrie100! Thanks for working on this. It would be nice if it only responded during resizing. Otherwise the event is constantly firing. You may want to look at the [`onResize`](https://github.com/openstreetmap/iD/blob/107f669eddb6086662e2aed0ad3f30c93177e1c1/modules/ui/init.js#L393) function that's called every time the window or the sidebar is resized, or the sidebar [drag handler](https://github.com/openstreetmap/iD/blob/107f669eddb6086662e2aed0ad3f30c93177e1c1/modules/ui/sidebar.js#L66). It may be nice to add some sort of "sidebar did drag" even that the restriction editor can listen for.

jguthrie100:
I've changed the implementation so that the restrictions panel is redrawn everytime `ui.onResize` is called, however, it means that its called 1000+ times whilst dragging the bar even just a little bit across.

Do you know of any good coding practices to limit the call numbers (other than setInterval or having some rudimentary counter loop?) - unless its fine as is?

jguthrie100:
Improved the algorithm significantly. Now only redraws if the sidebar has moved above a specified threshold (in this case 2px - easily changed)

image

jguthrie100:
> It may be nice to add some sort of "sidebar did drag" event that the restriction editor can listen for.

@quincylvania I'm a bit unsure what you mean by that?

And is there anything I can change to get this merged? The redraw threshold could probably be lowered to 1px if there are concerns about it being a bit jaggy, and I'm assuming most browsers would be able to handle it (considering the entire map is redrawn so frequently whilst dragging around the map!)

bhousel:
Thanks @jguthrie100 & @quincylvania  - this looks pretty good to merge...
I might eventually switch over the custom event dispatch to a more idiomatic D3.js way of doing it, but for now it's cool..

PR #5509, opened 2018-11-20
Visually centers the change count in the save button

quincylvania:
**Before**
screen shot 2018-11-20 at 6 28 07 pm

**After**
screen shot 2018-11-20 at 6 24 35 pm


bhousel:
Nice 👍

PR #5516, opened 2018-11-22
Trackpad scroll-to-zoom while holding Shift or Option/Alt

### LINKED ISSUE #5508: Enable zooming with modifier key + two-finger trackpad scroll
With the recent addition of two-finger panning, it is no longer possible to zoom via scrolling on a trackpad. While pinch-to-zoom works as an alternative, apps such as Apple Maps also allow scroll-to-zoom while holding down Shift or a similar modifier key. iD should adopt this interaction.

quincylvania:
Closes #5508.

I didn't end up piggybacking on the Control key zooming since that has an extra scale multiplier for the pinch-to-zoom.

### DIFF
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -452,16 +452,6 @@ export function rendererMap(context) {
                 x2 = p0[0] - p1[0] * k2;
                 y2 = p0[1] - p1[1] * k2;
 
-            // 2 finger map panning (all browsers) - #5492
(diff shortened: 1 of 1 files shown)


bhousel:
> I didn't end up piggybacking on the Control key zooming since that has an extra scale multiplier for the pinch-to-zoom.

Thanks @quincylvania looks good... 👍

PR #5522, opened 2018-11-24
Add rendering for attraction=water_slide

hikemaniac:
This PR adds special rendering for attraction=water_slide. I've used the color for the water slide stroke from OpenStreetMap Carto and it looks like this:
water slide rendering


bhousel:
Thanks @hikemaniac 👍

PR #5524, opened 2018-11-24
Amend TagInfo API calls to use current locale

### LINKED ISSUE #5523: Translations: Where do the tooltips come from?
I am referring to tooltips like this: 
> Predominantly houses and apartments buildings.

# Where do they come from?

- I cannot find them in the repo https://github.com/openstreetmap/iD/search?q=predominantly&unscoped_q=predominantly
- And not in the translation tool https://www.transifex.com/openstreetmap/id-editor/translate/#en_GB/presets?q=text%3Apredominantly (however searching there is quite tricky)

# How can I add more of them?

I think this feature is great. I would like to extend it / contribute to it.

jguthrie100:
Fixes #5523

### DIFF
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -301,7 +301,6 @@ export default {
 
     docs: function(params, callback) {
         var doRequest = params.debounce ? debouncedRequest : request;
-        params = _extend({ lang: currentLocale }, params);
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @jguthrie100 👍

Issue #5527, opened 2018-11-25
Add guard rail preset


RudyTheDev:
That is, `barrier=guard_rail`. These are all over the place when micro-mapping, but the only `barrier` type that's not searchable.

bhousel:
done in #5528

PR #5528, opened 2018-11-25
Added guard rail preset

RudyTheDev:
This implements https://github.com/openstreetmap/iD/issues/5527. I match-scored it at 0.75, so it doesn't show up for "rail" as the first entries, because it's much more likely the user meant "railway" or "railroad" or whatever.

One issue I have is that `presets.json` has all the new recently added named stores, banks, etc. They got deleted when the file got auto-generated when I built the source. I manually restored them for the PR. But I can't for the life of me figure out where this data is supposed to actually come from for the build to properly include.


data/presets/presets/barrier/guard_rail.json
```diff
@@ -0,0 +1,21 @@
+{
+    "icon": "maki-barrier",
+    "geometry": [
+        "vertex",
```
bhousel: Can it really be a vertex?  (node on a way)

RudyTheDev: Whoops, I copied the preset from kerb and it specified vertex as valid, so that's where it's from. I didn't think to remove it or check against other barrier types. Pushed the fix.

RudyTheDev:
>  You'll need to `npm install` to grab the latest version of that.

Ah, thanks. I didn't think to update packages. For some reason I thought everything's local.

data/presets/presets/barrier/guard_rail.json
bhousel: No problem- just making sure!

PR #5529, opened 2018-11-27
 Add one-sided triangular markers to ways with sides (e.g. natural=cliff).

### LINKED ISSUE #1475: Better rendering for tags with left/right semantics (e.g. coastline, cliff)
The following tags make ways "non-reversible":(aka the semantics depend on way direction) natural=cliff, natural=coastline,    barrier=retaining_wall and man_made=embankment. 

There is no "good" way of handling these outside of showing an "are you sure" alert or similar, this needs to be done for both explicit reversal and implicit (when merging ways).

Updated list of tags:

natural=cliff
natural=coastline
barrier=retaining_wall
barrier=kerb
barrier=guard_rail 
man_made=embankment. 
barrier=city_wall if two_sided != yes

some people believe that waterway=\* belongs in this list too

huonw:
This generalizes the oneway arrow logic for adding SVG markers along a
line. Using that functionality, certain tags get arrows on their
right-hand side, indicating which side is "inside", e.g. the
right-side of a cliff is the lower side.

The list of tags considered to be sided (unless there's a
two_sided=yes tag) is:

- natural=cliff
- natural=coastline
- barrier=retaining_wall
- barrier=kerb
- barrier=guard_rail
- barrier=city_wall
- man_made=embankment

The triangles attempt to be reminiscent of the triangles used for
rendering cliffs on OSM (and elsewhere). The different tags get
different renderings (e.g. colors that match the main way, and
different spacings). In addition, natural=coastline is special-cased
to have blue markers (despite having a green way), to emphasise that
the "inside" of a coastline is the water.

The triangles attempt to be reminiscent of the triangles used for
rendering cliffs on OSM (and elsewhere).

Fixes #1475.

Picture:

id-triangles

(The two lines for barrier=city_wall demonstrates the boundary of lengths when a new marker appears/disappears.)

(This also adds tests for oneway arrows to check that the generalization doesn't break it.)

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -205,7 +205,8 @@ text {
 }
 
 .onewaygroup path.oneway,
-.viewfieldgroup path.viewfield {
+.viewfieldgroup path.viewfield,
+.sidedgroup path.sided {
     stroke-width: 6px;
 }
 
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -56,3 +56,19 @@ export var osmPavedTags = {
         'grade1': true
     }
 };
+
+export var osmRightSideIsInsideTags = {
+    'natural': {
+        'cliff': true,
+        'coastline': 'coastline',
+    },
+    'barrier': {
+        'retaining_wall': true,
+        'kerb': true,
+        'guard_rail': true,
+        'city_wall': true,
+    },
+    'man_made': {
+        'embankment': true
+    }
+};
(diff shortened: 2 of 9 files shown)


bhousel:
This is sooo good - thank you @huonw 🎉

huonw:
Wow, thanks for the fast merge!

RudyTheDev:
First of all, I love this! Time to to go and click `V` a thousand times.

However, I would suggest these be much subtler. In particular, guard rails are placed along the roads, which means they will run parallel to the road and the arrows will point at the road. In practice, this means they overlap (some here are in the wrong direction):

noisy

Here's a more zoomed-in example, which would be a typical "editing zoom":

noisy2

I would say the arrows should be no more than half the current size and without a gap between the line.

bhousel:
Good suggestions @RudyTheDev - I've spun it off into a new issue #5533

PR #5532, opened 2018-11-29
Adds preset for Railing

RudyTheDev:
Today I learned I've been tagging `barrier=guard_rail` when I should have been tagging `barrier=fence; fence_type=railing`...

This adds the preset for the correct way of tagging pedestrian-oriented direction-less railing versus the vehicle road-side direction-based guard rails. That is, [this](https://www.barriersdirect.co.uk/assets/0002/7009/DSC_0097-3.jpg) vs [this](https://www.baizlaw.com/wp-content/uploads/2015/10/guardrail.jpg).

guard rail

The conflict here is that both can be validly called "guard rail", but they have different OSM meanings. And the guard rail has an actual direction, which means all incorrectly tagged railings are now obvious thanks to https://github.com/openstreetmap/iD/pull/5529.


bhousel:
Awesome, thanks @RudyTheDev 👍 
I love that the linear direction rendering is already paying dividends 😄

Issue #2656, opened 2015-05-22
Move presets / data into a separate repository


gmaclennan:
It is hard to track code changes between all the changes and issues around presets.

Keeping the data folder as a separate repository would also make it easier for those of us using iD with custom presets, and in general keep the issues cleaner for each.

jgpacker:
I believe it is possible to create a separate repository and add it as a *git submodule* to this one, while preserving those files's history (using some of gits history-rewriting commands). Though I'm not sure how we could migrate the related Github issues to the new presets repository.

It should also be verified if the presets are cleanly separated in different directories, because otherwise we would need to refactor current code to accomodate this change (and that could be a deal-breaker unless it was a really trivial change).

bhousel:
> It should also be verified if the presets are cleanly separated in different directories

Yes the presets are cleanly isolated under `data/presets`.  It would be very easy to just remove all the preset files and build script into another repository, and that's basically what I'm proposing here: osmlab/editor-presets#2

bhousel:
Bumping the priority of this issue.  As soon as reasonably possible, I will:
- [ ] move OSM presets to separate repository
- [ ] create a separate similar repository for HOT presets

HOT really needs their own presets because:
* HOT uses a far more limited preset subset than OSM does
* HOT is more tag-focused, so we shouldn't try to hide, simplify, or localize, unfortunately named OSM tags (like `unclassified` and `track`)
* Any little changes we make on the OSM side shouldn't invalidate all of the HOT training materials

This also helps me out because now quibbles about proper tagging can be moved elsewhere.

gmaclennan:
I created a separate module for building presets, which expects data in the same structure as `data/presets`: https://github.com/digidem/id-presets-builder

We use this for building custom presets for different deployments. Hopefully this can be useful for others working with custom presets and also be used when OSM presets move to a separate repository.

**Differences from the iD `build.js`**

Most of the code is pulled from [build.js](https://github.com/openstreetmap/iD/blob/master/build.js) with the following differences:

- Everything is async, which speeds things up a bit.
- Additional schema checks for [`defaults.json`](https://github.com/digidem/id-presets-builder/blob/master/schema/defaults.json) and [category files](https://github.com/digidem/id-presets-builder/blob/master/schema/category.json) - it may be useful to pull this into `build.js`.
- Additional checks that any categories or presets in `defaults.json` actually exist.
[...]

PR #5582, opened 2018-12-09
Define additional preset-specific fields for the "Add field" dropdown

### LINKED ISSUE #4871: Hide some preset fields by default
In many cases, it makes sense to offer a field in the “Add field” dropdown for a particular preset but not display it by default for every feature that uses the preset. The idea is to make the tag more discoverable (so users don’t have to know the raw tag name and values) but avoid making the field too prominent or take up valuable screen real estate (because the field wouldn’t apply to enough features).

Currently, it’s only possible to make a field behave this way by setting `universal` to `true`. However, that adds it to the “Add field” dropdown for all presets, whereas in the following cases, we only want it to be available on some presets:

https://github.com/openstreetmap/iD/issues/4816#issuecomment-366731895
#4522
https://github.com/openstreetmap/iD/issues/4722#issuecomment-366487679

Perhaps a preset JSON file could have an `optional_fields` property alongside the usual `fields` property. These optional fields could appear at the top of the “Add field” dropdown, above the universal fields.

quincylvania:
Closes #4871.

As discussed in the issue, this PR adds a `moreFields` property to presets that allows for defining non-default fields that the user can add manually. This lets us include fields for uncommon or esoteric tags that a minority of mappers may want use without cluttering up the default UI.

For example, here is the definition for Hotel with two of the more advanced fields moved to `moreFields`:

```JSON
{
    "icon": "maki-lodging",
    "fields": [
        "name",
        "brand",
        "operator",
        "address",
        "building_area",
        "smoking",
        "rooms",
        "internet_access",
        "internet_access/fee"
    ],
    "moreFields": [
        "stars",
        "internet_access/ssid"
    ],
    "geometry": [
        "point",
        "area"
    ],
    "tags": {
        "tourism": "hotel"
    },
    "name": "Hotel"
}
```

Resulting in this UI: 

screen shot 2018-12-09 at 1 52 54 pm

Hotel is one of only a few presets actually affected in this PR, which is meant primarily to lay the groundwork for future data changes.

I did not go so far as to remove universal fields. There can be further discussion on that issue once we see how people end up using `moreFields`.


bhousel:
Looks great!  Go ahead and merge if it's ready 👍

bhousel:
cc @nyurik - This is a non-breaking schema change to the preset format.. I think it's valuable to be able to iterate quickly on changes like this, instead of waiting for community buy-in to change the preset format that we use.

On this issue we had considered making the presets a separate project, but it's not something I want to do right now:  https://github.com/openstreetmap/iD/issues/2656#issuecomment-399775749

nyurik:
Thanks @bhousel , I haven't poked at it yet - want to finish the migration to KeyDescription & ValueDescription (e.g. just migrated Pl:KeyDescription, which was using their own version all this time). TODO - Pl:ValueDescription, and the bot needs a lot of rewriting love. I will poke at it afterwards.

PR #5583, opened 2018-12-10
Mechanism for displaying fields conditionally based on tags

### LINKED ISSUE #5581: Conditional field display based on tags
Sometimes it only makes sense to display a field if certain tags are present on the feature. For example, the Payment Types field would be useful on presets such as Parking Lot, Telephone, or Toilets, but only when `fee=yes`. Another use would be in #5580, where the field for `diet:vegetarian` would be redundant if `diet:vegan=yes`.

I propose finding a generalized solution where conditions can be defined in JSON and not in code.

Related: #4871

quincylvania:
Closes #5581.

This is a bare-bones implementation of the issue. It adds a `prerequisiteTag` property to the field schema that can restrict fields to show only in one of the following cases:

- A given key/value tag is present
- A given key is present with any value
- A given key is present but it does not have the given value

For example, this hides the Denomination field unless there is already a `religion` tag.

```js
{
    "key": "denomination",
    "type": "combo",
    "label": "Denomination",
    "prerequisiteTag": {
        "key": "religion"
    }
}
```

This hides the Internet Access Fee field unless the feature has an `internet_access` tag not equal to `no`.

```js
{
    "key": "internet_access:fee",
    "type": "check",
    "label": "Internet Access Fee",
    "prerequisiteTag": {
        "key": "internet_access",
        "valueNot": "no"
    }
}
```

I tried to keep it simple. If more complex conditionals are needed in the future, we can revise or build upon the design.


mmd-osm:
For the first example, what happens to those 8000 existing pois where _denomination_ is set, but no _religion_? How would you know _religion_ is missing, if you can't see either one?

quincylvania:
@mmd-osm Thanks for your feedback!

Fields with values are shown even if the tag prerequisite isn't met. So if there is a `denomination` tag, the Denomination field will appear even if there is no `religion` tag. If the `denomination` tag is removed before a `religion` tag is set, the Denomination field will disappear.

Also, fields defined in `fields` or `moreFields` without prerequisites will still show up. So for the School preset, the Religion fields would still appear even if empty, no matter the value of `denomination` or any other tag.

PR #5594, opened 2018-12-13
Add apartment complex preset

kreed:
I see a lot of new mappers use the building=apartments preset to tag entire apartment complexes, resulting in huge, bogus buildings. I'm hoping adding a preset for apartment complexes will help with that.


bhousel:
Thanks @kreed , looks good 👍

PR #5606, opened 2018-12-15
Standing water fill pattern and water "barriers"

RudyTheDev:
This adds a fill pattern for standing water for `landuse=reservoir`/`natural=water`;`water=reservoir`:

res

Also the same for a fountain as area:

fount

Additionally barrier-like stroke for `man_made=breakwater` and `man_made=groyne` ways and areas.

gr way gr area

Such as this off-shore example:

bw


bhousel:
Thanks @RudyTheDev!

Issue #4876, opened 2018-03-11
Splitting roads can change order of ways in relation


ockendenjo:
A number of the bus_route relations in Lancaster, UK have recently become invalid.

User:gurglypipe has been modifying the roads in the City Centre and has  split the roads in a few places. These are the relevant changesets.
https://www.openstreetmap.org/changeset/56998682
https://www.openstreetmap.org/changeset/56997445

See relation: https://www.openstreetmap.org/relation/6536546

I link to two files:
[Relation version 37 - though to be valid](https://github.com/openstreetmap/iD/files/1800707/rel_37.txt)
[Relation version 39 - after some roads were split using ID](https://github.com/openstreetmap/iD/files/1800708/rel_39.txt)

User:gurglypipe report that they have not specifically modified the relation

Using a diff tool on files, several ways have moved order in the relation. This should not happen when roads are split. This relation contains several loops, which might be why the problem has occurred??

ockendenjo:
![Kdiff showing differences](https://user-images.githubusercontent.com/29466887/37257032-579bcec4-255b-11e8-946b-eb73ed0cf142.jpeg)

bhousel:
Thanks @ockendenjo - this is related to #4589 

The fix for that bug might have introduced this one, or it could be a case that iD always got wrong.  It's very hard to order the member ways correctly in a relation that loops back over itself, especially when the relation is very large and iD hasn't downloaded all of its members.   

My _guess_ is that one of those ways inside the moved range was what was split, but iD hadn't downloaded the other parts of the relation - this is why iD joined w473970093 to itself.

The best way to address this issue would be to reduce the issue to a minimal condition and write a test case for it in here:
 https://github.com/openstreetmap/iD/blob/master/test/spec/osm/multipolygon.js#L147

ockendenjo:
The ways were already ordered correctly in the relation so ID would not have needed to reorder them.
The new segment of road should either have inserted before or after the existing (shortened) way, depending on the travel direction. Calculating the direction in which the way is travelled might be the difficult thing, and probably depends on the ways either side being downloaded?

Even if ID got the before/after decision wrong this would probably have been better than reordering all the ways

I've also come across a [case](https://www.openstreetmap.org/relation/7919795) where splitting a road has caused the stops/platforms in the relation to be reordered. This is presumably a side effect of the roads being re-ordered incorrectly?
[...]

PR #5624, opened 2018-12-18
Add more tests for splitting ways in relations containing loops

ockendenjo:
I've added some tests for the splitting of route relations containing loops as described in bug #4876 

There are probably too many combinations of ways directions, loop traversal direction to test them all, but I can add more combinations if it's helpful.


bhousel:
> There are probably too many combinations of ways directions, loop traversal direction to test them all, but I can add more combinations if it's helpful.

Thank you @ockendenjo!  Testing how iD works with more kinds of route topologies is always a great thing 👍

PR #5627, opened 2018-12-19
Extend data model for validation issues

gaoxm:
1. Add a class to represent the validation issue
2. Extend the data model for an validation issue to
  (1) add a severity level field (useful for identify save-blocking issues later)
  (2) replace single entity with an array of entities (useful for issues involving multiple entities)
  (3) add a coordinates field for highlighting the location of the issue on the map
  (4) add a fixes field for possible automatic fixes

3. Update existing validation modules to use the new data model

### DIFF
--- a/modules/core/history.js
+++ b/modules/core/history.js
@@ -281,9 +281,12 @@ export function coreHistory(context) {
 
 
         validate: function(changes) {
-            return _flatten(_map(Validations, function(fn) {
-                return fn(context)(changes, _stack[_index].graph);
-            }));
+            return _flatten(_map(
+                _filter(Validations, _isFunction),
+                function(fn) {
+                    return fn(context)(changes, _stack[_index].graph);
+                }
+            ));
         },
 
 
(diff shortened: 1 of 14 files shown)


modules/validations/validation_issue.js
```diff
@@ -0,0 +1,41 @@
...
+var ValidationIssueSeverity = Object.freeze({
+    warning: 'warning',
+    error: 'error',
+});
+
+
+export { ValidationIssueType, ValidationIssueSeverity };
+
+
+export function validationIssue(attrs) {
```
wonga00: can this be `ValidationIssue`?

modules/core/history.js
```diff
@@ -281,9 +281,12 @@ export function coreHistory(context) {
 
 
         validate: function(changes) {
-            return _flatten(_map(Validations, function(fn) {
-                return fn(context)(changes, _stack[_index].graph);
-            }));
+            return _flatten(_map(
+                _filter(Validations, _isFunction),
```
wonga00: i think this extra check should probably happen just once when the `Validations` is created

modules/validations/validation_issue.js
gaoxm: I was trying to follow the convention of other "classes" like geoExtent, osmEntity, etc.

PR #5632, opened 2018-12-19
Change the initial inspector of the sidebar upon feature selection

quincylvania:
Old behavior: The preset list is shown when selecting a feature without tags (except `area=yes`) and that is not member of a relation. Otherwise show the editor.

New behavior: The preset list is shown when selecting a new feature or a `vertex` without tags. Otherwise show the editor.

### DIFF
--- a/modules/osm/entity.js
+++ b/modules/osm/entity.js
@@ -143,11 +143,13 @@ osmEntity.prototype = {
     },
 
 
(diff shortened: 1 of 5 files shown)


bhousel:
Sounds good to me!
Also nice to rename `isUsed()` to something more descriptive.

PR #5636, opened 2018-12-20
open combobox if anywhere on box clicked...

### LINKED ISSUE #5596: Combobox should toggle on clicking anywhere on the field, not just the carat
Click on fields should show list of options. In the previous release, click on a field in the sidebar after selecting something showed list of options, now it doesn't work. We have to click on the arrow to show list of options. It would be great we can have the old behavior.

maxgrossman:
closes #5596

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1906,6 +1906,7 @@ div.combobox {
     margin-left: -30px;
     vertical-align: top;
     cursor: pointer;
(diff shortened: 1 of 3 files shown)


bhousel:
Looks ok, thanks @maxgrossman!
We really should figure out what's wrong with that one Wikipedia test on Node 10  🤔

PR #5640, opened 2018-12-20
Import Custom Validation Rule by linking to json generated from MapCSS

abalosc1:
custom mapcss rule

To Test:
Add Point with Healthcare Facility preset but don't add a name tag
https://api.myjson.com/bins/lo480


abalosc1:
Also set up the framework to build in some validations from the JOSM MapCSS Tag Checks
sidewalk
More of these can be brought in once we build out
https://github.com/radiant-maxar/mapcss-parse

quincylvania:
Hmm this never proceeded and lots has changed about the validation system in the last year…

@abalosc1 Do you mind if we close this in favor of a revised implementation?

PR #5645, opened 2018-12-21
Improve cross way check

gaoxm:
1. Add more way types and relation types to check
2. Add more legit crossing cases
3. Add tests for 1 and 2

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -1,3 +1,4 @@
+import _clone from 'lodash-es/clone';
 import { geoExtent, geoLineIntersection } from '../geo';
 import { set as d3_set } from 'd3-collection';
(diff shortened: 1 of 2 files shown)


modules/validations/crossing_ways.js
```diff
@@ -39,10 +40,104 @@ export function validationHighwayCrossingOtherWays() {
...
+    }
+
+    function extendTagsByInferredLayer(tags, way) {
+        if (!hasTag(tags, 'layer')) {
+            tags.layer = way.layer().toString();
+        }
+        return tags;
+    }
+
+    function isLegitCrossing(way1, featureType1, way2, featureType2, graph) {
```
gaoxm: @bhousel @quincylvania please take a look at all the cases here and comment whichever seems incorrect.

PR #5646, opened 2018-12-21
add validation for connecting highway with extension

gaoxm:
but avoid connect when the edge goes in parallel to the other road.
also added tests for the validation.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -930,6 +930,9 @@ en:
     crossing_ways:
       message: Crossing ways without connection
       tooltip: "Roads are crossing other roads, buildings, railroads, or waterways without connection nodes or a bridge tag."
+    highway_almost_junction:
+      message: Almost junction
+      tooltip: "This node is very close but not connected to way {wid}."
     fix:
       delete_feature:
         title: Delete this feature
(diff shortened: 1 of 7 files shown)


SilentSpike:
I believe it should be re-worded to something along the lines of "Almost Junction" as commonly found in QA tools. Reason being that `Potential connection with extension` doesn't sound like an error, but rather a suggestion.

Same for the tooltip, it seems to be suggesting a connection should be made when really the user should investigate more closely and either mark with `noexit` (if there truly is no connection) or map whatever missing connections there are.

gaoxm:
> I believe it should be re-worded to something along the lines of "Almost Junction" as commonly found in QA tools. Reason being that `Potential connection with extension` doesn't sound like an error, but rather a suggestion.
> 
> Same for the tooltip, it seems to be suggesting a connection should be made when really the user should investigate more closely and either mark with `noexit` (if there truly is no connection) or map whatever missing connections there are.

Makes sense. @bhousel @quincylvania would you guys also suggest a better name/message for this? JOSM calls this "way ends near another highway": https://wiki.openstreetmap.org/wiki/JOSM/Validator

gaoxm:
@bhousel @quincylvania ping - do you think I should update the name and message for the validation check before merging this PR?

bhousel:
Hey @gaoxm  - Yes the KeepRight version of this check looks like:
```
          title: 'Almost junction'
          description: 'This node is very close but not connected to way {var1}.'
```

I agree with @SilentSpike , that i like the name "Almost Junction" a bit better.

gaoxm:
@bhousel @quincylvania validation name and messages are all updated.

Issue #3776, opened 2017-01-18
Use Pt: wiki pages in pt-BR iD


virgilinojuca:
Recently, the Portuguese (`Pt:`) and Brazilian Portuguese (`Pt-br:`) wikis were unified into a single namespace (`Pt:`) so that the Portuguese-speaking community can join forces. There are no `Pt-br` pages with content in the wiki anymore. All of them were either merged with their correspondent `Pt:` pages or moved to `Pt:`. That means that the old `Pt-br:` pages are just redirects now.

So now we have a problem in iD. Currently, when in the pt-BR interface, which will continue to exist, iD looks up for documentation in the now deprecated `Pt-br:` namespace, so we only get a warning that the page was moved (#3337). Can you guys change the lookup namespace to `Pt:` in the pt-BR iD?

bhousel:
Ok, I guess this is a step further beyond #3337.  Can someone else (@naoliv?) confirm that all `pt-BR:` pages on osm wiki are now just redirects to `pt:` pages?   If so, I'm fine with special casing this instead of showing the redirect text.

naoliv:
Verified and they are all redirects now.

PR #5647, opened 2018-12-22
OSM Multilingual data directly from wiki (wikibase data items)

nyurik:
This completes the work by @bhousel (thanks for getting me started!)

* Takes data directly from the Wikibase data items (OSM Wiki)
   https://wiki.openstreetmap.org/wiki/OpenStreetMap:Data_Items
* Understands the difference in regions - e.g. will show different
  images depending on the local settings
* Perf: Single request will get both the tag and key description
* Slightly different wording to encourage description editing/translation

image

| RU has a custom image | EN has no image |
|-----|-----|
| image | image |

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -405,7 +405,6 @@ en:
   inspector:
     no_documentation_combination: There is no documentation available for this tag combination
     no_documentation_key: There is no documentation available for this key
-    documentation_redirect: This documentation has been redirected to a new page
     show_more: Show More
     view_on_osm: View on openstreetmap.org
     all_fields: All fields
@@ -418,6 +417,7 @@ en:
     choose: Select feature type
     results: "{n} results for {search}"
     reference: View on OpenStreetMap Wiki
+    edit_reference: Edit or translate on OSM Wiki
     back_tooltip: Change feature
     remove: Remove
     search: Search
--- a/modules/services/osm_wikibase.js
+++ b/modules/services/osm_wikibase.js
@@ -1,0 +1,74 @@
+import _debounce from 'lodash-es/debounce';
+import _forEach from 'lodash-es/forEach';
+
+import { json as d3_json } from 'd3-request';
+
+import { utilQsString } from '../util';
+import { currentLocale } from '../util/locale';
+
+
+var apibase = 'https://wiki.openstreetmap.org/w/api.php';
+var _inflight = {};
+var _wikibaseCache = {};
+
+
+var debouncedRequest = _debounce(request, 500, { leading: false });
+
+function request(url, callback) {
+    if (_inflight[url]) return;
+
+    _inflight[url] = d3_json(url, function (err, data) {
+        delete _inflight[url];
+        callback(err, data);
+    });
+}
+
+
+export default {
+
+    init: function() {
+        _inflight = {};
+        _wikibaseCache = {};
+    },
+
+
+    reset: function() {
+        _forEach(_inflight, function(req) { req.abort(); });
+        _inflight = {};
+    },
+
+
+    docs: function(params, callback) {
+        var doRequest = params.debounce ? debouncedRequest : request;
+
+        // if (params.value) path = 'tag/wiki_pages?';
+        // else if (params.rtype) path = 'relation/wiki_pages?';
+
+        var obj = {
+            action: 'wbgetentities',
+            sites: 'wiki',
+            titles: 'Tag:amenity=parking',
+            languages: 'en',
+            origin: '*',
+            formatversion: 2,
+            format: 'json'
+        }
+        var url = apibase + '?' + utilQsString(obj);
+        doRequest(url, function(err, d) {
+            if (err) {
+                callback(err);
+            } else {
+                _wikibaseCache[url] = d.data;
+                callback(null, d.data);
+            }
+        });
+    },
+
+
+    apibase: function(_) {
+        if (!arguments.length) return apibase;
+        apibase = _;
+        return this;
+    }
+
+};
--- a/modules/services/osm_wikibase.js
+++ b/modules/services/osm_wikibase.js
@@ -4,7 +4,6 @@ import _forEach from 'lodash-es/forEach';
 import { json as d3_json } from 'd3-request';
 
 import { utilQsString } from '../util';
-import { currentLocale } from '../util/locale';
 
 
 var apibase = 'https://wiki.openstreetmap.org/w/api.php';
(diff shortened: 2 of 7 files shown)


modules/services/osm_wikibase.js
```diff
@@ -0,0 +1,185 @@
...
+
+
+    reset: function() {
+        _forEach(_inflight, function(req) { req.abort(); });
+        _inflight = {};
+    },
+
+
+    /** List of data items representing language regions.
+     *  To regenerate, use Sophox query:  http://tinyurl.com/y6v9ne2c (every instance of Q6999)
```
1ec5: This seems like a list that will grow over time, perhaps even as frequently as once per release cycle. Consider splitting it out into a separate file (JSON perhaps) that can be updated as part of iD’s build process.

modules/services/osm_wikibase.js
```diff
@@ -0,0 +1,185 @@
...
+     *  A less accurate list can be seen here (everything that links to Q6999):
+     *  https://wiki.openstreetmap.org/w/index.php?title=Special%3AWhatLinksHere&target=Item%3AQ6999&namespace=120
+     */
+    regionCodes: {
+        ar: 'Q7780', az: 'Q7781', bg: 'Q7782', bn: 'Q7783', ca: 'Q7784', cs: 'Q7785', da: 'Q7786',
+        de: 'Q6994', el: 'Q7787', es: 'Q7788', et: 'Q7789', fa: 'Q7790', fi: 'Q7791', fr: 'Q7792',
+        gl: 'Q7793', hr: 'Q7794', ht: 'Q7795', hu: 'Q7796', id: 'Q7797', it: 'Q7798', ja: 'Q7799',
+        ko: 'Q7800', lt: 'Q7801', lv: 'Q7802', ms: 'Q7803', nl: 'Q7804', no: 'Q7805', pl: 'Q7806',
+        pt: 'Q7807', ro: 'Q7808', ru: 'Q7809', sk: 'Q7810', sq: 'Q7811', sv: 'Q7812', tr: 'Q7813',
+        uk: 'Q7814', vi: 'Q7815', yue: 'Q7816', 'zh-hans': 'Q7817', 'zh-hant': 'Q7818',
```
1ec5: It would be nice to have all of [iD’s current localizations](https://github.com/openstreetmap/iD/tree/master/dist/locales) represented here if possible.

modules/services/osm_wikibase.js
```diff
@@ -0,0 +1,185 @@
...
+        _inflight = {};
+    },
+
+
+    /** List of data items representing language regions.
+     *  To regenerate, use Sophox query:  http://tinyurl.com/y6v9ne2c (every instance of Q6999)
+     *  A less accurate list can be seen here (everything that links to Q6999):
+     *  https://wiki.openstreetmap.org/w/index.php?title=Special%3AWhatLinksHere&target=Item%3AQ6999&namespace=120
+     */
+    regionCodes: {
```
1ec5: It looks like this is a list of locales, rather than regions, to use [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag) terminology. Other parts of iD use the term “locale” to refer to this concept, so I think we should be consistent here too.

modules/services/osm_wikibase.js
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+
+        if (!titles.length) {
+            // Nothing to do, we already had everything in the cache
+            return callback(null, result);
+        }
+
+        var obj = {
+            action: 'wbgetentities',
```
1ec5: Do you think there’s an opportunity to merge this code with the existing Wikidata item lookup functionality somehow, given that they’re both Wikibase instances?

https://github.com/openstreetmap/iD/blob/3de03439e23b5f605c43b3667361990cd168ab94/modules/services/wikidata.js#L30-L38

modules/ui/tag_reference.js
```diff
@@ -10,53 +7,61 @@ import { t } from '../util/locale';
...
 
-
+    /**
+     * @returns {{itemTitle: String, description: String, image: String|null}|null}
+     **/
     function findLocal(data) {
-        var locale = utilDetect().locale.toLowerCase();
-        var localized;
-
-        if (locale !== 'pt-br') {  // see #3776, prefer 'pt' over 'pt-br'
```
1ec5: For the benefit of others reading this PR, I’m pretty sure the MediaWiki API handles #3776 and other language fallbacks implicitly.

modules/ui/tag_reference.js
```diff
@@ -10,53 +7,61 @@ import { t } from '../util/locale';
...
+            } else {
+                image = wikibase.claimToValue(entity, 'P28', langCode);
+                if (image) {
+                    url = 'https://wiki.openstreetmap.org/w/index.php';
+                }
+            }
+            if (image) {
+                result.image = {
+                    url: url,
+                    title: 'Special:Redirect/file/' + image
```
1ec5: Special:Redirect downloads the full image, which can be [huge](https://wiki.openstreetmap.org/wiki/Item:Q145#P4) for such a small piece of screen real estate.

[This MediaWiki API query](https://wiki.openstreetmap.org/wiki/Special:ApiSandbox#action=query&format=json&prop=imageinfo&titles=File%3ADollar%20General%2C%20Glen%20St.%20Mary.JPG&iiprop=url|size&iiurlwidth=100) gets the 100-pixel-wide thumbnail of an image given the image name. A `maxage` parameter would also make sense, since we don’t expect the images to change very frequently.

data/core.yaml
```diff
@@ -418,6 +417,7 @@ en:
     choose: Select feature type
     results: "{n} results for {search}"
     reference: View on OpenStreetMap Wiki
+    edit_reference: Edit or translate on OSM Wiki
```
1ec5: Is this string used anywhere?

modules/services/osm_wikibase.js
nyurik: I would actually want to shrink this list, not grow it :)  The reason for this list is to support all the existing discrepancies in the wiki, e.g. status = `in-use` in some languages, and `defacto` in other. This makes very little sense to encourage such differentiation, so I created the smallest possible list (only the ones that actually conflicted)

modules/services/osm_wikibase.js
bhousel: I'd rather not, at least right now.  They are 2 different wikibases.  (Also, I don't see us ever adding more).

nyurik: Not sure TBH.  They do use the same API, but they tailor to wikidata vs osmwiki structure differently. For example, osmwiki uses specific properties and qualifiers to limit the data, has a different sitelink handling, etc. Eventually it might make sense to have a separate utilities library for the common functionality, e.g. to parse claims, but I think services are sufficiently different to warrant separate code - different properties, different lookup methods.

bhousel: Actually we could probably get rid of `services/wikipedia.js` and fold what that does into `services/wikidata.js` (but again lets do that another time)

modules/ui/tag_reference.js
nyurik: @1ec5 this is not correct, run these commands:
```
url -L 'https://en.wikipedia.org/w/index.php?title=Special:Redirect/file/File:Dollar_General,_Glen_St._Mary.JPG&width=100&height=100' > small.jpg
curl -L 'https://en.wikipedia.org/w/index.php?title=Special:Redirect/file/File:Dollar_General,_Glen_St._Mary.JPG' > large.jpg
ls -al small.jpg large.jpg
```
small.jpg is 2749 bytes

modules/services/osm_wikibase.js
bhousel: Why can't the service just fetch these on `init()` ?

modules/ui/tag_reference.js
1ec5: Ah, I missed that the `width` and `height` parameters were being tacked onto the end. As long as using Special:Redirect this way doesn’t raise server-side performance issues like [thumb.php](https://meta.wikimedia.org/wiki/Talk:Www.wikipedia.org_template/2006#thumb.php), then this is much simpler than what I suggested. 🙂

nyurik: Heh, that was a while ago :))   Redirect in theory should just [generate a URL](https://github.com/wikimedia/mediawiki/blob/0e56ff447bbb74df007f2573725baaa61ed2745c/includes/specials/SpecialRedirect.php#L111) according to MediaWiki rules, and redirect to the actual image store - which in turn will handle the missing size to dynamically scale it.

modules/services/osm_wikibase.js
nyurik: Does `init()` support async?  In any case, this would be a fast-enough call to do on the first request to the service, and cache it.  Patch is on its way...

bhousel: The taginfo interface does a similar lookup on `init()` to fetch popular keys.  Just try to make sure that if it fails iD will still work ok (maybe just without documentation lookups)

https://github.com/openstreetmap/iD/blob/e51f2eb1e706e369d8e8518818d6ee5747a9cf15/modules/services/taginfo.js#L202-L208

data/core.yaml
nyurik: Good catch!  It got lost in a merge accident :)

pnorman:
Does this have any implications for editing when the wiki is offline?

bhousel:
> Does this have any implications for editing when the wiki is offline?

Not really - the built in tag documentation already comes from the wiki.  So this PR is just to request the documentation directly from the wikibase API rather than going through taginfo API for it.

bhousel:
Thanks @nyurik  - looks good to merge 👍

bhousel:
> I'd rather expect a link to take me to a Wiki page with more human-readable explanations as to what the key is about (e.g. https://wiki.openstreetmap.org/wiki/Key%3Amaxheight).

@1ec5 @nyurik - should we also leave an OSM wiki sitelink there, if exists a suitable one?

I thought we were keeping them, but then again I'm pretty 👎 about the OSM wiki these days, so I'm ok with not including it.  😕

nyurik:
I do think we should have both -- a link to add/fix/translate short descriptions, and a link to the wiki page of it exists (could be two separate lines). To make it simple to detect existence in a given language. I will need to populate the existence of the wiki page in the data items.

nyurik:
P.S. example of the data item with translation info: https://wiki.openstreetmap.org/wiki/Item:Q104#P31

Issue #5620, opened 2018-12-18
Add Horse Racing Track preset


quincylvania:
Tags: `leisure=track` and `sport=horse_racing`

See https://wiki.openstreetmap.org/wiki/Tag:sport%3Dhorse_racing

PR #5649, opened 2018-12-22
Add sport=equestrian

tordans:
This started to fix https://github.com/openstreetmap/iD/issues/5620.

However, I think `sport=horse_racing` is already part of https://github.com/openstreetmap/iD/blob/master/data/presets/fields/sport_racing_nonmotor.json I think the issue is already solved.

I checked the recommended combinations (see https://wiki.openstreetmap.org/wiki/Tag:sport%3Dhorse_racing) like `leisure=track` and they either all use the sport_racing_nonmotor field list or a generic sport dropdown which also includes `horse_racing`.

I checked the JOSM presets and could not find anything else that we should add here - https://josm.openstreetmap.de/browser/trunk/data/defaultpresets.xml.

--- 

However, I noticed that `sport=equestrian` was missing from the list, so this PR adds it.
See https://wiki.openstreetmap.org/wiki/Tag:sport%3Dequestrian

Is there anything else like translations that needs to happen?


bhousel:
> However, I think `sport=horse_racing` is already part of https://github.com/openstreetmap/iD/blob/master/data/presets/fields/sport_racing_nonmotor.json I think the issue is already solved.

I think what @quincylvania wants is a preset that already has the `sport=horce_racing` added.  So we already have a "Running Track" preset, and we'd copy that one over to a new one "Horse Racing Track".

> However, I noticed that `sport=equestrian` was missing from the list, so this PR adds it.
> See https://wiki.openstreetmap.org/wiki/Tag:sport%3Dequestrian

I'm going to close this - I'm pretty sure combining `sport=equestrian` with `leisure=track` is the wrong thing to do (especially since `sport=horce_racing` is already on the list).

Issue #5644, opened 2018-12-21
Allow to overwrite editor language via URL


tordans:
I would be great to specify the language for iD, so I can test it the way most code contributors see it and make screenshots for issues that everyone understands.

Some research
- In https://github.com/openstreetmap/iD/issues/1666 (2013) is was said this is already possible, but I don't think that is true anymore, since I dont see any URL param
  - UPDATE: Its possible via the preferred language input on the OSM website user profile. See https://github.com/openstreetmap/iD/issues/3842 for an old screenshot of this UI element.
- I guess https://github.com/openstreetmap/iD/blob/master/modules/util/detect.js#L62-L63 is the place that handles the language

Suggested solution
- An optional URL param `language=en` to force the editor to english

bhousel:
done in #5650

PR #5650, opened 2018-12-22
Allow locale and language being set via URL param

tordans:
With the hash-url `locale=en-US` or `locale=de-DE` one can force a locale and language regardless of the given language from the osm-website-settings.

This is to solve https://github.com/openstreetmap/iD/issues/5644.

However, while looking into this I understood that the interface language is in fact pulled from the OSM-website-profile-language (not the browser language). So there is a way to change the language, by editing the preferred languages in the profile.

So while this PR makes changing the language a bit easier and quicker, I am not sure if it is worth the added complexity.

In case this should be continued, those thinks are todo IMO:
- [x] This line https://github.com/openstreetmap/iD/blob/master/modules/util/detect.js#L93 is duplicated now, should it be removed?
- [x] Should there be specs for this? If so, I don't know enough about specs to write good ones.

### DIFF
--- a/modules/util/detect.js
+++ b/modules/util/detect.js
@@ -59,24 +59,31 @@ export function utilDetect(force) {
     // Added due to incomplete svg style support. See #715
     detected.opera = (detected.browser.toLowerCase() === 'opera' && parseFloat(detected.version) < 15 );
 
-    detected.locale = (navigator.language || navigator.userLanguage || 'en-US');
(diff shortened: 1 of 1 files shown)


tordans:
@bhousel ready for review. Thanks.

bhousel:
Thanks @tordans  - looks good!
Could you also add a short description of the parameter here: https://github.com/openstreetmap/iD/blob/master/API.md

Issue #5652, opened 2018-12-23
highway=crossing has wrong "unmarked crossing" description 


Komzpa:
Hi,

`highway=crossing` has wrong description "Uncontrolled crossing". Uncontrolled crossing is `crossing=uncontrolled`. `highway=crossing` is umbrella for all crossings, often controlled ones too.

Context: https://github.com/openstreetmap/iD/commit/1ffca3ce85d630539c7ebb11454a91770bd76536

bhousel:
Can you share a screenshot?

I think “Uncontrolled” has been used very inconsistently, and I didn’t think iD is anymore calling them that in the UI anymore. 

Ideally people would remap these as `marked` or `unmarked`, to resolve the ambiguity. 

There are lots of old tickets that you can read about this - I don’t feel like looking them up right now.

tohaklim:
I believe this is what @Komzpa is referring to:
![image](https://user-images.githubusercontent.com/7071423/50388135-f574da00-071e-11e9-8597-6c7de55b984c.png)
Both options have the same generic description and the same wiki link. I am not necessarily against cleaning up crossing tags, but not sure if auto-completing a `crossing=marked` tag that is not even on the wiki is best? The aforementioned wiki page already has one example of iD introducing a questionable preset with `crossing=zebra`

1ec5:
If you select a feature that’s only tagged `highway=crossing`, or perhaps one that’s tagged `highway=crossing` `crossing=traffic_signals`, iD uses the Unmarked Crosswalk preset. It used to be that `highway=crossing` without a `crossing=*` tag would use the generic Pedestrian Crossing preset without referring to whether pavement markings are present, but the generic preset has been renamed to use the same title as the `crossing=unmarked` preset. I believe that’s what @Komzpa is referring to, not the fact that the documentation is identical between the two presets.

`highway=crossing` only | `crossing=unmarked`
----|----
![generic](https://user-images.githubusercontent.com/1231218/50388242-17da0980-06c5-11e9-844c-bf43e8e876de.png) | ![unmarked](https://user-images.githubusercontent.com/1231218/50388240-14df1900-06c5-11e9-808f-5255da26e94e.png)

[...]

Issue #5205, opened 2018-08-05
single preset or better extended preset for crossing=traffic_signals


Lukas458:
Where pedestrians are crossing the road with traffic lights, there can be added more informationen about the crossing than just tactile_paving=* and kerb=*. What do you think, wouldn't it be useful or would it somehow be possible for crossing=traffic_signals, to add _button_operated=yes/no, traffic_signals:arrow=* traffic_signals:vibration=yes/no and traffic_signals:sound=*_? I think this would encourage people to look at this and adding it in an easy way.

In most these are all facilities used by disabled people, but that's the same for e. g. tactile_paving.

1ec5:
As someone who’s mapped wayyy too many signalized crosswalks, I would find it very convenient if the Marked Crosswalk and Unmarked Crossing presets could have a Pedestrian Signal checkbox. This is closely related to the request above but smaller in scope, so hopefully more feasible to implement cleanly.

Currently, one of the popular `crossing=*`values is [`traffic_signals`](https://wiki.openstreetmap.org/wiki/Tag:crossing%3Dtraffic_signals), which indicates that pedestrians using the crossing must wait for a walk light, potentially after pressing a “beg button”. But even before 1ffca3ce85d630539c7ebb11454a91770bd76536, iD’s presets have been focused on `crossing` values that describe the appearance of a crossing rather than its signalization.

Signalization is important, sometimes moreso than pavement markings. For example, routers need to penalize signalized crossings when giving walking directions, regardless of whether the crossing is marked. (The marked-versus-unmarked distinction is generally more meaningful to renderers.)

According to taginfo, 3% of all instances of `traffic_signals=*` are on `crossing=*` features. @bhousel pointed out in https://github.com/openstreetmap/iD/issues/5205#issuecomment-410784580 that this tag is ambiguous, because mappers could intend it to represent either pedestrian or motorist signals at a crossing node.

There are a few ways to resolve the ambiguity:

[...]

PR #5654, opened 2018-12-23
Add Pedestrian Signals field to crossing presets

1ec5:
Added a Pedestrian Signals field to all the crossing presets so that mappers can indicate the presence of [crossing signals](https://en.wikipedia.org/wiki/Pedestrian_crossing#Signals). The checkbox adds the `traffic_signals=yes` tag to the way or node.

civic-center

`crossing=unmarked` | `crossing=marked` | `railway=crossing`
----|----|----
crossing=unmarked | crossing=marked | railway=crossing

(Apologies for the missing images in these screenshots; please use your imagination.)

This change encourages mappers to provide information about a crosswalk’s signalization, which is important for routing (think ETAs). Both marked and unmarked crossings can be signalized or unsignalized. Prior to this change, mappers were encouraged to indicate whether there are pavement markings (which is primarily relevant to renderers); they could change the Type field to `traffic_signals` to indicate signalization, but at the cost of losing markedness, and iD would discourage such a change by reverting to the “Unmarked Crosswalk” preset (#5652).

This PR’s choice of `traffic_signals=yes` diverges from the much more common, documented [`crossing=traffic_signals`](https://wiki.openstreetmap.org/wiki/Key:crossing#The_crossing_tag) tag. However, `traffic_signals=yes` is the most widely used tagging scheme that’s compatible with the `crossing=marked`/`unmarked` scheme adopted in 1ffca3ce85d630539c7ebb11454a91770bd76536. The `footway=crossing` `traffic_signals=*` and `highway=crossing` `traffic_signals=*` combinations are already used [2,303 times](http://overpass-turbo.eu/s/EJU) worldwide, mostly in Europe. ([`traffic_signals:foot`](https://taginfo.openstreetmap.org/keys/traffic_signals%3Afoot#overview) is more common, but it’s primarily used in traffic sign mapping, disconnected from any roads or paths.) This PR is also consistent with @nbolten’s [draft proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/crossing%3Dmarked) for `crossing=marked`/`unmarked`.

As in https://github.com/openstreetmap/iD/issues/5205#issuecomment-410784580, I was initially concerned that the use of `traffic_signals=*` on a crosswalk/roadway intersection node could cause the node to be interpreted as a vehicular traffic light by data consumers. However, vehicular traffic lights and crossing signals are mutually exclusive: in the absence of `highway=traffic_signals`, `traffic_signals=*` expresses a “has a” relationship, not an “is a” relationship. Moreover, for there to be a conflict, the node would have to be tagged `highway=traffic_signals;crossing`, which is a poorly supported, [officially discouraged](https://wiki.openstreetmap.org/wiki/Multiple_values#Avoid_MV_tagging_if...) tagging style.

This PR only deals with traffic control devices that require pedestrians to wait. Traffic control devices requiring motorists to wait would be covered by a documented [`flashing_lights`](https://wiki.openstreetmap.org/wiki/Key:crossing#Additional_tags) tag that can be added in a separate PR. Other follow-up improvements could include support for the `traffic_signals:sound`, `traffic_signals:vibration`, and `button_operated` tags.


bhousel:
I took a look at https://wiki.openstreetmap.org/wiki/Key:traffic_signals
and now I'm furious.

Forget it.

There is no way I'm going to support `traffic_signals=yes` for pedestrian signals, after the wiki folks aren't even ok with iD using `traffic_signals=signal` for a normal traffic signal - a tagging that was accepted just not very widespread before iD started doing it.

The OSM Wiki needs to end.   Seriously.   It's ruining this project.

1ec5:
You’re right, that page was over the top, so I [reworded it](https://wiki.openstreetmap.org/wiki/Special:Diff/1761409). I’ll go back to the drawing board and find a less controversial way to let mappers indicate signalization, because I continue to believe that it’s a more important aspect of a crosswalk than the present of pavement markings.

PR #5659, opened 2018-12-28
Add sidewalk field preset for highways

SilentSpike:
Tagging sidewalk availability is valuable for pedestrian routing so I thought it was appropriate to make it easier by adding a field for it to iD.

A couple of thoughts came up as I tested this was working:

- I've limited the `"options"` to those that make sense (`no`/`left`/`right`/`both`/`separate`), but perhaps I should specify them using `"strings"` to prevent users typing in arbitrary values?
- I also wonder if there's a good way to explain what `separate` means here (mapped as a separate way) without requiring the user to go to the wiki page (is there a way to add a tooltip to dropdown options?).
- Perhaps this field shouldn't be in the service road preset? I can't think off the top of my head of any service roads with sidewalks.
- You'll also notice I opted for `no` over `none` as it has a higher usage [according to taginfo](https://taginfo.openstreetmap.org/keys/?key=sidewalk#values), but have no strong feelings on this decision.

I considered the possibility of something more complicated like the cycleway field preset since further specification of surface/bicycle access/etc. can be given. However, I figure the most valuable information is the presence/absence of sidewalks and that could always be developed later on if desired.


bhousel:
I think I'd rather see them mapped as separate ways, sorry.

SilentSpike:
I actually agree 🤷‍♂️  However, I think more people are willing to add this valuable information in the form of tags now. Then as more things get mapped more people get interested in adding more detail in the form of separate ways.

Issue #5699, opened 2019-01-12
right margin for keyhints in rtl locales


iriman:
left margin present in ltr locales:

![en](https://user-images.githubusercontent.com/8467903/51071719-e53f8300-166a-11e9-9ab2-b315a68876e2.png)

right margin is needed for rtl locales:
![fa](https://user-images.githubusercontent.com/8467903/51071718-e53f8300-166a-11e9-98c2-f30a035e9d84.png)

Proposed fix:
```
[dir='rtl'] .tooltip-inner .keyhint {
    margin-left: 0;
    margin-right: 5px;
}
```

PR #5700, opened 2019-01-12
margin-right for keyhints in rtl locale

iriman:
re #5699


bhousel:
Thanks again @iriman I really appreciate it 👍

Issue #5687, opened 2019-01-08
Fix a few misaligned labels with right-to-left layout


iriman:
Some parts of these lables are hidden with Persian localization. Could you please adjust them. thanks

- [x] This appears when we search for a feature. (`geocoder.search`)

![2](https://user-images.githubusercontent.com/8467903/50834159-92af4f80-1368-11e9-95ce-bc8e1240ecf0.png)

- [x] This one is title of turn restrictions help. (`help.field.restrictions.title`)

![1](https://user-images.githubusercontent.com/8467903/50834160-9347e600-1368-11e9-9736-4149c4bd49fd.png)

PR #5704, opened 2019-01-13
fix for some rtl issues

iriman:
fix for #5705 and second part of #5687

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2275,6 +2275,12 @@ div.combobox {
     padding-left: 10px;
 }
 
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks again @iriman !

By the way, for stuff like the membership editor heading section, it's usually ok and a bit simpler to switch the original padding rule to something like `padding: 0 10px` (padding on both sides) instead of adding a rtl rule with `padding-right` to override the original `padding-left`.   What you did is ok too  👍

iriman:
Ah yes :+1: thanks

Issue #5692, opened 2019-01-09
rtl & right-align for tooltips of "edit menu" in rtl locales


iriman:
![untitled](https://user-images.githubusercontent.com/8467903/50917356-8dc6ca80-1452-11e9-8d5e-e4ee11cdeb05.png)

bhousel:
I think #5708 should close this?

PR #5708, opened 2019-01-14
[RTL] place tooltip under #id-container to inherit rtl

iriman:
re #5692


bhousel:
Good catch, thanks @iriman 👍

iriman:
really finally! after a bit struggling :)

bhousel:
> really finally! after a bit struggling :)

haha I couldn't tell!   You're doing great  🙇

Issue #5611, opened 2018-12-16
Place of Worship preset: "Religion" field disappears after filling-in some data


my-na-me:
see movie:
![pow](https://user-images.githubusercontent.com/27923772/50050608-7836dd00-0100-11e9-800b-dd98a63c4bd9.gif)

bhousel:
cc @quincylvania

qsmmorgan:
Hi @my-na-me! So it looks like it’s working as intended, but I can see how it is confusing.

When you set the religion, the preset subtlety updates from Place of Worship to Church, which has the same fields except for Religion. We might improve this by calling the latter preset Christian Church and/or keeping the Religion field around so people can easily view or correct the religion tag.

quincylvania:
Hi @my-na-me! So it looks like it’s working as intended, but I can see how it is confusing.

When you set the religion, the preset subtlety updates from Place of Worship to Church, which has the same fields except for Religion. We might improve this by calling the latter preset Christian Church and/or keeping the Religion field around so people can easily view or correct the religion tag.

bhousel:
> When you set the religion, the preset subtlety updates from Place of Worship to Church, which has the same fields except for Religion. We might improve this by calling the latter preset Christian Church and/or keeping the Religion field around so people can easily view or correct the religion tag.

Oh yeah good catch @quincylvania ..  I like the proposed solution to make these presets less confusing.

RudyTheDev:
I think the preset selector button should visibly highlight itself briefly whenever it gets changed other than directly.

bhousel:
I made these changes per above comments:
- Make sure the "religion" field will not disappear (I added it for all the religion-specific amenities)
- Make sure that the religion appears in the preset label (e.g. renamed "Church" -> "Christian Church")

PR #5712, opened 2019-01-14
Preset field inheritance

### LINKED ISSUE #5710: Preset field inheritance
There is a lot of duplicated field data in the presets. For instance, most of the `shop` presets have the same fields:

```JSON
    "fields": [
        "name",
        "operator",
        "address",
        "building_area",
        "opening_hours",
        "payment_multi"
    ]
```

If we want to change this default list, we need to do it dozens of times to be consistent. This issue is compounded by the new `moreFields` property.

The presets should inherit their `fields` and `moreFields` properties from the "parent" preset if they are not manually defined. For instance, `shop/convenience` should automatically take the fields of `shop` if the user doesn't provide more explicit values.

A special field value such as `{inherit}` could allow for adding onto the inherited fields, like so:

```JSON
{
    "icon": "maki-clothing-store",
    "fields": [
        "{inherit}",
        "clothes"
    ],
    "geometry": [
        "point",
        "area"
    ],
    "tags": {
        "shop": "clothes"
    },
    "name": "Clothing Store"
}
```

An alternative solution would be to allow targeting a specific preset, but I think "parent" inheritance fits most cases.

quincylvania:
Closes #5710.

The biggest user-facing implication of this change is that, by default, specified presets will retain the generic fields that define them. So the `shop` field will be present on essentially all `shop` features. I actually don't mind this behavior. It let's users quickly fix mistagged or accidentally-tagged features (see #5611), and more clearly indicates how the tagging/preset system works. If we do go this route I think it'd be wise to rename fields named `Type` with more specific labels.

screen shot 2019-01-15 at 10 31 13 am

**Todo**
- [x] Update the documentation


bhousel:
Hmm, can we resolve these in preset.init() instead of at build time?   I imagine that people who are replacing them with a custom set of presets would want to use this feature too..

quincylvania:
> Hmm, can we resolve these in preset.init() instead of at build time? I imagine that people who are replacing them with a custom set of presets would want to use this feature too..

Oh that's a good idea, thanks! Let me know if anything else about this needs work.

bhousel:
> The biggest user-facing implication of this change is that, by default, specified presets will retain the generic fields that define them. So the shop field will be present on essentially all shop features. I actually don't mind this behavior. 

Hmm I kind of don't like this part, but mainly because the tags themselves are arbitrary and messy and we can't easily translate them.  In the example shown above, showing "Shop Type" of `doityourself` would be confusing to both non-English and English speakers.  I'm imagining that people who don't know any better might try to replace that with "do it yourself", thinking that they are helping. 

I see the main purpose of iD's preset system as: hiding the ugliness of OSM tagging from most users.

I really like the idea of preset field inheritance though!  It would make a lot of the preset definitions much simpler.

You said in #5710:
> An alternative solution would be to allow targeting a specific preset, but I think "parent" inheritance fits most cases.

I wanted to think about this a bit before responding, and I really like this idea of targeting a specific preset a bit better than the `{inherit}` keyword.  The reason is because of how arbitrary OSM tags are.  We want to inherit the "building like" fields (address, opening hours) for `shop=*` but also lots of `amenity=*`, `tourism=*`, `leisure=*`.. 

I don't see OSM tagging as having a clear inheritance where, e.g. all the amenities get some fields and all the leisure ones get different fields.  It's all mixed up.

Would you be ok with - instead of `{inherit}` - allowing a preset id to appear in there?   
The preset ids are automatically generated from the file path and name, so for example, all vending machines could inherit fields from `{amenity/vending}`.

quincylvania:
> I'm not sure what the best method is for ignoring the "Type" fields of target presets. I might strip the target's fields that map to the tags defining the source preset, but I'm not sure that's desirable in all cases.

I ended up doing this. I'd consider changing it back once we can figure out translatable labels for common `combo` field values. It's not exactly intuitive that some fields are inherited but not all. For now I think it works alright. This PR is ready to merge on my end.

screen shot 2019-01-15 at 1 55 29 pm

PR #5714, opened 2019-01-15
taginfo.json: remove unused embankment tags

goldfndr:
These embankment tag values aren't offered in the dropdown list.


bhousel:
Weird!  We should figure out how they got there.  Because `taginfo.json` is generated by [`build_data.js`](https://github.com/openstreetmap/iD/blob/master/build_data.js), I'm afraid that editing the file like this won't actually fix the underlying problem, so I'm going to close this PR and move it to a new issue instead.

PR #5716, opened 2019-01-15
Switch brand suggestion preset generation over to inherit fields

bhousel:
This changes the brand suggestion preset generation over to use the new inherit fields feature from #5712 - @quincylvania can you take a look?

This also attempts to include `cuisine` and `vending` tags in the `presetID` - in case a more specific preset exists for those things.  This fixes a minor annoyance where by only matching `key/value`, the code would find the "fast food" preset instead of the "chicken fast food" or "pizza fast food" preset.

It means that now Chick-fil-a has a chicken icon 🍗 and Sbarro has a pizza icon. 🍕


quincylvania:
@bhousel Great! I'd go ahead and merge.

PR #5717, opened 2019-01-16
Add wiki links from tag description

nyurik:
This feature requires P31 property to be present
in the data item. For an example, see
https://wiki.openstreetmap.org/wiki/Item:Q104#P31

Once the bot populates them in the data items,
it will automatically work for all existing keys/tags.

image

The above text has been changed:
• View documentation  (if the wiki page is in the user's language)
• View documentation in English (if the wiki page is in English, and the current language is not English)

### Manual Testing
Add a `bridge:movable` key, and click on the info icon next to it.


bhousel:
> View documentation in French (this is shown when neither local nor EN version is available. The language name will always be shown in English, even if the rest of the string is translated)

I'm pretty strongly against this option.  Not only is it difficult to localize the strings to display the "View Documentation in French" - but it's not very useful to users to offer to show them documentation in a language that they don't speak.

nyurik:
Per my message on Slack, in the modern world, Google translate is usually good enough to read any language -- this is something I do almost daily. That said, since you are so insistent on not having it, removed.

bhousel:
Thanks @nyurik 👍

Issue #4812, opened 2018-02-17
support {header()} in Custom Background URLs


jidanni:
Please see https://help.openstreetmap.org/answer_link/55381/ .

jidanni:
Also
> Normally JOSM ignores any elements that it does not recognize. However, it will skip the entire imagery entry, when one of the unsupported tags has the attribute mandatory='true'.

So iD, instead of sending such {header()} etc. unknown elements  down the pipe, still hanging on to the back of the URL, should at least strip them...

bhousel:
Unfortunately that's not really possible to adjust the referrer header from within a web application.  The browser sets it for us and we can't change it.  Sorry!

This also came up for discussion on #3420

jidanni:
OK, but it "behooves" you to throw an error:
```
Dear User: you tried to use a directive that we did not understand or
cannot support, within your URL template.

Rather than sending your directive, still embedded in the URL
uninterpreted, to the server, which would be ** NEVER what your intended
** and indeed ** possibly quite embarrassing ** and just plain looks silly too, please instead clean up
your URL template.

```

bhousel:
> OK, but it "behooves" you to throw an error:

Isn't raising errors about custom imagery covered in one of the other half dozen issues you just opened?

jidanni:
I don't see any for "catch unsupported URL template directives (that
should never be sent raw)'

PR #5738, opened 2019-01-19
Replace wkid/w/s/n/e in WMS URL templates

1ec5:
Added support for `{wkid}`, `{w}`, `{s}`, `{n}`, and `{e}` tokens in WMS-style background imagery templates, as described in [this JOSM documentation](https://josm.openstreetmap.de/wiki/Maps#WebMapServicesWMS).

The `{wkid}` token is particularly handy for `exportImage` endpoints in ArcGIS ImageServer services. (In fact, it appears multiple times, whereas the previous code only replaced each token once.) Currently, osmlab/editor-layer-index#611 works around the lack of this token by hard-coding a projection, but JOSM can use the `projection` property correctly: https://github.com/osmlab/editor-layer-index/commit/95680b2a3b9f468748365deb37943c7f1f871955#r31964799. Once this PR lands, we can modify the editor layer index’s tests to accept `{wkid}` as an alternative to `{proj}`.

Thanks to @don-vip for making me aware of these additional tokens.

### DIFF
--- a/modules/renderer/background_source.js
+++ b/modules/renderer/background_source.js
@@ -124,13 +124,33 @@ export function rendererBackgroundSource(data) {
                 }
             }).bind(this);
 
+            var tileSize = this.tileSize;
+            var projection = this.projection;
             var minXmaxY = tileToProjectedCoords(coord[0], coord[1], coord[2]);
             var maxXminY = tileToProjectedCoords(coord[0]+1, coord[1]+1, coord[2]);
-            return template
-                .replace('{width}', this.tileSize)
-                .replace('{height}', this.tileSize)
-                .replace('{proj}', this.projection)
-                .replace('{bbox}', minXmaxY.x + ',' + maxXminY.y + ',' + maxXminY.x + ',' + minXmaxY.y);
+            return template.replace(/\{(\w+)\}/g, function (token, key) {
+              switch (key) {
+                case 'width':
+                case 'height':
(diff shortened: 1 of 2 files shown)


1ec5:
The only other unsupported token is `{header(…, …)`, which was requested in #4812. Apparently the Fetch API does support [custom headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers), but that’s different enough that it should be a separate PR. Moreover, #4812 requested `Referer`, which is [immutable](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name). (I’d expect that overriding the referrer would only be necessary if one is using an endpoint without permission.)

bhousel:
> The only other unsupported token is `{header(…, …)`, which was requested in #4812. Apparently the Fetch API does support [custom headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers), but that’s different enough that it should be a separate PR. Moreover, #4812 requested `Referer`, which is [immutable](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name). (I’d expect that overriding the referrer would only be necessary if one is using an endpoint without permission.)

The problem here is that we don't request the imagery via XHR or Fetch - the browser requests the imagery when we create `` tags.

1ec5:
Ah, good point. 😅 So in other words, this PR gets us to parity with JOSM to the extent possible.

Issue #5683, opened 2019-01-07
Add ImproveOSM Q/A layer


bhousel:
Following #5201 - we can add a Q/A layer for [ImproveOSM](https://improveosm.org/)

The three types of errors that it detects are - possible missing oneways, possible missing turn restrictions, and possible missing roads.

There is no documentation for the API, but it seems very simple and similar to the KeepRight API (get issues, mark as fixed/invalid).  They use a fork of an older version of iD, and we can just inspect the network to see with their forked iD does.  

cc @mvexel 

<img width="935" alt="screenshot 2019-01-07 13 36 29" src="https://user-images.githubusercontent.com/38784/50786328-40473380-1281-11e9-86b0-5e9278409d75.png">

boothym:
This would be a good addition, as the improve-osm iD fork is outdated.

SilentSpike:
I have a semi-working implementation of this. Currently only retrieving missing one-ways and turn restrictions as they're easy to locate a specific point for.

Missing geometry issues are presented as a tile filled with points by ImproveOSM, assuming we want to be consistent across services and have errors displayed as an icon at a point, my current line of thinking is to just take an average of those (similar for one-ways which are found as way segments between two given nodes and shown as arrows, currently I just use the node at one end, but the midpoint makes more sense unless we want to display them differently).

Obviously the KeepRight icon shown here is just temporary, if anyone has a more suitable one for ImproveOSM let me know 😃 

![image](https://user-images.githubusercontent.com/5459452/51441250-9a0b1d00-1cc7-11e9-84e3-18c7f44b4fdc.png)

![image](https://user-images.githubusercontent.com/5459452/51441181-cbcfb400-1cc6-11e9-89ce-8715d01ed34c.png)

bhousel:
done in #5739 🎉

PR #5739, opened 2019-01-21
Add ImproveOSM Q/A Layer

SilentSpike:
Ref #5683 

This PR is still WIP, but opened for progress tracking and collaborative purposes.

My current priorities:
- [x] Query all error types
- [x] Render errors in viewport
  - [x] Add an appropriate icon
- [x] Display error details to user
- [x] Implement error resolving (mark as false/fixed)
- [x] ~~Implement error commenting~~ (appears to be non-functional)
- [x] Cleanup and feedback incorporation

There's a lot of similar and duplicate code to the KeepRight implementation that could probably be made more general and shared between the two (and future services). I could probably do this too, but would rather get this working first then worry about that in a later PR.


SilentSpike:
Doing some more work on this today and it actually seems like the ImproveOSM comment system doesn't work. Can anyone else try it on their site and confirm this?

bhousel:
Thanks @SilentSpike !   This looks pretty great.. FYI we are preparing an iD release for this week, so it won't be in January, but I'd like to try to get it into the February release.

SilentSpike:
All critical functionality is in, really just needs implementation cleanup and decisions (i.e. do we want all future QA layers to have consistent UX or each have their own icons/behaviour).

For the time being I've made these errors render the same as points with icons (but coloured). From my latest commit message:

> This allows more diverse representation of the error subject at a glance than relying on colour alone.
> 
> However, it would be good to have a generic error icon that can contain icons which is differentiated from the point icon for clarity.
> 
> Sidebar header currently still uses the bolt icon until I figure out how best to deal with that. Also the font awesome icons don't seem to work, perhaps there's additional code needed for those that I've missed.

Here's a preview of what that looks like.

image

bhousel:
> All critical functionality is in, really just needs implementation cleanup and decisions (i.e. do we want all future QA layers to have consistent UX or each have their own icons/behaviour).

Awesome @SilentSpike , looking forward to trying this out.  I think it's ok for the different Q/A layers to have their own icons for now.  We probably won't add more than a handful.   We might be able to generalize what the sidebar does.

SilentSpike:
@bhousel I think different icons are actually useful to see which service is being interacted with.

My main line of thinking with the more generic icon + detail icon implemented here is geared towards future Osmose support. It's similar to this service in that there's no distinguishable icon to use, but unlike this there are many more error types that could potentially be supported in future to differentiate between.

In terms of the current status of this, the current desirable list in my head ordered by ratio of significance/ease:
- [ ] I should probably bump turn restriction error positions slightly so they're not right on the junction node. An easy QOL improvement.
- [ ] Review UI strings
  - [ ] Could add identified start/end nodes to one-way errors
  - [ ] Decide if happy with wording and information given
  - [ ] Need to update the help string that explains QA layers to reflect this implementation
- [ ] Believe I may have unsafely used the `userDetails` function to get OSM username (it's also possible this isn't actually needed to post updates to the errors and because commenting doesn't work might be worth checking that).
- [ ] Decide whether we want to track the errors closed using their external identifiers (they're based on unique features of each error, not very pretty readable numbers like KeepRight).
- [ ] More appropriate error positioning for oneway errors, may be beyond my current ability. Currently just takes the average of the start and end of the way segment in question as the API only serves these lat/lon. This works fine for relatively straight segments, but if the segment is curved it can become far off the way.
- [ ] Decide if we want to add links to the error on ImproveOSM (no way that I see to link directly to it, but we could link to the location or something).

Beyond that:
In future I could see all services defining their errors as children of a base error class which has the properties common to all (position, name, description, category, subcategory, icon, external ID).

In this way a lot of the code could be made general to work with these common values, then there could be some cross-service consistency in things like colouring (based on category) to aid workflow (e.g. if someone wanted to investigate all navigation related errors they'd just look for a certain colour).

bhousel:
I played around with this a bunch today.. I think it's very nearly ready to merge 🎉 

> My main line of thinking with the more generic icon + detail icon implemented here is geared towards future Osmose support. 

I agree, this is a good approach..

> * [ ]  Review UI strings - Decide if happy with wording and information given

They look pretty good to me.. What do you think about calling it an ImproveOSM "Issue" or "Detection" instead of "Error".. Possibly also include some way to show the user the confidence score.  There are a lot of false positives in this dataset and I don't want to give users the impression that these are all real errors that need fixing.

> * [ ]  Believe I may have unsafely used the `userDetails` function to get OSM username (it's also possible this isn't actually needed to post updates to the errors and because commenting doesn't work might be worth checking that).

Yes, I tried this on the main ImproveOSM site and I couldn't get it to work there either.  It's ok to leave the commenting feature out for now.

> * [ ]  Decide whether we want to track the errors closed using their external identifiers (they're based on unique features of each error, not very pretty readable numbers like KeepRight).

Ah yeah I noticed this too.  It might be best to just invent an id string out of the `loc` of the error, like "-74.00765,40.96369"  (truncate to 5 decimals?).  Though if users close a lot of issues we'll overrun the 255 char limit in the changeset tag.  Maybe this is a good use for [open location codes](https://www.npmjs.com/package/open-location-code)?  They'd be a slightly shorter than coordinate pairs.

> * [x]  More appropriate error positioning for oneway errors, may be beyond my current ability. Currently just takes the average of the start and end of the way segment in question as the API only serves these lat/lon. This works fine for relatively straight segments, but if the segment is curved it can become far off the way.

What you did looks pretty good for now!  I was thinking of drawing some shape or arrows over the whole way (for missing oneways) or over the from/via/to (for missing turn restrictions)  like on the ImproveOSM site, but we don't have to.

> * [ ]  Decide if we want to add links to the error on ImproveOSM (no way that I see to link directly to it, but we could link to the location or something).

Yeah without true stable ids, I don't see a good way to do it.   We can skip this part.


> Beyond that:
> In future I could see all services defining their errors as children of a base error class which has the properties common to all (position, name, description, category, subcategory, icon, external ID).

yes!
 
> In this way a lot of the code could be made general to work with these common values, then there could be some cross-service consistency in things like colouring (based on category) to aid workflow (e.g. if someone wanted to investigate all navigation related errors they'd just look for a certain colour).

Yes, this would be great to have some consistency.  FWIW, I did put a bunch of thought into the KeepRight colors.  I probably spent a few hours adjusting them so that they
- have enough contrast with each other and with map features
- reduce the overall number of unique colors by grouping similar kinds of errors
- chose colors based on how frequently the issues/markers occur on the map

Anyway, great work!  I'm happy to merge it whenever you feel that it's finished.

SilentSpike:
Rebased to master and squashed a few trivial commits in the process.

In terms of functionality I'm happy with what's here. The code itself can probably be improved/cleaner in places, but I figure let's get this into the master branch since it's working and improve from there.

I haven't used open location codes here as I didn't want to introduce a new dependency just for this, but perhaps worth using if they'll also be used elsewhere.

You'll probably want to review https://github.com/openstreetmap/iD/pull/5739/commits/e9397aa14f868151d8299d46ee4bb27ab2a1d22f in particular before merging as it's the first time I've had to properly consider how scope and callbacks work in javascript.

bhousel:
> In terms of functionality I'm happy with what's here. The code itself can probably be improved/cleaner in places, but I figure let's get this into the master branch since it's working and improve from there.

Yes, I agree.. I think it looks great (and PRs get harder to merge the longer they sit).

> I haven't used open location codes here as I didn't want to introduce a new dependency just for this, but perhaps worth using if they'll also be used elsewhere.

Sounds good. I might make an attempt at it this weekend...

> You'll probably want to review [e9397aa](https://github.com/openstreetmap/iD/commit/e9397aa14f868151d8299d46ee4bb27ab2a1d22f) in particular before merging as it's the first time I've had to properly consider how scope and callbacks work in javascript.

You did great 👍 iD is a fantastic project to learn on.  Also, ping me on Slack anytime if you ever have any questions, or are looking for more to do!  😄

mvexel:
@SilentSpike Do you still need someone to investigate the comments on the ImproveOSM side?

SilentSpike:
@mvexel Do you mean to fix the functionality on the ImproveOSM end?

We've confirmed that they don't seem to be functional, but if that is updated I'll happily add in the support on this end.

beataj-telenav:
The comment functionality should work for the ImproveOSM layers, at least in the ImproveOSM plugin works as expected. 
You can add a comment with an optional status change to a list of ImproveOSM entities by using the "**comment**" operation from the ImproveOSM API's. 

Example of comment operation requests:
  **Missing Roads:**  
    {
      "username": "test",
      "text": "There is no road there",
      "status": "INVALID",
      "targetIds":
       [
           {
          "x": 100,
          "y": 100
          },
       {
          "x": 100,
          "y": 101
       }
      ]
   }
     
 **One ways (Direction of flow)**
      {
        "username": "zero",
         "text": "There is no road there",
         "status": "INVALID",
        "targetIds":
         [
           {
            "wayId": 100,
            "fromNodeId": 100100,
            "toNodeId": 100200
           },
          {
           "wayId": 100,
            "fromNodeId": 100200,
            "toNodeId": 100300
           }
          ]
       }  

 **Turn restrictions** 
     {
       "username": "zero",
       "text": "There is no road there",
       "status": "INVALID",
       "targetIds":
            [
            "abc", "def"
            ]
       } 
Let me know if you need further assistance with the ImproveOSM API's.

SilentSpike:
@beataj-telenav Awesome thank you! I was only going by the requests I observed being sent by the improveosm.org editor so was totally unaware of the `retrieveComments` operation (and posting new comments weren't showing up there).

With this information I'll see if I can get commenting working here in iD 😃

beataj-telenav:
@SilentSpike :  I forgot to add a **retrieveComments** example for the One-way (Direction of flow) component. In this case the request would be: https://directionofflow.skobbler.net/directionOfFlowService/retrieveComments?wayId=229550180&fromNodeId=151993197&toNodeId=2381277816. 
Let me know if you have any further questions regarding the ImproveOSM API's or need any specification (parameters, status codes, available filters etc). We don't have any public documentation published for these components, but I can provide the necessary documentation for integration if needed.

Btw: great work with the integration, I'm glad that our components will be integrated also to the main iD version.

SilentSpike:
@beataj-telenav Managed to suss those out myself 😉 

image

I have working comments now, but one question for you: the timestamp served, what format is it in (as you can see it came out as the wrong date in the above screenshot)?

beataj-telenav:
@SilentSpike The timestamp is the number of seconds elapsed since 1970-01-01 00:00:00 UTC, not counting leap seconds. In order to display the correct value, you need to format it based on the current timezone. (In the JOSM plugin we first transform the value into milliseconds and then format based on the user timezone).

beataj-telenav:
@SilentSpike For the missing roads data in some cases, you might get -1 for the **numberOfTrips** field. This happens for data that came from a 3rd party. In the JOSM plugin for these cases, we display "not available".

Issue #2058, opened 2013-12-13
Straighten should work on selected subpaths, not be limited to straightening entire ways


WillPittenger:
To often I want to straighten part of a way, but not the whole way.  Or I may want to align parts of a street represented by multiple ways.  The existing straighten tool works only on ways and will straighten the entire way.

So if there are three or more nodes selected, but no ways, add a tool to the context menu that would let me straighten those nodes regardless of which ways they are part of.  (After the straighten operation is complete, it may be appropriate to look for redundant nodes.)

bhousel:
Done in #5740

PR #5740, opened 2019-01-21
Straighten multiple ways in one operation

jguthrie100:
Implements aspects of #2058 (multiple way straightening)

This change allows the user to select a number of joined ways and straighten them all in one operation.

- I've written a number of additional tests to cover some of the new scenarios

straighten


jguthrie100:
One thing to consider - should joined ways that go in opposite directions be able to be straightened together.
My changes only allow ways that go in the same direction to be straightened, but I could change it to allow for direction to not be an issue..

bhousel:
> Implements aspects of #2058 (multiple way straightening)

Nice, I'll have to check this a bit more thoroughly later this week 👍 

> One thing to consider - should joined ways that go in opposite directions be able to be straightened together.
> My changes only allow ways that go in the same direction to be straightened, but I could change it to allow for direction to not be an issue..

Sure - I don't think the direction matters.

jguthrie100:
Fixed it so that connected ways going in different directions can be straightened

straighten-multiple-directions

Apologies for the commit-history chaos!

bhousel:
Thanks @jguthrie100 , I just merged this..
(It doesn't show as merged on GitHub because I rebased it to avoid conflicts)

PR #5744, opened 2019-01-21
Update .gitattributes to use lf eol for .js files

SilentSpike:
Running "npm test" currently spews out many eol errors if you don't have git configured not to convert line endings to crlf.

As I understand it this change should override that (see: https://help.github.com/articles/dealing-with-line-endings/).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 * text=auto
 *.bat eol=crlf
+*.js eol=lf


bhousel:
Thanks @SilentSpike this seems ok!


> Adding an .editorconfig file can also ensure editors don't mess with the line endings too (and enforce proper spacing + trim whitespace, etc.).

Yes, this would be great too.  It's really just 4-space soft tabs and trim trailing whitespace.

Issue #4745, opened 2018-01-25
avoid invalid polygone


Marc-marc-marc:
Thanks for the new trip to avoid creating a self-crossing invalid polygon.
but I see that a user is still able to make a invalid polygon that make a "forward/backward" with a duplicate node
https://www.openstreetmap.org/way/90262899
duplicate node https://www.openstreetmap.org/node/5356991998
node "external" to the polygon https://www.openstreetmap.org/node/5356991994
it would be great if iD could prevent the user from creating invalid polygons

bhousel:
It's difficult to see what they did exactly, but there are still plenty of ways to create invalid polygons. 

Here's a situation where I dragged a midpoint from one line on an outer way over to a different line on the same multipolygon.  It probably shouldn't be allowed.

![invalid 2](https://user-images.githubusercontent.com/38784/35425400-9f8725e0-0228-11e8-8466-423789e065e6.gif)

Also, iD only checks geometry in drag node and drawing modes.  You can still select multiple things and just move them or rotate them to cause a self intersection.

![invalid](https://user-images.githubusercontent.com/38784/35425454-ef468616-0228-11e8-8b7e-e03c40887333.gif)

Marc-marc-marc:
Thanks for considering this issue.
I have sent a message to this user to ask how he does exactly.
Did you prefer that I leave the error for some days or a printscreen before fixing it or the history of the object is enough ?

bhousel:
> Did you prefer that I leave the error for some days or a printscreen before fixing it or the history of the object is enough ?

Looks like it was fixed in OSM already, but this is ok since I put some screen captures in the ticket.

mmd-osm:
_Merge_ is also bypassing self-intersection checks...

![peek 2018-01-27 21-34](https://user-images.githubusercontent.com/5842757/35476074-12fc9ed8-03aa-11e8-8582-01b406bef9ed.gif)

PR #5745, opened 2019-01-22
Add check to stop joining ways if resulting way intersects itself

jguthrie100:
Implements intersection check, as raised in #4745

join-intersect


bhousel:
This is fantastic @jguthrie100 !  Thanks and I appreciate you adding the tests too 🙇

PR #5816, opened 2019-01-27
Fix street imagery hover and selection behaviour

### LINKED ISSUE #5494: Mapillary yellow highlight icon doesn't persist
Small issue, but noticed some different behaviour between Bing Streetside and Mapillary.

If you click on a Mapillary marker to open an image, then select an OSM object or even move your mouse away the yellow highlight disappears. It only reappears if you mouse over a Mapillary trace again.

This does not happen with Streetside or OSC, where the selected image is always highlighted in yellow. Makes it much easier to see where the image is on the map, and where it's pointing.

Occurs in both Firefox and Chrome (Windows10).

SilentSpike:
Closes #5494 

~~So the cause of that issue was that whenever the svg layer `update()` method was triggered the "selected" and "highlighted" classes which get added by method `setStyles()` in the service are not being retained. The simple fix here is to just re-apply `setStyles()` with every update (there's likely a nicer way to just preserve them in the first place).~~

Since found to be false. See my comments below.

In testing, further issue with the interaction between multiple active street imagery layers was discovered. This turned out to be because the `reset` parameter of `setStyles()` was being used for style updates due to svg layer events (as far as I can see it is only needed for style updates due to interaction with the viewport - where the selected bubble/sequence is changed indirectly).

Beyond that I also changed the `mouseover` and `mouseout` events to `mouseenter` and `mouseleave` respectively. This is because the former fire for each child element of the svg group individually which is unnecessary.

There's still the issue of disabling any street imagery layer deselecting the current bubble which isn't part of that layer. However, I will open a separate issue for that (and PR if nobody beats me to it).

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -101,7 +101,7 @@
     cursor: pointer;
 }
 
-.viewfield-group.selected * {
+.viewfield-group.currentView * {
     fill: #ffee00 !important;
 }
 .viewfield-group.hovered * {
@@ -125,7 +125,7 @@
     stroke-opacity: 0.9;
     fill-opacity: 0.9;
 }
-.viewfield-group.highlighted.selected circle {
+.viewfield-group.highlighted.currentView circle {
     stroke: #222;
     stroke-width: 2;
     stroke-opacity: 1;
@@ -145,12 +145,12 @@
     stroke-width: 1;
     fill-opacity: 0.8;
 }
-.viewfield-group.highlighted.selected .viewfield {
+.viewfield-group.highlighted.currentView .viewfield {
     stroke-width: 1;
     fill-opacity: 0.9;
 }
 
-.viewfield-group.selected .viewfield-scale {
+.viewfield-group.currentView .viewfield-scale {
     transform: scale(2,2);
 }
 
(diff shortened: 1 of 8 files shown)


SilentSpike:
Demonstration of the fixed behaviour:
https://gfycat.com/ArcticUnfinishedInexpectatumpleco

SilentSpike:
You know, on further testing I actually don't think the `update()` function is the source of the issue. I think refreshing the style during it just appears to fix things because it updates so frequently.

The real source is *probably* the `select` class being removed unintentionally from multiple layers when deselecting an osm entity.

SilentSpike:
Confirmed, the actual source of the original issue is this line right here:
https://github.com/openstreetmap/iD/blob/49442afe0cbd81bb71775712ddbb88971167e693/modules/modes/select.js#L539-L541

Some debug shows that it's removing the class from all of these elements:
image

bhousel:
great analysis @SilentSpike  - yeah I think we will need to use a class other than `selected` for the imagery markers.

SilentSpike:
Done, this should now be good for review 👍

bhousel:
Looks great, thanks so much @SilentSpike 👍

PR #5826, opened 2019-01-29
Fix turn:lanes changing when way is reversed

### LINKED ISSUE #5674: Turn lanes reversed when reversing a way
When reversing the direction of a roadway, the values within the `turn:lanes`, `turn:lanes:forward`, `turn:lanes:backward`, and `turn:lanes:both_ways` tags are incorrectly reversed.

Steps to reproduce:

1. Draw or select a way tagged `highway=tertiary` `lanes=2` `turn:lanes:forward=right`.
1. Reverse the way’s direction.

Expected results:
`highway=tertiary` `lanes=2` `turn:lanes:backward=right`

Actual results:
`highway=tertiary` `lanes=2` `turn:lanes:backward=left`

It’s correct that `forward` became `backward`, but not that `right` became `left`. The left/right replacement behavior is intended to maintain consistency in tags that indicate a side of the road. However, the “left” and “right” in `turn:*` tags indicate maneuvers, not sides of the road. The `backward` direction before the reversal is the same as the `forward` direction after the reversal, so the tag shouldn’t start referring to a different connecting way.

This issue only reproduces with the values `left` and `right`, not with more complex `turn:lanes` values, such as `left|right` or `slight_left`, making this issue more surprising when it does occur. It’s especially subtle because turn lanes are only surfaced in the raw tag editor.

SilentSpike:
Closes #5674 

Couldn't think of any other keys where left/right value is relative to the key and not the way direction.

### DIFF
--- a/modules/actions/reverse.js
+++ b/modules/actions/reverse.js
@@ -20,6 +20,7 @@ References:
 export function actionReverse(wayID, options) {
     var ignoreKey = /^.*(_|:)?(description|name|note|website|ref|source|comment|watch|attribution)(_|:)?/;
     var numeric = /^([+\-]?)(?=[\d.])/;
+    var turn_lanes = /^turn:lanes:?/;
     var keyReplacements = [
         [/:right$/, ':left'],
         [/:left$/, ':right'],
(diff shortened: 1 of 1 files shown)


bhousel:
> I was wondering, why not just add it to the `ignoreKey` regex?

Actually nevermind, that thing already looks plenty complicated..

SilentSpike:
@bhousel I totally could have 😅 For some reason it didn't click that the `ignoreKey` regex was used to ignore values of those keys and not the keys themselves

SilentSpike:
We might actually need some special code here anyway though as I think `turn:lanes:both_ways` (or `turn:lanes` on a non-one-way road) might actually need special behaviour (the lane order is reversed when the way direction is reversed, therefore the `left|straight|straight|right` needs to also be reversed?)

Issue #3130, opened 2016-05-23
QA tools/warnings


aawiseman:
There was some discussion on the HOT email list about some potential QA tools or QA-related warnings in iD that would be helpful:
* [x] 1. when there are non-squared buildings (especially when there are a lot of them) -- #2624 and #3085 mention this
* [ ] 2. duplicated/overlapping polygons of the same/similar tags (which can be an issue when people map outside their squares, or just sloppy/quick editing)
* [x] 3. roads overlapping/going through polygons, especially buildings
* [X] 4. disconnected roads (that are just floating and not connected to anything, which is bad for HOT responses and for rendering, mentioned in #1281) (#3786)
* [x] 5. roads that cross and should intersect but aren't connected to each other
* [ ] 6. roads that are connected with buildings (I think due to snapping) - #476 mentions this, maybe it's not an error? But to me a road shouldn't be coincident with a building edge, which sometime happens

aawiseman:
Oh another one, it might be good to highlight features that have fixme= in them!

bhousel:
I'm all for adding validation to iD where it makes sense, but I don't want to overwhelm new users with a bunch of messages about things that they won't understand, and might not even be responsible for. 

Dedicated QA tools like Keepright and Osmose do great jobs at catching a lot of these conditions, and HOT has a pretty effective structure in place to have new mappers' edits reviewed by experienced validators.  

The last thing I want to do is turn away enthusiastic new mappers by nagging them about their work.  

But I think this list is a good start.. I'll think about these items and see if we can make some improvements, either in the iD training materials or the pre-save validations that iD runs.

aawiseman:
I think in the pre-save validations would be great.

Or something related, maybe a little popup that comes up as you are mapping, for example if you do ten buildings in a row that aren't squared. That may be a heavier lift, however.

quincylvania:
So a lot's changed in validation since this issue was posted!

- We've done points 1, 3, 4, and 5.
- I think #5977 covers point 6.
- Point 2 is already covered for buildings and will be covered for water by #7050. I'm not sure that other types of area crossings are a large concern but we can open separate issues for them if needed.

I'm closing this since remaining actionable items are pretty well covered by other issues 👍

PR #5830, opened 2019-01-29
Issue manager and advanced validation

### LINKED ISSUE #1669: Validation: crossing ways, same layer, no common node
> crossing ways without common node at the same layer

Especially when both share common type, eg. both highways or waterways (most common).

via #1657

### LINKED ISSUE #4591: Should iD just offer to upgrade tags?
In #4585 we added a bunch of new Public Transport presets.  iD will still recognize the legacy tags on a feature, but users really wouldn't know to change that preset to the newer Public Transport version.

Should we just add a button or something to let the user know if they are looking something that can be upgraded?  Should we just silently upgrade the tags? 

This would be for situations like in Public Transport and Healthcare tagging where there is an old tag and a new tag and it's totally clear how to do the upgrade.. 

e.g. `highway=bus_stop` -> `public_transport=platform + bus=yes + highway=bus_stop`

### LINKED ISSUE #5765: Adding a multipolygon should open the area preset browser
Currently when you select Multipolygon from the relation preset browser you are brought to the entity editor and have to go back to select the area preset. We should simplify this by bringing up the area preset browser immediately after selecting Multipolygon.

The will be especially important once the validation branch is merged. The user will be see an error on `type=multipolygon` relations with no other tags, but it does not make sense to warn them for new features.

### LINKED ISSUE #5870: Warning for relation without a type tag
Relations must have a [type](https://wiki.openstreetmap.org/wiki/Key:type) tag. iD should warn users when this is missing. [KeepRight does this already.](https://wiki.openstreetmap.org/wiki/Keep_Right/180_relations_without_type)

See #5830 for upcoming UI improvements that will surface issues immediately as you edit.

quincylvania:
**Overview**
- Issues pane as the central hub for validation
- Issues list for the selected feature in the editor sidebar
- "Quick fix" buttons to assist in resolving issues
- Instant validation while editing
- Crossing ways and almost junction warnings via @gaoxm

**To-Do Before Merge**

- [x] Validate upon restoring after page reload
- [x] Add tooltip to Upload button when disabled saying that errors (but not warnings) are blocking the upload
- [x] Correct the "Connect the features" fix for the crossing ways issue
- [x] Fix code tests
- [x] Restore "too many deletions" warning
- [x] Add issue counts to the list titles
- [x] Zoom to ways and relations when clicking their issues
- [x] Add "everything is okay" message to the Issues pane when no issues are detected
- [x] Fix pane list tooltip placement
- [x] Add appropriate tags to junction of crossing ways connection fix (e.g. ford, level crossing)
- [x] Improve crossings ways validation performance
- [x] Style fixes as clickable

**To-Do Eventually**

- [ ] Build "All features" validation mode where issues with everything nearby or visible is flagged, not just edited features
- [ ] Re-validate features only as-needed due to further edits
- [x] Don't count vertices toward "many deletions" warning
- [x] Quick-fix to close way when tags imply area
- [x] Use preset data for "tags imply area" validation
- [x] Don't show almost junctions issue if a connection would cause self-intersection
- [x] Use existing nodes where possible on almost junction fix
- [x] Use existing nodes where possible on crossing ways fix
- [x] Use warnings instead of errors for features the user didn't create or modify this session
- [x] Highlight corresponding entities when hovering over an issue (like with relations)
- [x] Show issue badge on the pane toggle button
- [x] Icons for fixes
- [x] Recommend fixes even if they can't be performed automatically
- [ ] Show warning and error badges on the map
- [ ] Re-think issue display in the commit sidebar
- [ ] Add list of all rules (issue types) and allow toggling certain rules on and off
- [ ] Integrate KeepRight and ImproveOSM into the Issues pane/system?

**Issues**
- Closes #1669
- Closes #4591
- Closes #5765
- Closes #5870
- Completes some of #3130

**Special Thanks** to @gaoxm, @wonga00, @abalosc1, @maxgrossman, and @brianhatchl for code and feedback!

screen shot 2019-02-04 at 1 51 16 pm
screen shot 2019-02-14 at 4 16 54 pm
screen shot 2019-02-14 at 4 15 32 pm

upgrade tags demo

screen shot 2019-02-04 at 1 51 43 pm


quincylvania:
> For the "point has no descriptive tags" issue - will this warning appear when for example, only a name tag has been added?

Yes! This is actually the existing behavior for the "untagged point" warning you may see when uploading changes.

> And what happens with amenity=swimming_pool tag if it is also tagged building=*? Because that is likely to be leisure=sport_centre + sport=swimming instead.

For deprecated tag warnings, we're going off of the wiki and real usage as much as possible. The wiki recommends `leisure=swimming_pool` as a replacement: https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dswimming_pool

It also doesn't appear that `building` is a common combination so I'm not worried too much about it! https://taginfo.openstreetmap.org/tags/?key=amenity&value=swimming_pool#combinations

quincylvania:
@bhousel I'm feeling good about the state of this PR! Take a look and let me know if anything needs changing. There are still lots of ways we can improve validation but I'd say let's merge this and then open issues for the remaining ideas.

SilentSpike:
Testing this branch out a bit this evening, I have some immediate UX feedback.

image

When I see this in the sidebar, it's not obvious that "Connect the features" is an optional automated action I can select. Additionally (and more importantly), it comes across like iD is telling me to connect the features to solve the issue (which might not even be the right thing to do).

Experienced users may pick up on this, but this could mislead someone using iD for the first time.

SilentSpike:
On a more positive note, I love how seamless this all is 😃 Look forward to an "all in view" mode and hopefully contributing some validations in future 👍

quincylvania:
> When I see this in the sidebar, it's not obvious that "Connect the features" is an optional automated action I can select. Additionally (and more importantly), it comes across like iD is telling me to connect the features to solve the issue (which might not even be the right thing to do).

Good point, @SilentSpike. How about something like this?

screen shot 2019-02-12 at 8 38 45 am

With the "Quick Fixes" label, we hopefully imply that these are assistive buttons but not required actions. We can give more detail in a tooltip when hovering over the label. I also like the idea that the entire top area of the issue UI doesn't need to be a button.

modules/validations/maprules.js
```diff
@@ -2,6 +2,9 @@ import { services } from '../services';
 
 
 export function validationMaprules() {
+    validation.type = 'maprules';
```
quincylvania: This probably shouldn't be before the declaration of `validation`.

bhousel: oh thanks sorry - I was going quickly and didn't realize it wasn't like the others.

modules/ui/entity_issues.js
```diff
@@ -117,12 +117,6 @@ export function uiEntityIssues(context) {
             .merge(itemsEnter)
             .classed('expanded', function(d, i) { return i === _expanded; });
 
-
-        items.select('button.label')     // propagate bound data
-            .attr('title', function(d) {
-                return t('issues.' + d.severity + 's.icon_tooltip');
-            });
-
```
quincylvania: FYI the `title` tooltip was originally just on the icon, to help correlate the icon with the severity title. Not sure how it ended up on the whole label. We probably don't need it anyway.

bhousel: Nice - that's what I assumed.. It makes some sense on the icon part only, but not really necessary since these icons aren't for clicking.

quincylvania: I was vaguely thinking that the icon could be the "zoom-to-issue" button too. This would make more sense once we show the badges on the map, since the icons would be the same. But even in this case we wouldn't use the same tooltip text.

quincylvania:
screen shot 2019-02-14 at 11 28 10 am

@bhousel @SilentSpike What about a fix list like this? This way we can recommend ways to resolve the issue without quick actions, and fixes with actions appear clickable.

PR #5836, opened 2019-02-01
do not close image viewer if street level imagery layers are still rendered

### LINKED ISSUE #5829: Disabling street imagery layer always closes viewport
As per my findings while working on #5816

If multiple street imagery layers are active and the viewport is currently showing an image from one layer, disabling any other street imagery layer closes the viewport when it shouldn't.

Expected behaviour:
- Viewport remains open unless currently shown image is part of the disabled layer.

maxgrossman:
closes #5829

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -264,7 +264,8 @@ div.hide,
 form.hide,
 button.hide,
 a.hide,
-li.hide {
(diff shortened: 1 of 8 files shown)


bhousel:
Cool thanks!  👍

PR #5855, opened 2019-02-04
Use a generic QA error object

SilentSpike:
This basically introduces a new `qaError` object to replace the `krError` and `iOsmError` objects.

This object require a `service` property to be initialised so that it can be interacted with generally (if we define the same generic functions in each service).

Additionally if `error_type` is initialised data can be read from `data/qa_errors.json` where I'm currently specifying an `icon` and `category` (only for ImproveOSM at the moment). Category is currently unused, but I'd like to convert error colouring to be based on this for some cross-service consistency.


modules/services/improveOSM.js
```diff
@@ -210,12 +210,13 @@ export default {
...
 
-                                var d = new iOsmError({
+                                var d = new qaError({
+                                    // Info required for every error
                                     loc: loc,
-                                    comments: null,
-                                    error_subtype: '',
+                                    service: 'improveOSM',
                                     error_type: k,
-                                    icon: 'fas-long-arrow-alt-right',
```
bhousel: Can you add this icon back?  It got missed because I only added it last night  😆

bhousel: or are this in data/qa_errors.json now?  thats ok too..

SilentSpike: Yeah this has been moved to the .json

I figure values we're setting in iD and not those being read from the service API can go there.

SilentSpike:
I'll push a fix for that error and then I'm happy for this to be merged (if you think this is the right way to go, I'm not super experienced with this kind of framework design so there may be a better way).

There's more to do, but can be separately done once this is in place.

bhousel:
Thanks @SilentSpike, I just merged this 👍

Issue #5859, opened 2019-02-05
Add preset for Golf Paths / Cartpaths


chadrockey:
I'm looking to extend similar to this:
https://github.com/openstreetmap/iD/blob/6ce2c5ea46b6bb89f8a3a3dd9d5794a920bbf610/data/presets/categories/golf.json

For lines types to bundle Golf->Hole with Golf->path (intended only for foot traffic) and Golf->cartpath (available for golfcarts).

These presets should create something like: highway=path and golf_cart=designated

My main question is, in order to add a value to the tag key "golf", such as "golf":"path" or "golf":"cartpath" is this mostly communal and metadata driven or is there a formal review process?

Once I find out how much formality is needed, I'll create the relevant json files for the presets and make a PR.

Originating discussion:
https://github.com/gravitystorm/openstreetmap-carto/issues/3650

bhousel:
This is all I could find about them on the OSM wiki:
https://wiki.openstreetmap.org/wiki/Key:golf_cart

What you suggested for `highway=path`+`golf_cart=designated` seems good.

> My main question is, in order to add a value to the tag key "golf", such as "golf":"path" or "golf":"cartpath" is this mostly communal and metadata driven or is there a formal review process?

Communal and metadata-driven.  It is already used [almost 10k times](https://taginfo.openstreetmap.org/keys/golf_cart) and it's a real thing that I'm sure people want to map, so adding a preset to iD is no problem.

chadrockey:
Nice!  Both are in use!
https://taginfo.openstreetmap.org/keys/golf#values

I'll get started on this immediately.

bhousel:
> I don't care which tags are chosen for this preset, but I have found that most golf cart paths are tagged as `highway=track` in osm, and not `highway=path`.

per https://github.com/openstreetmap/iD/pull/5862#issuecomment-460771112
We're going to use new `golf=*` tags because many of these paths don't connect to the main routing graph, and we don't want the disconnected golf paths to be flagged as routing errors.

bhousel:
Done in #5862

PR #5862, opened 2019-02-05
Presets for golf paths and cartpaths and driving ranges

chadrockey:
Adds presets to correctly tag golf paths.

golf: cartpath is intended for the small, single vehicle electric or gas powered golf carts on an actual golf course.  North American/USA style golf.

golf: path is intended for walking only paths, typically in more traditional golf where walking is more common.  Intended for the connecting routes between holes on an actual golf course.

Related issue: https://github.com/openstreetmap/iD/issues/5859

Looking for feedback on correctness and how to validate my changes.


chadrockey:
Also added a driving range preset since it's the 10th most common golf tag and has a wiki entry.

chadrockey:
@bhousel Thanks!  Turns out that json linters can't catch : inside of the quotes.  :)

As far as the tags, I got them from here:
https://taginfo.openstreetmap.org/keys/golf#values

Cartpath and Path are 11 and 13th most common.  If the tags are accepted here, I'll look into what else is required to standardize the in-use tags and will at minimum complete the wiki entries with suitability licensed example photos.

My motivation is that I'm using the rest of the golf: tags for metadata analysis.  What I'd like to see is the standard for "natural walking trails", which I understand to be highway: footway (walking only, not wide enough to be a track or intended for vehicles) and highway: path (with golfcart designated as suggested).

The idea being that there is a standard tag for rendering and the golf: tags seem to be "extras" across the board.  For example, they are mostly landuse: grass, but that grass happens to be a golf:green.  With this in mind, I tried to find the most suitable walking paths and small vehicle paths and use add the "extra" golf tags.

Specifically, I'm trying to make it clear which paths are part of the design of golfcourses vs merely near a golf course.  There may be a hiking trail near a golf course, but a path architected to go from golf green to the following tee box isn't for general use and is a crucial part of the golf course.

Please let me know if the taginfo is valid and acceptable proof of use with my understanding of the golf: tags in general.

Thank you for reviewing.  Will I be able to test these changes if I follow through with the npm install?  I'd like to see that the icons, descriptions, and nesting are all correct.

bhousel:
> Turns out that json linters can't catch : inside of the quotes. :)

Yes, it's not strictly an error - you can put anything in the quotes.

> As far as the tags, I got them from here:
> https://taginfo.openstreetmap.org/keys/golf#values
> 
> Cartpath and Path are 11 and 13th most common. If the tags are accepted here, I'll look into what else is required to standardize the in-use tags and will at minimum complete the wiki entries with suitability licensed example photos.

Sure but a few hundred uses is not really established:

screenshot 2019-02-05 13 10 59

You can use them if you really want, but again my strong recommendation is to subtag under the `highway=*` stuff, since that is a more normal way of inventing tags.

I'd do this: 
`highway=path`+`path=golf`
`highway=track`+`track=golf`+`tracktype=grade1`

This looks strange to me:
`highway=path`+`golf=path`
`highway=track`+`golf=cartpath`

> Thank you for reviewing. Will I be able to test these changes if I follow through with the npm install? I'd like to see that the icons, descriptions, and nesting are all correct.

Yes!

chadrockey:
@bhousel Can you comment if the wiki is out of date?
https://wiki.openstreetmap.org/wiki/Key:highway#Paths

Most golf cart paths are not wide enough for tractors or jeeps, so this suggests that "path" should be used.  And while "path" suits both the walking and golf cart paths, now there isn't a clear distinction.

I would say that walking paths most resemble hiking paths, with maybe nicer grass and less dirt.

So the two possibilities I see:
1) Use path for both presets, but set "golf_cart": "designated" for the cart paths.
2) Use path for golf carts, use footpath for the walking trails.

In either case, should "golf_cart": "designated" be included?

chadrockey:
@bhousel I updated this in the only way I could find that avoids connectivity errors in my editor.  I did try everything out locally and it's exactly what I wanted in terms of the UI and labeling.  Very clear and easy to search!

Ultimately, our reasoning for these paths is that they are not highways or routing paths - you can't really take them on route from a point A to a point B.  They are more descriptive pieces of the map.  Even within a golf course, there are often segments and routes that aren't for routing but just for traffic management.  For example, there may be no clear path between a tee and a green because you go to where your ball is, not in a consistent path.  Routing in this instance is out of scope for Open Street Map, it's not necessary to always navigate from "Hole 15 to Hole 7".  For this reasoning, we're proposing to only use golf: cartpath and golf: path in these situations, primarily to avoid connectivity errors.

In situations where the routes are NOT intended solely for the game of golf, then the usual highway rules will take precedent.  For example, a theme park has "golf cart paths" that its employees and emergency personnel use to travel in areas off limits to the guests.  These are NOT golf: cartpaths, but are the usual highway: path or highway:track.  This will need to be a clear distinction on the wiki pages.

For situations where both situations apply: example: a retirement community built around a golf course where the course's cart paths serve the purpose for the game as well as transportation routes for the residents, then the most appropriate tagging will be highway path/track with the golf:cartpath tag added. 
 
Do you think these cases should be handled in the wiki?  Or should we add another preset or two here so that these difficult cases are handled better for users?

Maybe the set in order of Game -> Transportation:
Golf Walking Path (golf: path)
Golf Cartpath (golf: cartpath)
Golf Cartpath Multi Purpose (highway: track + golf:cartpath ; expected to connect)
(and I think the final case for non-Golf game cartpaths would be for generic Path or generic Track)

The last open item is that there's obviously no rendering for these.  Is there a secondary tag available that will provide clear rendering?  If not, I will work with the rendering project to add rendering for these  "new" tags.

chadrockey:
For what it's worth, I prototyped a first pass at this idea and I think it works well.  It preserves the usage of highway:track for the larger, multipurpose golf cart paths, while supporting the game "golf:cartpath" tag.

I looked for a quick example and found this area.  Here you can see a Golf Cartpath transition into a special portion of the road where residents drive their own golf carts, even when they are not playing the game.  After a while, the highway turns back into a Golf Cartpath and is only used/intended when actually playing.

multipurpose

bhousel:
> Golf Cartpath Multi Purpose (highway: track + golf:cartpath ; expected to connect)

Could you use `highway=service` for this instead of `highway=track`?   `highway=track` is really for roads which go off into the wilderness and might be dangerous to drive on because of obstacles or whatever.

otherwise looks good 👍 

> The last open item is that there's obviously no rendering for these. Is there a secondary tag available that will provide clear rendering? If not, I will work with the rendering project to add rendering for these "new" tags.

No, sorry.. We usually don't consider whether or not a feature will be rendered on the main osm-carto style, because people can make their own renderings if they want to, and the osm-carto project moves kind of slowly (some of this is due to technical limitations of how they render - they need to change their database schema to support new tags sometimes).

chadrockey:
@bhousel Service sounds like a good fit!  Thanks!

OK, all updated.  Do I need to remove data/taginfo from the PR?

bhousel:
> OK, all updated. Do I need to remove data/taginfo from the PR?

Thanks @chadrockey looks good!  

No need to change the taginfo stuff, I will merge and rebuild everything, and make some updates to the .css files so the paths render nicely  👍

chadrockey:
@quincylvania Do you have a timeline of when this might go live?  I'd like to make a video tutorial but don't want to do so until these new features are in.  I could record from a local instance of iD, so I might do that.

Thanks!

quincylvania:
@chadrockey We're aiming for a release later this week! Though it can take another day or so until it's live on [openstreetmap.org](https://openstreetmap.org)

PR #5867, opened 2019-02-05
Add comments to ImproveOSM issues

SilentSpike:
If you want to test this out, I've left a few test comments on the oneway issue found at `&map=17.00/64.08049/-21.94070`.

Things to fix before merging:
- [x] Clear the comment box when submitting a comment
- [x] Don't retrieve the comments every time the UI is shown. Can probably get away with just doing the first time for now.

Preview:
image

### DIFF
--- a/modules/services/improveOSM.js
+++ b/modules/services/improveOSM.js
@@ -329,6 +329,29 @@ export default {
         });
     },
 
+    // Test case:
+    // http://missingroads.skobbler.net/missingGeoService/retrieveComments?tileX=137495&tileY=89379
+    getComments: function(d, callback) {
+        var key = d.error_key;
+        var qParams = {};
+
+        if (key === 'ow') {
+            qParams = d.identifier;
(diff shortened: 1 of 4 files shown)


SilentSpike:
Special thanks to @beataj-telenav for the API knowledge to get these working ❤️

Something I forgot to mention is that the avatars will all be the default as shown in the preview image (because as far as I can see you need the user uid to retrieve their information via the osm api). I considered just remove that element altogether, but think it's worth keeping to make visually clear that these are comments from a user.

bhousel:
> Something I forgot to mention is that the avatars will all be the default as shown in the preview image (because as far as I can see you need the user uid to retrieve their information via the osm api). I considered just remove that element altogether, but think it's worth keeping to make visually clear that these are comments from a user.

Nice @SilentSpike this looks really great!
I will see if there is a way to get the full userdetails.. We are doing something similar for the Notes feature, but you're right that it is easier with the uid.

PR #5875, opened 2019-02-07
Dont snap to non-vertices

### LINKED ISSUE #5811: Don't snap to point features that can't be vertices when line drawing
If the preset for a point feature does not include the `vertex` geometry, iD should avoid snapping to those nodes during line drawing.

One example where this would be useful is in adding buildings or sidewalks to a downtown area with many businesses tagged as points near their front entrances. While the user can hold down option to turn off snapping, they may not be aware of that feature or may be mapping on a touch device.

maxgrossman:
closes 

#5811

### DIFF
--- a/modules/behavior/draw.js
+++ b/modules/behavior/draw.js
@@ -13,6 +13,7 @@ import { behaviorTail } from './tail';
 import { geoChooseEdge, geoVecLength } from '../geo';
 import { utilKeybinding, utilRebind } from '../util';
 
+import _isEmpty from 'lodash-es/isEmpty';
 
 var _usedTails = {};
 var _disableSpace = false;
@@ -26,7 +27,7 @@ export function behaviorDraw(context) {
 
     var keybinding = utilKeybinding('draw');
 
-    var hover = behaviorHover(context).altDisables(true)
+    var hover = behaviorHover(context).altDisables(true).ignoreVertex(true)
         .on('hover', context.ui().sidebar.hover);
     var tail = behaviorTail();
     var edit = behaviorEdit(context);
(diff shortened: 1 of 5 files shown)


bhousel:
Looks good, thanks @maxgrossman 👍

quincylvania:
@bhousel FYI Max told me he added `wip` since there were still some issues with the on-map highlighting, but I guess we can work on that in time!

maxgrossman:
hey guys! You guys beat me to dropping a note about a bug I've found and that presently I'm not sure the best way to fix.

There's an intermittent behavior where the `entity.id` and `_newId` in the hover behavior remain the same such that the node we're trying to snap to does not receive its orange outline. 

I think the relevant lines of code are [here](https://github.com/openstreetmap/iD/blob/master/modules/behavior/hover.js#L143)

Here's a gif of me trying to debug it yesterday...

not_finding_new_id

Just by comparing the 2 lines below the second to last `it is null` log with the failing bottom line, the hover code in the 'success' scenario gets called twice, making the 2 ids are different.

Figuring out how/why it only called once when it fails is something I still need to figure out...

Also note that the non-hover scenario does not cascade, just happens once, then you click a new node, hover over the snappable and it is back to the expectation.

quincylvania:
@maxgrossman Thanks for uncovering and looking into that bug! Turns out it's actually in the release branch too. This issue is [here](https://github.com/openstreetmap/iD/blob/96a1ec92c7e43c011c8e4a29cff3d25df0b7ac96/modules/behavior/hover.js#L141). `_newId` is set as the first node the user hovers over after entering the draw mode. It could be any node, so we need to force it to be the node the user is continuing from.

quincylvania:
Fixed in 0b86be7bcb9d4ba13ba8ee71559001ce3d64cf1d

PR #5879, opened 2019-02-09
Fix negative number of trips in ImproveOSM issues

SilentSpike:
See https://github.com/openstreetmap/iD/pull/5739#issuecomment-460786055

This is something of a temporary solution as error titles and descriptions will be pretty easily generalised so the UI code is likely to change.


bhousel:
Thanks @SilentSpike 👍

PR #5884, opened 2019-02-12
Disable preset buttons for hidden layers

### LINKED ISSUE #5876: Disable preset buttons for hidden layers
It's not always obvious when a set of map features are hidden. This can lead to users re-mapping features that are already mapped but not shown. This is especially an issue with buildings, since they will automatically be hidden if there are too many.

We should disable the presets associated with the hidden layers, showing a tooltip on hover that informs users of the situation.

screen shot 2019-02-08 at 4 15 27 pm

screen shot 2019-02-08 at 4 10 01 pm

quincylvania:
Closes #5876.

It'd be great to get a review since this changes some rather hot code.

screen shot 2019-02-12 at 2 41 56 pm
screen shot 2019-02-12 at 2 41 09 pm

### DIFF
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -95,8 +95,12 @@ export function rendererFeatures(context) {
             defaultMax: (max || Infinity),
             enable: function() { this.enabled = true; this.currentMax = this.defaultMax; },
             disable: function() { this.enabled = false; this.currentMax = 0; },
-            hidden: function() { return !this.enabled || !context.editable() || this.count > this.currentMax * _cullFactor; },
-            autoHidden: function() { return this.hidden() && this.enabled; }
+            hidden: function() {
+                return !context.editable() ||
+                    (this.count === 0 && !this.enabled) ||
+                    this.count > this.currentMax * _cullFactor;
+            },
+            autoHidden: function() { return this.hidden() && this.currentMax > 0; }
         };
     }
 
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -95,96 +95,96 @@ export function rendererFeatures(context) {
             defaultMax: (max || Infinity),
             enable: function() { this.enabled = true; this.currentMax = this.defaultMax; },
             disable: function() { this.enabled = false; this.currentMax = 0; },
-            hidden: function() { return !context.editable() || this.count > this.currentMax * _cullFactor; },
-            autoHidden: function() { return this.hidden() && this.currentMax > 0; }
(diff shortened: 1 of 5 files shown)


modules/renderer/features.js
```diff
@@ -95,104 +95,104 @@ export function rendererFeatures(context) {
             defaultMax: (max || Infinity),
             enable: function() { this.enabled = true; this.currentMax = this.defaultMax; },
             disable: function() { this.enabled = false; this.currentMax = 0; },
-            hidden: function() { return !context.editable() || this.count > this.currentMax * _cullFactor; },
-            autoHidden: function() { return this.hidden() && this.currentMax > 0; }
+            hidden: function() { return !this.enabled || !context.editable() || this.count > this.currentMax * _cullFactor; },
+            autoHidden: function() { return this.hidden() && this.enabled; }
```
bhousel: I think the changes on 98-99 would change how the filtering menu works.. `enabled` is whether the user has deliberately checked the check box, but `hidden` and `autoHidden` will update the checkboxes if the user zooms out or a feature like buildings or points is getting culled because there are too many of them.

modules/renderer/features.js
```diff
@@ -412,6 +412,18 @@ export function rendererFeatures(context) {
     };
 
 
+    features.isHiddenPreset = function(preset, geometry) {
+        if (!_hidden.length) return false;
+        if (!preset.tags) return false;
+        for (var i = 0; i < _hidden.length; i++) {
+            if (_features[_hidden[i]].filter(preset.tags, geometry)) {
```
bhousel: Probably want to use `addTags` (the ones that will get added) instead of `tags` (the ones that will get matched).  They are usually the same, but not necessarily so.

Also, this might not work for situations where applying a preset will set a tag because of a field.  So if you are hiding buildings, and the preset has `building_area` field (for example "Office"), and you draw an area, that feature will get a `building=yes` tag assigned.

modules/renderer/features.js
quincylvania: Oh okay. There was an issue where `hidden()` would return `true` if `count` was `0` even if `enabled` was `false`. In the latest commit I restored the old behavior while accounting for this case.

modules/renderer/features.js
quincylvania: Good point. I just used `setTags` with `{}` to get `addTags` and the field defaults.

bhousel: nice 👍

PR #5885, opened 2019-02-12
Persist forced visible entities on move/rotate

SilentSpike:
Allows these operations to be performed on entities that are forced visible without them disappearing during the move/rotation.


bhousel:
Thanks @SilentSpike 👍

Issue #5882, opened 2019-02-11
Allow choosing the feature type before drawing


quincylvania:
**Issue**

Currently, there is a single workflow for adding a new feature:

1. Choose a geometry type (Point, Line, or Area)
2. Draw the geometry for the generic feature
3. Search for and select a feature type

There are a number of downsides to this approach. For one, new users often stumble over figuring out what geometry works with the type of feature they want to add. Also, this workflow forces use of the sidebar after every single addition, making fullscreen editing difficult and potential mobile editing untenable.

Further, iD could be more proactive if it knew what feature the user was drawing, and for example could:
- Instantly flag or preventing preset-specific validation errors (like waterway-railway connections)
- Automatically add appropriate tags to junction nodes (like railway-highway level crossings)
- Provide drawing instructions specific to the feature (like tracing cliffs with the base to the right)

**Solution**

I propose adding an additional (not replacement) workflow:

1. Search for and select a feature type
2. Choose a compatible geometry type (Point, Line, or Area)
3. Draw the geometry

I mocked-up a possible UI:

<img width="346" alt="screen shot 2019-02-10 at 11 39 45 am" src="https://user-images.githubusercontent.com/2046746/52566460-ceaa5a00-2dd7-11e9-9593-790586635a89.png">

<img width="359" alt="screen shot 2019-02-10 at 11 39 54 am" src="https://user-images.githubusercontent.com/2046746/52566472-d833c200-2dd7-11e9-904e-2b9a90f8f8cc.png">

<img width="366" alt="screen shot 2019-02-10 at 11 40 00 am" src="https://user-images.githubusercontent.com/2046746/52566479-dbc74900-2dd7-11e9-9c01-f41ffbb75f2e.png">

With #5873, favorite presets would appear like this, allowing one-click selection of both the preset and geometry type.

<img width="471" alt="screen shot 2019-02-10 at 11 42 23 am" src="https://user-images.githubusercontent.com/2046746/52566802-aff89300-2dd8-11e9-8536-cfeaaadca747.png">

The default favorites would be the generic geometry types. These buttons would work identically to the current ones.

<img width="474" alt="screen shot 2019-02-10 at 11 41 35 am" src="https://user-images.githubusercontent.com/2046746/52566899-ecc48a00-2dd8-11e9-9f5d-62f1b791aad4.png">

#5839 is a prerequisite issue, since we'll need the line icons to be more flexible.

tordans:
[...]

Issue #5873, opened 2019-02-07
Quick feature creation from favorite presets


quincylvania:
Users often map many instances of just a few types of features. We should let users select "favorite" presets and show these as a mode button alongside Point, Line, and Area. This will allow creating features with a predetermined preset in one step rather than two. For the sake of space, only the most recent should be shown in the button, with the rest accessible from a dropdown menu.

An alternative to favoriting would be to show the most recent, but this might confuse new users and fail to give advanced users enough control.

This is related to #590 but differs in that it doesn't affect the preset list in the sidebar and doesn't account for user-specified tags, just the built-in presets.

![preset mode](https://user-images.githubusercontent.com/2046746/52412354-807e1980-2aac-11e9-9e72-de8ecde2ded6.png)
![preset mode selected](https://user-images.githubusercontent.com/2046746/52412353-807e1980-2aac-11e9-8ebf-539309dcdd0d.png)

Cc @brianhatchl who demoed a live implementation of this feature and inspired this issue. Let's get this into the main iD!

brianhatchl:
Hey!  I've worked a bit on this https://github.com/brianhatchl/iD/tree/favorite-preset and came up with this approach for "favorite-ing" a preset.

![peek 2019-02-07 13-40](https://user-images.githubusercontent.com/2600932/52434798-34e46380-2ade-11e9-9ff6-1bab643bcadd.gif)

Instead of a dropdown, I was thinking of limiting the number of presets to three (kind of like old radio preset buttons).  New favorites above three would overwrite the current ones in some fashion.  Maybe FIFO or maybe just updating the last one.  That way the dropdown couldn't grow too long and approach becoming a copy of the preset list in the sidebar.

It would be helpful for the favorite selections to be persisted in local storage.

In this approach, the preset favorite also tracks the geometry.  One hang up I have right now is that the favorite icon gives no indication of the geometry type.  There is more that can be done to include something like the `icon-frame` used by the larger preset icons.
![image](https://user-images.githubusercontent.com/2600932/52438455-4b42ed00-2ae7-11e9-8afc-ea6c5f97ea19.png)


I was originally thinking that all preset favorites would be locked drawing modes, but I do like your idea of locking any of the modes with double-click.

The branch above doesn't have any of the mode logic yet.  The original live demo you referenced was for a "More like this" context menu item.

[...]

PR #5888, opened 2019-02-13
Dynamic line previews in line preset icons

### LINKED ISSUE #5839: Line icons used on Area features look ugly
This happens when we use the wrong kind of icon on a feature.  

screenshot 2019-02-01 16 01 02

I'd like to fix this by getting rid of the "line" and "relation" icons.  We could just have normal icons, then add the line/relation graphic below the icon with code, only if we need to.

screenshot 2019-02-01 16 05 27

quincylvania:
Closes #5839. Re: #5873, #5882.

- Remove embedded line geometry from iD line feature icons
- Add dynamic line previews to all line preset icons
- Use the same stroke styling for line icons as when rendering the lines in the map
- Icon styling can be overridden. For example, the stroke/casing of cycle paths is reversed for the light background

This PR does not address embedded geometry in relation icons.

- [x] The `stroke-dasharray` could probably be improved for railways, ferry routes

Nicer icons when using line preset as areas:
screen shot 2019-02-13 at 4 17 07 pm

A selection of the new icons:
screen shot 2019-02-14 at 9 59 22 am
screen shot 2019-02-13 at 4 15 01 pm
screen shot 2019-02-13 at 4 14 53 pm
screen shot 2019-02-14 at 9 59 05 am
screen shot 2019-02-13 at 4 14 19 pm
screen shot 2019-02-13 at 4 14 09 pm
screen shot 2019-02-13 at 4 16 16 pm
screen shot 2019-02-13 at 4 15 32 pm
screen shot 2019-02-13 at 4 15 17 pm


SilentSpike:
It sounds crazy and I've checked your maths (`10.5 + 6/2 = 13.5`), but I swear the vertices are slightly lower than the line 😆

quincylvania:
> It sounds crazy and I've checked your maths (`10.5 + 6/2 = 13.5`), but I swear the vertices are slightly lower than the line 😆

@SilentSpike Are you seeing this in the images or live? Now I'm worried that certain browsers or screens don't like the partial-pixel alignment.

SilentSpike:
@quincylvania I noticed it particularly in the residential road image, then checked a live instance and see the same. Took a screenshot and see a 1px difference:

discrep

bhousel:
> @SilentSpike Are you seeing this in the images or live? Now I'm worried that certain browsers or screens don't like the partial-pixel alignment.

Maybe try -
- setting an explicit `viewBox=0 0 40 20` attribute on the svg.  (width and height should do the same thing, but I'm not 100% sure).
- changing `lineSvg`'s height from 20 to 19 or 21?  (you can open Chrome's dev tools and play with this value live)

Retina screens should handle the half pixel alignment ok - non retina screens would blur it.  Also SVG strokes are measured from the center of the stroke 

OH and I'd draw those vertices as svg circles, not rounded divs.  Actually that's probably the real issue and the lines are probably fine.

quincylvania:
@bhousel Thanks for the tips! I implemented them and this should be ready for review.

@SilentSpike I changed the line casing in the icons to use full-pixel increments, which I think was the problem. Let me know if you're still seeing issues. I updated the railways screenshot so you can check that, or preferably on live.

quincylvania:
@bhousel Thanks! I could do the relation ones but I'm not sure those unblock new features like this did for #5873. Feel free to create an issue.

bhousel:
> @bhousel Thanks! I could do the relation ones but I'm not sure that would unblock new features like this did for #5873. Feel free to create an issue.

It seems kind of silly not to just do all of them at the same time.

Issue #5874, opened 2019-02-07
Allow locking of the draw mode


quincylvania:
When adding many features of the same type, it becomes unnecessarily repetitive to keep selecting the draw mode (point, line, or area). There should be a mechanism to lock into a mode so it stays active even after adding a feature. Double-clicking would be a simple way to lock, with a single click sufficient to unlock.

This would be especially powerful combined with #5873 since then users could quickly add many features of the same type with no clicks required in-between additions.

While theoretically this would make it easier to accidentally add many untagged features, #5830 will prevent their upload and help users tag or remove them.

Draw mode buttons with quick presets after #5873:
![preset mode](https://user-images.githubusercontent.com/2046746/52414217-b4a80900-2ab1-11e9-8bd2-c6e159251238.png)
With single feature draw mode active (on click or hotkey):
![preset mode selected](https://user-images.githubusercontent.com/2046746/52414218-b5409f80-2ab1-11e9-9df4-a51b610cd7c6.png)
Possible UI with locked draw mode active (on double click):
![preset mode locked](https://user-images.githubusercontent.com/2046746/52414216-b4a80900-2ab1-11e9-9b88-09859a521c63.png)

quincylvania:
I'm working on this today, targeting a "Repeat" toolbar button that appears during adding/drawing.

 
![Screen Shot 2019-05-30 at 10 24 55 AM](https://user-images.githubusercontent.com/2046746/58639667-6bc8ca00-82c5-11e9-938e-c757087fd820.png)

quincylvania:
I did this. Get ready for speed!

🏎💨💨💨

![repeat demo](https://user-images.githubusercontent.com/2046746/58645877-2f4f9b00-82d2-11e9-8e7a-bb216bcaee40.gif)

sun-geo:
:-)

PR #5908, opened 2019-02-18
Add double-click locking of draw mode

brianhatchl:
refs #5874

### DIFF
--- a/modules/ui/modes.js
+++ b/modules/ui/modes.js
@@ -51,7 +51,8 @@ export function uiModes(context) {
         context
             .on('enter.editor', function(entered) {
                 selection.selectAll('button.add-button')
-                    .classed('active', function(mode) { return entered.button === mode.button; });
+                    .classed('active', function(mode) { return entered.button === mode.button; })
+                    .classed('locked', function(mode) { return context.lockMode() != null && context.lockMode().id === mode.id; });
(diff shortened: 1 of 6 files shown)


modules/ui/modes.js
```diff
@@ -76,6 +77,27 @@ export function uiModes(context) {
 
 
         var debouncedUpdate = _debounce(update, 500, { leading: true, trailing: true });
+        var debouncedClick = _debounce(clickModeButton, 300);
```
brianhatchl: @quincylvania Not sure if a debounced function is the best way to support a click and a dblclick on an element, but it works for my click rate.

quincylvania:
Thanks, @brianhatchl, I'll take a look!

quincylvania:
So at the moment we're targeting #5960 which will actually hide the mode buttons as soon as you enter the drawing mode. Double-clicking wouldn't be a viable interaction at that point, particularly for unlocking the mode. 

Alternatively we could have a lock button in the contextual UI. This is far more discoverable anyway and would work better on touchscreens.

screen shot 2019-02-27 at 4 17 59 pm

bhousel:
Closing here, as @quincylvania said we're currently heading towards a contextual secondary menu that will include a lock button.

PR #5914, opened 2019-02-20
Favorite preset

brianhatchl:
refs #5873


brianhatchl:
@quincylvania I didn't get the preset favorite/tag-reference buttons sharing the same vertical space figured out (my kung-fu for making fiddly css changes is lacking :grin:).  But I did get the favorite preset/modes wired up.  It relies on changes to mode constructors that will hopefully pass muster.

The favorites get written to local storage.  This code is in the context because I needed access to the favorites from the favorite star button as well as the modes; and also needed to be able to fire an event when favorites change to trigger the modes update method.

The preset favorites have a limit of 3 currently (no drop down at this point).  I also rationalized that the tooltip could describe the geometry type bound to the favorite (instead of figuring out the icon geometry modifier).

quincylvania:
Thanks @brianhatchl! I'm working on fixes for the 2.14.2 point release right now. I'll take a look once things settle down!

quincylvania:
@brianhatchl I ended up tweaking a lot of little things like the icons and tooltips but the underlying code seems to do the trick. Thanks again!

@bhousel I'm happy with this as an initial implementation. It could still use some polish as we build out #5882 and #5960 but I think we can merge it as soon as you can give it a look over.

favoriting demo

quincylvania:
Further considerations:

- We should allow more than 3 favorites and have some mechanism for when they don't all fit.
- When not all the buttons fit, we should remove favorite labels starting right to left and not necessarily remove them all at once. (I'd also be okay with never showing the favorite labels. They're somewhat ungainly.)
- There should be a way to remove a favorite without having to have a feature of that preset selected.
- The ability to drag-and-drop reorder favorites would also be cool.
- The icons look weird when the button is selected but this won't be an issue after #5960.
- The keyboard shortcuts should continue down the number row, so `4` selects the first favorite, etc.
- After #5882, the Point, Line, and Area buttons should not be given special treatment. They would be favorited generic presets that the user can unfavorite to remove altogether.

bhousel:
> @bhousel I'm happy with this as an initial implementation. It could still use some polish as we build out #5882 and #5960 but I think we can merge it as soon as you can give it a look over.

This looks great!   I took a quick look over the code and it all looks great 👍  Go ahead and merge it so people can start using it and we can collect some early feedback..

PR #5940, opened 2019-02-23
Presets for LGBTQ+ venues

rory:
Here I add a few presets for LGBTQ+ venues, based on the [`lgbtq`](https://wiki.openstreetmap.org/wiki/Key:lgbtq) tagging scheme.


quincylvania:
Hi @rory! Thanks for opening this—could you change just a few little organization details?

- Instead of paths like `amenity/lgbtq_bar.json` can you change them to be like `amenity/bar/lgbtq.json`?
- Once you do that, you can delete the `fields` and `moreFields` properties. They will inherit the same fields as the base preset.

Then we can merge ➡️🆔

rory:
I didn't know you could do that with presets. I have made those changes and updated the branch.

quincylvania:
Looks great, thanks @rory 👍

> I didn't know you could do that with presets.

We added field inheritance relatively recently (2.13.0?). It's been great for keeping similar presets consistent and avoiding duplicate data.

PR #5950, opened 2019-02-24
Fix pressing Escape with only one active node in area

### LINKED ISSUE #5941: Escape throws an error when drawing area with only one point
```
[Error] TypeError: undefined is not an object (evaluating 'entity.id')
	isInvalidGeometry (iD.js:60348)
	checkGeometry (iD.js:60313)
	finish (iD.js:60522)
	call (iD.js:786)
	ret (iD.js:30157)
	testBindings (iD.js:28419)
	bubble (iD.js:28477)
	(anonymous function) (iD.js:1527)
```

Possibly related: #5911

jguthrie100:
Fixes #5941

### DIFF
--- a/modules/behavior/draw_way.js
+++ b/modules/behavior/draw_way.js
@@ -125,6 +125,8 @@ export function behaviorDrawWay(context, wayID, index, mode, startGraph, baselin
 
             if (origWay.isClosed()) { // Check if Area
                 if (finishDraw) {
+                    if (nodes.length < 3) return false;
+
                     nodes.splice(-2, 1);
                     entity = nodes[nodes.length-2];
                 } else {


jguthrie100:
press-escape-with-one-node-in-area

quincylvania:
Thanks, @jguthrie100 👍👍

maxgrossman:
...yay! Awesome for coming in for the quick bug fix @jguthrie100! 

Though I'm still stuck on feeling it is a little weird that some times the graph's `childNodes` method returns the # of nodes that the way obj has and other times it does not. Could it be that there a more problematic graph-related bug we need to investigate that this relatively innocuous bug has just brought light to.

quincylvania:
@bhousel Do you know of any reason `childNodes` wouldn't return all the nodes?

bhousel:
> @bhousel Do you know of any reason childNodes wouldn't return all the nodes?

Maybe it was called on a stale graph?  I can check this week..

PR #5953, opened 2019-02-24
add amenity=toilet to amenity=toilets validator rule

### LINKED ISSUE #5952: validator: replace amenity=toilet by amenity=toilets
See https://taginfo.openstreetmap.org/tags/amenity=toilet

Idea from https://josm.openstreetmap.de/browser/josm/trunk/data/validator/deprecated.mapcss

matkoniecz:
fixes #5952

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -65,6 +65,10 @@
       "replace": {"leisure": "swimming_pool"}
     },
     {
+      "old": {"amenity": "toilet"},
+      "replace": {"amenity": "toilets"}
+    },
+    {
       "old": {"amenity": "vending_machine", "vending": "news_papers"},
       "replace": {"amenity": "vending_machine", "vending": "newspapers"}
     },


matkoniecz:
https://josm.openstreetmap.de/browser/josm/trunk/data/validator/deprecated.mapcss has some other ideas

How many deprecated tags should be present to make adding a good idea? 1000?

bhousel:
> How many deprecated tags should be present to make adding a good idea? 1000? 500? 100?

I’d prefer make tagging decisions based on improving the user experience for mappers, the needs of data consumers, and common sense, rather than picking an arbitrary number.

quincylvania:
@matkoniecz Thanks, great suggestion!

matkoniecz:
> I’d prefer make tagging decisions based on improving the user experience for mappers, the needs of data consumers, and common sense, rather than picking an arbitrary number.

Obviously, all of that is important and good validator rule must match all that requirements. But I think that it makes little sense to spend time on adding rule to handle deprecated tag present 2 times worldwide and not used anymore.

bhousel:
> But I think that it makes little sense to spend time on adding rule to handle deprecated tag present 2 times worldwide and not used anymore.

Nice! Glad your version of “common sense” matches up with mine :smile:

matkoniecz:
I just wanted to check is it matching well - for me good level for adding new rule would be 1000 uses of deprecated tag worldwide (obviously, assuming that all other requirements for deprecation rule are fullfilled).

PR #5956, opened 2019-02-24
building:color -> building:colour deprecation

matkoniecz:
https://taginfo.openstreetmap.org/keys/building%3Acolor

inspired by https://josm.openstreetmap.de/browser/josm/trunk/data/validator/deprecated.mapcss#L569

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -93,6 +93,10 @@
       "replace": {"entrance": "*"}
     },
     {
+      "old": {"building:color": "*"},
+      "replace": {"building:colour": "$1"}
+    },
+    {
       "old": {"building:roof:colour": "*"},
       "replace": {"roof:colour": "$1"}
     },


matkoniecz:
note: how can I test this one? I want to be sure that it will remove old tag, not just add new one. I am assuming it will work based on very similar rules but I want to be sure.

quincylvania:
> note: how can I test this one? I want to be sure that it will remove old tag, not just add new one. I am assuming it will work based on very similar rules but I want to be sure.

You can just create a feature and add the deprecated tag in the All Tags editor. The warning will appear and you can click a fix and inspect the results.

bhousel:
It would be fantastic to have unit tests for every validator tho..

matkoniecz:
Is there any that I can base it on? i would happily add based on similar ones but I am not going to invent way to test it.

quincylvania:
@matkoniecz Thanks!

> Is there any that I can base it on? i would happily add based on similar ones but I am not going to invent way to test it.

We have https://github.com/openstreetmap/iD/blob/master/test/spec/validations/deprecated_tag.js. It just checks to make sure the tags are flagged. It doesn't even test the fix yet.

PR #5957, opened 2019-02-24
landuse=conservation to boundary=protected_area

matkoniecz:
https://taginfo.openstreetmap.org/tags/landuse=conservation 

idea from https://josm.openstreetmap.de/browser/josm/trunk/data/validator/deprecated.mapcss#L776

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -172,6 +172,10 @@
       "replace": {"highway": "road", "incline": "unpaved"}
     },
     {
+      "old": {"landuse": "conservation"},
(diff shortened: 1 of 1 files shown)


matkoniecz:
note: how can I test this one? I want to be sure that it will remove landuse value, not just add boundary value. In the worst case it may fall into infinite loop with warning raised for `landuse=conservation boundary=protected_area` with upgrade, not removing `landuse` tag.

matkoniecz:
@quincylvania It passed manual testing.

quincylvania:
@matkoniecz Thanks 👍

PR #5961, opened 2019-02-24
add focusable=false added by iD startup process

matkoniecz:
Building iD (https://github.com/openstreetmap/iD#building-id) on my computer resulted in changes to svg files.

I have no idea whatever it is result of bug (in that case treat it as a bug report), misconfiguration on my computer (in that case sorry) or desirable change (in that case merging it would help to reduce noise in command line, building program should not cause code to change).


quincylvania:
@matkoniecz The font awesome icons are added by a dependency. Make sure you run `npm install` first to make sure you have the latest version. These files as checked into iD aren't critical since they're autogenerated and will be replaced with the latest versions when releasing.

matkoniecz:
I followed https://github.com/openstreetmap/iD#building-id and run `npm install`

Issue #2205, opened 2014-04-24
Add square angles action on nodes


finetjul:
In addition to the "square corners" action, it would be great if such feature exist on single nodes.

Sometimes ways are not straight lines but not a square either (say a parking aisle case), it is a challenge to get all the angles of the way flat or square.

It would be helpful I believe to manually select a node, and square or flatten it with regard to its neighbors.
When a node has multiple neighbors (>2), then it would only flatten/square the angle formed by the 2 lastly selected nodes.

PR #5999, opened 2019-03-01
Improve orthogonalization action

### LINKED ISSUE #1979: Odd behaviour: square-an-area moves connected way
I have a parking-space http://www.openstreetmap.org/browse/way/244586733
and a driveway leading to it http://www.openstreetmap.org/browse/way/246844101

When using "square corners" on the parking-space, 
the point where the driveway joins it moves to the lower left corner.

I tried to reproduce it, building a similar area+road elsewhere,
but there, squaring works as expected.  
But undoing this test, leaves me with 
an active undo-button (but "nothing to undo"), 
as well as "Save: 1" (but nothing to save).  

iD 1.3.2, Opera 17.0, WinXP

### LINKED ISSUE #2472: Squaring-Error
"Square an area" should align the corners of an area / building,
to beautify it to a more right-angled shape.
But sometimes this does not work right, and ends up with an unuseble shape,
e.g. one corner is lost (or 2 points moved to the same coordinates).

When mapping small villages, there are quite a lot of old buildings with odd shapes.
This is where I encounter the buggy behaviour of "square an area" more frequently,
but it can happen with simple 4-cornerd houses too. 
(I just didn't bother to write a bugreport yet)

Here some examples from the village [Rossdorf/Zeilhard](http://www.openstreetmap.org/#map=16/49.84368/8.79194) :
- [Way/316791604](http://www.openstreetmap.org/way/316791604) - simple house
- [Way/316801542](http://www.openstreetmap.org/way/316801542) - 8-corner house gets distorted
- [Way/316794583](http://www.openstreetmap.org/way/316794583) - another big building with "indent".
- [Way/316873578](http://www.openstreetmap.org/way/316873578) - some kind of storage building
- [Way/316794582](http://www.openstreetmap.org/way/316794582) - Farm-building
- [Way/320705754](http://www.openstreetmap.org/way/320705754)

Find all those places I marked with notes: http://overpass-turbo.eu/s/6tN

Note: since the report, there has been an automatic mass-edit, 
that might have corrected the data that leads to the error,
see http://www.openstreetmap.org/changeset/27512311

### LINKED ISSUE #5093: Orthogonalizing paths
Hello,

I just wanted to orthogonalize a path, and couldn't figure out a way to do so with iD. It looks like the “square” option (for areas) is replaced with a “straighten” option (for paths).

Is there a rationale for not supporting squaring paths? As squaring is strictly more powerful than straightening, I don't really understand why the two are treated differently.

Anyway, thank you for your work!

### LINKED ISSUE #5967: Perfect squares cannot be squared as they are not squarish
Currently when a perfectly square way (such as [this one](https://www.openstreetmap.org/way/542259792)) is attempted to be squared by either <kbd>s</kbd> or `right click` + `square`, it returns the message `This can't be made square because it is not squarish.`

This can be confusing as it obviously looks square and the user may experience an existential crisis as a result of being told a square is not squarish. This would seem to be a race case but it would be nice to have a condition of some sort that gives a different message or just squares it anyway.

bhousel:
This PR fixes several long-standing issues with the orthogonalization action.

- [x] closes #1979 - don't move connected way
- [x] closes #2472 - improve algorithm to handle edge cases
- [x] closes #5967 - don't show "not a square" error message for perfect squares
- [x] closes #5093 - allow squaring of paths/lines
- [x] #2205 ?  - allow squaring a single edge when selecting a vertex 
- [x] add tests for the new vector math functions
- [ ] ~maybe use them in some other actions (straighten, etc)~ (will do after #5740)

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -74,12 +74,15 @@ en:
     orthogonalize:
       title: Square
       description:
+        vertex: Square this corner.
         line: Square the corners of this line.
         area: Square the corners of this area.
       key: Q
       annotation:
+        vertex: Squared a single corner.
         line: Squared the corners of a line.
         area: Squared the corners of an area.
+      end_vertex: This can't be squared because it is an end node.
       square_enough: This can't be made more square than it already is.
       not_squarish: This can't be made square because it is not squarish.
       too_large: This can't be made square because not enough of it is currently visible.
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -80,6 +80,7 @@ en:
       annotation:
         line: Squared the corners of a line.
         area: Squared the corners of an area.
+      square_enough: This can't be made more square than it already is.
       not_squarish: This can't be made square because it is not squarish.
       too_large: This can't be made square because not enough of it is currently visible.
       connected_to_hidden: This can't be made square because it is connected to a hidden feature.
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -76,7 +76,7 @@ en:
       description:
         line: Square the corners of this line.
         area: Square the corners of this area.
-      key: S
+      key: Q
       annotation:
         line: Squared the corners of a line.
         area: Squared the corners of an area.
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -105,6 +105,7 @@
                     "line": "Squared the corners of a line.",
                     "area": "Squared the corners of an area."
                 },
+                "square_enough": "This can't be made more square than it already is.",
                 "not_squarish": "This can't be made square because it is not squarish.",
                 "too_large": "This can't be made square because not enough of it is currently visible.",
                 "connected_to_hidden": "This can't be made square because it is connected to a hidden feature."
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -100,7 +100,7 @@
                     "line": "Square the corners of this line.",
                     "area": "Square the corners of this area."
                 },
-                "key": "S",
+                "key": "Q",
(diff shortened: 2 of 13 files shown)


bhousel:
So.. actionOrthogonalize can square lines now!

square line

However.. We already use the <kbd>S</kbd> keyboard shortcut for straightening lines, so it's kind of a problem having 2 menu commands with the same shortcut.

quincylvania:
> We already use the S keyboard shortcut for straightening lines, so it's kind of a problem having 2 menu commands with the same shortcut.

Could we intelligently choose square instead of straighten for lines when the line "can't be straightened because it bends too much?" I guess there are some instances where both are allowed though, and more importantly the user may mean to do one and get the other. I suggest switching the square shortcut to <kbd>Q</kbd>.

bhousel:
> I suggest switching the square shortcut to Q.

Yeah that would neatly solve the problem, and [it's what JOSM uses too](https://josm.openstreetmap.de/wiki/Help/Action/OrthogonalizeShape) .. I'll do it.

bhousel:
> I guess there are some instances where both are allowed though, and more importantly the user may mean to do one and get the other.

Yeah it's actually really easy to construct a line that can be either squared or straightened... It's nice to have both options now!

square and straighten line

bhousel:
I added squaring a single vertex for #2205... It doesn't seem as useful as I thought it would be - just squaring one vertex in isolation of all the other ones - but maybe people will want it.

square a vertex

bhousel:
I'm going to merge this.. The test failures seem related to #6001 / #6002   🤔

Issue #5988, opened 2019-02-27
hash-url-params treated differently from regular params? (Was: Is the `?locale=en`-feature broken?)


tordans:
With https://github.com/openstreetmap/iD/pull/5650 we introduced a feature that allows to set the editor locale via an url param ([API](https://github.com/openstreetmap/iD/blob/master/API.md)). However, I cannot get it to work anymore.

Example: https://www.openstreetmap.org/edit#map=19/50.81751/6.36161&locale=fr

Current behavior: It removes the param from the URL but does not change the language of the UI.

Did anything change? Or am I doing it wrong?

mmd-osm:
You're doing it wrong™

Correct link is: https://www.openstreetmap.org/edit?editor=id&locale=fr#map=19/50.81751/6.36161

tordans:
Thanks @mmd-osm.

I am pretty sure that at some point iD treated hash-params the same as regular params – in the PR I wrote "With the hash-url locale=en-US … one can force a locale and language".

But maybe that changes. I wonder if that is a bug or by design?

mmd-osm:
Well, only the standalone iD version seems to support hash-url at this time:

http://preview.ideditor.com/release/#background=Bing&locale=fr&map=5.58/65.130/35.646

PR #6005, opened 2019-03-03
Clarify API.md description for `locale` param

tordans:
I tried to make the documentation a bit more specific since I ran into this at https://github.com/openstreetmap/iD/issues/5988.

_Changes:_
- Highlight the different URL param-patterns (hash vs. regular)
- Add an example with regional locale `en-US`
- Add `locale` param to the list of osm-website-params with a bit more info about the different default-behavior there

This is a follow up to https://github.com/openstreetmap/iD/pull/5650 and based on the feedback in https://github.com/openstreetmap/iD/issues/5988. Thanks @quincylvania for https://github.com/openstreetmap/iD/commit/07327eb9c7d138c70c1b96fdcad261c921e1cf00, btw.

### DIFF
--- a/API.md
+++ b/API.md
@@ -6,7 +6,7 @@ This file documents efforts toward establishing a public API for iD.
 
 iD supports several URL parameters. When constructing a URL to a standalone instance
 of iD (e.g. `http://preview.ideditor.com/release/`), the following parameters are available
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @tordans 👍

PR #6006, opened 2019-03-03
[usability] add suggestion text below search bar

### LINKED ISSUE #5214: Improve search
### Hint text
iD's search box is currently too mysterious.  Let's add some hint text below it to help the user know what to use it for.

screenshot 2018-08-10 10 28 18

### Beyond that
I'd also like to think about what other kinds of things we can add here. 
For example Plus Codes are [currently trendy](https://github.com/openstreetmap/openstreetmap-website/pull/1818#issuecomment-412023682)

This is one of those weird situations where we actually aren't using a big block of UI for anything, so adding to it costs us almost nothing and can be done in a way that improves the UX.  (usually, adding features degrades UX). 

I'd be ok with supporting more in this box (like plus code, geohash, wikidata ids, even what.three.words) if it helps someone find a spot of earth in OSM and add stuff to it.

jmcelroy5:
Adds some suggestions below the search bar to help users get started. Uses [this copy](https://github.com/openstreetmap/iD/issues/5214#issuecomment-412212612) written by @manfredbrandl 

Question: is searching by plus code actually supported?

Fixes #5214 

screen shot 2019-03-03 at 10 00 38 am

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -875,6 +875,13 @@ a.hide-toggle {
 .feature-list  {
     width: 100%;
 }
+
+.search-hints {
+    padding: 20px;
+    font-size: 14px;
+    text-align: left;
+}
+
 .no-results-item,
 .geocode-item,
 .feature-list-item {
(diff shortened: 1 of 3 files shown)


bhousel:
Sorry for taking a while to get back to you on this.. I'm going to close because it doesn't really do what I asked for in #5214.  I know that @manfredbrandl suggested some block of text, but I really prefer the scannability of displaying the search hints in a list.

PR #6010, opened 2019-03-04
Select feature type before drawing

### LINKED ISSUE #5882: Allow choosing the feature type before drawing

### LINKED ISSUE #5889: Move loading indicator out of the top bar
There's not much reason for the loading indicator to be in the top bar since it's not a button. It takes up space that will be even more valuable as we move toward mobile support. It also looks blurry on retina screens. I suggest moving it to the bottom bar with the other status items.

screen shot 2019-02-14 at 11 45 46 am

Tangentially, can API calls get progress feedback? It'd be cool to show the completed/remaining amounts rather than an indeterminate spinner.

quincylvania:
Closes #5882, closes #5889. 

This PR replaces the Point, Line, and Area buttons with a search bar where users select the feature type before drawing the feature. So far I've found this workflow much more satisfying in practice than the old buttons, but I'd love to hear user feedback!

select type first demo

Presets of different geometry are mixed together, so users don't need to know what geometry to draw in advance.

screen shot 2019-03-04 at 2 49 06 pm

Presets with multiple geometries are expandable.

screen shot 2019-03-04 at 2 47 09 pm 

Mappers who prefer the old behavior can simply favorite the generic presets and map as before.

use old as favorites

Details:
- ~~`1`~~ <kbd>Tab</kbd> focuses the search bar.
- The number keys select the favorited presets.
- The results can be navigated with the arrow keys.
- Focusing the search bar highlights the contents so you can begin a new search immediately.
- Notes have been moved to a separate button.
- Categories are now geometry-agnostic.
- Only presets that allow point, line, and area features are supported for now.

To-do before release:

- [ ] Update walkthrough
- [x] Add support for vertex-only features

To-do eventually:
- Add default list when no text is yet entered
- Add tabs for quickly browsing Recents, Favorites, and Categories
- Add support for relations


iandees:
Ok this is a really great idea. Can there be a left-handed keyboard shortcut to put focus into the search box so that the right hand can stay on the mouse and the left can specify a feature type?

quincylvania:
> Can there be a left-handed keyboard shortcut to put focus into the search box so that the right hand can stay on the mouse and the left can specify a feature type?

Sure, I don't see why we can't use <kbd>0</kbd> too.

iandees:
Apologies I got too excited and didn't read that you already had "1" as a focus shortcut. Nevermind!

The only other thought that came to mind is if the walkthrough would need to be adjusted to introduce new mappers to this style.

quincylvania:
> Apologies I got too excited and didn't read that you already had "1" as a focus shortcut. Nevermind!

Oh no problem. I'll just leave it as-is then unless somebody really needs a second key.

> The only other thought that came to mind is if the walkthrough would need to be adjusted to introduce new mappers to this style.

Yep. We'll certainly need to update the walkthrough before the release.

bhousel:
Love this!  I'm excited about how much more mobile-friendly this looks.

We will really have to redo the walkthrough significantly (I see @iandees just mentioned this).  I'm pretty ok with this though.

> Support for vertex-only features (and eventually relations)

I'm curious why vertex features don't work for this flow?  That seems like a big issue.  (I can live without relations 😆)

quincylvania:
> I'm curious why vertex features don't work for this flow?

They'll work fine, we just have to add extra logic to require that they can only be added to ways. Otherwise the newly added point won't match the preset.

tastrax:
Can the selected feature remain until changed? That way, if you are doing a lot of one thing you dont need to continually select it each time.

quincylvania:
> Can the selected feature remain until changed? That way, if you are doing a lot of one thing you dont need to continually select it each time.

@tastrax This sounds like what we're planning in #5874. Locking can't be automatic since then you couldn't click-to-select features.

JamesKingdom:
This looks great! 👍 
I see that `1 focuses the search bar. `
Would it be possible to have an option where the 1,2 and 3 keys still work for node, line and area respectively?  I much prefer selecting these with the keyboard.

quincylvania:
> Would it be possible to have an option where the 1,2 and 3 keys still work for node, line and area respectively? I much prefer selecting these with the keyboard.

Keys <kbd>2</kbd> through <kbd>0</kbd> will select the favorites, so you could set node, line, and area to be  <kbd>2</kbd>, <kbd>3</kbd>, and <kbd>4</kbd>. We need <kbd>1</kbd> for focusing the search field itself.

bhousel:
I'd also like to try to find good way to preserve <kbd>1</kbd><kbd>2</kbd><kbd>3</kbd> for Point, Line, Area.  It's been this way for a very long time and it would be really disruptive to many of our users to change it.

Could we use something like <kbd>tab</kbd> to focus the search, and then leave Point, Line, Area favorites by default (but the user can choose to unfavorite them and replace them with others).

manfredbrandl:
> Could we use something like tab to focus the search, and then leave Point, Line, Area favorites by default (but the user can choose to unfavorite them and replace them with others).
Maybe the ^ left of 1 could be used? For me tab which is usually used to switch focus forward to the next item.

quincylvania:
> Maybe the ^ left of 1 could be used? For me tab which is usually used to switch focus forward to the next item.

On US keyboards the <kbd>`</kbd> key is to the left of <kbd>1</kbd>. iD uses this for toggling the sidebar or else it'd be a good choice.

slhh:
@bhousel 
>I'd also like to try to find good way to preserve <kbd>1</kbd><kbd>2</kbd><kbd>3</kbd> for Point, Line, Area. It's been this way for a very long time and it would be really disruptive to many of our users to change it.

+1
We can let these keys focus the add feature search box and also preselect a geomety type, limiting search results to matching presets.

This would generate a "select feature type before drawing" workflow, which is as efficient as the old workflow. I think the old workflow could be dropped in this case.

We might use <kbd>0</kbd> to focus the feature type search box without preselecting a geometry if desired. However, I don't see any usecase of the geometry independent feature type search, which does really need a shortcut.

quincylvania:
I added drag-and-drop support for reordering and removing favorites.

favorites drag and drop

BjornRasmussen:
Will iD remember which presets I've marked as favorites? I would hate to have to set them all up again each time I edit.

quincylvania:
> Will iD remember which presets I've marked as favorites? I would hate to have to set them all up again each time I edit.

Yep. iD remembers your favorites, including the order.

BjornRasmussen:
> Could we use something like tab to focus the search, and then leave Point, Line, Area favorites by default (but the user can choose to unfavorite them and replace them with others). 
I agree with this.

bhousel:
This seems pretty great!  I'm happy if you want to merge it so people can try it out, and we can get a jump on rebuilding the tutorial.

slhh:
@quincylvania 
>I added drag-and-drop support for reordering and removing favorites.

This gives me an idea to set the favorites:

- Make the items of the feature type search dragable to the favorites bar.
- Make the feature type of an selected object dragable from the inspector to the favorits bar.

This could avoid to clutter the user interface with any control to set the favorites.

quincylvania:
> - Make the items of the feature type search dragable to the favorites bar.
> - Make the feature type of an selected object dragable from the inspector to the favorits bar.
> This could avoid to clutter the user interface with any control to set the favorites.

@slhh Good ideas! It'd be cool to make iD more drag-and-drop friendly throughout the UI. I wouldn't get rid of the favoriting button though since drag-and-drop is less discoverable and more difficult for users who have trouble with motor functions.

bhousel:
> This gives me an idea to set the favorites:
> 
> * Make the items of the feature type search dragable to the favorites bar.
> * Make the feature type of an selected object dragable from the inspector to the favorits bar.
> 
> This could avoid to clutter the user interface with any control to set the favorites.

I'm still pretty happy with using the ⭐️star to favorite.  I think people can figure that out and it doesn't clutter anything.

slhh:
@quincylvania 
In many cases user want to map features which do already nearby.
How about including the preset of the currently selected element in the favorite bar temporarily.
When no element is selected, the preset of the most recently selected one might be used.

When the temporary favorite is placed slightly separated from the other ones, dragging the temporary one towards the other ones could make the temporary one permanent. A tooltip can make this discoverable, and a short drag distance would likely be less difficult for users who have trouble with motor functions.

Example workflow to add a park bench:

- Select a nearby park bench.
- Press the button of the temporary favorite.
- Place the new park bench.

slhh:
@quincylvania 
>Mappers who prefer the old behavior can simply favorite the generic presets and map as before.

This requires some inconsistant behaviour.
When e.g. favoriting an untagged line (e.g. a multipolygon member), applying the favorite should not ask for the feature type, as is is already defined as an untagged line, and the user might need access to memberships or tagging.
In contrast, using the generic favorite to get the old workflow, needs asking for the feature type.

slhh:
@quincylvania 
The add a feature search field occupies much valuable space in the top bar. Until it gets focussed, it doesn't have any other function than being a click target to get it focussed.
The unfocussed add a feature search field should be replaced in-place by 4 buttons:

- A generic "Add" or "Add feature" button 
- Icon only buttons for point, line, and area geometry

When clicked, each button should unhide and focus the feature type search field and hide all these buttons. In addition, the geometry buttons should preselect the geometry type to filter the search results.

This would reinstate the full efficience of the old workflow to the new workflow.

quincylvania:
@slhh 
> How about including the preset of the currently selected element in the favorite bar temporarily.

I like this.

> When no element is selected, the preset of the most recently selected one might be used.

Instead of just one and just temporarily, I made it so the recently-used presets live alongside the pinned favorites. Let me know what you think of this in practice.

> When e.g. favoriting an untagged line (e.g. a multipolygon member), applying the favorite should not ask for the feature type

The generic preset favorites do work differently than most favorites, but for good reason. We need to ask for the feature type to exactly replicate the behavior of the existing mode buttons.

> When clicked, each button should unhide and focus the feature type search field and hide all these buttons. In addition, the geometry buttons should preselect the geometry type to filter the search results.

This seems like a more complicated and less customizable version of just favoriting the point/line/area presets. I did make those the defaults, so the buttons will appear next to the search field upon launching iD. Filtering the search results by geometry would be nice, but I think that control should be in the results popover itself.

slhh:
@quincylvania 
>>    How about including the preset of the currently selected element in the favorite bar temporarily.
>
>I like this.
>
>>    When no element is selected, the preset of the most recently selected one might be used.
>
>Instead of just one and just temporarily, I made it so the recently-used presets live alongside the pinned favorites. Let me know what you think of this in practice.

Keeping some recently-used presets makes sense, but the preset of the most recently selected element should be included, even if the preset was not applied in the current session.
This is required to get the easy workflow of cloning the feature type from already existing elements, without any need to search even for the first time.

bhousel:
> Keeping some recently-used presets makes sense, but the preset of the most recently selected element should be included, even if the preset was not applied in the current session.

Yes, the appearance of the preset ribbon will persist between sessions.

brianhatchl:
I had a quick play with this.  Overall I like the new search bar.  I saw the discussion above and in slack, so I may be in the minority on this, but my one nitpick is that recent presets are restored to the toolbar on refresh.  It seems to quickly lead to toolbar button-itis with no way to clean it up (such as un-favoriting provides).

I would be okay if recents were shown in the toolbar during the session, but not restored on reload.  I realize this is not how the sidebar preset editor behaves (recent presets are restored from previous session).

Issue #5342, opened 2018-09-29
Tamil characters are broken


mohamedmahir:
Hi 

I found an issue with road-ways which is not displaying tamil text correctly. 
https://www.openstreetmap.org/edit?editor=id#map=21/13.07963/80.28161

Road should display:முத்துசுவாமி சாலை

but it is splitted and showing like below.

ம+ு+த்த+ு+ச+ு+வ+ா+மி ச+ா+ல+ை

Please refer the unicode characters for tamil here.
http://www.fileformat.info/info/unicode/block/tamil/list.htm

in chrome console, i see error as below:

//Refused to load the font 'https://cdn.joinhoney.com/fonts/icons/honey-icons.woff?vbeftd' because it violates the following Content Security Policy directive: "font-src 'none'".//

Thanks
Mahir

mmd-osm:
Follow up for: 
https://github.com/openstreetmap/openstreetmap-website/issues/2004

You didn’t mention the error message last time, so maybe this is really a csp issue.

bhousel:
My guess is that the CSP issue is unrelated - that's a browser extension trying to load some other icons.

I just tried it out on my machine (Chrome 68.0.3440.106 on macOS 10.13.6) and it does indeed look wrong.

<img width="561" alt="screenshot 2018-09-29 11 04 45" src="https://user-images.githubusercontent.com/38784/46247317-99c4da00-c3d7-11e8-83fe-4fd958e039b1.png">

The issue seems to be only with characters drawn along the textpaths, because the road name looks correct in the sidebar:

<img width="375" alt="screenshot 2018-09-29 11 04 58" src="https://user-images.githubusercontent.com/38784/46247321-9e898e00-c3d7-11e8-84f9-9c51dc9d4c64.png">

cc @mapmeld who knows much more about this than I do..

mapmeld:
I just checked and Hindi has the same issue, only in roads, and I don't see it in Firefox (so, it's a Chrome/Webkit SVG textPath issue). Arabic is also doing something new with word order. I don't get the font-loading error - only thing which fails to download is piwik js.

![screen shot 2018-09-29 at 8 41 00 am](https://user-images.githubusercontent.com/643918/46247686-e1416b00-c3c3-11e8-9008-7b339b2e414d.png)

My guess is that Chrome recently changed textPath in an unexpected and uncool way.

[...]

PR #6018, opened 2019-03-07
Fix Arabic script road labels on Chrome

mapmeld:
Long time no see! Good news: I have updated my workaround for Arabic road labels in Chrome! (updated text below as I test the problem)

## The Problem

Current broken label in Erbil at real and magnified size (expected: شەقامی کوردستان )

screen shot 2019-03-06 at 11 02 41 pm

This label helped me understand what was unnoticed or broken by Chrome since our earlier fix. The first/rightmost two letters شە get broken off and reattached to the end/left of the label, even though there is no whitespace char there? This is strange and happens on other road labels and letters.

## The Solution

In December I published <a href="https://github.com/Georeactor/alif-toolkit">alif-toolkit</a> with a JSON structure of all Arabic Unicode presentation forms. This replaces our old array with a more comprehensive list. I am investigating exactly what else should happen next (is it presentation chars reversed, or words / baselines?) but here's what I have

شەقامی کوردستان  now corrected

screen shot 2019-03-06 at 10 40 51 pm

Making sure our numbers stay LTR in Kuwait City:

screen shot 2019-03-06 at 10 36 30 pm

Hebrew labels unchanged and looking correct!

screen shot 2019-03-06 at 10 23 51 pm

## Quick Todo

One TODO is I need to fix Arabic digits, thanks

Also, bad news: #5342 (Tamil and other South Asia / Southeast Asian complex scripts, likely also Dhivehi) is still unfixed since our conversations with Google Chrome team. You can see examples of it failing on https://mapmeld.com/text-path-problems/ . We cannot use the same presentation-characters fix as Arabic, because there are no Unicode characters representing all of the possible shapes and combinations.

### DIFF
--- a/modules/util/svg_paths_rtl_fix.js
+++ b/modules/util/svg_paths_rtl_fix.js
@@ -1,16 +1,16 @@
 // see https://github.com/openstreetmap/iD/pull/3707
 // https://gist.github.com/mapmeld/556b09ddec07a2044c76e1ef45f01c60
 
-import { BaselineSplitter, WordShaper } from 'alif-toolkit';
+import { WordShaper } from 'alif-toolkit';
 
 export var rtlRegex = /[\u0590-\u05FF\u0600-\u06FF\u0780-\u07BF]/;
 
 export function fixRTLTextForSvg(inputText) {
-    let ret = '', rtlBuffer = [];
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks for the update @mapmeld this is really great 👍 
I appreciate your work on making the text shaping look as good as it can!

PR #6036, opened 2019-03-10
update public transport route presets

nlehuby:


quincylvania:
@nlehuby Looks great, thanks! 👍

PR #6039, opened 2019-03-12
Add landuse=industrial to scrap_yard's tags.

quotquot:
If I understand correctly, any preset `iD/data/presets/presets/foo/bar/baz.json` should have `foo=bar` in its tags, along with a test on `baz`.

If this is so, then `iD/data/presets/presets/landuse/industrial/scrap_yard.json` is missing `landuse=industrial`, which `slaughterhouse.json` in the same directory has.

This PR adds `landuse=industrial` to `scrap_yard.json`.

### DIFF
--- a/data/presets/presets/landuse/industrial/scrap_yard.json
+++ b/data/presets/presets/landuse/industrial/scrap_yard.json
@@ -11,6 +11,7 @@
         "area"
     ],
     "tags": {
+        "landuse": "industrial",
         "industrial": "scrap_yard"
     },
     "addTags": {


quincylvania:
@quotquot Nice job noticing this!

In this particular case, `landuse=industrial` is included in the preset's `addTags` property instead of the `tags` property. It may seem quirky, but this pattern gives us some benefits. It means that the tag isn't needed to match the preset, but will be added when you set the preset. In the next version of iD, this also means that features with `industrial=scrap_yard` but not `landuse=industrial` will be flagged by the validator.

PR #6076, opened 2019-03-21
Relocate fill patterns to individual presets

RudyTheDev:
Follow-up on https://github.com/openstreetmap/iD/pull/5491#issuecomment-439534028 -- I moved the data from hard-coded `patterns` variable in `areas.js` to the individual json preset files as `"fill"` field for ease of maintenance and to follow the preset-based design.

The two syntaxes in preset jsons are:

    "fill": "grass"

and

    "fill": [
        "forest",
        { "leaf_type": "broadleaved", "pattern": "forest_broadleaved" },
        { "leaf_type": "needleleaved", "pattern": "forest_needleleaved" }
    ]

I added a corresponding entry to `preset.json`, but I have no idea how to describe `"type"` field when the value can be one of two structures and the child may also be one of two structures. This means it doesn't validate the type at build-time, which is bad-ish. The implementation assumes it will be in one of the expected syntaxes. Not sure what to do about this?

This change means the `drawAreas` now looks up presets for each area it draws, which I imagine is costly. But it seems like other places do the same, like `labels.js`. I couldn't find the preset being previously stored anywhere in entity or any sort of cached lookup I could do. Is there a better way to do this or do we have to look up the preset every time?


bhousel:
> I added a corresponding entry to `preset.json`, but I have no idea how to describe `"type"` field when the value can be one of two structures and the child may also be one of two structures. This means it doesn't validate the type at build-time, which is bad-ish. The implementation assumes it will be in one of the expected syntaxes. Not sure what to do about this?

I'm pretty sure the JSON schema validator can do this, but I'd have to look up how.
I do wonder whether we can make this a little simpler.. I can't tell from looking at the second example how it works, but I know it does some kind of tag evaluations to pick a fill pattern.

> This change means the `drawAreas` now looks up presets for each area it draws, which I imagine is costly. But it seems like other places do the same, like `labels.js`. I couldn't find the preset being previously stored anywhere in entity or any sort of cached lookup I could do.

It is actually cached: `preset.match` wraps what it does in a "transient" which caches the result in the graph itself.  If the graph doesn't change (which it doesn't while rendering) it returns the same result. (a fancy word for this is ["memoization"](https://en.wikipedia.org/wiki/Memoization))

https://github.com/openstreetmap/iD/blob/e585090192ae8463b9bea266f733ad54c0b48b57/modules/presets/index.js#L44-L55

quincylvania:
> I do wonder whether we can make this a little simpler.. I can't tell from looking at the second example how it works, but I know it does some kind of tag evaluations to pick a fill pattern.

I suggest only using the simple `"fill": "grass"` syntax and creating further-specified presets where needed.

RudyTheDev:
> I suggest only using the simple `"fill": "grass"` syntax and creating further-specified presets where needed.

While "Broadleaf Forest" or "Marsh Wetland" or "Muslim Cemetery" and such make sense as presets, the example where it doesn't quite work is `traffic_calming=island` + `surface=grass` or you'd have to create a "Grass Traffic Island" preset.

Another case is something like `landuse=forest` + `leaf_type=broadleaved` + `leaf_cycle=evergreen` (which has 60k usages), which would need a "Broadleaf Evergreen Forest" preset.

I don't know iD's stance on having such specific presets? I imagine the problem is that now the users expect all major variants to have a preset ("Asphalt Traffic Island", "Needleleaved Evergreen Forest", etc.). The way I implemented tag evaluation before was to not require new presets just to add fill patterns. But I imagine we don't need that if it's okay to add presets.

quincylvania:
> Another case is something like landuse=forest + leaf_type=broadleaved + leaf_cycle=evergreen (which has 60k usages), which would need a "Broadleaf Evergreen Forest" preset.

I didn't see this combination in your PR so I didn't consider it. Were you planning on adding fills this specific? I'm not saying if you should or not, it might be fine! I was thinking presets like "Broadleaved Woods" would be okay and the `leaf_cycle` would still just be a field.

The traffic island one is a bit of a problem. I would be okay with hardcoding little edge cases like that for the sake of avoiding the complex JSON property.

bhousel:
I'd rather not create separate presets for all the surface types.

It's possible this might not belong in presets after all... I feel bad for pushing you in that direction with [this comment](https://github.com/openstreetmap/iD/pull/5491#issuecomment-439534028)  but I was definitely thinking it would be simple and not need to special case different rules for different tags.  

Since this PR doesn't actually bring anything new functionality, is it ok if we close it but think about other options?  We also have a stretch goal to avoid using CSS and SVG for so much, and switch over to rendering more things with Canvas, so I think having cartographic style decisions stored somewhere in code (and not presets) might be where iD is headed.  I thought we had an issue for this, but I don't see it right now.  #3471 has some, but it's kind of gone stale.

RudyTheDev:
The problem that I see with `areas.js` is that it needs to repeat and hard-code the tags even when presets exist. If presets like "Broadleaf Forest" or "Marsh Wetland" or "Muslim Cemetery" can be added, then the JSON can indeed be just `"fill": "grass"`. Whereas `areas.js` would still need to repeat all the tags. With presets, only the edge cases like "Grass Traffic Island" need to be hard-coded like quincylvania says.

I personally don't like hard-coding even the special cases, but I can see the benefits of keeping the preset JSON clean. On the other hand, hard-coding _all_ the cases doesn't seem like a good practice to me. One can always migrate data or change the implementation, but funky stuff happens when the two are mixed. I would personally argue that separating data from business is always worth it.

That said, I can't comment on how this might affects eventual Canvas transition. I imagine all the CSS stuff breaks, although fill patterns look like just names that `defs.js` translates to CSS-able file names.

bhousel:
> The problem that I see with `areas.js` is that it needs to repeat and hard-code the tags even when presets exist.

Yes this is not ideal to have the fillpatterns hardcoded in this file.. However I'm ok with having it this way for right now, and I don't want to add a bunch of extra-specific presets like "Muslim Cemetery" just to let people declare fill patterns for them.  

I'm going to close this PR because styling info is not something I want to add to the presets right now (and I do feel really bad about closing it!)  We need to think about where style declarations should live eventually - I don't have the bandwidth for it right now.

PR #6137, opened 2019-04-03
Add preset subtitle property

### LINKED ISSUE #6131: Add subtitle property to presets
Mappers should feel confident that they're selecting the appropriate preset for the thing they want to map. We should show an optional subtile below the preset's title, similarly to how we show brand preset titles in two parts:

54547771-ca480280-497c-11e9-8c0f-af5fb28ea289

This would help us differentiate similar presets (Casino vs. Adult Gaming Center, Drain vs. Ditch) and further define unusual feature types (Adit, Cycle Path Crossing). It would also let us simplify long preset names (Nordic or Crosscountry Piste/Ski Trail).

The subtitle would be character-limited and would not a replacement for the full description text from the OSM wiki.

### Examples

**Adit**
Horizontal Mine Entrance

**Cross-country Ski Trail**
Nordic Piste

quincylvania:
Closes #6131.

This adds an optional `subtitle` property for presets that displays below the name. It also switches name suggestion presets to use it, removing some hacky code.

I didn't merge this directly since it has some special implications for translations that need to be worked out. The usefulness of the subtitle depends on the language, and thus there is not a one-to-one mapping of what presets should get subtitles in English versus any other given language.

Update: I made it so subtitles don't default to English if there's no translation available.

Screen Shot 2019-04-02 at 7 58 22 PM
Screen Shot 2019-04-02 at 8 08 55 PM
Screen Shot 2019-04-02 at 8 09 37 PM


westnordost:
The mechanism currently possible to deliberately not translate a `terms` field as a translator would be to simply enter `<>`. The import from Transifex removes this field in this case because it removes everything in between `<` and `>` while in Transifex itself I think it should turn up as "translated".

This method is not described or communicated anywhere to the translators yet, but it could, and the same method could be used for the subtitle descriptions.

quincylvania:
Putting this feature on hold. See https://github.com/openstreetmap/iD/pull/6412#issuecomment-500976378.

Issue #2248, opened 2014-06-05
'Straighten' on long ways can separate junction points


skylerbunny:
In a case where you have a long way, and are on one end of it (but iD knows that way is straight 'enough' to allow 'straighten' as a choice), it's possible to straighten a way from one end, and have it break junction points on the other end (presumably where loading ways has fallen out of scope for the editor, so it doesn't know to carry the junction points along).

Example: Consider this way: http://www.openstreetmap.org/way/277201243 Then, zoom to the north end, and enter the iD editor. Straighten will be given as an option for the way. However, if you use it, as you scroll south, you'll eventually notice that junctions on crossing ways haven't been carried along, leaving the way orphaned.

This is sort of a data loss bug, in the sense that junctions and navigation can be unintentionally broken by a user who doesn't realize this can happen. This would especially be true in cases where one is dealing with a large city grid, and just wants to 'straighten the roads', again, not realizing that ways further out from the work area are being broken apart.

bhousel:
Planning to use `iD.restrictions` to fix this too.
WIP on this: 8467433

verdy-p:
Why not just straightening ways only on the part which is fully in the fully loaded area ?

This means that you could create one or two extra nodes where they cross that area, but still allow the fully loaded section between these nodes to be straightened.

The same logic could be used for orthogonalization (or circularization), to limit the scope of the action only to the viewport, not touching whay is not visible: it would be a convenient way to limit the action: we just need to zoom into the interesting area where we want to "regularize" the drawings of the selected object.

In my opinion it is always unsafe to edits geometries of objects that are not fully visible, but it is safe to do it on partial objects

verdy-p:
Note: this logic works as long as the visible area is fully loaded. We need a visible indication for this, and as long as this is not the case, these regularization options should be disabled.

Note however that the server API frequently has issues (returns HTTP 500 errors due to timeout/resources) when querying parent objects of an object selected by id (this generally happens when there are several parent relations, as it seems that the server needlessly need to load all their children and fails with an "internal out of memory", when we are just interested in just getting the reference of parents, for exampe when selecting nodes on important highways where there are multiple bus routes including long distance ones, or admin boundaries for low admin_levels and multiple boundaries sharing this way).

Such server-side API issues will most often not occur if we query the API by area (that area can be extremely small, a few centimeters, around a single node in the selection): we still have the full list of parent objects (all parent relations for any object, plus the parent ways only for nodes) and the server replies immediately without failing !

[...]

Issue #6086, opened 2019-03-22
Avoid rerendering the sidebar so frequently


bhousel:
We currently rerender the sidebar on `history.change` events, but this is often unnecessary.

For example, here's a graph showing a frame while straightening a line.

![Screenshot 2019-03-22 14 35 24](https://user-images.githubusercontent.com/38784/54845459-c3d4c600-4caf-11e9-8f40-4b66c63b0b06.png)

- `_act` is the actual action of straightening the line
- `immediateRedraw` does some stuff as a result of that, mostly redrawing the vertices (ok fine but maybe some room for improvement)
- `historyChanged` is redrawing the sidebar and all of the fields on it (totally unnecessary)

Issue #6054, opened 2019-03-17
Performance degraded after validation was added


tordans:
This is probably similar to https://github.com/openstreetmap/iD/issues/5971.

I noticed that round about the time when validation where added, the editing in Berlin is really slow. This is for both the public version as well as http://preview.ideditor.com/master/#background=Berlin-2018&disable_features=boundaries&map=20.45/52.48086/13.44218.

How can we help figure out what is going on? Would one of those Google Chrome rending-graphs help? I don't know much about them, so I would need a tutorial what to provide.

My system did not change recently, so it's either updates to the OS, Chrome or iD that change the performance.

MacBook Pro (Retina, Mitte 2012), 2,3 GHz Intel Core i7, 16 GB 1600 MHz DDR3
Operating System Mac OSX 10.14.x
Web Browser Chrome 71.0.3578.98

**Update**

- Its really bad when I type my commit message. I type and nothing happens and then the text appears all at once.
- Would screen-videos help?

bhousel:
Yeah this is #5971 - I don't think it should have been closed.

> How can we help figure out what is going on? Would one of those Google Chrome rending-graphs help? I don't know much about them, so I would need a tutorial what to provide.

Yes, profiling in the Google Chrome dev tools would be very helpful.

quincylvania:
From my own testing on this, the degraded performance is almost entirely due to the crossing ways check. iD currently revalidates everything you've edited after every further change, so if you're editing long roads or such this can add up fairly quickly. We have #5901 to only revalidate features you've actually edited since the last validation. We should do that soon.

> Yeah this is #5971 - I don't think it should have been closed.

Oops.

bhousel:
#4731 might be related

bhousel:
This is much better now after #6140 ..

tordans:
@bhousel I used the preview system some more and first thought its just the changeset comment field (the separate issue I created) but now I feel big performance issues in other places.

I tried to debug it better this time.

Location: http://preview.ideditor.com/master/#background=GEOSN-DOP-RGB&disable_features=boundaries&id=w151467550&map=19.97/50.99388/12.80095

**Video:** https://www.dropbox.com/s/azpnxtrdlqly35u/id-performance-reflow.mov?dl=0
There are quite a few seconds between clicking something and being able to use it.

To me it looks like one issue is query?

First edit (outside the video):
[...]

Issue #5938, opened 2019-02-22
False warning for highways connected only to unloaded highways


SilentSpike:
See this scenario:
![image](https://user-images.githubusercontent.com/5459452/53269609-3ec2b500-36e1-11e9-8dd7-9409a8d7ec4b.png)

If you open location `map=18/57.08568/-2.14301` and hover over the highway, it is flagged as disconnected. In reality, it is connected, but just so long that (I assume) iD hasn't loaded the data where it connects to the rest of the network.

quincylvania:
@SilentSpike Yep, I've seen this plenty of times in testing. It'd be nice to resolve this accurately by downloading more nearby/connecting data (see #5180). In the meantime we could ignore disconnected highways that overlap non-downloaded tiles.

bhousel:
related #2248

pnorman:
> @SilentSpike Yep, I've seen this plenty of times in testing. It'd be nice to resolve this accurately by downloading more nearby/connecting data (see #5180). 

You'd have to issue map calls along the length of the way, which could get expensive. You'll also then up with a bunch of new ways which exit the downloaded area.

> In the meantime we could ignore disconnected highways that overlap non-downloaded tiles.

I think this is the better choice. Errors of this type outside the downloaded area were obviously not created in the current editing session.

bhousel:
> In the meantime we could ignore disconnected highways that overlap non-downloaded tiles.

I just added this to #6140  🎉 

I also noticed that both the `disconnected_ways` and `almost_junction` validations are vulnerable to this, so I'm fixing it in both places. 

In this example below, the left tile was loaded first, and an `almost_junction` warning was raised before the right tile loaded.

<img width="997" alt="Screenshot 2019-04-10 12 14 02" src="https://user-images.githubusercontent.com/38784/55899788-ababe480-5b93-11e9-8267-0ed5eed7c371.png">

I'll think more about this too, because anyplace in the code that we test the `parentWays` of a vertex that might not be on the visible map is risky.  

For example, we check `parentWays` of vertices in the turn restriction editor, but the vertex has to be visible and selected for iD to even do anything.  Might be worth adding defensive code anyway, in case the intersection extends off the screen and into unloaded map.

bhousel:
Done in #6140

PR #6140, opened 2019-04-04
Validation and change performance

bhousel:
This PR will attack a bunch of related performance issues..
Basically we can do slightly more work in `coreDifference` to track what kind of stuff got changed in the difference, so that other parts of the iD code can use this information to avoid expensive processing.

- [x] #6086 - avoid rerendering the sidebar for geometry-only changes

- [x] check `coreTree`, which can use the improved difference to avoid expensive delete/reinsert of entity bounding boxes

- [x] #6054 / #5901 / #5898 - validator performance
We are currently validating all the changes in the user's edit history every time the history changes significantly - and we should instead run validation per-entity.  

- [x] #5906 - validate unedited features as well as edited features
I accidentally implemented this today, and was surprised by how well it worked.  Needs some UI to let the user opt in to it. 

- [x] #5938 - false warning about disconnected features that extend beyond the loaded map 
These are already bad, but we see them a lot once we try validating all the features per #5906

- [x] #2248 / #4890 - should really fix these old bad bugs once we have the code to let other parts of iD code ask the osm service what tiles have been loaded per #5938

- [ ] considering: avoid rendering some geometry in map.js#draw
This depends knowing which tags affect style.  e.g. `highway` controls the color of lines, `oneway` controls oneway markers, etc.   There is no great solution to this problem, since style still lives in both JS and CSS for now.

- [ ] Add ability to flag a warning as a false positive.  Something like "Don't warn me about this again"

- [ ] Add lots of tests


bhousel:
I'm posting an update here because I've been working on this for a while.  It's really coming along well.  Here's what is on my mind:

### performance
- iD now knows more about what changed in an edit, and we can use that information in `coreDiffrence` to avoid extra renders of the sidebar.
- We're using deferred callbacks a lot more now.  Profiling iD has gotten more interesting, because as some things have gotten faster it has exposed other things for us to improve.  For example, d3.zoom is causing reflow now.  I think upgrading to d3.v5 might fix this (#4929), which we can finally do now that #6087 polyfilled the stuff it needs.

### validation
- We've changed from validating everything the user touched on every history change (very frequent) to validating just the changed features at specific times, and this seems to be working really well.
- iD can now check whether data tiles have been loaded, and will prevent certain actions (like straightening especially) if we need data from unloaded tiles, and will schedule loading of that data during idle times.  This closes #2248, which is probably the worst open bug in iD. 
- I've added the ability for some validation fixes to be done automatic, and added a wrench 🔧 button.  I still need to add a Fix All button or something similar.  This is a feature JOSM has had for a long time, so it feels good to add it to iD too.
- The issue tooltips are driving me nuts.  I don't feel like having the same tooltip "some tags change over time.." repeated 1000 times in the list is helpful, and the tooltip placement is wrong when the list grows very long.  It's not using `scrollTop` correctly or something.
- We have checkboxes under the "Rules" section to enable/disable rules, but they are currently not working.  I have to put this code back.
- I've added radio buttons for what to validate and where.  I really like these.  Fun fact:  iD is actually validating everything loaded from OSM - the buttons really just filter the list.  I'm testing this around Manhattan and it actually performs fine.  Validation is deferred using the `requestIdleCallback` feature in modern browsers.
- I'm a bit uneasy with the deferred validation.  The validator keeps a "validatedGraph" reference, knowing that the graph can change between validation calls.  The individual validators all still check `context.entities()` and `context.graph()` a lot.  I think I would prefer to have each validator module to accept a `coreDifference` instead of a list of entities. 
- `crossing_ways` is still the biggest performance drag.  I don't understand the code very well, but I see lots of things in there that could probably be faster.
- `outdated_tags` is interesting.  I'd prefer for it do a trial run of changing the tags, then we could show the user exactly what tags are going to be changed.  What I really want to see is almost something like "git diff" or "patch" style of output with tags added/removed/modified. I could actually build this with the `node-diff3` module we use.  I don't think most users want to see this, but it would make me feel better anyway.
- I don't know what to do about `many_deletions`.  It's the only one that runs on the whole history instead of a list of entities.  I have it commented out right now, and I'm not even sure whether we want to keep it, or whether it helps anyone.  What are users supposed to do when they see this?  Every time I've seen it, I have been cleaning up overnoded ways, or deleting circular ways from the end of cul-de-sacs.  Then I think "huh ok" but it's not an actionable warning. 


Here's a screenshot of what it looks like today:
Screenshot 2019-04-14 12 08 15

bhousel:
This doesn't do everything that I would like it to, but it's good to merge now.   The validator is really impressive, and I'd love for people to try it out in http://preview.ideditor.com/master

We're going to continue to do a lot of testing this week, and see if we can add a few more improvements before the next release ✨

Issue #6135, opened 2019-04-02
Flag sources with incompatible licenses (e.g. Google Maps)


quincylvania:
We currently flag changeset comments that mention Google:

<img width="390" alt="Screen Shot 2019-04-02 at 7 41 02 AM" src="https://user-images.githubusercontent.com/2046746/55400080-de623700-551a-11e9-8cd8-d0bbc4c45cdc.png">

We should create an "Incompatible Data Source" validation warning that similarly flags `source` tags with such text on features.

Edit: This comment previously mentioned turning the changeset warning into a native validation issue. This would involve validating the actual changeset, which the validator isn't set up to do right now. We could consider this later in a separate issue if we really want to.

PR #6146, opened 2019-04-05
Allow for multiple invalid sources

jguthrie100:
Updates the fix for #6135 and just adds a loop that loops through a list of invalid sources


quincylvania:
@jguthrie100 Looks great, thanks!!

Issue #6149, opened 2019-04-06
Unable to disconnect closed way at endpoint node


quincylvania:
Disconnecting a closed way from other ways at its endpoint doesn't work when both the node and the way are selected. This is particularly annoying since it's not clear which point is the endpoint on closed ways.

<img width="592" alt="Screen Shot 2019-04-06 at 4 23 28 PM" src="https://user-images.githubusercontent.com/2046746/55676348-7435f300-5888-11e9-87ea-16a25b0a735c.png">

PR #6161, opened 2019-04-08
Leave way as closed when disconnecting

jguthrie100:
This adds to the fix in #6149 and makes it so that when a closed way is selected as well as the node to disconnect, the closed way gets disconnected from anything it is connected to, but remains closed.

I think this stay truer to the existing functionality where when a way is selected, the way is disconnected independently and every other connection remains as it was. Chances are that the user just wants to disconnect the closed way from the junction node, rather than opening up the way as well (which they can just do after if that was actually their intention)

If they _do_ want to disconnect everything at the junction node, they can just right click the node without selecting any connected ways, and everything will be disconnected as usual.

disconnect_closed_way2

disconnect_closed_way1

I've also added a fix so that the option to disconnect doesn't even appear unless all selected ways are connected to the selected node, to save confusion on what actually will be disconnected, which is closer to how it was before #6149 was fixed.

disconnect_closed_way3

### DIFF
--- a/modules/actions/disconnect.js
+++ b/modules/actions/disconnect.js
@@ -59,7 +59,7 @@ export function actionDisconnect(nodeId, newNodeId) {
             } else {
                 way.nodes.forEach(function(waynode, index) {
                     if (waynode === nodeId) {
-                        if (way.isClosed() && parentWays.length > 1 && wayIds && wayIds.includes(way.id) && index === way.nodes.length-1) {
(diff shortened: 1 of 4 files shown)


quincylvania:
@jguthrie100 Thanks so much! This is the way I should have solved it to begin with 🙈

jguthrie100:
We live and learn! 😄

Issue #6155, opened 2019-04-08
Add preset for natural=shingle


Mahabarata:
Hi,

in Transifex, for natural=*, we can find the words scrub, wood, cliff and so on so these words can be translated in any langage.
But shingle is not in Transifex : so in ID natural=shingle has no translation but the general translation for 'natural' which is quite bad.

Could you fix that ?
Best regards

bhousel:
sure, thanks for letting us know!
https://wiki.openstreetmap.org/wiki/Tag:natural%3Dshingle

Mahabarata:
Sorry to come back : I just try the other one (in the list of ID for "natural") and find some strange things :
- in ID, natural=valley is not translated in french : but natural=valley is in transifex with a good translation
- natural=rock (170 000 objects in taginfo) is, like shingle, not in transifex

Best regards

PR #6180, opened 2019-04-13
Update natural.json

dhruv1701:


bhousel:
Looks like this fails because there is not yet a preset for `natural/shingle`  (re: #6155)
Please check the [preset README](https://github.com/openstreetmap/iD/blob/master/data/presets/README.md) for more information on how to add one, and look at other examples under `data/presets/presets/natural/*.json` - there is probably one you can copy to get started.

dhruv1701:
Thanks for helping.

PR #6183, opened 2019-04-14
Add validation for nordic pistes

yvecai:
While piste:difficulty is more than just useful for the data user, piste:grooming for cross-country skiing appears to defaults to quite different practices across the world.


bhousel:
Same as my response on #6184 - `addTags` shouldn't add empty tags.

PR #6184, opened 2019-04-14
Add validation for downhill

yvecai:
Piste:difficulty is more than just useful for the data user


bhousel:
Sorry, I don't think we want to add empty tags by default.  This would have the effect of leaving them this way if the user doesn't fill them in.

PR #6187, opened 2019-04-14
shingle.json

dhruv1701:


bhousel:
I'm sorry, but I can't accept this.  You can't just add a bunch of duplicates in the `tags` section like that.  Also the "retweet" icon is really not a very descriptive icon for this feature.

PR #6188, opened 2019-04-15
Use address as backup label

### LINKED ISSUE #1524: Render address points and addresses

quincylvania:
I'm not sure if this would close fully #1524 but it's at least a half step toward it.

Screen Shot 2019-04-15 at 3 15 12 PM
Screen Shot 2019-04-15 at 1 57 32 PM


modules/util/util.js
```diff
@@ -97,15 +97,22 @@ export function utilGetAllNodes(ids, graph) {
...
+        var network = entity.tags.cycle_network || entity.tags.network;
         if (network) {
             name = network + ' ' + name;
         }
     }
 
+    if (!name && entity.tags['addr:housenumber']) {
+        name = entity.tags['addr:housenumber'];
+        if (entity.tags['addr:street']) {
+            name += ' ' + entity.tags['addr:street'];
```
1ec5: In some countries, the street name comes before the house number. Would it be feasible to use the address-formats.json as a guide to how to format these addresses?

1ec5: In areas where both buildings and addresses have been mapped or imported, there may still be freestanding address points for units within a multi-tenant building. Adding `addr:unit` to the display name would disambiguate these closely spaced points.

Issue #6177, opened 2019-04-13
Brainstorming: Improving address mapping and validation


tordans:
Inspired by a presentation on a German OSM(…) conference ([YouTube, Audio in German](https://www.youtube.com/watch?v=AIBi2aD-4n0)), I looked at https://addresses.lorenz.lu/oad_dach/countries which is a tool about address validation in OSM.

There are a few use cases that could be handled by iD to improve address data in OSM.

This ticket is meant more inspirational, than a specific feature request. Especially since some of the goals can be solved in different ways. I suggest we collect ideas here and create separate tickets once we haven a clearer idea what is worth solving ATM.

**UseCase: Copy paste an address (as node or when copy-pasting a house).**
I copy paste houses all the time since its less clicks to copy+move+rotate than to draw them from scratch. This has a higher risk of copying the address as well.
- iD could validate „is there another house in the viewport with the same address, if yes, suggest checking it“. Actions would be „change address = edit object“ or „ignore“

**UseCase: Copy paste a house which has an address to another road nearby, then forget to change the road.**
  - iD could validate „if the house that I edited has a street name, check if the street nearby has the same name. If not, show a warning.“ Actions would be „copy street name from street to the house“, „edit address = edit object“, „ignore“

**UseCase: I create a new house next to a road but do not add an address.**
- iD could provide a feature, saying „copy street name from the road nearby“ (just the street name).
- And in case, there is a house nearby with address: „copy address (without number) from the house next to this one“ (street name, zip, city name).
- Those feature could be part of the validator, or a separate notice / action / button on the address-preset-section.
  - Or maybe both, a separate feature but also a validator.

- Another take on this, where to add those address validations via an API to https://github.com/britiger/osm_address_db/issues/18. Or to add the validation rules to KeepRight instead. This would not provide live feedback, but at least show the issues inside the editor.

**UseCase: I copy a house with a name attribute.**
- iD could validate the name the same way as the house number (see above). This is interesting for all attributes that should usually not be copy-pasted. So also URLs, Owner. What else?
- Maybe the preset-JSON is a place where we mark those features that should be validated on copy-paste actions since they should be unique? (I would not remove them automatically on copy-paste.)

**UseCase: Complete Address street … after adding the number on the go**
[...]

PR #6189, opened 2019-04-15
Validate against duplicate name/address tags

jguthrie100:
Implements some of #6177 

This is a bit of a work in progress as I don't know if it would be better to combine it with a more thorough validation process (although it works as is), but it checks whether there are duplicates in the current graph of various tags (at the moment just _name_ and both _addr:housenumber_ and _addr:street_ together.)

Includes tests

validate_road_name

validate_duplicate_address

### DIFF
--- a/modules/validations/duplicate_tags.js
+++ b/modules/validations/duplicate_tags.js
@@ -31,7 +31,7 @@ export function validationDuplicateTags() {
                                     return entity.tags[t];
                                 }).join(',');
 
-            if (!entityTagValue.match(/^,+$/) && existingTagValues.indexOf(entityTagValue.toLowerCase()) !== -1) {
+            if (!entityTagValue.match(/^,*$/) && existingTagValues.indexOf(entityTagValue.toLowerCase()) !== -1) {
                 issues.push(new validationIssue({
                             type: type,
                             severity: 'warning',
--- a/modules/validations/duplicate_tags.js
+++ b/modules/validations/duplicate_tags.js
@@ -7,7 +7,7 @@ export function validationDuplicateTags() {
     var type = 'duplicate_tags';
     var tagsToCheck = [
                         ['Name', ['name']],
-                        ['Address', ['addr:housenumber', 'addr:street']],
+                        ['Address', ['addr:unit', 'addr:housenumber', 'addr:street']],
                         // Add more tags to check here
                       ];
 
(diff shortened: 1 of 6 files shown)


BjornRasmussen:
I think that the first example would be incorrect in 90% of all cases - chances are that you are just adding the other direction of a pre-existing road in OSM if you see this warning.

The second example, duplicate address warnings, looks great!

modules/validations/duplicate_tags.js
```diff
@@ -0,0 +1,77 @@
+import { actionChangeTags } from '../actions/index';
+import { t } from '../util/locale';
+import { validationIssue, validationIssueFix } from '../core/validator';
+
+
+export function validationDuplicateTags() {
+    var type = 'duplicate_tags';
+    var tagsToCheck = [
+                        ['Name', ['name']],
+                        ['Address', ['addr:housenumber', 'addr:street']],
```
1ec5: Since this code only checks `addr:housenumber` and `addr:street`, it’ll incorrectly flag features that share `addr:housenumber` and `addr:street` but differ in `addr:unit`. This can happen for instance when there are multiple tenants in the same building, each mapped as a point.

Nakaner:
Duplicated addresses aren't unusual and this pull request will lead to a large number of false positives. I fear that the validation rules is asking for trouble.

POIs often have `addr:*=*`tags even if multiple of them are located in the same building. There is no consensus whether this is good or bad. An editor should not enforce any point of view in this issue.

Addressing is difficult and different from country to country, from region to region. There are exemptions for all rules in that field. Duplicated addresses might be an error in many cases. An editor aiming to inexperienced users who cannot form their own opinion on difficult OSM issues, should avoid to issue warnings with a high rate of false positives because its users will blindly (they don't it better) follow the instructions in order to make the software happy.

There are a few ways to solve this issue:

* Don't show duplicated addresses as errors if they intersect (e.g. POI node in building).
* The error message explains that there are valid cases: "Multiple features have the same addresses. This might be an error but there are a lot of cases when duplicated addresses are fine. Ignore this message if you are unsure."
* Hide validation errors with a too high rate of false positives for new users.
* Disable the rule by default and enable it if the community in a country asks to do so. Then it will be there decision to accept the issues related with inexperienced users and false positives.

bhousel:
Ok, going to close this since some local communities in OSM seem to like their addresses duplicated.  #6177 is tagged as "considering" meaning, we might never do it.

cserpell:
I love this kind of checks... I think duplicate addresses are because users don't know how to use addr:unit and addr:door, and this would be a good way to learn.

polarbearing:
You did not tell in which country you are using 'addr:unit and addr:door'. 
They are not a typical part of a postal address in Germany, where I live. On the other hand, multiple businesses having the same postal address (addr:*) is typical, they are distinguished by their name. Thus I fully agree with Nakaner, to avoid flagging them as errors.

cserpell:
Interesting. I am from Chile. The case of same exact address here happens, but it is rare. Usually you can distinguish them with unit, when it is a different property or building, and door, when it is the same building, but different office numbers, for example. Thanks!

PR #6190, opened 2019-04-15
Make incompatible Google source data more generic

jguthrie100:
Change incompatible_source locale translation lookup to pass in 'Google' as an argument, rather than having it hardcoded into the text

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1297,10 +1297,10 @@ en:
       tip: "Names should be the actual, on-the-ground names of features."
     incompatible_source:
       title: Incompatible Data Sources
-      google:
+      source:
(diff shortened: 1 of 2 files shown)


quincylvania:
@jguthrie100 Thanks for the idea, but are you okay waiting to add a generic option once people want to add more sources? Otherwise the translators will have to re-translate the strings arbitrarily. I structured the yaml that way on purpose so we could add more source strings later, including generic ones. I kinda like having specific messages for Google since it's probably the most common and it lets us phrase the message more sensically.

jguthrie100:
Sure, it was just a quick idea I had going through the code

Issue #6185, opened 2019-04-14
Allow users to paste several "k=v" tag pairs into the raw tag editor


famo:
It would be nice if we could use the "Add field" textbox to add new tags, e.g. by pasting 
`roof:levels=1` into it.
Currently one has to scroll all the way down to the end of the current tags list, hit + and then paste - this is cumbersome.

Furthermore it would be awesome, if we could enter several tags at once this way, e.g. by pasting: `roof:levels=1 roof:shape=gabled`.

bhousel:
> It would be nice if we could use the "Add field" textbox to add new tags, e.g. by pasting
> `roof:levels=1` into it.

Sorry, but I disagree with repurposing the "Add Field" dropdown for this.  

If you are copy pasting tags like this, you should probably just hide the "All fields" section completely and use the "All tags" section.


> Furthermore it would be awesome, if we could enter several tags at once this way, e.g. by pasting: `roof:levels=1 roof:shape=gabled`.

Yeah this makes sense.. Currently it only supports pasting a single key-value pair, but we could allow people to paste several.

famo:
> If you are copy pasting tags like this, you should probably just hide the "All fields" section completely and use the "All tags" section.

Well, I use both. But I see now that the "Add field" box really belongs to the fields section.

> Yeah this makes sense.. Currently it only supports pasting a single key-value pair, but we could allow people to paste several.

+1
Still tho, really looking for (speed) improvements here. How about an open textbox by default ("Add tags:")? Maybe even on top of the existing tags?

bhousel:
> Still tho, really looking for (speed) improvements here.

I totally understand.. iD has to balance a lot of design concerns and we value simplicity over mapper efficiency and speed, so you might want to use a different tool like JOSM in the mean time.

I do want to make pasting tags work great, but this really is a niche need (most users don't and shouldn't care about what the tags are), so I don't want to just add a textbox for people to paste tags into and close the issue.  This actually is a very good idea and we might build something like that, but make it spark joy, you know?

tordans:
> Furthermore it would be awesome, if we could enter several tags at once this way, e.g. by pasting: `roof:levels=1 roof:shape=gabled`.

This sounds a lot like what I tried to commented/request at https://github.com/openstreetmap/iD/pull/5070#issuecomment-396031571. Maybe this new issue will nudge @AndreasHae to have another look at it :-).

AndreasHae:
[...]

PR #6211, opened 2019-04-22
Handle multiple tags when splitting up in tag editor

AndreasHae:
As requested by @tordans in #5070 and #6185:

When pasting multiple tags like "amenity=restaurant name=The Restaurant", they shall be split up and added separately. My implementation assumes that there will never be spaces in keys, so it also handles cases like "name=The Restaurant amenity=restaurant" correctly, and splits it up at the last space before the key.

### DIFF
--- a/modules/ui/raw_tag_editor.js
+++ b/modules/ui/raw_tag_editor.js
@@ -329,13 +329,13 @@ export function uiRawTagEditor(context) {
             }
 
             // if the key looks like "key=value key2=value2", split them up - #5024
-            var keys = (kNew.match(/[\w_]+=/g) || []).map(function (key) { return key.slice(0, -1)});
+            var keys = (kNew.match(/[\w_]+=/g) || []).map(function (key) { return key.slice(0, -1); });
             var vals = keys.length === 0
                     ? []
                     : kNew
-                        .split(new RegExp(keys.map(function (key) { return key.replace('_', '\\_') }).join('|')))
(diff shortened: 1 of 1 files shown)


AndreasHae:
Huh, seems like the build system complains about missing semicolons, although they are present.

bhousel:
> Huh, seems like the build system complains about missing semicolons, although they are present.

You need to put them in the anonymous functions too.. for example:

```js
list.map(function(item) { return item.trim(); });
```

tordans:
@AndrewHain <3

Will it also work if I copy multi line key-value pairs which are something like `key=value\nkey2=value2` (`\n` meaning "however JS "sees" a line break). (I am bad at reading regex.)

AndreasHae:
Oh, thanks @bhousel! Didn't see them.. What's weird though is that the build passed just fine for me locally.

AndreasHae:
@tordans yes, that should work!

quincylvania:
> What's weird though is that the build passed just fine for me locally.

@AndreasHae You need to run `npm run lint` to see some syntax errors like missing semicolons that the build and mocha tests won't catch.

slhh:
@AndreasHae 
Assuming no spaces in keys seems to be ok, but in addition you are likely assuming no equal signs in values, which could for example fail in case of URLs, conditional restrictions (operators =, <=, >=), strange names. I think a space is insufficient to separate tags.

AndreasHae:
Good point @slhh.
Do you think I should limit the splitting only to newlines?

bhousel:
Thanks @AndreasHae  - this looks ok 👍

PR #6245, opened 2019-04-27
d3 v5

bhousel:
Upgrades iD to use d3 v5
The main change here is replacing the code that used d3-request to now use d3-fetch, and big updates to the test code to handle this change, such as adding a sinon.stub for the fetch API, and making most of those tests work async.


bhousel:
Good enough for now..

PR #6246, opened 2019-04-27
Add aerialway=j-bar to presets

westnordost:
I noticed that from the `aerialway` presets, `j-bar`, described in the wiki as a type of `aerialway=drag_lift` seems to be missing. T-bar already exists in the presets.

https://wiki.openstreetmap.org/wiki/Tag:aerialway%3Dj-bar


bhousel:
Thanks @westnordost 👍

Issue #3579, opened 2016-11-12
Enhancement: Preset for more leisures and aquariums.


sguinetti:
Similar to #3559 Add POIs of leisure places. The purpose is to make it more varied for family entertainment than mentioning "parks" or "gardens".

- leisure=resort (example: country clubs)
-- Info: https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dresort

- leisure=dance (example: salsoteca)
-- Info: https://wiki.openstreetmap.org/wiki/Tag:leisure%3Ddance
-- Proposed page: https://wiki.openstreetmap.org/wiki/Proposed_features/dance

- leisure=water_park
-- Info: https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dwater_park
-- Info 2: http://gestion.pe/tendencias/siete-parques-acuaticos-que-usted-conocer-2100377

- tourism = aquarium
-- Similar to zoos.
-- Info: https://wiki.openstreetmap.org/wiki/Tag:tourism%3Daquarium
-- Proposed page: https://wiki.openstreetmap.org/wiki/Proposed_features/Aquarium

Please and thanks.

bhousel:
I think we have `leisure=water_park` already, but all the other ones sound good.. Thanks for the suggestions!

PR #6251, opened 2019-04-28
Deprecate amenity=dancing_school

westnordost:
Also deprecated in the wiki and a replacement is available, see https://wiki.openstreetmap.org/wiki/Tag%3Aamenity%3Ddancing_school

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -29,6 +29,10 @@
       "replace": {"amenity": "community_centre"}
     },
     {
+      "old": {"amenity": "dancing_school"},
+      "replace": {"leisure": "dance", "dance:teaching": "yes" }
+    },
+    {
       "old": {"amenity": "dog_bin"},
       "replace": {"amenity": "waste_basket", "waste": "dog_excrement"}
     },


bhousel:
Thanks for all your contributions today @westnordost 👏

simonpoole:
A facility providing professional dancing instruction may be co-located with a dancing hall, but just as well could be completely separate depending on the audience and scope (for example providing vocational education). 

Just because somebody engaged in wiki-fiddling is no reason to jump to conclusions.

westnordost:
Okay, so the concept is that `leisure=dance` is a "(general purpose) dancing hall" and `amenity=dancing_school` is, well, a dancing school? And the latter can be within the former?

Then let's revert this and make the dancing school a preset and somebody should clear up the wiki.

bhousel:
> Then let's revert this and make the dancing school a preset and 

Dance Hall [has been a preset since 2016](https://github.com/openstreetmap/iD/blob/master/data/presets/presets/leisure/dance.json), see #3579 and I followed the recommendation on the wiki at the time.

Dance school [has been a preset since 2018](https://github.com/openstreetmap/iD/blob/master/data/presets/presets/leisure/dancing_school.json), see #4846.  The contributor followed the recommendation on the wiki at the time.

I agree that the wiki could use some cleanup, but we don't need to revert anything.  We are following the guidelines the best we can, and we don't need multiple presets for Dance Halls - it's totally fine to deprecate the `amenity` one.

westnordost:
I think this is about the name. Sure, you may get tuition at a dancing venue, but this doesn't make the dancing venue a dancing school.

For example, a car dealership `shop=car` may also offer repair services (`service:vehicle:car_repair=yes` I supposed), but that doesn't make the car dealership an automobile repair shop `shop=car_repair`. These two are usually quite some different things, at least are referred to differently (as I just did).

Whether this is really about that, I am not sure, because the wiki is not really clear here. But at least it would make sense this way and would be consistent.

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/issues/373 proposes revert of that PR

matkoniecz:
https://github.com/openstreetmap/id-tagging-schema/issues/1177#issuecomment-2040103619 actually reverts it, I made broken PR

PR #6252, opened 2019-04-28
Deprecate amenity/leisure=club/social_club

westnordost:
See https://wiki.openstreetmap.org/wiki/Key%3Aclub , club is documented as a top-level tag

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -25,6 +25,10 @@
       "replace": {"tourism": "information"}
     },
     {
+      "old": {"amenity": "club"},
+      "replace": {"club": "*"}
+    },
+    {
       "old": {"amenity": "community_center"},
       "replace": {"amenity": "community_centre"}
     },
@@ -107,6 +111,10 @@
     {
       "old": {"amenity": "sloped_curb"},
       "replace": {"kerb": "lowered"}
+    },
+    {
+      "old": {"amenity": "social_club"},
+      "replace": {"club": "*"}
     },
     {
       "old": {"amenity": "swimming_pool"},
@@ -417,6 +425,10 @@
     },
     {
       "old": {"leisure": "club"},
+      "replace": {"club": "*"}
+    },
+    {
+      "old": {"leisure": "social_club"},
       "replace": {"club": "*"}
     },
     {


trigpoint:
This change is begining to cause issues and loose valuable information in the UK.

Amenity=social_club is in fact a type of pub or bar, which is operated by its members, the wiki does seem to have misunderstood the term, possibly based on wikipedia.

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dsocial_club

Thanks Phil

quincylvania:
@trigpoint Hmm I can remove the deprecations, but I'd suggest working to change the tagging to something like `amenity=pub` + `club=social`. It's difficult for data consumers to support low-usage tags that are specific to one country—I wouldn't be surprised if many POI maps incidentally don't show these.

westnordost:
Hi @trigpoint, sorry, I didn't know that a social club in Britain carries a special meaning and is not simply a synonym for club when I created this PR.
It's some time since I created the PR so I don't remember exactly the source or motivation of it, but usually all PRs I made to the iD presets were based on some description found in the wiki.

I also just realized that documentation about this feature was just added a few days ago by @SK53.

The important point is how is it guaranteed that all the places that are currently tagged as amenity=social_club actually fall under that definition and were not just synonyms of amenity=club as I assumed it to be?

Or in other words: Does the recently added documentation reflect the actual use of the tag?

As said, my assumption back in April was *probably* based on some information found in the wiki at that time in one way or another, so it must be assumed that other mappers may have assumed the same.

To summarize:
amenity=social_club is not a good tag because it is very specific to one country and can be misinterpreted as something else by non-British mappers without clear documentation. Something so close to a pub should not have an own top level tag. 

Though, as it turns out now that it is something very different from an amenity=club, the deprecation rule is wrong and must be removed.

If however there is evidence that the tag is ambiguos in that other clubs have been tagged as social_club than British social clubs, a deprecation of the tag should be discussed in favor of a tagging that clearly separates clubs from British social clubs. Maybe there is already one.

I'm on the road for the rest of the week, so I can't do the research right now, maybe someone else? Interesting questions are:
- Are all amenity=social_club the British social clubs? How much % are probably mistaggings?
- Is amenity=social_club the only way how British social clubs are tagged currently? What other methods are there, if any, and how do they compare in usage?

trigpoint:
Thank you Quincy and Tobias
I have looked at a sample of the usage, England and Wales only so far and the names all match things I would expect to see, xxx Working Mens Club, yyy Sports and Social Club, zzz Royal British Legion. 
I will do a more thorough check and report back.

quincylvania:
@trigpoint It'd also be helpful if you or someone could determine whether `leisure=social_club` is used a synonym. Right now it has approximately the same usage as `amenity=social_club` but without documentation. If it's the same then I can change iD's upgrade path.

quincylvania:
Also: `social_facility=social_club`

westnordost:
I added my research and opinion at the discussion page to the article. 
https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dsocial_club#Current_usage_.26_Comparison_to_other_methods_to_tag_the_same_thing

If anything comes out of it, I'll open another PR.

simonpoole:
It is fundamental to OSM that it is possible to invent and use tags that only have regional significance, if the facility/object fits in to a larger schema then obviously it is preferable to use that (which however in this case is not the case), but that is the limit to which we should try to normalize the world. 

BTW I can easily come up with a dozen plus tags that are Germany specific,, not to mention things that only relevant in the US of A. Lets start deleting those.

SK53:
A few quick points somewhat disjointed I'm afriad . The tag has been in use for at least 9 years, but prior to that people may have used amenity=pub or possibly leisure=social_club. The decision to introduce this tag was not taken lightly, and on the basis of detailed local knowledge and considerations. Note that we have at least one equivalent tag for places to drink which are similarly distinct and valued in a particular culture (amenity=biergarten).

The tag came into use because, although, in may ways similar to pubs, they are not pubs (notionally, at least, serving drinks requires that 'customers' be members of 2 days standing or guests of members), and have a different [licence regime](https://www.gov.uk/club-premises-certificate). 

Knowing which ones can be treated like a pub (i.e,, wandering in and ordering a drink) and which are pretty much closed to any but the relevant social group usually requires precise local knowledge which is beyond the scope of OSM. 

Such places are tagged in a large number of ways: amenity=pub, amenity=bar, amenity=social_club, leisure=social_club, club=social etc. Many pub & bar tags predate any use of the other tags.  A quick scan through leisure=social_club reveals that the vast majority indeed have names which are typical of this type of social club (British Legion, Working Man, Sports and Social, Conservative Club, Labour Club, Liberal Club, Comrades Club, Ex-Serviceman's Club etc). There are one or two Rugby clubs which better fit in the club namespace, a few associated with a company (BBC Club), and one or two based on ethnicity (which may fit, or may be better with club or amenity=community_centre). For instance in Nottingham the Irish & Polish Clubs are definitely social_clubs, whereas others are not). I would estimate 80-90% represent synonyms.

quincylvania:
So I'm not going to spend the time or energy to debate tags with low usage like this. I removed the `leisure=social_club` deprecation as well so now iD doesn't have an opinion on any `social_club` tags.

In general I'm fine with supporting tags for region-specific features as long as they're defined and documented clearly enough that mappers and data consumers can easily understand or ignore them.

SK53:
That's fine, but I find the tone a little derogatory. 

I've had to invest considerable time to justify a tag I've been using for 10 years or more, and I don't get paid for it. I would ask you to consider using a more respectful tone. I do appreciate that you get a lot of things coming in and clearly it's sensible to do triage & devote your time to particular tasks, but don't write things which sort of sound as if the time of others is any less valuable than yours.

I think this can be closed, but two more general points to consider:

The NSI is surfacing undocumented tags: that's great it's providing an incentive to document them which is also valuable. On the other hand moving too quickly may cause useful tagging to be lost. If there was a way to highlight tags which are being caught before each version goes live, we can get the best of both worlds: better documentation, better guidance of mappers to useful tags.

Regionally significant tags will become commoner, as we get more people mapping in non-Western countries, where existing tags may not support important cultural features (e.g., [circumcision huts in Lesotho](https://wiki.openstreetmap.org/w/index.php?title=WikiProject_Lesotho&oldid=1551455#Special_Lesotho_Content) or [cultivated land types](https://wiki.openstreetmap.org/wiki/Key:farming_system) in Nepal). It's worth considering how wemight want to handle this in the future.

simonpoole:
> 
> 
> So I'm not going to spend the time or energy to debate tags with low usage like this.

I think there is a bit of a misunderstanding here. The whole point of the discussion is that iD should simply do nothing when presented with an unknown (for many values of unknown) tag. 

Actual deprecation (as in there is a generally accepted more popular alternative) of tags in OSM is very very rare, and should be investigated carefully before being implemented in any editor.

I don't have current numbers, but all of the editors presets hover around having presets for perhaps 10-20% of tagging for objects that are not just one-offs (it isn't just a question of numbers, some things are simply fairly rare irl, say opera houses). This is probably slightly less visible in iD because of taginfo integration, but in the grand scheme of things that doesn't make a big difference. In any case there should be no expectation that there are presets for "everything" and that something for which there is no documentation or relatively low use, is unimportant.

quincylvania:
> That's fine, but I find the tone a little derogatory.

@SK53 Oh no, I'm sorry my post came across that way! I admit re-reading my words they sound dismissive. This was not my intention. I should have been more thoughtful and I apologize.

westnordost:
@simonpoole This discussion belongs into the wiki.

PR #6253, opened 2019-04-28
Add shop=general for general store

westnordost:
The tag is used 2800+ times and following a recent discussion on the tagging mailing list, the tag is pretty well documented now and also what is the difference to `shop=convenience`, `shop=variety_store`, `shop=department_store`, `shop=hardware` and `shop=doityourself` and so on.

See https://wiki.openstreetmap.org/wiki/Tag:shop%3Dgeneral

Also, there is a good wikipedia article about it: https://en.wikipedia.org/wiki/General_store

### DIFF
--- a/data/presets/presets/shop/general.json
+++ b/data/presets/presets/shop/general.json
@@ -1,0 +1,12 @@
+{
+    "icon": "maki-shop",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "shop": "general"
+    },
+    "terms": ["village shop"],
+    "name": "General store"
+}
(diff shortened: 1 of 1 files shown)


data/presets/presets/shop/general.json
```diff
@@ -0,0 +1,12 @@
...
+    "icon": "maki-shop",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "shop": "general"
+    },
+    "terms": ["village shop"],
+    "name": "General store"
```
bhousel: capitalize like "General Store"

bhousel: Thanks.. The terms don't have to be capitalized, they are for searching, not display.

PR #6254, opened 2019-04-28
Add preset for building=hangar

westnordost:
Also, make the hangar building preset have a lower matchScore so that first the` aeroway=hangar` is found. Add `building=hangar` to the aeroway preset.

See https://wiki.openstreetmap.org/wiki/Tag:building%3Dhangar

### DIFF
--- a/data/presets/presets/building/hangar.json
+++ b/data/presets/presets/building/hangar.json
@@ -1,0 +1,13 @@
+{
+    "fields": [
+        "name"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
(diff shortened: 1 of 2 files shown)


data/presets/presets/building/hangar.json
```diff
@@ -0,0 +1,13 @@
...
+        "name"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "building": "hangar"
+    },
+    "matchScore": 0.5,
+    "name": "Hangar building"
```
bhousel: "Hangar Building"

PR #6257, opened 2019-04-28
Add castle tower preset

westnordost:
See https://wiki.openstreetmap.org/wiki/Tag%3Atower%3Atype%3Ddefensive , used 3800+ times

### DIFF
--- a/data/presets/presets/man_made/tower/defensive.json
+++ b/data/presets/presets/man_made/tower/defensive.json
@@ -1,0 +1,20 @@
+{
+    "icon": "maki-castle",
+    "geometry": [
+        "point",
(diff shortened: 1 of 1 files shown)


data/presets/presets/man_made/tower/defensive.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "reference": {
+        "key": "tower:type",
+        "value": "defensive"
+    },
+    "terms": [
+        "Defensive tower",
+        "Castle tower"
+    ],
+    "name": "Fortified tower"
```
bhousel: "Fortified Tower"

Issue #6007, opened 2019-03-03
christian church versus church building


eehpcm:
There are times when I map a disused church, and use the "Church building" preset which gives building=church.  I'm happy with that, it's the way I'd expect it to work.  The "Chapel building" preset gives similar results with building=chapel.

There are other times when I map an active church - it's still a place of worship and it's in a building that is recognisable as a church (the sort of building that, if you saw it without knowing what it was, you'd think "church").  But this gives building=yes, which I have to manually change to building=church.

The same thing applies to chapel, except there's no preset (that is obvious to me) for a chapel that is an active place of worship, just a preset for a building that looks like a chapel but no longer is used as such.

Yes, there are Christian places of worship in buildings that are not recognizable as churches or chapels.  One chapel near me (which looks like a chapel) is thinking of relocating to a larger, industrial building in a business park.  One church near me was purpose-built  as a church but is not recognizable as such (it looks like somebody on drugs was given a Lego set of mismatched parts, and would probably be in the top-10 list of "Ugliest churches in Britain").  But **most** active churches I map are building=church, not building=yes and **most** active chapels I map are building=chapel, not building=yes.

Any chance of optimizing the presets for the common case?

bhousel:
It's a bit complicated, here's why:
The `building=yes` tag gets added because of the [`building_area`](https://github.com/openstreetmap/iD/blob/master/data/presets/fields/building_area.json) field.  This is a pretty common field that is used on a lot of presets that should probably get a `building` tag, but only when they are mapped as areas.  It doesn't do anything if the feature is mapped as a standalone node.

So for us to add a `building=church` tag, we'd need to make a different kind of `building_area` field just for churches.  And then people would want different building fields for everything.

If we instead added `building=church` directly to the Church preset `addTags` section, it would add that tag even if the Church were mapped as a node, and this not really correct.  (People don't like to see `building` tags on nodes).

I'm not sure what the best solution is.  A while ago I suggested creating a special kind of field for building details that could handle all of this stuff (also #1829, https://github.com/openstreetmap/iD/issues/2550#issuecomment-78273632).

eehpcm:
@bhousel 

>  I'm not sure what the best solution is

Me neither.

It's not a major problem, but if you ever come up with something I'd appreciate it.

matkoniecz:
> Any chance of optimizing the presets for the common case?

Note that as result of this uncommon cases (place of worship in non-church building) will start to be tagged poorly.

[...]

PR #6259, opened 2019-04-28
Add presets for bicycle parkings...

westnordost:
... that are not essentially just a type of bicycle stands but a different thing (by name). See https://wiki.openstreetmap.org/wiki/Key:bicycle_parking

### DIFF
--- a/data/presets/presets/amenity/bicycle_parking/building.json
+++ b/data/presets/presets/amenity/bicycle_parking/building.json
@@ -17,7 +17,9 @@
         "key": "bicycle_parking"
     },
     "terms": [
-        "Multi-storey bicycle park"
+        "Multi-Storey Bicycle Park",
+        "Multi-Storey Bike Park",
+        "Bike Parking Station"
     ],
-    "name": "Bicycle parking station"
+    "name": "Bicycle Parking Station"
 }
--- a/data/presets/presets/amenity/bicycle_parking/building.json
+++ b/data/presets/presets/amenity/bicycle_parking/building.json
@@ -1,0 +1,23 @@
+{
+    "icon": "maki-bicycle",
+    "fields": [
+        "{amenity/bicycle_parking}",
+        "opening_hours"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "amenity": "bicycle_parking",
+        "bicycle_parking": "building"
+    },
+    "reference": {
+        "key": "bicycle_parking"
+    },
+    "terms": [
+        "Multi-storey bicycle park"
+    ],
+    "name": "Bicycle parking station"
+}
--- a/data/presets/presets/amenity/bicycle_parking/shed.json
+++ b/data/presets/presets/amenity/bicycle_parking/shed.json
@@ -2,7 +2,7 @@
     "icon": "maki-bicycle",
     "fields": [
         "{amenity/bicycle_parking}",
-        "building"
+        "building_area"
     ],
     "geometry": [
         "point",
@@ -10,16 +10,6 @@
         "area"
     ],
     "tags": {
-        "amenity": "bicycle_parking",
-        "bicycle_parking": "shed"
-    },
-    "addTags": {
-        "building": "shed",
-        "amenity": "bicycle_parking",
-        "bicycle_parking": "shed"
-    },
-    "removeTags": {
-        "building": "shed",
         "amenity": "bicycle_parking",
         "bicycle_parking": "shed"
     },
--- a/data/presets/presets/amenity/bicycle_parking/shed.json
+++ b/data/presets/presets/amenity/bicycle_parking/shed.json
@@ -1,0 +1,33 @@
+{
+    "icon": "maki-bicycle",
+    "fields": [
+        "{amenity/bicycle_parking}",
+        "building"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "amenity": "bicycle_parking",
+        "bicycle_parking": "shed"
+    },
+    "addTags": {
+        "building": "shed",
+        "amenity": "bicycle_parking",
+        "bicycle_parking": "shed"
+    },
+    "removeTags": {
+        "building": "shed",
+        "amenity": "bicycle_parking",
+        "bicycle_parking": "shed"
+    },
+    "reference": {
+        "key": "bicycle_parking"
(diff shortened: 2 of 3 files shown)


tordans:
I wonder, if this also changes the meaning of the regular "Bicycle parking" preset to be meant for "on street without cover". If so, the cover=no could be the smart default there (to be changed by users if required).

westnordost:
No, because there are also normal stands  which simply have a roof/cover but are open to the sides. In my area, often found at train and metro stations.

There is a StreetComplete quest that asks for bicycle stands if they are covered or not.

So, it is definitely a possible field for bicycle stands, but not automatically "no".

Am 29. April 2019 07:34:32 MESZ schrieb Tobias <notifications@github.com>:
>I wonder, if this also changes the meaning of the regular "Bicycle
>parking" preset to be meant for "on street without cover". If so, the
>cover=no could be the smart default there (to be changed by users if
>required).

data/presets/presets/amenity/bicycle_parking/building.json
```diff
@@ -0,0 +1,25 @@
...
+        "opening_hours"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "amenity": "bicycle_parking",
+        "bicycle_parking": "building"
```
bhousel: hmm this one is weird..
Is it really `bicycle_parking=building` and not `building=bicycle_parking`?
Can we change that before it gains traction?

data/presets/presets/amenity/bicycle_parking/shed.json
```diff
@@ -0,0 +1,33 @@
...
+        "point",
+        "vertex",
+        "area"
+    ],
+    "tags": {
+        "amenity": "bicycle_parking",
+        "bicycle_parking": "shed"
+    },
+    "addTags": {
+        "building": "shed",
```
bhousel: So this is a problem because we don't want to add `building=shed` if the user maps it as a point.  This is why for presets that can be either points or areas, we include the `building_area` field.  It will set `building=yes` only if the thing is mapped as an area.

data/presets/presets/amenity/bicycle_parking/building.json
westnordost: These are two different things:

`bicycle_parking=building`  defines the type of bicycle parking: That it is in an enclosed building.

`building=bicycle_parking` is not defined in the wiki, but I guess it would be a building that was built for the purpose of serving as a bicycle parking station.

The difference is, that the bicycle parking doesn't have to fill the whole building but may be some area or point within the building. So, much like `parking=multi-storey` vs `building=parking`.

The `bicycle_parking=building` is already used by StreetComplete by a year or so. And much longer documented like that.

data/presets/presets/amenity/bicycle_parking/shed.json
westnordost: Can you give me a hint what I need to change to make this preset add `building=shed + bicycle_parking=shed` if it is and area and only `bicycle_parking=shed` if it is a point?

bhousel: We can't change it right now (this is also #6007) .  So I think best to just leave `building` out of addTags and removeTags, and instead add a `building_area` field (which will set it to `building=yes` if it is an area).  This is the best we can do for right now and still not put `building` tags on points.

westnordost: Okay, done

data/presets/presets/amenity/bicycle_parking/building.json
quincylvania: I agree with Bryan, this preset is confusing. I'd prefer to just use `indoor=yes` and reserve the `bicycle_parking` tag for the actual method, like stands, lockers, or whatnot.

westnordost: `indoor` is for indoor-mapping only but `bicycle_parking=building` may and is often tagged on the whole building (but doesn't have to, as mentioned).

westnordost: On the wiki page, someone else raised a similar concern as Bryan in 2017: https://wiki.openstreetmap.org/wiki/Talk:Key:bicycle_parking#Not_sure_if_bicycle_parking.3Dbuilding_makes_sense

westnordost: To summarize, tagging [bicycle parking stations](https://en.wikipedia.org/wiki/Bicycle_parking_station) as `bicycle_parking=building` is missing an alternative because it could be tagged as an indoor-node  but also the building itself. I see three options, please choose:
1. drop the  bicycle parking station preset from this PR
2. limit the bicycle parking station preset to only areas and set `building=bicycle_parking` instead
3. use `bicycle_parking=building` after all, leave PR as-is

quincylvania: Hmm I was assuming this preset was for bike parking inside buildings with other purposes, but I guess [bike parking garages are a thing](https://www.psfk.com/2017/08/new-bicycle-garage-netherlands-largest-in-the-world.html)? If that's what this tag is actually for, then I'm fine with option 2 as long as we rename the preset to "Bicycle Parking Garage" or similar.

westnordost: Yes, the Netherlands are known worldwide for their advanced bicycle infrastructure. They call these apparently mostly "[fietsenstalling stationsplein](https://www.bezoek-utrecht.nl/wat-te-doen/fietsen/fietsenstallingen/3440301639/fietsenstalling-stationsplein)" (literally: bicycle storage station square).

[Here is a Google StreetView of such a location](https://www.google.com/maps/@52.0929459,5.1137695,3a,55.7y,143.28h,84.77t/data=!3m6!1e1!3m4!1scRA6JV6ql6h0389fOHS7zg!2e0!7i13312!8i6656) (note that bicycle "street" ). [The same location in OpenStreetMap](https://www.openstreetmap.org/node/2466665935) is already tagged.
Though, note that with option 2, the preset would not work for cases like these because it is mapped as a node, much like a `amenity=parking_entrance`.

I will change the PR to option 2 later this evening (UTC+2) or otherwise on Saturday.

westnordost:
Wait, I did not change anything yet!

quincylvania:
@westnordost I know! I figured I'd just take care of it so the translators can get started.

westnordost:
Ok cool

PR #6260, opened 2019-04-28
Added laundromat preset and self_service field

westnordost:
Added the laundromat preset because it often has quite a different name in languages (including English) than a generic (staff-operated) laundry. See https://en.wikipedia.org/wiki/Self-service_laundry

Also, added the `self_service` field https://wiki.openstreetmap.org/wiki/Key%3Aself_service , used 7k+ times and added this field to the presets that often have self-service: fuel stations, car washes and laundries. Not sure if it should be in `fields` or `moreFields`.

### DIFF
--- a/data/presets/presets/shop/laundry/self_service_laundry.json
+++ b/data/presets/presets/shop/laundry/self_service_laundry.json
@@ -1,0 +1,19 @@
+{
+    "icon": "maki-laundry",
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "tags": {
+        "shop": "laundry",
+        "self_service": "yes"
+    },
+    "terms": [
+        "coin laundry",
+        "laundromat",
+        "coin wash",
+        "launderette",
+        "washateria"
+    ],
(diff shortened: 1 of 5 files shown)


westnordost:
Since case doesn't matter for terms, I thought I provide them already in the form they should be when aliases are introduced (cause these terms are actually aliases) :-)

data/presets/presets/shop/laundry/self_service_laundry.json
```diff
@@ -0,0 +1,19 @@
...
+        "self_service": "yes"
+    },
+    "terms": [
+        "Coin Laundry",
+        "Laundromat",
+        "Coin Wash",
+        "Launderette",
+        "Washateria"
+    ],
+    "name": "Self-service Laundry"
```
quincylvania: Oh could you also capitalize "Service" here? Thanks!

quincylvania: I took care of it so no worries. Thanks for this PR!

PR #6263, opened 2019-04-29
Revert "Deprecate amenity=dancing_school"

westnordost:
Reverts openstreetmap/iD#6251

The wiki is/was misleading here. Apparently, the `amenity=dancing_school` is just the school, while `leisure=dance` is generally a venue where people can dance.


bhousel:
We already have a Dance School preset

Issue #6241, opened 2019-04-25
Flag duplicate or extremely close adjacent nodes in a way


quincylvania:
It's generally undesirable to have adjacent nodes in a way that are actually or nearly in the same place, but it can be difficult to find such instances visually. iD should flag these and offer to merge the points.

<img width="404" alt="Screen Shot 2019-04-24 at 5 10 01 PM" src="https://user-images.githubusercontent.com/2046746/56701458-2147a280-66b4-11e9-924b-310b1d025e57.png">

bhousel:
Haha strong agree.. these drive me nuts.

BjornRasmussen:
Same here

SilentSpike:
Semi-related to this, what about offering to simplify straight lines with multiple untagged/connected nodes along their length?

bhousel:
> Semi-related to this, what about offering to simplify straight lines with multiple untagged/connected nodes along their length?

We don't currently offer a great line simplification operation in iD.  This is being tracked on #3116.

bhousel:
done in #6267

PR #6267, opened 2019-04-30
add validation for very close node on road

gaoxm:
Addresses issue #6241

Also made a change in inspector, so that when an untagged entity has issues, entity editor is shown to make the issues visible. Otherwise it's a confusing experience for users when they click on an item on the issues pane but don't see the issue show up in the inspector.

Using a cheap check by delta on lat/lon degrees, since 1-e5 degree is ~1 meter. Let me know if the great-circle geo-distance is preferred.

Tested in browser and see the issue gets flagged; also verified the merge operation works.

Screen Shot 2019-04-29 at 10 54 16 PM

cc @wonga00 @chrisklaiber

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1403,6 +1403,11 @@ en:
     unknown_road:
       message: "{feature} has no classification"
       reference: "Roads without a specific type may not appear in maps or routing."
+    dupe_node_on_road:
+      title: Very close nodes on road
+      message: "Very close nodes on road"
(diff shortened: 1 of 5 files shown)


bhousel:
Looks great, thanks @gaoxm 👍 

> Using a cheap check by delta on lat/lon degrees, since 1-e5 degree is ~1 meter. Let me know if the great-circle geo-distance is preferred.

Awesome, for such tiny distances there's no need to do great circle 😄

PR #6270, opened 2019-04-30
inherit preset.removeTags from preset.addTags if not defined [Ready for Review]

westnordost:
I am thinking of ways to make preset definitions shorter and easier to maintain.

My suggestion is to let `removeTags` default to `addTags` if it is not defined, just as they both currently default to `tags`.

I would be willing to work on it, namely adapt the tests and other code points if I missed them, and remove the `removeTags` fields in the existing presets where they have become unnecessary. But I want to check first if the PR would be merged before I really start working on it.

Less code in presets means less time to write new ones and maintain them, and also less room for mistakes when doing so.


quincylvania:
Seems fine to me.

data/presets/presets/amenity/parking/underground.json
```diff
@@ -18,11 +18,6 @@
         "parking": "underground",
         "layer": "-1"
     },
-    "removeTags": {
-        "amenity": "parking",
-        "parking": "underground",
-        "layer": "-1"
-    },
```
westnordost: Would make `"layer": "*"` make sense here for `"removeTags"`? Then it should stay.

build_data.js
```diff
@@ -263,7 +263,6 @@ function suggestionsToPresets(presets) {
             geometry: preset.geometry,
             tags: Object.assign({}, preset.tags, wikidataTag),
             addTags: suggestion.tags,
-            removeTags: suggestion.tags,
```
westnordost: This change will btw somewhat reduce the size of the `presets.json`

data/presets/presets/amenity/parking/underground.json
bhousel: Actually in `removeTags` the values don't mean anything, only the keys.

build_data.js
bhousel: Yeah, tbh for iD v3 we should find a way to bundle the suggestions separately anyway.  They are huge.

data/presets/presets/amenity/parking/underground.json
westnordost: Oh, then I can remove some more `"removeTags"`, except if you think they should be kept for better overview?

westnordost: Well, I think I will leave it because maybe in the future iD wants to only remove a tag if the value hasn't been changed to something else in the meantime. Removing it feels like a loss of information of the preset.

bhousel:
This looks ok, thank you!

PR #6274, opened 2019-04-30
Add more artwork presets

westnordost:
Added the next three most used (according to taginfo) artwork types: Graffiti, Installation and Bust.

I think it is important to
* have bust if statue exists as a preset
* have graffiti if mural exists as a preset
* and have installation if sculpture exists as a preset

to avoid mistaggings, e.g. people searching for "graffiti wall" and finding only the "mural" preset would probably tag it (incorrectly) as a mural. Same with the others


westnordost:
Uh damn, it is based on that other PR of mine. I will rebase it.

PR #6275, opened 2019-04-30
Add bust, graffiti and installation artwork type preset

westnordost:
Added the next three most used (according to taginfo) artwork types: Graffiti, Installation and Bust.

I think it is important to
* have bust if statue exists as a preset
* have graffiti if mural exists as a preset
* and have installation if sculpture exists as a preset

to avoid mistaggings, e.g. people searching for "graffiti wall" and finding only the "mural" preset would probably tag it (incorrectly) as a mural. Same with the others

### DIFF
--- a/data/presets/presets/tourism/artwork/graffiti.json
+++ b/data/presets/presets/tourism/artwork/graffiti.json
@@ -18,9 +18,9 @@
         "key": "artwork_type"
     },
     "terms": [
-        "graffiti",
         "Street Artwork",
-        "Guerilla Artwork"
+        "Guerilla Artwork",
+        "Graffiti Artwork"
     ],
-    "name": "Graffiti Artwork"
+    "name": "Graffiti"
 }
--- a/data/presets/presets/tourism/artwork/graffiti.json
+++ b/data/presets/presets/tourism/artwork/graffiti.json
@@ -1,0 +1,26 @@
+{
+    "icon": "maki-art-gallery",
+    "fields": [
+        "name",
+        "artist"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "line",
+        "area"
+    ],
+    "tags": {
+        "tourism": "artwork",
+        "artwork_type": "graffiti"
+    },
+    "reference": {
+        "key": "artwork_type"
+    },
+    "terms": [
+        "graffiti",
+        "Street Artwork",
+        "Guerilla Artwork"
+    ],
+    "name": "Graffiti Artwork"
+}
--- a/data/presets/presets/tourism/artwork/installation.json
+++ b/data/presets/presets/tourism/artwork/installation.json
@@ -1,0 +1,26 @@
+{
+    "icon": "maki-art-gallery",
+    "fields": [
+        "name",
+        "artist"
+    ],
+    "geometry": [
+        "point",
+        "vertex",
+        "line",
+        "area"
+    ],
+    "tags": {
+        "tourism": "artwork",
+        "artwork_type": "installation"
+    },
+    "reference": {
+        "key": "artwork_type"
+    },
+    "terms": [
+        "interactive art",
+        "intervention art",
+        "modern art"
+    ],
+    "name": "Installation Artwork"
+}
--- a/data/presets/presets/tourism/artwork/installation.json
+++ b/data/presets/presets/tourism/artwork/installation.json
@@ -22,5 +22,5 @@
         "intervention art",
         "modern art"
     ],
-    "name": "Installation Artwork"
+    "name": "Art Installation"
 }
--- a/data/presets/presets/tourism/artwork/bust.json
+++ b/data/presets/presets/tourism/artwork/bust.json
@@ -1,0 +1,22 @@
+{
+    "icon": "maki-art-gallery",
+    "fields": [
+        "name",
+        "artist",
+        "material"
+    ],
+    "geometry": [
+        "point", "vertex"
+    ],
+    "tags": {
+        "tourism": "artwork",
+        "artwork_type": "bust"
+    },
+    "reference": {
+        "key": "artwork_type"
+    },
+    "terms": [
+        "figure"
+    ],
+    "name": "Bust"
+}


data/presets/presets/tourism/artwork/graffiti.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    "reference": {
+        "key": "artwork_type"
+    },
+    "terms": [
+        "graffiti",
+        "Street Artwork",
+        "Guerilla Artwork"
+    ],
+    "name": "Graffiti Artwork"
```
quincylvania: Can we just call it Graffiti?

data/presets/presets/tourism/artwork/installation.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    "reference": {
+        "key": "artwork_type"
+    },
+    "terms": [
+        "interactive art",
+        "intervention art",
+        "modern art"
+    ],
+    "name": "Installation Artwork"
```
quincylvania: I think this one is confusing and I'm not sure we should add it. It looks like installations are mostly indoors and often overlap with `sculpture`. If we do add it then "Art Installation" would be my preferred name for it.

data/presets/presets/tourism/artwork/graffiti.json
westnordost: Maybe... the problem is, that not all graffiti is art. Actually, most of graffiti is not art, but just juvenile scribbling. This wouldn't and shouldn't be tagged with artwork. Perhaps if it is called just "Graffiti", people might get the wrong idea about what graffiti should be mapped.

An uncertainty that could certainly be solved with your subtitles.

quincylvania: I'm actually fine with any graffiti being tagged this way. If we specify this as only for "art" graffiti then we'll have people arguing over what's art and what's not art. A statue is a statue based on its form, not its quality. I think we can treat this the same.

data/presets/presets/tourism/artwork/installation.json
westnordost: Sounds right. I am not a native speaker.

It's true, the overlap with sculpture is there, but that is exactly why I think it is important to add it. However, it is just an overlap. A sculpture is just a sculpture, while an installation art can be so much more, is is much more free in what it can be, might be interactive, might be lights, something with music, AR/VR (<- well, this won't appear on the OSM of course), some "machine",... etc. 

Here are some cool examples by the way, enjoy :-)
- [blinkenlights](https://www.youtube.com/watch?v=PZRUtKYCpms)
- [Wasserlichtspiele Planten un Blomen in Hamburg](https://www.youtube.com/watch?v=qi9SXH3Fr8c)
- [Wintergatan Marble Machine](https://www.youtube.com/watch?v=IvUU8joBb1Q)

quincylvania: Okay, sounds fine with the new name.

data/presets/presets/tourism/artwork/graffiti.json
westnordost: What is art and what is not art will always be subjective. But this is nothing iD contributors need to deal with.

Fact is, the tag is `tourism=artwork`. As iD does not expose the tag names (immediately) to contributors, this information - that it is about art - would be lost if it is not mentioned somehow in the name.

[This, for example, is (also) graffiti.](https://proxy.duckduckgo.com/iu/?u=http%3A%2F%2F1.bp.blogspot.com%2F_VezUbgtAF0c%2FTAjzJ7hbMzI%2FAAAAAAAACzg%2FBhhc0mcPss4%2Fs1600%2Fgraffiti_tags.jpg&f=1) The most common form I see around my area, actually.

An (art)work of graffiti is something very different from that.

quincylvania: > An (art)work of graffiti is something very different from that.

I guess we just disagree on this point. To me graffiti is graffiti. This seems like a silly thing to argue about so I suggest we just hold off on adding this preset until we add the subtitle property.

SilentSpike: I may be totally wrong on this, but I believe there's a weird ambiguous legal distinction (in many places) between what is called "street art" and what is called "graffiti". As far as I can tell, they're the same thing except the former is deemed to have some worth. Quite an interesting topic to read up on.

westnordost:
> Can you include the derived changes from `npm run build` with your preset PRs? Thanks!

Yes I could (need to install node first but no problem), but do you really want this?

The generated `preset.json` is something that will very often result in conflict. If I had run `npm run build` for every PR adding or modifying presets last Sunday, there would have been a merge conflict to solve for every single PR after each one of them was merged.

quincylvania:
@westnordost We usually advise contributors to include the derived data so no followup work is needed, but it's not a big deal either way.

bhousel:
> > Can you include the derived changes from `npm run build` with your preset PRs? Thanks!
> 
> Yes I could (need to install node first but no problem), but do you really want this?

It's also important because the generated files are used by Transifex so that our translation teams can get started translating the preset names and terms.

(But like @quincylvania said - not a big deal either way.. We're happy to run the build for people who don't do it.)

data/presets/presets/tourism/artwork/graffiti.json
tordans: Dont we have the same issue with other presets and tags that have subjective importance? Like trees. First only very important trees where added. Now there are a lot of them on roads and such. And new use cases and data explorations come out of them. Some renderer still complain that people add too many trees which they need to filter; some projects are happy to calculate the green-ness of a hood. IMO mapper will use their own judgement here just fine; it happened before :).

Naming: I suggest keeping the actual name short but checking the info-(i) description and maybe make sure that it hints at artworks, not at scribbles/tagging (which will be pulled from Wikibase ATM, right?).

westnordost:
Ok then I can maybe do it after I am done with a "batch" of preset PRs :-)

data/presets/presets/tourism/artwork/graffiti.json
westnordost: Maybe my comments made it appear I have a stronger opinion about it that I actually do. I just wanted to show that I gave thought to the name choice.
Just "Graffiti" is fine for me too, especially since it's still the artwork icon and the tags are as always still visible further below.

quincylvania: Thanks for the thoughtful discussion, everyone. Apologies if I came across as grumpy.

quincylvania:
Looks good, @westnordost 👍

Issue #5497, opened 2018-11-18
Add presets to basin tag


Adamant36:
It would be cool if there was presets on basins for basin=* tag and also the intermittent=* tag. As it is, only 20,401 out of 212,641 basins are tagged with which type they are and only 2,213 are tagged as intermittent. I imagine the numbers for both would increase a lot if there were presets for them. 

Thanks

bhousel:
Sure - looking here for reference:  https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dbasin

I have questions, and I'm mostly asking because I don't want people to complain later.

> I imagine the numbers for both would increase a lot if there were presets for them.

Is this a thing people really want?  Just "increasing numbers" isn't a good enough reason to provide fields for things in iD.  Maybe the reason the numbers are low is because the the tags are not interesting or useful.

Would people really be able to tag whether it's an `infiltration`/`detention`/`retention` basin?

Aren't all basins really `intermittent=yes`?  They are for when it rains, right?

Trying to see things from the perspective of a casual contributor - I just don't see the point.  If people really want this, I'm ok with adding it, but I don't like to add fields just because someone wrote some things on the OSM wiki.   (This reminds me a lot of the fire hydrants in #5488)

Adamant36:
I can understand where your coming from with the retention type tags. The fire hydrant thing was a little ridiculous also. But I think its important to add intermittent tags to all water bodies in general if they are intermittent because they display differently on the based on it and intermittent is just a useful thing to know overall. More so with basins though because looking at it a from a higher zoom level it might not be obvious what's a basin and what's just a pond. The intermittent tag helps with that to some degree. Id assume there are some basins that aren't intermittent to, like nuclear waste/industrial/agriculture/water works retention basins that are full pretty much all the time. Although they don't seem to be covered in the wiki for basin types for some reason, but people do tag those types of things as basins "shrug." So, leave out the type. I'm fine with that. Intermittent should at least be in there. Or maybe you could make the same argument for removing it from the ponds preset.

bhousel:
I'm ok with adding both `basin` and `intermittent` tags..  Again, just really wanted to discuss it first.

- If somebody is using a tag and they really care about it, then we should try to add it.  (`intermittent` definitely meets this category, and `basin` seems pretty harmless)
[...]

PR #6278, opened 2019-05-01
add presets for types of wetland [Require Feedback]

westnordost:
Added the different types of wetland as presets. Also, corrected the geometry - should only be used on areas.

The `natural=wetland` preset contains the fields `salt` and `tidal`. I am not sure how to proceed with the sub-presets because some wetland types are `salt=yes` or `no` by definition, same with tidal.

**Should for example `salt=yes` automatically be added for `saltmarsh` and `mangrove` or the field be left away in cases where it is clear from the tag?**

Overview:
<table>
<tr><th></th><th>salt</th><th>tidal</th><th>intermittent</th></tr>
<tr><th colspan="4">Woody</th></tr>
<tr><td>swamp</td><td>well, there is <tt>mangrove</tt></td><td></td><td> </td></tr>
<tr><td>mangrove</td><td>always</td><td>always</td><td> </td></tr>
<tr><th colspan="4">Boggy</th></tr>
<tr><td>bog</td><td>never</td><td></td><td> </td></tr>
<tr><td>fen</td><td>never</td><td></td><td> </td></tr>
<tr><th colspan="4">Grassy</th></tr>
<tr><td>marsh</td><td>well, there is <tt>saltmarsh</tt></td><td> </td><td> </td></tr>
<tr><td>wet_meadow</td><td>well, there is <tt>saltmarsh</tt></td><td> </td><td>always</td></tr>
<tr><td>saltmarsh</td><td>always</td><td>always</td><td> </td></tr>
<tr><td>reedbed</td><td>well, there is <tt>saltmarsh</tt></td><td> </td><td> </td></tr>
<tr><th colspan="4">Muddy</th></tr>
<tr><td>tidalflat</td><td></td><td>always</td><td> </td></tr>
</table>

### DIFF
--- a/data/presets/presets/natural/wetland.json
+++ b/data/presets/presets/natural/wetland.json
@@ -5,27 +5,15 @@
         "salt",
         "tidal"
     ],
-    "geometry": [
-        "point",
-        "area"
-    ],
+    "geometry": [ "area" ],
     "tags": {
         "natural": "wetland"
     },
     "terms": [
-        "bog",
-        "fen",
-        "marsh",
         "mire",
         "moor",
         "muskeg",
-        "peatland",
-        "quagmire",
-        "reedbed",
-        "saltmarsh",
-        "swamp",
-        "tidalflat",
-        "wet meadow"
+        "quagmire"
     ],
     "name": "Wetland"
 }
--- a/data/presets/presets/natural/wetland/bog.json
+++ b/data/presets/presets/natural/wetland/bog.json
@@ -1,0 +1,20 @@
+{
+    "icon": "maki-wetland",
+    "geometry": [ "area" ],
+    "tags": {
+        "natural": "wetland",
+        "wetland": "bog"
+    },
+    "reference": {
+        "key": "wetland",
+        "value": "bog"
+    },
+    "terms": [
+        "Bogland",
+        "Peatland",
+        "peat",
+        "acidic",
+        "rain-fed"
+    ],
+    "name": "Bog"
+}
--- a/data/presets/presets/natural/wetland/fen.json
+++ b/data/presets/presets/natural/wetland/fen.json
@@ -1,0 +1,18 @@
+{
+    "icon": "maki-wetland",
+    "geometry": [ "area" ],
+    "tags": {
+        "natural": "wetland",
+        "wetland": "fen"
+    },
+    "reference": {
+        "key": "wetland",
+        "value": "fen"
+    },
+    "terms": [
+        "Peatland",
+        "peat",
+        "groundwater-fed"
+    ],
+    "name": "Fen"
+}
--- a/data/presets/presets/natural/wetland/mangrove.json
+++ b/data/presets/presets/natural/wetland/mangrove.json
@@ -1,0 +1,18 @@
+{
+    "icon": "maki-wetland",
+    "geometry": [ "area" ],
+    "tags": {
+        "natural": "wetland",
+        "wetland": "mangrove"
+    },
+    "reference": {
+        "key": "wetland",
+        "value": "mangrove"
+    },
+    "terms": [
+        "Mangrove Forest",
+        "Mangal",
+        "mangroves"
+    ],
+    "name": "Mangrove Swamp"
+}
--- a/data/presets/presets/natural/wetland/marsh.json
+++ b/data/presets/presets/natural/wetland/marsh.json
@@ -1,0 +1,17 @@
+{
+    "icon": "maki-wetland",
+    "geometry": [ "area" ],
+    "tags": {
+        "natural": "wetland",
+        "wetland": "marsh"
+    },
+    "reference": {
+        "key": "wetland",
+        "value": "marsh"
+    },
+    "terms": [
+        "grass",
+        "wet grass"
+    ],
+    "name": "Marsh"
+}
--- a/data/presets/presets/natural/wetland/reedbed.json
+++ b/data/presets/presets/natural/wetland/reedbed.json
@@ -1,0 +1,17 @@
+{
+    "icon": "maki-wetland",
+    "geometry": [ "area" ],
+    "tags": {
+        "natural": "wetland",
+        "wetland": "reedbed"
+    },
+    "reference": {
+        "key": "wetland",
+        "value": "reedbed"
+    },
+    "terms": [
+        "reeds",
+        "bulrushes"
+    ],
+    "name": "Reedbed"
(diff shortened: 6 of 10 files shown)


bhousel:
I'm kind of feeling like these are unnecessarily specific.  I'd prefer to not add these.

westnordost:
Do you mean `salt` or the wetland types in general?

If it is the latter, I can provide you with some information that may change your mind, as I did a little research on that today. Also, the [wetland key is used almost **a million times**](https://taginfo.openstreetmap.org/keys/wetland#values), so it is not exactly a fringe key :-).

To summarize my research, a swamp, a marsh, a mangrove and a tidalflat differ quite a lot from another. The rest is a little more difficult to tell apart and may require a bit of specialist knowledge, true.

bhousel:
> The rest is a little more difficult to tell apart and may require a bit of specialist knowledge (all of wet_meadow, saltmarshes, reedbeds are actually types of marshes).

Yes - I think it's this point, that it requires specialist knowledge to tag correctly, which makes me prefer to just offer people a generic "wetland" preset, and let mappers fill in the type if they know it.

I don't know if people could even translate stuff like "Fen" to all 80 languages.

We don't need a preset for every specific thing.

bhousel:
Worth reading:
https://en.wikipedia.org/wiki/Overchoice
https://en.wikipedia.org/wiki/The_Paradox_of_Choice
https://en.wikipedia.org/wiki/Analysis_paralysis

Basically, offering people too many choices between very similar things makes them unhappy.

quincylvania:
> I think it's this point, that it requires specialist knowledge to tag correctly, which makes me prefer to just offer people a generic "wetland" preset, and let mappers fill in the type if they know it.

I'm with Bryan on this. If a casual user sees a half-dozen alternatives for wetlands then they'll likely pick the wrong one and will probably feel unsure and therefore bad about their choice either way.

westnordost:
Wait, can we open this again? I had no time yet to answer today and I would like to make a case for at least some of the wetland presets.

Am 1. Mai 2019 18:16:09 MESZ schrieb Quincy Morgan <notifications@github.com>:
>> I think it's this point, that it requires specialist knowledge to tag
>correctly, which makes me prefer to just offer people a generic
>"wetland" preset, and let mappers fill in the type if they know it.
>
>I'm with Bryan on this. If a casual user sees a half-dozen alternatives
>for wetlands then they'll likely pick the wrong one and will probably
>feel unsure and therefore bad about their choice either way.

bhousel:
They can already pick it this way: 

Screenshot 2019-05-01 13 42 11

quincylvania:
@westnordost Okay, I suppose some of them like `tidalflat` might be fine.

westnordost:
This will probably be a three-part answer. First, I want to show you as promised that some of the different types of wetlands are indeed as different as forest, bushland, heath, meadow and beaches (only waterlogged). And thus, can be easily told apart:

<table>
<tr><th></th><th>criminally short description</th><th>photo</th></tr>
<tr><td>swamp</td><td>waterlogged <b>forest</b></td><td></td></tr>
<tr><td>marsh</td><td>waterlogged <b>grassland</b></td><td></td></tr>
<tr><td>tidal flat</td><td>mud beach</td><td></td></tr>
<tr><td>mangrove</td><td>tropical saltwater swamp</td><td></td></tr>
</table>

A bog (or moor) would be waterlogged **heath** but I did not find any photos that clearly separate it from a marsh. (Because grassland and heath have a similar appearance, I suppose).

quincylvania:
I'm okay with adding Tidal Flat and maybe Mangrove presets because I think those are pretty differentiated, at least in English. I'm still against the others; I couldn't have told you the difference between swamps, marshes, and wetlands.

It would be an easier argument to add these once we add the subtitle property.

westnordost:
Second, in response @bhousel's comment:

Perhaps I misunderstood how iD displays feature presets because I do not see where a user is confronted with choices:

**In the search**, if he inputs _"swamp"_ in the search field, he will get only the swamp preset and only that, not bog, not marsh etc. I did not find a view in iD that would display the different wetland feature presets in a list next to each other.

westnordost:
Thirdly, I understand that the primary problem @bhousel sees here is that in common language, "swamp", "marsh", "moor", "mire", "wetlands" and so on are used more or less interchangibly and/or the common folk don't know the difference, regardless of how big the actual difference actual is.

Okay, @quincylvania beat me to it suggesting the solution: short descriptions. Another already implementable would be to create a new icon for swamp being the wetland-forest.

Still, I want to ask if whether I with these three replies I convinced you to add swamp, marsh, tidal flat and mangrove and drop from this PR bog, fen, reedbed, saltmarsh and wet meadow?

quincylvania:
> In the search, if he inputs "swamp" in the search field, he will get only the swamp preset and only that, not bog, not marsh etc. I did not find a view in iD that would display the different wetland feature presets in a list next to each other.

@westnordost The problem isn't choice alone. Only showing a single choice is still a problem if it's not correct and the user doesn't know there are other options. Maybe it's a marsh but the user typed "swamp" not knowing the difference.

I'd be okay with Tidal Flat like I said before. Once we have subtitles we can reconsider the others but we're not going to add them now.

Is there a particular reason you want these presets? It seems like a lot of effort to just make things more complicated. Most data consumers will never care about the difference between wetland types anyway.

westnordost:
> Is there a particular reason you want these presets? 

No particular reason. But there is a general reason why I am adding a lot of presets and that is
1. I find the iD presets to be the most complete and well maintained, so I see potential in the iD presets being used for wider use cases than just for the iD editor. Namely, to have a general [dictionary of map features to and from OSM tags](https://github.com/westnordost/osmfeatures). Outside the context of OSM editors, for example to be able to easily search a consumer map by just typing the localized name of the feature, would be somewhat of a killer feature.
2. I have a private list of OSM tags to German localization that I want to contribute but want to make sure first that the presets in that list already exist in iD presets ([actually it's not that private](https://github.com/westnordost/tagdictionary/blob/master/strings/names/de.yml))

westnordost:
Anyway, if this is your final word on this matter, feel free to close this again. I will then maybe come back to it once the short descriptions are implemented.

bhousel:
This reminds me of the basins:  https://github.com/openstreetmap/iD/issues/5497

I was ok with adding a field so that users who know and care about such things can pick whether it's an `infiltration`/`detention`/`retention` basin.   But adding 3 presets for those things would have been too much.

Issue #5869, opened 2019-02-05
Disconnected highway warning should search deeper than a single connection


quincylvania:
iD has a warning for disconnected highways, but right now it can be satisfied by a single connection. Highways connected to each other but not the larger road network should still be flagged. This would help prevent "routing islands" like the one pictured:

<img width="501" alt="screen shot 2019-02-05 at 5 03 03 pm" src="https://user-images.githubusercontent.com/2046746/52307315-4c014500-2968-11e9-821e-b2f050e2472f.png">

tordans:
I noticed a few of those while looking at KeepRight-Issues. So they are covered in some of the validation options at least.

quincylvania:
We did this in 2.15!

PR #6284, opened 2019-05-01
Improve validation for disconnected highways #5869

Bonkles:
This addresses issue #5869. 

(This code was originally by @gaoxm- Thanks Ming!) 

Old behavior: mappers could create new highways that didn't attach to any existing highways. If the new highways were all connected to each other, no warning flag would present itself. 

New behavior: mappers will receive a warning if they create any new highways that aren't ultimately connected to existing ones. 

Validation catching two new connected primary (orange) highways that are otherwise disconnected from the rest of the highways: 
image

Issue is no longer flagged when the two new highways are connected: 
image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1353,7 +1353,9 @@ en:
       tip: "Find unconnected highways and paths"
       highway:
         message: "{highway} is disconnected from other roads and paths"
+        message_new_road: "New {highway} is disconnected from existing road network"
         reference: "Highways should connect to other highways or building entrances."
+        reference_new_road: "New roads should connect to existing road network or building entrances."
     fixme_tag:
       title: '"Fix Me" Requests'
       message: '{feature} has a "Fix Me" request'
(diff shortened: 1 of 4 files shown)


Bonkles:
Seems that travis is unhappy with my submission- taking a look at what the issue is!

gaoxm:
cc @wonga00 @chrisklaiber

quincylvania:
Thanks for the work!

I'd really like this validation to work for all routing islands, not just new ways. We could have a single issue for each routing island, like "Roads aren't connected to the primary road network", and when you click it all the ways in that island are selected.

quincylvania:
I'm also fine with merging this as-is and then we can update it as-needed.

bhousel:
> I'd really like this validation to work for all routing islands, not just new ways. We could have a single issue for each routing island, like "Roads aren't connected to the primary road network", and when you click it all the ways in that island are selected.

Yes I was thinking this too..  But like you said, I'm also ok with merging it as-is and we can develop it more.

PR #6294, opened 2019-05-02
Add an empty favicon for preview.ideditor.com

tordans:
There are multiple calls to the favicon on the preview.ideditor.com system in one session:
 > GET http://preview.ideditor.com/favicon.ico 403 (Forbidden)

According to https://stackoverflow.com/questions/1321878/how-to-prevent-favicon-ico-requests#comment50777781_13416784 this should show an empty favicon which will prevent the browser to try loading the missing icon.

- Is this the right index.html file that is used for the preview build?
- Is this the smart way to solve this? – Please just close it otherwise :).

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -4,7 +4,7 @@
         <meta charset='utf-8'>
         <title>iD</title>
         <link rel='stylesheet' href='iD.css'>
-        <link rel="icon" type="image/png" href="data:image/png;base64,iVBORw0KGgo=">
+        <link rel='icon' type='image/png' href='data:image/png;base64,iVBORw0KGgo='>
 
         <!-- mobile devices -->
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks this looks good! 
I might eventually replace it with an actual icon though. 😆

quincylvania:
An actual icon would be great! While we're at it we should add `apple-touch-icon` support. This is used when adding iD to the home screen on iOS. This is helpful since there's no Safari UI in the way when you launch a webpage from a home screen shortcut.

tordans:
Thanks.

--- 

> I might eventually replace it with an actual icon though.

…
> While we're at it we should add apple-touch-icon support

I did not go there, since both are present for the common use case of 
view-source:https://www.openstreetmap.org/edit

It's just the preview that had the 404.

PR #6297, opened 2019-05-02
Remove crazyegg from the preview-index.html

tordans:
It looks like it is more than 6 years old https://github.com/openstreetmap/iD/blame/372bb7d892048d67415087f474688f85391bab2a/dist/index.html#L28 and the linked URL seems to be broken:
> $ curl http://dnn506yrbagrg.cloudfront.net/pages/scripts/0013/6714.js
> "undefined"==typeof CE2&&(CE2={}),CE2.uid=136714,CE2.status="no data available";

So to me it looks like it is save to remove it(?)

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -28,12 +28,6 @@
               var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
             })();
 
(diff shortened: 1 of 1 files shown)


bhousel:
Yup  🔪

PR #6302, opened 2019-05-03
Text raw tag editor / Copy-paste tags

### LINKED ISSUE #839: Copy-paste tags

### LINKED ISSUE #6185: Allow users to paste several "k=v" tag pairs into the raw tag editor

bhousel:
This change adds ability to toggle between the list view and a text view of tags, so that people can make faster tag edits or copy/paste tags.

closes #6185 
closes #839


bhousel:
It works like this ✨

text tag editing

modules/ui/raw_tag_editor.js
```diff
@@ -75,13 +81,71 @@ export function uiRawTagEditor(context) {
...
+            .data([0]);
+
+        textarea = textarea.enter()
+            .append('textarea')
+            .attr('class', 'tag-text' + (_tagView !== 'text' ? ' hide' : ''))
+            .call(utilNoAuto)
+            .merge(textarea);
+
+        textarea
+            .attr('rows', rowData.length + 1)
```
1ec5: Can we disable spell checking and autocorrect? Safari’s autocorrect is pretty aggressive inside textareas, and the 〰️ underlines get pretty annoying with raw tags.

tordans:
Does iD have code already that is able to resize the height of an textarea based on its content? (Like https://css-tricks.com/textarea-tricks/#article-header-id-6) I would not add it just for this element (it's quite a bit of JS just for this\*\*), but if it only needs to be "activated" for this textarea, it would be nice for the last bit of the GIF where the text area has a scrollbar.
Bildschirmfoto 2019-05-03 um 07 08 27
\*\*) The `<div contenteditable>` approach from the article above might be a less-code solution that works in this case; also maybe for the autocorrect issue?

sun-geo:
Wow, happy to see any progress for (bulk) copy/paste function :-)

Yes, maybe the window height could increased a litte bit same time while when pasting the new lines? At that moment:
y-window-height

modules/ui/raw_tag_editor.js
bhousel: I dunno sometimes it does the right thing :trollface: 

Screenshot 2019-05-03 10 09 51

bhousel:
This is good enough to merge now.. 
Please try it out in http://preview.ideditor.com/master/ and open any followup issues that you find!

mmd-osm:
Some folks add new line characters in the description: https://www.openstreetmap.org/way/463013906

Would this work as is?

bhousel:
> Some folks add new line characters in the description: https://www.openstreetmap.org/way/463013906
> 
> Would this work as is?

Yep, that's handled in 16ec25753  
I'm using `JSON.stringify` to escape newlines and other unprintable characters.

mmd-osm:
ok, looks good, newline got replaced by an `\n`

I found one issue with two rows having the same key: one of them gets discarded silently when I switch back to the list.

```
sport=soccer
sport=rugby
```

quincylvania:
> I found one issue with two rows having the same key: one of them gets silently discarded when I switch back to the list.

I think the preferred behavior would be to overwrite the old value, if any.

bhousel:
> I think the preferred behavior would be to overwrite the old value, if any.

It just keeps whichever is the last one in the list.  I'm ok with this, since I'd expect people to paste their new values at the end of the list.

sun-geo:
Maybe change the blue and grey colour:
image

(blue for clickable items, like "zoom to this" / "All fields" / "All tags")?

bhousel:
> Maybe change the blue and grey colour:
> (blue for clickable items, like "zoom to this" / "All fields" / "All tags")?

Oh yeah, I was thinking about this too.  I do like using a consistent blue for links. Maybe I could make it look more like these toggles we use elsewhere:

toggle

sun-geo:
yes.

Or just one blue button, with remaining blue colour, but changing list icon <=> cursor icon? I'm not sure...

PR #6313, opened 2019-05-04
add preset for water=wastewater

westnordost:
A settling or clarifying basin for wastewater. Used 17+k times and [documented in the wiki](https://wiki.openstreetmap.org/wiki/Tag:water%3Dwastewater).

Usually found in a wastewater plant, but on the countryside, it may also look like an innocent pond (if there are no signs).

I am not sure if I hit the correct English name here, I did not find a wikipedia article that is only about the basin, not the whole wastewater plant.

### DIFF
--- a/data/presets/presets/natural/water/wastewater.json
+++ b/data/presets/presets/natural/water/wastewater.json
@@ -1,0 +1,23 @@
+{
+    "icon": "maki-water",
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "natural": "water",
+        "water": "wastewater"
+    },
+    "reference": {
+        "key": "water",
+        "value": "wastewater"
+    },
+    "terms": [
+        "excrement",
+        "shit",
+        "sewage",
+        "wastewater",
+        "Settling Basin",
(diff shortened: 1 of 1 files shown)


bhousel:
> I am not sure if I hit the correct English name here, I did not find a wikipedia article that is only about the basin, not the whole wastewater plant.

It sounds good to me - I think most English speakers would understand it.

bhousel:
> I did not find a wikipedia article that is only about the basin, not the whole wastewater plant.

btw, I guess it would probably be some kind of `landuse=industrial`+`industrial=?`  if people wanted to map a wastewater treatment plant.

westnordost:
It's man_made=wastewater_plant I think 

Am 5. Mai 2019 05:22:03 MESZ schrieb Bryan Housel <notifications@github.com>:
>> I did not find a wikipedia article that is only about the basin, not
>the whole wastewater plant.
>
>btw, I guess it would probably be some kind of
>`landuse=industrial`+`industrial=?`  if people wanted to map a
>wastewater treatment plant.

PR #6321, opened 2019-05-05
add castle-related presets

westnordost:
- [`man_made=cross`](https://wiki.openstreetmap.org/wiki/Tag%3Aman_made%3Dcross), documented since 2011, used 4600+ times. This is a "cross of no religious or historic importance", so, usually a **summit cross**.
- [`water=moat`](https://wiki.openstreetmap.org/wiki/Key:water), documented since 2013, used 400+ times
- [`historic=city_gate`](https://wiki.openstreetmap.org/wiki/Tag%3Ahistoric%3Dcity_gate), documented since 2010, used 3000+ times
- [`barrier=sally_port`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dsally_port), documented since 2012/2016, used 4000+ times, approved. A side gate or **postern** in castles, e.g. for sorties.
- [`historic=fort`](https://wiki.openstreetmap.org/wiki/Tag%3Ahistoric%3Dfort), documented since 2012/2016, used 3500 times
- [`castle_type=fortress`](https://wiki.openstreetmap.org/wiki/Tag%3Acastle_type%3Dfortress), documented since 2014 in German wiki and in English wiki since 2015, used 1400+ times. Heavily fortified castle for military use.
- [`castle_type=palace`](https://wiki.openstreetmap.org/wiki/Tag%3Acastle_type%3Dpalace), documented since 2014 in German wiki and in English wiki since 2015, used 1300+ times. A royal residence.
There is also `historic=palace`, but it is used only 380+ times and was never documented.
- [`castle_type=stately`](https://wiki.openstreetmap.org/wiki/Tag%3Acastle_type%3Dstately), documented since 2014 in German wiki only, translated to English wiki in 2017, used 4000 times. A residence of the nobility.
- [`historic=manor`](https://wiki.openstreetmap.org/wiki/Tag%3Ahistoric%3Dmanor), documented since 2011, used 5000+ times. Or mansion. A residence on an estate.
There is also `castle_type=manor`, but it is used much less often (2300 times) and was documented later (2015). There is no discussion in the wiki whether `historic=manor` should be replaced by that `castle_type=manor` tagging.

How exactly some of the castle types presented here differ from another, for example manor vs stately, is I think not so important because in the end, what counts is the denomination used on the plaques / tourist guides / info booklets / wikipedia article. So, the mapper mapping with iD will likely input the denomination on this info material there and expect to get the correct preset for that.

By the way, "castle" may be used rather as a generic term for most of the above, that is why I did not add the `castle_type=defensive` - this would be the typical medieval castle.

### DIFF
--- a/data/presets/presets/barrier/sally_port.json
+++ b/data/presets/presets/barrier/sally_port.json
@@ -1,0 +1,14 @@
+{
+    "icon": "fas-dungeon",
+    "geometry": [
+        "vertex"
+    ],
+    "tags": {
+        "barrier": "sally_port"
+    },
+    "terms": [
+        "Postern",
+        "castle side gate"
+    ],
+    "name": "Sally Port"
+}
--- a/data/presets/presets/historic/castle/fortress.json
+++ b/data/presets/presets/historic/castle/fortress.json
@@ -1,0 +1,26 @@
+{
+    "icon": "maki-castle",
+    "fields": [
+        "name",
+        "building_area",
+        "access_simple",
+        "start_date"
+    ],
+    "geometry": [
+        "point",
+        "area"
+    ],
+    "reference": {
+        "key": "castle_type",
+        "value": "fortress"
+    },
+    "tags": {
+        "historic": "castle",
+        "castle_type": "fortress"
+    },
+    "terms": [
+        "citadel",
+        "military"
+    ],
+    "name": "Historic Fortress"
+}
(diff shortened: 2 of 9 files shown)


bhousel:
Seems fine, thanks @westnordost  👍 
Remember to try to use Title Case for the preset names.

westnordost:
Oh right, I forgot it with the summit cross

PR #6322, opened 2019-05-05
fix case for summit cross

westnordost:


bhousel:
I already fixed it, but thanks 👍

PR #6334, opened 2019-05-07
New 'G' Hotkey:  Highlight geometry edits

Bonkles:
This hotkey toggle highlights any geometry edits that the user has made in their current session. useful for users to determinine what they've changed at a glance! Original work credited to @deepbasu007 . 

Note: this is for geometry changes only- tag changes will not show up in the highlighting.  The following GIF shows this behavior. 

highlight_edited_ways


Bonkles:
Just discussed this one with @bhousel and @quincylvania. I discovered some bugs with this implementation: 
* it does not honor 'wireframe mode' 
* it also mistakenly displays ways as 'edited' when connected to a node whose tags have changed. 

I will close this PR for now. Expect this work to return after the release of 2.15 with: 
1) wireframe mode awareness
2) different styling for tag changes vs. geometry changes
3) the ability to show the original unedited map changes alongside the current edits.

PR #6337, opened 2019-05-08
create highway=emergency_bay preset

westnordost:
Used 5000+ times, documented since 2014 in German wiki and since 2017 in English wiki: https://wiki.openstreetmap.org/wiki/Tag:highway%3Demergency_bay

According to taginfo, mostly used in Germany, Austria, Switzerland, Italy and France so far.

I am not sure if "Highway Emergency Bay" is the correct word for it in English, I did not find any wikipedia article or translation about it.


westnordost:
Uh darn, I didn't add the .json extension, wait a moment...

quincylvania:
Hmm I'd really prefer not to introduce another `highway` type for routers to deal with. I'd recommend `highway=service` with a `service` subtag if these really need to be mapped as separate ways.

This might be okay for vertices similar to how we have `highway=passing_place`.

westnordost:
So, remove the `line`? More than 4 out of 5 times, the tag is used on vertices rather than lines.

quincylvania:
@westnordost I think we can accept this with just `vertex` geometry. I'd use a vehicle icon and change the name to "Emergency Stopping Place" or similar.

westnordost:
Ok, done

quincylvania:
@westnordost Thanks!

PR #6338, opened 2019-05-08
Add runaway truck ramp preset

westnordost:
Used 800+ times so far, documented and approved since 2012

See https://wiki.openstreetmap.org/wiki/Tag%3Ahighway%3Descape

### DIFF
--- a/data/presets/presets/highway/escape.json
+++ b/data/presets/presets/highway/escape.json
@@ -1,0 +1,15 @@
+{
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "highway": "escape"
+    },
+    "terms": [
+        "Runaway Truck Lane",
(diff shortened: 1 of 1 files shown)


bhousel:
Same as with #6337 - this should be a `highway=service` = `service=something`

westnordost:
Did you read the approved proposal and the preceeding discussion?

https://wiki.openstreetmap.org/wiki/Proposed_features/escape_lane

service=escape_lane was the original proposal it seems but it was rejected and instead highway=escape was chosen for the reasons discussed there.

Am 9. Mai 2019 16:05:38 MESZ schrieb Bryan Housel <notifications@github.com>:
>Same as with #6337 - this should be a `highway=service` =
>`service=something`

PR #6351, opened 2019-05-12
Update for playground equipment presets

FelisPimeja:
Additional equipment types according to wiki:

- activity panel
- aerial rotator
- climbing wall
- excersise
- hopscotch
- playhouse
- sledding
- splash pad
- teen shelter
- trampoline
- water

Additional description fields for playground equipment:

- blind
- sitting_disability
- walking_disability
- min_age
- max_age
- theme

### DIFF
--- a/data/presets/presets/playground/activitypanel.json
+++ b/data/presets/presets/playground/activitypanel.json
@@ -1,5 +1,19 @@
 {
     "icon": "maki-playground",
+    "fields": [
+        "playground/theme",
+        "playground/min_age",
+        "playground/max_age",
+        "playground/baby",
+        "wheelchair"
+    ],
+    "moreFields": [
+        "name",
+        "playground/blind",
+        "playground/walking_disability",
+        "playground/sitting_disability",
+        "material"
+    ],
     "geometry": [
         "point"
     ],
--- a/data/presets/presets/playground/activitypanel.json
+++ b/data/presets/presets/playground/activitypanel.json
@@ -1,0 +1,10 @@
+{
+    "icon": "maki-playground",
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "playground": "acrivitypanel"
+    },
+    "name": "Playground Activity Panel"
+}
--- a/data/presets/presets/playground/aerial_rotator.json
+++ b/data/presets/presets/playground/aerial_rotator.json
@@ -1,0 +1,10 @@
+{
+    "icon": "maki-playground",
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "playground": "aerialrotator"
+    },
+    "name": "Playground Overhead Rotator"
+}
--- a/data/presets/presets/playground/aerial_rotator.json
+++ b/data/presets/presets/playground/aerial_rotator.json
@@ -1,5 +1,17 @@
 {
     "icon": "maki-playground",
+    "fields": [
+        "playground/theme",
+        "playground/min_age",
+        "playground/max_age"
+    ],
+    "moreFields": [
+        "name",
+        "playground/blind",
+        "playground/walking_disability",
+        "playground/sitting_disability",
+        "material"
+    ],
     "geometry": [
         "point"
     ],
--- a/data/presets/presets/playground/balance_beam.json
+++ b/data/presets/presets/playground/balance_beam.json
@@ -1,5 +1,17 @@
 {
     "icon": "maki-playground",
+     "fields": [
+        "playground/theme",
+        "playground/min_age",
+        "playground/max_age",
+        "playground/baby"
+    ],
+    "moreFields": [
+        "name",
+        "playground/blind",
+        "playground/sitting_disability",
+        "material"
+    ],
     "geometry": [
         "point",
         "line"
--- a/data/presets/presets/playground/exercise.json
+++ b/data/presets/presets/playground/exercise.json
@@ -1,5 +1,18 @@
 {
     "icon": "maki-playground",
+    "fields": [
+        "playground/theme",
+        "playground/min_age",
+        "playground/max_age",
+        "wheelchair"
+    ],
+    "moreFields": [
+        "name",
+        "playground/blind",
+        "playground/walking_disability",
+        "playground/sitting_disability",
+        "material"
+    ],
     "geometry": [
         "point"
     ],
--- a/data/presets/presets/playground/exercise.json
+++ b/data/presets/presets/playground/exercise.json
@@ -1,0 +1,10 @@
+{
+    "icon": "maki-playground",
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "playground": "exercise"
+    },
+    "name": "Playground Exercise Equipment"
+}
--- a/data/presets/fields/playground/baby.json
+++ b/data/presets/fields/playground/baby.json
@@ -1,5 +1,5 @@
 {
     "key": "baby",
     "type": "check",
-    "label": "Baby Seat"
+    "label": "Primarily designed for babies"
 }
--- a/data/presets/fields/playground/blind.json
+++ b/data/presets/fields/playground/blind.json
@@ -1,0 +1,6 @@
+{
+    "key": "blind",
+    "type": "combo",
+    "label": "Can be used by blind people",
+    "options": ["yes", "no", "limited"]
+}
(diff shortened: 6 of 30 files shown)


tordans:
Hi FelisPimeja, I am a big fan of mapping playground equipment.

> Additional equipment types according to wiki:
> 
> * …

Personally, I used many of those manually in the past. Sounds good to me to add them. 

Some are more used than other, though. 
I wonder if there should be a cut to add them only if they are used at least 100 times?
Eg aerialrotator is used only 56 x, yet.

- https://taginfo.openstreetmap.org/keys/playground#values

> Additional description fields for playground equipment:

However, I am very unsure about the following key. Most of them don't even have a wiki page. They are not used a lot.

I know they are [all "officially" from the proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Playground_Equipment), but in my opinion they should be used more widely to be added to the preset.

> * blind

No wiki page. No info on the "Blind"-Wiki-Collection. Only 141 usages.
- https://wiki.openstreetmap.org/wiki/Key:blind
- https://taginfo.openstreetmap.org/keys/blind
- https://wiki.openstreetmap.org/wiki/OSM_for_the_blind

> * sitting_disability
> * walking_disability

Same as blind.
I never saw them in use. And would not know how to use them.
- No wiki https://wiki.openstreetmap.org/wiki/Key:sitting_disability, https://wiki.openstreetmap.org/wiki/Key:walking_disability
- used 90x only https://taginfo.openstreetmap.org/search?q=walking_disability
- used 80x only https://taginfo.openstreetmap.org/search?q=sitting_disability

> * min_age
> * max_age

I always wondered what to do with those keys on playground structures. There is no sign on such structures that I saw which would help me with this. I could make something up, but thats bad mapping IMO. I understand the usecases the wiki shows outside of playground cases, I just dont see why one would add them to a playground element.
- https://wiki.openstreetmap.org/wiki/Key:min_age
  - used only 496 times in combination with playground https://taginfo.openstreetmap.org/keys/min_age#combinations
- https://wiki.openstreetmap.org/wiki/Key:max_age
  - used ~400 times in combination with playground https://taginfo.openstreetmap.org/keys/max_age#combinations

> * theme

I tried using this key for playground but it never felt helpful. I kind of makes sense to me on the leisure=playground, but not on playground=trampoline and such. And again, nearly no one uses it:
- 16x theme + leisure https://taginfo.openstreetmap.org/keys/theme#combinations
- 16x theme + playground https://taginfo.openstreetmap.org/keys/theme#combinations

Personally, I would skip all the additional tags and just go with the most used playground=* values. And review all this in a year or two to see how mapping evolved in the meantime.

--- 

PS: The whole area of playground mapping would get an even bigger boost IMO, if we got OSM-Carto to render the equipment on the higher/highest zoom level. The github discussion about this got stuck a while back, unfortunately.

FelisPimeja:
Ok, @tordans, thank you for detailed explanation. I see your point of view and it makes sence to me. Almost agree about playground values count and mostly about description fields. Except for those:

- **min_age**/**max_age** isn't popular with playground equipment yet (though it's 522/482 times already used - good dynamics). And children usually better deside themselves what they want to play with : ). But where I live (Moscow, Russia) we have information sings on lots of new playgrounds telling what equipment for what ages is designed (see the image). So that information may be helpfull in some way.

IMG_9742

- **theme** tag isn't so helpfull, as for me. But it may be used to describe stylization for lots of playground structures (castle, tower, ship, etc) and spring riders (cars, motorcycles, horses and lots of other animals). And it's funny!

Anyway I'll close this PR and open a new one to make it cleaner. Just tell me what do you think about my suggestions above.

tordans:
Hi!

> min_age/max_age (…) But where I live (Moscow, Russia) we have information sings on lots of new playgrounds telling what equipment for what ages is designed (see the image).

Great example. Looks like a good reason to me to keep the tags.

> theme (…)

Yes, and maybe I would use it more, once more values are suggested due to more usage after the preset.

> Anyway I'll close this PR and open a new one to make it cleaner.

Maybe wait for some feedback by Bryan and Quincy, first.

FelisPimeja:
Ok. Thank you! Looking for further suggestions.

data/presets/presets/playground/balance_beam.json
```diff
@@ -1,5 +1,17 @@
 {
     "icon": "maki-playground",
+     "fields": [
+        "playground/theme",
```
quincylvania: Tip: Instead of repeating the fields for each playground preset you can set the `fields` and `moreFields` of `_playground.json` and iD will inherit them automatically. You can still override these properties in individual files if you want.

If this doesn't make sense then don't worry about it.

data/presets/presets/playground/exercise.json
```diff
@@ -0,0 +1,23 @@
...
+        "playground/blind",
+        "playground/walking_disability",
+        "playground/sitting_disability",
+        "material"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "playground": "exercise"
```
quincylvania: I know this is somewhat different than `leisure=fitness_station` but perhaps not different enough to warrant a second tag…

data/presets/presets/playground/aerial_rotator.json
```diff
@@ -0,0 +1,22 @@
...
+        "playground/blind",
+        "playground/walking_disability",
+        "playground/sitting_disability",
+        "material"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "tags": {
+        "playground": "aerialrotator"
```
quincylvania: This has pretty low usage, we might wait to add it.

data/presets/presets/playground/activitypanel.json
```diff
@@ -0,0 +1,24 @@
+{
+    "icon": "maki-playground",
+    "fields": [
+        "playground/theme",
+        "playground/min_age",
```
quincylvania: I think it's okay to add `min_age` and `max_age` to these, but I'm noticing it was a past mistake for these fields to be under the `playground` folder. They can also be used for other features like nightclubs. We can fix this later.

quincylvania:
Lacking further work on this, I just went ahead and manually picked out the most salient changes. Thanks for getting this started, @FelisPimeja, and sorry I couldn't merge it directly!

matkoniecz:
Is it intentional effect that `playground:theme=playground` got wildly popular?

https://github.com/openstreetmap/iD/commit/8a6fa241df1b32a4fe32ebd33dfa256b04483c18 that appears to result in wild growth of `playground:theme`, mostly `playground:theme=playground`

See https://taginfo.openstreetmap.org/tags/playground%3Atheme=playground#chronology https://taginfo.openstreetmap.org/keys/playground%3Atheme#chronology that now accounts for over half of uses of this tag.

See https://taginfo.openstreetmap.org/keys/playground%3Atheme#values

If anyone knows what this tag means they are welcome to answer at https://wiki.openstreetmap.org/wiki/Talk:Key:playground:theme

(from quick review it seems to be used mostly by newbie confused mappers due to prominent display in iD presets inviting to fill something there, with `playground` given as an example - so I expect that this data is meaningless)

1ec5:
8a6fa241df1b32a4fe32ebd33dfa256b04483c18 just added a plain `combo` field for `playground:theme` but no hard-coded options, so the field suggests the most common values for that key from taginfo. Most likely the `playground` value was already near the top, so at best this field contributed to a feedback loop. One possible solution would be to define well-known values of that key in [the field in id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/blob/bd3711852471a6fd51db2283b14dfe63ef2e31b4/data/fields/playground/theme.json), which will make those values look more legitimate than the ones pulled straight from taginfo.

matkoniecz:
https://taginfo.openstreetmap.org/tags/playground%3Atheme=playground#chronology https://taginfo.openstreetmap.org/keys/playground%3Atheme#chronology indicate that `playground` value was unused before iD started to request to fill theme.

Maybe it would be better not as a field available for every single playground? Most playgrounds have no theme. Maybe `playground` value should be interpreted as "no theme here for this playground".

1ec5:
The id-tagging-schema repo would be a better place to discuss our options for resolving this issue. As far as I can tell, though, it’s a complete mystery as to how this value became more popular than `castle`, which was the most popular at the time this PR was released.

PR #6352, opened 2019-05-12
Add filter building:part and indoor

tordans:
- Extract the building:part from the building filter into it's own filter
- Add indoor filtering
- Update api doc


tordans:
@bhousel are you open to this topic and PR? I was just curious and looked into it for fun …

Also: Is this close to what needs doing? I am not sure if it is really just those few lines.

quincylvania:
I haven't thought about it too much but I imagine these filters would be useful.

bhousel:
Yes @tordans I think this looks ok 👍 
I'd test these around places that have lots of building parts or indoor features mapped, to make sure they are catching everything you mean for them to  (I'm not super familiar with the tags used for these things).

quincylvania:
Pretty much everything indoors will have an `indoor` tag except for some `highway=corridor`. iD should probably recommend adding `indoor=yes` or `indoor=corridor` to those.

PR #6367, opened 2019-05-15
Allow "layer" tags alone to resolve building crossing issues

### LINKED ISSUE #5924: Drive-Through crosses Roof warning
I was editing a building that has a covered drive-through and got a "Drive-Through crosses Roof" warning. I've followed the [Tag:building=roof](https://wiki.openstreetmap.org/wiki/Tag:building%3Droof) suggestions and made the roof part a separate area with `building:roof` and `layer:1`. The drive-through way has no Structure annotations. See the following pic:

image

I'm using 64-bit Firefox 65.0.1 on 64-bit Linux and using iD editor 2.14.1

I think I've done this edit right, but iD is warning me here. Is this a false positive?

Additionally the pop-up descriptive text when you hover over the warning says "Highways crossing buildings should use bridges, tunnels, coverings, or entrances. I searched the OSM wiki for "coverings", to see if that would be what I want to use here, but that term doesn't exist in the wiki. So the descriptive text also led me down the wrong road to resolve this.

quincylvania:
Closes #5924.

This lets instances of buildings crossing highway/railways/waterways to be disambiguated by simply setting different `layer` tags. The `covered` tag is ignored altogether.

Leaving this as a draft since I'm still not 100% sure this is the desired behavior. 

Screen Shot 2019-05-15 at 12 13 05 PM

This also makes makes the Layer editor visible in the Structure field even without a structure tag if a "layer" tag is present

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1334,9 +1334,9 @@ en:
       building-building:
         reference: "Buildings should not intersect except on different layers."
       building-highway:
-        reference: "Highways crossing buildings should use bridges, tunnels, coverings, or entrances."
+        reference: "Highways crossing buildings should use bridges, tunnels, or different layers."
       building-railway:
-        reference: "Railways crossing buildings should use bridges or tunnels."
+        reference: "Railways crossing buildings should use bridges, tunnels, or different layers."
       building-waterway:
         reference: "Waterways crossing buildings should use tunnels or different layers."
       highway-highway:
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -1637,10 +1637,10 @@
                     "reference": "Buildings should not intersect except on different layers."
                 },
                 "building-highway": {
-                    "reference": "Highways crossing buildings should use bridges, tunnels, coverings, or entrances."
+                    "reference": "Highways crossing buildings should use bridges, tunnels, or different layers."
                 },
                 "building-railway": {
-                    "reference": "Railways crossing buildings should use bridges or tunnels."
+                    "reference": "Railways crossing buildings should use bridges, tunnels, or different layers."
                 },
                 "building-waterway": {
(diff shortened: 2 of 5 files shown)


BjornRasmussen:
Thanks for this!! The JOSM validator currently works like this, so this will be nice to avoid confusion about the validator

bhousel:
I'm wary of this approach, because I still think the ground has an implied layer 0, and features below it may not render.

BjornRasmussen:
I agree with @bhousel - we need to make sure that, for example, a service road under a roof doesn't get the tag `layer=-1`

cserpell:
Please make this through, this issue is driving me nuts with some new editors that do not understand that adding the layer=1 to a building=roof (specially in gas stations) is enough to make all service ways below to be ok.

SilentSpike:
@cserpell According to the `layer` tag [wiki page](https://wiki.openstreetmap.org/wiki/Key:layer) this is not true

>When ways are passing on different levels apply layer=* only to the way which also has the bridge/tunnel attribute. Only ways with one of the tags/attributes tunnel=*, bridge=*, highway=steps, highway=elevator, covered=* should be tagged with the layer tag, similar for railways and waterways.

I'm not sure where steps and elevators come into it, but this matches my understanding of the layer tag. The feature still must be tagged appropriately

matkoniecz:
Fortunately negative layer does not mean that something is underground, explicit `location=underground` should be used to tag this.

quincylvania:
Does this approach (not requiring `covered=yes`) sound okay if iD wouldn't assign negative layers?

bhousel:
If you want to hard-code rules for roofs, go ahead.

quincylvania:
> If you want to hard-code rules for roofs, go ahead.

I don't particularly. But I _think_ if any building has `layer=1` with a crossing road then it's implied that the building covers the road and iD shouldn't flag it.

PR #6376, opened 2019-05-17
Flag routing islands

quincylvania:
This upgrades the disconnected ways validation to flag entire routing islands as single issues.

Screen Shot 2019-05-16 at 10 59 28 PM

It also incorporates ferry routes as first-class routable ways. This means ferry routes can resolve or extend routing islands and that disconnected ferry routes are flagged.

Routable ways are considered connected to the larger network if they connect to building entrances or parking garage entrances, or overlap unloaded tiles.

The code could be made more efficient eventually but it's okay for now.

@bhousel please give it a lookover and merge when ready. Make any changes you need.


bhousel:
This looks really good!  Thanks @quincylvania 👏

PR #6386, opened 2019-05-18
Allow user to adjust the threshold for the unsquare building warning

bhousel:
This is a followup from #6234, which introduced an unsquare building validator and fixer.
This PR allows users to adjust the threshold for flagging the feature as unsquare.

We have received a lot of positive feedback about the unsquare building validator from people who have tried it out (and a lot of negative feedback from people who.. obviously haven't tried it out 🙄 )

Anyway, @quincylvania and I have spent many hours testing this feature across many different cities and building types.  We've found that there really isn't a magic number for squareness that works everywhere.

This PR allows people to change the number if they want to.  **The default value will be set to something conservative,** so that it only raises a warning if the building is almost square already.  

[Extremely bad mapping of mostly square buildings](http://preview.ideditor.com/master/#background=Mapbox&disable_features=boundaries&map=18.20/40.56675/-74.63549&offset=5.16,-2.98) can benefit from increasing the number. 

Worth mentioning again, that default behavior of the iD validator only validates things that the user has edited themselves in their session.

Screenshot 2019-05-18 15 46 54


BjornRasmussen:
> I think we need to use the same degree threshold for the fix and autofix as for the detection. Otherwise iD could square larger diagonals than the user specified. Or if the user's threshold is greater than 13 then running the fix may not resolve the issue.

This seems risky, since it gives contributors the power to destroy a lot of data.  Things in iD are usually safe in this way, so giving people the ability to destroy data will make them think that using the tool is safe at any threshold.

quincylvania:
> This seems risky, since it gives contributors the power to destroy a lot of data. Things in iD are usually safe in this way, so giving people the ability to destroy data will make them think that using the tool is safe at any threshold.

@BjornRasmussen I don't see how this lets users "destroy" any data. It just makes the fix match the threshold they manually set, giving expected results. This makes the tool safer by making it more predictable.

BjornRasmussen:
Autofix should only be available in cases where no damage can be done

bhousel:
> This seems risky, since it gives contributors the power to destroy a lot of data. Things in iD are usually safe in this way, so giving people the ability to destroy data will make them think that using the tool is safe at any threshold.

I see your point about making users feel safe to use the tool, but this is why the default threshold would be set very conservatively, and only get changed by someone who is really trying to clean up a mess.

It's actually very easy to mess up OSM with _any_ editor.  iD isn't special in this way.  A user can do more serious damage by disconnecting motorways than by squaring buildings, and I still think that anyone really determined to damage the map will probably upgrade to a more powerful editor or just write their own tool to do so.

BjornRasmussen:
I think that the warnings should be shown if the manually set threshold is met, and the autofix should be available only if the buildings couldn't be damaged by squaring.  This makes sense, since the issues still show up as expected.

BjornRasmussen:
By autofix, I mean the ability to fix hundreds of buildings at once.  Quickfix, which is available when selecting a building, should be available for all warnings.

BjornRasmussen:
>  anyone really determined to damage the map will probably upgrade to a more powerful editor or just write their own tool to do so.

I was thinking more of newbies trying to resolve all warnings in their area, and then setting the threshold up super high to catch "all" warnings.  Then, they might mechanically fix all of the issues, and square buildings that shouldn't be square without looking at those buildings.

quincylvania:
> I was thinking more of newbies trying to resolve all warnings in their area, and then setting the threshold up super high to catch "all" warnings. Then, they might mechanically fix all of the issues, and square buildings that shouldn't be square without looking at those buildings.

@BjornRasmussen I'd rather see how this plays out in practice. Would a "newbie" really find the issues pane, set it to validate "Everything", open the Rules section, set a high threshold, pan over a lot of stuff, click "Fix All", then save hundreds of changes without thinking twice? I think most new mappers are very hesitant to change too much.

quincylvania:
I addressed the issues I mentioned earlier and I think this is ready to merge!

bhousel:
> I addressed the issues I mentioned earlier and I think this is ready to merge!

Thanks so much for finishing this up, looks great  🙌

PR #6389, opened 2019-05-19
Support special styling for Wikidata-tagged features

bhousel:
I'm experimenting with the idea of giving a special style to features that have been tagged with a `wikidata` or `brand:wikidata` tag.  These features have some extra protections put in place to prevent accidental name changes and deletions, so it would be helpful to indicate that they are special somehow.  

Screenshot 2019-05-18 23 56 49

I kind of _don't_ like the green, but might keep it or do something else.  Maybe I'll leave the text halos white.  Or maybe something similar to how verified accounts on Twitter get a special badge.  I don't want something that steals the user's attention so much.


BjornRasmussen:
How about keep the text halo white, and make the icon light gray instead of green?
Buildings wouldn't be changed by this, though, so perhaps they could get something else

Adamant36:
"How about keep the text halo white, and make the icon light gray instead of green?"

That was totally going to be my suggestion.

bhousel:
Another try, this time with just subtle green stroke.  It's better, but still not sure if I like it.
I don't really want to make them grey, because they aren't really readonly.

Screenshot 2019-05-20 00 33 06

1ec5:
A halo would conflict with the blue halo for relation members. How about boldfacing, italicizing, or underlining the label (when shown), assuming that most Wikidata-linked features would have names? This approach has the additional benefit of being useful to those with color blindness.

bhousel:
I ended up doing this

- Markers with locked fields have a gray outline and slight gray fill
- Locked fields get a lock 🔒 
- Locked fields don't get revert ↩️or delete buttons 🗑 
- Same calculation as before for locking a field
  - if `wikidata` tag is set, lock `name` 
  - if `brand:wikidata` tag is set, lock either `name` or `brand` 

Screenshot 2019-05-20 17 52 22

Screenshot 2019-05-20 17 54 11

Screenshot 2019-05-20 17 53 56

Issue #6139, opened 2019-04-03
Presets: Separate terms into synonyms and keywords


westnordost:
I recently had a closer look at the data in `/data/presets/` and I must say, it's pretty awesome. Such a wealth of information and so well structured, great job guys! I hope this would be used more in other OSM projects. (As a matter of fact, I am currently working on a (Java-)library that should serve as a tags<->names dictionary which will exclusively scrape it's data from here)

So, to the point, I have a suggestion that could improve searchability of the presets even more. Currently, there is the `name` and there are the `terms`. I suggest to split up `terms` into "further names" (synonyms) and keywords.

### Advantages I see
1. search matches with synonyms can get a higher priority than for keywords - possibly just below `name` and even before `suggestions` (brand names)
2. for items for which the search matches a synonym, that synonym could be displayed in the result list instead of the primary translation/most-well-known one (which might, depending on the area, not actually be the most-well-known one). Currently, only that translation is shown (Spielbank is a synonym for Kasino in German):  ![spielbank](https://user-images.githubusercontent.com/4661658/55515764-a6b6d480-566b-11e9-874b-b13ddf9fd7fe.png) 
The current behavior is understandable because iD can't know whether a term in `terms` is actually a synonym or just a keyword (and a keyword shouldn't be displayed in the result list). 

### Dealing with current translations
An unproblematic migration in respect to the current translations could work like this: A new translatable field named `synonyms` is introduced (or `name` field being made into `names`) and all translations in `terms` (perhaps renamed to `keywords` for clarity) just stay where they are. The behavior does not change. Then, translators can gradually move terms that are actually synonyms into the other field to profit from the features the separation brings. 

What do you think?

BjornRasmussen:
That would be great step towards improving the searchability of presets. +1

quincylvania:
> I recently had a closer look at the data in /data/presets/ and I must say, it's pretty awesome. Such a wealth of information and so well structured, great job guys!

@westnordost On behalf of the project, thank you! Our presets are especially great because so many individuals have contributed to them. A new preset makes for an excellent first PR.

> I hope this would be used more in other OSM projects. (As a matter of fact, I am currently working on a (Java-)library that should serve as a tags<->names dictionary which will exclusively scrape it's data from here)

Sounds terrific! Be aware that the preset schema can change fairly often to suit the needs of iD.

>  I suggest to split up terms into "further names" (synonyms) and keywords.

[...]

PR #6412, opened 2019-05-22
Add preset aliases property [WIP, Feedback required]

westnordost:
Implementation for #6139

Work in progress. Basically done except for the UI.

There are a few basic questions that need to be cleared before I can continue with the implementation:
1. extending Transifex `name` key with extended descriptions for translators okay? @1ec5 had some comments on that
2. refactor of `presetCollection.search` to return a list of matches instead of simply the presets okay? This has some more implications for the UI code because the UI that works with the results of this function needs to be refactored to be able to display these (alongside "raw" presets, recents, categories, ...). Advice how to do that is very welcome.
3. reconsider #6298 in the frame of this PR? It is now a bit inconsistent that `aliases` are a string array while `terms` is a comma separated string.

More detailed comments in the code.

### DIFF
--- a/build_data.js
+++ b/build_data.js
@@ -294,7 +294,7 @@ function generatePresets(tstrings, faIcons, tnpIcons) {
         validate(file, preset, presetSchema);
 
         tstrings.presets[id] = {
-            name: preset.name,
+            name: [preset.name].concat(preset.aliases || []).join(','),
             terms: (preset.terms || []).join(',')
         };
 
@@ -351,14 +351,17 @@ function generateTranslations(fields, presets, tstrings) {
         var tags = p.tags || {};
         var keys = Object.keys(tags);
 
+        var tagsDescription = keys.map(function(k) { return k + '=' + tags[k]; }).join(', ');
+
         if (keys.length) {
-            preset['name#'] = keys.map(function(k) { return k + '=' + tags[k]; }).join(', ');
+            preset['name#'] = `Name for '${tagsDescription}'. Aliases may be added after the name, separated by commas.`;
         }
         if (p.searchable !== false) {
+            preset['terms#'] = `Related terms beside the name and aliases with which the preset for '${tagsDescription}' should be found, separated by commas. '<>' if none are eligible.`;
             if (p.terms && p.terms.length) {
-                preset['terms#'] = 'terms: ' + p.terms.join();
+                preset['terms#'] += ' English terms: ' + p.terms.join();
             }
-            preset.terms = '<translate with synonyms or related terms for \'' + preset.name + '\', separated by commas>';
+            preset.terms = '<>';
         } else {
             delete preset.terms;
         }
@@ -652,8 +655,28 @@ function writeEnJson(tstrings) {
         var imagery = YAML.load(data[1]);
         var community = YAML.load(data[2]);
 
+        function transifexToPreset(transifex) {
+            var names = transifex.name.trim().split(/\s*,+\s*/);
+            var preset = {};
+            preset.name = names[0];
+            if(names.length > 1) preset.aliases = names.slice(1);
+            // terms stay Transifex'ied (comma separated string)
+            if(transifex.terms.length) preset.terms = transifex.terms;
+            return preset;
+        }
+
+        var tstringPresets = {};
+        for (const id of Object.keys(tstrings.presets)) {
+            tstringPresets[id] = transifexToPreset(tstrings.presets[id]);
+        }
+        var enjsonPresetStrings = {
+            categories: tstrings.categories,
+            fields: tstrings.fields,
+            presets: tstringPresets
+        };
+
         var enjson = core;
-        enjson.en.presets = tstrings;
+        enjson.en.presets = enjsonPresetStrings;
         enjson.en.imagery = imagery.en.imagery;
         enjson.en.community = community.en;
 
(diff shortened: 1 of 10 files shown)


build_data.js
```diff
@@ -351,14 +351,17 @@ function generateTranslations(fields, presets, tstrings) {
...
         }
         if (p.searchable !== false) {
+            preset['terms#'] = `Related terms beside the name and aliases with which the preset for '${tagsDescription}' should be found, separated by commas. '<>' if none are eligible.`;
             if (p.terms && p.terms.length) {
-                preset['terms#'] = 'terms: ' + p.terms.join();
+                preset['terms#'] += ' English terms: ' + p.terms.join();
             }
-            preset.terms = '<translate with synonyms or related terms for \'' + preset.name + '\', separated by commas>';
+            preset.terms = '<>';
         } else {
```
westnordost: The `....#` stuff is the description that is shown in Transifex as "Developer Notes".

The description for `name` mentions the possibility to add aliases after the name.
The description for `terms` now also mentions what they are used for and how to explicitly omit them. The previous message directly in the translation was sometimes misunderstood by translators so that they literally translated the English terms.

cc @1ec5

build_data.js
```diff
@@ -652,8 +655,28 @@ function writeEnJson(tstrings) {
         var imagery = YAML.load(data[1]);
         var community = YAML.load(data[2]);
 
+        function transifexToPreset(transifex) {
+            var names = transifex.name.trim().split(/\s*,+\s*/);
+            var preset = {};
+            preset.name = names[0];
+            if(names.length > 1) preset.aliases = names.slice(1);
+            // terms stay Transifex'ied (comma separated string)
+            if(transifex.terms.length) preset.terms = transifex.terms;
```
westnordost: #6298

modules/presets/collection.js
```diff
@@ -58,26 +58,48 @@ export function presetCollection(collection) {
...
+            // take those presets from the source array for which the given function
+            // returns a match and sort the resulting list of matches
+            function take(array, fn) {
+                return utilArrayMapTruthy(array, function (val, i) {
+                    var r = fn.call(null, val, i, array);
+                    if (r) array[i] = null;
+                    return r;
+                }).sort(sortMatches);
+            }
+
```
westnordost: Besides making this `filter + map + sort` more concise in the code below, this is also a performance improvement: Presets that are added to the results are not searched again.

modules/presets/collection.js
```diff
@@ -58,26 +58,48 @@ export function presetCollection(collection) {
...
+                var bCompare = getMatchWord(b).toLowerCase();
 
-                // priority if search string matches preset name exactly - #4325
+                // priority if search string matches word exactly - #4325
                 if (value === aCompare) return -1;
                 if (value === bCompare) return 1;
 
+                // priority for smaller edit distance
+                var i = a.fuzziness - b.fuzziness;
+                if (i !== 0) return i;
```
westnordost: Added this since it is now easy to get this information

modules/presets/collection.js
```diff
@@ -93,81 +115,102 @@ export function presetCollection(collection) {
...
-                });
+            var leading_tag_values = take(searchable, function(a) {
+                var tagValues = Object.values(a.tags || []);
+                for (var i = 0; i < tagValues.length; i++) {
+                    var tagValue = tagValues[i];
+                    if (tagValue !== '*' && leading(tagValue)) {
+                        return { preset: a, tagValue: tagValue };
+                    }
+                }
+            });
```
westnordost: Both `tagValue` and `term` could as well be just booleans `isTagMatch`, `isTermMatch` or similar or as well be left out completely because it is currently not used in the UI besides sorting (which did not happen before). It is simply "free" to record this info at this point.

modules/presets/collection.js
```diff
@@ -93,81 +115,102 @@ export function presetCollection(collection) {
...
-                    return a.dist - b.dist;
-                }).map(function(a) {
-                    return a.preset;
-                });
+            var similar_suggestions = take(suggestions, function(a) {
+                var dist = utilEditDistance(value, a.originalName);
+                if (dist + Math.min(value.length - a.originalName.length, 0) < 1) {
+                    return { preset: a, fuzziness: dist };
+                }
+            });
```
westnordost: `fuzziness` is used for sorting and could be used (in the future) to display those items differently.

modules/presets/collection.js
```diff
@@ -93,81 +115,102 @@ export function presetCollection(collection) {
...
                 if (typeof geometry === 'string') {
-                    results.push(presets.fallback(geometry));
+                    results.push({ preset: presets.fallback(geometry) });
                 } else {
                     geometry.forEach(function(geom) {
-                        results.push(presets.fallback(geom));
+                        results.push({ preset: presets.fallback(geom) });
                     });
                 }
             }
```
westnordost: **Question**: If the desired `geometry` is already passed in this function as a parameter, why is the whole collection not filtered to only return presets that can be used with this geometry? Is this is a bug?

quincylvania:
@westnordost I've been thinking about this a lot in light of recent events, and unfortunately came to the conclusion that we should put features like this and #6137 on hold.

iD's interpretation of tagging via its presets is a continuous source of conflict that has grated on the project. We've started to think of ways to separate OSM tagging from iD (like #6483), and until then I'd rather not make any major preset changes.

Sorry for all the work you put into this.

Issue #5688, opened 2019-01-08
Fix tanning salon preset


boothym:
I recently added a tanning salon by typing and selecting the preset, but realised it tags the place as `shop=beauty` and `beauty=tanning`. 

[According to the wiki](https://wiki.openstreetmap.org/wiki/Tag:leisure=tanning_salon) the correct tag is `leisure=tanning_salon`, which is what I remember manually tagging a couple of years ago, and certainly makes more sense as it's not really a "shop".

Of the 851 which are tagged with the shop/beauty combo, 772 were last modified after 2016-11-15, the date iD 2.0.0 was released, which makes it seems like those have come from the iD preset.

bhousel:
[According to this other page on the wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbeauty) `shop=beauty` includes tanning salons too.

[And the talk page for `leisure=tanning_salon`](https://wiki.openstreetmap.org/wiki/Talk:Tag:leisure%3Dtanning_salon) is just people talking about how it doesn't fit under "leisure".

[And the talk page for `shop=beauty`](https://wiki.openstreetmap.org/wiki/Talk:Tag:shop%3Dbeauty) lists tanning as the first example in 2009.

Sorry but since both tags have really low numbers, and I really think it fits more in with beauty shops (like nail salons) than leisure activities (like bowling alleys and sports centers), I'm inclined to just leave it as-is.

PR #6446, opened 2019-05-28
Deprecate leisure/tanning_salon

tas50:
The correct way to tag a tanning salon is a beauty shop that offers
tanning as a service.

Signed-off-by: Tim Smith <tsmith@chef.io>

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -460,6 +460,10 @@
       "replace": {"club": "*"}
     },
     {
+      "old": {"leisure": "tanning_salon"},
+      "replace": {"shop": "beauty", "beauty": "tanning"}
+    },
+    {
       "old": {"leisure": "video_arcade"},
       "replace": {"leisure": "amusement_arcade"}
     },


1ec5:
There’s some previous discussion in #5688 that I didn’t notice until now. In osmlab/name-suggestion-index@ca9a777a522e86e3f7ecfbe7c66643471b8cbf72, I went with both `shop=beauty` `beauty=tanning` and `leisure=tanning_salon`, because [this wiki page](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dtanning_salon) makes it sound like `leisure=tanning_salon` is also intended for _outdoor_ tanning salons, which I personally would want to tag as `leisure` rather than `shop`.

I’ve never encountered an outdoor tanning salon in the U.S., but maybe those exist in other countries? 🤷‍♂ If they are a thing, then `shop=beauty` `beauty=tanning` seems like a subset of `leisure=tanning_salon`, so maybe all three tags should go on an indoor tanning salon?

bhousel:
I'm very sure I also remember a lengthy thread on the tagging mailing list several years ago about why `leisure=tanning_salon` should be deprecated.  But I can't find it now and I guess nobody ever updated the wiki.

tas50:
Any chance someone has an example of one of these outdoor tanning salons?

quincylvania:
@simonpoole I'm interpreting "tanning salon" as "a place where you can pay to receive a tan". Please explain if you think there are appreciably different types of tanning salons.

simonpoole:
They are typically self-service / automated tanning bed facilities and very very distinct from what a typical shop=beauty would be. And just because they are not common in the states is no reason to force us under your rule.

quincylvania:
@simonpoole 

> They are typically self-service / automated tanning bed facilities and very very distinct from what a typical shop=beauty would be.

Okay, I'll take your word for it. I have zero interest in debating tanning of all topics. We can reverse this for now.

> And just because they are not common in the states is no reason to force us under your rule.

My impression is that you resent a few American developers making decisions that impact OSM in any significant way. I get that. The setup is imperfect. But please understand that we are professionals. My job is to make it easy for anyone to contribute good data with iD, and for that data to be easy to consume. I'm not driven by ego or some bizarre desire to force my will on others. I'm always open to reasoned ideas from people acting in good faith.

quincylvania:
Reverted in 02a8fa626077c74e20879101a8885509a02b54b5.

PR #6455, opened 2019-05-30
New search terms for various presets 

BjornRasmussen:
With the new choose-then-draw method of adding features, it is more important than ever before to have proficient search terms for whatever people come up with typing in.  This pull request adds missing search terms to presets that have been annoyingly difficult for me to find in the past.


quincylvania:
@BjornRasmussen Awesome, thanks 👍

PR #6456, opened 2019-05-30
Typos and wording

StoyanDimitrov:


quincylvania:
Hi @StoyanDimitrov 👋

We handle translation through [Transifex](https://www.transifex.com/openstreetmap/id-editor/dashboard/), so that's where you'll need to make the changes. These are generated files, so they'll be overwritten otherwise. See more details [here](https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#translating).

Thanks for your interest in making iD better!

PR #6469, opened 2019-05-31
Change "Direction" to "Direction Affected" for direction fields on vertexes.

BjornRasmussen:
The "Direction" field in vertexes in iD such as stop signs or traffic signals is quite unclear currently, since "Direction" could either mean "Direction of traffic affected" or "Direction that the sign/signal faces".  I was confused about this when I first started mapping, so decided to just not add that information.  Many people have asked on places like the osmus slack about this as well.
This pull request changes "Direction" to "Direction affected" to make this easier for new mappers to understand without checking wiki pages.

One question: the `direction_vertex` field is, as far as I know, only used in situations where there is a "Direction **affected**", so let me know if changing the label of that particular field will cause any issues.


BjornRasmussen:
> All checks have failed
Hmm... Is there code that relies on the label being a certain value? I can't think of any way this change could have broken anything.

quincylvania:
This seems like a good idea! A few implementation changes first:

- We use Title Case for preset and field names, could you use "Direction Affected"?
- Everything under `dist` is derived data and will be overwritten if edited. You'll need to change the name in the `data/presets/fields/direction_vertex.json` file and then run `npm run build` (or we can do that for you)

Feel free to ping me on Slack if you have questions!

BjornRasmussen:
Ok, I've edited the correct files now and ran `npm run build`.
I tested it, and everything looks good now!
image

PR #6494, opened 2019-06-05
Add email and website format validation

### LINKED ISSUE #5866: Warnings for website and email address tags with invalid formats
Email addresses and URLs are easy to validate with regex. We should flag errors with `email` and `website` tag format.

See #5830 for ~~upcoming~~ implemented UI improvements that will surface issues immediately as you edit.

SilentSpike:
Closes #5866

Simple validation following the HTML5 standard for emails as we don't expect POIs to have convoluted email addresses (so there should be minimal false positives). For websites I'm only checking that the scheme is present currently. 

Only checks the `website` and `email` tags as these are what iD currently supports with fields.

I was unsure how best to handle the `reference` property for the `validationIssue` objects because both issues can be present at once on the same entity (so using one function and changing the text via variable isn't viable). Just defined two functions for now since it works, can be refactored as desired.

The main thing to get right here are the translated strings. I think the website one should probably be reworded as it can flag instances where the user sees a seemingly valid site (e.g. `www.example.com`).

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1443,10 +1443,12 @@ en:
       title: Invalid Formatting
       tip: Find features tagged with invalid value formatting
       email:
-        message: '{feature} is tagged with an invalid email address'
+        message: '{feature} has an invalid email address: {email}'
+        message_multi: '{feature} has multiple invalid email addresses: {email}'
         reference: 'Email addresses must be of the form "local-part@domain".'
       website:
-        message: '{feature} is tagged with an invalid website'
+        message: '{feature} has a website with no scheme: {site}'
+        message_multi: '{feature} has multiple websites with no scheme: {site}'
         reference: 'Websites should start with a "http" or "https" scheme.'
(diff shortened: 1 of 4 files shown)


mmd-osm:
I think the email check disregards that you can specify multiple email addresses in the value which are being separated by a semicolon: https://overpass-turbo.eu/s/JIF

SilentSpike:
@mmd-osm Good point, easy to fix 👍

quincylvania:
@SilentSpike Looks great, thanks!! 👏

FYI I'm going to tweak the message strings to make them a little.

mmd-osm:
EMail addresses can have umlauts in their domain name. This one would be flagged as invalid, although it isn't: https://www.openstreetmap.org/node/5849987202

SilentSpike:
The leading and trailing space is stripped out by iD already and not a problem.

Will need to think about how best to handle special characters 🤔

PR #6503, opened 2019-06-07
Normalized public_bookcase Editor with Wiki Page.

ToastHawaii:
Normalized public_bookcase Editor with Wiki Page and add some useful fields from my own experience in mapping public bookcases.


quincylvania:
@ToastHawaii Thanks! For your localization change to be permanent you'll have to [change the string on Transifex](https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#translating).

PR #6505, opened 2019-06-07
Fix email validation for Unicode characters

SilentSpike:
Issue raised by https://github.com/openstreetmap/iD/pull/6494#issuecomment-499620002.

I just modified the regex to instead use negated lists for characters which can only appear inside a quoted string email address (the likelihood of a POI - or anyone - using such an email is minuscule). The other option would have been to put Unicode character ranges inside the square brackets, but this way means we don't accidentally exclude any more valid characters.

Also fixes handling for list values with excess whitespace by stripping it out.

### DIFF
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -9,9 +9,9 @@ export function validationFormatting() {
         var issues = [];
 
         function isValidEmail(email) {
-            // Same regex as used by HTML5 "email" inputs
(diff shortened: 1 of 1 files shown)


SilentSpike:
For reference: https://en.wikipedia.org/wiki/Email_address#Syntax

quincylvania:
Thanks, @SilentSpike!

PR #6562, opened 2019-06-21
Rejoin way if structure choice is reverted

### LINKED ISSUE #6559: Don't split ways with the "structure" change (when drawing) if the structure isn't changed
I really love the new structure tool for drawing streams, since it lets me create tunnels very quickly, but sometimes, I accidentally press `S` switching the structure to `Tunnel`, so I switch it back.  This starts a new way, however, instead of continuing the old way, which is something I then have to go back and fix later.

jguthrie100:
Fixes #6559 
Way is rejoined back to a single way if structure change is reverted

rejoin_structure_revert

### DIFF
--- a/modules/ui/tools/structure.js
+++ b/modules/ui/tools/structure.js
@@ -15,8 +15,6 @@ import { actionJoin } from '../../actions/join';
 import { modeDrawLine } from '../../modes/draw_line';
 import { osmWay } from '../../osm/way';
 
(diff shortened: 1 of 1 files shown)


BjornRasmussen:
Nice!

quincylvania:
@jguthrie100 Looks great, thanks! Impressive to open a PR against a brand new feature 💪

PR #6564, opened 2019-06-21
Add public_bookcase type and address field

ToastHawaii:
Add public_bookcase type field based on wiki page https://wiki.openstreetmap.org/wiki/Key:public_bookcase:type and the address field for public bookcases with location=indoor.


quincylvania:
@ToastHawaii I'm okay with adding the Type field, but I'm not sure about adding all those explicit options. Most of them don't have very much usage. How about leaving them off so the suggestions will just be the most-used options from Taginfo? That way mappers can create new ones and the list can grow organically.

quincylvania:
@ToastHawaii Great, thanks! Can you also change the field name to just "Type"? This is how other type fields are named.

ToastHawaii:
Thanks for the tip. I removed te explicit options.

quincylvania:
@ToastHawaii Thanks for your patience! This will be in the next release, probably next week.

ToastHawaii:
Thanks for the good work. 
Always a pleasure.

Issue #4139, opened 2017-07-10
Adjust preset list based on connected or nearby features


bhousel:
Currently the preset list is ordered by *most recently used* followed by the *default common* presets.   We can improve the ranking of some presets based on connected or nearby features.

Examples: 

##### Connected features
* When adding a highway-highway vertex, show Traffic Signals at the top
* When adding a motorway-motorway_link vertex, show Motorway Junction at the top
* When adding a highway-footway vertex, show Pedestrian Crossing at the top
* When adding a preset to the endpoint of a residential highway, show Turning Circle at the top

##### Nearby features (also see everything in #4138)
* When looking at a bank, show ATMs at the top
* When looking at a park, show Picnic Tables at the top 
* When looking at a gas station, show Convenience Store, Compressed Air, Vacuum at the top
* When looking at a post office, show Post Box at the top

etc..

cc @1ec5

1ec5:
There’s a neverending list of presets we could helpfully suggest, so it’ll be a an area for continuous improvement – and a boon for micromapping. 😄

More connected features:

* When a vertex of Residential Street is selected, show Pedestrian Crosswalk, Traffic Calming
* When a vertex of Building is selected, show Entrance
* When adding a line connected to Motorway or Trunk Road, show Motorway Link or Trunk Link, respectively

More nearby features:

* When adding an area inside School Grounds, show School Building, Playground, Sport Pitch
* When adding a line or area inside Golf Course, show the Golf group
* When adding a line near Residential Road, show Sidewalk
* When adding a line inside Building, show Corridor
* When adding a line inside Park, show Footpath
* When adding a point inside Park, show Fountain
* When adding a point near Secondary Road etc., show Bus Stop
* When adding an area near Church, show Graveyard, School Building
* When adding an area inside Residential Area, show House, Apartments, Garage

bhousel:
Good suggestions, @1ec5. 

Also I want to add, we have traditionally avoided building this kind of OSM tagging logic into iD itself, (as you said, it's a "neverending list of presets").  So the rules for this issue and #4138 should exist in the preset definitions, not in JavaScript code.  Something like:

```js
{
    "icon": "park",
    "geometry": [
        "point",
        "area"
    ],
    "related": {
        "contains": [
            "leisure/picnic_table",
            "amenity/bench",
            "amenity/bbq",
            "amenity/fountain",
            ...
        ],
        "nearby": [
            "highway/footpath",
            ...
        ]
    },
[...]

PR #6583, opened 2019-06-26
Show presets related to nearby features in the preset browser

quincylvania:
- Use entity groups to specify which features are found in particular "zones" defined by the presence of other features
- Display nearby presets and presets related to nearby features as defaults in the preset browser (re: #4139)


quincylvania:
I restructured this PR today. It makes iD populate the default preset list with the presets of in-view features as well as presets that are often found near those. This is not only a convenience, but could inspire mappers to add features they hadn't known they could add.

The "often found nearby" presets are specified in JSON. The possible domains are legion, but I added a few basic ones to start, including:

Park:
Screen Shot 2019-07-11 at 4 14 01 PM
Airport:
Screen Shot 2019-07-11 at 4 07 43 PM
Theme park:
Screen Shot 2019-07-11 at 4 10 27 PM

PR #6587, opened 2019-06-27
Add cairn preset

SilentSpike:
Have been mapping a lot of these recently (common hiking feature here in Scotland). Seems like this is the generally accepted tagging with [much more usage](https://taginfo.openstreetmap.org/tags/man_made=cairn) than [`landmark=cairn`](https://taginfo.openstreetmap.org/tags/landmark=cairn).


bhousel:
These were merged through 2.15 branch as
9ec7c57e4 2019-06-30 Stage new cairn preset file
0e012db20 2019-06-30 Add Cairn preset

PR #6595, opened 2019-06-30
Improving the translation

Ruta89:


bhousel:
Hi, thanks for the report. Our translations are handled through [Transifex](https://www.transifex.com/openstreetmap/id-editor/dashboard/), so you should sign up for an account there to fix issues with the translations.  The contents of the `pl.json` file will be replaced by whatever is in Transifex the next time we make a release of iD.

PR #6677, opened 2019-07-24
Minor CSS fix

mapmeld:
I was testing out OSM iD V3 (awesome, by the way! 😎) and I noticed that the trash button has a border-radius on both left and right sides in right-to-left language layout.  Quick fix

Screen Shot 2019-07-24 at 3 00 44 PM


quincylvania:
@mapmeld Thanks!!

PR #6697, opened 2019-07-30
Update Clearcut preset

rivermont:
This change adds `natural=scrub` to the `man_made=clearcut` preset. As suggested [on the wiki](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dclearcut#Rendering), this will make clearcuts render on the map while providing accurate tagging.

There is still a discussion going on on the [Talk page](https://wiki.openstreetmap.org/wiki/Talk:Tag:man_made%3Dclearcut) as to whether this tag should actually be used in conjunction with `man_made=clearcut`, or whether `man_made=clearcut` should exist at all. However I think that for iD this change should be fine. Open to discussion though!

### DIFF
--- a/data/presets/presets/man_made/clearcut.json
+++ b/data/presets/presets/man_made/clearcut.json
@@ -4,7 +4,8 @@
         "area"
     ],
     "tags": {
-        "man_made": "clearcut"
+        "man_made": "clearcut",
+        "natural": "scrub"
     },
     "terms": [
         "cut",


quincylvania:
@rivermont Thanks for the thought, but we're trying not to enact any tag changes in iD that haven't reached consensus (without a very good reason). The usage of `man_made=clearcut` is relatively low and not well defined, so how about letting mappers pick additional tags on a case-by-case basis? If rendering is an issue then I'd try contacting the renderers teams themselves to get this tag on the map.

PR #6701, opened 2019-07-31
Minor keyboard accessibility improvements

Abbe98:
General/minor improvements to keyboard accessibility

 - Makes more buttons and links in the main interface accessible by using the tab key
 - Aligns some hover and focus styles


quincylvania:
Thanks @Abbe98! This looks useful. I'll keep an eye out for tabbing issues.

Issue #4990, opened 2018-04-17
Language Names not translated in post-upload-screen


manfredbrandl:
It would be nice to allow translation of "de", "en" etc.
![image](https://user-images.githubusercontent.com/5360273/38855813-e112a7aa-4224-11e8-99de-cd4b73bb4ef5.png)

bhousel:
Ah yeah, I punted on that just to get something deployed quickly.  We might actually get this automatically when we do #2457 - I haven't looked into CLDR much lately, but #4964 will start pulling pieces of it into iD as a dependency.

PR #6702, opened 2019-07-31
CLDR Languages

### LINKED ISSUE #2457: Use CLDR for language list source

quincylvania:
Closes #2457 and #4990.

This gives iD much more knowledge about languages using the [`cldr-localenames-full`](https://github.com/unicode-cldr/cldr-localenames-full) module.

- We're no longer limited to the wikipedia language list
- Language names can be translated as needed without going through Transifex
- Only translations for iD's supported languages are checked in

Screen Shot 2019-07-31 at 1 42 54 PM

Screen Shot 2019-07-31 at 1 41 41 PM

- [x] Future work: nicer names for languages in a certain script when CLDR doesn't provide a name. Luckily CLDR does provide translated script names. For example, `sr-Latn` is currently displayed as "Serbian (sr-Latn)" in English when it should be "Serbian (Latin)".


bhousel:
This looks pretty good to me!  Feel free to merge 👍

PR #6720, opened 2019-08-05
Add bar and outdoor seating to main preset field for restaurants

Adamant36:
Thought id see how this works by doing a PR to add the bar and outdoor seating presets to the main fields thing for restaurants. If it's something you guys want in the main preset field, that's fine. It at least gave me a chance to test the environment and learn how presets work if nothing else. It's pretty snazzy.


quincylvania:
@Adamant36 Thanks for taking the time to setup the development environment! Presets are a great way to get started. Let me know if you have questions.

I'm not sure we necessarily want these fields to appear by default though. A long initial list may overwhelm new mappers. Plus those fields aren't necessarily more important than those like `delivery` or `website`.

Adamant36:
Totally. I agree. Thanks. I'll let you know if I have any questions. That's fine on the presets. Cafe has an outdoor seating preset. So I thought it would be cool on restaurants also. It's not a big deal if it or bar is just part of the drop down selector with the other presets though. As I agree they aren't necessarily more important then any of the other options there. Given that, now that I have things setup I'll just close this and keep an out for other issues I can work on.

PR #6725, opened 2019-08-06
Add vending=bottle_return preset

ENT8R:
This might be useful as a preset, as, at least in Germany, every store that sells bottles also has to provide a machine to return those bottles. (There is even a law for that 😄: https://www.gesetze-im-internet.de/verpackg/__15.html)
The current usage in OpenStreetMap is actually rather low (below 100), but maybe this increases when there is a preset for this machine...

### DIFF
--- a/data/presets/presets/amenity/vending_machine/bottle_return.json
+++ b/data/presets/presets/amenity/vending_machine/bottle_return.json
@@ -1,0 +1,22 @@
+{
+    "icon": "temaki-vending_machine",
+    "fields": [
+        "vending",
+        "operator"
+    ],
+    "geometry": [
+        "point"
+    ],
+    "terms": [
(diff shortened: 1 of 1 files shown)


tas50:
This would be a very useful preset in  Oregon where every supermarket legally has a bottle deposit vending machine. They're generally tagged as recycling centers which is kind of right and kind of wrong.

quincylvania:
@ENT8R Thanks!

I'm not sure I'd personally classify these as vending machines since they receive things rather than "vend" things, but it's fine if this is the tag people are using. We can always change it later if someone comes up with a better tag.

carstenhag:
@quincylvenia hey, they are commonly called Reverse Vending Machines in the retail sector, afaik. Yeah, it's strange.

PR #6740, opened 2019-08-11
Fix "one way draw from start" bug

### LINKED ISSUE #6737: Using the validation fix "Continue drawing from start" corrupts iD V3
To reproduce:
1: Create a oneway road that ends on another road, but doesn't connect to anything at the start.
2: Select the feature, and then select the validation fix "Continue drawing from start"
3: iD is now corrupted, and these issues are printed to the console:
image

jguthrie100:
Fixes #6737 

extend_one_way


quincylvania:
@jguthrie100 Thanks for the fix! There was actually an underlying issue where `addMode` should never just be true and should sometimes be undefined, so I had to implement an alternative solution.

PR #6752, opened 2019-08-13
Stop penalising forums and mailing lists

woodpeck:
Until now, Facebook, Slack, and Reddit were always listed above forums, mailinglists, and IRC for a same-sized area. This fails to give a true picture of the OSM community in all but a small number of countries. Furthermore,
* Facebook, Slack, and Reddit are closed, proprietary platforms where you can only participate by giving up some of your privacy and we should not over-eagerly promote them;
* Facebook and Reddit are proven to harbour the most obnoxious people on the planet, and we have no control over these platforms to exclude people we don't like (whereas we have, and do exert, this control on the forums and mailing lists).

It is ok to list Facebook/Reddit/Slack if the community in a country has chosen to use them, but it is not ok to make the knee-jerk assumption that as soon as there's any Facebook/Reddit/Slack channel in one country, it should automatically be recommended above a forum or mailing list.

### DIFF
--- a/modules/ui/success.js
+++ b/modules/ui/success.js
@@ -12,16 +12,17 @@ import { utilRebind } from '../util/rebind';
 export function uiSuccess(context) {
     var MAXEVENTS = 2;
 
-    // All else being equal, rank more "social" communities higher
+    // All else being equal, rank communities higher when
+    // OSM has more influence on who's there
     // (anything not in this list receives no adjustment)
     var COMMUNITYRANK = {
         'meetup': +5,
-        'slack': +4,
-        'facebook': +3,
-        'reddit': +2,
-        'forum': -2,
-        'mailinglist': -3,
-        'irc': -4
+        'slack': -1,
+        'facebook': -3,
+        'reddit': -2,
+        'forum': +3,
+        'mailinglist': +2,
+        'irc': +1
     };
 
     var detected = utilDetect();


bhousel:
Yes this is an open issue in the community index itself, see
https://github.com/osmlab/osm-community-index/issues/114

I'd strongly prefer for local communities to control the sort order of their own entries rather than using hardcoded values in iD.  So anyway, that's the place to fix the issue, not in iD.

I'm not bothering to respond to your other points, but I had a chuckle about where you feel the "most obnoxious people on the planet" spend their time.  Frederik, you are _so consistently wrong_ on so many things that sometimes I feel we're talking about two different OSMs.

rory:
> I had a chuckle about where you feel the "most obnoxious people on the planet" spend their time

Ummm.... we're talking about reddit here.... The [OpenStreetMap reddit (/r/openstreetmap)](https://www.reddit.com/r/openstreetmap/) is pretty good and nice and helpful, I contribute often to it. I've used reddit for 10+ years, it's a great resource.

But c'mon this is the site that's full of [some of the worst of the internet](https://en.wikipedia.org/wiki/Controversial_Reddit_communities), and is referenced many times on [SPLC's guide to "Male Supremacy"](https://www.splcenter.org/fighting-hate/extremist-files/ideology/male-supremacy). Are you _really_ sure you want to defend a site which steps up to protect [GamerGate](https://www.polygon.com/2018/7/13/17568556/kotakuinaction-reddit-mod-shut-down-administrator)?

bhousel:
@rory This is only about the [OpenStreetMap reddit (/r/openstreetmap)](https://www.reddit.com/r/openstreetmap/) (or any other subreddits people might set up for local OSM talk - I don't know of any), and how they might rank relative to other things.   

The people there are mostly nice, and there are moderators (spanholz and others) who mostly keep it that way.

Please don't drag the other shitty things that people do elsewhere on Reddit into the discussion.  It's not relevant here, and I'm not "defending" it, and that's a bullshit thing of you to suggest.

> _Update_: For the record, I can totally understand why one would want to "demote" IRC & the mailing lists, that's certainly something reasonable people can disagree about, and discuss.

There is a comment in the code saying why we rank them the way we do.  @woodpeck tried to remove it with this PR and replace it with his own opinion.

👉 The point of showing the community index _in iD_ is to let mappers know that there are people around where they have edited that care about the map.  The old post-save screen had some stuff about "share your changes on social media" and put the user first.  The current post-save screen has stuff like "reach out to people around you" and puts the community first.  This was a direct response to the kinds of edits we saw people doing when the Pokemon Go craze first hit OSM.  

A potential Pokemon Go vandal or someone who just dropped in to make some play edits and evaluate the project isn't going to be convinced to care about mapping by seeing that there is a mailing list.  They might potentially be converted into a useful mapper by seeing that there is a Meetup or Reddit group and that this is a thing that actual humans can do for fun.

It's worth mentioning too that other apps which use the community-index might use it differently than iD does.  "community" is not just a list of links to show people.

Komzpa:
> ` we have no control over these platforms `

https://wiki.osmfoundation.org/wiki/FAQ#What_is_the_purpose_of_the_OpenStreetMap_Foundation.3F
```
What is the purpose of the OpenStreetMap Foundation?
The goal of the foundation is to support but not control the OpenStreetMap project.
```

joostschouppe:
Both versions clearly show what @woodpeck and @bhousel value. I don't feel comfortable with either of you being in control of what is shown worldwide. I'd like people to be directed to the channel that is most likely to help them, no further values attached. Changing the sort order on a global level does not fix that, and makes smaller countries a victim of your "geopolitics".

woodpeck:
@joostschouppe, thank you for stating that the current situation is undesirable. The issue could be somewhat mitigated by allowing the repository that collects the available channels to indicate a preferred sort order. Of course we'd have to make sure that pull requests to *that* repository are then handled by someone independent.

joostschouppe:
I believe on a local level there will be much less contention:
* in most cases the choice will be quite obvious (as there is usually one, maybe two, dominant channels)
* in most cases the usual power-to-whom-does-the-work will apply (which we usually don't find problematic)
* in most cases, in case of conflict the parties will have a more personal relationship and work things out together
It's true that when things do go wrong, it's another chance for the admins to make gutsy decisions. But at least that would be a relatively small decision affecting not too many people. And it wouldn't be as hard to find a solution that makes everybody equally unhappy.

I would prefer to drop the sorting altogether until a better solution is implemented though.

PR #6754, opened 2019-08-13
Fix generic name validation for known brands

SilentSpike:
Currently a generic name warning can be given for a known brand (see http://preview.ideditor.com/master/#background=Bing&id=n6573433998) and this fixes that.

I figure it makes sense for an entity with a regular wikidata tag too which is why I've used `hasWikidata`.


quincylvania:
@SilentSpike Thanks! Nice improvement.

PR #6782, opened 2019-08-21
feat: add Mapillary Map Features layer

kratico:
This PR adds Mapillary Map Features layer


quincylvania:
@kratico Looks like an interesting feature… Could you target the `2.15` branch instead of `master`? The `master` branch contains all the iD v3 breaking changes and won't be release for awhile yet. If we pull into `2.15` then this feature could be released much sooner. We routinely merge `2.15` into `master` so the changes will be on both branches. Thanks!

Issue #6820, opened 2019-09-01
Separate "Imagery Info / Report a Problem" into two links


jidanni:
In the Backrgound menu there is a
```
[/] Imagery Info / Report a Problem
```
link.

The user thinking that clicking "Imagery Info" will show him the date of the image,
finds instead that this is all one link. So I would separate it into two.
One for Imagery Info, one for Report a Problem.

Issue #6397, opened 2019-05-21
Be sure all shortcuts have corresponding 'longcuts' in the menus


jidanni:
The Keyboard shortcuts should all be real shortcuts.
I.e., they all should have corresponding "longcuts".
i.e., they should all be items already available in various menus, not standalone.
Not only would users be able to find them easier when looking among related items in the menus,
but that would make them available in cases when the shortcut was unusable (#5306).

bhousel:
I agree that we should not over-rely on keyboard entry, for eventual touch/mobile device support.
We'll think about how to best do this over the next few months as we plan out iD v3.

quincylvania:
I added toggle buttons for all the panels that were shortcut-only: Background, Location, History, Measurement. I also did #2508, so copy and paste of features can now be performed without a keyboard.

While there are still some shortcuts that don't correspond directly to UI controls (zoom in/out by a lot, nudge features with arrow keys, select first/last/next vertex in a way), these are all things users can essentially accomplish via other interactions.

We should continue to pursue feature parity no matter what combination of input devices the mapper is using. See also #7713.

l29ah:
It seems to me like the `Find features matching search text` thing is now shortcut-only, and i couldn't even trigger it with a shortcut: Ctrl-F triggers Firefox's text searching box, not iD's.

PR #6839, opened 2019-09-09
Add button to toggle the Background info panel

quincylvania:
This reorganizes the detailed background buttons for #6397 and #6820.

Screen Shot 2019-09-09 at 12 25 31 PM

I'm opening this PR since I know exposing advanced UI like this has been controversial in the past. I for one think it's worth making this panel more discoverable and mobile-friendly.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -562,7 +562,7 @@ en:
     switch: Switch back to this background
     custom: Custom
     overlays: Overlays
-    imagery_source_faq: Imagery Info / Report a Problem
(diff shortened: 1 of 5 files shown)


bhousel:
I like this 👍   I've always thought we should have some way (aside from a secret keyboard combination) of toggling the panels.  Especially for mobile users.

Checkboxes like this are ok, but a better solution might eventually be to have a dedicated toolbar with icons someplace that a user can get to.  Maybe under a hamburger menu or something.

PR #6841, opened 2019-09-09
add uniq num to end of old modes; re-render modes after from external.

### LINKED ISSUE #6840: fix external presets/modes race condition
I was testing out maprules/iD integration at SOTMUS, and the 2.15 release introduced an unexpected race condition between the http request for external presets and the modes toolbar rendering.

When `fromExternal` (what gets the from the maprules service) gets called in the context it whacks the contents of the old presets variable and sticks all the presets in an `externalPresets` variable. However, the modes ui gets rendered in reference to that old presets variable.

The fix I'll put together is to add an event after we go and get external presets to tell the modes class re-render the modes toolbar. 

I mentioned it briefly to @quincylvania, but there's an id function used when we bind the tools configs to their html on the top toolbar such just updating the modes and re-rendering doesn't actually update the bound data. So, I introduced a small update to the modes module to add a hash to the end of the modes.id so it is distinct and will re-render.

ref #6664

maxgrossman:
closes #6840

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -562,11 +562,12 @@ export function coreContext() {
     if (presetsParameter && presetsParameter.indexOf('://') !== -1) {
         // assume URL of external presets file
 
-        presets.fromExternal(external, function(externalPresets) {
+        presets.fromExternal(presetsParameter, function(externalPresets) {
             context.presets = function() { return externalPresets; }; // default + external presets...
             osmSetAreaKeys(presets.areaKeys());
             osmSetPointTags(presets.pointTags());
             osmSetVertexTags(presets.vertexTags());
+            context.presets().call('newPresets');
         });
     } else {
         var addablePresetIDs;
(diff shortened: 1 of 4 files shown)


quincylvania:
@maxgrossman Thanks! One question… does MapRules actually use the toolbar in `2.15`? I assumed this was a v3 issue since then we show the custom presets in the toolbar.

maxgrossman:
Hey sorry for delay. 

I don't think it is so much that maprules uses the toolbar, but rather that the toolbar uses the `context.presets()` function, which maprules mutates.

It is something like this...

When iD first gets initied, `context.presets = function() { return presets }`.
If you have maprules configured, after its successful request it makes that function return a newly created variable in its callback `externalPresets`. It also flushes the initial presets variable (I think because in from external we call `presets.init()` or something.

This creates badness with the toolbar where the toolbar is initialized before this update happens and looks at the initial presets variable, which is empty. So when you interact with the toolbar it creates an NPE because the tools' attached presets' are undefined or null (can't remember which).

I guess writing that out makes me wonder if rather than overwriting the function, could we overwrite the presets with something like `presets = customPresets` in the `fromExternal` callback.

That might just fix this without even all the new updates I made to the toolbar class...

PR #6857, opened 2019-09-14
Update stile preset (again)

ewnh:
Attempt 2 at adding the stile and material fields to the stile preset, sans weird commit spam.


quincylvania:
Thanks, @ewnh!

PR #6863, opened 2019-09-16
Add support for switching levels (indoor) presets

danielsjf:
Adds:
* Escalator
* Stairs
* Elevator

All defined by the [simple indoor tagging](https://wiki.openstreetmap.org/wiki/Simple_Indoor_Tagging) scheme.


data/presets/presets/indoor/elevator.json
```diff
@@ -0,0 +1,20 @@
...
+    "tags": {
+        "indoor": "room",
+        "highway": "elevator"
+    },
+    "terms": [
+        "elevator",
+        "lift"
+    ],
+    "matchScore": 0.8,
+    "name": "Indoor elevator"
```
quincylvania: How about "Indoor Elevator Shaft"?

data/presets/presets/indoor/elevator.json
```diff
@@ -0,0 +1,20 @@
+{
+    "icon":"temaki-elevator",
+    "fields": [
+        "level",
+        "name"
```
quincylvania: I'd try `ref` instead of `name` for these since they're more likely to have reference codes than real names.

data/presets/presets/indoor/stairs.json
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    "tags": {
+        "indoor": "room",
+        "stairs": "yes"
+    },
+    "terms": [
+        "stairs"
+    ],
+    "matchScore": 0.8,
+    "name": "Indoor stairs"
```
quincylvania: "Indoor Stairwell"?

data/presets/presets/indoor/escalator.json
```diff
@@ -0,0 +1,21 @@
...
+    "tags": {
+        "indoor": "room",
+        "stairs": "yes",
+        "conveying": "yes"
+    },
+    "terms": [
+        "escalator"
+    ],
+    "matchScore": 0.8,
+    "name": "Indoor escalator"
```
quincylvania: Hmm a room dedicated to escalators doesn't seem particularly common. Can we just skip this preset and use the `conveying` field for the stairwell preset instead?

data/presets/presets/indoor/elevator.json
```diff
@@ -0,0 +1,20 @@
+{
+    "icon":"temaki-elevator",
+    "fields": [
+        "level",
```
quincylvania: You can use the `level_semi` field I added recently. This lets mappers more easily specify that a feature spans multiple floors.

data/presets/presets/indoor/elevator.json
```diff
@@ -0,0 +1,20 @@
...
+        "name"
+    ],
+    "geometry": [
+        "area"
+    ],
+    "tags": {
+        "indoor": "room",
+        "highway": "elevator"
```
quincylvania: Hmm using `highway=elevator` on areas can be complicated for routers to handle. Any chance we could use `elevator=yes` for this? This would be like how we map `highway=corridor` separately from `indoor=corridor`.

data/presets/presets/indoor/escalator.json
danielsjf: @quincylvania, I followed the Simple Indoor Tagging scheme as documented as closely as possible. I agree with some of your comments related to the exact wording. However, I'm not sure if we want to deviate from the official recommendation. There have been a few iterations on the scheme and it seems to have stabilised around this.

Link to the details on the connections between levels:
https://wiki.openstreetmap.org/wiki/Simple_Indoor_Tagging#Modelling_connections_between_different_levels

data/presets/presets/indoor/elevator.json
danielsjf: @quincylvania, I followed the Simple Indoor Tagging scheme as documented as closely as possible. I agree with some of your comments related to the exact wording. However, I'm not sure if we want to deviate from the official recommendation. There have been a few iterations on the scheme and it seems to have stabilised around this.

Link to the details on the connections between levels:
https://wiki.openstreetmap.org/wiki/Simple_Indoor_Tagging#Modelling_connections_between_different_levels

danielsjf:
@quincylvania, I agree with most changes and I have adapted them. However, see two specific issues where I've commented. Those were aligned with the OSM documentation. Not sure if we want to deviate from the official documentation.

data/presets/presets/indoor/escalator.json
quincylvania: I think the tags are okay but adding a preset for it seems potentially confusing (indoor escalator room?) and not super useful.

quincylvania:
> However, see two specific issues where I've commented. Those were aligned with the OSM documentation. Not sure if we want to deviate from the official documentation.

@danielsjf This is fair. I recommend waiting to resolve elevators and escalators and just adding the stairwell one to iD right now.

data/presets/presets/indoor/elevator.json
danielsjf: Or maybe a non-searchable tag would work for now? Then at least we'll have the symbols.

quincylvania: > Since these elevators will always be inside, they will never be connected to another way and therefore also shouldn't be a direct issue with routing.

The issue here is that elevators should be connected to other highways so routers can use them for indoor routing. This is relevant even before detailed indoor routing. For example, wheelchair users in hilly cities will often enter a building, take an elevator up, and exit on a higher level rather than climb a hill.

> Apart from this, any recommendations on how to proceed if we would actually like to modify this tagging scheme?

Looking more into this, there are only [about 380 ways](https://overpass-turbo.eu/s/MxZ) with `highway=elevator` and `indoor=room` right now. There's currently no great method of changing tagging consensus, but with so few uses perhaps this can be accomplished via the tagging mailing list…

> Or maybe a non-searchable preset would work for now? Then at least we'll have the symbols.

I don't really want iD mappers to see this preset in the wild, have it look sort of official, and wonder why they can't add it themselves.

data/presets/presets/indoor/escalator.json
quincylvania: > However, I don't have a strong opinion here so if you would like to merge both I'm fine with that.

Yeah let's just add the stairwell for now and use a field to specify if they stairs are actually escalators. Maybe later I could be convinced otherwise 😆

data/presets/presets/indoor/elevator.json
danielsjf: > The issue here is that elevators should be connected to other highways so routers can use them for indoor routing. This is relevant even before detailed indoor routing. For example, wheelchair users in hilly cities will often enter a building, take an elevator up, and exit on a higher level rather than climb a hill.

I'm a bit confused now. You're initial argument was that with this scheme, we would tag a a closed way as a highway which might confuse routers. Now you say that the elevators should be highways but in fact have to be connected to other highways? Not to say that I don't agree with your arguments about routing.

In order to facilitate the discussion on changing the indoor tagging scheme, what would you suggest to change?

quincylvania: Ah, apologies for the confusion. My contention is that `highway=elevator` should only be used on nodes along the footway network and not on ways at all. 

Screen Shot 2019-09-25 at 10 41 03 AM

I suggest switching to `elevator=yes`, which means the thing (the room) _has_ an elevator, while `highway=elevator` means it _is_ an elevator. In fact, I'm just going to merge this and make the change myself. It seems like a valid use of an existing tag.

quincylvania:
@danielsjf Thanks for working through this! I made a couple other tweaks including switching to `elevator=yes`. They will be in the next iD release, 2.15.6.

danielsjf:
Sweet, thanks a lot for merging it. I agree with the changed tags. Do you want to start an official osm procedure or do we leave it as is?

quincylvania:
@danielsjf I'm not really sure what "official" process would even be suited to this. Both tags are already in use. Though it would be nice to identify indoor data consumers and let them know which tag iD will support.

PR #6879, opened 2019-09-24
Added some fields to amenity:charging_station

wopfel:
I've added some fields to the charging station amenity, based on https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station.

I hope everything fits. I something's wrong, please let me know :-)

The types are based on /data/presets/README.md.

Kind regards,
Bernd

### DIFF
--- a/data/presets/fields/authentication/app.json
+++ b/data/presets/fields/authentication/app.json
@@ -1,0 +1,5 @@
+{
+    "key": "authentication:app",
+    "type": "check",
+    "label": "Authentication by app"
+}
--- a/data/presets/fields/authentication/none.json
+++ b/data/presets/fields/authentication/none.json
@@ -1,0 +1,5 @@
+{
+    "key": "authentication:none",
+    "type": "check",
+    "label": "No authentication required"
+}
--- a/data/presets/fields/authentication/short_message.json
+++ b/data/presets/fields/authentication/short_message.json
@@ -1,0 +1,5 @@
+{
+    "key": "authentication:short_message",
+    "type": "check",
+    "label": "Authentication by Short message (SMS)"
+}
--- a/data/presets/fields/car.json
+++ b/data/presets/fields/car.json
@@ -1,0 +1,5 @@
+{
+    "key": "car",
+    "type": "defaultCheck",
+    "label": "Electric cars"
+}
--- a/data/presets/fields/socket/chademo.json
+++ b/data/presets/fields/socket/chademo.json
@@ -1,0 +1,7 @@
+{
+    "key": "socket:chademo",
+    "type": "number",
+    "minValue": 0,
+    "label": "Chademo sockets",
+    "placeholder": "1, 2..."
+}
--- a/data/presets/fields/socket/chademo/output.json
+++ b/data/presets/fields/socket/chademo/output.json
@@ -1,0 +1,7 @@
+{
+    "key": "socket:chademo:output",
+    "type": "combo",
+    "label": "Chademo Power Output",
+    "placeholder": "11 kW, 22 kW...",
+    "options": ["11 kW", "22 kW", "43 kW", "50 kW"]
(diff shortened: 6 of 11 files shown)


bhousel:
Sorry, but I'm going to close this.. These fields just seem to esoteric and complicated for casual users to be able to be able to fill them in.

Having a bunch of checkboxes for various authentication methods..
Socket Type 2 and Socket Chadmo..

It feels like the tags weren't really thought out too well.

wopfel:
Thanks for your comment, @bhousel. I'm sorry I don't get the point. Casual users don't have to fill them in. I think users of charging stations are indeed able to fill most fields. For this I moved some fields to the 'more fields' section. 
Besides that, the tags are all described in the osm wiki and people are actually using them. 
Would you merge a smaller commit? What are the conditions for 'tags thought out well'?
I'd be happy if some extensions would be merged just for easier tagging charging stations.

bhousel:
ok I can provide some specific feedback on the tagging:

##### socket

Screenshot 2019-10-29 15 47 32

We can't support tags like this, where the user needs to supply something in the middle of the key part of the tag.  (some Seamark tags are this way too).  Really, tagging the amperage and wattage of each socket and of the overall station doesn't seem that important anyway - that would be like tagging how fast the gasoline comes out of a gasoline pump.

Tagging _what types of sockets are available_ is definitely important.  I would have gone with something more OSM-standard like a semicolon-delimited tag `charging:sockets=type2;chademo` or a namespaced tag like `charging:socket:type2=yes`/`charging:socket:chademo=yes`.  iD supports both of these by letting users pick things from a dropdown that we seed with values from taginfo.


##### authentication

Screenshot 2019-10-29 15 50 34

We also can't support a bunch of value options after a namespaced key.  We can support a multikey like `authentication:*` and the user can add which ones are supported (they will be tagged as `authentication:thing1=yes`,`authentication:thing2=yes`).  We can also support semicolon-delimited values like `authentication=thing1;thing2`.  

We don't have the bandwidth to support a tagging scheme that is really different from how most of the other tags work in OSM.  I mean - [seamark](https://wiki.openstreetmap.org/wiki/Key:seamark) is also documented in the wiki and people are using it, but that doesn't mean that it will all be supported in iD.  We support a small subset that casual users can understand.  You can still use whatever tags you want, but your users will need to use the raw tag editor to enter them.

wopfel:
Hi @bhousel, thanks for taking the time and explaining the circumstances.

Kind regards,
Bernd

ghost:
I think both socket and output are very important, since without the right socket, you cannot charge at all, and the time needed to charge can vary from a few minutes to an entire day according to the output. IMHO, without them, the charging_station POI is virtually useless. To ensure good quality information, we need to allow casual users to contribute seamlessly, i.e. have proper presets in ID editor.

[Another problem](https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dcharging_station#Multiple_CCS_chargers_with_various_output_on_the_same_station_?) : the current tagging instructions make two outputs for the same socket impossible to tag.

Given the difficulty of modifying ID editor, I think tagging instructions should be modified on the [wiki page](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station). (I am new to OSM so please let me know if it is not the way to go)

I am not a data expert, so I see only one solution that solves the problems mentioned :
Each tuple (socket, output, voltage, capacity) would be stored under the `config<i>` tag, where i is an arbitrary id.
Example :

```
config1:socket=css
config1:output=250
config1:voltage=400
config1:capacity=4

config2:socket=css
config2:output=100
config2:voltage=400
config2:capacity=2

config3:socket=type2
config3:output=22
config3:voltage=400
config3:capacity=2
```

Please let me know what is wrong and whether it is compatible with ID editor presets.

Best regards

PR #6907, opened 2019-10-02
Filter out generic tools in recents

### LINKED ISSUE #6888: Don't show geometry presets in recents when Geometries tool is shown
Showing these twice is redundant.

Screen Shot 2019-09-26 at 9 39 48 PM

Aundron:
Closes #6888.

When Geometries is toggled off they are visible in Recents.
Screenshot 2019-10-02 at 13 26 00

When Geometries is toggled on they are filtered out of Recents.
Screenshot 2019-10-02 at 13 26 13


quincylvania:
@Aundron Thanks!! Looks like a good solution 👍

PR #6911, opened 2019-10-03
(2.15) Allow "layer" tags alone to resolve building crossing issues

### LINKED ISSUE #5924: Drive-Through crosses Roof warning

quincylvania:
Moved from #6367 to target 2.15.

Closes #5924.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1369,9 +1369,9 @@ en:
       building-building:
         reference: "Buildings should not intersect except on different layers."
       building-highway:
-        reference: "Highways crossing buildings should use bridges, tunnels, coverings, or entrances."
+        reference: "Highways crossing buildings should use bridges, tunnels, or different layers."
       building-railway:
-        reference: "Railways crossing buildings should use bridges or tunnels."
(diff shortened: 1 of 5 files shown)


bhousel:
I still think this is actually wrong, but I'm just going to merge it anyway.

bhousel:
I merged this via the `2.x` branch, that's why GitHub didn't notice.

PR #6915, opened 2019-10-05
Custom WMS imagery

### LINKED ISSUE #4977: Custom WMS imagery

1ec5:
A custom background imagery URL template can now include WMS tokens like `{bbox}`. A new button in the Custom Background Settings dialog generates a URL template from a WMS GetCapabilities URL.

before
wms
after

The WMS GetCapabilities UI and the logic for converting the GetCapabilities document into a URL template are rather crude for now, and I’m also not wedded to the idea of a separate dialog. Should the Custom Background Settings dialog just accept the GetCapabilities URL in its main text field and asynchronously fetch the GetCapabilities XML when clicking OK? Should that dialog have a separate tab for WMS, to keep the instructions from getting unwieldy? Whatever the case, the JavaScript `prompt()` will need to be replaced by a proper custom dialog, since some WMS servers require the user to choose from among multiple named layers.

Fixes #4977.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -755,7 +755,10 @@ en:
       header: Custom Background Settings
       instructions: "Enter a tile URL template. Valid tokens are:\n   {zoom} or {z}, {x}, {y} for Z/X/Y tile scheme\n   {-y} or {ty} for flipped TMS-style Y coordinates\n   {u} for quadtile scheme\n   {switch:a,b,c} for DNS server multiplexing\n\nExample:\n{example}"
       template:
-        placeholder: Enter a url template
+        placeholder: Enter a URL template
+      generate_from_wms: WMS
+      wms:
+        instructions: "Enter a WMS server GetCapabilities URL, for example:\n\nhttps://www.example.org/geoserver/gwc/service/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=getcapabilities&TILED=true\n\nhttps://www.example.com/arcgis/services/Orthophotography/ImageServer/WMSServer?request=GetCapabilities&service=WMS"
     custom_data:
       tooltip: Edit custom data layer
       header: Custom Map Data Settings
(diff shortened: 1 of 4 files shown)


mmd-osm:
I gave it a try on https://deploy-preview-6915--ideditor.netlify.com/ (preview is pretty cool btw!) on  the first WMS service I could find on the web: [link](https://gis-bldam-brandenburg.de/ows/bodendenkmale?Service=WMS&Version=1.3.0&Request=GetCapabilities)

Unfortunately, Firefox won't let me use that site due to same origin policy - CORS header ‘Access-Control-Allow-Origin’ missing. The error popup reports a: TypeError: NetworkError when attempting to fetch resource.

1ec5:
Unlike JOSM, iD is subject to CORS and that’s going to be a challenge for any WMS workflow. One possibility would be to fall back to an upload form (or text area) where the user can provide the contents of the GetCapabilities XML file.

This PR does make it possible to take the `wms:` URL that JOSM generates and use it as a standard custom background imagery URL template, `{bbox}` and all. The WMS button doesn’t do a whole lot with the GetCapabilities file; it just uses it to detect MIME types and layer names, and potentially projections. But if we create a dedicated custom dialog for WMS, then all that can be entered manually anyways.

bhousel:
It looks like a good start, but I can't merge it as-is, so this might have to sit for a while.

1ec5:
Yeah, I’m not ready to call it done either. Say, I only went down the path of trying to support GetCapabilities URLs because I wasn’t sure if people could be comfortable deriving the WMS URL templates manually. But what if we land support for WMS URL parameters (9fbf0c30558adafcda1b3ccf5380eb11e1c81e06) first, then figure out the UI for GetCapabilities separately?

bhousel:
I ended up including a limited version of this in #7428..  

For now, iD can support custom wms templates but we will assume EPSG:3857 for the projection.  (I didn't have time to add an input field for the user to type in some different projection or validate it against the limited list of projections iD actually supports)

PR #6921, opened 2019-10-07
fix: mapillary pagination and minimum detection count

cbeddow:
Changed Mapillary API URLs to include a parameter enabling pagination to fix a bug with missing traffic signs, and also changed a parameter indicating minimum number of photos a traffic sign must be visible in before being vision as a point feature (from 1 previously to 2 now).


quincylvania:
Thanks for the fix! 🤜💥🤛

PR #6979, opened 2019-10-28
fix(mode/rotate): polygonHull sometimes returns array of length 2

### LINKED ISSUE #6977: Rotating multiple points blanks OSM data layer
Rotating a multiple selection of three points in close proximity causes the entire OSM data layer to go blank.

Steps to reproduce:

1. Go to https://preview.ideditor.com/master/#map=20.00/39.27909/-84.40139
1. Select all three picnic tables (nodes [6923427636](https://www.openstreetmap.org/node/6923427636), [6923427637](https://www.openstreetmap.org/node/6923427637), and [6923427638](https://www.openstreetmap.org/node/6923427638)).
1. Rotate this multiple selection.

Expected results: Two of the picnic tables rotate around the middle one.

Actual results: The entire OSM data layer goes blank. Sometimes undoing momentarily restores some of the data, but it quickly vanishes again when changing the selection. The data model seems to be unaffected; I’m able to save existing changes up to but not including the rotation even after things go haywire.

The same thing happens if I duplicate the three points and rotate those points, but not if I draw three new points from scratch and rotate them. So I think it may be specific to the coordinates of these points relative to each other.

Reproduces in both Safari and Firefox, in both 2.16.0 and master (as of defcbe7088bbaa165bc9cfe7c4896904c17ff7a2). In the Firefox console, I see a warning about `translate(NaN,NaN)` being [set on the `transform` attribute](https://github.com/d3/d3-selection/blob/8b260d2529bcf87a7d78a00a8395584ad0e052ff/src/selection/attr.js#L31).

hackily:
d3_polygonHull() sometimes returns an array of length 2, which d3_polygonCentroid() cannot handle.

This adds a check, and falls back to using geoVecInterp().

This fixes #6977

### DIFF
--- a/modules/modes/rotate.js
+++ b/modules/modes/rotate.js
@@ -95,7 +95,7 @@ export function modeRotate(context, entityIDs) {
             _pivot = geoVecInterp(points[0], points[1], 0.5);
         } else {
             var polygonHull = d3_polygonHull(points);
-            if(polygonHull.length === 2) {
+            if (polygonHull.length === 2) {
                 _pivot = geoVecInterp(points[0], points[1], 0.5);
             } else {
                 _pivot = d3_polygonCentroid(d3_polygonHull(points));
--- a/modules/modes/rotate.js
+++ b/modules/modes/rotate.js
@@ -69,14 +69,7 @@ export function modeRotate(context, entityIDs) {
 
             var nodes = utilGetAllNodes(entityIDs, context.graph());
             var points = nodes.map(function(n) { return projection(n.loc); });
-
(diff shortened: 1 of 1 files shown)


bhousel:
Great catch @hackily!  Thanks for the fix, and welcome!  👋

bhousel:
@hackily For fixes like this, can you target the `2.x` branch?  We're using `master` for development of iD v3, which won't be available for a while.

I ended up cherry picking the fix onto the `2.x` branch and merging, that's why GitHub doesn't show it as merged.

Issue #6985, opened 2019-10-28
Add better icon for communication tower


matkoniecz:
checked in preview - still has icon that fits best for observation tower 

https://preview.ideditor.com/master/#background=Bing&disable_features=boundaries&map=16.97/59.77751/9.06189

bhousel:
I made some new icons for these..

<img width="373" alt="Screenshot 2019-11-04 13 24 47" src="https://user-images.githubusercontent.com/38784/68146686-8d7a4d00-ff06-11e9-90ac-ca83825c3361.png">

quincylvania:
@bhousel Love these! Now mappers (myself included) can finally figure out the difference between towers and masts.

PR #6987, opened 2019-10-29
fix(icon): Use communications tower icon for communications tower feature

hackily:
As was pointed out in #6985, the icon for communications tower does not match.

This uses the icon `maki-communications-tower` for the `Communication Tower` feature.


quincylvania:
Thanks for the PR, @hackily! So we try to have different icons for masts and towers since OSM makes a distinction between them. We're already using this icon for the communication mast presets so we'll have to find or draw something more tower-y. See https://wiki.openstreetmap.org/wiki/Tag:tower:type=communication

bhousel:
Sounds good @quincylvania - I was actually drawing a new one last night (I also don’t really like the maki one).. will ping you later today

Issue #6816, opened 2019-08-29
Disable Circularize operation when way is already circular


quincylvania:
We disable the Square operation when a way "can't be made more square than it already is". We should have this check for Circularize too so users won't see an enabled button that does nothing.

PR #6993, opened 2019-10-29
Fix issue #6816

CarycaKatarzyna:
It solves issue [#6816](https://github.com/openstreetmap/iD/issues/6816).I disable circularize if it's already circular. I recognize circular object through compare distances between nodes and centroid and test if central angles are smaller than maxAngle.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -100,6 +100,7 @@ en:
       too_large: This can't be made circular because not enough of it is currently visible.
       connected_to_hidden: This can't be made circular because it is connected to a hidden feature.
       not_downloaded: This can't be made circular because parts of it have not yet been downloaded.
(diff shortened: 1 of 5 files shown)


BjornRasmussen:
Would this disable the circle tool for shapes with just 4 or five nodes?  These are _technically_ not much different from normal circular things, but look much worse.

CarycaKatarzyna:
It depends on maxAngle(it means max central angle) defined in the circular tool. Now maxAngle is equal 20°. So circularize button would be still available for shapes with 4 or 5 nodes.

bhousel:
Thanks @CarycaKatarzyna and welcome!!  👋  
This looks pretty good to me, and I love that you added tests too..

PR #7039, opened 2019-11-11
Add more attraction presets

hikemaniac:
This PR adds presets for

- attraction=kiddie_ride
- attraction=log_flume
- attraction=swing_carousel

according to the [wiki](https://wiki.openstreetmap.org/wiki/Key:attraction).


data/presets/presets/attraction/kiddie_ride.json
```diff
@@ -0,0 +1,21 @@
...
+        "payment_multi_fee",
+        "charge_fee"
+    ],
+    "moreFields": [
+        "{attraction}",
+        "max_age",
+        "min_age"
+    ],
+    "geometry": [
+        "point"
```
bhousel: Should this one also have "area" ?

data/presets/presets/attraction/log_flume.json
```diff
@@ -0,0 +1,25 @@
+{
+    "icon": "temaki-tower",
```
bhousel: We should probably pick a better icon for a log flume.. Maybe `maki-ferry` for now until we come up with a better one.

data/presets/presets/attraction/kiddie_ride.json
hikemaniac: I'd agree with the wiki here. Usually, those kiddy rides are just small horses, cars or helicopters moving to and fro. In my opinion they are too small to be mapped as areas.

bhousel: Oh ok!  Didn't realize that's what it was for.

data/presets/presets/attraction/log_flume.json
hikemaniac: Thank you for pointing that out, I just pushed a commit changing the icon. 👍

Issue #6617, opened 2019-07-04
'Use a bridge' message should provide the bridge it had in mind


jidanni:
User sees
```
Service Road crosses 食水嵙溪
￼
Connect the features
Use a bridge or tunnel
Reposition the features
Ignore this issue
Highways crossing waterways should use bridges, tunnels, or fords.
```
User clicks 'Use a bridge or tunnel' and nothing happens.
Perhaps a ready made bridge should appear where the message thinks it should.
Then the user could
- grab and adjust its shape, if necessary.
- hit Undo to make it go away.

Indeed providing the bridge would be much better than the user doing it himself, as it is not that simple as first laying down a bridge, and then laying down a road.

quincylvania:
I'm pretty sure this is similar to https://github.com/openstreetmap/iD/issues/6501#issuecomment-500100028. I think we should try it.

jidanni:
Yes!
All I know is that user A makes a stream.
And later user B makes a highway that crosses the stream.
So even before saving, at some point suddenly the editor should "burst in" and say, "would you like me to create a properly attached bridge for you? Just remember to stretch it if it doesn't fit correctly..."

quincylvania:
This was done by @CarycaKatarzyna in #7055 🎉

Issue #6501, opened 2019-06-06
Tool for changing the structure while drawing highways/railways


quincylvania:
The ribbon UI speeds up mapping quite a bit, but it is still cumbersome to add bridges and tunnels. There should be a Structure toolbar item to quickly toggle between these while drawing. For instance, the workflow could work like this:

1. The user is drawing a road that suddenly becomes a bridge
2. The user toggles the Structure toolbar item from None to Bridge
3. A second road feature with `bridge=yes` is started from the last-placed node 
4. The user draws the bridge portion
5. The user toggles from Bridge back to None
6. A third road feature without `bridge=yes` is started
7. The user keep drawing
8. When finished, all added features are selected

The could be very efficient when combined with a keyboard shortcut. Users could draw very long roads in complex environments without needing to stop, restart, or go to the inspector.

tordans:
I like better bridge-support. Its a lot of clicks ATM …

Another idea, thinking about this the other way around:

1. The stream is there already in this case
2. I draw a fully highway
3. The validator notes the highway<>stream
4. The validator has a "create bridge" which splits the highway segment (and a "create tunnel" which splits the stream)
   - The new bridge has all the tags of the highway but also the bridge tags
5. I re-adjust the nodes where the bride starts/ends

Advantage of this: I can tag the highway with all the details like name, surface, ref … – and only then split the bride. With keeps me from adding all those later. The later issue could also be solved with the "select multiple segments and overwrite key-values for all of the"-feature (couldn't find the issue right away.)

famo:
I also really would like to see some (time) improvements in this area.

Quincy's approach is more inline with the workflow of drawing a new way.

@tordans approach has the big advantage that it is also applicable for existing ways and uses the validator (which is already there).

That being said, the two approaches don't exclude each other (we can have both). However I think the latter one has the bigger usability (time saving) impact, since it applies to existing ways.

PS:
There is no tunnel below a bridge, but `covered=yes`. If we could have a "create covered" that would be a real time saver! (1-click solution since it can be directly attached in case of buildings, and man_made bridge.)

quincylvania:
I added the basic implementation of this feature. I think it has a lot of potential to speed up this type of mapping!

![structure tool demo](https://user-images.githubusercontent.com/2046746/59866220-167a5880-9359-11e9-9d53-b786d2082ce7.gif)

danieldegroot2:
> PS:
[...]

PR #7055, opened 2019-11-19
Fix issue #6617

CarycaKatarzyna:
This concerns issue [#6617](https://github.com/openstreetmap/iD/issues/6617) . When user clicks on "Use a bridge"/"Use a tunnel" it splits edge if is long and creates bridge/tunnel. Later user may correct bridge/tunnel location by manually shifting nodes.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1652,8 +1652,9 @@ en:
       upgrade_tags:
         title: Upgrade the tags
         annotation: Upgraded old tags.
-      use_bridge_or_tunnel:
-        title: Use a bridge or tunnel
+      use_bridge:
+        title: Use a bridge
+        annotation: Created bridge
       use_different_layers:
         title: Use different layers
       use_different_layers_or_levels:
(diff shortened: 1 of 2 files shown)


quincylvania:
Hi @CarycaKatarzyna! Thanks so much for this, it looks really great. I merged it as-is and then followed up with a few tweaks.

quincylvania:
Here's a screencapture. I this feature will be really helpful and will hopefully reduce instances where mappers erroneously select the connection fix since it was the easiest. 

add structure fix demo

bhousel:
So awesome 🎉 
I've wanted this for a while...

tordans:
Awesome, this is perfect for [the way I map](https://github.com/openstreetmap/iD/issues/6501#issuecomment-500100028). Thanks!

matkoniecz:
One more thank you, this is something really useful! It is nice to see iD editing getting easier.

I am now adding missing bridges/tunnels (more than 300 cases of broken data, in Rymanów and in Beijing), I was not expecting that iD has a chance to be better than JOSM for this task.

PR #7059, opened 2019-11-21
Add support for big open air chess board.

ToastHawaii:


quincylvania:
@ToastHawaii Thanks!! I changed the icon to differentiate it from other chess presets and changed the name to make it more standard in American English. You can translate at https://www.transifex.com/openstreetmap/id-editor/translate/.

PR #7075, opened 2019-11-29
Updating Chinese address scheme

koaber:
Currently in China, people are translating the generic province/city/district placeholders into country-specific schemes, which may cause some problems when mapping other countries using the same fields.

screenshot2

so I made a small fix to add Chinese-specific fields for separation of generic schemes, similar to the ones did in Japan and Vietnam. (see also [Administrative divisions of China](https://en.wikipedia.org/wiki/Administrative_divisions_of_China#Summary))

Screen Shot 2019-11-29 at 12 19 36


quincylvania:
@koaber Thanks and welcome! I don't know a lot about this so I'll trust your knowledge 😄

PR #7082, opened 2019-12-03
add tactile paving to steps

stragu:
Adding this field in the main ones, just like for street crossings, as it is very important for accessibility and safety.


quincylvania:
@stragu Thanks!

PR #7095, opened 2019-12-07
Add Osmose Q/A layer

### LINKED ISSUE #5682: Add Osmose Q/A layer
Following #5201 - we can add a Q/A layer for [Osmose](http://osmose.openstreetmap.fr/en)

Some information about the ruleset and API can be found on the [osm wiki](https://wiki.openstreetmap.org/wiki/Osmose).

The list of checks that Osmose performs is [here](https://wiki.openstreetmap.org/wiki/Osmose/issues) - we'll want to support a reasonable subset of these.

Similar to KeepRight, we can just pick a single marker style and a few colors to differentiate, rather than trying to match the literal osmose icons, which don't seem to mean anything.

screenshot 2019-01-07 12 28 39

SilentSpike:
Closes #5682

This is a functional Q/A layer with support for a reasonable number of issue types. Supporting new issues is as simple as giving them an icon entry in the `qa_errors.json` data file.

Some implementation notes:
- Osmose issues are uniquely identified by a unique `item` and `class` combination (both integer values). To identify specific issues we concatenate as a string `item-class`.
- Adding a new issue type to iD is as simple as adding an icon entry for the `item-class` in `data/qa_errors.json`.
- When requesting issues for a tile, we use the `item=` filter since Osmose tiles can return a maximum number of issues (we want to show the user as many as possible).
- All issue colors and static translation strings (besides UI headings) are pulled in from the Osmose API when the layer is first enabled. To reduce the overall data being requested, this is done using a `langs=` filter and as individual requests for each `item-class` (as Promises).
  - The API automatically falls back to English if a string is untranslated or if the requested locale doesn't exist in Osmose
- The list of elements and a dynamic details string (contains information like keys and values of the issue) associated with each issue has to be requested individually - this is done upon UI creation.
- The UI subsections (other than the issue description) are only created as needed, in order:
  - Description (always created)
  - Issue Details (the dynamic string)
  - Issue Elements (list of element links)
  - Fix Guidelines (string giving direction on how to fix the issue)
  - Common Mistakes (string giving information on traps to avoid while fixing)

Preview of the UI in action:
image


SilentSpike:
Going to have a look through the types of errors today and see about implementing support for a suitable subset.

Can also reduce the amount of information we're requesting from the API by passing in the `item=` parameter to request only the categories we're interested in (likely still need to filter out some sub-types on our end).

quincylvania:
@SilentSpike Wow, thanks for this PR! It'll make a great addition to iD's QA toolkit. I haven't had a chance to review it yet but we should try to get it merged before it lingers too long.

One important consideration is that the PR targets `master`. While I appreciate that effort, I think we'd have more success targeting `2.x`. The v3 assistant is still very much in testing and is unlikely to get merged in its current form.

SilentSpike:
>I think we'd have more success targeting 2.x

The new code is so nice though 😆 but yeah good point, I think the only real difference here is the UI stuff so should be easy enough

quincylvania:
> The new code is so nice though

@SilentSpike Hah thanks! Since you already wrote the v3 UI, we might as well keep that for the `master` branch. But we still need a `2.x` UI.

SilentSpike:
@quincylvania Trying to think of the easiest method in git to do this. I now have two local branches, one in the same state as this remote branch and once which I've rebased off of `master` and onto `2.x` 🤔

SilentSpike:
Actually looks like it's a good thing this wasn't able to be quickly merged as the Osmose API has changed and isn't currently serving element IDs (needed for highlighting and error descriptions).

Could potentially switch to using the old 2.0 API.

mmd-osm:
> Could potentially switch to using the old 2.0 API.

Why don't you discuss this with @frodrigo and @jocelynj? I think they would be happy to advise.

SilentSpike:
I've re-targeted this PR to `2.x` now and force pushed the appropriate changes. Still have the old `3.0` code locally and will see about using that once this PR is done.

After opening a discussion at the Osmose frontend repository (https://github.com/osm-fr/osmose-frontend/issues/193), I've tweaked the code to use a different API endpoint for tiles as it's slightly more convenient for iD and should be quick to serve for Osmose.

This PR is no longer feature complete as I need to implement specific issue detail fetching when the UI loads to get the associated elements for highlighting and error descriptions.

There's also potential to use the provided translation strings for titles and descriptions - although I'm not sure they're all written in a way iD would typically present information to users.

quincylvania:
@SilentSpike Thanks for your continued work on this! Let us know if you need any advice/assistance and when this is ready for review.

> There's also potential to use the provided translation strings for titles and descriptions - although I'm not sure they're all written in a way iD would typically present information to users.

Hmm we don't really have any other external translation integrations like this but it'd be worth looking into if there are are lots of strings.

frodrigo:
For now the Osmose analyzers (backend) have more than 500 strings. But I am on integrating doc to transifex. Each class of issue will have a title, a description, a how to fix, a tips on trap to avoid and example plus a contextual message with placeholder. For now only title and contextual message are present. The process of adding the others is ongoing. The text with placeholder are already replaced and translated on Osmose API, the placeholder are not available at the API level.

Osmose analyzers also run JOSM validator and reuse the JOSM translation.

The Osmose API aka Frontend aka Website also contains issues reported by other parties. We have no the hand over thus texts and availability of translations.

frodrigo:
Note this API http://osmose.openstreetmap.fr/api/0.3beta/items

SilentSpike:
@quincylvania This PR is now feature complete and ready for review. I suspect the translation strings may need some tweaking, but everything else should be good 🤞

As for using the supplied translations, some considerations:
1. If I counted right there are 85 new strings here and for comparison around 120 KeepRight strings.
2. I find that the Osmose strings seem to vary in quality:
    1. Some would definitely fit into the UI here as titles (e.g. "Deprecated tag").
    2. Some serve more as a readable identifier for the error type rather than a title (e.g. "Symbol char" which I've named "Unexpected Symbol").
    3. Some are more descriptive than a title (e.g. "Name present but missing main tag" which I've named "Unknown Named Object").

Also, may want to hold off on merging this until the Osmose 0.3 API is out of beta as presumably the endpoints will change.

SilentSpike:
Here's an updated preview showing the rendering of many more error types:

image

I should say I've just used the same colours as the Osmose website for now, but they could probably be tweaked a bit to look nicer in iD (the dark purple for example doesn't give much contrast with the icons).

frodrigo:
It is better to use "issue" than "error" on buttons.

The ignore button is not well named. It is not an ignore action, but flag the issue as false positive for the current user, but all the users. Then no more displayed.

SilentSpike:
Easily changed, those are just reused strings from the other QA services. We actually seem to interchange "error" and "issue" in a few places would be good to make those all consistent.

quincylvania:
>1. If I counted right there are 85 new strings here and for comparison around 120 KeepRight strings.
>2. I find that the Osmose strings seem to vary in quality

@SilentSpike Okay, that's not _too_ many. Let's continue with separate translations since it's easiest and allows us to use wording suited to iD mappers.

> Also, may want to hold off on merging this until the Osmose 0.3 API is out of beta as presumably the endpoints will change.

@frodrigo Thanks for your continued advice. When do you expect the v0.3 API to be stable?

frodrigo:
> @SilentSpike Okay, that's not too many. Let's continue with separate translations since it's easiest and allows us to use wording suited to iD mappers.

I am interested in improving the wording directly into Osmose. Plus it can help to avoid re-wording/translation in iD.

> @frodrigo Thanks for your continued advice. When do you expect the v0.3 API to be stable?

When it will be ready! The API 0.3 was started in summer '18, so I do not know I will be completed. Sure, it will take some more months. But if you choose to use the 0.3beta we will pay attention to it. Some other few softwares already use it.

SilentSpike:
@quincylvania Perhaps we could update and use the Osmose title strings for now since they're available - then once descriptions are available consider using those too. That removes around 40 strings from here.

In terms of implementing that here, we'd probably want to request http://osmose.openstreetmap.fr/api/0.3beta/items (1.2MB) when the layer is first enabled and cache the titles we'll be using so that they can be displayed immediately in the UI.

Can the user change locale mid-session? If not then we could only cache the titles for the `currentLocale` and make life easier.

@frodrigo This would mean moving some of the descriptive elements from Osmose title strings into subtitles, as iD would tend to have titles with only 2-4 words, for example:
1. "Missing access way to parking" -> "Missing parking access"
  msgstr: "There is no highway leading to this car park"
2. "{0} Street object {1} observed around but not associated tags" -> "Missing Street Object"
  msgstr: "{1} detected nearby, but no \"{0}\" tagging"

Is that fine as far as Osmose is concerned? I can make a PR to the backend to update all the strings for issues supported in this PR.

frodrigo:
> In terms of implementing that here, we'd probably want to request

http://osmose.openstreetmap.fr/api/0.3beta/items (1.2MB)

We can add a parameter to filter only on one language and largely reduce the size.

> This would mean moving some of the descriptive elements from Osmose title strings into subtitles, as iD would tend to have titles with only 2-4 words, for example:

The item level have short label, few words. The titles are static strings, but the subtitles are dynamics string with place holder. Subtitles are optional and cannot be switched.

Titles was build along the time by different people, and yes can be improved. But I am not sure we can reduce all titles to few words.

Title like "{0} Street object {1} observed around but not associated tags" are from MapCSS JSOM validator. May be switch can be done on Osmose side with MapCSS title.

Note. The integration of documentation aside the titles is in progress (backend is done, frontend not yet). A long description will be available and can be filled with the actual title. This documentation will be available in http://osmose.openstreetmap.fr/api/0.3beta/items probably with parameter filters to lighten the content.

SilentSpike:
@frodrigo Had a quick discussion and we agree that it'd be ideal to use the Osmose strings directly and align our interests.

In terms of the API "wants" to make this possible for iD:
1. Language filter to reduce data when requesting static strings for caching
2. Long descriptions available via API - will these be dynamic strings or also static?

In the meantime I've forked the backend and will see about updating some of the English titles

frodrigo:
Glad to hear.

1. I am working on extending http://osmose.openstreetmap.fr/api/0.3beta/items I am adding a new endpoint http://osmose.openstreetmap.fr/api/0.3beta/items/{item}/class/{class} to get only one class, maybe this one could be interesting for you. I will add a language filter, I also need it.

2. API only provide static string. But the subtile is from a place holder string, the replacement take place in the backend. It is only on subtitle.

Please wait a few days before changing titles. We are close the merge a branch changing all analyzers on this code line to add more docs: https://github.com/osm-fr/osmose-backend/commits/doc

jocelynj:
I've just pushed the doc branch to master, and updated file "po/osmose-backend.pot" with a lot of new English strings. @SilentSpike: if you were able to review some strings before we push to transifex, it would be very nice, as it would simplify the job of translators.

SilentSpike:
@jocelynj Taking a look now, I see you guys are abusing the `.pot` file a bit and using the `msgid` fallback for English translations. Does this not make life more difficult because the English strings can't be updated without creating an entirely new string to translate?

frodrigo:
@SilentSpike Yes and No. It is not abuse, but the old way to do it. The translation software are aware of this and able to follow the string changes.

SilentSpike:
Now using Osmose title strings directly and the code is all set up to easily be extended to descriptions too and make the requests with filtering.

Have to change the way dynamic details are presented in the UI (previously integrated directly into descriptions), here's how that looks at the moment:
image
image

SilentSpike:
@quincylvania This is now feature complete and ready for review.

The final UI looks something like:
image

The subsections here change depending on which strings are present for a given issue type (all have a description though - although many currently don't on the Osmose side of things).

All of the string content is now provided by the Osmose API as well as the icon colours. So to add new issue types it's literally now just a case of adding an icon entry into `qa_errors.json`.

Some errors have an associated dynamic string with details (like which key/value is causing the error) which will always be in English until the `langs=` parameter is added to the API endpoint for those. However, this code already provides that parameter so those will automatically become translated once it's there.

SilentSpike:
Not sure what travis' issue is here, but I ran `npm test` locally and it only fails on the `iD.uiFieldWikipedia` and `iD.validaions.suspicious_name` tests which were an issue in `2.x` when I rebased onto it

bhousel:
> Not sure what travis' issue is here, but I ran `npm test` locally and it only fails on the `iD.uiFieldWikipedia` and `iD.validaions.suspicious_name` tests which were an issue in `2.x` when I rebased onto it

Yeah those are my fault - sorry!
I'm fixing them now.

modules/ui/osmose_details.js
```diff
@@ -0,0 +1,199 @@
...
+    let _error;
+    const unknown = t('inspector.unknown');
+
+
+    function issueString(d, type) {
+        if (!d) return unknown;
+
+        // Issue description supplied by Osmose
+        var s = services.osmose.getStrings(d.error_type);
+        return (type in s) ? s[type] : unknown;
```
quincylvania: If the description is missing you may as well not show the Description section instead of showing "Unknown".

scripts/build_data.js
```diff
@@ -172,6 +171,27 @@ function validate(file, instance, schema) {
 }
 
 
+function readQAErrorIcons(faIcons, tnpIcons) {
```
quincylvania: Great to see these added programmatically!

modules/services/osmose.js
```diff
@@ -0,0 +1,355 @@
+import RBush from 'rbush';
```
quincylvania: I didn't review this file closely… I'm assuming it works fine since the issues load correctly 😅

modules/svg/osmose.js
```diff
@@ -0,0 +1,265 @@
...
+            .attr('class', 'stroke');
+
+        markersEnter
+            .append('polygon')
+            .attr('fill', d => getService().getColor(d.item))
+            .call(markerPath, 'qa_error-fill');
+
+        markersEnter
+            .append('use')
+            .attr('transform', 'translate(-5.5, -21)')
```
quincylvania: The icons aren't quite centered in the pin head. This is probably a pre-existing problem with OSM and ImproveOSM markers too.

Screen Shot 2020-02-04 at 2 02 24 PM

modules/ui/commit.js
```diff
@@ -25,7 +25,11 @@ var readOnlyTags = [
     /^host$/,
     /^locale$/,
     /^warnings:/,
-    /^resolved:/
+    /^resolved:/,
+    /^closed:note$/,
```
quincylvania: Hmm can we actually move this behavior to a different PR? It's a good idea but I see it as its own new thing.

modules/ui/osmose_details.js
```diff
@@ -0,0 +1,199 @@
...
+                .text(d => d)
+                .each(function() {
+                    var link = d3_select(this);
+                    var entityID = this.textContent;
+                    var entity = context.hasEntity(entityID);
+
+                    // Add click handler
+                    link
+                        .on('mouseenter', () => {
+                            context.surface().selectAll(utilEntityOrMemberSelector([entityID], context.graph()))
```
quincylvania: For consistency, we might try using the blue highlight instead of the red for this and the other issue types but we can change them later.

data/core.yaml
```diff
@@ -824,6 +827,13 @@ en:
   cannot_zoom: "Cannot zoom out further in current mode."
   full_screen: Toggle Full Screen
   QA:
+    osmose:
+      title: Osmose Issue
+      detail_title: Issue Details
+      elems_title: Issue Elements
+      fix_title: Fix Guidelines
+      trap_title: Common Mistakes
+      translation: Translations supplied by Osmose
```
quincylvania: Might just say "Translate", and maybe hide the button unless there are untranslated strings onscreen.

Separately, we should add text to iD's documentation about where you can translate the Osmose strings.

modules/ui/osmose_editor.js
```diff
@@ -0,0 +1,170 @@
...
+
+import { uiOsmoseDetails } from './osmose_details';
+import { uiOsmoseHeader } from './osmose_header';
+import { uiQuickLinks } from './quick_links';
+import { uiTooltipHtml } from './tooltipHtml';
+
+import { utilRebind } from '../util';
+
+
+export function uiOsmoseEditor(context) {
```
quincylvania: Is there any link format we can use for a "View on Osmose" button in the footer?

data/core.yaml
```diff
@@ -630,6 +630,9 @@ en:
       improveOSM:
         tooltip: Missing data automatically detected by improveosm.org
         title: ImproveOSM Issues
+      osmose:
+        tooltip: Automatically detected map issues from osmose.openstreetmap.fr
```
quincylvania: Hmm I know this is the same structure as the KeepRight description but both could use improvement. Maybe "Data issues detected by osmose.openstreetmap.fr"?

data/qa_errors.json
```diff
@@ -1,36 +1,92 @@
 {
     "services": {
         "improveOSM": {
-            "errorTypes": {
-                "ow": {
-                    "icon": "fas-long-arrow-alt-right",
-                    "category": "routing"
```
quincylvania: I guess the categories aren't used for anything?

SilentSpike: Yeah they were just a half implemented idea of my own with the thinking that issue styling could be made more simple once Osmose support was added (because there are so many issue types). Thankfully Osmose makes that easy already so I've just removed them.

modules/ui/osmose_editor.js
SilentSpike: There's the following information page:

```
https://osmose.openstreetmap.fr/error/e2a7b8fa-3ae8-c698-e488-7363adc27577
```

Alternatively we could link to the actual map:

```
https://osmose.openstreetmap.fr/map/?item=5070&zoom=17&lat=58.2055999&lon=-6.8280015
```

modules/ui/osmose_details.js
SilentSpike: Yeah there are a few behaviours such as this which are shared between the three QA layers which should probably all be changed together (tried to avoid differing in this PR for that reason)

quincylvania: I figured. This release will be a good opportunity to suss some of these problems out.

data/core.yaml
SilentSpike: I believe there's no way for us to check whether untranslated strings are present (since the API automatically falls back to English strings and they all get cached under the current language code).

My thinking with not using a "call to action" here is that if there are issues with the strings or missing translations then a user inclined to improve them knows where they come from - and if there aren't then it's still nice to acknowledge they're supplied by a 3rd party.

quincylvania: Hmm we could fetch the English strings too and see if they're equivalent? The issue for me is that the string is sort of long and overly draws my attention. "Translations by Osmose" might be a small improvement. It'd also help to put the button under the grey box or to at least give it a greater margin.

Another issue is that I _think_ we use a verb phrase everywhere else we use the outlink icon. Something like "View translations" would keep that while being less "asky". We could also use a different icon.

modules/services/osmose.js
SilentSpike: Yeah pretty confident about everything in this file I've spent so much time looking at it 😄 I want to convert more of it to using Promises instead of callbacks (just did for the issue detail requests), but can do that separate from this PR.

modules/ui/osmose_editor.js
quincylvania: Hmm the raw page isn't pretty but the map page doesn't show the expanded issue so let's go with the first one anyway.

kymckay: Added this in af0e37f7c56cc2adfd11bcef72fce8756a43f749

frodrigo:
Just note that the API 0.3 of Osmose is now released. No change in the API, just it is now 0.3 in place of 0.3beta. The 0.3beta is still valid URL path and redirect to 0.3.

kymckay:
> Just note that the API 0.3 of Osmose is now released. No change in the API, just it is now 0.3 in place of 0.3beta. The 0.3beta is still valid URL path and redirect to 0.3.

Great to hear! I'll update the iD source to use the new path 🥳

Issue #7090, opened 2019-12-05
Ghost menu items after undoing 'create bridge'


rivermont:
iD v3 d415f36

Steps to reproduce:
1. Have highway over waterway
2. Select highway and choose 'add bridge' option in Crossing Way error box
2a. Should have three ways selected; the three sections of the road
3. <kbd>CTRL</kbd>+<kbd>Z</kbd>, don't deselect
4. Should now have menu that matches picture. iD thinks the three roads are still there and menu is for editing multiple objects

![image](https://user-images.githubusercontent.com/25139386/70199241-33c09a80-16df-11ea-91ee-17bac11035f8.png)

quincylvania:
@rivermont Thanks, this also happens on the `2.x` branch so we should try to fix it before release.

quincylvania:
Fixed for v2 in 9aa27f68c3dc7b9d6bd08bf1f3fbf58546459aa5 and for v3 in #7103.

PR #7103, opened 2019-12-10
fix issue #7090

anonomis:
An easier way to recreate the issue is by creating two ways of different type then select both and undo one of the lines. This patch removes the removed selected ids from the editor by manipulating _entityIDs in historyChange().

### DIFF
--- a/modules/ui/entity_editor.js
+++ b/modules/ui/entity_editor.js
@@ -252,7 +252,8 @@ export function uiEntityEditor(context) {
                     difference.didChange.deletion;
             if (!significant) return;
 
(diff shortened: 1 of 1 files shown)


quincylvania:
Hi @anonomis, welcome and thanks for the simple fix 👋

FYI the `2.x` branch code for this works a little differently so I had to implement a separate fix in 9aa27f68c3dc7b9d6bd08bf1f3fbf58546459aa5.

anonomis:
Hi Quincy thanks for your awesome work!

I was unsure if I was supposed to use the setter or not.

Happy to be of assistance! ☺️

Issue #6623, opened 2019-07-05
Connecting to endpoint when drawing line creates closed way


quincylvania:
Easiest to just demo:

![closed line bug](https://user-images.githubusercontent.com/2046746/60724753-3d15c300-9f05-11e9-96c1-d4d3433df175.gif)

DujaOSM:
...or a feature, even if unintended? I can see that this behavior can be useful, e.g. it allows assigning areal tags immediately, without needing to go through hoops of adding area=yes or a relation. Personally, I'm undecided.

BjornRasmussen:
Actually, it doesn't.  It just messes up the drawing of ways, but didn't convert them to areas

quincylvania:
Even if the bug is useful to someone somewhere, I'd think the vast majority of users would find this behavior unexpected almost all the time.

PR #7106, opened 2019-12-10
finish drawing closed way

CarycaKatarzyna:
Closes [#6623](https://github.com/openstreetmap/iD/issues/6623) . Finish drawing closed way by click on first node.

### DIFF
--- a/modules/behavior/draw_way.js
+++ b/modules/behavior/draw_way.js
@@ -359,6 +359,12 @@ export function behaviorDrawWay(context, wayID, index, mode, startGraph, baselin
         if ((d && d.properties && d.properties.nope) || context.surface().classed('nope')) {
             return;   // can't click here
         }
+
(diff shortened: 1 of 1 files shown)


quincylvania:
Hi @CarycaKatarzyna, thanks for looking at this! However, we don't actually want to end drawing upon clicking the endpoint. We want to continue drawing without keeping the way circular. I've actually been looking into this and the solution is sort of delicate so I'm going to self-assign it if you don't mind. Feel free to comment on issues you're working on so we can work out this stuff ahead of time 😄

CarycaKatarzyna:
Thank you for the review, @quincylvania ! Finish drawing was my first idea for expected behavior and I haven't thought about different approach. I'm sorry for the mistake. Ok, take this issue if you say so. I will try to solve something else

Issue #4234, opened 2017-08-16
Add keyboard shortcut for Upload button


1ec5:
Adding a keyboard shortcut for the Upload button in the Save panel would make it more convenient to save a changeset. Over time, as additional fields are added to the panel, the shortcut would become more important for making the Upload button easily accessible.

Personally as a Mac user, I’m a fan of ⌘↩ as a shortcut for the default action in any modal. But I don’t know whether <kbd>Ctrl</kbd><kbd>Enter</kbd> is similarly intuitive for Windows and Linux users. Another option would be to add `accesskey="u"` or somesuch, which could be more discoverable.

bhousel:
I thought about this more and I don't think we should do it.  I think the risk is too great that a user who is typing stuff in the comment box might accidentally trigger the keyboard shortcut that uploads before they are ready to do it.

Besides, a keyboard shortcut for the upload button doesn't really save any time, since the user will just be sitting there for some seconds waiting for the upload to finish.

pnorman:
> Personally as a Mac user, I’m a fan of ⌘↩ as a shortcut for the default action in any modal. But I don’t know whether CtrlEnter is similarly intuitive for Windows and Linux users

Just to note, the convention from many sites is enter sends, and shift-enter gives a newline.

PR #7108, opened 2019-12-11
Submit changesets on Ctrl+Enter

BANOnotIT:


bhousel:
oh we discussed this in #4234

I'm someone who _regularly_ submits stuff before its ready because of this keyboard shortcut, (unfinished tweets on twitter, unfinished emails in gmail, unfinished comments on github), so I'm definitely in the anti-cmd-enter camp.

I would probably feel differently if submitting a changeset were a consequence-free thing that could easily be undone if it were done accidentally, or if people could amend their changeset comments later - but it's not like that.

Anyway, if people really want this, I won't block it, but just wanted to raise that it was a thing discussed before, and we didn't do it because reasons.

BANOnotIT:
> I would probably feel differently if submitting a changeset were a consequence-free thing that could easily be undone if it were done accidentally, or if people could amend their changeset comments later - but it's not like that.

I'm agree that hot key in such important place is quite dangerous. So I set it on ctrl+enter in favor of UX of old interface. 

If there're users that use ctrl+enter in other applications and that harms them there is some uncommon hot key: ctrl+shift+enter. I really don't know interfaces where such simple action is hidden under 3-key hot key, so it can't be misstyped I think. But such hot-key requires some labels so users can use it.

And if there's any possobility to try A/B testing for ctrl+enter it'll be vary cool! For now I can add shift+* in code and do some simple research in RU-OSM chat.

1ec5:
If we do make the shortcut more complicated, like Ctrl+Shift+Enter, perhaps pressing Ctrl+Enter could display a tip about using Ctrl+Shift+Enter for discoverability. (For reference, MediaWiki’s Visual Editor saves an edit when hitting Ctrl+Enter, so pressing Enter displays a note about Ctrl+Enter.)

As the person who originally requested #4234, I would prefer Ctrl+Enter over something more complicated. I don’t often tweet, but to me it seems like there’s a difference because you still have to type something out before saving the thing that could land you in hot water. (Maybe we should link to [RevertUI](http://revert.osmz.ru/) in the changeset confirmation dialog though…)

BANOnotIT:
> I would prefer Ctrl+Enter over something more complicated.

I'm pretty agree with the point that Ctrl+Enter has many variable meanings in many interfaces and experiance from other interfaces can play a bad joke in this case. There's no real way to revert changes in OSM, so referring to revert.osmz.ru may misinform users.

> Ctrl+Enter could display a tip about using Ctrl+Shift+Enter for discoverability.

It'll be good I think. But I don't know how to do it in this project. It may take some time for me to find how to do this. I think it can be like an "google warning" but it requires some state.

bhousel:
I've thought more about this and I really think we should not do it.  Thanks for the PR, but sorry!

PR #7126, opened 2019-12-15
increase width of input field

iriman:
It is covered by buttons (for input value 1)
image


quincylvania:
@iriman What browser are you seeing this in? Seems that different browsers have different behavior here.

quincylvania:
I just changed the input to use a constant width so hopefully we don't have to worry about it anymore.

iriman:
It's Windows 7 with classic theme and latest Firefox

PR #7147, opened 2019-12-21
Update address & phone formats in bolivia

51114u9:
Currently in Bolivia, no exist postal code. Also, I update phone format following official format


quincylvania:
@51114u9 Looks great, thanks! This will be in the next release scheduled for this week.

Issue #7040, opened 2019-11-12
Privacy Policy


bhousel:
We've been working with @kathleenlu09 from the [LWG](https://wiki.osmfoundation.org/wiki/Licensing_Working_Group) on a privacy policy for iD which will bring us in compliance with the [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation).   Thank you Kathleen for all your work on this! 🙇 

Though this work has been ongoing since September and this has been mentioned in other channels and in the [LWG minutes](https://wiki.osmfoundation.org/wiki/Licensing_Working_Group/Minutes/2019-09-12#iD_leakage_of_data_to_Facebook), we should have an issue here for visibility too.  

The key points for iD will be:
* Make the privacy policy available on GitHub
* Show a link to the notice at startup and require users to accept it
* Account for all the data that we collect about a user. While iD doesn't have a concept of "server logs", user activity does get saved as changeset tags, and this includes things like the user's name, count of changesets, walkthrough completion status, etc.
* Account for all of our third party assets that we fetch  (There is a lot!  This includes aerial imagery, streetview imagery and detections, Q/A detections, OSM note gravatars, and assets from third parties like the brand logos or wikidata commons images.  e.g. #7028, #6660, #6361, #5167, #5823, and many others).
* Making sure that all of these assets can be opted out of (imagery already can, not the other things).

This is also separate from #5017, which is about making any GDPR related changes for the OSM API.  Also worth mentioning that this privacy policy wouldn't cover any forks or other instances of iD  that handle data differently - such as [RapiD](https://mapwith.ai/) or [frodrigo/iD](https://github.com/frodrigo/iD/).  Those organizations will need their own privacy policy.

Nakaner:
Where can I find the link to the privacy policy in the user interface?

I don't think that this issue should be closed by commit ada4fb5814a51c3e49e451149a54e1469b6c971c. Leaving the question about the usage of resources from graph.facebook.com aside, the privacy policy is still not sufficient. I compared it to my checklist and was not able to tick all the boxes:

* [ ] name and contact details of the responsible person (this can be a company) [1],
* [ ] if the responsible entity has a data protection officer (it might be required), his/her name and contact details [1],
* [x] the purposes of the processing for which the personal data are intended
* [ ] the legal basis for the processing [2],
* [x] if the processing is based on point (f) of Article 6(1), the legitimate interests pursued by the controller or by a third party,
[...]

Issue #7028, opened 2019-11-07
Brand tag linking to non existend logo picture


fnordson:
brand=Rewe
brand:wikidata=Q16968817
brand:wikipedia=en:REWE

shows a broken image link: 

![rewe](https://user-images.githubusercontent.com/5624605/68403708-ab051c00-017d-11ea-838f-e42fa2cc519b.png)

File seems to link to: https://graph.facebook.com/Rewe/picture?type=large
The logo is also available here: https://de.wikipedia.org/wiki/Rewe_(Markt)

scaidermern:
@developers:
Is it really necessary to tell Facebook about an OSM user currently editing a supermarket? Can you please remove this nonsense? Either host these images on your server or just use a generic icon. But stop linking to Facebook and similar organizations. Respect our privacy, please.

bhousel:
I've seen this reported a few times.. What's probably happening is either your browser or a plugin is blocking the call to facebook to get the image.

We are going to:
* add an `onError` handler to show the fallback preset image
* add an option so that users can disable the brand icons

@scaidermern Please be nice.  Features that give users feedback and improve the quality of the data in OSM are not "nonsense".  Maybe you should have more respect for the work that the maintainers have put into making iD a good editor.

scaidermern:
> @scaidermern Please be nice. Features that give users feedback and improve the quality of the data in OSM are not "nonsense". Maybe you should have more respect for the work that the maintainers have put into making iD a good editor.

I didn't had the intention to insult anybody. iD is indeed a great editor and makes editing really easy for new mappers. However there is no need to link to such external resources and to undermine user's privacy. People aren't expecting automatic calls to Facebook in the background when firing up their OSM editor.

bhousel:
> However there is no need to link to such external resources and to undermine user's privacy.

How do you think the background imagery works?

scaidermern:
> > However there is no need to link to such external resources and to undermine user's privacy.
> 
> How do you think the background imagery works?

The same way. However the user is able to choose which aerial imagery to use.

fnordson:
> I've seen this reported a few times.. What's probably happening is either your browser or a plugin is blocking the call to facebook to get the image.

I agree this is probably it. 

I also agree with @scaidermern on the privacy issue especially because there are better sources and the facebook ping is really unnecessary.

bhousel:
> because there are better sources

What are the better sources of "all the logos for all the brands in the world"? 

[...]

PR #7161, opened 2019-12-24
Require opt in for third party images

rory:
iD v2.17.0 has added a way to [opt out of sending data to third party sites](https://github.com/openstreetmap/iD/blob/v2.17.0/CHANGELOG.md#tada-new-features), but the default is send data, and one must opt out. I suggest flipping that, and making such third party request opt in. This small patch implements that change (I think). Opt in, rather that opt out, is the gold standard for privacy, and we should hold ourselves to high standards with human rights issues like privacy.

There are many conversations to be had about privacy and usability, and we should continue to have those conversations. In the mean time, this quick change makes us private-by-default until we can come to agreement and decide on the best way forward.

cf.: #7028 #7040


rory:
I'm closing this and replacing it with #7162 . I think I made a PR/branch against something else, and there are hundreds of commits. PR against `master` is sensible

PR #7162, opened 2019-12-24
Make third party data require opt in

rory:
iD v2.17.0 has added a way to opt out of sending data to third party sites, but the default is send data, and one must opt out. I suggest flipping that, and making such third party request opt in. This small patch implements that change (I think). Opt in, rather that opt out, is the gold standard for privacy, and we should hold ourselves to high standards with human rights issues like privacy.

There are many conversations to be had about privacy and usability, and we should continue to have those conversations. In the mean time, this quick change makes us private-by-default until we can come to agreement and decide on the best way forward.

cf.: #7028 #7040 https://github.com/openstreetmap/openstreetmap-website/pull/2474

(whelp, I tried a PR in #7161 but I had some weird git history thing)


bhousel:
Sorry @rory but I strongly disagree.. I'm not comfortable intentionally breaking features in our software that provide value - and we have all previously agreed that letting users opt out of them is fine.  We put a lot of work and thought into the NSI feature, and while I'm not surprised, I'm seriously disappointed in you dropping your PR in here like this one day after the 2.17 release.  **Shame on you.** Please don't perform stunts like this or troll our repository in this way - it's not welcome. 

Taken to the extreme, we'd also need to opt-out the user from seeing satellite imagery, and force them to go looking through menus to figure out how to turn it on. They wouldn't even be able to use the tutorial!

Also, we can't accept PRs that touch the Privacy policy like this.

qeef:
Dear @bhousel, please avoid personal attacks. It's against [Code of Conduct](https://github.com/openstreetmap/iD/blob/2.x/CODE_OF_CONDUCT.md). Thanks.

bhousel:
> Dear @bhousel, please avoid personal attacks. It's against [Code of Conduct](https://github.com/openstreetmap/iD/blob/2.x/CODE_OF_CONDUCT.md). Thanks.

@qeef If you really feel like I've made a personal attack and want to open a report, please email the OSM-US CoC team - [details here](https://www.openstreetmap.us/2019/12/osmus-code-of-conduct-committee).

Issue #1727, opened 2013-08-23
Panning the map is jaggy and slow.


viesturz:
Steps to reproduce:
- open a editor with around 100 objects or so.
- drag the map around and let go.
- map update takes about 0.5 sec on core i7. 
- the lag gets higher as object count increases, easily getting to few seconds.
- this really disrupts workflow, making clicks unreliable and lagging visual feedback until the update is finished.

A quick profile shows this is mostly due to SVG updates being slow.
Have you considered using canvas instead? Kinect.JS?

Tested on: chrome, firefox, win8, x64.

PR #7188, opened 2019-12-27
OSM API /map call: support Overpass JSON format 

mmd-osm:
---

This pull request is about parsing performance, for context see https://github.com/openstreetmap/iD/issues/1727#issuecomment-26099201

---
The code in here adds parsing logic for an OSM/[Overpass based JSON](http://overpass-api.de/output_formats.html) format, which I implemented as part of https://github.com/zerebubuth/openstreetmap-cgimap/pull/153.

As the deployment of the JSON format to production is currently delayed due to missing support on the Rails port, the code in here is "silent" and inactive. To enable the new JSON parser, a few more changes would be needed:

*  loadFromAPI: 
    - call to `return parseXML(payload, callback, options);` needs to be replaced by  `return parseJSON(payload, callback, options); `
    - `d3_xml(url, { signal: controller.signal })` needs to be replaced by `d3_json(url, { signal: controller.signal })` 

* loadTile 
    - API endpoint expects ".json" suffix to trigger JSON output, hence `var path = '/api/0.6/map?bbox=';` needs to be replaced by `var path = '/api/0.6/map.json?bbox=';` 

(replacing would depend on some flag to allow switching between XML and JSON formats)

JSON output can be tested on the dev instance _master.apis.dev.openstreetmap.org_. Halifax in Nova Scotia has plenty of test data.

Unit tests are currently missing. I’m not sure if we can simply copy the existing XML ones and have them parse json instead. Or maybe something else?

---

Upstream dependencies:
* https://github.com/openstreetmap/openstreetmap-website/pull/2485 (for moving JSON support to production)

### DIFF
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -2,6 +2,7 @@ import _throttle from 'lodash-es/throttle';
 
 import { dispatch as d3_dispatch } from 'd3-dispatch';
 import { xml as d3_xml } from 'd3-fetch';
+import { json as d3_json } from 'd3-fetch';
 
 import osmAuth from 'osm-auth';
 import RBush from 'rbush';
@@ -89,6 +90,14 @@ function getNodes(obj) {
     return nodes;
 }
 
+function getNodesJSON(obj) {
+    var elems = obj.nodes;
+    var nodes = new Array(elems.length);
+    for (var i = 0, l = elems.length; i < l; i++) {
+        nodes[i] = 'n' + elems[i];
+    }
+    return nodes;
+}
 
 function getTags(obj) {
     var elems = obj.getElementsByTagName('tag');
@@ -116,6 +125,19 @@ function getMembers(obj) {
     return members;
 }
 
+function getMembersJSON(obj) {
+    var elems = obj.members;
+    var members = new Array(elems.length);
+    for (var i = 0, l = elems.length; i < l; i++) {
+        var attrs = elems[i];
+        members[i] = {
+            id: attrs.type[0] + attrs.ref,
+            type: attrs.type,
+            role: attrs.role
+        };
+    }
+    return members;
+}
 
 function getVisible(attrs) {
     return (!attrs.visible || attrs.visible.value !== 'false');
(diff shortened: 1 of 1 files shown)


bhousel:
Nice!  It would be great to use this someday!

> Unit tests are currently missing. I’m not sure if we can simply copy the existing XML ones and have them parse json instead. Or maybe something else?

Yes, I'd modify `loadFromAPI` somehow.. It currently just calls `parseXML`, but if it could inspect they payload, then you could test it against both XML and JSON payloads to make sure it works with both.

I'm just going to merge as-is, since it's dead code anyway..

mmd-osm:
In my loadFromAPI method, I'm deriving a `use_json` flag based on the path being passed on as paramter: `var use_json = path.includes(".json");`  and do all switching based on that flag. `loadTile` is the place to either go for a `map?bbox` or `map.json?bbox` URL. 

I'm not quite sure what's next here, i.e. if we should add some optional URL parameter to enable/disable JSON (like we do say for backgrounds), or have some try/fallback mechanism.

bhousel:
> Currently, I don't see the need to JSON enable any of the other endpoints. Let's focus on /map first. I'm not quite sure what's next here, i.e. if we should add some optional URL parameter to enable/disable JSON (like we do say for boundaries), or have some try/fallback mechanism.

Can we test this yet, maybe against dev server?

mmd-osm:
Yes, you can already test the cgimap version against https://master.apis.dev.openstreetmap.org

mmd-osm:
I've noticed @quincylvania changed some of the object lookups to also use json. Currently there's a hardcoded `visible` value in jsonparsers.{node,way,relation}:

```
visible: true,
```

Now looking at https://api.openstreetmap.org/api/0.6/node/3006673214/4.json, you'll notice that in case the object has been deleted, there will be an additional "visible" attribute.

My suggestion would be to use the `obj.visible` flag if is it part of the API payload, otherwise assume "true" as default:


```
visible: obj.visible ?? true,
```

(or something similar accomplishing the same task)

quincylvania:
> I've noticed @quincylvania changed some of the object lookups to also use json

@mmd-osm Yep! I tried to use as many of the JSON endpoints as possible

> My suggestion would be to use the obj.visible flag if is it part of the API payload, otherwise assume "true" as default

Sounds good, thanks for flagging this 👍

PR #7204, opened 2020-01-03
Add field for segregated=*

hikemaniac:
This pull request adds a field for [segregated](https://wiki.openstreetmap.org/wiki/Key%3Asegregated)=* to the `Cycle & Foot Path` preset.

<br>

By the way, I wish you a happy new year, good luck and a bright future for iD. Keep up the great work! 🎉


quincylvania:
@hikemaniac Seems like a fine field, but what do you think of using "Separation" in the label instead of "Segregation"? Segregation is a loaded term in many places.

hikemaniac:
@quincylvania You are absolutely right, sorry about that. :sweat_smile: I just adapted the name of the key without thinking of a better name for the field. I've changed it with my latest commit. 👍

quincylvania:
@hikemaniac No problem! Thanks for your PR and happy new year :)

PR #7237, opened 2020-01-09
add emergency=landing_site preset

andrewharvey:
https://wiki.openstreetmap.org/wiki/Tag:emergency=landing_site


quincylvania:
@andrewharvey Looks great, thanks!

PR #7244, opened 2020-01-15
Change page title

CarycaKatarzyna:
It changes page title in respect of selected items and map location. The title looks like one of:
1. iD - Europe
 _when one item is selected_
2. iD - Europe and 1 more
 _when many items are selected_
3. iD - (0.00,0.00)
 _no item is selected - it is only map location_

Let me know if you prefer other way of representing title.

### DIFF
--- a/modules/behavior/hash.js
+++ b/modules/behavior/hash.js
@@ -51,16 +51,18 @@ export function behaviorHash(context) {
         if (selected.length) {
             newParams.id = selected.join(',');
         }
-        updateTitle(selected);
-
+ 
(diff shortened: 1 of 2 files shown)


quincylvania:
@Lukas458 I incidentally closed this by renaming the default branch. You'll need to open another PR for this before we can move forward… apologies!

quincylvania:
@CarycaKatarzyna I incidentally closed this by renaming the default branch. You'll need to open another PR for this before we can move forward… apologies!

PR #7252, opened 2020-01-15
Added "Shared Path" localisation to en-AU

GenericError:
Shared path is the official and legal term in Australia for a path where cyclists and pedestrians can both legally travel on the same path together

### DIFF
--- a/dist/locales/en-AU.json
+++ b/dist/locales/en-AU.json
@@ -76,6 +76,9 @@
                 "healthcare/blood_donation": {
                     "name": "Blood Donor Centre"
                 },
+                "highway/cycleway/bicycle_foot": {
+                    "name": "Shared Path"
+                },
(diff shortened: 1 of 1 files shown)


mmd-osm:
Transifex would be the right place to fix translations: https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#translating

quincylvania:
@GenericError Thanks, but like @mmd-osm said, you'll have to change this in Transifex.

TLDuhamel:
I would suggest that "shared path" and "shared use path" should also be in included in the _terms_ for the Cycle & Foot Path preset. The term is definitely heavily used in NZ and AU, but are also used worldwide: https://en.wikipedia.org/wiki/Shared_use_path

quincylvania:
@TLDuhamel Sounds good, I added the terms 👍

PR #7253, opened 2020-01-16
Lower confusion in text of multipolygon_create

mikini:
Addresses the confusion raised and acknowledged by @bhousel on Transifex in comments to core string with key=help.relations.multipolygon_create (https://www.transifex.com/openstreetmap/id-editor/translate/#da/core/125401279).

* elaborate that a line as inner needs be closed
* remove wording that area as inner needs to be of "different kind" (current iD happily accepts "same kind" inner area as outer, fx. building area inside building area)
* elaborate about tagging of inner as area with distinct features (presumably what was meant by "different kind")

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1324,7 +1324,7 @@ en:
       relation_types_h: "Relation Types"
       multipolygon_h: "Multipolygons"
       multipolygon: "A *multipolygon* relation is a group of one or more *outer* features and one or more inner features. The outer features define the outer edges of the multipolygon, and the inner features define sub-areas or holes cut out from the inside of the multipolygon."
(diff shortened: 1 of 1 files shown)


quincylvania:
Thanks for the PR, @mikini. I'm thinking we don't need to be too verbose here. How about just "draw the outer edge as an area and the inner edge as a closed line or area"?

mikini:
Sorry for letting this slip. Your suggestion would definitely be more comprehensible. It doesn't emphasize the independence of the inner geometry besides its role on the multipolygon, as I attempted to address, but I agree that the discussion is probably out of scope here.

I get a lot of conflicts pulling develop into this branch attempting to commit a fix, but I've now allowed maintainer edits (sorry for neglecting that), feel free to edit as you see fit here, or I could make another PR if your prefer that.

quincylvania:
@mikini I closed this incidentally since we renamed the `master` branch to `develop`. But I'd say let's move on anyway since we're already going to be rewriting a lot of the help documentation for the v3 release. I'll keep this point in mind, thanks for bringing it up.

PR #7254, opened 2020-01-16
Modified Icon, Fixes #6986

### LINKED ISSUE #6986: Use blue for "waterway" group and waterfall to make them easier to recognize
Currently icons for both of them are not really recognizable. I see that at leas some icons are using color, but I am not sure about strategy for doing this (I checked https://github.com/openstreetmap/iD/blob/master/data/presets/README.md ).

Selection_003

abhisheknaiidu:
Fixes #6986

iD1


quincylvania:
Hi @abhisheknaiidu, your change makes every vertex icon blue. This isn't reflected in your screenshot. Instead we want just the waterway category icon to be blue. Try inspecting the folder icon element in the web inspector to see what CSS selector you need.

PR #7268, opened 2020-01-18
Add preset for amenity=exhibition_centre

hikemaniac:
With this PR a preset for `amenity=exhibition_centre` is added.

Running `npm run build` cluttered presets.json and en.json with lots of data from `name-suggestion-index` and `osm-community-index`, I hope that's okay. 😕


quincylvania:
> Running npm run build cluttered presets.json and en.json with lots of data from name-suggestion-index and osm-community-index

@hikemaniac You probably need to run `npm install` again to update these dependencies before running `npm run build`.

hikemaniac:
Oh, thank you for explaining! :blush::+1: I'll quickly rebuild it.

quincylvania:
@hikemaniac Looks great, thanks!

Adamant36:
@quincylvania can you add a preset for amenity=conference_centre also then so they aren't miss-tagged? It would also be good if this preset wasn't used for "fair." A fair is a cimpletely different thing and I don't think there's consensus on how to tag a fair grounds at this point.

hikemaniac:
Hi @Adamant36! iD already has a preset for [`amenity=conference_centre`](https://github.com/openstreetmap/iD/blob/2.x/data/presets/presets/amenity/conference_centre.json). I added 'fair' as a term, because 'trade fair' seems to be a common term for such exhibitions and 'fair' is listed as a synonym on the Merriam-Webster Dictionary for both [exhibition](https://www.merriam-webster.com/thesaurus/exhibition) and [exposition](https://www.merriam-webster.com/thesaurus/exposition). Would you still recommend that I remove it? I am not a native speaker so feedback is very welcome.

quincylvania:
I think it's fine to leave "fair" as a term. Exposition centers in cold climates often host fair-like or fair-branded events.

We also don't need to be totally pedantic with search terms, we just want the preset to be discoverable by mappers who don't know what iD calls the thing they're thinking of.

Issue #7272, opened 2020-01-20
Add a surgeon preset


HeyItsAdam:
Would be nice to have, could also have a tag for the surgeon's speciality (plastic surgeon, oral surgeon, etc.).

ferdinand0101:
I would like to suggest to expand the suggestion to all the healthcare:speciality keys

quincylvania:
Hi and welcome, @HeyItsAdam. Do you have a tag in mind for this? Healthcare tagging isn't super well defined.

maro-21:
I don't think we should have a separate preset for each of the doctors... There are a lot of specialities
If you want to add a surgeon, you can add
```
amenity=doctors
healthcare=doctor
healthcare:speciality=surgery
```
or `healthcare:speciality=plastic_surgery` or `healthcare:speciality=oral_surgery`.

Instead of adding preset(s), I suggest to edit Doctor preset and add all the popular specialities to
`terms: medic*,physician`
so that they are easier to find.

> I would like to suggest to expand the suggestion to all the healthcare:speciality keys

Yeah. It would be good this field work similar to `cuisine` field. When I type "ser" in the cuisine field I have suggestions like "dessert/brasserie/serbian/rotisserie" but when I type "surg" in the healthcare:speciality field, I don't have such suggestions. Do they work differently?

ferdinand0101:
so could it be implemented those presets for doctors and hospital presets ?

ferdinand0101:
what I have done there is added a cuisine like tag suggestion box (not sure if it works) and one preset

ferdinand0101:
note I just realized presets for hospitals aren't needed and should take care of them selves once the doctor presets exist

bhousel:
> I don't think we should have a separate preset for each of the doctors... There are a lot of specialities.

Agreed.. sorry we didn't close this sooner!

PR #7283, opened 2020-01-22
(Don't merge yet)This is a small start for heathcare:speciality pre sets

ferdinand0101:
#7272 
This is a small test for Allergology 
I'd like to know if I'm doing this correctly.
Thanks.

### DIFF
--- a/data/presets/fields/healthcare/speciality
+++ b/data/presets/fields/healthcare/speciality
@@ -1,0 +1,5 @@
+{
+    "key": "healthcare:speciality",
+    "type": "semiCombo",
+    "label": "Heathcare specialities"
+}
--- a/data/presets/presets/amenity/doctors/allergology
+++ b/data/presets/presets/amenity/doctors/allergology
@@ -28,7 +28,7 @@
     ],
     "tags": {
         "amenity": "doctors"
-        "healthcare:speciality":"allergology"
+        "healthcare:speciality": "allergology"
     },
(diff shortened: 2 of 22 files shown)


bhousel:
@ferdinand0101 Let's not do this.. It's generating a lot of notifications and I don't think we want to add a preset for every kind of medical specialty.

ferdinand0101:
okay
1. sorry about the notifications I din't think about that when pushing each one to my branch
2. I understood it from the issue that while maybe not every most should be added

ferdinand0101:
okay I'll create a new pr which should enable the field to be list but I won't do anything with the pre sets

bhousel:
> 2\. understood it from the issue that while maybe not every most should be added

I really don't think we want presets for these at all..

_I'm trying to take a hands-off approach to iD these days, and let people work through tagging issues on their own.._ but.. there are hundreds of medical specialties and it's a multi-valued tag.  We can't really add a preset for every one right?   Many doctors offices will have a handful of specialties.

bhousel:
> okay I'll create a new pr which should enable the field to be list but I won't do anything with the pre sets

Doesn't iD have this already?

I see this:

Screenshot 2020-01-23 10 25 20

ferdinand0101:
Yes it seems like it has it.
But when I was in the folder creating it it wasn't there and I had to create it but now I can also see it anyways I guess it's all done.

PR #7289, opened 2020-01-24
Add simple access to advertising

ToastHawaii:
"Who can change the content of the board?" option added based on the Wiki page.

### DIFF
--- a/data/presets/presets/advertising/billboard.json
+++ b/data/presets/presets/advertising/billboard.json
@@ -1,9 +1,15 @@
 {
     "icon": "temaki-billboard",
     "fields": [
+        "operator",
+        "lit",
+        "support",
+        "visibility"
+    ],
+    "moreFields": [
         "access_simple",
         "direction",
-        "lit"
+        "height"
     ],
     "geometry": [
         "point",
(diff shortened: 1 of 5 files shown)


quincylvania:
@ToastHawaii Hmm the meaning of this field seems ambiguous on these… I could also assume it refers to who can view or approach the feature.

ToastHawaii:
@quincylvania Thanks for pointing that out. I add the visibility tag and moved the access tag to the more field section. I think this helps to improve the clarity.

Both are part of the documentation on the wiki pages eg. for board https://wiki.openstreetmap.org/wiki/Tag%3Aadvertising%3Dboard. The access property was add for 3 years.
Based on the taginfo the access tag is used many times in comination mit advertising https://taginfo.openstreetmap.org/keys/advertising#combinations 

Do you think it needs more? Is there anything I can do to make it less ambiguous?

quincylvania:
@ToastHawaii Okay, if people are using this tag then it seems okay to have it under `moreFields`. Thanks for understanding 🤙

ToastHawaii:
@quincylvania Thanks

PR #7309, opened 2020-01-31
Fix almost junction behaviour for close endpoints

### LINKED ISSUE #7201: strange connection of two paths
I found on Osmose two close paths that need connecting with each other and used iD's "connect features" function but, instead of extend one path to the other and connect endpoints, it created new node and the paths overlap.
https://www.openstreetmap.org/edit?way=684989044#map=22/51.09896/16.95324
iD

SilentSpike:
Closes #7201 

Basically, if an intersection is found, it now checks:
1. Are any of the intersected way's endpoints within `EXTEND_DIST - NODE_JOIN_DIST` of the node being extended from
2. If there are, which one has smallest change in angle under threshold `atan(NODE_JOIN_DIST / EXTEND_DIST)`

These thresholds were arrived at by considering the edge-most cases where we'd want endpoint joining to kick in:
(the distance threshold, we want to catch an effectively parallel way that would be joined just outside of the existing node joining distance)
Parallel

(the angle threshold, a perpendicular way would already be joined to the end node if the intersection occurs at the joining distance, so we want to catch cases where the change in edge angle is less than the angle of this triangle)
Perpendicular

Some before/after applying the quick fix for different test cases:

image image

(this one is both endpoints of the same way)
image image

(both close, connects to end which creates smallest change in angle)
image image


SilentSpike:
I think I could optimise this by running the nearby endpoint detection and angle checks in the quick fix code rather than in the validation itself. Will try to adapt as such 👍

SilentSpike:
Ready for review now, due to all indentation changes in `almost_junction.js` I'll point out most significant changes on lines:
- 21-27: declaring constants which control the validation and quick fix behaviour
- 99-110: where I've changed the quick fix
- 212-246: where I've added the functions used by above
- 314: now includes id of the node before the end node - also saved into the issue `entityIds` for use in the quick fix

quincylvania:
@SilentSpike Thanks so much!! I didn't trace every line of code but it appears to work as expected. Plus the code tests are a big reassurance. Just merged it as-is but we can make changes on the main branch if needed.

PR #7344, opened 2020-02-09
Add field for depth=*

hikemaniac:
This pull request adds a field for [depth](https://wiki.openstreetmap.org/wiki/Key:depth)=*.


quincylvania:
@hikemaniac Hmm the `depth` tag seems problematic. Things like caves, lakes, and rivers have variable depths—a tag like `maxdepth` would be less ambiguous.

But the [wiki](https://wiki.openstreetmap.org/wiki/Key%3Adepth) seems to indicate that `depth` refers to the depth at which a feature is located, like for a shipwreck, not the total depth.

quincylvania:
Yeah I don't think we can support the `depth` tag as defined in this PR :/

hikemaniac:
@quincylvania I'm sorry to only get to reply this late, I was on holiday ☺️.

From doing a few Overpass queries, I got the impression that depth is used to indicate the depth of the deepest point. This can of course vary for water bodies, since water levels can rise or fall but typically the difference stays within a meter (if we're not looking at arid regions or intermittent water bodies).
The wiki is not very specific in my opinion. It states that

> Depth is a property referring to how deep a feature is according to charted depth

which could both be applied for the depth a feature is located in (e.g. a shipwreck) but also the depth e.g a lake has at its deepest point.
However, it explicitly mentions fords, rivers and canals and inland lakes and depth seems to be actively used on those according to my Overpass queries.

[Wiki permalink](https://wiki.openstreetmap.org/w/index.php?title=Key:depth&oldid=1955923)

quincylvania:
> which could both be applied for the depth a feature is located in (e.g. a shipwreck) but also the depth e.g a lake has at its deepest point.

Yeah I suppose it can read both ways and people seem to use it both ways. This should really get cleaned up. How would you tag `depth` for a basin inside a cavern? Or a cave entrance at the bottom of a lake?

I don't know if data consumers really use this tag, but it would be nice to support on fords. The depth makes a big difference for traversing.

hikemaniac:
I agree that the tag is ambiguous and in these cases it is really not clear what is meant with it. Especially since iD is supposed to be as user-friendly as possible, I believe it's best not to burden users with such interpretations.
From my intuition I would have considered `depth` to work the same as `height` and another tag would have to be used for position data, analogous to `ele`. However, until the situation is clarified, it's probably better to leave it out of iD.

Supporting this on fords might be a good idea, I think the situation is at least clear for _them_ 😆. If you want to I can open another pull request to add this.

quincylvania:
> If you want to I can open another pull request to add this.

Sure, I'd accept another PR with a Depth field for fords.

Issue #7347, opened 2020-02-11
Add spike strip preset


1ec5:
[`barrier=spikes`](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dspikes) is for a spike strip, also known as traffic spikes, tire shredders, one-way traffic treadles, stingers, stop sticks, and a tire deflation device. An icon for this preset could depict the spikes themselves and possibly a tire being punctured. The preset should only appear for a node on a way.

animesh-007:
I would like to work on it.

animesh-007:
I tried to solve the issue. Any suggestions on it?
@bhousel @quincylvania

animesh-007:
I tried to solve the issue with proper naming this time. sorry for the inconvenience.

bhousel:
done in #7354

PR #7350, opened 2020-02-12
Added spike strip preset

animesh-007:
added spike strip preset fixes issue #7347


data/presets/presets/barrier/spike.json
```diff
@@ -0,0 +1,12 @@
+{
+    "icon": "maki-fence",
+    
+    "geometry": [
+        "line"
+    ],
+    "tags": {
+        "barrier": "spike"
+    },
+    "name": "Spike",
```
1ec5: These things go by many names, but I’m unaware of it being called simply “spike”. The most common term seems to be “spike strip”.

animesh-007: Thank you, I just renamed it to "spike strip". Any other suggestions.

bhousel: ok, so here is the documentation:
https://wiki.openstreetmap.org/wiki/Tag:barrier=spikes

So: 
The tags should be `"barrier": "spikes"`.  
The filename should be `spikes.json`.
The name should be `"Spike Strip"`.  
The geometry should be `["vertex"]`.

animesh-007: Made these changes in the file.

bhousel: > Made those changes in the file.

Sorry but the file is still not named correctly or using the correct tag.  I'm going to close this PR.

PR #7362, opened 2020-02-16
Add Osmose translation to contributing guidelines

kymckay:
Wasn't entirely sure where to insert this so figured I'd open a pull request for review instead of just pushing.


quincylvania:
@SilentSpike Thanks! You can just push stuff like this. Developer documentation is important but it's not somewhere we really need to sweat the details 😅

Issue #7297, opened 2020-01-28
raw raw editor: idea of fully support to copy just keys


sun-geo:
Thanks again for the great raw raw-editor (tag-copy) function.

Recently i tried to copy just some shown keys (instead of often used well working full tag copies actions). The copy function seems working, but the transformation of the key information from the raw-raw-editor view to the raw-editor view is currently not available see movie:

![raw-editor copy key support](https://user-images.githubusercontent.com/47290585/73270571-3ae7be80-41df-11ea-902b-7b9f41740abc.gif)

bhousel:
oh yeah.. thanks for flagging this.
It probably just doesn't work with empty values.
(the parsing step expects a bunch of `k=v\n` pairs)

DujaOSM:
The "raw raw" editor erases empty tags as soon as the control loses keyboard focus. Similar thing happens when you e.g. select a new language for an object's name but do not populate the value immediately -- the control and the associated tag are lost soon after.

I've always assumed it's a feature, not a bug, although I also find it slightly annoying.

sun-geo:
> oh yeah.. thanks for flagging this.

You are welcome. I think it's nice-to-have (not urgent low priority)

cz2h:
Hi, I'd like to contribute to this feature and I would like to double check if I understand this feature properly.

We just simply want to display tags with emtpy contents right ? And there is no other specification.

quincylvania:
This was fixed in #7363 by @zengchu2 🎉

PR #7363, opened 2020-02-17
Enable using empty string as value for tags from text area

cz2h:
Attempt to address the issue described from #7297 . 
Add extra value checking to let ''  treated as non-empty strings for tags' value.

A little **concern** about the warning poped up as shown below, I am not sure if the warning is expected. 
After typing `tag1=` in the text area, the tag is udpated in the mode as expected but there pops up a warning.
openstreetmapPR1


quincylvania:
@zengchu2 Thanks for the PR and for your patience! Looks great, welcome 👋

By the way, the validation issue was separate and I've fixed it.

Issue #6601, opened 2019-07-01
Move the "Zoom to this" button to the toolbar


quincylvania:
The current button looks sort of out of place and takes up vertical height. With the contextual toolbar, we can show a Zoom To item when a feature or note is selected.

![Screen Shot 2019-07-01 at 11 23 15 AM](https://user-images.githubusercontent.com/2046746/60448102-e6ea0c80-9bf2-11e9-930d-fc14bc1cb92e.png)

quincylvania:
This seems better.

![Screen Shot 2019-07-01 at 1 08 57 PM](https://user-images.githubusercontent.com/2046746/60454283-74ccf400-9c01-11e9-9d8d-c9444d974615.png)

bhousel:
Nice 👍 I like this better too..

quincylvania:
Reflecting more on this, it's not great for a zoom control to live alongside edit controls. I'm going to try putting this in the map alongside the other zoom controls. It'd be disabled when nothing is selected. We'll have the space after #7368.

<img width="84" alt="Screen Shot 2020-02-26 at 4 21 14 PM" src="https://user-images.githubusercontent.com/2046746/75400450-26691580-58b4-11ea-8b74-e5f050603f4c.png">

PR #7391, opened 2020-02-27
Move the "zoom to this" button to the map controls

quincylvania:
Re: #6601 

Opening a PR for this in order to gather feedback. It's nice to have all the zoom controls together and the design is neater than the text button. But does it make sense to have a selection-aware control separate from the inspector?

zoom to selection map control

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -417,6 +417,7 @@ button[disabled].action:hover {
 
 .icon.light {
     color: #fff;
+    fill: currentColor;
 }
 .icon.created {
     color: #00ca07;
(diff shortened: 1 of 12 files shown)


bhousel:
Looks good 👍

tordans:
@quincylvania I just noticed the "zoom to" button/feature on a relation member and I wonder, if that is a duplicate of this feature – just for relation members. If so, maybe it could be removed in favour of this newly positioned button?

Map: https://preview.ideditor.com/master/#background=Berlin-2018&disable_features=boundaries&id=r933526&locale=en&map=20.00/52.47423/13.44983

Bildschirmfoto 2020-03-14 um 21 38 28 Kopie

quincylvania:
@tordans Nice thought, but I think the relation member button is still useful since you can use it without changing the selection. I often want to focus on a single member of a big relation, say an island in a lake, where the extent of the full relation is significantly different than the extent of the one member.

Issue #7035, opened 2019-11-10
add support for payment:others=no


Lukas458:
Hello,
I think it might be a good propose for having directly under the field where people can specify a shop's/restaurant's etc. **accepted** payment methods
a "check" field where people can set a stick for "the list of accepted payments is complete". This would then set the **payment:others=no**-tag. This is for marking that all accepted payments are listed. The tag's usage on taginfo is here: https://taginfo.openstreetmap.org/keys/payment%3Aothers#values

Lukas458:
and the wiki-related line is here: https://wiki.openstreetmap.org/wiki/Key:payment#Showing_that_the_list_of_accepted_payments_is_complete

quincylvania:
We could possibly do this but we'd need to make a custom field element.

Lukas458:
I see. This seems to be a bit difficult as I read from your post?

quincylvania:
> This seems to be a bit difficult as I read from your post?

@Lukas458 Yes, we don't have any other fields like this right now that we could copy.

bhousel:
I don't think we can do this easily with the multivalued fields.

PR #7413, opened 2020-03-04
Create payment_others_no

Elefant-aus-Wuppertal:
A first try to let iD support payment:others=no to show that the list of all payment methods accepted is complete, see https://github.com/openstreetmap/iD/issues/7035


quincylvania:
Thanks for your interest, @Lukas458, but the issue has already been closed. We really don't have a good way of supporting this type of data in iD right now :/

Elefant-aus-Wuppertal:
hmm okay, it's bad to hear that but of course, I will accept it.

PR #7428, opened 2020-03-12
Imagery index

### LINKED ISSUE #7425: Replace editor-layer-index with @ideditor/imagery-index
> We already did have an ELI issue for this (https://github.com/osmlab/editor-layer-index/issues/214, which I mention above as a blocker), and I submitted a PR to fix that and all the other issues with ELI that we have (https://github.com/osmlab/editor-layer-index/pull/490) and nobody wanted it.  This was like 2 years ago.

_Originally posted by @bhousel in https://github.com/openstreetmap/iD/issues/4994#issuecomment-582718226_

The past 2 weeks I've had some time to dive in and clean up the editor-layer-index project to make it something that works better for us.  I reviewed every single source file and rewrote all the code.

Today I released an initial version and I'm feeling good about switching iD over to use it!
The code is here:  https://github.com/ideditor/imagery-index

This builds off of the work done in that previous pull request, but adds a bunch more.

#### ✨What's new:

- Built an awesome preview site - https://ideditor.github.io/imagery-index/  (seriously, check it out)
- Separated .geojson files from imagery source .json files to promote reuse
- Leveraged [country-coder](https://github.com/ideditor/country-coder) and [location-conflation](https://github.com/ideditor/location-conflation) for defining geographic regions (so we can avoid shipping geojson for boundaries that iD already has)
- Added many geometry checks to correct geojson winding order, coordinate precision, prettify file formatting
- Fixed all the GeoJSON MultiPolygons (they were encoded as Polygons before)
- Used Javascript everywhere for ease of maintenance
- Published software to npm with semantic version number (We can load these files from a CDN now with guarantee that they will be compatible with iD.  We don't have to wait for an iD release to deliver the latest imagery to our users)
- Included backward-compatible editor-layer-index style files (`imagery.json`,`imagery.geojson`,`imagery.xml`)

### Still TODO - I'd like to:
- Restore the historic imagery and wmts sources, I just skipped these for the initial version to save time (iD does not include old imagery sources).
- Improve the `best` flag. We can think of a better way to say which sources are best where.
- Add a lot more checks to make sure that privacy policies and licenses are set for all the sources. Willing to do this work myself, but would appreciate help.  I'd like to work towards making these fields required for all sources.
- Move all the icons back into the repository so that they are saved somewhere stable, and find better quality version of them (many currently reference files that are 404 or unstable, or .ico files)
- Produce a better report about how the index differs from JOSM's index

Excited to make the switch 🍾

bhousel:
This switches the source of our imagery from ELI to imagery_index.

As followup I'd like to allow iD projects to work with alternate imagery indexes (and preset indexes).  But at least now the bulk of the work is done, to switch from "bundled" imagery (which only changes when iD releases happen) to "fetched" imagery (which can be released on different schedule from iD).

closes #7425


bhousel:
Pretty excited about this part
Screenshot 2020-03-12 17 48 54

quincylvania (approved):

quincylvania:
Due to strong push-back from some influential members of the community, I've reverted the changes from this PR for now. I'm also locking conversation in this and #7425 until the best medium for discussing a path forward is determined.

PR #7436, opened 2020-03-17
Add japanese number phone format

sguinetti:
Accourding with [Wikipedia](https://en.wikipedia.org/wiki/Telephone_numbers_in_Japan), the number format is +81-[areacode]-[number]. For example, in Tokyo is +81-3-1234-5678. This makes work easier for mappers.


bhousel:
Thanks! 👍

PR #7442, opened 2020-03-21
adding the option "internet_access:fee=customers"

Elefant-aus-Wuppertal:


Elefant-aus-Wuppertal:
internet_access:fee ist not only a boolean just "yes/no"-key, see https://taginfo.openstreetmap.org/keys/internet_access%3Afee#values and https://wiki.openstreetmap.org/wiki/Key:internet_access:fee

quincylvania:
@Lukas458 Thanks! I tweaked the labels somewhat to be a bit more self-explanatory in English.

PR #7496, opened 2020-04-01
v3 UI

### LINKED ISSUE #5701: Mapping power lines
I just added a power line.  Not my first, but maybe my longest.  Over 50 poles.  Which prompted me to submit an issue that has occurred to me every time I've mapped a power line.

By the mechanics of power lines, if the line deviates from a straight line it can **only** happen at a tower/pylon/pole.  I don't know about others, but I map a node at every tower/pylon/pole I can see (often visible by itself, sometimes apparent as a shadow) and nowhere else.  Which means, after I've drawn the line I then have to revisit every node and tag it accordingly.

Could we have an dialogue option (or other method), after completing the line, to make every node a tower/pole?  It's rather tedious and mind-numbing to do it by hand.

### LINKED ISSUE #5873: Quick feature creation from favorite presets

### LINKED ISSUE #5874: Allow locking of the draw mode

### LINKED ISSUE #5882: Allow choosing the feature type before drawing

### LINKED ISSUE #5960: Contextual top bar when drawing
_This issue is strongly related to #5882, which should be done before or with this._

The existing top bar is relatively static no matter what mode the user is in. While drawing an area it looks like this:

screen shot 2019-02-24 at 1 53 51 pm

There are several buttons we can remove during drawing to free up space:

- Remove the Save button. There's no need to upload in the middle of drawing.
- Remove the buttons for the other drawing modes. We disable clicking on them anyway during drawing.

With the extra space we can add UI elements specific to drawing:

- A Cancel or Done button to exit the mode. This would benefit everyone, but especially touchscreen users who currently have no great way to finish drawing.
- Instructions for drawing the geometry type. After #5882 this can also be specific to the preset.
- A toggle for line/orthogonal/arc drawing. Finally, a home for #2699. Ideally these would work for both lines and areas you could switch segment types in the middle of drawing a way.
- A control for locking/unlocking the draw mode (see #5874).
- After #5882 we could even consider controls specific to the feature type. Like toggling between adding poles or towers when drawing power lines (see #5701). 

screen shot 2019-02-24 at 1 55 56 pm

### LINKED ISSUE #6468: Use new preset browser for changing presets
The new toolbar preset browser can be used for adding features, but the old browser is still required for changing presets or setting a preset after drawing. This results in twice the number of UIs for a user to learn and twice the amount of development work to maintain them. We should completely replace the old browser with the new one.

The new popover browser has a number of advantages:

- Displays more results at once
- Usable anywhere, not just in a sidebar
- Can display mixed point, line, and area results

Naturally, we would hide presets with incompatible geometry.

 Rough mockup:

Screen Shot 2019-05-31 at 11 58 06 AM

### LINKED ISSUE #6501: Tool for changing the structure while drawing highways/railways

### LINKED ISSUE #6665: Improve presets=* feature (Favorites, UI)
My goal is to create a single purpose editor by changing the iD config. This editor should only allow and show what is need for the current single purpose so learning the tool is as easy as possible and users cannot break too much by accident.

As as follow up to https://github.com/openstreetmap/iD/issues/6553, I suggest …

a. Automatically make all presets of the presets-parameter (list and external file) a favorite in the new toolbar.

Example: Right now, one can create a singe purpose editor but the favorites are empty at the start, which makes it harder to spot the single purpose action. http://preview.ideditor.com/master/#background=Berlin-2019&disable_features=boundaries&locale=en&map=20.48/52.47762/13.44972&presets=amenity/bicycle_parking

b. I also suggest to hide the "Add Feature"-Feature, whenever all current presets are part of the favorites-bar. The dorpdown does not have any advantage in this situations and only makes the UI more complex.

c. I also suggest to hide the default "Favorites" "add node, way, area" whenever the presets=* feature is used, since those presets already specify the type of mapping and allow choosing the correct type (node, way…) via the preset.

### LINKED ISSUE #6717: Add UI to toggle toolbar items
The v3 contextual toolbar opens up lots of possibilities for operation discoverability and advanced tools in iD (see https://github.com/openstreetmap/iD/issues/6684#issuecomment-515455350). However, it runs the risk of becoming cluttered.

61954615-2d7b1e80-af87-11e9-8410-29dbe79f01b7

If a mapper never uses a tool like Circularize—or relies solely on the keyboard shortcut—it's a waste to always show the button.

We should add a way for users to quickly add and remove items from the toolbar. Logically, this could be a toolbar item itself.

Screen Shot 2019-08-03 at 1 20 09 PM
Screen Shot 2019-08-03 at 1 20 39 PM

This would let us keep the default toolbar stripped down while still offering tools for power mappers. For example:

- Extract cycleway or sidewalk (#6541)
- Query features at low zoom (#6662)
- Follow a path (#1614)

In addition, features that live elsewhere could have be turned into optional, redundant tools. These could replace the need to dive into menus or use keyboard shortcuts for some users.

- Switch backgrounds
- Adjust imagery offset
- Toggle minimap, streetview, info panels, etc.
- Navigate issues

As part of #6483, these toolbar items could be modular so developers could create and distribute their own tool plugins.

quincylvania:
This brings an improved version of the toolbar from from the `v3-prototype` branch to the `develop` branch.

Screen Shot 2020-04-01 at 10 15 13 AM


tahini:
We are interested by this feature. We do a lot of OpenStreetMap editing, and this divides by 2 the number of clicks per edit! We have our local instance of iD with a lot of local changes. It is based on v3-prototype which hasn't been updated since 2020 and now we would like to contribute our changes back.

We have time and hands to put on this. How should we go about working on this feature? Could we rebase/update this branch? Or try to split it in smaller chunks for easier review? Is the community interested in merging this? Will this be reviewed?

Thanks

cc @kaligrafy

danieldegroot2:
@quincylvania @tyrasd could you give some reply to the above and/or more information (to-do's) for people interested in this PR, if possible?

quincylvania:
Closing this PR as very stale.

For those interested, I never completed this design since it was a major departure from the existing workflow and would have been too disruptive. Though I think there were some cool ideas in here which I would love to see become optional tools for advanced users, and/or become rolled into a different editor.

kaligrafy:
Just a note here, my fork using this favorites menu makes my edits at least 5 to 10 times faster and I can also add keyboard shortcuts to each favorite. Worth considering...

kaligrafy:
I would suggest reopening this as this can make edits way more efficient for a lot of users, even new users.

quincylvania:
@kaligrafy I'm glad you're finding it useful! The problem is this code is 4+ years old and I'm no longer working on iD anyway. I'll leave the `v3-toolbar` branch up if anyone wants to use it for anything.

PR #7501, opened 2020-04-04
Fix confusion between allotment gardens and community gardens

stragu:
The current "Community Garden" present assigns the tag `landuse=allotments`.

According to both the [allotment](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dallotments) and [garden:type](https://wiki.openstreetmap.org/wiki/Key:garden:type) wiki pages, allotment gardens and community gardens are distinct.

This pull request fixes the issue, with the existing preset renamed to "Allotment Garden", and the `leisure=garden` preset extended with the `garden:type` field, which includes `garden:type=community`.

### DIFF
--- a/data/presets/fields/garden/type.json
+++ b/data/presets/fields/garden/type.json
@@ -1,0 +1,15 @@
+{
+    "key": "garden:type",
+    "type": "combo",
+    "label": "Garden Type",
+    "strings": {
+        "options": {
+            "arboretum": "Arboretum",
+            "botanical": "Botanical",
+            "community": "Community",
+            "residential": "Residential",
+            "roof_garden": "Roof Garden",
+            "show_garden": "Show Garden"
+        }
+    }
+}
--- a/data/presets/presets/landuse/allotments.json
+++ b/data/presets/presets/landuse/allotments.json
@@ -20,5 +20,5 @@
         "allotment",
         "garden"
     ],
-    "name": "Community Garden"
+    "name": "Allotment Garden"
 }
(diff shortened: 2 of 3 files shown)


quincylvania:
Thanks, @stragu!

Normally I'd leave this alone since these are called "community gardens" in the US, but it's indeed true that the phrase can refer to two different types of features whereas `landuse=allotments` is specific to one.

I changed the named to "Garden Allotments" because it's helpful to indicate a plural if mappers aren't too familiar with this feature type. Someone can edit the name for the UK locale if needed.

I also removed the explicit options for the Garden Type field since they're not very well established yet and they currently block someone from entering a custom value.

stragu:
Thanks, @quincylvania 

What's your feeling about also creating a dedicated preset for botanic gardens and community gardens? Ignoring the most popular `garden:type=residential` and the supposedly erroneous `garden:type=private` (shouldn't it be an `access` value?), they are the two most popular values. `garden:type=botanical` has its own wiki page.

https://taginfo.openstreetmap.org/keys/garden%3Atype#values

quincylvania:
> What's your feeling about also creating a dedicated preset for botanic gardens and community gardens?

@stragu Sounds fine, I added them. Hopefully this will help mappers realize that `leisure=garden` is for the POI, not flowerbed landuse.

The `garden:type=private` tag indeed seems erroneous but it's not clear if it should be `access=private` or `operator:type=private`. There are many privately-owned and -operated gardens that are open to the public. I added the Operator Type field to help people suss this out.

PR #7507, opened 2020-04-10
Add field for ruins=*

hikemaniac:
This PR adds a field for _[ruins](https://wiki.openstreetmap.org/wiki/Key:ruins)=*_ to the `Ruins` preset.


quincylvania:
@hikemaniac Hmm are you sure the `ruins` key should be used this way? Mostly it looks like people are just adding `ruins=yes` to `historic` things and buildings.

hikemaniac:
@quincylvania Sorry for the late reply. I see your point and even though the wiki mentions that

> In some cases the value of the "ruins" tag has been used to provide additional information

it is only used 4088 times (5.1 % of total use) with values other than "yes" or "no". The wiki page for [historic=ruins](https://wiki.openstreetmap.org/wiki/Tag:historic%3Druins) suggests `ruins=*` as a useful combination to add more information and claims that

>  Although they both provide similar information, a form such as historic=ruins, ruins=castle suggests a structure that is mainly in ruins, while a form such as historic=castle, ruins=yes suggests a structure that still appears much like a castle. 

so it seems to make a semantic difference. While I believe it makes sense to add this to the `Ruins` preset, the choice between a different `historic=*` preset or `Ruins` might be made more confusing by the addition of the `ruins=*` field. What is your opinion on this?

quincylvania:
Thinking more about this, it does seem safer and more correct to say "these are ruins, this was a church" than to say "this is a church, it is in ruins". It's not a church anymore. Though supporting one pattern doesn't mean mappers can't use the other.

PR #7522, opened 2020-04-20
Upgrade alif-toolkit to avoid warning installing

mondeja:
`alif-toolkit@1.2.6` has request as dependency and this warning was raised running `npm install`:
```
alif-toolkit > request@2.88.2: request has been deprecated, see https://github.com/request/...
```
After this commit, `alif-toolkit@1.2.9` hasn't request as dependency.


quincylvania:
@mondeja Thanks!

PR #7548, opened 2020-04-28
Remove corridors from path category

### LINKED ISSUE #7478: highway=corridor isn't an indoor feature
After unchecking "Indoor Features" iD doesn't hide corridors [highway=corridor, indoor=yes], for it this is a "path".

JamesKingdom:
This removes corridors from the path category, so that they are only affected by the indoor features category.
Fixes #7478


quincylvania:
Looks good, @JamesKingdom, thanks!!

PR #7558, opened 2020-04-30
Add support for man_made=planter

UgnilJoZ:
Good morning,

I felt like [man_made=planter](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dplanter) was missing in iD, so I added a preset.
Did I do everything right or is there additional work to do?

Cheers!


UgnilJoZ:
If my request is invalid for some reasons, just tell me :)

quincylvania:
@UgnilJoZ I'm hesitating because I'm afraid this tag is too similar to `landuse=flowerbed`. A planter is essentially a raised flowerbed, no?

UgnilJoZ:
Planters are indeed mostly elevated - but they contain not only flowers, but also shrubs and small trees. There are many planters I wouldn't tag as flowerbed.

quincylvania:
@UgnilJoZ Closing this since we actually moved iD's preset data to a [new repository](https://github.com/openstreetmap/id-tagging-schema). I suggest opening an issue there for discussion if you'd still like this to be supported.

PR #7559, opened 2020-05-01
Ignore Thunderforest Outdoors layer

hikemaniac:
This PR discards the Thunderforest Outdoors layer, as it needs an API key to be displayed without a watermark. Currently only a black background is shown, since the URL still contains the `{apikey}` parameter and is therefore invalid.

This was reported in osmlab/editor-layer-index#577.


quincylvania:
@hikemaniac Thanks!

Issue #7549, opened 2020-04-28
Rename Metal Construction preset


maro-21:
Isn't the name for craft=metal_construction "Metal Construction" ambiguous in English?
It suggests that this is a kind of construction, but this supposed to be an office of a company.
Maybe we should change it to "Metal Construction (company)" or similar?

1ec5:
“Metalworking” would be more intuitive from my (American English) perspective.

animesh-007:
@1ec5 I have made suggested changes.

maro-21:
Suggested name was "Metalworking", not "Metal Working".
https://en.wikipedia.org/wiki/Metalworking

The tag remains the same, `craft=metal_construction`, but you changed it to `craft=metal_working`.

animesh-007:
Thanks for reviewing.
I have made the suggested changes.

quincylvania:
Done in #7568.

PR #7568, opened 2020-05-05
renamed metal construction preset

animesh-007:
Renamed Metal Construction preset to Metal Working #7549 
Screen Shot 2020-05-05 at 4 48 23 PM

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -4681,8 +4681,8 @@ en:
         name: Locksmith
       craft/metal_construction:
         # craft=metal_construction
-        name: Metalworking
-        terms: '<translate with synonyms or related terms for ''Metalworking'', separated by commas>'
+        name: Metalworker
+        terms: '<translate with synonyms or related terms for ''Metalworker'', separated by commas>'
       craft/painter:
         # craft=painter
         name: Painter
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -4681,8 +4681,8 @@ en:
         name: Locksmith
       craft/metal_construction:
         # craft=metal_construction
-        name: Metal Construction
-        terms: '<translate with synonyms or related terms for ''Metal Construction'', separated by commas>'
+        name: Metalworking
+        terms: '<translate with synonyms or related terms for ''Metalworking'', separated by commas>'
       craft/painter:
         # craft=painter
         name: Painter
--- a/data/presets/presets.json
+++ b/data/presets/presets.json
@@ -416,7 +416,7 @@
   "craft/insulator": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "insulation"}, "name": "Insulator"},
   "craft/joiner": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "joiner"}, "terms": ["furniture"], "name": "Joiner"},
   "craft/key_cutter": {"icon": "fas-key", "geometry": ["point", "area"], "tags": {"craft": "key_cutter"}, "name": "Key Cutter"},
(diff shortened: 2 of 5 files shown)


data/presets/presets/craft/metal_working.json
```diff
@@ -7,5 +7,5 @@
     "tags": {
         "craft": "metal_construction"
     },
-    "name": "Metal Construction"
+    "name": "Metalworking"
```
1ec5: Let’s move this file back to where it was. I don’t think anything depends on the file names being consistent with the tags, but it’ll be easier to find the preset at the old name.

data/presets.yaml
```diff
@@ -4467,10 +4467,10 @@ en:
       craft/locksmith:
         # craft=locksmith
         name: Locksmith
-      craft/metal_construction:
+      craft/metal_working:
```
1ec5:
```suggestion
      craft/metal_construction:
```

data/presets/presets.json
```diff
@@ -410,7 +410,7 @@
     "craft/insulator": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "insulation"}, "name": "Insulator"},
     "craft/joiner": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "joiner"}, "terms": ["furniture"], "name": "Joiner"},
     "craft/key_cutter": {"icon": "fas-key", "geometry": ["point", "area"], "tags": {"craft": "key_cutter"}, "name": "Key Cutter"},
-    "craft/metal_construction": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "metal_construction"}, "name": "Metal Construction"},
+    "craft/metal_working": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "metal_construction"}, "name": "Metalworking"},
```
1ec5:
```suggestion
    "craft/metal_construction": {"icon": "temaki-tools", "geometry": ["point", "area"], "tags": {"craft": "metal_construction"}, "name": "Metalworking"},
```

1ec5 (commented):
Thanks for making this change! There are a few things left to clean up so that the raw tag remains `craft=metal_construction`. Once you make these changes, remember to rebuild iD (`npm install && npm run all`) to update the remaining files.

animesh-007:
@1ec5 Thanks for suggestions. I have made those changes.

data/presets/presets.json
1ec5: I implemented this change and force-pushed to your branch.

dist/locales/en.json
```diff
@@ -10708,10 +10708,6 @@
                 "name": "OpenStreetMap Galicia Twitter",
                 "description": "Follow us on Twitter at {url}"
             },
-            "Galicia-Xeoinquedos-Twitter": {
```
1ec5: @quincylvania I ended up introducing a lot of diffs related to the community index and NSI after rebasing and building; might want to double-check that I’m not blowing anything away due to something missing in my configuration.

1ec5 (approved):

dist/locales/en.json
quincylvania: Hmm you probably need to run `npm install` to sync up your versions of these dependencies before building again.

quincylvania:
What do you all think about the name "Metalworker" instead? Lots of the `craft` presets are like "Plumber", "Roofer", "Painter", "Photographer"…

dist/locales/en.json
1ec5: Thanks, that worked.

1ec5:
That seems reasonable to me.

quincylvania:
@animesh-007 @1ec5 Thanks!!

Issue #7432, opened 2020-03-13
Hover state persistst when tapping buttons on touch devices


quincylvania:
Mobile Safari gives tapped buttons the `:hover` CSS styling, which shows them as pressed. However, this doesn't always properly disappear after the tap is complete. It's not a big deal but it looks sloppy and can be confusing.

quincylvania:
Turns out the [`hover` media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover) can by used to apply CSS only to environments where the primary pointer can hover. I wrapped all of iD's `:hover` selectors in media queries and added `:active` selectors so all buttons still react when pressed on touch devices.

PR #7607, opened 2020-05-16
Add highlight to multi-select sidebar on hover of entity

### LINKED ISSUE #2949: With multiple features selected, indicate which one is mouse-hovered in the sidebar multiselect list

jgscherber:
This is to resolve #2949 and is a WIP

This is my first time working on this project (and first time working in a large JS codebase) so please don't hold back on style/design criticisms. I want to thank you for the Architecture readme that you have. It really helped with initially finding my way around.

Here's a demo of the highlight
highlight_demo

I ran into a couple issues while working on it
* After adding the `hoverModeSelect` callback, performing a click-and-drag results in iD dropping out of "select" mode. Any ideas what's causing that?
* The discussion in #2949 says the resolution to this issue should also have the sidebar scroll into view the element in the list that's being highlighted - how can I check if an element is visible in the viewport?
* This change adds a class `.hover` to perform the styling which duplicates the pseudo-class `:hover` -- is it better for the `:hover` pseudo-class be removed and only styled using the `.hover` class


quincylvania:
@jgscherber Thanks for working on this! I'll take a look and give you feedback this week.

modules/ui/sidebar.js
```diff
@@ -113,6 +113,24 @@ export function uiSidebar(context) {
             .append('div')
             .attr('class', 'inspector-hidden inspector-wrap fr');
 
+        var hoverModeSelect = function hoverModeSelect(datum) {
```
quincylvania, line 156: No need to name this twice, you can pick either style.

modules/ui/sidebar.js
```diff
@@ -113,6 +113,24 @@ export function uiSidebar(context) {
...
+
+            if (context.mode().selectedIDs().length > 1 
+                && (datum instanceof osmEntity)){
+                
+                var elements = context.container().selectAll('.feature-list-item')
+                    .filter(function (node) {
+                        return node.id === datum.id;
+                });
+                
+                if (elements.size() > 0){
```
quincylvania, line 167: This works fine but there's a `.empty()` function that's even easier.

quincylvania:
@jgscherber Looks pretty good! We could merge as-is but I'll give you a chance to take another look if you want. Let me know if you have any questions.

> After adding the hoverModeSelect callback, performing a click-and-drag results in iD dropping out of "select" mode. Any ideas what's causing that?

This was a bug with other changes I'd been making in the `develop` branch. If you remerge this should be fixed.

> The discussion in #2949 says the resolution to this issue should also have the sidebar scroll into view the element in the list that's being highlighted - how can I check if an element is visible in the viewport?

I think we might actually live without this. I feel like it'd be disorienting more often than not. But to answer your question you can compare the list item element's position to [`scrollTop`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop) of the inspector.

> This change adds a class .hover to perform the styling which duplicates the pseudo-class :hover -- is it better for the :hover pseudo-class be removed and only styled using the .hover class

Nah, there's no harm in two separate classes. I had to move all the `:hover` declarations to media queries anyway for #7432.

jgscherber:
I'm planning on changing your two suggestions. I'll fix my rebase also

PR #7620, opened 2020-05-19
Remove force "major_light" from all lighthouses.

Eric-Sparks:
West Marine should be tagged with the appropriate "chandler" small craft facilities seamark tags.


Eric-Sparks:
Also, not all lighthouses are "light_major" seamarks as some are decommissioned and only have a smaller, special light affixed.

quincylvania:
Hi and welcome, @Eric-Sparks.

The West Marine preset comes from the [name-suggestion-index](https://github.com/osmlab/name-suggestion-index) project, so you'll have to bring it up there.

The lighthouse tag we can fix but you'll have to change the `iD/data/presets/presets/man_made/lighthouse.json` file. The `data/presets/presets.json` file is derived and would be overwritten.

Eric-Sparks:
> The West Marine preset comes from the [name-suggestion-index](https://github.com/osmlab/name-suggestion-index) project, so you'll have to bring it up there.

Great, I was looking for where to bring that up.  Thanks.

> The lighthouse tag we can fix but you'll have to change the `iD/data/presets/presets/man_made/lighthouse.json` file. The `data/presets/presets.json` file is derived and would be overwritten.

I've made the appropriate changes to lighthouse.json.  I'll remove the `data/presets/presets.json` file from the pull request since that doesn't seem to need changing.

Thanks!

Eric-Sparks:
Ugh, github is so infuriating sometimes... starting over.

PR #7621, opened 2020-05-19
 Remove force "major_light" from all lighthouses.

Eric-Sparks:
Not all lighthouses are light_major as some are no longer active.


quincylvania:
Looks good, thanks @Eric-Sparks 👍

Eric-Sparks:
Awesome, thank you!

PR #7622, opened 2020-05-19
Update marina.json

Eric-Sparks:
Moved some fields from 'moreFields' to 'Fields' to get better visability into what information is important to document with marinas.
Also added a few tags that will help capture information that is helpful.
Added proper seamark tags.

### DIFF
--- a/data/presets/presets/leisure/marina.json
+++ b/data/presets/presets/leisure/marina.json
@@ -35,8 +35,8 @@
         "boat"
     ],
     "tags": {
-        "leisure": "marina"
-        "seamark:type": "harbour"
+        "leisure": "marina",
+        "seamark:type": "harbour",
         "seamark:harbour:category": "marina"
     },
     "name": "Marina"
--- a/data/presets/presets/leisure/marina.json
+++ b/data/presets/presets/leisure/marina.json
@@ -3,24 +3,28 @@
     "fields": [
         "name",
         "operator",
-        "capacity",
+        "phone",
+        "email",
+        "website",
         "fee",
         "payment_multi_fee",
         "charge_fee",
         "sanitary_dump_station",
-        "power_supply"
+        "internet_access",
+        "internet_access/fee",
+        "maxlength",
+        "maxwidth",
+        "maxdraft",
+        "maxairdraft",
+        "power_supply",
+        "vhf_channel"
     ],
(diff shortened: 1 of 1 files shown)


Eric-Sparks:
I'm not exactly sure how to propose such a change, but the sanitary_dump_station for boats is different from RVs.  Simply marking it as "yes" doesn't really map it correctly and could lead to some issues with RVers navigating to the wrong location.

Actually, I think I just found the proper tag...  waterway=sanitary_dump_station instead of using amenity=sanitary_dump_station.

data/presets/presets/leisure/marina.json
```diff
@@ -3,24 +3,28 @@
...
         "sanitary_dump_station",
-        "power_supply"
+        "internet_access",
+        "internet_access/fee",
+        "maxlength",
+        "maxwidth",
+        "maxdraft",
+        "maxairdraft",
+        "power_supply",
+        "vhf_channel"
```
quincylvania, line 20: I don't think all these fields exist yet. You'll have to add them in files under `data/presets/fields`.

data/presets/presets/leisure/marina.json
```diff
@@ -31,7 +35,9 @@
         "boat"
     ],
     "tags": {
-        "leisure": "marina"
+        "leisure": "marina",
+        "seamark:type": "harbour",
+        "seamark:harbour:category": "marina"
```
quincylvania, line 40: These should go under `addTags` since otherwise this preset won't match any existing features that don't have them.

data/presets/presets/leisure/marina.json
```diff
@@ -3,24 +3,28 @@
     "fields": [
         "name",
         "operator",
-        "capacity",
+        "phone",
```
quincylvania, line 6: We try to limit the `fields` array to only a few of the most important fields since some mappers feel like they have to fill out everything. Lots of complex default fields can become intimidating. Please keep the advanced fields under `moreFields`.

quincylvania (changes requested):
Looks like there are a few issues here. You can review the [presets readme](https://github.com/openstreetmap/iD/blob/develop/data/presets/README.md) to learn more about the file format. Also, be sure to run `npm run build:data` to check for errors and include the derived data changes as well.

PR #7629, opened 2020-05-24
show when and who last modified a feature

### LINKED ISSUE #9691: When entity is selected show "last modified" in left pane
### Description

Let's say I am about to change some road -- add surface information -- I click on the road, in the left pane I see all existing features for it. The one information that would be handy is "Last modified: 2020-07-12" (example).

Rationale: to avoid ping-pong effect I generally stop making changes when the last modification is fresh (like somebody put grade 2, and I want to change it to grade 4). But if the change is old (7 years ago) it indicates that road simply changed in time. Such information does not take too much screen space and it would be valuable.

Currently I have to switch to OSM inspector, which is time consuming (when making a lot of changes).

Extra feature would be showing tooltip with commit comment when moving mouse over "Last modified ..." block.

### Screenshots

_No response_

k-yle:
Closes #9691

It would be very helpful if you could see when a feature was last modified from within the iD editor. I've changed the `View on openstreetmap.org` message to `Last modified by <user> xx <days/months/...> ago`. The original message is still shown when you hover over the link.

image

&nbsp;

~~I haven't added localization yet - wasn't sure if it's okay to use the new [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) API.~~

### DIFF
--- a/modules/ui/view_on_osm.js
+++ b/modules/ui/view_on_osm.js
@@ -1,6 +1,7 @@
 import { t } from '../core/localizer';
-import { osmEntity, osmNote } from '../osm';
+import { osmEntity, osmNote, osmRelation, osmWay } from '../osm';
 import { svgIcon } from '../svg/icon';
+import { getRelativeDate } from '../util/date';
 
 
 export function uiViewOnOSM(context) {
@@ -31,9 +32,23 @@ export function uiViewOnOSM(context) {
             .attr('href', url)
             .call(svgIcon('#iD-icon-out-link', 'inline'));
 
-        linkEnter
-            .append('span')
-            .call(t.append('inspector.view_on_osm'));
+
+        if (_what && !(_what instanceof osmNote)) {
+            // node/way/relation
+            const { user, timestamp } = uiViewOnOSM.findLastModifiedChild(context.history().base(), _what);
+
+            linkEnter
+                .append('span')
+                .text(t('inspector.last_modified', {
+                    timeago: getRelativeDate(new Date(timestamp)),
+                    user
+                }))
+                .attr('title', t('inspector.view_on_osm'));
+        } else {
+            linkEnter
+                .append('span')
+                .call(t.append('inspector.view_on_osm'));
+        }
     }
 
 
@@ -45,3 +60,33 @@ export function uiViewOnOSM(context) {
 
     return viewOnOSM;
 }
+
+
+/**
+ * @param {iD.Graph} graph
+ * @param {iD.OsmEntity} feature
+ */
+uiViewOnOSM.findLastModifiedChild = (graph, feature) => {
+    let latest = feature;
+
+    /** @param {iD.OsmEntity} obj */
+    function recurseChilds(obj) {
+        if (obj.timestamp > latest.timestamp) {
+            latest = obj;
+        }
+        if (obj instanceof osmWay) {
+            obj.nodes
+                .map(id => graph.hasEntity(id))
+                .filter(Boolean)
+                .forEach(recurseChilds);
+        } else if (obj instanceof osmRelation) {
+            obj.members
+                .map(m => graph.hasEntity(m.id))
+                .filter(e => e instanceof osmWay || e instanceof osmRelation)
+                .forEach(recurseChilds);
+        }
+    }
+
+    recurseChilds(feature);
+    return latest;
+};
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Consider other name-like tags for labelling features, such as `lock_name` ([#9588], thanks [@k-yle])
 * Always take reduced map size due to open side panels into account when zooming to selected entities
 * Limit landuse tags that are considered for "Landuse" area features ([#11184], thanks [@youssefelzedy])
+* When a feature is selected, show when and who last modified it ([#7629], thanks [@k-yle])
 * Render climbing routes and via ferrata ways as dotted lines ([#11133], thanks [@harahu])
 #### :scissors: Operations
 #### :camera: Street-Level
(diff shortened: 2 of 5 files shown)


1ec5 (commented):
Neat, this saves me from having to open the well-hidden Info panel. I wonder if the relative timestamp could be shared with that panel for consistency too.

> I haven't added localization yet - wasn't sure if it's okay to use the new Intl.RelativeTimeFormat API.

This API would be ideal, but it still needs a fallback for Safari, which doesn’t support it.

The format string will also need to be localized just like the one that is being replaced.

k-yle:
I've made the changes to localize the last modified message. 

Not sure how to implement a fallback for Safari - currently, if `Intl.RelativeTimeFormat` is supported it produces a string like "Last modified **6 months ago** by Kyle". Otherwise, it just displays the date in the user's locale e.g. "Last modified **on 26/05/2020** by Kyle".

The rest of the text is also localized.

Dimitar5555:
Safari has implemented RelativeTimeFormat according to https://caniuse.com/?search=Intl.RelativeTimeFormat.

k-yle:
yeah, there's nothing blocking this PR anymore. I have a slightly different version (adf8ea66) deployed to https://kyle.kiwi/iD 

Since this PR has been sitting around for 3 years I doubt it will ever get merged though...

tordans:
Would love to see this merged. Had the need for this several times this week.

tyrasd (commented):
At first, I noticed that this essentially duplicates the information already displayed in the History panel. But as that panel is not very well known, there is merit to use the "open on osm" link for some extra information. And I think it is actually a good thing to make the mappers behind the data more visible, and the data vintage can also be useful while mapping, e.g. to find out if a mapped building that is not visible on aerial imagery is so new that it wasn't yet captured, or the opposite.

However, what if we used this opportunity to be a little more precise in what we mean by “_last edited by_”: The current version of this PR just naively displays the timestamp/user metadata of the respective raw OSM entity. Which, granted, is also what the osm website displays, but in other places iD is more user friendly: e.g. when uploading a changeset, a moved building is not represented as _you changed 4+ nodes_, but rather this is represented as _one changed building_ feature:



See below for a suggestion to find by whom and when the selected feature was _actually_ last modified. (The history panel would still display the raw OSM entity metadata, for those who need that.)

What do you think?

modules/ui/view_on_osm.js
```diff
@@ -31,9 +32,21 @@ export function uiViewOnOSM(context) {
...
-            .append('span')
-            .call(t.append('inspector.view_on_osm'));
+
+        if (_what && !(_what instanceof osmNote)) {
+            // node/way/relation
+            linkEnter
+                .append('span')
+                .text(t('inspector.last_modified', {
+                    timeago: getRelativeDate(new Date(_what.timestamp)),
+                    user: _what.user
```
tyrasd: Code for the suggestion above.

```suggestion

            let timestamp = _what.timestamp;
            let user = _what.user;
            const graph = context.history().base();
            function recurseChilds(obj) {
                if (obj.timestamp > timestamp) {
                    timestamp = obj.timestamp;
                    user = obj.user;
                }
                if (obj instanceof osmWay) {
                    obj.nodes
                        .map(graph.hasEntity.bind(graph))
                        .filter(Boolean)
                        .forEach(recurseChilds);
                } else if (obj instanceof osmRelation) {
                    obj.members
                        .map(m => graph.hasEntity(m.id))
                        .filter(e => e instanceof osmWay || e instanceof osmRelation)
                        .forEach(recurseChilds);
                }
            }
            recurseChilds(_what);

            linkEnter
                .append('span')
                .text(t('inspector.last_modified', {
                    timeago: getRelativeDate(new Date(timestamp)),
                    user
```

It might be possible to simplify it a little bit by not to going as far as to find the changes of a relation by looking at the changes to a relation's way's nodes. And/or: for something like a `restriction` relation it might even make sense to skip the recursion. :thinking:

k-yle: this is something i considered, but as you mentioned, there are two major drawbacks:
1. it would not match what the OSM website displays, which is bound to cause confusion. Especially since this is a hyperlink that leads to a page with conflicting information. 
2. as soon as we start special-casing some relations (where we do not recurse down), then the feature becomes less useful because no one will understand its behaviour and which tags we special-case

so my suggestion is to make this hyperlink exactly match the OSM data model, and in the future maybe the history panel could use the intelligent approach? since there's also more space there

tordans: For when we look into the history panel more… maybe we find a way there to also explain to users this special behavior of OSM which only increases version and date when tags are changed. So "last modified" is more like "Last tag modification".

tyrasd: I'm actually leaning quite strongly towards also considering the minor edits in this quite prominent location:

* This is the consistent abstraction behavior within the editor: e.g. when a mapper moves a building, the upload dialog shows that _the building_ was edited. Now, if the footer would only show the last major version of the object, it would not credit the mapper with their edit, and it could be confusing for the mapper.
* One of the main use cases of the "last modified on" information is to evaluate whether a selected feature is fresher then the background imagery or the opposite. In that case, the geometry-only changes of a _move_ operation are very relevant.
* I'd consider the fact that the website doesn't do this properly as a deficiency of the website, which might be fixed eventually. Also, strictly speaking, the website does not attribute the user as the "last modifier", but just labels them as "edited by".
* I cannot think of a situation where only the last major version's metadata is important to know while mapping.
* For relations: yeah, it might indeed be confusing if the behavior depends on the type. Let's just keep it consistent then, and not look at the `type` tag.

tyrasd: PS: if we come to the conclusion that the discrepancy between the link text and the way the osm website displays the information is a major blocker for this suggestion, that could be also addressed differently. For example by just printing the "last modified on … by …" information as text and change the "open on osm" link to an OSM icon that links to osm.org. 



//edit: this would also have the additional benefits to be able to hotlink the osm user profile, and the relative date could have a hover text to show the absolute date instead, and the date could maybe even link to the changeset of the last edit.

k-yle: done but without the special-casing for some tags. If we're happy with the mismatch in iD vs the website, then I've got no other concerns – we can just see if people actually get confused.

[this is a good test case](https://osm.org/way/1206128777) where I edited every child node last week but the website says "2 years ago"[^1]

[^1]: disclaimer: this example is staged, I was thinking of this PR while mapping and intentionally did not add/delete/reorder any nodes 🤓

tyrasd (approved):

PR #7630, opened 2020-05-24
Add `real_fire=yes/no` preset to pubs

amandasaurus:
Pub culture is an important part of Irish & British culture (amoung others), this is probably why OSM has had a distinction between `amenity=pub` versus `amenity=bar`. There are many properties people care about, one is the presence of a real fire ( e.g [WhatPub allows one to search for pubs which have/don't have a real fire](https://whatpub.com/search?q=map) ). This patch adds the `real_fire=yes/no` preset to `amenity=pub`.

I don't think it's possible to have *fields* in a preset to be customized per country, if it was, I'd suggest limiting this to pubs in Ireland & Britain.


trigpoint:
Thank you Rory, this is a good addition and I support it.

quincylvania:
So we recently did add the ability to [limit fields to countries](https://github.com/openstreetmap/iD/blob/78a4b75e55f49ed54fe0a0d1d8348ababf47aa82/data/presets/schema/field.json#L194-L211), but I don't think it's necessary to do so in this case. There are pubs all over the world in the European style where this tag might be useful.

I admit I don't love the name Real Fire—does this include gas fires or just wood fires? Just fireplaces or also open firepits?

amandasaurus:
> I admit I don't love the name Real Fire—does this include gas fires or just wood fires? Just fireplaces or also open firepits?

It's a bit weird alright, but it's a very common term. (In normal times) you could approach anyone on the street in Ireland or Britain and ask “Does [pub] have a real fire?” and they know what you mean.

It means a fireplace, where a venue would regularly light a fire there (plenty of old fireplaces are blocked up). Anything that looks like the image on the wiki: https://wiki.openstreetmap.org/wiki/File:Giardino_Mountain_fireplace_and_lounge_-_panoramio.jpg

The fuel doesn't have to be wood. There's coal or, in Ireland, [peat briquettes](https://en.wikipedia.org/wiki/Briquette#Peat_briquettes). As per the wiki, a gas fire would not have the same aroma, sound, or "ambience", so wouldn't commonly be included.

“Open firepits”? That sounds like a campsite, where the `fireplace` tag is used. You don't get "pits" indoors in Irish (etc) pubs

quincylvania:
@rory Okay, since the term is more specific than the name alone implies, and since this is clear to people in Britain and Ireland, I would suggest that we limit the field to those countries for now. We can't expect other people to know this field includes coal fires but not gas fires, particularly when the name is translated to all of iD's languages.

In the future I would suggest picking descriptive tags that can easily be used anywhere, e.g. `fireplace:fuel:wood=yes`, etc.

amandasaurus:
“Limiting it to IE & UK” works for me. 🙂

PR #7642, opened 2020-05-31
Header size didn't include padding above buttons

### LINKED ISSUE #7640: Mid-way toggle is too close to top bar to easily grab
image

The toggle seems to stay a fixed distance from any edge of the map view, but with the top bar coverting part of the view the mid-way toggle is right on edge of it, making it hard to grab.

jgscherber:
Resolves #7640 

This is the spacing after the changes
cleared_header

**Root cause**
The header size in `trimmedExtent` was set to the size of only the inner `div` (i.e. from the top of the buttons to the bottom of the bar). I changed the value to include the padding space above the buttons.
header_height

I also confirmed the footer size of 30px is correct
footer_height

### DIFF
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -981,7 +981,7 @@ export function rendererMap(context) {
 
     map.trimmedExtent = function(val) {
         if (!arguments.length) {
-            var headerY = 60;
+            var headerY = 71; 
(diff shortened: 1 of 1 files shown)


quincylvania:
@jgscherber Thanks a bunch!

PR #7649, opened 2020-06-02
Add follow action and operation

### LINKED ISSUE #1614: Follow a path
In other mapping software like ArcGIS and OCAD there is an option for following a polyline/polygon when creating a new object instead of adding all nodes one by one.

This function would be amazing and time saving if iD had it!

kaligrafy:
closes #1614


kaligrafy:
I'll close for now, a better/up to date version would be needed

PR #7651, opened 2020-06-03
Add some build tips encountered on Raspberry Pi.

jleedev:


CONTRIBUTING.md
```diff
@@ -354,6 +354,22 @@ You can rebuild iD completely with the command `npm run all`.
...
+### Build Troubleshooting
+
+**`npm install` fails with "PhantomJS not found on PATH … Try to install
+PhantomJS globally".**
+Then you should do exactly that, such as `apt install phantomjs`.
+
+**`npm install` fails with "Command failed: /usr/bin/phantomjs --version
+qt.qpa.screen: QXcbConnection: Could not connect to display".**
+Add `QT_QPA_PLATFORM=offscreen` to your environment. (This will unblock
+installation but running the tests on a Raspberry Pi has not been attempted.)
```
pnorman, line 366: I don't think specific hardware should be mentioned in the documentation.

CONTRIBUTING.md
```diff
@@ -354,6 +354,22 @@ You can rebuild iD completely with the command `npm run all`.
 iD will be built to the `dist` directory. This directory is self-contained; you can copy it
 into the public directory of your webserver to deploy iD.
 
+### Build Troubleshooting
+
+**`npm install` fails with "PhantomJS not found on PATH … Try to install
+PhantomJS globally".**
+Then you should do exactly that, such as `apt install phantomjs`.
```
pnorman, line 361: If https://github.com/openstreetmap/id#prerequisites is outdated, it should be updated

CONTRIBUTING.md
```diff
@@ -354,6 +354,22 @@ You can rebuild iD completely with the command `npm run all`.
...
+
+**`npm install` fails with "Command failed: /usr/bin/phantomjs --version
+qt.qpa.screen: QXcbConnection: Could not connect to display".**
```
pnorman, line 364: What OS did you encounter this on? This looks like a bad phantomjs install, not an iD issue

pnorman (changes requested):

jleedev:
Most of this has suffered from bit rot, as phantomjs isn't even being used anymore, so I'll close this.

Just some notes:

- https://github.com/openstreetmap/id#prerequisites doesn't say anything about dependencies, but https://github.com/openstreetmap/iD/blob/develop/CONTRIBUTING.md#tests says that karma needs CHROME_BIN pointing at the path to chromium, which seems good.
- The phantomjs install really was that flaky, _and_ would leave the node_modules in a broken state that `npm install` didn't fix.
- The low memory situation is still valid, but not really worth mentioning. (But wow, Karma manages to consume 1.5G of memory just to tell me that it can't find Chrome.)

PR #7696, opened 2020-06-11
Add rollup replace plugin for env vars subst.

ssipos90:
Hello,

I've added the ability to provide some of vars as env variables at rollup time. We are using this on our private OSM instance in a dockerized environment.

Also, I refactored the rollup config to prevent duplicate code from accumulating.

### DIFF
--- a/modules/config.js
+++ b/modules/config.js
@@ -1,0 +1,4 @@
+/* eslint-disable no-undef */
+export const osm_urlroot = OSM_URLROOT;
+export const oauth_consumer_key = OSM_OAUTH_CONSUMER_KEY;
+export const oauth_secret = OSM_OAUTH_SECRET;
--- a/modules/services/osm.js
+++ b/modules/services/osm.js
@@ -12,14 +12,15 @@ import { geoExtent, geoRawMercator, geoVecAdd, geoZoomToScale } from '../geo';
 import { osmEntity, osmNode, osmNote, osmRelation, osmWay } from '../osm';
 import { utilArrayChunk, utilArrayGroupBy, utilArrayUniq, utilRebind, utilTiler, utilQsString } from '../util';
 
+import { osm_urlroot, oauth_consumer_key, oauth_secret } from '../config';
 
 var tiler = utilTiler();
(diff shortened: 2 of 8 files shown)


modules/config.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable no-undef */
```
ssipos90, line 1: This file is necessary to avoid the eslint rule, otherwise is redundant.

modules/services/osm.js
```diff
@@ -12,14 +12,15 @@ import { geoExtent, geoRawMercator, geoVecAdd, geoZoomToScale } from '../geo';
 import { osmEntity, osmNode, osmNote, osmRelation, osmWay } from '../osm';
 import { utilArrayChunk, utilArrayGroupBy, utilArrayUniq, utilRebind, utilTiler, utilQsString } from '../util';
 
+import { osm_urlroot, oauth_consumer_key, oauth_secret } from '../config';
 
 var tiler = utilTiler();
 var dispatch = d3_dispatch('apiStatusChange', 'authLoading', 'authDone', 'change', 'loading', 'loaded', 'loadedNotes');
-var urlroot = 'https://www.openstreetmap.org';
+var urlroot = osm_urlroot;
```
ssipos90, line 19: This var is duped because there's a setting down somewhere that can override this and importing creates const.

ssipos90:
Not worth it.

quincylvania:
> Not worth it.

@ssipos90 I haven't had a chance to look at this yet, but I'd be happy to work through it with you!

ssipos90:
I closed it because it's not the best approach regarding the auth mechanism, however it can be used for other build time vars.

`osmAuth` should be configurable at runtime (overriding the `urlroot` and `oauth_*` stuff when creating the context by calling the `switch` function or smtn) and this is something that I plan working on in the near future.

bhousel:
I would prefer to not refactor the config files.  I know there is duplicate code there to support different build targets, but I think it's valuable to be able to open a config file and see everything that it does without needing to look elsewhere.

ssipos90:
I've created another PR (#7720) that doesn't refactor Rollup config files, as requested.

Issue #6836, opened 2019-09-06
Preset: Cycle & Foot Path should produce highway=path rather than highway=cycleway


naomap:
The preset highway/cycleway/bicycle_foot.json produces **highway=cycleway** + foot=designated + bicycle=designated. 

According to the **case S5** in https://wiki.openstreetmap.org/wiki/Bicycle#Miscellaneous, a "Cycle and Foot path" should instead produce  **highway=path** + foot=designated + bicycle=designated.

This bug causes a lot of burden in the Greater Paris area, where a few mappers have damaged a lot of data.

sun-geo:
I agree, same tagging (highway=**path** + foot=designated + bicycle=designated) is a tool recommending for German signs, see:

http://osmtools.de/traffic_signs/?signs=240
highway=path
bicycle=designated
foot=designated
segregated=no
traffic_sign=DE:240

or

http://osmtools.de/traffic_signs/?signs=241
highway=path
bicycle=designated
foot=designated
segregated=yes
traffic_sign=DE:241

BjornRasmussen:
I personally prefer using `highway=cycleway` on any path that is designated for both cyclists and pedestrians, since those paths look more like cycleways than what I would consider a "path".  Although the two methods are equivalent, I think that using cycleway implies that the way is more than just a path in the woods, but rather one that is more formal, and suitable for cyclists.
So really, it doesn't actually matter which tags are used on this preset, because data consumers will interpret them all the same, so I don't think we should change the preset without a good reason.

charlesmillet:
> I personally prefer using `highway=cycleway` on any path that is designated for both cyclists and pedestrians, since those paths look more like cycleways than what I would consider a "path". Although the two methods are equivalent, I think that using cycleway implies that the way is more than just a path in the woods, but rather one that is more formal, and suitable for cyclists.
> So really, it doesn't actually matter which tags are used on this preset, because data consumers will interpret them all the same, so I don't think we should change the preset without a good reason.

This is a long standing subject concerning the path. I used to prefer to chose highway=footway or highway=cycleway rather than use highway=path but I admitted that highway=path was something very generic, justly adapted to those cases where you can't tell and that some additional information like surface, width, segregation, etc. were widely sufficient to distinguish with simple paths.

naomap:
[...]

PR #7703, opened 2020-06-14
bicycle_foot: Add "de" to notCountryCodes

henry4442:
see: https://github.com/openstreetmap/iD/issues/6836#issuecomment-643802110


quincylvania:
@henry4442 Thanks!

PR #7707, opened 2020-06-16
Adding FOMI ortophotos

7F11:


quincylvania:
Welcome, @7F11. iD's imagery data currently comes from the [editor-layer-index](https://github.com/osmlab/editor-layer-index/) project. You'll have to open a PR over there, since this file will be overwritten when we do a release.

7F11:
> Welcome, @7F11. iD's imagery data currently comes from the [editor-layer-index](https://github.com/osmlab/editor-layer-index/) project. You'll have to open a PR over there, since this file will be overwritten when we do a release.

Thanks for the hint.

PR #7720, opened 2020-06-19
Add rollup replace plugin for env vars subst

ssipos90:
As requested, I've created a new pull request in favor of reopening the old one.  (#7696)

Meanwhile, I've found a way to override the urlroot and stuff at runtime (eg: in index.html), by calling the switch function before creating the context.

```javascript
          // ...
          iD.services.osm.switch({
            'urlroot': 'https://...',
            'oauth_consumer_key': '....',
            'oauth_secret': '...'
          });
          var context = iD.coreContext()
         // ...
```


package.json
```diff
@@ -90,6 +90,7 @@
     "js-yaml": "^3.14.0",
     "json-stringify-pretty-compact": "^2.0.0",
     "jsonschema": "^1.1.0",
+    "lodash": "^4.17.15",
```
ssipos90, line 93: This was already used as a dependency, so I added it as a dev dependency and used it, as `lodash-es` doesn't work because rollup doesn't transpile `node_modules`.

k-yle:
These changes were incorporated in c8a3cf1 using a somewhat different approach, so I think this PR can be closed. Thanks for kick-starting the work @ssipos90

tordans:
Closing based on https://github.com/openstreetmap/iD/pull/7720#issuecomment-2151925367

PR #7740, opened 2020-06-25
Add duplicate address validator

### LINKED ISSUE #2137: Need a duplicate address validation
As more and more address nodes are added (imported) into OSM, having a check for duplicates would help prevent adding duplicates. If iD spots a duplicate the user should be notified and recommended that the nodes be combined. 

Which is a related request, make it easy to combine a POI, i.e. restaurant with an address node.

bhousel:
Here's a duplicate address validator that we built for RapiD...
Hope you find it useful!
previous discussion: https://github.com/facebookincubator/RapiD/pull/134


Nakaner:
There have been a couple of issues and a closed pull request related to duplicated addresses:

* 6177: [Brainstorming: Improving address mapping and validation](https://github.com/openstreetmap/iD/issues/6177)
* 6189: [Validate against duplicate name/address tags](https://github.com/openstreetmap/iD/pull/6189)
* 2137: [Need a duplicate address validation](https://github.com/openstreetmap/iD/issues/2137)

In the discussion of #2137, the corner case of multiple POIs having the same address located in the same building, and the case of addresses mapped on the entrance node of the building with POI nodes in the building having the same address was mentioned. That issue was closed in 2014 with the following comment (note that this was a couple of years prior to iD getting a validator):
> I think this is best implemented in a third-party QA tool (detecting duplicates after the fact) first, where its utility can be proven and all the corner cases can be worked out without false positives impacting a large number contributors. I think this may be a difficult validation to get a high reliability on.

#6177 is a list of ideas what validation related to addresses in OSM can do. The ticket is labelled "considering".

#6189 tried to implement a subset of #6177. I myself warned that addressing differs from country to country. Duplication which is not liked in one country is established and accepted mapping elsewhere. I mentioned the following issue that should kept in mind:
> POIs often have addr:*=*tags even if multiple of them are located in the same building. There is no consensus whether this is good or bad. An editor should not enforce any point of view in this issue.

@bhousel What's the difference compared to #6189? Is your new implementation aware of POIs (nodes and ways) having the same address than the building they are located in? Is it aware of POIs having the same address as other POIs in the same building (building not having an address)? How does it handle these cases? If it does not address these issues, there is no or almost no improvement compared to #6189 and asking for exactly the same trouble you avoided about one year ago by closing pull request #6189.

If you prefer to get this merged into iD nevertheless, please add `addr:city` and `addr:postcode` to the list of tags to be compared. Otherwise you miss the following case: If a street crosses city boundaries but has the same name on both sides, two buildings having the same number on both sides of the boundary have different cities and likely different postcodes and are therefore different addresses. And even within cities, street names can be duplicated as long as the street is unique within a postal code area. The example sounds strange and likely leads to a lot of confusion among delivery services etc. but in real world, any strange corner case exists.

bhousel:
Thanks @Nakaner for your comments..

As you noted, #2137 was many years before iD had a validator, so our thinking has changed a bunch.  John's comments about external validation from 2014 aren't relevant today.

I closed #6189 mainly because you seemed opposed to the idea.  It is true that 6189 would have raised a bunch of false positives


> What's the difference compared to #6189? Is your new implementation aware of POIs (nodes and ways) having the same address than the building they are located in? Is it aware of POIs having the same address as other POIs in the same building (building not having an address)? How does it handle these cases?

Yes, this code is a bit different - I'll try to outline the differences:
* 6189 compared tags against all entities downloaded to iD's graph at the time, this PR compares tags only against entities within 50m of one another
* 6189 simply raised a warning if tags are the same, this PR applies a heuristic to decide whether the addresses share housenumbers without being different in other address tags (the `isDuplicate` code is pretty readable)
* 6189 did not handle housenumbers with semicolon or comma delimited lists, this PR does.
* 6189 did case insensitive compare, and this PR does not (6189 is better here, I will update this PR to compare case insensitive, to catch duplicates like "Apt 2B"/"Apt 2b")
* 6189 had tests, and this PR does not (6189 is better here too, I will add tests)

> If you prefer to get this merged into iD nevertheless, please add `addr:city` and `addr:postcode` to the list of tags to be compared.

Good idea, I'll make the change.. That will catch the case you mentioned where a street with same name crosses a city or postcode boundary and has the same numbers on either side.

modules/validations/duplicate_address.js
```diff
@@ -0,0 +1,136 @@
...
+          different = true;   // values don't match
+        } else {
+          dupe.tags[k] = val1;
+        }
+      }
+    });
+    if (different) return false;
+
+    // if any of these tags don't match, *and one is present*, it's not a duplicate..
+    ['addr:door', 'addr:unit', 'addr:flats', 'addr:floor'].forEach(k => {
```
Nakaner, line 56: Shouldn't you swap this list of keys with the list in line 22 (`['addr:street', 'addr:place', 'addr:block']`)? Otherwise two objects with identical house number but one with `addr:street=ABC Street` and the other one with `addr:place=DEF Farm` are not detected as different.

Nakaner:
Thank you for your explanations, @bhousel. I think that your implementation is better than the old one.

I expect no duplicate to be detected for the following test case: `addr:street=ABC Street` + `addr:housenumber=1` vs. `addr:place=ABC Farm` + `addr:housenumber=1` However, your implementation does not detect the two as different because it takes street and place only into account if any of these tags is set for both objects. IMHO, a value being set for one key and not being set for another key makes it different to me.

Did you intend to treat the following case as duplicate: `addr:street=ABC Street` + `addr:housenumber=2` + `addr:floor=4` + `office=it` + `name=Fast Computing Ltd.` vs. `addr:street=ABC Street` + `addr:housenumber=10` + `building=office` + `building:levels=8` + `name=Office Tower`

bhousel:
> Did you intend to treat the following case as duplicate: `addr:street=ABC Street` + `addr:housenumber=2` + `addr:floor=4` + `office=it` + `name=Fast Computing Ltd.` vs. `addr:street=ABC Street` + `addr:housenumber=10` + `building=office` + `building:levels=8` + `name=Office Tower`

no, because the `addr:housenumber` are different, it shouldn't trigger a warning.

Nakaner:
> > Did you intend to treat the following case as duplicate: `addr:street=ABC Street` + `addr:housenumber=2` + `addr:floor=4` + `office=it` + `name=Fast Computing Ltd.` vs. `addr:street=ABC Street` + `addr:housenumber=10` + `building=office` + `building:levels=8` + `name=Office Tower`
> 
> no, because the `addr:housenumber` are different, it shouldn't trigger a warning.

Sorry, I should read my comments before I post them.

The correct test case is (identical house number, otherwise it's a bit too simple): `addr:street=ABC Street` + `addr:housenumber=2` + `addr:floor=4` + `office=it` + `name=Fast Computing Ltd.` vs. `addr:street=ABC Street` + `addr:housenumber=2` + `building=office` + `building:levels=8` + `name=Office Tower`

bhousel:
> I expect no duplicate to be detected for the following test case: `addr:street=ABC Street` + `addr:housenumber=1` vs. `addr:place=ABC Farm` + `addr:housenumber=1` However, your implementation does not detect the two as different because it takes street and place only into account if any of these tags is set for both objects. IMHO, a value being set for one key and not being set for another key makes it different to me.

I agree - it should not trigger a warning in that case.

The intent of the code is to trigger a warning for:
`addr:housenumber=1` and `addr:street` not set
vs.
`addr:housenumber=1` and `addr:street=Something`

We probably should also trigger a warning if street and place are the same thing:
`addr:housenumber=1` and `addr:street=Something`
vs.
`addr:housenumber=1` and `addr:place=Something`

bhousel:
> The correct test case is (identical house number, otherwise it's a bit too simple): `addr:street=ABC Street` + `addr:housenumber=2` + `addr:floor=4` + `office=it` + `name=Fast Computing Ltd.` vs. `addr:street=ABC Street` + `addr:housenumber=2` + `building=office` + `building:levels=8` + `name=Office Tower`

These would turn up different because one of them has an `addr:floor=4` (only one needs to have it)

bhousel:
I've pushed a few additional commits to address the feedback we received - thanks!:

- [x] Consider `addr:city` and `addr:postcode` tags when comparing addresses
- [x] Perform case insensitive address comparisons
- [x] Add duplicate address tests 
- [x] Handle `addr:street` and `addr:place` tags interchangeably

quincylvania:
@bhousel Thanks for opening this! It'd be great to get a version of it into main iD.

I loaded this branch up in the Netherlands, a country with extreme address coverage, and received a ton of warnings. [This building](https://www.openstreetmap.org/way/272582963) alone triggers 28 issues, since it only has an `addr:street` tag which matches that of 28 nearby features. I'd say iD could be more generous here and only worry about features with house numbers. What do you think?

I haven't looked at the code yet but I figured I'd bring things up as I discover them.

bhousel:
> I loaded this branch up in the Netherlands, a country with extreme address coverage, and received a ton of warnings. [This building](https://www.openstreetmap.org/way/272582963) alone triggers 28 issues, since it only has an `addr:street` tag which matches that of 28 nearby features. I'd say iD could be more generous here and only worry about features with house numbers. What do you think?

Ah yeah, we tried it out in Zanzibar and the same thing happened.  It was actually a pretty impressive stress test.

Screenshot 2020-07-07 15 47 26

I'll rewrite it to work more like the "disconnected path" validation, where it raises a single issue and attaches it to `n` entities, rather than creating `n²` issues.

kaligrafy:
What is the progres on this? should it be bumped or closed?

PR #7742, opened 2020-06-26
Fix reference in help to renamed offset UI section

mikini:
Use current text of locale key:background.fix_misalignment in
key:help.imagery.offset (renamed in
11035127a34c125f844245e279c736c8ed0d5810).


quincylvania:
@mikini Thanks!

mikini:
You`re welcome, just stumbling upon these things while working on the Danish translation.

But thinking it ought to be possible somehow to reference other keys like this to pull them in from the translation automatically. But I see this might become rather complex requiring circular reference detection and such.

quincylvania:
> But thinking it ought to be possible somehow to reference other keys like this to pull them in from the translation automatically.

It'd be tough to do this automatically, but straightforward to do manually in code. Definitely worth considering in order to avoid mismatches like this.

Issue #6620, opened 2019-07-04
support handrail:left/right/center


Lukas458:
What's about to give the handrail's position with handrail:right=yes/no, handrail:left=yes/no and sometimes handrail:center=yes[/no]?

See the usage statistics at https://wiki.openstreetmap.org/wiki/Key:handrail

maybe it could be possible to create a handrail - position field for highway=steps like the one for the cycleway on highways?

PR #7747, opened 2020-06-27
Created the field "Handrail" for #6620

Elefant-aus-Wuppertal:
See #6620 - created a field for supporting handrail:right/center/left, as the one for "cycleway" is.

### DIFF
--- a/data/presets/fields/handrail
+++ b/data/presets/fields/handrail
@@ -1,0 +1,22 @@
+{
+    "keys": ["handrail:left", "handrail:center", "handrail:right"],
+    "type": "handrail",
+    "placeholder": "unknown",
+    "strings": {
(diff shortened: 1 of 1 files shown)


quincylvania:
@Lukas458 The cycleway field has custom logic and can't be duplicated like this. `handrail` isn't a supported field type so this should have thrown an error when building the data.

Elefant-aus-Wuppertal:
Oh sorry, I didn't know cycleway is a special field. Hmm... What do we gonna do now about this?

quincylvania:
@Lukas458 The best thing would probably be to generalize the cycleway field to make it work for similar data.

Elefant-aus-Wuppertal:
Yeah, I would like that.

Issue #7282, opened 2020-01-22
Pasting an OSM link in the search bar should detect the way/node/rel in that url


thibaultmol:
Being able to just paste an osm link and have it realize what it means would be convenient 
For example "https://www.openstreetmap.org/way/323660491" should just show that way
If someone sends you a link, you wouldn't have to extract the id from the url. You can just paste the whole thing in the search box from your clipboard

boothym:
You can already paste in w/323660491 and iD recognises it as a way (same with n or r). Perhaps iD could strip the OSM url and then the full word "way/node/relation" could also be detected?

As an aside, it would be good if "note/123456789" was recognised by iD so you could zoom to a particular note.

kymckay:
Resolved in #7775

PR #7775, opened 2020-07-09
Flexible Element Search

### LINKED ISSUE #7627: More flexible format in 'Search Features' when searching by element id
Currently, User can search for a specific feature with a 'Search Features' query like `n1234`. When I copy an element from JOSM (or other places) I get a string like `node 1234`. But when I search in iD for this string, I get nothing. When I search for `1234` it offers `node 1234` as an option.

I suggest supporting search queries of the form `node 1234`.

After looking at the code...
https://github.com/openstreetmap/iD/blob/63aa47f02f06534eb3311ec869659cfdffc74f42/modules/ui/feature_list.js#L125
 it could use the regex
`/^(node|way|relation|[nwr]) ?([0-9]+)$/i`
and in four places immediately blow, replace `idMatch[1]` with `idMatch[1][0]`

But this would only support English. A step further would be to get the user's current language, and support the localized word for each element type.

blackboxlogic:
Fix https://github.com/openstreetmap/iD/issues/7627
Accept a variety of formats in Feature Search when specifying an element by type and id.
For example: n123, n 123, node123, node 123 or even https://www.openstreetmap.org/node/123/history#map=19/43.66140/-70.25415

### DIFF
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -122,11 +122,11 @@ export function uiFeatureList(context) {
 
             if (!q) return result;
 
-            var idMatch = q.match(/^(.*[^\w])?(node|n|way|w|relation|r)[^\w]?([0-9]+)([^\w].*)?$/i);
+            var idMatch = q.match(/(?:^|\W)(node|way|relation|[nwr])\W?0*([1-9]\d*)(?:\W|$)/i);
 
             if (idMatch) {
-                var elemType = idMatch[2][0];
-                var elemId = idMatch[3];
+                var elemType = idMatch[1].charAt(0);
+                var elemId = idMatch[2];
                 result.push({
                     id: elemType + elemId,
                     geometry: elemType === 'n' ? 'point' : elemType === 'w' ? 'line' : 'relation',
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -125,8 +125,8 @@ export function uiFeatureList(context) {
             var idMatch = q.match(/^(.*[^\w])?(node|n|way|w|relation|r)[^\w]?([0-9]+)([^\w].*)?$/i);
 
             if (idMatch) {
-				var elemType = idMatch[2][0];
-				var elemId = idMatch[3];
+                var elemType = idMatch[2][0];
+                var elemId = idMatch[3];
                 result.push({
                     id: elemType + elemId,
                     geometry: elemType === 'n' ? 'point' : elemType === 'w' ? 'line' : 'relation',
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -122,14 +122,16 @@ export function uiFeatureList(context) {
 
             if (!q) return result;
 
-            var idMatch = q.match(/^([nwr])([0-9]+)$/);
+            var idMatch = q.match(/^(.*[^\w])?(node|n|way|w|relation|r)[^\w]?([0-9]+)([^\w].*)?$/i);
 
             if (idMatch) {
+				var elemType = idMatch[2][0];
+				var elemId = idMatch[3];
                 result.push({
-                    id: idMatch[0],
-                    geometry: idMatch[1] === 'n' ? 'point' : idMatch[1] === 'w' ? 'line' : 'relation',
-                    type: idMatch[1] === 'n' ? t('inspector.node') : idMatch[1] === 'w' ? t('inspector.way') : t('inspector.relation'),
-                    name: idMatch[2]
+                    id: elemType + elemId,
+                    geometry: elemType === 'n' ? 'point' : elemType === 'w' ? 'line' : 'relation',
+                    type: elemType === 'n' ? t('inspector.node') : elemType === 'w' ? t('inspector.way') : t('inspector.relation'),
+                    name: elemId
                 });
             }
 


blackboxlogic:
I ran `npm test` and got errors that didn't look related to my changes.

modules/ui/feature_list.js
```diff
@@ -122,14 +122,16 @@ export function uiFeatureList(context) {
 
             if (!q) return result;
 
-            var idMatch = q.match(/^([nwr])([0-9]+)$/);
+            var idMatch = q.match(/^(.*[^\w])?(node|n|way|w|relation|r)[^\w]?([0-9]+)([^\w].*)?$/i);
```
kymckay: Regarding the regex, I'm not sure I see the point in matching from the start, `^`, to the end, `$`, if between those and the `elemType` we're matching anything, `.*` followed by a non-alphanumeric character, `[^\w]`, all of which doesn't even have to be present because of the `?`.

For the same results, you could simplify to:
```suggestion
            var idMatch = q.match(/(node|way|relation|[nwr])\W?(\d+)/i);
```

Also any reason we're limiting to only a single non-alphanumeric character in-between the number and type (`?` instead of `*`)?

modules/ui/feature_list.js
```diff
@@ -122,14 +122,16 @@ export function uiFeatureList(context) {
...
 
             if (idMatch) {
+                var elemType = idMatch[2][0];
```
kymckay: I assume this is intending to take the first character of the string (index 0)?

If so, this is invalid syntax in Javascript (I think...it's been a while), I'd use `.charAt(0)`.

modules/ui/feature_list.js
```diff
@@ -122,14 +122,16 @@ export function uiFeatureList(context) {
...
+                var elemId = idMatch[3];
```
kymckay: If you were to change to my simplified regex as suggested these two indexes both drop 1 (so this becomes 2 and the other becomes 1).

modules/ui/feature_list.js
blackboxlogic: Regarding `^(.*[^\w])?`...
My concern was, while increasing the match-ability, I should attempt to avoid false positives like:
"Doctor360" offering "Relation 360"
I achieved this by requiring the element type (in this case `r`) to be the start of the string, or immediately after _any_ delimiter. Perhaps that could have been simplified: `(^|[^\w])`?
In your opinion, is avoiding false positives:
* Not worth the complexity and cpu time
* Should be achieved a different way (if yes, what do you think of `(^|[^\w])`?)

modules/ui/feature_list.js
blackboxlogic: I'm not experience in JS, thanks for pointing this out. I'll commit a change (including this) with the result of your other feedback.

modules/ui/feature_list.js
blackboxlogic: Oh, and keep in mind, I'm also trying to match `https://www.openstreetmap.org/node/101828281/history`, from https://github.com/openstreetmap/iD/issues/7282, so trailing stuff is ok, as long as there is a delimiter first. Perhaps it should end with `($|[^\w])`

So, what do you think of:
```js
var idMatch = q.match(/(^|[^\w])(node|way|relation|[nwr])[^\w]?(\d+)([^\w]|$)/i);
```

kymckay: Avoiding false positives is probably worth it so users aren't getting unexpected search results (UX takes priority especially because there's basically no complexity/overhead added). I understand the intention now, then yeah in that case I'd go with

```suggestion
            var idMatch = q.match(/(?:^|\W)(node|way|relation|[nwr])\W?(\d+)(?:\W|$)/i);
```

Note the non-capturing groups before and after which means those won't be returned in the array (so relevant indexes will be 1 and 2).

kymckay:
We can add some tests for all the cases we want to match/don't match too so that this improved functionality never gets accidentally broken 👍 

Don't worry about that in this PR though, I'd be happy to do that kind of maintenance after merge

modules/ui/feature_list.js
blackboxlogic: That looks great, I hadn't used non-capturing groups before. Slightly adjusted to account for the case of leading zero: match and use `node 0123`, don't match `node 000`, which is [supported](https://www.openstreetmap.org/api/0.6/node/0123) by the API, apparently.
```js
var idMatch = q.match(/(?:^|\W)(node|way|relation|[nwr])\W?0*([1-9]\d*)(?:\W|$)/i);
```

blackboxlogic:
I wanted to add tests, but didn't know where to put them, or how best to extract the function for testing. Thanks for handling it.
Positive Cases:
n1, N1, n01, n11, w1, r1, node1, way1, relation1, n 1, node 1, ?n1, n?1, n1?, ?n?1, ?n1?, n?1?, ?n?1?, 1?n?1?n, n?1?1, n?n?1
www.example.com/node/1/history?lat=100&lon=100
Negative Cases:
nn1, n1n, 1n1, 1n, n??1, n0, n, 1, [empty string]

quincylvania:
@blackboxlogic Thanks for working on this! Welcome to the dev side of iD 🔨

@SilentSpike Thanks a bunch for taking an interest. The code looks reasonable to me at first glance, but I'll defer to you. Feel free to merge and tweak this as needed. (Tests are always stellar too, since you mentioned them.)

Oh, and if anyone's feeling inspired it might be a good idea to break out some of this functionality into separate functions instead of having everything in `features()`.

kymckay:
@blackboxlogic I've tested this and it's good to go, will merge as is now 👍

If you'd like to learn how to add the tests yourself, let me know and I an point you in the right directions 😃 If not, I don't mind taking care of it.

Issue #7738, opened 2020-06-25
add some more keys to the post_box-preset


Elefant-aus-Wuppertal:
Hi,
I think for amenity=post_box, there could be added some more things to its preset:
- how is the post box supported (support=*)
- how is the post box constructed (post_box:type=*, see https://wiki.openstreetmap.org/wiki/Key:post_box:type)
- what colour is it (colour=*)
- what material is it (material=*)

Samarath:
I would like to work on this.

quincylvania:
@Samarath Great, welcome! Check out our [contributing guide](https://github.com/openstreetmap/iD/blob/develop/CONTRIBUTING.md) and the [presets readme](https://github.com/openstreetmap/iD/blob/develop/data/presets/README.md). Let me know if you have any questions.

manvisingh07:
I am interested, I would like to work on this.

Adamant36:
@Lukas458 I was looking at the usage for post_box:type and they all seem to be about the type of support. So, do you know how it would be different then the support tag? Because it looks like they are used the same to me.

Elefant-aus-Wuppertal:
Hmm, okay. In my thoughts there would be a difference between these keys. When you look at post_box:type's wiki page, you can for example find the value "lamp". A post box "lamp" is the kind of its shape to me. Here is what we call "lamp" as a post box in Germany: https://wiki.openstreetmap.org/wiki/File:BRIEFKASTEN_DER_DEUTSCHEN_POST_in_BREITH%C3%9CLEN.JPG

As support, the picture shows a pole. But the support of the "lamp" post box could also be a wall, please look at it in this pic: https://wiki.openstreetmap.org/wiki/File:Gelber_Briefkasten_der_Deutschen_Post.JPG

The same thing for post_box:type=pillar. The post box can be mounted onto the ground directly, but in Germany there's often a little concrete pedestal: https://wiki.openstreetmap.org/wiki/File:Briefkasten_Gie%C3%9Fen_2.jpg
A little one, but its there.

Adamant36:
[...]

PR #7777, opened 2020-07-10
added colour, material, and support to post_box preset

yyazdi13:
Added more fields to the post_box preset, in relation to issue #7738. Did not include post_box:type because it seems like it was decided against for now.


quincylvania:
@yyazdi13 Looks good, thanks a lot! Welcome to the project

PR #7779, opened 2020-07-12
Model Shop Preset Added

nisargshh:
Added a model shop present when adding for a point or area. Has the same icon as Hobby.

### DIFF
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -8230,6 +8230,11 @@ en:
         # shop=mobile_phone
         name: Mobile Phone Store
         terms: '<translate with synonyms or related terms for ''Mobile Phone Store'', separated by commas>'
+      shop/model:
+        # shop=model
+        name: Model Shop
+        # 'terms: manga,figurine,hobby'
+        terms: '<translate with synonyms or related terms for ''Model Shop'', separated by commas>'
       shop/money_lender:
         # shop=money_lender
         name: Money Lender
--- a/data/presets.yaml
+++ b/data/presets.yaml
@@ -8233,7 +8233,7 @@ en:
       shop/model:
         # shop=model
         name: Model Shop
-        # 'terms: manga,figurine,hobby'
+        # 'terms: Toys,Games,Crafts'
         terms: '<translate with synonyms or related terms for ''Model Shop'', separated by commas>'
       shop/money_lender:
         # shop=money_lender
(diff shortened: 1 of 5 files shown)


Adamant36:
Its questionable that manga should be a term. This tag is specifically for model shops. Not other types of shops that might sell models as a side thing. Manga related toys aren't really models as defined by the tag anyway. Same goes for figurines. There's actually shop=collector for that type of stuff already. Although, I don't think its supported in iD Editor yet. Anyway, this tag is purely confined to scale models.

nisargshh:
Could give some examples of tags that would be better suited for model shop?

Adamant36:
I can't think of any off the top of my head. Maybe look at some locations on the map and see if there are any similarities in their names or something.

nisargshh:
I edited the tags and the icon. @quincylvania Was this a relevant change? If not would going back be relevant or would it face the issue of what @Adamant36 was saying.

Adamant36:
I still don't think the terms work. Model shops don't really sell games and people who put together models don't consider them toys. Terms should be relevant to what people are going to be searching for and I don't think either of those are. Personally, I wouldn't do a search for either in iD Editor when adding a model shop, but people searching for a games store preset will just go with shop=model when it comes up because that's what searching for "games" comes up with. Same goes for crafts. What's wrong with just going with something like "model shop" for the preset and calling it at that? I think that would be fine. There isn't really any more specific synonyms for it that work IMO and there really doesn't have to be as long as people can find the preset when adding model shops. 

BTW, I'm just asking and it's only general thoughts, but the Wiki is pretty clear that this should only be used for a specific thing and for instance the shop=shoes preset doesn't come up when you search for "clothes" in iD Editor when you search for it. Even though clothing store and a shoe store are more related then a gaming store and a model shop.  Plus, if it's that un-clear where and how this tag should be used maybe there shouldn't be a preset for it at this point. At least until it's discussed and what it actually applies to is better hashed out.

quincylvania:
Thanks a lot, @nisargshh, welcome to the project!

@Adamant36 I know you're trying to be helpful, but it feels discouraging when you go into that much detail critiquing a minor aspect of someone else's work. The terms don't have to be perfect; they aren't shown in the UI and it's easy to update them later. The important thing is supporting new contributors.

Adamant36:
@quincylvania sorry if it came off that way. It wasn't how I intended it to.

quincylvania:
@Adamant36 Cool, no worries.

boothym:
@nisargshh didn't notice this PR until now - thanks for adding the preset! 😄 

@quincylvania and thanks for updating the terms, those look good

nisargshh:
@boothym Thanks, looking to do some more contributions.

boothym:
Actually, it would be good if shop=craft had the term "hobby" as well because that can also be called a hobby shop.

https://github.com/openstreetmap/iD/blob/72a99673a0328ba4f44c58dd110797cc733f7007/data/presets/presets/shop/craft.json

PR #7811, opened 2020-07-20
Craft Shop

nisargshh:
As mentioned in #7779 add the terms hobby to craft shop

### DIFF
--- a/data/presets/presets/shop/craft.json
+++ b/data/presets/presets/shop/craft.json
@@ -1,7 +1,7 @@
 {
     "icon": "fas-cut",
     "geometry": [
-        "point",
+        "point", 
         "area"
     ],
     "tags": {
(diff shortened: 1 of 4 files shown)


data/presets/presets/shop/craft.json
```diff
@@ -1,16 +1,9 @@
 {
     "icon": "fas-cut",
-    "geometry": [
-        "point",
-        "area"
-    ],
+    "geometry": ["point", "area"],
```
quincylvania: It's not a big deal but please leave arrays on multiple lines if that's how they're already formatted.

nisargshh: My bad I thought the lint would fix that.

quincylvania: We don't autoformat the preset JSON at the moment…

nisargshh:
Commited requested changes

quincylvania:
@nisargshh Thanks!

Issue #7467, opened 2020-03-26
Search features box should trim spaces


boothym:
As per title - sometimes when copying nodes/way/relation IDs from overpass-turbo, I select a leading or following space and include it when pasting. Would be good if these were trimmed.

kymckay:
I don't think the actual string in the UI should be trimmed, however searching would benefit from stripping out white space for the actual search (for better fuzzy matching)

boothym:
Perhaps it can be restricted to numeric strings then? Because if I paste in " 107775" or "107775 " (minus the quotation marks) then iD doesn't display the node/way/relation links.

kymckay:
Apologies, I think I wasn't super clear.

My point is the actual UI string should not change, but for actual searching it could be stripped of whitespace. This would solve the issue here, but also for cases where a user may search for "Pizza Express" when the preset is "PizzaExpress", etc.

NateGrobe:
Hi. I'm new to open source and was wondering if I could work on this issue?

quincylvania:
@NateGrobe Welcome, go ahead and give it a shot! The [modules/ui.feature_list.js](https://github.com/openstreetmap/iD/blob/develop/modules/ui/feature_list.js) is probably the place for this

NateGrobe:
I'm having a small issue getting the project to run. I'm getting the error:
Error: Could not load ~/iD/node_modules/es5-ext/array/index.js#/clear (imported by node_modules/ds6-set/polyfill.js)

I checked the imports in es6-map/polyfill.js and they seem fine. The import is as follows:
var clear = require('es5-ext/array/#/clear')

Switching it to an es6 imports didn't fix the issue. I'm just not sure why index.js is being added seemingly randomly to the import path. 

I tested this on Windows 10 Pro and PopOs 20.04 and got the same error both times.

quincylvania:
@NateGrobe Hmm I haven't seen that issue before. I assume you went through all the [installation](https://github.com/openstreetmap/iD#installation) steps? Make sure you're running Node 10, 12, or 14. It'd be easier to discuss if you care to join our [slack](https://osmus-slack.herokuapp.com/).

nisargshh:
I've been greeted with the same issue when cloning the project again running `npm run all` My node version is v12.16.2.

quincylvania:
@nisargshh I tried a clean install and am been getting this issue now too. Something must have changed with an in-range dependency but I haven't figured out what yet.

[...]

PR #7816, opened 2020-07-20
added function to strip whitespace from searches

NateGrobe:
This is my solution for issue #7467

### DIFF
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -106,9 +106,9 @@ export function uiFeatureList(context) {
             drawList();
         }
 
-      function stripWhitespace(val) {
-        return val.split('').filter(char => char !== ' ').join('');
-      }
+        function stripWhitespace(val) {
+          return val.split('').filter(char => char !== ' ').join('');
+        }
 
         function mapDrawn(e) {
             if (e.full) {
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -107,7 +107,7 @@ export function uiFeatureList(context) {
         }
 
         function stripWhitespace(val) {
-          return val.split('').filter(char => char !== ' ').join('');
+            return val.replace(/ /g, '');
         }
 
         function mapDrawn(e) {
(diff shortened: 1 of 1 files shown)


secan:
Hi,
may I suggest you a simpler approach to solve the issue? Rather than splitting the string into an array of single characters, filtering out the white spaces and re-joining the array of characters in a single string, you could simply use the JavaScript bult-in function `String.prototype.replace()` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)

```javascript
function stripWhitespace(val) {
    return val.replace(/ /g, '');
}
```

NateGrobe:
Hey,
I agree that replace is a better solution for this.
Out of curiosity, is there any advantage to using regex over a simple whitespace like `return val.replace(' ', '');`?

secan:
If you use  a string (e.g. `' '`) only the first occurrence will be replaced while the regex allows you to replace all the occurrences (the "g" in `\ \g` stands for "globally").

> Out of curiosity, is there any advantage to using regex over a simple whitespace like `return val.replace(' ', '');`?

NateGrobe:
Oh ok makes sense. Thanks!

secan:
You're welcome

quincylvania:
@NateGrobe Thanks for working on this! Sorry for the delay in responding, no one had been working on iD full-time for awhile.

Unfortunately, I merged this prematurely and had to revert. In testing, I noticed there's an issue with the original task described in #7467. If we only strip whitespace from the query and not the test strings, then it makes it impossible to search for things with whitespaces, like "South Broad Street". If you're still interested in working on iD feel free to open a new PR and we can work through a solution 🙂

@secan Thanks for the assistance here.

ahmedragab96:
Hello @quincylvania, I am new to open source contributing and I'm interested to work on fixing this with you if possible.

quincylvania:
@ahmedragab96 Thanks, but I think this is something @NateGrobe or myself should take care of. There are plenty of [other issues](https://github.com/openstreetmap/iD/issues) you can check out!

PR #7832, opened 2020-07-22
Add portapotty preset

### LINKED ISSUE #7783: Add portapotty preset
I think the correct tags for a portapotty are `amenity=toilets` and `toilets:disposal=chemical`, but I'm not smart enough to remember this.

Actual: Type in 'portapotty' and come up with nothing. Try the 'toilets' preset, then frown at the refinement options, since I'm not sure if this only applies to indoor facilities.

Expected: Select the 'portapotty' preset and be happy.

Nimisha94:
Added a portapotty preset. Fixes #7783


data/presets/presets/amenity/toilets/disposal/portapotty.json
```diff
@@ -0,0 +1,31 @@
...
+    "tags": {
+        "amenity": "toilets",
+        "toilets:disposal": "chemical",
+        "portable": "yes"
+    },
+    "reference": {
+        "key": "toilets:disposal",
+        "value": "chemical"
+    },
+    "name": "Porta Potty"
```
quincylvania: How about calling this by the generic term "Portable Toilet"? There are loads of slang and regional names we can put in `terms` later.

data/presets/presets/amenity/toilets/disposal/portapotty.json
```diff
@@ -0,0 +1,31 @@
...
+        "restroom",
+        "privy",
+        "portable toilet",
+        "chemical toilet"
+    ],
+    "tags": {
+        "amenity": "toilets",
+        "toilets:disposal": "chemical",
+        "portable": "yes"
+    },
```
quincylvania: Could use make this the `addTags` property and just use `amenity=toilets` + `portable=yes` for `tags`? This will mean the `toilets:disposal=chemical` tag won't be required to match features to the preset, but will be added when the feature is created.

quincylvania:
Hi @Nimisha94. Thanks for waiting, iD development had been on hold for awhile. Could you address the couple things I pointed out?

quincylvania:
@Nimisha94 Looks great, thanks a lot! Welcome to iD development 🚀

Issue #7772, opened 2020-07-08
Can't undo to the last vertex of a polygon


jidanni:
On https://ideditor.netlify.app/ start making an area.
Now undo the steps one by one with CTRL+Z.

You will see that with a polygon with vertices
A B C D E
you can't just undo down to just
A
because any further undoing beyond
A B
with just zap the whole thing.

PR #7837, opened 2020-07-24
Undo Fix

nisargshh:
Fixed the bug faced in issue https://github.com/openstreetmap/iD/issues/7772

### DIFF
--- a/modules/core/history.js
+++ b/modules/core/history.js
@@ -204,9 +204,8 @@ export function coreHistory(context) {
 
             var previousStack = _stack[_index];
             var previous = previousStack.graph;
-            while (_index > 0) {
+            if (_index > 0) {
                 _index--;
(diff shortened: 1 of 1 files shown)


quincylvania (changes requested):
@nisargshh Thanks for looking into this. This solution affects all undo operations, and would unfortunately break some functionality. [This file](https://github.com/openstreetmap/iD/blob/develop/modules/behavior/draw_way.js) is probably what needs updating instead, since only way drawing is affected.

nisargshh:
I was not able to find any other solution, the undo function is undoing the vertices and not the nodes and could not find a way to undo the nodes instead.

quincylvania:
@nisargshh That's okay, thanks for trying! This is a surprisingly complicated area of the code.

Issue #7342, opened 2020-02-07
Mapillary Date Filter


zlavergne:
I'm working on adding a date filter to the Mapillary layer. The plan is to use a calendar date picker similar to the Mapillary web app. I'm thinking it will live inside the current Mapillary box of the `Photo Overlays` list. 

I should have an initial version next week to show off and make sure we're all on the same page of how this will look.

tordans:
Hi @zlavergne this sounds like https://github.com/openstreetmap/iD/issues/4518, right?

Personally, I will be very happy with any kind of filtering!

Just as a suggestion, I wonder if a calendar date picker is even the easiest (first) step.
My personal use cases are all about reducing the clutter of too many images, like …
- only show recent images – last 30 days, last 12 month, last 3 years
- only show images from specific users (only me, only someone; this is https://github.com/openstreetmap/iD/issues/5307)
- ideally there where a "only show images that are taken from sidewalks, not streets – but AFAIK this is not a filter the API provides

But back to the date picker: In my mind the pre defined selections would be faster to choose than the date picker and less complexity. – Depending on the UseCase you have in mind of course.

cbeddow:
JOSM uses this:

> only show recent images – last 30 days, last 12 month, last 3 years

This we've found is not feasible for many users, who need a specific date ranger.

> only show images from specific users

The username filter will be great, but lower demand. 

> ideally there where a "only show images that are taken from sidewalks, not streets

A lot of backend on the Mapillary side to first classify the images this way, based on content and GPS, not reliable to do for now. :) But would be really good.

quincylvania:
#4518 is indeed photo date filtering, but I'll leave this open since Zach seems to be working on something specific to Mapillary.

> >     only show recent images – last 30 days, last 12 month, last 3 years
> 
> This we've found is not feasible for many users, who need a specific date ranger.

I could see how mapping teams might need to see a specific date range, but most iD users just need the photo overlay filtered down to a manageable amount of data. Date pickers are finicky and aren't useful if you don't know beforehand what dates to enter. Filtering by the previous 30/90/365 days is simpler but is blind to the actual data clustering.

I think my preferred solution is a dropdown for filtering to the most recent 1/2/3/5/10 capture sequences in the visible map area. This could be set to something low by default so mappers don't get sequence spaghetti right away. Then we could consider date filtering (or even just filtering to specific sequences) as an advanced option.

cbeddow:
We will test some prototypes and share results!

[...]

Issue #5307, opened 2018-09-12
Show just one single user's Mapillary images


koninklijke:
Mapillary has collected many millions of photos, but it can be hard to pick the ones one contirbutor took of a particular feature from amongst the many green markers shown. This is irrespective whether those would be my own, or pics from someone whom I know visited that area and took pics of that particular feature, and would be as opposed to the standard straight-ahead traffic-sign stuff which may show that there is, for example a regional walking network marker, but lacks the detail which direction to go in order to reach which particular other node. 
That filter is already available on the Mapillary site, this request is to implement it when overlaying Mapillary photos in iD editor, please.

nickplesha:
Hi, you can try the username filter in this PR: https://github.com/openstreetmap/iD/pull/7847 let me know if you have any feedback or suggestions.

quincylvania:
Done by @nickplesha in #7847, followup in #8076.

You can now filter photos to those by one or more specific users for all layers: Mapillary, OpenStreetCam, and even Bing Streetside if you somehow wanted to 😄

Issue #4518, opened 2017-11-09
Date filter for street level imagery trips


mvexel:
Broken out from #4512 

Density of OSC / Mapillary traces can be prohibitively high in places. A date filter would be nice to focus on newer images. 

Here's how we do it in the OSC JOSM plugin:

![osc-josm-filter](https://user-images.githubusercontent.com/120452/32613518-58151fb0-c528-11e7-8945-b6eac8b7e2de.gif)

1ec5:
I wonder if iD could automatically filter by date (or by age?) when the traces get too dense, similar to the feature filters that turn on automatically when (for instance) you view an area of the map with a certain number of buildings.

davebrny:
yea some places are like a jackson pollock painting so any type of filter would be a great help. by date, by user or even filtering out certain users

joecaruso:
Are there any update on this?  In my area (Raleigh, NC) there's starting to be a lot of coverage with both Mapillary and OSC, which is great!  The problem is that there are so many trips that iD is starting to become unusable in certain areas.  I now have to open a separate window with the Mapillary website to use as a viewer (where I can do filtering), and then move them in tandem as I make edits.

nickplesha:
Hi, you can try the date filter in this PR: https://github.com/openstreetmap/iD/pull/7847 let me know if you have any feedback or suggestions.

quincylvania:
Done by @nickplesha in #7847, followup in #8076. ✨📸✨

You can now filter by exact dates. I'd like to add another option later to just show the "most recent", whatever that may be, but this is already a big step forward.

Issue #6398, opened 2019-05-21
Read photo id from query string for street-level imagery 


cbeddow:
Example:

`https://www.openstreetmap.org/edit?editor=id#map=18/39.93450741752976/-105.0910279993947?photo_overlay=mapillary&photo=o6y6pgnHjz7vdEbjY7YCtg`

This link, for example, would make sure the Mapillary photo overlay is turned on, and then initiate the MapillaryJS viewer and execute `moveToKey('o6y6pgnHjz7vdEbjY7YCtg')` which would open that specific image, so a user can start editing right away.

planemad:
This is a great suggestion. I find the Mapillary explorer pretty powerful to navigate through various imagery/filters and would love to jump to iD from a specific Mapillary image and retain context of what i wanted to edit.

cbeddow:
Yes if we can get this working then we'll link directly from Mapillary's website to open the same current photo in iD, ready to edit :)

EuropeanWaterProject:
This would be  very helpful... has any progress been made on this ?

quincylvania:
> has any progress been made on this ?

@EuropeanWaterProject Not yet…

quincylvania:
@nickplesha implemented this for Mapillary photos in #7885!

I'm going to work on achieving parity for OpenStreetCam and Bing Streetside. As part of supporting multiple services, I prefixed the format like `photo=mapillary-eZxLgJ9sHstg0ySKE3plLw`. We can't rely on `photo_overlay=mapillary` since that key supports multiple layers, e.g. `photo_overlay=streetside;mapillary;openstreetcam`. Plus this lets iD open the correct layer automatically if `photo_overlay` is left off.

tordans:
@quincylvania I know I am late to the party, just a quick though: Reading `photo=mapillary-eZxLgJ9sHstg0ySKE3plLw` makes me think of an ID that is actually `mapillary-eZxLgJ9sHstg0ySKE3plLw`. But reading your comment, its more of a identifier/category+ID which IMO could be shown a bit better by something like `photo=mapillary/eZxLgJ9sHstg0ySKE3plLw` or `photo=mapillary;eZxLgJ9sHstg0ySKE3plLw` or `photo=mapillary+eZxLgJ9sHstg0ySKE3plLw`.

quincylvania:
@tordans Hmm, that's a good point. The slash might work better. We can easily change this since we haven't released it yet.

PR #7847, opened 2020-07-27
Filtering for photo overlay layers

nickplesha:
Hello, this PR adds filtering for photo overlay layers. Included are date and username filters.

The filters are applied to all photo overlays (Mapillary, Openstreetcam and Bing Streetside) at the same time.

Filter UI is in the `Photo Overlays` section. Will become visible when filterable photo layers are turned on.

filters

## Changes
- Added date (from and to) filters for photo layers. Uses default browser date picker (input with type `date`).
- Added username filter for photo layers

Let me know if you have any questions or suggestions. There has been discussion in the linked issues on how the filters should look and behave so I'm open to modifying it based on feedback.

Related issues: https://github.com/openstreetmap/iD/issues/7342 https://github.com/openstreetmap/iD/issues/5307 https://github.com/openstreetmap/iD/issues/4518

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -168,7 +168,8 @@ input[type=search],
 input[type=number],
 input[type=url],
 input[type=tel],
-input[type=email] {
+input[type=email],
+input[type=date] {
     background-color: #fff;
     color: #333;
     border: 1px solid #ccc;
@@ -177,6 +178,11 @@ input[type=email] {
     border-radius: 4px;
     text-overflow: ellipsis;
     overflow: hidden;
+}
+input.list-item-input {
+    height: 20px;
+    padding: 0px 4px;
+    width: 160px;
 }
 .ideditor[dir='rtl'] textarea,
 .ideditor[dir='rtl'] input[type=text],
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -740,6 +740,16 @@ en:
       panoramic:
         title: "Panoramic Photos"
         tooltip: "360° photos"
+    date_filter:
+      fromDate:
+        title: "From"
+        tooltip: "Show photos taken after this date"
+      toDate:
+        title: "To"
+        tooltip: "Show photos taken before this date"
+    username_filter:
+      title: "Username"
+      tooltip: "Show only photos by this user"
   feature:
     points:
       description: Points
--- a/dist/locales/en.json
+++ b/dist/locales/en.json
@@ -928,6 +928,20 @@
                     "title": "Panoramic Photos",
                     "tooltip": "360° photos"
                 }
+            },
+            "date_filter": {
+                "fromDate": {
+                    "title": "From",
+                    "tooltip": "Show photos taken after this date"
+                },
+                "toDate": {
+                    "title": "To",
+                    "tooltip": "Show photos taken before this date"
+                }
+            },
+            "username_filter": {
+                "title": "Username",
+                "tooltip": "Show only photos by this user"
             }
         },
         "feature": {
--- a/modules/renderer/photos.js
+++ b/modules/renderer/photos.js
@@ -9,6 +9,10 @@ export function rendererPhotos(context) {
     var _layerIDs = ['streetside', 'mapillary', 'mapillary-map-features', 'mapillary-signs', 'openstreetcam'];
     var _allPhotoTypes = ['flat', 'panoramic'];
     var _shownPhotoTypes = _allPhotoTypes.slice();   // shallow copy
+    var _dateFilters = ['fromDate', 'toDate'];
+    var _fromDate;
+    var _toDate;
+    var _username;
 
     function photos() {}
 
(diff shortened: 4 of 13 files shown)


nickplesha:
Deployed a preview for testing here: https://id-filtering.netlify.app/

Let me know if you need anything else to merge this. @quincylvania

tordans:
@nickplesha this looks great, thank you.

I am having issues with adding the date by keyboard.
Video: https://share.icloud.com/photos/0EZn0Xn7orcCDNbhbMf4_GTdQ

My goal is to type 1.8.2020 to just get my newest uploads of an area, but the day-segment (first segment) is reset once I add the second date-segement (month).

Location: https://id-filtering.netlify.app/#background=Berlin-2019&disable_features=building_parts,indoor,boundaries,water,rail,pistes,aerialways,power,past_future&map=19.65/52.46765/13.44186&photo_overlay=mapillary

tordans:
This might be off topic for this PR: The logo looks like it is out of place comparted to the text.
Bildschirmfoto 2020-08-10 um 14 21 05

Google Chrome Canary, Mac os.
Location: https://id-filtering.netlify.app/#background=Berlin-2019&disable_features=building_parts,indoor,boundaries,water,rail,pistes,aerialways,power,past_future&map=19.65/52.46765/13.44186&photo_overlay=mapillary

nickplesha:
@tordans Thanks for the bug report, I'll look into it!

The fix for the attribution CSS is in my new PR: https://github.com/openstreetmap/iD/pull/7885

nickplesha:
Fixed in https://github.com/openstreetmap/iD/pull/7847/commits/e324a00e544b434758fed73f4385a74b331046b3

cbeddow:
@quincylvania checking in here, let me know if you think this looks okay or needs further action?

tordans:
@nickplesha I think I have a bug at https://id-filtering.netlify.app/#background=Berlin-2019&disable_features=building_parts,indoor,boundaries,water,rail,pistes,aerialways,power,past_future&map=18.81/52.49935/13.44527&photo_overlay=mapillary

Steps:
Open URL
1. add filter 'tordans' user - wait a sec
2. add filter date from 1.1.2020 - wait a sec

What I see (screenshot) is image circles disappear but should not; the line stays, I can also still navigate images via < and > that are not visible as circles.

image

tordans:
Feature request: Add the filter to the URL. When I open an old URL or share it, it should use the same filters, IMO. The rest of the UI works the same way AFAIK.

quincylvania:
@cbeddow @nickplesha I'm planning to finish up #6398 and then work on merging this. I'd like to get it into the next release.

nickplesha:
Thanks for the bug report @tordans! I see what you reported but after a couple of minutes of waiting the dots do appear. The line is visible sooner as there's not that many sequences there (those are loaded separately from images) but the number of images is very high in that area and the particular images you're filtering for load after a long wait. I can see how this becomes unusable if there's no feedback on the loading process.

I'll look into finding a way to show a loading indicator for Mapillary imagery in the UI that would let users know that some images might still appear after loading. The whole filtering feature is mainly aimed at making it easier to work in areas with super-dense coverage so this is important but I think the loading indicator should be in a separate PR: @tordans @quincylvania @cbeddow any ideas on what the loading indicator should look like?

quincylvania:
@nickplesha Can you give me access to commit to this branch?

tordans:
@nickplesha thanks for looking into it. I checked the Network view for the [URL above](https://id-filtering.netlify.app/#background=Berlin-2019&disable_features=boundaries&map=19.81/52.49967/13.44489&photo_overlay=mapillary) and I think there is something happening there that is not ideal. It looks like, the filter do not apply to the network-calls. All network calls go though but only get filtered afterwards, so they are no used at all(?). Also, the images-calls each are quite slow. Some are around 300ms, but a few are 5-40 seconds before the next one starts loading. It  looks to me like the gain of improving this would be bigger here than having a loading indicator.

About the loading indicator:

It takes a really long time, so IMO a simple spinner is not enough. It would look like it is stuck. Especially if all the loaded data gets filtered so its not visible; it looks like nothing is happening. To work around this, maybe a counter could be added: 
Example:
``` 
[x] Mapillary          <Spinner> 12/31
     [_] Map Objects
     …
``` 

Where the 12 counts up whenever an image-call is done.

nickplesha:
@tordans Thanks for the ideas for the loading indicator, I'll check if the number of pages of results per bbox is available from the API and then we can use that number to calculate progress.

The filter is applied only when drawing images and not to the requests that retrieve them for a couple of reasons:
1. The OSC and Microsoft APIs might not have the filtering capabilities that Mapllary has and the plan is for these filters to apply to all image sources.
2. I was aiming for filtering that would be similar to what Leaflet and Mapbox GL are doing, this allows us to skip creating more requests when filters are changed or turned on/off.
3. Filtering only on drawing allows us to swap the way we load the images entirely while maintaining the same filtering code. This allows us to e.g. replace making those dozens of API calls with simply using our vector tiles and the filtering will still work the same after that.

@quincylvania I'm looking into adding you to the branch, the process for adding contributors has changed a bit on our side and I'm trying to figure it out. If you'd like to move faster let me know if you'd like me to create a PR from this branch to a new branch of your choosing so you can take over.

quincylvania:
@nickplesha There used to be an option when you open a PR to allow commits from maintainers, I'm not sure if that's still possible. Feel free to open a new PR if needed. Another potential option is to add me as a collaborator to the mapillary/iD repo. Targeting a different branch might work but isn't ideal.

PR #7848, opened 2020-07-27
Update core.yaml

manfredbrandl:
Unite severals lines into line 873 to avoid cr/lr.


quincylvania:
@manfredbrandl Great, thanks 👍

PR #7852, opened 2020-07-27
Added Irish Pub preset (`theme=irish`)

amandasaurus:
It's excluded from Ireland, because that doesn't make as much sense.
Visitors to Ireland might falsely tag every pub as an Irish Pub.


quincylvania:
@rory Looks good, thanks!

PR #7859, opened 2020-07-29
Use correct term in help.streetlevel for photo overlays.

mikini:
Help text uses the term "layer" to refer to "Map Data" (key:map_data.title) panel's "Photo Overlays" (key:photo_overlays.title).


quincylvania:
@mikini Thanks!

Issue #7865, opened 2020-07-31
Edit Office presets


maro-21:
Add Website and Phone as default fields. This will improve and speed up editing.

quincylvania:
Done in #7875 by @thibaultmol

PR #7875, opened 2020-08-04
Set Website and Phone as default flags on Office

thibaultmol:
Don't think I've ever made a pull request on iD, but this should be good I think

Adding Website and Phone as default flags instead of optional ones on 'Office' (issue #7865  )


thibaultmol:
@quincylvania could you help me figure out why the build is failing? the only thing I changed were those fields :thinking:

quincylvania:
The CI failure looks unrelated.

quincylvania:
@thibaultmol Looks great, thanks!!

Issue #7883, opened 2020-08-06
Open links in OSM notes in new tab


Spanholz:
For example notes from StreetComplete often have a linked image. These links should be opened in a new tab instead of the current one.

AdityaSharma911:
Can you please assign this issue to me? I will change the target associated with this linked image so that it opens in a blank tab.

jesse-ashmore:
I've been looking into this briefly but I can't find where link tags are constructed. I'm guessing there's some sort of user-input parsing library that's being used to clean any comment data, that also detects and wraps links? Maybe this is coming from the API. I think I might need to use a CSS selector and manually append a `target=_blank` attr for any link tags pointing away from OSM hosts (something that would move your view within the window).

Any ideas? :)

jesse-ashmore:
Okay, so I've gone on the assumption and written a helper function to preprocess any comments for links, adding target=_blank attribute to any which point to 'external domains'.

At the moment, I'm considering any link whose hostname does not include "osm.org" or "openstreetmap.org" as an external domain.

There are 3 things to note with this:
1. There may be more domains than just these two I've considered so far
1. A domain such as foosm.org would be considered non-external, due to containing "osm.org" in the host
1. I don't have a way of differentiating between within-map and within-site links, such as user profile vs. map node. It would probably be preferable to open a within-site but out-of-map link in a new tab. I'd need a full list of all endpoints of either category, based on how I'm implementing it now.

Here's a preview: https://www.youtube.com/watch?v=MFpImKkY9gA

---
Here's what I've written for it (wary of making a PR just yet):

```javascript

// This is all in note_comments.js
mainEnter
  .append('div')
  .attr('class', 'comment-text')
  .html((d) => preprocessLinkTags(d.html));

/**
 * Add target="_blank" to any comment links pointing away from the site
 * That is, any link with domain other than those in ignoredDomains.
 */ 
const preprocessLinkTags = (commentTextHtml) => {
  if (typeof commentTextHtml !== 'string') return commentTextHtml;
  try {
    const template = document.createElement('template');
    const trimmedCommentTextHtml = commentTextHtml.trim();
    template.innerHTML = trimmedCommentTextHtml;

    const links = Array.from(template.content.querySelectorAll('a'));
    const ignoredDomains = ['openstreetmap.org', 'osm.org']; // Possibly more?

    // Filter links based on their hostname not including domains in ignoredDomains
    links
      .filter((link) => {
        const hostname = new URL(link).hostname;
[...]

PR #7893, opened 2020-08-10
Open external links in OSM notes in a new tab

jesse-ashmore:
This addresses #7883.

The solution determines if a link is external using a basic hostname check, and if so, applies the attribute `target="_blank"` to the link tag.

At the moment, a link is external if its hostname contains either `.osm.org` or `.openstreetmap.org`. The prefixed `.` is used to allow for any prefix such as `api06.dev.osm.org` or `www.osm.org`. 

The idea here is that most internal links will probably be pointing to nodes on the map, so opening a second map in a new tab is not ideal. However, if the link is to a non-map route, such as a user profile, then this will unfortunately not open in a new tab.

Open to suggestions regarding implementation, but as a v1 this should cover a lot of cases.

### DIFF
--- a/modules/ui/note_comments.js
+++ b/modules/ui/note_comments.js
@@ -68,7 +68,12 @@ export function uiNoteComments() {
         mainEnter
             .append('div')
             .attr('class', 'comment-text')
-            .html(function(d) { return d.html; });
+            .html(function(d) { return d.html; })
+            .selectAll('a')
+                .filter(isExternalLink)
+                .attr('rel', 'noopener')
+                .attr('rel', 'nofollow')
+                .attr('target', '_blank');
 
         comments
             .call(replaceAvatars);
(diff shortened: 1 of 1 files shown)


modules/ui/note_comments.js
```diff
@@ -68,7 +68,12 @@ export function uiNoteComments() {
         mainEnter
             .append('div')
             .attr('class', 'comment-text')
-            .html(function(d) { return d.html; });
+            .html(function(d) { return d.html; })
+            .selectAll('a')
+                .filter(isExternalLink)
+                .attr('rel', 'noopener')
+                .attr('rel', 'nofollow')
```
quincylvania: Setting attributes via `attr` in D3 will override the existing value, so if you want both values you'll have to do `                .attr('rel', 'noopener nofollow')` instead.

quincylvania:
Hi and welcome, @JeeZeh. Thanks for looking into this! Apologies for keeping you waiting, there were a couple months where no one was working full-time on iD.

My main thought here is that even internal links should open in new tabs. iD is kind of a weird, self-contained app that doesn't smoothly transition to other browser pages within a session. Plus, people run versions of iD on all sorts of domains so we can't assume `openstreetmap.org` is necessarily internal. What do you think?

jesse-ashmore:
All good, glad it is being looked at now :) 

I'm all for covering the widest case here and just having all links open externally. It's also by far the less hacky solution here. 

I'll update with this and the attr change you suggested soon.

quincylvania:
@JeeZeh Thanks a lot! Good thinking around internal/external links but in this case we can keep it simple. Welcome to iD development 👍

PR #7894, opened 2020-08-11
Fix export of `geoVecLengthSquare`

til-schneider:
This fixes a typo (missing `.`): file path is `./vector.js`, not `/vector.js`


quincylvania:
@til-schneider Thanks 👍

PR #7895, opened 2020-08-11
Fix geometries detection in preset_fields.js

til-schneider:
`renderDisclosureContent` in `preset_fields.js` tries to detect the geometry types of the selected entities, but the result is always `[]`.

Reason: The handler of `_entityIDs.reduce` always returns `true`. So the result of the `reduce` call is `true` as well. `Object.keys(true)` results in `[]`.

Fix: The handler of `_entityIDs.reduce` should return the map which collects the geometry types, so `geometries` will be be something like `["line", "vertex"]`.


quincylvania:
@til-schneider Nice catch.

Issue #7900, opened 2020-08-14
Adding a preset field "parking=*" (amenity=parking_entrance)


sun-geo:
Adding a preset / field of `parking=*` (for main tag `amenity=parking_entrance`), see wiki:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparking_entrance#Additional_tags

nafees87n:
@sun-geo @quincylvania I want to do this. What I understood is that I'll have to add the `parking.json` field to the fields key array in `data/presets/presets/amenity/parking_entrance.json` . 
I am new to open source. I need help to understand the work to be done for this issue.

quincylvania:
Done in #7909 by @brokemyspoke.

PR #7909, opened 2020-08-17
add parking preset for parking_entrance

brokemyspoke:
Hi, first time attempt to contribute on this codebase so i hope i addressed correctly... https://github.com/openstreetmap/iD/issues/7900

One note on approach, I wasnt sure if it was possible to reuse the existing parking.json and constrain options so opted to create a variant since only two options are valid for parking_entrance.


quincylvania:
@brokemyspoke Looks great! Thanks a lot and welcome to open source 😄

PR #7932, opened 2020-08-26
Improve performance of raw tag editor

til-schneider:
The detection whether `_entityIDs ` includes a relation is performed many times inside a loop. Instead this should be done only once (lazy).

Besides I changed the name of the variable to `includesRelation` because the old name (`isRelation`) could be misunderstood.

### DIFF
--- a/modules/services/osm_wikibase.js
+++ b/modules/services/osm_wikibase.js
@@ -124,7 +124,6 @@ export default {
     // {
     //   key: 'string',
     //   value: 'string',
-    //   rtype: 'string',
     //   langCode: 'string'
     // }
     //
@@ -133,7 +132,7 @@ export default {
         var that = this;
         var titles = [];
         var result = {};
-        var rtypeSitelink = params.rtype ? ('Relation:' + params.rtype).replace(/_/g, ' ').trim() : false;
+        var rtypeSitelink = (params.key === 'type' && params.value) ? ('Relation:' + params.value).replace(/_/g, ' ').trim() : false;
         var keySitelink = params.key ? this.toSitelink(params.key) : false;
         var tagSitelink = (params.key && params.value) ? this.toSitelink(params.key, params.value) : false;
         var localeSitelink;
@@ -241,10 +240,6 @@ export default {
     // {
     //   key: 'string',     // required
     //   value: 'string'    // optional
-    // }
-    //   -or-
-    // {
-    //   rtype: 'rtype'     // relation type  (e.g. 'multipolygon')
     // }
     //
     // Get an result object used to display tag documentation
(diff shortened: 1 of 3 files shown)


bhousel:
Can we just always use the relation doc lookup for the `type=` key?  We probably don't need to actually check whether the user has selected a relation at all.

It seems like using `type` for anything other than the relation type is deprecated:
https://wiki.openstreetmap.org/wiki/Key:type

So it's not like looking up `type=someval` will return anything meaningful if the user has selected some feature that is not a relation but somehow still has a `type` tag.

til-schneider:
Using `type` was already part of the old logic which I didn't change.

I can't really answer whether the `type` check is OK or whether it should work differently or whether it should be dropped altogether. But from a pure UI perspective, I think it's a little bit strange that the tag reference works in any way differently if one of my selected entities happens to be a relation.

til-schneider:
I investigated a little bit more:
- The `type` check was added 5 years ago (on 8 Oct 2015) and it used to be applied to a single entity. See [b980c40](https://github.com/openstreetmap/iD/commit/b980c40dab107efd3ac74563ae6c93cbc3ea69d1).
- Years later (on 1 Jan 2020) the whole tag editor was change to support multiple selected entities. See [eda51f6](https://github.com/openstreetmap/iD/commit/eda51f6835bc7807fb68fad1bfa854bdd11d078f#diff-7c094a185e433eb9ee800d82d975cf83R253-R256).

So maybe a different behavior of the UI for relations was a good idea in the version 5 years ago, but with the support of multi-selection doesn't make sense any more?

til-schneider:
According to [taginfo](https://taginfo.openstreetmap.org/keys/type) there are still 102.108 nodes (0.06%) and 92.926 ways (0.01%) having the `type` tag.

til-schneider:
The `type` check basically decides whether `uiTagReference` is called with `{ rtype: d.value }` or with `{ key: d.key, value: d.value }`. So I checked the code to see what difference this does.

 In the end `rtype` is used by `getEntity` of `modules/services/osm_wikibase.js` to decide whether to use:
- A link describing the relation type like https://wiki.openstreetmap.org/wiki/Relation:multipolygon
- Or a link describing the tag's value like https://wiki.openstreetmap.org/wiki/Tag:highway=motorway
- Or - if the value link doesn't exist - a link describing the tag itself like https://wiki.openstreetmap.org/wiki/Key:highway

So I think this whole logic should be moved to `osm_wikibase.js`:
- `uiTagReference` should always be called with `{ key: d.key, value: d.value }`.
- `osm_wikibase.js` (and `uiTagReference` which acts as proxy) should remove the option `rtype`.
- `osm_wikibase.js` should use the following logic:
  - If `key` is `'type'`: Check whether the `Relation:${value}` page exists
  - If not: Check whether the `Tag:${key}=${value}` page exists
  - If not: Use the `Key:${key}` page

Should I change this PR accordingly?

bhousel:
> Should I change this PR accordingly?

Your suggestion looks good to me 👍

til-schneider:
Done.

til-schneider:
Removed the last doc change since all parameters are optional.

quincylvania:
@til-schneider Thank you! Definitely nice to simplify this. I was able to make a few more tweaks based on your changes.

PR #7980, opened 2020-09-11
[Performance] Calculate height of the textarea instead of letting the browser do it

bjornstar:
I like using the text editor for tags, but I noticed it was a bit slow to render.

Using the performance trace tool in Chrome, you can see a forced reflow when the height is unset and the browser reflows the page to calculate the new height:
openstreetmap-before

If we calculate the height ourselves, we can save a nice chunk of time:
openstreetmap-after

There are two modes for the text area, this is the result of letting the browser calculate the height:
textarea-before-edit
textarea-before-readonly

And this is when we supply the height ourselves:
textarea-after-edit
textarea-after-readonly

In the worst case I saw a savings of 60ms in the onClick handler on my machine, which is a noticeable improvement.

### DIFF
--- a/modules/ui/sections/raw_tag_editor.js
+++ b/modules/ui/sections/raw_tag_editor.js
@@ -285,8 +285,11 @@ export function uiSectionRawTagEditor(id, context) {
         if (_tagView !== 'text') return;
 
         var selection = d3_select(this);
-        selection.style('height', null);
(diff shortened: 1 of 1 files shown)


quincylvania:
@bjornstar Wow, that's a handy improvement. Thanks a lot! Especially for the trace and previews. Welcome to the project 👍

PR #7986, opened 2020-09-13
Added grocery store search terms

TheAdventurer64:


data/presets/presets/shop/supermarket.json
```diff
@@ -12,7 +12,11 @@
     "terms": [
         "grocery",
         "store",
-        "shop"
+        "shop",
+        "grocery store"
```
quincylvania, line 16: The build failed because you're missing commas after these values, but I'll just add them for you.

quincylvania:
@TheAdventurer64 Thanks, welcome to open source development!

One tip, you can run `npm run build` to check for errors and generate the derived data for your changes.

data/presets/presets/shop/supermarket.json
TheAdventurer64: Thanks!

TheAdventurer64:
How exactly do you run commands?

quincylvania:
> How exactly do you run commands?

From a command line tool like Terminal on macOS or cmd.exe on Windows. If you haven't used one before, there are lots of [tutorials](https://www.webfx.com/blog/web-design/command-line/). The command line is an important part of developing most software, including NodeJS apps like iD.

PR #7991, opened 2020-09-14
Support language-specific pluralization

### LINKED ISSUE #597: Better i18n plural support

quincylvania:
Closes #597.

This is essentially a reimplementation of #4964 (shout-out to @1ec5!), but for the current codebase.

We can also use the built-in [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules) function instead of an external dependency, since it's now supported in all major browsers. iD will attempt to fallback gracefully on legacy browsers.

### DIFF
--- a/modules/core/localizer.js
+++ b/modules/core/localizer.js
@@ -191,25 +191,44 @@ export function coreLocalizer() {
             });
     };
 
+    localizer.pluralRule = function(number) {
+      return pluralRule(number, _localeCode);
+    };
+
+    // Returns the plural rule for the given `number` with the given `localeCode`.
+    // One of: `zero`, `one`, `two`, `few`, `many`, `other`
+    function pluralRule(number, localeCode) {
+
+      // modern browsers have this functionality built-in
+      const rules = 'Intl' in window && Intl.PluralRules && new Intl.PluralRules(localeCode);
+      if (rules) {
+        return rules.select(number);
+      }
+
+      // fallback to basic one/other, as in English
+      if (number === 1) return 'one';
+      return 'other';
+    }
+
     /**
     * Given a string identifier, try to find that string in the current
     * language, and return it.  This function will be called recursively
     * with locale `en` if a string can not be found in the requested language.
     *
-    * @param  {string}   s             string identifier
+    * @param  {string}   stringId      string identifier
     * @param  {object?}  replacements  token replacements and default string
     * @param  {string?}  locale        locale to use (defaults to currentLocale)
     * @return {string?}  localized string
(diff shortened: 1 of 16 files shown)


modules/core/localizer.js
```diff
@@ -220,24 +239,50 @@ export function coreLocalizer() {
...
+                    // string and hope it makes sense.
+                    result = Object.values(result)[0];
+                  }
+                }
             }
+            if (typeof result === 'string') {
+              for (let k in replacements) {
+                const token = `{${k}}`;
+                const regex = new RegExp(token, 'g');
+                result = result.replace(regex, replacements[k]);
```
1ec5, line 264: This might be a good place to apply some [number formatting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) both for grouping characters (really big changesets!) and non-European numerals (for languages like Arabic and Hindi).

1ec5 (approved):
Looks like iD uses Ruby-style YAML, so Transifex should recognize the `one` and `other` subkeys as [plural rules](https://docs.transifex.com/formats/yaml#section-plurals). :+1:

quincylvania:
> Looks like iD uses Ruby-style YAML, so Transifex should recognize the one and other subkeys as plural rules. 👍

@1ec5 Yep, I made the changes with that documentation in mind. I'll debug if Transifex doesn't pick it up for some reason.

modules/core/localizer.js
quincylvania: Agreed, this is why I left strings like `{n} result for {search}` instead of `1 result for {search}`. I think I'll add formatting separately though.

quincylvania:
Looks like Transifex is reacting as expected! 

Screen Shot 2020-09-15 at 11 08 12 AM

quincylvania:
@1ec5 Do you know if we need to enable plural forms for strings like this that are based on an amount but don't include the amount?

```
      too_bendy:
        single: This can't be straightened because it bends too much.
        multiple: These can't be straightened because they bend too much.
```

maro-21:
> Do you know if we need to enable plural forms for strings like this that are based on an amount but don't include the amount?
> 
> ```
>       too_bendy:
>         single: This can't be straightened because it bends too much.
>         multiple: These can't be straightened because they bend too much.
> ```

It's not necessary when there is no numeral.

PR #7995, opened 2020-09-15
Fix GPS coordinate searches returning OSM elements

kymckay:
Simply let a location search match take priority over an ID pattern match (#7627)


quincylvania:
@SilentSpike Looks great, thanks!

Issue #7996, opened 2020-09-15
Language fallback based on the user's preferences


quincylvania:
If a string is untranslated in the user's preferred locale, iD currently falls back only to English. However, modern browsers let users specify [multiple ordered language preferences](https://support.mozilla.org/en-US/kb/use-firefox-another-language). iD should attempt to fallback to one or more of these before English.

Re: #7713

quincylvania:
I did this! iD will now attempt to use all the user's specified languages before defaulting to English. In the below example, the set locale is Greek, but untranslated terms fallback to the user's next language, German. Since I did #7994, and German currently has 100% coverage, the English locale is never loaded at all 💪

In addition, iD will now fallback to generic locales from regional locales when possible, e.g. `pt-BR` to `pt`.

<img width="1102" alt="Screen Shot 2020-09-15 at 5 47 23 PM" src="https://user-images.githubusercontent.com/2046746/93269334-c61cc100-f77c-11ea-96a6-591c75b23363.png">

PR #7998, opened 2020-09-16
Add `lang` attribute to translated text

### LINKED ISSUE #7963: Support `lang` HTML attribute for multilingual text
The [`lang`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) attribute is used on HTML elements to specify the language, script, and locale of the contained text. Encoding this info explicitly helps web users in a number of ways, including:

- Pronunciation in text-to-speech (critical for users of screen readers especially)
- Formatting, such as how to break words when wrapping text between lines
- Auto-transliteration, where a browser or extension translates foreign text into the user's language

Since iD isn't 100% translated into all languages, many mappers experience a mix of English and non-English strings. Adding `lang` attributes will improve this experience.

Re: #7713

quincylvania:
Closes #7963.

Turns out that adding a `lang` attribute to every text snippet in iD is a sweeping proposition 😬

This type of change isn't ideal, but I think it's necessary in making iD a first-class multilingual app. We can't really have [mixed text](https://github.com/openstreetmap/iD/issues/7996#issuecomment-693003909) throughout the UI without specifying what text is in what language. And some users of iD will always see mixed text unless we had a professional translation team that reaches 100% coverage in every language for every release (highly unlikely).


quincylvania:
I ended up making this backwards-compatible, so developers won't see different behavior from the `t` function. But now when we're displaying a translated string in the UI we should use the `t.html` function, which returns the text wrapped in a `span` element with the proper `lang` attribute. There's also a new `coreLocalizer.tInfo` function which returns both the text and locale, which is useful when you want to handle things manually.

Haven't yet achieved 100% coverage of the UI, but I'm happy with the groundwork so I went ahead and merged.

PR #8002, opened 2020-09-16
Deprecate bench:seats for bench:capacity

### LINKED ISSUE #7886: Deprecate bench+capacity in favor of bench+seats
There are quite a few `amenity=bench + capacity=<int>` out there.
- About 6k https://taginfo.openstreetmap.org/tags/amenity=bench#combinations
- Query https://overpass-turbo.eu/s/WR3

But the wiki says, that it should be `amenity=bench + seats=<int>` instead.
- https://wiki.openstreetmap.org/wiki/Key:seats
  > Do not use capacity=* in this context
- https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbench
  > seats=* Number of seats
- And it is used a lot more: 200k+ https://taginfo.openstreetmap.org/tags/amenity=bench#combinations

There was a proposal about this, which was rejected, but the seats key has won "de facto" regardless based on Wiki + Numbers.
- https://wiki.openstreetmap.org/wiki/Proposed_features/Attributes#seats.3D.2A

**Suggested change:**

Deprecate the capacity value on amenity=bench to be updated via the validator to seats.

In case both exist, remove the capacity and keep the seats.

mikenath223:
Fixes #7886

### DIFF
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -228,6 +228,10 @@
     "replace": {"barrier": "fence", "fence_type": "wood"}
   },
   {
+    "old": {"bench": "capacity"},
+    "replace": {"bench": "seats"}
+  },
+  {
     "old": {"bicycle:oneway": "*"},
     "replace": {"oneway:bicycle": "$1"}
   },
--- a/data/deprecated.json
+++ b/data/deprecated.json
@@ -228,8 +228,8 @@
     "replace": {"barrier": "fence", "fence_type": "wood"}
   },
   {
-    "old": {"bench": "capacity"},
-    "replace": {"bench": "seats"}
+    "old": {"amenity": "bench", "capacity": "*"},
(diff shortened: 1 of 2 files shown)


data/deprecated.json
```diff
@@ -227,6 +227,10 @@
     "old": {"barrier": "wood_fence"},
     "replace": {"barrier": "fence", "fence_type": "wood"}
   },
+  {
+    "old": {"bench": "capacity"},
```
quincylvania: The old tags for this would be `amenity=bench` + `capacity=*`. The replacement tags would then be `amenity=bench` + `seats=$1`.

mikenath223:
Thanks. Changes made.

quincylvania:
@mikenath223 Great, thank you!

mikenath223:
The pleasure is mine. Thanks 😊

PR #8012, opened 2020-09-21
Enable scaling the selection via hotkeys

### LINKED ISSUE #3372: Ability to resize a selected area
When an area is selected, add the ability to resize the object.

This would be especially useful after pasting an area with the same shape but not the same size.

quincylvania:
Closes #3372.

This implements my proposal in https://github.com/openstreetmap/iD/issues/3372#issuecomment-696161712 of scaling the selection by pressing the <kbd>–</kbd><kbd>+</kbd> keys while holding <kbd>Shift</kbd>. To support high detail situations, the scale factor is relatively small, but you can hold <kbd>Alt</kbd>/<kbd>Option</kbd> to scale by a wider factor.

The operation is disabled and gives error feedback in the following situations:

- The selection is already very small and you're trying to scale down.
- The selection isn't mostly visible.
- The selection contains relations that aren't fully downloaded.
- The selection includes nodes that are on undownloaded tiles.
- The selection includes nodes connected to hidden features.

Further work would involve scaling without a keyboard, e.g. via a mouse or multitouch. Though there's no rush since scaling isn't a highly common workflow and mappers can always scale manually by moving individual nodes.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -400,6 +400,28 @@ en:
         line: Reversed a line.
         lines: Reversed multiple lines.
         features: Reversed multiple features.
+    scale:
+      annotation:
+        down:
+          feature:
+            one: Scaled down a feature.
+            other: "Scaled down {n} features."
+        up:
+          feature:
+            one: Scaled up a feature.
+            other: "Scaled up {n} features."
+      too_small:
+        single: This feature can't be scaled down because it would become too small.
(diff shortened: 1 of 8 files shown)


quincylvania:
Here's a demo of a copy -> paste -> scale workflow. Circles are fairly easy to draw in iD, but you can see how this will be especially useful with irregular shapes.

scale_keys

PR #8013, opened 2020-09-22
Added clubhouse preset

TheAdventurer64:


quincylvania:
Thanks, @TheAdventurer64. I changed the name to Golf Clubhouse since there are other types of clubhouses. I also added some fields and terms.

TheAdventurer64:
👍 Thank you for getting back to me on this. I appreciate the preset field improvements, as well as the fixes for the search terms. But I have to ask: Why did you edit the wig extensions?

quincylvania:
> Why did you edit the wig extensions?

@TheAdventurer64 What do you mean? I don't think this relates to the clubhouse preset…

TheAdventurer64:
Well, you edited the wig extension preset in 3e3f56f.

quincylvania:
@TheAdventurer64 That's a link to the merge commit for this PR, so I'm afraid you're mistaken. In #8054 we added a search term to the Wig Shop preset, maybe that's what you mean?

TheAdventurer64:
Yes.

quincylvania:
@TheAdventurer64 We often add more search terms for presets so they'll appear when someone types in related words. In this case, extention is a misspelling of extension, but instead of changing it we just added both.

PR #8031, opened 2020-09-28
Added boat storage preset

### LINKED ISSUE #8000: Adding amenity=boat_storage to presets
Seems like a tag worth supporting: https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dboat_storage

animesh-007:
Fixes #8000 
Added boat storage preset.


quincylvania:
@animesh-007 Thanks!

PR #8039, opened 2020-09-30
Added Lounger preset

### LINKED ISSUE #7947: Add preset amenity=lounger
I think it would be useful to have https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dlounger as a preset, even if usage is not so high with 440+ cases ATM.

For search terms, I suggest to also add "bench", so people who are looking for a sub-tagging of bench will find this Preset and can choose between a regular bench or a lounger.

Additional fields on the preset
- I think, all of those that `amenity=bench` has
- I would skip the seasonal field for now and wait if people are missing it; I saw controversial discussions about it

Btw, one could also tag those loungers with `amenity=bench + bench:type=lounger` (see https://wiki.openstreetmap.org/wiki/Key:bench:type), but the main tag makes more sense and it also has more usage.

Personally, I would vote for deprecating the bench:type=lounger (and similar values, see https://taginfo.openstreetmap.org/keys/?key=bench%3Atype#values).

animesh-007:
Fixes #7947 
Added amenity lounger preset.


quincylvania:
@animesh-007 Thank you!

data/presets/presets/lounger.json
```diff
@@ -0,0 +1,32 @@
+{
+    "icon": "temaki-bench",
+    "fields": [
+        "backrest",
```
tordans, line 4: @quincylvania @animesh-007 this specific field sounds like something that should not be included, right? I would not know what a lounger without backrest looks like.

quincylvania: I changed some of the fields, see https://github.com/openstreetmap/iD/commit/1c3928160f68734aedadf977b723c833e8b19791.

PR #8046, opened 2020-10-02
improved readability for update_locales.js & update_imagery.js (template literals)

Croustys:


quincylvania:
Hi and welcome, @Croustys. I'd actually prefer to leave random syntax stuff like this alone unless we have a reason to change it, like a specific performance improvement or a broader standardization initiative. Otherwise these changes could cause merge conflicts and muddy the git blame.

Check out our [contributing guide](https://github.com/openstreetmap/iD/blob/develop/CONTRIBUTING.md) for info about helping the project.

PR #8056, opened 2020-10-06
Update to D3 v6

quincylvania:
Re: #8022 

The biggest change for iD from v6 is that d3.event is no longer global, and is instead passed directly to event listeners. While this pattern is more standard and feels modern, we have to update hundreds of calls. There are also times where iD used (or abused) the global nature of d3.event for some confusing mixed use case functions, so I'm trying to weed those out as well.


quincylvania:
Looks like we've hit a [snag](https://travis-ci.org/github/openstreetmap/iD/jobs/733713913) with the legacy build:

```
[!] (plugin buble) CompileError: Transforming generators is not implemented. Use `transforms: { generator: false }` to skip transformation and disable this error. (1:0)
2173node_modules/d3-array/src/merge.js (1:0)
```

Buble can't transform [generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*), an ES6 feature that D3 6 [adopts](https://github.com/d3/d3-selection/blob/master/src/selection/iterator.js). Assuming D3 doesn't revert this, our options are either to drop support for Internet Explorer and other legacy browsers (not really doable yet), find a different way to transpile these, or hold off on updating D3. Replacing buble might be the easiest, but there's also no rush to update.

quincylvania:
Turns out D3 v6 also uses advanced JavaScript features that our [browser-polyfills](https://www.npmjs.com/package/browser-polyfills) package didn't handle. I switched iD over to [core-js](https://www.npmjs.com/package/core-js) instead, which is integrated nicely with babel. The result is:

- Much broader polyfill coverage
- No need to bundle any polyfills with non-legacy build
- Only needed polyfills are bundled with the legacy build
- Smaller overall bundle sizes!

Issue #7129, opened 2019-12-16
Indicate units of measurement for elevation, height


cbeddow:
Related issues: https://github.com/openstreetmap/iD/issues/929, https://github.com/openstreetmap/iD/issues/5658

I know that when I add a `peak` for example I should add it in meters. I am in the US looking at USGS maps of Montana and dropping peak points, so first I convert on my own from feet to meters. It'd be good to have a reminder in the OSM iD interface for `elevation (meters)` so users don't just enter feet. But I think as some mention, having like a dropdown option for the units would be cool so when the user submits the changeset the units are taken into account, and the meters altitude is still submitted to OSM even if they want to use feet for input.

1ec5:
Also related: #3614.

quincylvania:
Closing this in favor of #3614. We should solve this for the general case so all fields that take length measurement values can have unit dropdowns.

PR #8075, opened 2020-10-12
Add road height field with unit dropdowns

### LINKED ISSUE #3614: Max Height field should have separate unit dropdown
The new “Max Height” (`maxheight=*`) field should have a dropdown for the unit, similar to the dropdown in the “Speed Limit” field. Acceptable units are listed at [this wiki page](https://wiki.openstreetmap.org/wiki/Key:maxheight) (and technically [this page](https://wiki.openstreetmap.org/wiki/Map_Features/Units) too, but measuring a vehicle’s height in miles would be ridiculous).

1ec5:
Implemented the `roadheight` field type based on the `roadspeed` field type. The new field has an explicit option to choose the unit of measurement (between meters and feet). When the field is set to feet, an additional input box lights up to allow the user to enter inches separately.

Initial state in Vietnam:
vn

`maxheight=4.8`:
m

Initial state in the U.S.:
us

`maxheight=12'`:
ft

`maxheight=12'6"`:
ft in

`maxheight=36"`:
in

`maxheight=below_default`:
below_default

Multiple selection:
multiple

Fixes #3614. Depends on ideditor/country-coder#30.


data/presets.yaml
```diff
@@ -1436,8 +1436,6 @@ en:
       maxheight:
         # maxheight=*
         label: Max Height
-        # maxheight field placeholder
-        placeholder: '4, 4.5, 5, 14''0", 14''6", 15''0"'
```
1ec5: This placeholder could probably be restored, but rewritten so that it only encourages the use of `.` as a decimal separator but doesn’t encourage putting `'` and `"` directly in the field.

data/core.yaml
```diff
@@ -724,6 +724,13 @@ en:
     add_fields: "Add field:"
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
+    maxheight:
+      # symbol for meters
+      meter: m
+      # abbreviation of feet
+      foot: ft
```
1ec5, line 771: These abbreviations are already present in the YAML file, but only as part of format strings:

https://github.com/openstreetmap/iD/blob/bb56584ce597028a96e28f87410d5f2741d7b4ad/data/core.yaml#L2278

In theory, we could reuse those strings to decide whether to display the unit dropdown to the left or right of the input box, but I don’t know if any language needs that extra complexity.

1ec5:
Regarding https://github.com/openstreetmap/iD/issues/7129#issuecomment-672971961, I think we could generalize this field to cover the other length fields (Height, Length, Width, and Elevation). However, if we ever add fields for `attraction:minheight` and `attraction:maxheight`, they’ll need a separate custom field because some countries measure human heights with different units than vehicle or building heights.

data/presets/schema/field.json
```diff
@@ -59,6 +59,7 @@
                 "identifier",
                 "localized",
                 "manyCombo",
+                "maxheight",
```
1ec5: This PR will need to be reworked now that #8175 has split out the field schema into a separate [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/) repository.

quincylvania: Yep unfortunately that's the case. I can work on reconciling them.

1ec5:
The new maxheight field type would be useful for openstreetmap/id-tagging-schema#232 as well.

data/presets/schema/field.json
1ec5: I finally got around to rebasing this PR. It’s a bit slimmer now, thanks to openstreetmap/id-tagging-schema@e0a72439df71dffbe2261ecdb1f6f2fdec17fdea.

Issue #3659, opened 2016-12-18
Transitioned actions


bhousel:
Some actions, like the new Reflect action (#3555), could be performed in a transitioned way.. It would be cool (and actually beneficial to users not so adept at spatial reasoning) to see the feature quickly flip across its axis, rather than instantly appear in its new configuration.   Circularize, Orthogonalize, and Straighten could do this too.

It could probably be implemented by allowing certain actions to have a property indicating whether or not the action is transitionable, and if so, use `d3` or `timer`/`requestAnimationFrame` in `history.js` to difference the graph and rewrite the `perform` action as series of tweened `replace` calls followed by a final `perform` call.

bhousel:
some discussion in #1734

bhousel:
This turned out all right!

![transitioned reflects2](https://cloud.githubusercontent.com/assets/38784/21447915/f888575e-c8a9-11e6-86b8-f676031c1e48.gif)

bhousel:
circles too 👯 

![transitioned circularize](https://cloud.githubusercontent.com/assets/38784/21456865/2886b6f4-c8f9-11e6-8dc5-45b76e20cbc6.gif)

bhousel:
The transitions look like this slowed down to 500ms.  I think 150ms-ish feels better for actual editing, but this is fun too.

![slow transitions](https://cloud.githubusercontent.com/assets/38784/21457006/0a335616-c8fa-11e6-8069-a920720be0c2.gif)

bhousel:
.... 
 
![straighten transition](https://cloud.githubusercontent.com/assets/38784/21457763/acd27bb8-c8ff-11e6-9c81-41e75cf8f456.gif)

bhousel:
📐 😏  

![orthogonalize transition](https://cloud.githubusercontent.com/assets/38784/21458777/488bfabe-c907-11e6-8ec0-3a7c2ad2b9a3.gif)

slhh:
The visual impression should also be tested with real usecases like circularizing a nearly cirlular way. Maybe the animation time can be dependent on the amount of change (e.g. moving ditance of the nodes). For example the typical slight modifications could take 150ms, which is indeed better for editing, and such misuses like circularizing a bulding as shown above could be in the 500ms range, which makes accidential use more visible.

bhousel:
[...]

PR #8083, opened 2020-10-14
Tidy some things up.

teymour-aldridge:
Please let me know if (small) changes of such nature are not appreciated.

I obviously don't have the same appreciation for the software as the maintainers do, so I may be incorrect in my interpretations of the code.

I also added `.idea` (where IntelliJ editors store metadata) folder to the `.gitignore`

### DIFF
--- a/modules/actions/circularize.js
+++ b/modules/actions/circularize.js
@@ -2,14 +2,14 @@ import {median as d3_median} from 'd3-array';
 
 import {
     polygonArea as d3_polygonArea,
-    polygonCentroid as d3_polygonCentroid,
-    polygonHull as d3_polygonHull
+    polygonHull as d3_polygonHull,
+    polygonCentroid as d3_polygonCentroid
 } from 'd3-polygon';
 
-import {geoVecInterp, geoVecLength} from '../geo';
-import {osmNode} from '../osm/node';
-import {utilArrayUniq} from '../util';
-import {geoVecLengthSquare} from '../geo/vector';
+import { geoVecInterp, geoVecLength } from '../geo';
+import { osmNode } from '../osm/node';
+import { utilArrayUniq } from '../util';
+import { geoVecLengthSquare } from '../geo/vector';
 
 
 export function actionCircularize(wayId, projection, maxAngle) {
--- a/modules/actions/circularize.js
+++ b/modules/actions/circularize.js
@@ -133,7 +133,6 @@ export function actionCircularize(wayId, projection, maxAngle) {
                     var nearAngle = nearNodes[nodeId];
                     var dist = Math.abs(nearAngle - angle);
                     if (dist < min) {
-                        dist = min;
                         origNode = origNodes[nodeId];
                     }
                 }
@@ -252,7 +251,7 @@ export function actionCircularize(wayId, projection, maxAngle) {
                 return false;
             }
         }
-        
+
         //check if central angles are smaller than maxAngle
         for (i = 0; i<hull.length; i++){
             actualPoint = hull[i];
@@ -266,7 +265,7 @@ export function actionCircularize(wayId, projection, maxAngle) {
             if (angle > Math.PI){
                 angle = (2*Math.PI - angle);
             }
- 
+
             if (angle > maxAngle + epsilonAngle) {
                 return false;
             }
--- a/modules/actions/circularize.js
+++ b/modules/actions/circularize.js
@@ -1,4 +1,4 @@
-import {median as d3_median} from 'd3-array';
+import { median as d3_median } from 'd3-array';
 
 import {
     polygonArea as d3_polygonArea,
--- a/modules/actions/circularize.js
+++ b/modules/actions/circularize.js
@@ -128,10 +128,12 @@ export function actionCircularize(wayId, projection, maxAngle) {
                 ]);
 
                 // choose a nearnode to use as the original
+                var min = Infinity;
                 for (var nodeId in nearNodes) {
                     var nearAngle = nearNodes[nodeId];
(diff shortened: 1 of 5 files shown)


modules/actions/circularize.js
```diff
@@ -128,12 +128,10 @@ export function actionCircularize(wayId, projection, maxAngle) {
                 ]);
 
                 // choose a nearnode to use as the original
-                var min = Infinity;
                 for (var nodeId in nearNodes) {
                     var nearAngle = nearNodes[nodeId];
                     var dist = Math.abs(nearAngle - angle);
-                    if (dist < min) {
-                        dist = min;
```
bhousel, line 136: haha this should probably be `min = dist;`.  I always wondered why this code didn't work.

teymour-aldridge: Ah :) Will alter

teymour-aldridge: Surely the assignment isn't necessary atol, given that `min` is never used?

bhousel: `min` is used each time through the loop.  This loop is trying to find the nearest existing node for a newly added node to appear next to.  This was intended to be a cool effect that occurs [during the few milliseconds of transition when a feature is circularized](https://github.com/openstreetmap/iD/issues/3659#issuecomment-269004422).  Because I had the assignment backwards, instead all the newnodes pop out from the "first" node of the circle and then just work their way around the circle to where they should be.  I'm probably the only person who ever noticed this.

teymour-aldridge: Oh sorry, I thought the `min` was declared inside the loop :)
Yeah, that makes sense

modules/actions/circularize.js
```diff
@@ -1,15 +1,15 @@
...
 } from 'd3-polygon';
 
-import { geoVecInterp, geoVecLength } from '../geo';
-import { osmNode } from '../osm/node';
-import { utilArrayUniq } from '../util';
-import { geoVecLengthSquare } from '../geo/vector';
+import {geoVecInterp, geoVecLength} from '../geo';
+import {osmNode} from '../osm/node';
+import {utilArrayUniq} from '../util';
+import {geoVecLengthSquare} from '../geo/vector';
```
quincylvania: As you can see in most of iD's source files, the existing spacing is preferred, so could you change these back?

modules/ui/edit_menu.js
```diff
@@ -126,7 +126,7 @@ export function uiEditMenu(context) {
         }
 
         // update
-        buttons = buttonsEnter
+        buttonsEnter
```
quincylvania, line 129: We actually have to do the assignment when we merge, or else the results are discarded. See https://github.com/d3/d3-selection/blob/v2.0.0/README.md#selection_merge. Could you change this back?

modules/ui/shortcuts.js
```diff
@@ -74,8 +74,7 @@ export function uiShortcuts(context) {
             .append('span')
             .html(function (d) { return t.html(d.text); });
 
-        tabs = tabs
-            .merge(tabsEnter);
+        tabs.merge(tabsEnter);
```
quincylvania: Same here as in edit_menu.js.

modules/ui/shortcuts.js
```diff
@@ -235,8 +234,7 @@ export function uiShortcuts(context) {
             .html(function (d) { return d.text ? t.html(d.text) : '\u00a0'; });
 
 
-        shortcuts = shortcuts
-            .merge(shortcutsEnter);
+        shortcuts.merge(shortcutsEnter);
```
quincylvania: Same here as in edit_menu.js.

modules/osm/multipolygon.js
```diff
@@ -210,7 +210,6 @@ export function osmJoinWays(toJoin, graph) {
             }
 
             if (!nodes) {     // couldn't find a joinable way/member
-                doneSequence = true;
```
quincylvania, line 213: Does this change the functionality? If not, can we remove `doneSequence` entirely?

quincylvania (changes requested):
Welcome, @teymour-aldridge. Thanks for opening this, could you address the few things I mentioned? Let me know if you have questions.

modules/actions/circularize.js
teymour-aldridge: Ah – sorry, my editor gets a bit aggressive at times.

modules/ui/edit_menu.js
teymour-aldridge: Sure

teymour-aldridge: I might be wrong, but I think that the assignment isn't used (it's not captured in a closure, or used later in the program).

quincylvania: Oh okay, looks like you're right.

modules/ui/shortcuts.js
quincylvania: If `tabs` isn't used after this then we can skip the merge entirely.

modules/osm/multipolygon.js
teymour-aldridge: I think it's needed because of the `break` below to exit from the loop.

quincylvania (approved):

quincylvania:
@teymour-aldridge Thanks for working through the changes!

teymour-aldridge:
You’re welcome – hopefully they make things better!

My school are moving our volunteering to editing bits of OpenStreetMap to provide data for first responders in LEDCs so I thought it might be fun to write code for the editor instead of just editing data on the map :)

PR #8093, opened 2020-10-16
Add name field to park and ride preset

evansiroky:
I've been mapping a lot of park and rides lately. I often like to give them a name, but the name field is hidden by default. This PR makes the name field be shown by default when selecting the park and ride preset. Full disclosure: I am mapping a lot of these park and rides so that they can be used by OpenTripPlanner deployments that the company I work for supports. OpenTripPlanner uses the name of the park and ride lot in trip planning results.


quincylvania:
Hi and welcome, @evansiroky. This looks like a fine change, thanks! You can start using it right away on our [preview site](https://ideditor.netlify.app) if you want. Happy mapping.

PR #8095, opened 2020-10-16
Check to see if code is clean

henrychase-linux:
Check to see if code is clean


quincylvania:
Welcome, @henrychase-linux. I'm not sure what your intention is here, but merging an old version of iD into our working branch isn't all that desirable 😅

PR #8110, opened 2020-10-21
Update country-coder to v4.0.0

quincylvania:
Re: #8102 

I had this going into 2.19.0 but realized that iD would still be bundling version 3.2.0 as well, since some dependencies haven't been updated yet. This would have a large cost to bundle size, since twice the data would be included.

We can just leave this open until all the right dependencies are up-to-date.


bhousel:
I think we can get them all updated in the next few days..

manfredbrandl:
... in the next few weeks ...

bhousel:
> ... in the next few weeks ...

I know you're trying to be funny, but all the dependencies that use Country Coder actually were updated back at the end of October.

manfredbrandl:
@bhousel may I delete my last comment?

bhousel:
> @bhousel may I delete my last comment?

Sure

quincylvania:
For those wondering, this is still pending since it's related to a chain of updates around presets planned for the next major release, including:

- Moving presets to id-tagging-schema
- Supporting location-conflation sets in id-tagging-schema
- Separating name-suggestion-index presets from iD's core presets 

There's no rush to complete this part before the other parts are in line.

PR #8117, opened 2020-10-23
Replace nested ternary expressions

quincylvania:
Nested ternary expressions can be inaccessible to beginners and are unnecessarily difficult to read for developers of all levels. For an open source app like iD, I contend it's better to avoid them. This PR enables the [no-nested-ternary](https://eslint.org/docs/rules/no-nested-ternary) linting rule and replaces all existing usage with less complex syntax.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -53,6 +53,7 @@
         "no-loop-func": "error",
         "no-loss-of-precision": "error",
         "no-multi-str": "error",
+        "no-nested-ternary": "warn",
         "no-new": "error",
         "no-new-func": "error",
         "no-new-wrappers": "error",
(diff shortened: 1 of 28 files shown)


kymckay (approved):
A sensible change

quincylvania:
@bhousel Is there a wider reason you don't like this or just personal preference? I don't care strongly about this change, but I do think we should try to make the code as readable as possible for newcomers.

bhousel:
> @bhousel Is there a wider reason you don't like this or just personal preference?

Just personal preference - I find concise code much easier to skim over.

quincylvania:
Okay, maybe let's leave the rule disabled for now and just convert some of the most complex expressions.

https://github.com/openstreetmap/iD/blob/15c8e060f8444231076573db08383ca342bff0ce/modules/behavior/draw_way.js#L241-L242

PR #8125, opened 2020-10-27
Add support for CRS:84 projection

rbuffat:
WMS server typically support the CRS:84 projection, which is very similar to  EPSG:4326. The difference is, that CRS:84 is in lon/lat axis order, while EPSG:4326 has lat/lon axis order. This is relevant for WMS 1.3.0, as with this standard the bbox coordinates should be in order as defined in the CRS, while WMS versions prior 1.3.0 require lon/lat, respectively x,y axis ordering.  

In practice, it is unlikely that a server supports CRS:84 and not also EPSG:4326, but adding support for it is not a lot of work.

### DIFF
--- a/modules/renderer/background_source.js
+++ b/modules/renderer/background_source.js
@@ -142,6 +142,7 @@ export function rendererBackgroundSource(data) {
                 var lat = Math.atan(sinh(Math.PI * (1 - 2 * y / zoomSize)));
 
                 switch (source.projection) {
+                    case 'CRS:84':
                     case 'EPSG:4326':
                         return {
                             x: lon * 180 / Math.PI,
@@ -171,7 +172,10 @@ export function rendererBackgroundSource(data) {
                 case 'wkid':
                     return projection.replace(/^EPSG:/, '');
                 case 'bbox':
-                    // WMS 1.3 flips x/y for some coordinate systems including EPSG:4326 - #7557
+                    // WMS versions prior 1.3.0 require easting / northing (x,y) coordinate order for bbox parameter
+                    // WMS version 1.3.0 requires axis ordering as specified in the CRS - #7557
+                    // E.g. http://epsg.io/4326 > Coordinate system: ... Orientations: north, east.
+                    //      http://epsg.io/3857 > Coordinate system: ... Orientations: east, north.
                     if (projection === 'EPSG:4326' &&
                         // The CRS parameter implies version 1.3 (prior versions use SRS)
                         /VERSION=1.3|CRS={proj}/.test(source.template())) {
--- a/scripts/update_imagery.js
+++ b/scripts/update_imagery.js
@@ -75,7 +75,8 @@ const supportedWMSProjections = [
   'EPSG:102100',
   'EPSG:3785',
   // WGS 84 (Equirectangular)
-  'EPSG:4326'
+  'EPSG:4326',
+  'CRS:84'
 ];
 
 


rbuffat:
I forgot to add some background information:
See the [WMS 1.3.0 specification](http://portal.opengeospatial.org/files/?artifact_id=14416) 

The relevant parts are:

> 6.7.3.2 CRS namespace for CRS 
> EXAMPLE “CRS:84” refers to WGS 84 geographic longitude and latitude expressed in decimal degrees, with longitude
> ranging from −180 degrees to +180 degrees and latitude from −90 degrees to +90 degrees. 

> 6.7.4 Bounding boxes
> Bounding box values specify the portion of the Earth to be mapped through two pairs of coordinates in a specified
> Layer CRS. The first pair specifies the minimum coordinate values in the Layer CRS, the second specifies
> maximum coordinate values. Although for most CRSs with axes incrementing to the east and north this would be
> the lower left and upper right corners of the area of interest, the minimum and maximum values might be at other
> points in some instances. For example, when using geographic coordinates to describe an area over a pole, or
> when the layer CRS axes increment in directions other than east and north. The order in which ordinates in each
> pair are listed shall be as defined by the Layer CRS; x corresponds to the first axis in the Layer CRS and y to the
> second. This order may not coincide with the Map CS axis order i, j. The bounding box coordinate values shall be
> in the units defined for the Layer CRS.
>
> ...
>
> EXAMPLE 1 A \<BoundingBox\> metadata element for a Layer representing the entire Earth in the CRS:84 Layer CRS
> would be written as \<BoundingBox CRS="CRS:84" minx="-180" miny="-90" maxx="180" maxy="90"\>.
> A BBOX parameter requesting a map of the entire Earth would be written in this CRS as  BBOX=-180,-90,180,90.
> EXAMPLE 2 A \<BoundingBox\> representing the entire Earth in the EPSG:4326 Layer CRS would be written as
> \<BoundingBox CRS="EPSG:4326" minx="-90" miny="-180" maxx="90" maxy="180"\>.
> A BBOX parameter requesting a map of the entire Earth would be written in this CRS as
> BBOX=-90,-180,90,180.

matkoniecz:
Is there any practical use case or any known server where it would be useful?

Was it tested on any case serving that data?

> In practice, it is unlikely that a server supports CRS:84 and not also EPSG:4326, but adding support for it is not a lot of work.

well, adding code that needs to be maintained seems pointless then

1ec5:
In trying to find sources for editor-layer-index, I have occasionally encountered WMS layers that support CRS:84 but not EPSG:4326. Usually they support some other EPSG:* projection that iD does not. Unfortunately I’m unable to find these examples again, since I just moved on to alternatives.

matkoniecz:
If anyone would see anything where it can be tested it would be helpful to share it

tyrasd (approved):
It seems to be part of the specs, so I think it's fine to include this possibility to specify EPSG 4326 for the (unlikely?) case if a server will only offer it in this variant.

bryceco:
Might be relevant, it's a request for 3765 as an alias for 4326:
https://github.com/bryceco/GoMap/issues/832

danieldegroot2:
> If anyone would see anything where it can be tested it would be helpful to share it

@1ec5 @matkoniecz try searching GitHub.

Example `repo:osmlab/editor-layer-index "CRS:84" NOT "EPSG:4326"`

unless many more EPSG are identical..

-# also it is possible ELI files or WMS server content is inaccurate.

PR #8128, opened 2020-10-28
Walkthrough hangs on "line" module when splitting "Washington Street"

ricloy:
I think the problem is that operations.split.annotation now has two different values depending on a numerical argument which is missing in the current code.

I haven't tested my patch, I just copied the format from /modules/ui/intro/building.js .


quincylvania:
Woah, thanks a lot @ricloy for finding this and for the quick PR! 🙇‍♂️

PR #8141, opened 2020-10-30
Fix error when custom background url is empty

willemarcel:
When we select the custom layer on the background panel, it prevents the dialog to open, but if the URL is not defined, it raises an error on the console: `TypeError: Cannot read property 'preventDefault' of undefined`

I believe the best would be to open the dialog when the url is not set. This PR fixes it.


quincylvania:
@willemarcel Thanks for the bug fix! I changed it a bit so `d3_event` is only used by the event handler and then `editCustom` is more agnostic.

PR #8229, opened 2020-12-01
Unbundle tagging schema data

quincylvania:
This makes iD load the [presets and other tagging data](https://github.com/openstreetmap/id-tagging-schema) from the npm CDN at runtime, meaning we can update them without doing an iD release.

Since the presets require their own version-specific translations, those strings have been removed from iD's bundled locale files. `coreLocalizer` has been upgraded to handle multiple sets of locale files.

The name-suggestion-index presets have been temporarily removed for this, but we'll add them back as an external preset expansion pack.

One snag is that iD bundles only the fontawesome icons for the listed presets, so new FA icons can only be added with a new release. We should look into a better solution.


pnorman (changes requested):
This will need an update to the privacy policy to mention that it requests content and transmits PII to the NPM CDN.

kymckay:
@pnorman Could you clarify which PII is being transmit? I couldn't see that in the changes

quincylvania:
> This will need an update to the privacy policy to mention that it requests content and transmits PII to the NPM CDN.

@pnorman Hmm okay, we've already been fetching certain resources there since 2.18.0, but better late than never I suppose.

https://github.com/openstreetmap/iD/blob/de7ab7e5b81db2536b62139b950b13fb5365cdce/modules/core/file_fetcher.js#L22-L25

ireun:
It would be great if the openstreetmap/id-tagging-schema could be updated to include newest Translations. 

Also that repository seems to be kind of unmaintained, many pull requests are left for a half a year without any notice from the collaborators.

PR #8238, opened 2020-12-08
Repost: Support translatable options in all combo fields

### LINKED ISSUE #2708: Openstreetmap tag strings are not translatable
Before all, I am using iD in French if that matters.
- Hints

Currently when editing a route, one can see default hints untranslated:

> asphalt, unpaved, paved...

However for some others fields there are translated (name, mountain bike difficulty, some access fields). 
The point is: around 50% fields are translated, 50% aren't. Since French Transifex is 100%, I do not think this is a missing translation, can you confirm? 
- Keys

I don't know if it's possible to translate keys, but to take again the surface type field, when you select it you have to select a value in [asphalt, unpaved, paved, gravel, ...]. I know this is the English key and the wiki explains each of them, but would it be possible to either translate them in UI or display a translated equivalent within parenthesis for instance? If so, where would the keys be translated? Could it be done on the wiki so that other projects can use them as well?
Again, this is not specific to route surface type, many (all?) keys are proposed in a similar way and if the above behavior is implemented it should not be field specific.

quincylvania:
Closes #2708.

This is essentially a re-implementation of #7411, but on top of the unbundled preset system. I'll just repost what I wrote before: 

> This extends iD's existing mechanism for translating combo field options to support fields where all possible values may not be predetermined. The text styling indicates which values are raw tags and which are translated labels. It works for `semiCombo` and `multiCombo` as well as single-value combos. Multiselection is also supported.

This functionality opens the door to adding loads of new tagging data to the presets, which is the main reason I've held off on the change up until now. But the new, robust preset system should help alleviate headaches in a few ways:

- There's a focused repo for people to collaborate and debate changes
- We can now make tweaks without having to ship iD itself
- 100% translation completion on preset strings is now distinct from 100% completion on iD's other strings

Translation is a major factor here. I know people have been translating tags in the OSM Wikibase for awhile now, so we should see if there's a good way to avoid duplicate work.


matkoniecz:
>I  know people have been translating tags in the OSM Wikibase for awhile now

Interesting - I am not a fan of data items, but this is a case where they can be useful.

manfredbrandl:
Question: After this PR goes live will the translations for presets be used from the OSM-Wiki instead of Transifex?

quincylvania:
> Question: After this PR goes live will the translations for presets be used from the OSM-Wiki instead of Transifex?

@manfredbrandl No, but something like this might be future work.

PR #8241, opened 2020-12-09
Ignore another typo as it's in the changelog

peternewman:
We should perhaps add an ignore lines file

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -13,5 +13,5 @@ jobs:
         with:
           check_filenames: true
           skip: ./.git,./data/territory_languages.json,./data/imagery.json,./data/languages.json,./data/address_formats.json,./dist/locales,./docs/img,./dist/img
-          ignore_words_list: "auxilary,casette,cemetary,chancel,childs,extentions,files',froms,generat,guerilla,kindergarden,nd,specialties,vias,visibles,tos"
+          ignore_words_list: "auxilary,casette,cemetary,chancel,childs,extentions,files',froms,generat,guerilla,kindergarden,nd,specialties,vias,visibles,tos,sculpter"
           only_warn: 1


quincylvania:
@peternewman Thanks!

> We should perhaps add an ignore lines file

As in ignoring specific lines of specific files? Seems like that would be hard to maintain, especially for the changelog.

peternewman:
> @peternewman Thanks!
> 
> > We should perhaps add an ignore lines file
> 
> As in ignoring specific lines of specific files? Seems like that would be hard to maintain, especially for the changelog.

Yeah, although it currently means if someone adds the word sculpter in another file it won't flag it as it's ignored globally. This was we'd explicitly just skip it in the changelog and in terms.

Presumably it doesn't change in the changelog, someone adds it and so we'd add the line once.

quincylvania:
@peternewman I'm not sure I understand. If it's easy to demo or illustrate this then that might help 😄

> Presumably it doesn't change in the changelog, someone adds it and so we'd add the line once.

I mean if we say "ignore 'sculpter' on line 302 in CHANGELOG.md" and then add release notes for the next version so suddenly line 302 becomes line 427, then that's not very helpful.

peternewman:
> @peternewman I'm not sure I understand. If it's easy to demo or illustrate this then that might help 
> 
> > Presumably it doesn't change in the changelog, someone adds it and so we'd add the line once.
> 
> I mean if we say "ignore 'sculpter' on line 302 in CHANGELOG.md" and then add release notes for the next version so suddenly line 302 becomes line 427, then that's not very helpful.

It's more like ignore:
`* Rename sculpter to sculptor ([#4504], thanks [@simonpoole])`

So whenever ^line$ matches that pattern, it doesn't test it. Which technically means if thanks was misspelled as thaks and we didn't notice it would never complain.

peternewman:
Something like https://github.com/openstreetmap/iD/pull/8244 :smile:

PR #8253, opened 2020-12-15
stopped allowing undo/redo into a degenerate area

### LINKED ISSUE #7946: Disallow undoing to a degenerate area
After drawing an area and exiting the draw mode, you can undo to the point where the area only has two unique nodes. Two-node areas are degenerate and iD shouldn't create them. Attempting to undo to just one node properly removes the entire way for both lines and areas, so I think undoing areas to two nodes should also work like this.

Screen Shot 2020-09-02 at 1 24 37 PM

right1:
References https://github.com/openstreetmap/iD/issues/7946

Treats starting an area annotation the same as undefined on undo/redo

### DIFF
--- a/modules/core/history.js
+++ b/modules/core/history.js
@@ -8,6 +8,7 @@ import { coreGraph } from './graph';
 import { coreTree } from './tree';
 import { osmEntity } from '../osm/entity';
 import { uiLoading } from '../ui/loading';
+import { t } from './localizer';
 import {
     utilArrayDifference, utilArrayGroupBy, utilArrayUnion,
     utilObjectOmit, utilRebind, utilSessionMutex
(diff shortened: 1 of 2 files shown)


matkoniecz:
https://github.com/openstreetmap/iD/issues/7946 is reportedly fixed

matkoniecz:
It is still possible to select area drawing and undo to the point of having untagged way with two nodes and `area=yes`

tyrasd:
This workaround seems a bit too fragile (checking for the history's state annotation). Also, while one is still in the _add area_ behavior, it is on purpose that the undo/redo lets you go back to the degenerate initial state (if you decide to undo one of the first clicks), because you're still in the process of drawing the area. This PR would break that workflow.

Instead, it should probably be addressed in `draw_way.js` -> after the way is finished drawing, the degenerate state should be removed from the history.

PR #8254, opened 2020-12-15
Update osm-community-index to v3.0.0

bhousel:
The major version bump is because
- it no longer distributes `index.js`, `index.mjs` code bundles
   (which iD hasn't needed in a long time anyway)
- renamed `dist/features.json` -> `dist/featureCollection.json`


modules/ui/success.js
```diff
@@ -32,14 +32,20 @@ export function uiSuccess(context) {
...
 
         Object.values(ociResources).forEach(resource => {
-          const feature = loco.resolveLocationSet(resource.locationSet).feature;
-          let ociFeature = ociFeatures[feature.id];
-          if (!ociFeature) {
-            ociFeature = JSON.parse(JSON.stringify(feature));  // deep clone
-            ociFeature.properties.resourceIDs = new Set();
-            ociFeatures[feature.id] = ociFeature;
+          let feature;
+          try {
```
bhousel, line 36: Also this is wrapped in a `try` now because it can potentially fail to resolve a locationSet with a code from a newer version of country-coder.  (Some of them include things like `gb-nir`, `gb-sct` now)

quincylvania:
@bhousel Thanks a bundle!

PR #8258, opened 2020-12-18
Migrate Travis CI to GitHub Actions

bhousel:
Travis-ci.org is shutting down Dec 31, 2020 - Projects must either find a replacement or migrate to Travis-ci.com by then. Also, they are already severely limiting the amount of builds that projects can run.

This PR 
- Converts the Travis CI job into a GitHub Action.
- Adds another GitHub Action to push iD's source strings to Transifex 
(currently, iD runs a script to do this on Travis in an `after_success` hook). 

I haven't tested the Transifex job but it will probably work, or at worst, fail quietly.  

Transifex requires a `TX_TOKEN` environment variable to be assigned in the GitHub settings for iD.  If you don't have one already, you can generate a token in your [Transifex user settings page](https://www.transifex.com/user/settings/api/).


quincylvania:
@bhousel Thanks a bunch! It's nice to have this right on GitHub.

Issue #7619, opened 2020-05-19
Boundaries should be easily un-anchored from physical objects 


jjiglesiasg:
Boundaries are polytical NON existing objects that should be easily detached from actual physical objects like roads, rivers, etc.

The disconnection feature of the right click should be applicable for large sections of these "forced" juntions instead of node by node. Indeed roads/waterways and boundaries should not be possible to be connected at all. Due to different nature, and boundaries not being a real object.

Thanks

JJ

quincylvania:
@jjiglesiasg You can click-and-drag on the map while holding shift to lasso all the connection nodes and then use the Disconnect command to detach them all at once. Perhaps this is sufficient?

maro-21:
> roads and boundaries should not be possible to be connected at all

I don't agree. I had a situation when I had to connect a road to a boundary because of the names of the streets in two cities. If I didn't correct, overpass would find the name of a street which belongs to the other city.

jjiglesiasg:
Hi quincylvania. there is no possibility to disconnect multiple nodes because boundaries are part of relations and ID do not allow to disconnect following your example.

maro-21 I dont know under what exact circunstances your example falls, but again Boundaries are NOT existing objects and roads, waterways, etc, are REAL objects that needs to be adjusted, corrected, and redrawn, fighting with hundred of nodes connected with boundaries do Not make sense at all and for worst if they could not be easily detached.

Likewise power features, should not be possible to connect a road/waterway node to a power line node, they are objects of different nature...

BsRgds JJ

DujaOSM:
I propose that selecting a way (open or closed) and issuing "disconnect" (via context menu or D key) disconnects all its nodes from all adjacent ways, except for those that it shares a relation with. That would be a huge time saver.

 On many occasions, I've encountered good-faith but clueless mapping where everything is glued together and it's almost impossible to make any changes in geometry of objects without breaking something. Think a single node adjoining three administrative boundaries, four highways and two farmland polygons: try untangling that. In fact, now that iD has a powerful validation engine, I would like having a warning against gluing objects from different "layers" (e.g. boundaries, waterways, landuse and highways).

jjiglesiasg:
That's my bottom point, a mess all these connected areas or Political (Administrative) Boundaries to a way or river do not make sense at all for me at least...

kymckay:
Validator discussion here has some relevance to #6631

Original issue might be addressed by now implemented #7652

quincylvania:
[...]

Issue #7590, opened 2020-05-10
Multiselection on touch devices


quincylvania:
Without the ability to shift-click, it's impossible to select multiple features on touchscreens. This makes operations like merging ways unavailable, even with #7577. We should come up with some way to enable this functionality with no keyboard, perhaps via a mode in which tapping features adds or removes them from the selection.

PR #8264, opened 2020-12-22
Add shortcut to select parent way(s)

### LINKED ISSUE #1239: Handle coincident features (i.e. multiple things one spot)

### LINKED ISSUE #2225: Right now it is impossible to select between multiple overlapping polygons

### LINKED ISSUE #5009: Keep the way selected when navigating by home/end keys
When you have a way selected and press home/end keys, the first/last node of the way gets selected.
Is it possible to just pan the map to the first/last node but keep the way selected?
It is useful when you want to just check what is going on at the ends.
It would still be possible to get the selection to the end nodes by pressing the keys when a way node is selected.

### LINKED ISSUE #7375: Lasso tool - Enable entire features selection
Since now iD can edit multiple selected features #7276, the shift-click-drag (lasso tool) should select the entire feature once all its nodes are selected, instead of only selecting nodes as now.

The use case is when the user needs to select multiple areas to change the tagging of them. For example consider selecting multiple areas tagged as "house" to change them to "plot" preset. Now the user has to shift-click over each one of the feature instead of using the lasso tool.

1ec5:
Pressing <kbd>Ctrl</kbd><kbd>↑</kbd> on Windows or ⌘↑ on macOS selects the selected node’s highlighted parent way, or all the node’s parent ways if none is currently highlighted.

## Rationale

This keyboard shortcut is helpful for keyboard accessibility, reducing the need to use the mouse in some cases. For example, in this scenario, the crosswalk way’s hit target is crowded out by the crosswalk node’s icon and adjacent road and sidewalk ways:

crosswalk

Without this keyboard shortcut, you would have to switch to wireframe mode, zoom in, or meticulously move the mouse until the cursor changes to indicate a line selection. These workarounds would be inconvenient for users who prefer to stay at a single zoom level while mapping without switching modes.

The lasso selection tool (<kbd>Shift</kbd>-drag) only selects nodes, which is adequate if you want to delete or move entire lines or areas. If you want to change tags on multiple lines or areas, then you can use this keyboard shortcut after making a lasso selection.

The keyboard shortcut is especially helpful when multiple ways (lines or areas) are joined at multiple consecutive nodes, such as a power substation that shares its nodes with the fence that surrounds it, or an office building with multiple tenants occupying entire floors. Without this shortcut, users have to try their luck selecting the joined lines or areas, often having to resort to complicated workarounds like creating temporary nodes, disconnecting them, and deleting them one by one.

iD has long had a <kbd>\\</kbd> shortcut (also <kbd>Pause</kbd> on Windows) that works like Potlatch’s <kbd>/</kbd> to cycle among the ways that are joined at the selected node, but until now, it hasn’t had anything like Potlatch’s <kbd>W</kbd> shortcut to select the highlighted way. Potlatch is about to disappear from the Web, so it feels like an appropriate time to bring this feature to iD for safekeeping. (JOSM’s [UtilsPlugin2](https://josm.openstreetmap.de/wiki/Help/Plugin/UtilsPlugin2#Selection) plugin has a similar command, [Adjacent Ways](https://josm.openstreetmap.de/wiki/Help/Action/AdjacentWays).)

## Examples

The following screen recordings depict three hypothetical but realistic examples:

* A dog park surrounded by a fence with a gate at the entrance
* A road that follows a boundary; see https://github.com/openstreetmap/iD/issues/7619#issuecomment-749299874 for some concrete examples
* A three-story building, with a pizza joint occupying most of the ground floor, two offices occupying the entire middle and top floors respectively, and a solar panel on the rooftop

If you select a shared node and press <kbd>Ctrl</kbd><kbd>↑</kbd> or ⌘↑, all the parent ways are selected:

https://user-images.githubusercontent.com/1231218/102862671-4db24580-43e6-11eb-9e90-5a4d67564bd4.mov

Thanks to the multiselection support added in #1761, you can select the parent way you want from the Features list:

https://user-images.githubusercontent.com/1231218/102863482-80a90900-43e7-11eb-9354-5d1f23537b39.mov

If you select a shared node and highlight a specific parent way using <kbd>\\</kbd>, then pressing <kbd>Ctrl</kbd><kbd>↑</kbd> or ⌘↑ causes only that highlighted way to be selected:

https://user-images.githubusercontent.com/1231218/102863754-e2697300-43e7-11eb-80ff-2f933232fb12.mov

Intuitively, if you’re already following the nodes along a specific way (using the <kbd>[</kbd> and <kbd>]</kbd> shortcuts), then you’d want to select that way instead of having to choose it again from the Features list.

If you select multiple shared nodes, pressing <kbd>↑</kbd> or ⌘↑ selects all their parents, even if some of the nodes have different parents than others. In the scenario below, higher-resolution imagery shows that the rooftop solar panel doesn’t extend to the edges of the building, so you’d lasso-select all the nodes that make up the building, select the union of their parent ways, select the solar panel from the Features list, and use the Disconnect and Scale operations to refine the solar panel independently of the rest of the features:

https://user-images.githubusercontent.com/1231218/102927590-1c6c6080-444c-11eb-8231-8de1f2bdc100.mov

## Internationalization

iD seems to generally reserve letters for editing operations and uses control keys or punctuation for navigation, which makes sense. The problem is that there are very few control keys left that aren’t used by iD or the browser, and very few punctuation characters are universal among keyboard layouts. I went with <kbd>Ctrl</kbd><kbd>↑</kbd> on Windows and ⌘↑ on macOS because these are physical keys on every keyboard, regardless of locale.

This combination is unlikely to cause a conflict with browsers. On Windows, <kbd>Ctrl</kbd><kbd>↑</kbd> would move the caret to the beginning of the paragraph, but this shortcut would take effect when no text field has focus. On macOS, ⌘↑ would move the caret to the beginning of a text field or scroll to the top of the page, but iD doesn’t scroll like a page (cue request to jump to the North Pole). It’s also Finder’s shortcut for moving up to the containing folder.

## Other caveats

I’m unsure of a good way to expose this functionality to touchscreens. From #7590, it looks like iD does support multiple selection on touch screens, but I’m unsure about <kbd>[</kbd> and <kbd>]</kbd> navigation. Maybe we could add a context menu item that corresponds to the “Select parent way” command?

Fixes #1239, fixes #2225, fixes #5009, and fixes #7375.

### DIFF
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -159,7 +159,7 @@
             "text": "shortcuts.browsing.vertex_selected.last"
           },
           {
-            "shortcuts": [">"],
+            "shortcuts": ["<"],
             "text": "shortcuts.browsing.vertex_selected.parent"
           },
           {
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -159,6 +159,10 @@
             "text": "shortcuts.browsing.vertex_selected.last"
           },
           {
+            "shortcuts": [">"],
+            "text": "shortcuts.browsing.vertex_selected.parent"
+          },
+          {
             "shortcuts": ["\\", "shortcuts.key.pause"],
             "text": "shortcuts.browsing.vertex_selected.change_parent"
           }
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -159,7 +159,8 @@
             "text": "shortcuts.browsing.vertex_selected.last"
           },
           {
-            "shortcuts": ["<"],
+            "modifiers": ["⌘"],
+            "shortcuts": ["↑"],
             "text": "shortcuts.browsing.vertex_selected.parent"
           },
           {
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2258,6 +2258,7 @@ en:
         next: "Jump to next node"
         first: "Jump to first node"
         last: "Jump to last node"
+        parent: "Select parent way"
         change_parent: "Switch parent way"
     editing:
       title: "Editing"
--- a/modules/modes/select.js
+++ b/modules/modes/select.js
@@ -245,7 +245,7 @@ export function modeSelect(context, selectedIDs) {
             .on(utilKeybinding.minusKeys.map((key) => uiCmd('⇧' + key)), scaleSelection(1/1.05))
             .on(utilKeybinding.minusKeys.map((key) => uiCmd('⇧⌥' + key)), scaleSelection(1/Math.pow(1.05, 5)))
             .on(['\\', 'pause'], nextParent)
-            .on('<', selectParent)
+            .on(uiCmd('⌘↑'), selectParent)
             .on('⎋', esc, true);
 
         d3_select(document)
--- a/modules/modes/select.js
+++ b/modules/modes/select.js
@@ -565,11 +565,12 @@ export function modeSelect(context, selectedIDs) {
 
         function selectParent(d3_event) {
             d3_event.preventDefault();
-            if (_relatedParent) {
-                context.enter(
-                    modeSelect(context, [_relatedParent])
-                );
-            }
+            var parents = _relatedParent ? [_relatedParent] : commonParents();
+            if (!parents) return;
+
+            context.enter(
+                modeSelect(context, parents)
+            );
         }
     };
 
--- a/modules/modes/select.js
+++ b/modules/modes/select.js
@@ -245,6 +245,7 @@ export function modeSelect(context, selectedIDs) {
             .on(utilKeybinding.minusKeys.map((key) => uiCmd('⇧' + key)), scaleSelection(1/1.05))
             .on(utilKeybinding.minusKeys.map((key) => uiCmd('⇧⌥' + key)), scaleSelection(1/Math.pow(1.05, 5)))
             .on(['\\', 'pause'], nextParent)
+            .on('>', selectParent)
             .on('⎋', esc, true);
 
         d3_select(document)
@@ -559,6 +560,15 @@ export function modeSelect(context, selectedIDs) {
             if (_relatedParent) {
                 surface.selectAll(utilEntitySelector([_relatedParent]))
                     .classed('related', true);
+            }
+        }
+
+        function selectParent(d3_event) {
+            d3_event.preventDefault();
+            if (_relatedParent) {
+                context.enter(
+                    modeSelect(context, [_relatedParent])
+                );
             }
         }
     };
--- a/modules/modes/select.js
+++ b/modules/modes/select.js
@@ -23,7 +23,7 @@ import { osmNode, osmWay } from '../osm';
 import * as Operations from '../operations/index';
 import { uiCmd } from '../ui/cmd';
 import {
-    utilArrayIntersection, utilDeepMemberSelector, utilEntityOrDeepMemberSelector,
+    utilArrayIntersection, utilArrayUnion, utilDeepMemberSelector, utilEntityOrDeepMemberSelector,
     utilEntitySelector, utilKeybinding, utilTotalExtent, utilGetAllNodes
 } from '../util';
 
@@ -82,6 +82,33 @@ export function modeSelect(context, selectedIDs) {
 
         selectedIDs = ids;
         return true;
+    }
+
+
+    // find the common parent ways for nextVertex, previousVertex
+    function allParents() {
+        var graph = context.graph();
+        var allParents = [];
+
+        for (var i = 0; i < selectedIDs.length; i++) {
+            var entity = context.hasEntity(selectedIDs[i]);
+            if (!entity || entity.geometry(graph) !== 'vertex') {
+                return [];  // selection includes some not vertices
+            }
+
+            var currParents = graph.parentWays(entity).map(function(w) { return w.id; });
+            if (!allParents.length) {
+                allParents = currParents;
+                continue;
+            }
+
+            allParents = utilArrayUnion(allParents, currParents);
+            if (!allParents.length) {
+                return [];
+            }
+        }
+
+        return allParents;
     }
 
 
@@ -565,7 +592,7 @@ export function modeSelect(context, selectedIDs) {
 
         function selectParent(d3_event) {
             d3_event.preventDefault();
-            var parents = _relatedParent ? [_relatedParent] : commonParents();
+            var parents = _relatedParent ? [_relatedParent] : allParents();
             if (!parents) return;
 
             context.enter(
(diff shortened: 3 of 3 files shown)


data/shortcuts.json
```diff
@@ -158,6 +158,10 @@
             "shortcuts": ["}", "⇟"],
             "text": "shortcuts.browsing.vertex_selected.last"
           },
+          {
+            "shortcuts": ["<"],
```
1ec5: (Repeating this comment in a thread so it’s easier to discuss.)

I couldn’t think of an intuitive keyboard shortcut that was as convenient as the <kbd>[</kbd> and <kbd>]</kbd> shortcuts that it would complement. So I went with <kbd><</kbd>, based on (of all things) the [<code><</code> operator](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Recurse_down_.28.3E.29) in OverpassQL, which recurses up to the parent of a node or way. I’m open to other ideas.

<kbd><</kbd> is a common character on keyboard layouts due to its prevalence in basic math. Just to be on the safe side, I consulted [Apple’s visual guide to physical keyboard localizations](https://support.apple.com/en-us/HT201794) and checked each of the 169 keyboard layouts available on macOS:

* Hindi Devanagari, Khmer, Myanmar, Sinhala, and Tibetan Wylie lack a <kbd><</kbd> key, though I didn’t check whether there are any dead keys that provide access to this character.
* Chinese Pinyin, Chinese Shuangpin, and Thai lack a <kbd><</kbd> key, but Thai keyboards have a <kbd><</kbd>-labeled key, so I suspect the expectation is that users would switch to a QWERTY keyboard when necessary.

To accommodate these keyboard layouts, we could make the keyboard shortcut localizable, relying on an assumption that keyboard layouts generally correspond to languages. If we choose a letter as the keyboard shortcut instead, then it would definitely have to be localizable.

All the other keyboard layouts have a <kbd><</kbd> key, but some require pressing both Shift and AltGr at the same time, which can be inconvenient.

1ec5: I originally went with punctuation because it seems like iD generally reserves letters for editing operations and uses control keys or punctuation for navigation, which makes sense. The problem is that there are very few control keys left that aren’t used by iD or the browser, and very few punctuation characters are universal among keyboard layouts.

An alternative I didn’t consider earlier was <kbd>Ctrl</kbd><kbd>↑</kbd> on Windows or ⌘↑ on macOS. These keys are physical keys on every keyboard, regardless of locale.

This combination is unlikely to cause a conflict with browsers. On Windows, <kbd>Ctrl</kbd><kbd>↑</kbd> would move the caret to the beginning of the paragraph, but this shortcut would take effect when no text field has focus. On macOS, ⌘↑ would move the caret to the beginning of a text field or scroll to the top of the page, but iD doesn’t scroll like a page (cue request to jump to the North Pole). It’s also Finder’s shortcut for moving up to the containing folder.

1ec5: > but iD doesn’t scroll like a page

If the focus is somehow on the sidebar (but not one of its fields), then pressing ⌘↓ jumps to the bottom of the scroll view but ⌘↑ would select the parent ways instead of jumping to the top of the scroll view. I think this situation would be relatively rare, but hopefully there’s a way to special-case the shortcut to only occur when the canvas has focus. Otherwise, that would be a good argument for making the shortcut <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>↑</kbd> on Windows and ⇧⌘↑ on macOS.

quincylvania:
@1ec5 Thanks for this! I hadn't much used iD's vertex navigation keys before, but they're definitely an underrated accessibility feature. It's great to fill in this missing action. Some thoughts:

- The ⌘↑ shortcut sounds fine on paper, but in practice I find it cramped to do with one hand and too far from  <kbd>[</kbd>,  <kbd>]</kbd>, and <kbd>\\</kbd>. My first idea is to use <kbd>|</kbd>, but I'm not sure if that makes sense for international keyboard layouts. Another idea is to just use <kbd>Return</kbd>, but this might cause unexpected issues.
- We should also have a reverse shortcut to put the selection back on the vertex. This would be useful in case of accidental selection, or if you just want to inspect the way before doing something else.

quincylvania:
> My first idea is to use <kbd>|</kbd>

I did this for now, let me know what you think. You can press it again to toggle back to the vertex selection.

1ec5:
> The ⌘↑ shortcut sounds fine on paper, but in practice I find it cramped to do with one hand and too far from  <kbd>[</kbd>,  <kbd>]</kbd>, and <kbd>\\</kbd>. My first idea is to use <kbd>|</kbd>, but I'm not sure if that makes sense for international keyboard layouts. Another idea is to just use <kbd>Return</kbd>, but this might cause unexpected issues.

<kbd>|</kbd> will definitely be problematic. I considered that before <code>&lt;</code>, but there are a lot of keyboard layouts that completely omit that character. (I heard about it all the time from Vietnamese wiki contributors, because the MediaWiki wikitext syntax for internal links includes a <code>|</code>. We worked around it by placing a <kbd>|</kbd> button under the textbox.) The nice thing about the arrows is that we know they exist on every keyboard layout with the same number of keystrokes. If ⌘↑ is unergonomic on laptop keyboards, then we should continue hunting for more common punctuation or, failing that, choose a letter (even though letters are usually for editing operations). Or bind both ⌘↑ and <kbd>|</kbd> to the same operation.

> We should also have a reverse shortcut to put the selection back on the vertex. This would be useful in case of accidental selection, or if you just want to inspect the way before doing something else.

Yes, this is why I tried to find a key that has a companion key we could use for the reverse operation. If ⌘↑ goes “up” to the way, then ⌘↓ can go back down to the vertices (or just the previously selected vertex if known). Another possibility in the future could be to select parent relations using ⌘↑ or relation members using ⌘↓, for completeness.

1ec5:
#8298 continues the discussion about a key binding for this operation.

PR #8276, opened 2021-01-01
Label route relations with directions and waypoints

1ec5:
Extended the display name functionality so that route relations are distinguished by their directions and/or waypoints in addition to their networks and reference numbers.

## Rationale

By convention, the `name` key of many public transportation route relations is set to [a rigid format](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2073385#Step_2_-_Create_the_new_bus_relations) that indicates the route’s waypoints, which does not necessarily reflect the actual name used on the ground. For example, a route relation might be named “Downtown Jobs Express” in reality but `5X: Newtown => Downtown` in OSM.

This convention is primarily intended to disambiguate similar route relations in a list of relations, to avoid mistakes when adding members to a route relation, since it’s very common for there to be multiple routes with the same number in the same area. However, the standard is also to add the structured tags [`from`](https://wiki.openstreetmap.org/wiki/Key:from), [`to`](https://wiki.openstreetmap.org/wiki/Key:to), and [`via`](https://wiki.openstreetmap.org/wiki/Key:via), making the `name` convention redundant and counterproductive (as it obscures the structured data that consumers are more likely to use). Better editor support for the structured data will mitigate the pain that mappers might experience from deprecating the naming convention.

Along for the ride, this PR also extends the display name for road routes in countries that conform to the MUTCD. In these countries, a route’s two directions are distinguished by [cardinal directions](https://wiki.openstreetmap.org/wiki/Highway_Directions_In_The_United_States), which are either tagged as the member ways’ roles or as a [`direction`](https://wiki.openstreetmap.org/wiki/Key:direction) tag on child relations joined by a superrelation.

## Caveats

This PR does not pretty-print or localize the `direction` values. That would be worth doing later once the infrastructure is in place for more general localization of tag values.

Ideally, the additional detail in these names would only appear when there’s ambiguity between two relations in the relation list, similar to how iD automatically appends the relation ID to two relations with identical display names. However, that optimization would be a bit more invasive.

## Example

The following screenshot shows the “Choose a parent relation” dropdown after having loaded six relations, including two superrelations:

* `type=route` `route=road` `network=US:US` `ref=66`
   * `type=route` `route=road` `network=US:US` `ref=66` `direction=east`
   * `type=route` `route=road` `network=US:US` `ref=66` `direction=west`
* `type=route_master` `route_master=bus` `network=RTA` `ref=5`
   * `type=route` `route=bus` `network=RTA` `ref=5` `from=Downtown` `to=County Line`
   * `type=route` `route=bus` `network=RTA` `ref=5` `from=County Line` `to=Downtown`

relations


test/spec/util/util.js
```diff
@@ -225,4 +225,31 @@ describe('iD.util', function() {
             expect(iD.utilUnicodeCharsTruncated('😎😬😆😵😴😄🙂🤔', 255)).to.eql('😎😬😆😵😴😄🙂🤔');
         });
     });
+
+    describe('utilDisplayName', function() {
+        it('returns the name if tagged with a name', function() {
+            expect(iD.utilDisplayName({tags: {name: 'East Coast Greenway'}})).to.eql('East Coast Greenway');
+        });
+        it('distinguishes unnamed features by ref', function() {
+            expect(iD.utilDisplayName({tags: {ref: '66'}})).to.eql('66');
```
1ec5, line 234: This test is [failing](https://github.com/openstreetmap/iD/pull/8276/checks?check_run_id=1632225027):

```
      expected 'Missing en translation: inspector.display_name.ref' to deeply equal '66'
4063
```

Not sure why, since I did add the string to both the YAML file and the compiled JSON file.

data/core.yaml
```diff
@@ -732,6 +732,21 @@ en:
     add_fields: "Add field:"
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
+    display_name:
+      direction: "{direction}"
```
quincylvania, line 736: Do you think any languages need this and `ref` to be translatable? Seems unnecessary even if it makes the code cleaner.

1ec5: I wasn’t sure. Some languages like Chinese and Japanese tend to append a suffix to numbers or direction words, so if they feel the need to do that for the longer strings, they might want to do the same thing in these shorter strings for consistency. But I would expect most languages to leave these strings alone.

quincylvania: Okay, I'll leave it alone then.

test/spec/util/util.js
quincylvania: I'd forgotten that the code tests don't actually load the UI strings by default. It couldn't find this string since it didn't have any of the strings.

k-yle:
Hey @1ec5, this PR has been really helpful! would you consider extending it to show `{ref} {name}` if both exist, perhaps only for bus routes?

image

1ec5:
That’s a good idea. Can you open a separate issue to track making this change? It could be helpful for other kinds of routes too, not just bus routes. I think we’ll want to consider whether to also include the network for consistency with other formats that include the ref, or whether that would unnecessarily crowd out the name.

k-yle:
opened #8559

1ec5:
This feature made it into v2.20.0 by the way.

PR #8288, opened 2021-01-09
Debounce input events in the preset list.

jleedev:
On a slow machine, the preset list is painfully slow to search, as it
rerenders the list on every keystroke, which takes several hundred
milliseconds. By debouncing this, it is able to keep up.

Performance analysis:

Before: Typing the word 'stream' results in a 2.5s frame according to
Chrome's profiler. Each input event takes between 300 and 600 ms to run.

After: Typing the word 'stream' results in three 300ms frames. The first
and last input events take around 200ms, while the four in the middle
are ignored by the debouncer.

### DIFF
--- a/modules/ui/preset_list.js
+++ b/modules/ui/preset_list.js
@@ -1,5 +1,6 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
 import { select as d3_select } from 'd3-selection';
+import _debounce from 'lodash-es/debounce';
 
 import { presetManager } from '../presets';
 import { t, localizer } from '../core/localizer';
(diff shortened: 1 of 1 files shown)


bhousel:
Actually can we hold off on this change? 
I have a [nsi-v5 branch](https://github.com/openstreetmap/iD/tree/nsi-v5), work in progress on integrating the latest name-suggestion-index where I've sped up the preset_list considerably.  (it was recalculating the selected entity's extent on every keystroke).

jleedev:
Checked out that branch and profiled: Typing 'stream' produces a 1500ms frame. The `drawList` call is taking 200ms in its own right. I believe there's still value in making the search async.

bhousel:
> The `drawList` call is taking 200ms in its own right. I believe there's still value in making the search async.

Yes I agree!  Sorry, I should have phrased it more like - there is probably value in this change, I just don't want it to conflict what else I'm working on.

mbrzakovic (approved):

PR #8305, opened 2021-01-18
Name-suggestion-index v6

### LINKED ISSUE #5484: Autofill "operator:wikidata" using the Operator field
Currently, when typing in an operator for an object in iD, all I can enter is plain text.  Could a wikidata search be used to find the operator that the contributor is looking for and suggest it to them? 
For example, in the image below, I am editing a bus stop and trying to add the operator.  Instead of just including the text "Chapel Hill Transit", iD could autofill a tag, "operator:wikidata"="Q5073024".
image
I think that something like this could be very useful if implemented.

Also, for anyone super interested in the image, it is of [this location](https://www.openstreetmap.org/edit#map=19/35.90602/-79.04464).

### LINKED ISSUE #6055: Do not use directly discardNames name-suggestion-index
See https://github.com/openstreetmap/iD/issues/5930

`discardNames` from name suggestion index is directly used by iD.

So listing names there will cause iD to list them as invalid and encourage people to delete them. 

To avoid this one may add names to `discardKeys` that are not brands but can be used but it makes contributing to name-suggestion-index more complicated. See https://github.com/osmlab/name-suggestion-index/pull/2488#issuecomment-473830149

I think that it would be preferable to have iD maintain its own list, based on name-suggestion-index.

Just list filters in `discardNames` that are discarding invalid names (and use them in validator), list of filters that are harmful in iD and have ability to list filters not appearing in either first or second list.

---

In case of creation of separate set of entries it may be worth syncing with JOSM - see https://josm.openstreetmap.de/browser/josm/trunk/data/validator/unnecessary.mapcss#L93

### LINKED ISSUE #6408: Make tag diff more discoverable
Personally I was unaware there was a way to see the tag diff before applying an autofix until I saw it in a screenshot. Before learning about it I was actually undoing/redoing to check which tags changed.

I think to improve usability any autofix that changes tags should show the user the diff before it is applied (without having to click the information button), otherwise from user perspective you're applying a fix which you don't actually know what it's doing.

This can also serve as an indirect learning method to see which tagging is preferred.

Since UI space is precious and it'd probably look bad to always have the diff there, showing it when hovered over the fix seems reasonable.

### LINKED ISSUE #6890: problems with trying to add a name for post boxes that are tagged with brand
Hi,
please have a look at https://www.openstreetmap.org/node/2293452747

this post box is tagged with brand=Deutsche Post because it's not just the operator, but a kind of brand acting in the post business in germany. The issue is now that iD wants to add not only the correct brand:wikipedia tag, but also "name=Deutsche Post" which would be clearly wrong for a post box, wouldn't it?

### LINKED ISSUE #8143: Show what changes are suggested along with the apply button rather than hiding what changes will be done by default
being asked to apply unspecified changes

Rather than asking people to apply unspecified changes and requiring an extra click to see what change you will be making, iD should show what tagging changes it is proposing. Honestly, nobody should be clicking automatic edit buttons unless it's an approved import and their click is part of that approved process, but these tag upgrades are not an import or they could be done mechanically. I'm not proposing an annoying confirmation pop-up or anything, just fold out the proposed changes by default so people can tell, often at a glance, if it is expected and good.

expanded

The sentence "Some features should have additional tags." seems redundant to me, if we want to save some space while we're at it. The info button could be made into a collapse symbol or removed altogether since people can collapse the Issues section if they don't want to see this.

If there is no objection, I could write a patch to resolve this issue.

### LINKED ISSUE #8237: Name-suggestion-index preset pack
Following #8229, iD now fetches preset data at runtime instead of bundling it with the release. The name-suggestion-index maintains its own presets, and has started generating iD-compatible files (see https://github.com/osmlab/name-suggestion-index/issues/4718). We need to update iD to integrate presets from multiple sources.

### LINKED ISSUE #8271: iD 2.19.5 cannot distinguish gas station and c-store with the same name
I'm having this issue for multiple gas brands in the US:

Here is the reproduction: https://imgur.com/a/Pabf2TN

When two kinds of places share the same name, iD cannot pick up the one I clicked. And reselect category from "Feature Type" does not help. I can only bypass this weird behavior by deleting tag first then reselect the correct type of feature.

### LINKED ISSUE #8304: iD craftily changes POI type
I just typed the name of a butcher shop and after it, I clicked somewhere and the butcher shop was magically changed to a shoe shop with Russian name. WTH?
iD didn't even ask me if there are some "missing tags", just changed it.

Marko
Marko2

### LINKED ISSUE #8426: iD wants to change shop value despite better match in NSI
Yves Rocher can be tagged as `shop=beauty` or `shop=cosmetics` and both are valid an in NSI

But iD want to "upgrade" while changing shop type, without a good reason

https://www.openstreetmap.org/edit?editor=id&node=3845799681#map=19/51.05281/16.97383
screen06

From what I see NSI has correctly also `shop=cosmetics` - see https://github.com/osmlab/name-suggestion-index/blob/ea9d441ffb3e1594fca160910219929b4393eb12/data/brands/shop/cosmetics.json#L390

----------

triggered by https://forum.openstreetmap.org/viewtopic.php?pid=823920#p823920 where local mapper complains about people incorrectly changing `shop=cosmetics` into `shop=beauty` due to "upgrade" button in iD

bhousel:
This updates the iD code to support the latest name-suggestion-index.  I'll release v5 in the next few days, but I have tested this locally and works awesome.

Some enhancements:
- The presets and fields have been upgraded to fully support locationSets now
- Added a `coreLocations` module to resolve and cache locationSets used anywhere in iD (presets, community-index, name-suggestions).  Processing happens in chunks during browser idle callbacks.
- NSI presets are downloaded and merged in after iD starts up. The presets become available within a few seconds. 
- Upgraded parts of the code (field locking, map pin styling) that previously only supported `brand:wikidata` to now support the other kinds of features in NSI.  (flags, operators, transit networks)
- Upgraded the outdated tags and suspicious tags validators to use the latest NSI data and matcher.  
- Note about the validator: there is an interesting issue here in that the first few tiles downloaded from OSM won't be fully validated because the NSI data isn't ready yet.  This was always an issue before but it is a bit more annoying now.  I'd like to Promisify all the validators so they just run whenever they are ready to run (update: solved it in #8319 and merged it here)

I'll have to look and see what outstanding issues this closes, but it's all the ones I know about.

re: https://github.com/osmlab/name-suggestion-index/issues/4543
closes #8426
closes #8304
closes #8271 
closes #8237 
closes #8143
closes #6890
closes #6055
closes #6408
closes #5484


modules/validations/outdated_tags.js
```diff
@@ -17,58 +16,126 @@ let _nsi;
...
+
       return _nsi;
     })
     .catch(() => { /* ignore */ });
 
 
+  // Returns true if this tag key is a "namelike" tag that the NSI matcher would have indexed..
+  function isNamelike(k) {
+    const namePatterns = [
+      /^(flag:)?name$/i,                                             // e.g. `name`, `flag:name`
```
1ec5: openstreetmap/id-tagging-schema#114 requests a `flag:name` field for the Flagpole preset.

modules/validations/outdated_tags.js
```diff
@@ -17,58 +16,126 @@ let _nsi;
...
+      /^\w+_name:\w+$/i                                              // e.g. `alt_name:en`, `short_name:ru`
+    ];
+
+    return namePatterns.some(pattern => {
+      if (!pattern.test(k)) return false;    // k is not a name tag, skip
+
+      // There are a few exceptions to the namelike regexes.
+      // Usually a tag suffix contains a language code like `name:en`, `name:ru`
+      // but we want to exclude things like `operator:type`, `name:etymology`, etc..
+      if (/:(colour|type|left|right|etymology|wikipedia)$/.test(k)) return false;
```
1ec5: `name:pronunciation` is another common subkey of `name` that isn’t a language code. That said, IPA is very unlikely to match a name in NSI anyways.

modules/validations/outdated_tags.js
```diff
@@ -79,7 +146,7 @@ export function validationOutdatedTags() {
       }
     }
 
-    // add missing addTags..
+    // Add missing addTags from the detected preset
     let newTags = Object.assign({}, entity.tags);  // shallow copy
     if (preset.tags !== preset.addTags) {
       Object.keys(preset.addTags).forEach(k => {
```
1ec5, line 55: Each flag entry in NSI comes with a `flag:name` tag, but that’s mainly for internal reference. The `flag:name` key would logically be in the local language, whereas the value in NSI is specific to English, so it wouldn’t be appropriate to add the `flag:name` from NSI in non-English-speaking countries. (Apologies if the code below already handles that; I couldn’t tell.)

modules/validations/outdated_tags.js
```diff
@@ -93,63 +160,120 @@ export function validationOutdatedTags() {
...
+          const itemID = match[0].itemID;
+          const item = _nsi.ids.get(itemID);
+          const mainTag = item.mainTag;               // e.g. `brand:wikidata`
+          const itemQID = item.tags[mainTag];         // e.g. `brand:wikidata` qid
+          const notQID = newTags[`not:${mainTag}`];   // e.g. `not:brand:wikidata` qid
+
+          // Exceptions, throw out the match
+          if (
+            (!itemQID || itemQID === notQID) ||       // no `*:wikidata` or matched a `not:*:wikidata`
+            (newTags.office && !item.tags.office)     // feature may be a coprorate office for a brand? - #6416
```
1ec5: Would a corporate office still be a decent candidate for an `operator` preset?

modules/ui/fields/input.js
```diff
@@ -34,11 +34,41 @@ export function uiFieldText(field, context) {
...
+                }
+
+                // Features linked to Wikidata are likely important and should be protected
+                if (entity.tags.wikidata) return true;
+
+                var preset = presetManager.match(entity, context.graph());
+                var isSuggestion = preset && preset.suggestion;
+
+                // Lock the field if there is a value and a companion `*:wikidata` value
+                return isSuggestion && !!entity.tags[which] && !!entity.tags[which + ':wikidata'];
```
1ec5, line 54: `flag:name` goes with `flag:wikidata` rather than `flag:name:wikidata`. (Sorry, I didn’t think of the inconsistency when promoting `flag:name` instead of `flag`.)

Maybe this is a happy accident, since the mapper may need to come up with the flag name in the local language anyways. Also, for some kinds of flags, it’s common enough for the same flag to represent a different entity and be known by a different name. (The example that comes to mind is the flag of the Vatican City, which is also a religious flag representing the Holy See, but there are also plenty of examples among historical flags.)

1ec5 (commented):
Exciting to see more parts of NSI become accessible to ordinary mappers. This is the first time I’m looking at the matcher code, so it’s possible that you’ve already accounted for some of the feedback I’ve given below.

modules/core/file_fetcher.js
```diff
@@ -19,8 +19,14 @@ export function coreFileFetcher() {
     'keepRight': 'data/keepRight.min.json',
     'languages': 'data/languages.min.json',
     'locales': 'locales/index.min.json',
-    'nsi_brands': 'https://cdn.jsdelivr.net/npm/name-suggestion-index@4/dist/brands.min.json',
-    'nsi_filters': 'https://cdn.jsdelivr.net/npm/name-suggestion-index@4/dist/filters.min.json',
+
+    'nsi_presets': 'https://raw.githubusercontent.com/osmlab/name-suggestion-index/main/dist/presets/nsi-id-presets.min.json',
```
bhousel: Oh I should add - I don't actually want to load these from the GitHub `main` branch.. 
These URLs just for testing it.

modules/core/locations.js
```diff
@@ -0,0 +1,267 @@
...
+  //  After resolving and indexing, the Objects will be decorated with a
+  //  `locationSetID` property.
+  //  [
+  //   { id: 'preset1', locationSet: {…}, locationSetID: '+[Q2]' },
+  //   { id: 'preset2', locationSet: {…}, locationSetID: '+[Q30]' },
+  //   { id: 'preset3', locationSet: {…}, locationSetID: '+[Q2]' },
+  //   …
+  //  ]
+  //
+  //  Returns a Promise fullfilled when the resolving/indexing has been completed
```
bhousel: haha spellchecker

modules/ui/fields/input.js
bhousel: > Maybe this is a happy accident, since the mapper may need to come up with the flag name in the local language anyways.

Yes it is 👍   It means this code would lock a `flag` field, but not a `flag:name` field, which is what I think you want..

modules/validations/outdated_tags.js
```diff
@@ -17,58 +16,126 @@ let _nsi;
...
-  fileFetcher.get('nsi_brands')
-    .then(d => {
+
+  function delay(msec) {
+    return new Promise(resolve => {
+      window.setTimeout(resolve, msec);
+    });
+  }
+
+  // This Promise will fullfill after NSI presets are loaded and locations merged into the locationManager.
```
bhousel: Well at least my misspelling is consistent.

modules/validations/outdated_tags.js
bhousel: Yeah I _think_ that's ok - This code is just collecting values to attempt to match, so if collects a few extras it's probably fine, but will waste a few microseconds.

modules/validations/outdated_tags.js
bhousel: > (Apologies if the code below already handles that; I couldn’t tell.)

It doesn't but it should - this is a good catch..

modules/validations/outdated_tags.js
```diff
@@ -93,63 +160,120 @@ export function validationOutdatedTags() {
...
+          ) {
+            match = null;   // forget match and keep looking
+            continue;       // (it might make sense to stop looking, not sure)
+          }
+
+          // We are keeping the match at this point
+          subtype = 'noncanonical_brand';
+
+          // Preserve some tags values that we don't want NSI to overwrite.
+          const keepTags = ['takeaway', 'building']
```
bhousel: @1ec5 we should add `flag:name` here to leave people's local flag names alone.

modules/validations/outdated_tags.js
bhousel: I'm not sure.. I kind of think no, except in cases where it's really widespread or like a chain.  (Military recruiting centers maybe or DMVs?)

I suppose any business can also have a corporate office.  Do people tag them like `office=yes`?  Adding an extra preset for each one seems like it would just confuse users.

For reference, these are the brand-like `office=*` categories NSI tracks today:
Screen Shot 2021-01-18 at 7 17 31 PM

tas50:
@bhousel Any chance you can push v5 so this can be tested locally?

bhousel:
> @bhousel Any chance you can push v5 so this can be tested locally?

What I've been doing for now is using [`npm link`](https://dev.to/erinbush/npm-linking-and-unlinking-2h1g) to test locally:
```bash
cd path/to/name-suggestion-index
npm link
cd path/to/iD
npm link name-suggestion-index
```

This makes a symlink between the two projects, which is great for local development.. 
But as soon as I'm really sure we won't need to change anything else about NSI's matcher I'll release it.

modules/validations/outdated_tags.js
bhousel: I removed it anyway in a32553595

modules/validations/outdated_tags.js
bhousel: I did this in f61a3ef5e, and it works:

before | after
--- | ---
Screen Shot 2021-01-18 at 9 37 04 PM | Screen Shot 2021-01-18 at 9 42 27 PM

tas50:
@bhousel That is some JS magic right there. You're making me sad to be in Ruby land. I was able to fire up the branch just fine with link. It does look like some places are too aggressive suggesting name changes. That feels like the iD side of things.

Here's a Portland independent car repair shop that's tagged as repairing Volvos with brand:volvo

Screen Shot 2021-01-18 at 7 52 00 PM

Here's a hotel with a custom name set:

Screen Shot 2021-01-18 at 7 51 35 PM

This one is real strange, but it made me chuckle:

Screen Shot 2021-01-18 at 7 52 31 PM

bhousel:
> @bhousel That is some JS magic right there. You're making me sad to be in Ruby land. I was able to fire up the branch just fine with link. It does look like some places are too aggressive suggesting name changes. That feels like the iD side of things.

Thanks @tas50 - these are great examples.. Let me know if you see any more like this.
There is some logic in the field UI for "whether the name field should be readonly/locked" and I think I can put some similar logic into the validator for "whether the name tag should be replaced with what is in NSI".

I'm really curious about what's going on with the "Vacant" Applebees. Is it using a tag like `name:old` or `name:disused`? 🤔  
If you have the node/way id, I'd like to take a look!

tas50:
@bhousel It turns out the Applebees was decom'd wrong and had the official_name tag left so that sort of makes sense. I found a few more places where things are a bit too aggressive overwriting name fields:

Wireless stores that have a brand set. I think these are sort of mistagged, but that's a whole different issue. They're basically Verizon stores with an operator of XYZ, but they get tagged as XYZ with the brand of Verizon.
Screen Shot 2021-01-18 at 10 12 10 PM

1ec5:
> Wireless stores that have a brand set. I think these are sort of mistagged, but that's a whole different issue. They're basically Verizon stores with an operator of XYZ, but they get tagged as XYZ with the brand of Verizon.

It’s probably one of those Verizon Authorized Retailers where either the store name or Verizon could have top billing. It’s hard to say what other POI categories tend to have store-specific names; it probably varies from country to country.

bhousel:
@tas50 & @1ec5  I just pushed a change that should make the name overwriting much less aggressive.

Basically the rule works the same as the way it has worked before for Name field locking 🔒 :
* if the preset _does_ display a "Brand" field, the Brand field gets locked and the Name field is editable.  NSI _will not_ overwrite the name tag.
* if the preset _does not_ display a "Brand" field, the Name field is locked.  NSI _will_ overwrite the name tag.
* (same rules above apply for presets that use "Operator" instead of "Brand")

This means that we can control which categories we expect mappers to supply store-specific names (Car Dealerships, Hotels, Post Offices, etc).  AFAIK the presets already set up correctly, but we can adjust in the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) repo if needed..

UKChris-osm:
Some great with this @bhousel thank you 👍 

How's it looking with adding an operator within a location, such as an Ambulace Station in the [South West of England](https://ideditor.github.io/location-conflation/?locationSet=%7B%22include%22%3A%5B%22gb-south-west.geojson%22%5D%7D&referrer=nsi)?

Does placing a node / way in the area suggest [South Western Ambulance Service](https://nsi.guide/index.html?t=operators&k=emergency&v=ambulance_station#southwesternambulanceservice-bba5ce) as expected?

bhousel:
> How's it looking with adding an operator within a location, such as an Ambulace Station in the [South West of England](https://ideditor.github.io/location-conflation/?locationSet=%7B%22include%22%3A%5B%22gb-south-west.geojson%22%5D%7D&referrer=nsi)?
> Does placing a node / way in the area suggest [South Western Ambulance Service](https://nsi.guide/index.html?t=operators&k=emergency&v=ambulance_station#southwesternambulanceservice-bba5ce) as expected?

Yes, it is working wonderfully!  
Thanks for your help in contributing POIs for this region to the index 🙇  

I see this when I type "south" around **Plymouth, South West England**:

Screen Shot 2021-01-20 at 3 24 11 PM

I see this when I try it around **Cardiff, Wales**:

Screen Shot 2021-01-20 at 3 27 38 PM

UKChris-osm:
Fantastic :)

I've enjoyed contributing :)

What happens if you place a node in the South West but start typing "_Ambulance_" for example, will it suggest a branded operator Ambulance Station as the first result, or does the generic Ambulance Station POI within iD still dominate?

bhousel:
> What happens if you place a node in the South West but start typing "_Ambulance_" for example, will it suggest a branded operator Ambulance Station as the first result, or does the generic Ambulance Station POI within iD still dominate?

Screen Shot 2021-01-20 at 4 18 59 PM

It matches the generic presets first, but the branded ones rank pretty highly..
Here's what I see in Cornwall - does this look reasonable?
(I'm sure we can scope the locationSets a bit better, but at least we're not seeing anything from other countries in this list)

UKChris-osm:
Yes, looks like it's working great 👍 

I am hoping to add a few more .geojson's tomorrow / Friday as the first two items (North and East) aren't based in Cornwall, but I expect them to appear at the moment as they are "gb-eng" so shouldn't appear in, say, Wales.

Also, I guess pressing the "i" button next to each one above would display a bit of text from Wikipedia / Wikidata?

UKChris-osm:
> Here's what I see in Cornwall - does this look reasonable?

I've covered England with some .geojson files for each region, and assigned police / fire / ambulance stations to these, so the locationing should be better now :)

modules/validations/outdated_tags.js
```diff
@@ -93,63 +158,148 @@ export function validationOutdatedTags() {
...
+          const itemID = match[0].itemID;
+          const item = _nsi.ids.get(itemID);
+          const mainTag = item.mainTag;               // e.g. `brand:wikidata`
+          const itemQID = item.tags[mainTag];         // e.g. `brand:wikidata` qid
+          const notQID = newTags[`not:${mainTag}`];   // e.g. `not:brand:wikidata` qid
+
+          // Exceptions, throw out the match
+          if (
+            (!itemQID || itemQID === notQID) ||       // no `*:wikidata` or matched a `not:*:wikidata`
+            (newTags.office && !item.tags.office)     // feature may be a coprorate office for a brand? - #6416
```
peternewman: Typo:
```suggestion
            (newTags.office && !item.tags.office)     // feature may be a corporate office for a brand? - #6416
```

PRed into Codespell so would flag up eventually...

tas50:
@bhousel It looks like the new flag pole logic is a bit aggressive as well:

Flag pole with multiple flags defined:
image

Coast Guard signal flag:
image

modules/validations/outdated_tags.js
bhousel: fixed it in 3640e15fa, thanks for catching 👍

peternewman: Thanks!

bhousel:
> It looks like the new flag pole logic is a bit aggressive as well:

Thanks for catching these @tas50 .. I'll adjust how the code handles flagpoles (this is all new and we're still learning).

> Flag pole with multiple flags defined:

I think this one happened because there is a `country=US` tag.  
The [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dflagpole) says: "For national flags to indicate which country is represented on the flag (examples: United States of America, Germany, Tanzania)." so it's a bit incorrect to add this tag to a flagpole with multiple flags on it, but I can see why people do it and the wiki is just a suggestion anyway.  I'll add some code to 1) skip any feature where the namelike tag contains semicolons, 2) consider `country` after other values.  

My concern here is that people may tag things like `flag:name=NJ`+`country=US` which is again - kind of wrong but I can see why people would do it - and we don't want to suggest an "upgrade" to the United States national flag.  So I have to think about this more.

> Coast Guard signal flag:

This one probably happened because there is an `operator=` tag with a value like "US Coast Guard" or "USCG" or something.  On the NSI side we have a hierarchy where things are placed into "trees" and there is better logic for knowing which tags matter for which trees.  So I need to push this ability into the iD code too.

1ec5:
>> I think this one happened because there is a `country=US` tag.  
> 
> The [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dflagpole) says: "For national flags to indicate which country is represented on the flag (examples: United States of America, Germany, Tanzania)." so it's a bit incorrect to add this tag to a flagpole with multiple flags on it, but I can see why people do it and the wiki is just a suggestion anyway.  I'll add some code to 1) skip any feature where the namelike tag contains semicolons, 2) consider `country` after other values.  

Yeah, there’s no good answer to this one. `country=US` is correct for one part of `flag:wikidata` but there’s no correct value for the other part. In theory, we could come up with a lanes-like syntax so that `country=US|` would correspond to `flag:wikidata=Q42537|Q330517`, but the whole point of continuing to use `country` on flagpoles is graceful degradation in existing renderers like OSM2World (which doesn’t understand multiple flags). If the validator skips features with `flag:wikidata` set to multiple values, then I think that’s good enough to sidestep the larger issue of arrays-of-key-value-stores for now.

tas50:
I found another weird operator suggestion @bhousel where an emergency room tag is being removed

Screen Shot 2021-02-13 at 5 54 28 PM

tas50:
Amazon Lockers are also a bit tricky since each one has a custom name and it's handy to know that when you schedule delivery

image

peternewman:
> Amazon Lockers are also a bit tricky since each one has a custom name and it's handy to know that when you schedule delivery
> 
> image

Normally, like Weatherspoons pubs, we'd just remove the name in the preset:
https://github.com/osmlab/name-suggestion-index/blob/504a8a2a306abec8cba118a9f90c35a297cae765/data/brands/amenity/vending_machine.json#L49

Which will stop it overwriting.

At first I thought @bhousel was suggesting something smarter would happen here:
> * if the preset _does_ display a "Brand" field, the Brand field gets locked and the Name field is editable.  NSI _will not_ overwrite the name tag.

But I think that relies on the brand field being shown in the Parcel Pickup Locker preset (which perhaps it should).

I wonder if the other name "Hemlock" should be the ref perhaps?

Adamant36:
> I wonder if the other name "Hemlock" should be the ref perhaps?

That would be my guess. Or something else since the ref key is really more to do with numbers (I'd be extremely surprised if there were not legitimate reference numbers for Amazon Lockers that should be in the ref field and not just location names).

1ec5:
[`branch`](https://wiki.openstreetmap.org/wiki/Key:branch) might be more appropriate than `ref`. I don’t have a strong opinion on where to put that information for Amazon lockers, but sometimes a `name` or `official_name` is the only way a data consumer knows to concatenate `brand`, `ref`, `branch`, etc. into a single coherent, human-readable name. It’s common to give post offices individual branch names, so it makes sense to me that a non-office delivery point might be named similarly.

bhousel:
>Amazon Lockers are also a bit tricky since each one has a custom name and it's handy to know that when you schedule delivery

Good catch @tas50 - I put some thoughts here on the Green King PR, as this is a thing that keeps coming up and I have to figure out:
https://github.com/osmlab/name-suggestion-index/issues/4906
tl;dr - NSI should keep track of whether to apply the name tag as a default only, or stictly (overwrite what's there).

> I found another weird operator suggestion @bhousel where an emergency room tag is being removed

Ugh, also good catch, and I know why this is.. It's because `emergency=*` is also used as a toplevel/defining tag for things like ambulance and rescue stations, see here: https://github.com/osmlab/name-suggestion-index/tree/main/data/operators/emergency
When OSM uses a single tag in multiple ways, stuff always gets weird.

I'll fix these..

peternewman:
> That would be my guess. Or something else since the ref key is really more to do with numbers (I'd be extremely surprised if there were not legitimate reference numbers for Amazon Lockers that should be in the ref field

I'm sure somewhere in their database their is a more standardised identifier, but they seem to work really hard to hide it on their website. Probably because people can't accidentally mix up a name as 1243 for example. Similar to What 3 Words.

> and not just location names).

In my experience they are just an arbitrary dictionary word with no relevance to anything, let alone the location/site/area, it's just an identifier.

bhousel:
> In my experience they are just an arbitrary dictionary word with no relevance to anything, let alone the location/site/area, it's just an identifier.

The names are funny but I actually think they are prominently displayed in the app and in real life.. If you go to the location, it says that you are at "Nunzio" or "Pail".  For example Google Maps shows this:

Screen Shot 2021-02-16 at 10 27 06 AM

tas50:
Another one that might be a bit too aggressive. This walmart was tagged as a supemarket and a pharmacy, but named for the supermarket:

image

peternewman:
> > In my experience they are just an arbitrary dictionary word with no relevance to anything, let alone the location/site/area, it's just an identifier.
> 
> The names are funny but I actually think they are prominently displayed in the app and in real life.. If you go to the location, it says that you are at "Nunzio" or "Pail". For example Google Maps shows this:

Yes sorry, I didn't word that very well. I'd agree "Pail" is the most important way to know you're at the right Amazon Locker, but you need a map location to do that, you can't infer it's in Plainfield from the name (compared to say a supermarket where it would be "Costco Plainfield" for example).

Is there a beta site where we can start test-driving this @bhousel or do we need to do stuff locally to test it?

bhousel:
> Is there a beta site where we can start test-driving this @bhousel or do we need to do stuff locally to test it?

Yeah I would like this too - I think I can put something together soon.

I've been steadily picking my way though https://github.com/osmlab/name-suggestion-index/issues/4906 , which will give the iD validator a lot more flexibility in how it treats named things.

bhousel:
I spent a bunch of time on the NSI internals and we now have a lot more flexibility in how `name` and other tags are handled. 

I just tested some of the things that have come up in the testing and the latest version is looking pretty great 👍 :
- [x] Banner Goldfield Medical Center - don't overwrite `emergency` tag
- [x] Amazon Locker - allow custom `name`
- [x] Apple Store - allow custom `name`
- [x] Greene King - allow custom `name` 

This one is still tricky
- [x] Walmart Neighborhood Market + `amenity=pharmacy`

The validator is now able to prioritize the name and supermarket, but it does suggest to remove the `amenity=pharmacy` tag.  I sort of understand why, and I agree with the suggestion that the pharmacy tags should not be combined with the shop tags, but I might put something special into the code to avoid surprises like this, since this is probably a widespread mapping practice.


Screen Shot 2021-03-03 at 2 48 11 PM

I'll try to get a version of this up for people to play around with asap.

tas50:
Removing double tags might surprise some folks, but it's really the right thing to do for the map in general. I see a lot of combined drugstore/pharms and convenience store / gas stations. That's now how they should get tagged in OSM and it doesn't even align with how Apple or Google maps things. Having the editor provide another nudge to get users off the combined tagging model is a plus in my eyes.

bhousel:
I deployed this branch with the preview NSI v5 code to:  http://www.ideditor.org/nsi-v5/
Try it out!

UKChris-osm:
When trying a couple of samples in [South West of England](http://www.ideditor.org/nsi-v5/#background=Bing&disable_features=boundaries&map=16.80/50.97205/-3.18294):

- Typing "_Dev_" will suggest [Devon and Somerset Fire and Rescue Service](https://nsi.guide/index.html?t=operators&k=amenity&v=fire_station#devonandsomersetfireandrescueservice-ab1fb5), as expected, but when typing in "_fire station_" didn't suggest any NSI fire station entries at all. When typing "_Poli_" then Avon & Somerset police did show up, as did South West Ambulance when typing in "_Amb_".
- When typing in "_Emergency Phone_", Highways England didn't show up at all, which I was expecting.

Is this because the matcher only works on names, and not types?

bhousel:
> Is this because the matcher only works on names, and not types?

Hey @UKChris-osm - yes what you're seeing is because of how iD preset matching works.  It is set to return the core presets before the suggestion presets.

>  but when typing in "fire station" didn't suggest any NSI fire station entries at all.

I actually do see these but they are far down in the list below around 15 core presets.  There are just a lot of presets that match "fire"

> When typing in "Emergency Phone", Highways England didn't show up at all, which I was expecting.

This is weird to me too - I'll play around and see if I can get the iD preset matcher to return more relevant results.

ALSO - I just pushed the latest version to http://ideditor.org/nsi-v5/ for testing, which improves a bunch of things.
(we're tracking some issues both here and on https://github.com/osmlab/name-suggestion-index/issues/4543)

UKChris-osm:
> I actually do see these but they are far down in the list below around 15 core presets. There are just a lot of presets that match "fire".

Ah yes, I see now they are listed, just further down off screen. Is there scope to make the suggestions a two step process, so if I were to select the generic "_fire station_" iD preset, that the fire stations within the NSI would then be suggested in a secondary way?

Also, when I added a node as a generic fire station, I typed in "_Devon and Somerset_" as the operator, then clicked away, then clicked back, but no suggestion came up suggesting Devon and Somerset Fire and Rescue Service.

bhousel:
> Ah yes, I see now they are listed, just further down off screen. Is there scope to make the suggestions a two step process, so if I were to select the generic "_fire station_" iD preset, that the fire stations within the NSI would then be suggested in a secondary way?

I don't think we could introduce 2 steps of preset selection, but I can try to make the results more relevant.

> Also, when I added a node as a generic fire station, I typed in "_Devon and Somerset_" as the operator, then clicked away, then clicked back, but no suggestion came up suggesting Devon and Somerset Fire and Rescue Service.

The validator would suggest it if there is a `matchName` or some kind of alternate name like "Devon and Somerset".  
It needs to match something in NSI, by [key, value, name, location].

bhousel:
>> The validator is now able to prioritize the name and supermarket, but it does suggest to remove the amenity=pharmacy tag. I sort of understand why, and I agree with the suggestion that the pharmacy tags should not be combined with the shop tags, but I might put something special into the code to avoid surprises like this, since this is probably a widespread mapping practice.
>
> Removing double tags might surprise some folks, but it's really the right thing to do for the map in general. I see a lot of combined drugstore/pharms and convenience store / gas stations. That's now how they should get tagged in OSM and it doesn't even align with how Apple or Google maps things. Having the editor provide another nudge to get users off the combined tagging model is a plus in my eyes.

I agree 100% that it's the right thing to do, but I took out the code anyway because we have run into it a few times now:
- removing `emergency=yes` from a hospital [above](https://github.com/openstreetmap/iD/pull/8305#issuecomment-778707926) (because `emergency` is also a toplevel tag for phones, ambulance stations, etc), 
- removing `man_made=surveillance` tag [here](https://github.com/osmlab/name-suggestion-index/issues/4543#issuecomment-791913156) (because `man_made` is also a toplevel tag for flagpoles)
- removing `amenity=pharmacy` from Walmart Neighborhood Market [above](https://github.com/openstreetmap/iD/pull/8305#issuecomment-782565876) (because `amenity` and `shop` are both toplevel tags)

So for now, I think we have to just leave these in.  I adjusted the code to only remove unspecific tags like `amenity=yes`, or `shop=yes` (but not `emergency=yes` 🙄 ).

I'll try to think about whether someday we can find a way to nudge people towards following the ["one feature, one element" rule](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element), in a way that doesn't surprise or remove data..

This is really the biggest challenge with OSM's free tagging model: tags are used to identify _what a feature is_, but occasionally those same tags are used to identify _something a feature has_.

bhousel:
I pushed the latest work to https://ideditor.org/nsi-v5 - let me know what you think!

peternewman:
> In future the iD upgrade could also populate the branch/ref field where it's currently only included in the name.

_Originally posted by @peternewman in https://github.com/osmlab/name-suggestion-index/issues/4906#issuecomment-782896882_

I see it's now offering this thanks @bhousel e.g. for https://www.openstreetmap.org/way/217820293

It's currently also proposing to strip it from the name. Is there some config for when it appears on the store itself too (I don't know if it does in this case)/how's that all being handled?

UKChris-osm:
If a user selects the suggested tags from the NSI, should some sort of reference be included in the source tag, either on the element itself, or within the changeset log?

Adamant36:
"If a user selects the suggested tags from the NSI, should some sort of reference be included in the source tag, either on the element itself, or within the changeset log?" 

I've thought about that myself. At least with it being mentioned in a changeset comment (doing it on the element itself is a bad idea IMO). 

The problem is that the "anti-automated edit" crowed in OSM is way to bullish in their opposition to anything having to do with iD Editor or "automated edits", there isn't clear definition of what constitutes ones and the people who are ardently against them aren't willing to provide either, and so declaring how an edit was made if it was done in iD Editor with the NSI a recipe for targeting/harassment/and arbitrary blocks/reverts that are based more on "automated edit" fear mongering then if the edit was actually correct or not. 

Other software has presets and don't have to declare them as the source in changeset comments either. There isn't anything special about the NSI that it should have to compared to any other software that does similar things. 

The "source" could just as easily be "I verified this myself and I used the preset after doing so" anyway.

bhousel:
> It's currently also proposing to strip it from the name. Is there some config for when it appears on the store itself too (I don't know if it does in this case)/how's that all being handled?

Hey @peterneubauer - I hadn't replied to this before because I was still trying a few different approaches in code, but what I settled on was basically:
- the `name` tag is special and the code that looks for name candidates will also look for branches in the `name` tag
- it does it by breaking the name value into pieces and trying to match those fragments longest to shortest until it finds a name:
  - "TUI ReiseCenter Neuss Innenstadt" - no match
  - "TUI ReiseCenter Neuss" - no match
  - "TUI ReiseCenter" - match!  ("Neuss Innenstadt" is suggested as the `branch`)

This approach fails in a few interesting ways, but these are things I think we can handle in the name-suggestion-index with either separate entries for store formats or matchNames for alternate names.  I'm really interested in testing more of these.
- "Gap Kids", "Gap Factory Outlet" - (in these cases "Kids" and "Factory Outlet" are not branches).
- "KFC / Taco Bell" (It thinks "Taco Bell" is the branch, need to fix)

bhousel:
> Other software has presets and don't have to declare them as the source in changeset comments either. There isn't anything special about the NSI that it should have to compared to any other software that does similar things.

Yes this is a good point - NSI is really a preset pack + a separate matcher for alternate names.  I don't think of it as a "source" in the OSM sense.

bhousel:
I've been doing a bunch of testing with this the past few days, and I think it's about as good as it will get..
NSI v5 has been published, and this branch now pulls the latest compatible release of that.
I'd love for more people to test here and let me know if you notice any issues!  - https://ideditor.org/nsi-v5

lkernan:
https://ideditor.org/nsi-v5 seems to have stopped keeping up with changes to NSI somewhere around the 5.0.20210326 release commit.

UKChris-osm:
> https://ideditor.org/nsi-v5 seems to have stopped keeping up with changes to NSI somewhere around the 5.0.20210326 release commit.

I think @bhousel has changed this PR to use the "_released_" version of the NSI, rather than the pre-release that is generated when "_npm run dist_" is run.

bhousel:
> https://ideditor.org/nsi-v5 seems to have stopped keeping up with changes to NSI somewhere around the 5.0.20210326 release commit.

Yes!  20210326 is the latest published release. That's 2 days ago..

Screen Shot 2021-03-28 at 7 51 18 AM

And that is the version of the `nsi.min.json` that iD fetches from the CDN:
Screen Shot 2021-03-28 at 7 51 24 AM

So that's actually pretty awesome that iD can do this now.  It won't pull the latest code from the `main` branch, as that is often broken, but we can publish updates very frequently now.

lkernan:
> > https://ideditor.org/nsi-v5 seems to have stopped keeping up with changes to NSI somewhere around the 5.0.20210326 release commit.
> 
> I think @bhousel has changed this PR to use the "_released_" version of the NSI, rather than the pre-release that is generated when "_npm run dist_" is run.

Oops sorry, that’s what confused me.

UKChris-osm:
@lkernan A new release is now available to test which includes your updates to the NSI 👍

kymckay (commented):
I've gone through the code closely and it all looks good to me, plus the testing introduced for the new locations module is very reassuring. Strong use of promises and idle callbacks 👍

Only thing I'm curious about is in the use of `setTimeout` in the tests, I see the PhantomJS comment and assume the problem is some kind of race condition where the test "server" can terminate before those async tests ever resolve?

bhousel:
> Only thing I'm curious about is in the use of `setTimeout` in the tests, I see the PhantomJS comment and assume the problem is some kind of race condition where the test "server" can terminate before those async tests ever resolve?

It looks like I added those delays into the test code elsewhere in this commit: https://github.com/openstreetmap/iD/commit/c481f90e7df21bdb464fbfa34249b2b7fae7c1f7

"This also includes a bunch of tweaks to make the tests work
in both PhantomJS and modern browsers like Chrome.

Basically - introduce some more async into the test code so that
the coreData.get promise is guaranteed to settle.  Because in
PhantomJS the promise is polyfilled, and Chrome it's native, they
work slightly differently."

kymckay (approved):
Ah of course, I hadn't considered the async polyfill behaviour differing from the real thing - core-js really is a magical thing!

In that case, the code gets logical/semantic approval from me. Haven't played too much with the preview, but it definitely works and seems there are no lingering significant design concerns from those who have spent more time with it.

Just throwing some weight in towards getting this beast merged before it becomes a truly long-life branch. 😄

modules/core/context.js
```diff
@@ -557,7 +557,7 @@ export function coreContext() {
 
       Object.values(services).forEach(service => {
         if (service && typeof service.init === 'function') {
-          service.init();
+          service.init(context);
```
mbrzakovic: Is this needed? I can't seem to find service that requires context.

modules/core/locations.js
```diff
@@ -0,0 +1,267 @@
...
+  let _queue = [];
+  let _deferred = new Set();
+  let _inProcess;
+
+
+  // Returns a Promise to process the queue
+  function processQueue() {
+    if (!_queue.length) return Promise.resolve();
+
+    // console.log(`queue length ${_queue.length}`);
```
mbrzakovic, line 43: Few redundant debug comments.

modules/core/locations.js
```diff
@@ -0,0 +1,267 @@
...
+          _deferred.delete(handle);
+          // const t0 = performance.now();
+          chunk.forEach(resolveLocationSet);
+          // const t1 = performance.now();
+          // console.log('chunk processed in ' + (t1 - t0) + ' ms');
+          resolvePromise();
+        });
+        _deferred.add(handle);
+      })
+      .then(() => processQueue());
```
mbrzakovic, line 56: Nice trick 👌 (no-op)

modules/core/locations.js
```diff
@@ -0,0 +1,267 @@
...
+  // Performs the locationSet resolution, caches the result, and sets a `locationSetID` property on the object.
+  function resolveLocationSet(obj) {
+    if (obj.locationSetID) return;  // work was done already
+
+    try {
+      const locationSet = obj.locationSet;
+      if (!locationSet) {
+        throw new Error('object missing locationSet property');
+      }
+      const resolved = _loco.resolveLocationSet(locationSet);
```
mbrzakovic, line 72: I did some debugging, it seems that Presets_presets objects with locationSet that only has exclude (e.g.amenity/pub/irish in [preset_presets](https://github.com/openstreetmap/id-tagging-schema/blob/main/dist/presets.json)) will be treated as errors due to missing 'includes' and handled here as Q2. Is this expected?

modules/core/locations.js
```diff
@@ -0,0 +1,267 @@
...
+      const locationSetID = resolved.id;
+      obj.locationSetID = locationSetID;
+
+      if (!resolved.feature.geometry.coordinates.length || !resolved.feature.properties.area) {
+        throw new Error(`locationSet ${locationSetID} resolves to an empty feature.`);
+      }
+      if (!_resolvedFeatures[locationSetID]) {  // First time seeing this locationSet feature
```
mbrzakovic, line 79: I am curious:  _loco has internal caching, why is another _resolvedFeatures cache required?

modules/services/nsi.js
```diff
@@ -0,0 +1,668 @@
...
+  let foundSemi = false;
+  let testNameFragments = false;
+  let patterns;
+
+  // Patterns for matching OSM keys that might contain namelike values.
+  // These roughly correspond to the "trees" concept in name-suggestion-index,
+  let t = identifyTree(tags);
+  if (!t) return empty;
+
+  if (t === 'transit') {
```
mbrzakovic, line 272: This seems very similar to: https://github.com/osmlab/name-suggestion-index/blob/main/dist/trees.json
Why not just use that data?

modules/services/nsi.js
```diff
@@ -0,0 +1,668 @@
...
+
+
+// PRIVATE FUNCTIONS
+
+// `setNsiSources()`
+// Adds the sources to iD's filemap so we can start downloading data.
+//
+function setNsiSources() {
+  const sources = {
+    'nsi_data': 'https://cdn.jsdelivr.net/npm/name-suggestion-index@6.0/dist/nsi.min.json',
```
mbrzakovic: Just to double-check. The url (version part) should be changed manually on each package version bump?
If so, are there any suggestions for automatization of this part? Perhaps we could at least document these kind of updates. (no-op)

mbrzakovic (changes requested):
It was amazing to see presets, nsi and oci common locationSet functionality coming together.
This is surely a big improvement and sets a strong foundation to make very easy additions of locationSet specifics to any of these packages. Also, its great that several issues have been addressed.

Perhaps we could include this concept in Arhitecture documentation as it has impact across multiple files and its not that easy to digest.

I went through the code and added few minor comments and some curious questions.
Lets go through them and merge this change.

mbrzakovic:
I saw data/imagery.json being updated in this PR.
Can you please provide info on how are these imagery sources gathered?

modules/core/context.js
bhousel: probably not needed, I think all the services do not depend on context these days.

modules/core/locations.js
bhousel: They are great for debugging tho

modules/core/locations.js
bhousel: That is not expected, I think the json schema for these should require each locationSet to have an `include` property.  But I will check for a missing `include` and set it to `['Q2']` to mimic the old behavior of the presets (available worldwide).

modules/core/locations.js
bhousel: Good question, and I don't really remember why I wrote it this way.  
I think it might just have been defensive coding - this code makes a copy of the Object in `_resolvedFeatures` for whatever iD wants to do with it, instead of passing around references to the `_loco` internally cached Objects.

modules/services/nsi.js
bhousel: I don't really remember, but I think it was because:
-  in NSI it's easy to match what tree a feature belongs to, because the data is organized that way... 
-  in iD, it's a lot more complicated because it can only rely on a feature's tags, and those tags need to be matched in a certain order in some cases - primary before alternate + the exception for flag countries, and a feature can possibly match multiple trees.

modules/services/nsi.js
bhousel: I don't think we'd want to automate it.. The CDN will already serve the latest file that matches the requested [semver](https://semver.org/) number, so 6.0.xxxxxx.

I've been following semver on the NSI side, so if there is a major bump from 5 to 6, that means there is some incompatibility in the code that need to be addressed.  I always put stuff like this in the [changelog](https://github.com/osmlab/name-suggestion-index/blob/main/CHANGELOG.md).

Basically my thinking is:
- major bump - maybe some code needs to be rewritten, maybe data file format is different
- minor bump - code should still work, file format same maybe with new properties (NSI might get a new tree)
- patch bump - code is same, file format is same, some data is new. (NSI publishes a new patch about every week).

bhousel:
> I saw data/imagery.json being updated in this PR.
> Can you please provide info on how are these imagery sources gathered?

Yeah that's just ELI.  Anyone who works on iD needs to update it manually because ELI doesn't actually publish code with version numbers.  (so npm or package-lock can't work with it).  Everyone who installs iD is going to get whatever version of ELI happened to exist at the time - there will always be changes in `imagery.json`.  See #7425 or PM me for details.

modules/core/locations.js
mbrzakovic: ok :)

modules/core/locations.js
mbrzakovic: Got it. no-op for now as impact is low and priority is pushing this change through

modules/services/nsi.js
mbrzakovic: Got it. I agree.
Ok then I have only this small observation: When new major/minor packages version becomes available, maintainer would have to bump version in both package.json and url links. It would be more secure to have one singular place with version and for that info to propagate automatically wherever its needed.

bhousel: Oh yeah! I didn't understand before what you meant by automatization originally but I do now.  I think I can make a small improvement here so that this code constructs the URL to have the appropriate version number that matches whatever `package.json` has.

bhousel: I did this, check out 8db1c1f7f

bhousel: Can use this trick for a handful of other data files too, like osm-community-index, or the presets

tas50:
@bhousel is there still a URL floating around with this branch?

bhousel:
> @bhousel is there still a URL floating around with this branch?

Yes I've kept https://ideditor.org/nsi-v5 updated to pull latest released NSI

mbrzakovic (approved):

lgommans:
Why did this close #6408? It changed nothing in terms of seeing what you're changing, iD still invites people to blindly apply changes under their name.

PR #8309, opened 2021-01-21
Swap out w8r/martinez clipper for mfogel/polygon-clipping

bhousel:
This is a drop in replacement that performs better and yields more correct results
see also https://github.com/ideditor/location-conflation/pull/20

I tested this locally by loading up some custom vector tile layers to make sure that cross-tile polygon unioning is still working.. These vector tile layers really stress test it!

**Mapbox Streets**
Screen Shot 2021-01-21 at 11 22 29 AM

**Mapbox Terrain**
Screen Shot 2021-01-21 at 11 24 07 AM


bhousel:
### Before
Screen Shot 2021-01-21 at 11 43 06 AM


### After

Screen Shot 2021-01-21 at 11 43 34 AM

quincylvania:
@bhousel Thanks! Looks like a great upgrade ⬆️

PR #8310, opened 2021-01-21
Fix a few errors in note processing

bhousel:
- Notes still load over the xml api, so when the user is not authenticated,
  we can't use `d3_json` to fetch them (instead use `d3_xml`)
- Make sure to delete the deferred callback handles once the callbacks are
  processed (probably unnecessary, but good housekeeping)
- Remove a stray `if (!json.elements)` block from the `parseUserJSON`
  function. This was preventing the user profile from parsing, as the
  user's JSON has no such property (was probably copied from `parseJSON`)


quincylvania:
@bhousel Thanks!

PR #8319, opened 2021-01-25
Promisify validation

bhousel:
This would follow #8305 

Rewrite the validator in ES6/Promises, several improvements here:
- implements a validation work queue, jobs are run during browser idle callbacks
- when merging base entities, don't run validations 2x on both base and head graphs (this was wasteful)
- keep track of resolved issues in a separate set (it's not a simple compare of base/head anymore) 
This happens after validation queue is empty and avoids race conditions and inaccurate resolved counts

This isn't quite ready yet, but will be soon.  My plan is to make each of the validators either 
- return promises themselves that resolve whenever they are ready to do their thing  or
- have some status property so that the validator can just keep them queued until they are ready to do their thing (probably easier).

The issue I'm trying to solve here I mentioned on #8305 - some validators fetch data from elsewhere and can't properly validate until they have everything they need.

### DIFF
--- a/modules/core/validator.js
+++ b/modules/core/validator.js
@@ -21,14 +21,17 @@ export function coreValidator(context) {
   let _headCache = validationCache();  // issues after all user edits
   let _previousGraph = null;
 
-  let _deferred = new Set();   // Set( IdleCallback handles )
-  let _inProcess;              // Promise fulfilled when validation complete
+  let _deferredRIC = new Set();   // Set( RequestIdleCallback handles )
+  let _deferredST = new Set();    // Set( SetTimeout handles )
+  let _inProcess;                 // Promise fulfilled when validation complete
+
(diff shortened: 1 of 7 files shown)


bhousel:
> * have some status property so that the validator can just keep them queued until they are ready to do their thing (probably easier).

I did this 👍   This PR is good now.
Now validators can return a "provisional" result.  This is useful for situations like the `outdated_tags` validator, which may take a few seconds to be ready as it fetches the latest name-suggestion-index data at startup.
 
When a validation rule returns a provisional result, the validation engine will put that entity back into the validation queue after a short delay.

bhousel:
I'm happy with how this side quest turned out, so I'm going to merge it back to the nsi_v5 branch and continue from there.

The next logical step is to just use web workers for the stuff we're using `requestIdleCallback` for. I found a decent library for this, so I might do that in the next few days.

PR #8322, opened 2021-01-26
Fix wms detection to determine axis ordering

rbuffat:
The current WMS 1.3.0 detection only works for upper case parameters, but `version` and `crs` can also be lower-case.


quincylvania:
@rbuffat Thanks for the fix!

PR #8367, opened 2021-02-25
Use Mapillary vector tiles for images, sequences and map features

nickplesha:
Hello, this PR updates the Mapillary integration and now it uses our vector tiles as the data source for images, sequences and map features (traffic signs and points). 

With this change the loading of Mapillary data should be much faster (you should observe an improvement in loading speed of 1000%+ in very densely mapped areas).

### Changes

- Switched to using vector tiles instead of the Mapillary API
- Updated the data and cache structure to match new data source
- Fixed a bug with clicking on a Mapillary image marker
- Updated how Mapillay JS is initialized (version 3.0.0 introduced some breaking changes)
- Removed user filter for Mapillary images (vector tiles don't have usernames right now)
- Cleaned up the code and used a bit more modern JS where possible

You can test a preview of the changes here: https://id-mapillary-tiles.netlify.app/

Let me know if you have any question or suggestions.
Thanks!


nickplesha:
BTW, this also reduces the amount of data transferred by up to 50% while still showing more Mapillary data than before which will provide a better experience for users with slower internet connections.

1ec5:
This sounds fantastic! Comparing [this branch](https://id-mapillary-tiles.netlify.app/#map=20.00/37.33680/-121.88943&photo=mapillary/Yj7K_lI4TV57gHNXUoED3w&photo_overlay=mapillary) to [the development branch](https://ideditor.netlify.app/#map=20.00/37.33680/-121.88943&photo=mapillary/9EpB2pX6eYFKveLkwfLBvQ&photo_overlay=mapillary), the new implementation does load significantly faster, but I’m seeing far fewer sequences connecting the images, even though images belonging to the same sequence are highlighted together once one of the images is selected. Is that the intended behavior?

nickplesha:
@1ec5  Thanks for checking! Yep, I'll fix the sequence loading function - it seems it has been affected by the change in the cache structure, we should still be able to load all visible sequences.

nickplesha:
Fixed sequences - the problem was loading sequences that span multiple tiles so I adjusted how these are stored in the cache.

Also, refactored some old methods to further reduce the size of the code.

nickplesha:
Created a new PR: https://github.com/openstreetmap/iD/pull/8372 to fix merge conflicts and build errors.

PR #8372, opened 2021-03-01
Use Mapillary vector tiles and switch to v4

nickplesha:
**Edit**: This PR has now been updated to use the new (v4) Mapillary API. 

The old API will be shutting down soon so, in order to keep the Mapillary integration working, it would be great if this PR could be merged as soon as possible.

---

Hello, this PR updates the Mapillary integration and now it uses our vector tiles as the data source for images, sequences and map features (traffic signs and points).

With this change the loading of Mapillary data should be much faster (you should observe an improvement in loading speed of 1000%+ in very densely mapped areas).

### Changes
Switched to using vector tiles instead of the Mapillary API
Updated the data and cache structure to match new data source
Fixed a bug with clicking on a Mapillary image marker
Updated how Mapillay JS is initialized (version 3.0.0 introduced some breaking changes)
Removed user filter for Mapillary images (vector tiles don't have usernames right now)
Cleaned up the code and used a bit more modern JS where possible
You can test a preview of the changes here: https://id-mapillary-tiles.netlify.app/

Check the old PR for previous discussion: https://github.com/openstreetmap/iD/pull/8367 as this is a new branch with all the changes and bug fixes in a single commit, based on the latest develop

Let me know if you have any question or suggestions.
Thanks!


nickplesha:
@quincylvania This PR also prepares the Mapillary integration for the upcoming changes to our API. Here's a blog post with more details: https://blog.mapillary.com/update/2021/03/03/preparing-for-the-new-mapillary-api.html

The plan is to create another PR which would just update the URLs to the new API (when it's available) and this should hopefully make for a smooth transition.

tordans:
@nickplesha the speed improvements sound great!

However, I was really happy to finally see a username-filter added only recently https://github.com/openstreetmap/iD/issues/5307 and it would be a sad thing to see it go … 😿 

Looking at https://blog.mapillary.com/update/2021/03/03/preparing-for-the-new-mapillary-api.html it sounds like, that usernames will be filterable soon again, so you would be able to re-add them soon?
> Filtering will happen on the client side of the vector tiles (for example filtering image point data by a date range, a set of **usernames**, or spatial overlap with a polygon)

Or does this part reference the fact, that filtering need to be done in iD now, so it still could be there but needs to be reworked?

nickplesha:
@tordans Yes, the new vector tiles will have all the data necessary to enable filtering like we have in the existing API (v3). I've created this PR using our current vector tiles and will update to use the latest tiles as soon as they are available.

nickplesha:
Updated PR with the switch to the v4 Mapillary API and vector tiles.

modules/svg/mapillary_images.js
```diff
@@ -6,13 +6,12 @@ import { services } from '../services';
...
 export function svgMapillaryImages(projection, context, dispatch) {
-    var throttledRedraw = _throttle(function () { dispatch.call('change'); }, 1000);
-    var minZoom = 12;
-    var minMarkerZoom = 16;
-    var minViewfieldZoom = 18;
-    var layer = d3_select(null);
-    var _mapillary;
-    var viewerCompassAngle;
+    const throttledRedraw = _throttle(function () { dispatch.call('change'); }, 1000);
+    const minZoom = 12;
```
mbrzakovic, line 10: Curious on why is the minZoom set to 12 for mapillary svg's.
I don't think that users actually view these features on such high lod's.
For urban areas, the data to load/draw is huge and it makes a screen very unresponsive.
Perhaps we should reconsider to at least 14?

modules/services/mapillary.js
```diff
@@ -306,101 +242,97 @@ export default {
...
         if (_mlyCache) {
-            Object.values(_mlyCache.images.inflight).forEach(abortRequest);
-            Object.values(_mlyCache.image_detections.inflight).forEach(abortRequest);
-            Object.values(_mlyCache.map_features.inflight).forEach(abortRequest);
-            Object.values(_mlyCache.points.inflight).forEach(abortRequest);
-            Object.values(_mlyCache.sequences.inflight).forEach(abortRequest);
+            Object.values(_mlyCache.requests.inflight).forEach(function(request) { request.abort(); });
         }
 
         _mlyCache = {
```
mbrzakovic, line 251: I was looking into memory consumption while browsing around with mapillary features turned on, and observed that allocated memory is never cleaned up...
Perhaps we can consider the following:
- Clear cache periodically [time based, or check on each tileLoad if num of loaded tiles are above some threshold). 
 We can discuss which policy would be the best for choosing the tiles to delete.. I can give an example on top of my mind: Clear tiles that are most distant from current projection view.
- Clear cache when 'Photo Overlays' check boxes are toggled off. (or throttle/delay that operation for clicky users)

modules/services/mapillary.js
```diff
@@ -306,101 +242,97 @@ export default {
...
-        return true;
+        return lineStrings;
     },
 
 
+    // Load images in the visible area
     loadImages: function(projection) {
-        loadTiles('images', apibase + 'images?sort_by=key&', projection);
-        loadTiles('sequences', apibase + 'sequences?sort_by=key&', projection);
+        loadTiles('images', tileUrl, 14, projection);
```
mbrzakovic, line 314: Just curious if 14 provides the best perf 🤔

modules/services/mapillary.js
```diff
@@ -636,195 +576,138 @@ export default {
...
-
-        var selectedImageKey = _mlySelectedImageKey;
-        var selectedSequenceKey = selectedImageKey && this.getSequenceKeyForImageKey(selectedImageKey);
-        var selectedLineString = selectedSequenceKey && _mlyCache.sequences.lineString[selectedSequenceKey];
-        var selectedImageKeys = (selectedLineString && selectedLineString.properties.coordinateProperties.image_keys) || [];
-
-        // highlight sibling viewfields on either the selected or the hovered sequences
-        var highlightedImageKeys = utilArrayUnion(hoveredImageKeys, selectedImageKeys);
+    // Update the currently highlighted sequence and selected bubble.
+    setStyles: function(context, hovered) {
```
mbrzakovic, line 645: I've noticed a very small inconsistency here [not sure if it is a inconsistency or a feature 😁]
Repro: Toggle Mapillary imagery on, select one viewport (color to yellow), toggle imagery off.
The selected item stays selected and visible.
I saw that OpenStreetCam handles this by clearing all.

Actually I think it would be most convenient from ui perspective, to clear all and also clear the 'view photo' box.

mbrzakovic (approved):
First of all, great job on this @nickplesha!
The code is clean, and I found it easy to understand.
I had few curious comments which are non-blockers for this PR and if needed they can be addressed later on.

modules/services/mapillary.js
nickplesha: That was intentional, the current position of the image remains visible without the Mapillary imagery layer (it's a "separate" layer in mapillary_position.js). This is because you can turn on only traffic signs (for example) and then you can click on a traffic sign and see the position of the image looking at the sign even without turning on the imagery layer. This goal of this feature is to help the user figure out the actual position of the traffic sign based on the  detected sign, the image in the viewer and the image position marker on the map.

modules/services/mapillary.js
nickplesha: No, that's just the only zoom level where we have all image positions in our tiles. 😄

modules/services/mapillary.js
nickplesha: Suggestion one makes sense, we can set a threshold and clean up when it's reached. Ideally we would clean up the oldest loaded data first but we don't store a timestamp right now. I'll see if a simple solution can be found.

Suggestion two makes total sense and we can remove the cached data when the related layer is turned off.

modules/svg/mapillary_images.js
nickplesha: OK, I'll make it 14. I think it was 12 for a long time and I don't remember why.

nickplesha:
Fixed merge conflicts.

nickplesha:
@mbrzakovic Thanks for reviewing the code!

I was wondering if there are plans make a new release of iD soon? I'm asking as the currently deployed version still uses Mapillary API v3 which we plan to shut down on August 18th.

mbrzakovic:
> @mbrzakovic Thanks for reviewing the code!
> 
> I was wondering if there are plans make a new release of iD soon? I'm asking as the currently deployed version still uses Mapillary API v3 which we plan to shut down on August 18th.

Yes, I was going to ask you what is the eol date for v3, thanks for the info!
I can't say anything official yet but you can expect the answer/announcement in the next week or two.
Its good to know that we have time on this and hopefully we won't get too close to the deadline.

modules/svg/mapillary_images.js
bhousel: We decided on 12 years ago so that users could see at low zoom where the Mapillary coverage exists. (it only renders linestrings at that zoom).

modules/services/mapillary.js
bhousel: The caches get cleared when the `service.reset()` is called - this happens when the user saves their edits.  (all services work like this).

mbrzakovic:
Hey @nickplesha,

I was doing some local testing and found that translation for some mapillary map features are missing. 
Can you repro the attached img?
If so, I think its due to Mapillary v4 using new set of feature objects and iD/data/core.yaml[mapillary_map_features] section is out-of-date.
Where can the list of feature object names be found? Curious on how often does Mapillary update this list?

image

modules/svg/mapillary_images.js
mbrzakovic: Ok. No-op on this.

nickplesha:
@mbrzakovic I think it was from the map feature titles, but we don't use them at all so I've removed them and I think this should fix those errors.

mbrzakovic:
> @mbrzakovic I think it was from the map feature titles, but we don't use them at all so I've removed them and I think this should fix those errors.

@nickplesha yes this fixes the error, unfortunately the change also disables the tooltips (hover over the mapillary features).
Can you please look closer into the code to see if we could have the solution that would eliminate the error and keep the tooltip.
From the quick look it seems to me that we would require full list of possible feature object types labels (e.g.: object--sign--advertisement) so that we could include them into core.yaml [translation files] and enable tooltips for multiple languages.

nickplesha:
@mbrzakovic OK, so that's what's it for, I haven't noticed the tooltips before. :) I brought them back now, and I've updated core.yaml to include all of our map feature classes and their names.

mbrzakovic:
> @mbrzakovic OK, so that's what's it for, I haven't noticed the tooltips before. :) I brought them back now, and I've updated core.yaml to include all of our map feature classes and their names.

@nickplesha Yes, thanks.
Sorry but when I go to urban areas with console on I am still seeing few more missing:
mapillary_map_features.construction.flat.driveway
mapillary_map_features.construction.barrier.temporary
mapillary_map_features.object.traffic_sign.direction_front
mapillary_map_features.object.traffic_sign.front
mapillary_map_features.object.traffic_sign.back
mapillary_map_features.object.sign.other

As mentioned above, I am curious on how often does Mapillary update this list?

nickplesha:
@mbrzakovic Thanks, I've added those classes. The last time we updated the list was two years ago and I don't anticipate significant changes in the future.

PR #8398, opened 2021-03-12
Pass d3_event to click handler

bhousel:
Small bug where d3_event needs to be passed to the click handler on the feature_list if the user has pressed return to accept the first item from the list.


quincylvania:
@bhousel Thanks!

PR #8440, opened 2021-04-02
use addr: tags as a fallback name

### LINKED ISSUE #1524: Render address points and addresses

### LINKED ISSUE #1760: Display house numbers on residential buildings
When mapping a series of addresses, it would be really useful to display the house numbers so that you can easily keep track of how far you have got, without having to select the prior building to get the number, so that you don't end up with some off by one error.

### LINKED ISSUE #2001: Housenumbers are not displayed in edit-mode
so, to see where some are missing, 
you need to switch to view-mode, and back.

This is slow, and you have to remember 
the place you want to go next, and the surrounding numbers.

Is there a good reason Housenumbers are not shown in edit-mode ?
Otherwise I suggest to show them, at zoomlevel 17+

### LINKED ISSUE #3857: House number inside buildings
Hey,

in my area a lot of buildings are tagged with their address as part of the building. But I always have to click on the building to see if there is an address added. Like in this example it's Schrödingers houses, some have an adress others not:

Example 1

Is it possible to modify the inside of the shape of buildings to see directly the house number or if there is an adress already added. Maybe an example how it could look like:

Example 2

Thanks!

### LINKED ISSUE #7327: Show house number of buildings
It would be helpful to see the house number rendered on buildings. Now users have to click/highlight each building to check if it has address or not. It makes hard to do an address survey in iD.

I cannot see any open issue or pull request in this topic. There was a promissing PR #3840 several years ago, but sadly it wasn't merged. #1524 is similar but for address points.

### LINKED ISSUE #8744: Show addr:housenumbers
Got an idea!:
Instead of just [empty white markers that all look the same](https://www.openstreetmap.org/edit#map=20/24.23851/120.83489),
20211010T144415
how about also showing their addr:housenumber next to each one?!

k-yle:
Closes #8744, Closes #7327, Closes #3857, Closes #2001, Closes #1760, Closes #1524

When adding or editing addresses, the changeset upload page is not very helpful since all addresses are just called "Address". This makes it hard to review what changes you have made.

This PR makes use of the [`addr:`](https://wiki.osm.org/wiki/Key:addr) tags to provide a fallback name for a node/way/relation, if there is no [`name`](https://wiki.osm.org/wiki/Key:name) or [`ref`](https://wiki.osm.org/wiki/Key:ref) tag.

<table>
<tr>
	<td>before
	<td>after
<tr>
	<td>
	<td>
</table>

### DIFF
--- a/modules/util/util.js
+++ b/modules/util/util.js
@@ -185,13 +185,15 @@ export function utilGetAllNodes(ids, graph) {
 /**
  * @param {boolean} hideNetwork If true, the `network` tag will not be used in the name to prevent
  *                              it being shown twice (see PR #8707#discussion_r712658175)
+ *
+ * @param {boolean} isMapLabel If true, this name is for a label on the map. If falsy, it's for a
+ *                             label elsewhere in the UI.
  */
-export function utilDisplayName(entity, hideNetwork) {
+export function utilDisplayName(entity, hideNetwork, isMapLabel) {
     var localizedNameKey = 'name:' + localizer.languageCode().toLowerCase();
     var name = entity.tags[localizedNameKey] || entity.tags.name || '';
 
     var tags = {
-        addr: entity.tags['addr:housenumber'] || entity.tags['addr:housename'],
         direction: entity.tags.direction,
         from: entity.tags.from,
         name,
@@ -210,10 +212,6 @@ export function utilDisplayName(entity, hideNetwork) {
     // Non-routes tend to be labeled in many places besides the relation lists, such as the map, where brevity is important.
     if (!entity.tags.route && name) {
         return name;
-    }
-    // unnamed buildings or address nodes: show housenumber/housename
-    if (tags.addr) {
-        return tags.addr;
     }
 
     var keyComponents = [];
@@ -246,6 +244,7 @@ export function utilDisplayName(entity, hideNetwork) {
     }
 
     const alternativeNameKeys = [
+        'addr:housename',
         'alt_name',
         'official_name',
         'loc_name',
@@ -268,13 +267,36 @@ export function utilDisplayName(entity, hideNetwork) {
         }
     }
 
+    // as a last resort, use the street address as a name.
+    const unit = entity.tags['addr:unit'];
+    const housenumber = entity.tags['addr:housenumber'];
+    const streetOrPlace = entity.tags['addr:street'] || entity.tags['addr:place'];
+
+    if (!isMapLabel && unit && housenumber && streetOrPlace) {
+        return t('inspector.display_name_addr_with_unit', {
+            unit,
+            housenumber,
+            streetOrPlace,
+        });
+    }
+
+    if (!isMapLabel && housenumber && streetOrPlace) {
+        return t('inspector.display_name_addr', {
+            housenumber,
+            streetOrPlace,
+        });
+    }
+
+    // the housenumber can always be used, regardless of isMapLabel
+    if (housenumber) return housenumber;
+
     // no match found
     return '';
 }
 
 
 export function utilDisplayNameForPath(entity) {
-    var name = utilDisplayName(entity);
+    var name = utilDisplayName(entity, undefined, true);
     var isFirefox = utilDetect().browser.toLowerCase().indexOf('firefox') > -1;
     var isNewChromium = Number(utilDetect().version.split('.')[0]) >= 96.0;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # unreleased (v2.36.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* use `addr:` tags to label features with no name ([#8440], thanks [@k-yle])
 * The flip operation now works on nodes with no `direction` tag, to support quickly adding `direction` to features like traffic signs ([#9317], thanks [@k-yle])
 * Show "add new key" placeholder text for blank row in raw tag editor ([#11211], thanks [@bhavyaKhatri2703])
 * Consider other name-like tags for labelling features, such as `lock_name` ([#9588], thanks [@k-yle])
@@ -62,6 +63,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :mortar_board: Walkthrough / Help
 #### :hammer: Development
 
+[#8440]: https://github.com/openstreetmap/iD/pull/8440
 [#9317]: https://github.com/openstreetmap/iD/issues/9317
 [#9511]: https://github.com/openstreetmap/iD/pull/9511
 [#9588]: https://github.com/openstreetmap/iD/pull/9588
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -766,6 +766,8 @@ en:
     add_fields: "Add field:"
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
+    display_name_addr: "{housenumber} {streetOrPlace}"
+    display_name_addr_with_unit: "{unit}, {housenumber} {streetOrPlace}"
     display_name:
       direction: "{direction}"
       network: "{network}"
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -299,7 +299,7 @@ export function svgLabels(projection, context) {
             var preset = geometry === 'area' && presetManager.match(entity, graph);
             var icon = preset && !shouldSkipIcon(preset) && preset.icon;
 
-            if (!icon && !utilDisplayName(entity)) continue;
+            if (!icon && !utilDisplayName(entity, undefined, true)) continue;
 
             for (k = 0; k < labelStack.length; k++) {
                 var matchGeom = labelStack[k][0];
@@ -328,8 +328,9 @@ export function svgLabels(projection, context) {
                 entity = labelable[k][i];
                 geometry = entity.geometry(graph);
 
-                var getName = (geometry === 'line') ? utilDisplayNameForPath : utilDisplayName;
-                var name = getName(entity);
+                let name = geometry === 'line'
+                    ? utilDisplayNameForPath(entity)
+                    : utilDisplayName(entity, undefined, true);
                 var width = name && textWidth(name, fontSize, selection.select('g.layer-osm.labels').node());
                 var p = null;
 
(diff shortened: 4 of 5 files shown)


modules/util/util.js
```diff
@@ -218,6 +218,16 @@ export function utilDisplayName(entity) {
...
     }
 
+    // if there's still no name found, try addr:housename
+    if (!name && entity.tags['addr:housename']) {
+        name = entity.tags['addr:housename'];
+    }
+
+    // as a last resort, use the street address as a name
+    if (!name && entity.tags['addr:housenumber'] && entity.tags['addr:street']) {
+        name = entity.tags['addr:housenumber'] + ' ' + entity.tags['addr:street'];
```
1ec5: This format may not be appropriate in every locale. For example, in China in Chinese, an address goes from biggest to smallest, so the street would come before the house number. In Vietnam in Vietnamese, the word “Số” would ideally precede the house number. Something like this would at least make the format localizable by language:

```suggestion
        name = t('inspector.display_name_addr', {
            housenumber: entity.tags['addr:housenumber'],
            street: entity.tags['addr:street']
        });
```

Then core.yaml would set `inspector.display_name_addr` to `{housenumber} {street}` to match the expected English format.

Unfortunately, `t()` on its own can’t resolve any country-specific differences. Perhaps this functionality could be added to address_formats.json.

1ec5: Another simpler option would be to return just `addr:housenumber`, which would be short enough to label the points on the map canvas, though it could get ambiguous in the sidebar.

k-yle: @1ec5 good point, I've made that change.

Regarding your second comment - In the future, perhaps `utilDisplayName()` could be split into `utilDisplayShortName()` and  `utilDisplayLongName()`, then the short name could be used on the map (just `addr:housenumber`) and the long name used everywhere else. But I'll leave that for a separate PR

modules/util/util.js
```diff
@@ -218,6 +218,19 @@ export function utilDisplayName(entity) {
         name = t('inspector.display_name.' + keyComponents.join('_'), tags);
     }
 
+    // if there's still no name found, try addr:housename
+    if (!name && entity.tags['addr:housename']) {
+        name = entity.tags['addr:housename'];
+    }
+
+    // as a last resort, use the street address as a name
+    if (!name && entity.tags['addr:housenumber'] && entity.tags['addr:street']) {
```
1ec5: `addr:unit` may also be necessary to differentiate address points.

pnorman: It seems to me that you'd want to display the housenumber if there was no name and addr:street tag

pnorman (changes requested):

modules/util/util.js
k-yle: @1ec5 I just added `addr:unit`, but I also just read https://github.com/gravitystorm/openstreetmap-carto/issues/3568... In New Zealand it is `{unit}/{housenumber}` or `{unit}, {housenumber}`

@pnorman I didn't want to cause confusion by having a random number showing on the map if there is `addr:housenumber` but no `addr:street`. But if you think that's a good idea, I'll make that change. **EDIT: done**

1ec5: > In New Zealand it is {unit}/{housenumber} or {unit}, {housenumber}

Yeah, ultimately it’ll be more effective to read the existing address_formats.json, which defines the order of the fields in the address field after geocoding, and join them with spaces (or commas like Nominatim does). That would require a bit of refactoring, though, so it probably could be a separate task. For now, I suppose we could leave it up to translators to come up with a reasonable string that may or may not match the local addressing convention but would still be recognizable.

mbrzakovic:
I very much like the idea of using addr:tag values as displayName fallbacks.
There are 2 things I noted while looking into code:

1. I don't think we should use translations as a solution to address formatting. Correct me if I am wrong but my understanding was that translations depend on users locale, while the address display format(order) should be dependent on entity location (as done in uiFieldAddress). These two are orthogonal concepts, meaning user can prefer China locale and still edit an entity in US. 

- I think the idea, that @1ec5 suggested above, using address_formats.json + locale lookup based on entity location is a valid solution and does not take much additional work. I've setup up small [branch-change that demonstrates this.](https://github.com/openstreetmap/iD/commit/9d5d3a7e6a15e1fa387641c7eab306a96f14abf2)

2. utilDisplayName has a wide usage, so we should be more carefull with this change.
	e.g. The current code would also enable map label-rendering for nodes that have icons and have 'addr' tags which I am not sure we want...
image
Actually I see the opertunity for this PR to have a bigger impact by covering wider types of entities (not just address nodes) and wider use-cases (not just the original images in PR's initial comment)
e.g. Building/Multiselect - Features dropdown (this looks better then having just 3 x Apartment Building)
image

I am interested in your opinion on this.
Bellow is the full table of displayName usage, please look it over/populate in which situations do you think addr:fallback should be used. @pnorman @k-yle @1ec5 
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/MBRZAK~1/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/MBRZAK~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72" lang=en-US>



Item | Usage | Description | Alternative -   Full context | Addr fallback?
-- | -- | -- | -- | --
svgLabels | displayName,   displayNameForPath | Condition   and rendered text for map labels display | None.   No displayName => No rendered label | 
ui\uiSectionSelectionList | displayName | Sidebar   - Multiple select - items display | 'Featuretype' 'displayName' |  
ui\uiFeatureList | displayName | Sidebar   - Search (none selected) | 'Featuretype' 'displayName' | 
ui\fields\raw_member_editor | displayName | Inside   a relation when editting Members | 'Featuretype' 'displayName' |  
ui\fields\raw_membership_editor | displayName | Relations   (from editting standard node\way) | Relation   'displayName' |  
ui\fields\restrictions | displayName | Click   on Node (Traffic Signals), Turn Restriction, hint on hover | Yes.   displayName. If none then featureType |  
ui\sections\uiSectionChanges | displayName | After   clicking Save in sidebar - Listing name of changes | Created   'FeatureType': 'displayName' |  
keepRight,osmose,improveOSM | displayName | Click   on issue, inner text can contain displayName | Yes.   displayName. If none then featureType | 
uploader | displayName | Conflict   detection - Displaying conflicts | Yes.   displayName. If none then displayType + id |  
hash | displayLabel | Browser   tab name when entity is selected | Yes.   displayName. If none then featureType |  
paste | displayLabel | Paste   annotation | Yes.   displayName. If none then featureType |  
validations\* | displayLabel | Displaying   issues on Save menu | Yes.   displayName. If none then featureType |  



</body>

</html>

1ec5:
> Actually I see the opertunity for this PR to have a bigger impact by covering wider types of entities (not just address nodes) and wider use-cases (not just the original images in PR's initial comment)

That’s a good point. This non-name-based labeling parallels what we’ve already done for route relations, which are labeled in various places based on secondary tags like `from`. By labeling more tags, we would make mappers less likely to put redundant information in `name`.

> Bellow is the full table of displayName usage, please look it over/populate in which situations do you think addr:fallback should be used.

These situations all seem reasonable or at least harmless. (It seems unlikely that an address would show up in the turn restriction editor, for example, but it’s probably better to let that happen then complicate the code with a special case.) It is worth noting that some areas have a _lot_ of buildings with addresses but no names, or a lot of address nodes without names. In these areas, the user will see many more labels than before, similar to the effect in openstreetmap-carto. I think that change would be reasonable, but it’ll likely be a significant visual change in some places.

The screenshot above shows some “undefined undefined” labels on buildings on the map. Is that because the address format loads too late, or something else?

mbrzakovic:
@1ec5  Thanks for the comment.

> The screenshot above shows some “undefined undefined” labels on buildings on the map. Is that because the address format loads too late, or something else?


Great vision :) Please disregard it, I did a quick local change to demo the 'multiple features'.

mbrzakovic:
link - https://github.com/openstreetmap/iD/issues/1524

tyrasd (commented):
Unfortunately, this clashes a little bit with #10970:

The rendered housenumber-shields are only really designed to fit the housenumber part of the address (housenames being already a bit of a stretch). Also, for labels on buildings with addresses, the label would only render when zoomed in quite far, which is also less useful when mapping addresses.

Image

Like in the above example, even if the shield width were correctly calculated, it's not really necessary to repeat the initial of the road name n times for every address. 🤷 

Maybe it would make sense to have a flag in `utilDisplayName` to toggle the output to use a short variant (i.e. only housenumber/housename) for map labels if required?

Some additional edge cases to consider:
* render `addr:housename` (with or without `addr:street`?), maybe only if `addr:housenumber` is not present
* render `addr:place` when there is no `addr:street` (examples: [1](https://www.openstreetmap.org/way/197134299), [2](https://www.openstreetmap.org/way/125087193), also common in eastern Europe AFAIK)
* what about `addr:block_number` (common in Japan) and `addr:block` (e.g. India)?

k-yle:
Updated, now these changes do not apply to short labels in the map-renderer, only to long labels in other parts of the UI:

image

&nbsp;

* `addr:place` added
* `addr:block_number` and `addr:block` - it seems like these are often numeric (or start/end with numbers), so it might be confusing to use them in the label?

tyrasd (approved):

maro-21:
I didn't want to add a new issue, because it's just a missing space...
So... when trying to save my changeset [similar to the screenshot "after" in the first post] there is no space between `addr:housenumber` and `addr:street`, so I can see something like
 `Modified Address: 55Main Street`.

pnorman:
> I didn't want to add a new issue, because it's just a missing space...

It's a new bug, so it's best to open a new issue. That way it can be tracked and closed when its fixed.

k-yle:
@maro-21 what language are you using? that text is localised, so it might be an issue with the translation

maro-21:
> @maro-21 what language are you using? that text is localised, so it might be an issue with the translation

Polish. Is there no bug in other languages?

k-yle:
@maro-21 yeah it's a problem with the polish translations. You can [search transifex for `key:display_name_addr`](https://app.transifex.com/openstreetmap/id-editor/translate/#pl/core?q=key%3Adisplay_name_addr) to find the buggy string

Issue #8472, opened 2021-04-27
CrossEditableZoom - Side panel hover-select stays


mbrzakovic:
Sidepanel entity information stays if cross-editable zoom level is crossed by only one scroll while hovering on the entity. This enables populating information of entity that is not selected

**Steps to reproduce:**
1) Zoom level 16 hover over any entity
<img width="1279" alt="CrossEditZoomBug_p1" src="https://user-images.githubusercontent.com/78906108/116232287-38fe0980-a75a-11eb-9d69-2b4b83455cf5.png"> (mouse is hovering over item in center)
2) Zoom out using mouse scroll only once (minimally)
<img width="1277" alt="CrossEditZoomBug_p2" src="https://user-images.githubusercontent.com/78906108/116232479-78c4f100-a75a-11eb-9c4d-6c7e61a11b25.png">
3) Feel free to move mouse and go to sidepanel and populate entity
<img width="1277" alt="CrossEditZoomBug_p3" src="https://user-images.githubusercontent.com/78906108/116232593-a316ae80-a75a-11eb-9b77-4e34286399f9.png">

PR #8473, opened 2021-04-27
CrossEditableZoom event firing fix - Use fresh zoom level value to determine if event should be fired

mbrzakovic:
(solves #8472)

Sidepanel behavior is dependent on crossEditableZoom event which has small issue: 
- Map uses old zoom level value in zoomPan method to condition firing the abovementioned event.

Fix is simple: First transform the projection and then look into firing the event.

### DIFF
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -589,15 +589,6 @@ export function rendererMap(context) {
             return;  // no change
         }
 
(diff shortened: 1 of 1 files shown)


mbrzakovic:
This pr should also resolve small inconsistency with selected element animation (breathe) around cross Editable zoom.

PR #8523, opened 2021-06-01
Declare missed but used dependencies in package.json

artembert:
I found out that some of the dependencies aren't declared in the package.json. I used [depcheck ](https://www.npmjs.com/package/depcheck) to get the list of missed dependencies and added them to the package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "node-diff3": "2.1.0",
     "osm-auth": "1.1.0",
     "pannellum": "2.5.6",
+    "pbf": "^3.2.1",
     "polygon-clipping": "~0.15.1",
     "rbush": "3.0.1",
     "whatwg-fetch": "^3.4.1",


artembert:
@mbrzakovic, do you think the changes are meaningful?

mbrzakovic:
@artembert  thank for contributing! apologies for delayed response.
d3 is kind of special case as a d3 package only re-exports stable versions of d3-* subpackages. So we actually don't need to specifically add each d3-* subpackage as a dependency (and I also wouldn't want dependabot to ping on these).

pbf is a good catch, please do leave that line of code and we can close this PR.

artembert:
@mbrzakovic, thank you for your feedback! I updated my PR

artembert:
@mbrzakovic, could we merge this PR?

jguddas:
There are two instances of `import Protobuf from 'pbf';`, are you sure that pbf can be removed?

https://github.com/openstreetmap/iD/search?q=pbf

artembert:
> There are two instances of `import Protobuf from 'pbf';`, are you sure that pbf can be removed?
> 
> https://github.com/openstreetmap/iD/search?q=pbf

We will not delete it but declare it in "dependencies"

Issue #3604, opened 2016-11-21
Relation member navigation


slhh:
Exploring an existing relation is currently a nightmare using iD, especially if members can't be identified by the name shown in the members section. 
The current workflow is:

1. scroll down to members section, 
2. jump to member,
3. view data,
4. scroll down to membership section,
5. find the relation again,
6. jump to relation,
7. scroll down to members section,
8. count members to fine the next one
9. repeat from 2. for each member.

In order to simplyfy this, the vertex navigation can be generalized as a parent child navigation covering relation/member similar to way/vertex.
The shortcuts can be reused and a graphical equivalent like it is proposed in #3603 can be added to the membership editor.
This would reduce the workflow shown above to a single keystroke per member. Users need to be able to explore the existing data easily in order to understand OSM datastructures.

If acceptable, I might work on this issue myself laster.

bhousel:
I had considered this a little.  I think what you want to do is jump previous/next way along a relation, like a route relation.. But I'm not sure how it would really work in practice.  

It's easy to do previous/next vertex, because you usually see a bunch of vertices on the screen at once, so jumping between them is not confusing.

But with relations, child member ways could be really long.. The user wouldn't see very well where they jumped to.  Also a way can have many kinds of parent relations that don't follow similar semantics.. e.g. - you highlight a segment, and say "next"... Next what?  It could be the next segment in a route, or the vertex in a turn restriction, or the next segment in a multipolygon.  Or anything.  

I kind of have my doubts about how useful this would be.. Also I'd like to figure out how to style route membership first.

slhh:
> I think what you want to do is jump previous/next way along a relation, like a route relation.

This is one one appllication. Some other examples are:
-Visit members of a site to remove the name which has been added to the site.
- Explore a level of relation which are members of a toplevel relation. (site of sites, moultipolygons of multipogons, segmented routes, etc.)
- Check and tag the bus stations of a bus route

> But with relations, child member ways could be really long.. The user wouldn't see very well where they jumped to. 

The viewport should not be changed as long as the way starts in the viewport in order to let the user watch the start position. We can just let the user scroll manually to the other end, or maybe we can reuse the "\\" shortcut in order to center the viewport on the invisible end of the way.
Relations consisting of long members aren't the real focus for improvement anyway, mapping bus routes etc. in urban areas, where the highway is split every 50m due to turn lanes etc. is much more inefficient, and really needs improvement.

[...]

PR #8577, opened 2021-07-13
Select [selectChild] - differentiate parent/child selection

### LINKED ISSUE #8298: Internationalize the key binding for selecting a parent way
ddf868221a893c528454a510188dfcac48bce606 changed the key binding for selecting a parent way (#8264) from ⌘↑ to <kbd>|</kbd> and made it so that repeatedly pressing <kbd>|</kbd> would toggle between a vertex and its parent way. <kbd>|</kbd> makes some sense on an English QWERTY keyboard, where `|` is strongly associated with `{` and `}`, but it’s problematic for users of other languages and keyboard layouts. For example, among Mac keyboard layouts:

* Belarusian, Russian, Serbian, Ukrainian – ⇧⌥1 requires three fingers
* Croatian, Slovene – ⇧⌥Ž requires three fingers on two hands
* French (AZERTY) – ⇧⌥L requires three fingers on two hands
* German (QWERTZ), Finnish, Spanish – ⌥7 requires two hands and is difficult to touch-type
* Italian, Macedonian, Norwegian, Swedish – **no way to type `|`**
* Polish – ⌥9 requires two hands and is difficult to touch-type
* Romanian – ⇧⌥Ț requires three fingers on two hands

⌘↑ would’ve been more universal while remaining reasonably intuitive to English speakers. It could’ve formed part of a system of arrow shortcuts for navigating “up” and “down” among nodes, ways, and relations. It would be pretty straightforward to restore ⌘↑ as an alternative key binding for <kbd>|</kbd>, but it wouldn’t make sense to keep the toggling behavior with ⌘↑ – it would be necessary to introduce a ⌘↓ as well. Separate ⌘↑ and ⌘↓ shortcuts would be easier to describe in help than the current <kbd>|</kbd> toggle.

mbrzakovic:
Resolves: https://github.com/openstreetmap/iD/issues/8298


mbrzakovic:
This change can be more precise. 
1) selectChild is currently like 'undo-selection' from selectParent. I think its better if selectChild could be used stand-alone as well if 'undo-selection' is not possible. e.g.: (browsing-> selectWay -> selectChild).
2) In keyboard shortcuts ui menu, selectChild does not belong to 'With node selected' section. It should belong to new: "With node selected' section.
The next commit is addressing this.

mbrzakovic:
Keyboard shortcuts ui menu now. (change: 'help' section to bottom left, new 'with way selected' section on bottom right)
image

modules/modes/select.js
```diff
@@ -113,6 +113,31 @@ export function modeSelect(context, selectedIDs) {
...
 
+    // find the child nodes for selected ways
+    function childNodeIdsOfSelection(onlyCommon) {
+        var graph = context.graph();
+        var childs = [];
+
+        for (var i = 0; i < selectedIDs.length; i++) {
+            var entity = context.hasEntity(selectedIDs[i]);
+
+            if (!entity || !['area', 'line'].includes(entity.geometry(graph))){
```
1ec5, line 124: If the selection includes a node and a way, should this method return nothing, or should it return the node and the way’s vertices?

data/core.yaml
```diff
@@ -2301,6 +2301,9 @@ en:
         last: "Jump to last node"
         parent: "Select parent way"
         change_parent: "Switch parent way"
+      way_selected:
+        title: "With way selected"
+        child: "Select child node"
```
1ec5: I think this is a lot clearer. Maybe it should say “Select child nodes”, plural, to set the expectation that normally (always?) it would result in many nodes being selected.

1ec5 (approved):
Thanks for splitting apart this operation and making it easier to reach on the keyboard. 😃 This same set of shortcuts will be convenient for an eventual implementation of navigating between relations and their member ways via keyboard: #3604. I just have a question about an edge case that could reasonably be resolved either way.

modules/modes/select.js
mbrzakovic: Good question. we could go either way but I would prefer to return nothing in this case mainly for consistency reasons:
- We are explaining this shortcut through 'way(s) selected' section so It seems more straightforward having effect only if ways are selected
- selectParent only works if vertices are selected so it would make sense for its antonym to have similar type of constraints.

data/core.yaml
mbrzakovic: Let me provide more context and then I would like to hear your opinion @1ec5 
Method acts in following manner:
1) If its done after doing selectParent, it will return the node from which it started (what was so far toggle '|')
2) If no selectParent was done, then there are 2 cases:
- If single selection -> Will return all nodes
- If multiple selection -> Wil return only common nodes among selected candidates

It's hard to wrap this functionality into a user-friendly short name :)
I wanted to keep it simple and keep the 'antonym'-kind naming with 'Select parent way'.

However if we wanted to be more precise I would recommend:
' Select common child nodes'

mbrzakovic: Actually, I think your suggestion is perhaps the best from user perspective and here is why: Majority of use-cases and most likely first usage, of this feature will probably be case 2.single (abovementioned).  So, going with: 'Select child node**s**'

PR #8618, opened 2021-08-02
NSI related fixes

### LINKED ISSUE #8615: 2.20.0 tag validation issue
When adding Internet access tags to a hotel brand:

Before
image

After
image

### LINKED ISSUE #8617: NSI shouldn't match on `old_name` tag
Issue raised on Slack: `old_name` is not a name tag worth matching.

bhousel:
closes #8615 - preserve tag values that can be toplevel tags unless it's the toplevel tag for the matched item
closes #8617 - don't consider `old_name` as a name for matching

### DIFF
--- a/modules/services/nsi.js
+++ b/modules/services/nsi.js
@@ -349,6 +349,7 @@ function gatherNames(tags) {
   }
 
   function isNamelike(osmkey, which) {
+    if (osmkey === 'old_name') return false;
     return patterns[which].test(osmkey) && !notNames.test(osmkey);
   }
 }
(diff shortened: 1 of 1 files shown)


modules/services/nsi.js
```diff
@@ -349,6 +349,7 @@ function gatherNames(tags) {
   }
 
   function isNamelike(osmkey, which) {
+    if (osmkey === 'old_name') return false;
```
mbrzakovic, line 352: Perhaps a small detail but it would be cleaner to include this to 'notNames' exception list

mbrzakovic (approved):

modules/services/nsi.js
bhousel: I thought about that but `notNames` is already a pretty gnarly regex and I didn't want to make it worse.

PR #8623, opened 2021-08-04
Maxar Premium imagery template url

mbrzakovic:
Updated Maxar Premium Imagery template url
- Please note that the exact change is not visible in the code change itself due to encryption.
- Feel free to checkout the branch and test it out.


maraf24:
Are you sure this url was indeed for Maxar Premium? In my location it serves the same outdated tiles as Maxar Standard in JOSM.

1ec5:
See also #8852.

Issue #8326, opened 2021-01-27
"Connect the feature" repurposes stop node


bhousel:
reported in the RapiD repo here: https://github.com/facebookincubator/RapiD/issues/194
The "connect this way" validation should avoid repurposing an existing node with interesting tags.

> Situation:
> I had a place where there was a road that crossed a footpath and there was no node so RapiD warns that the two ways cross without a node - all good.
>
> The Issue:
>When I click on the "Connect the features" it took a stop node on the road that was close to the crossing and moved it and made it into a crossing (and stop) node.
>
>What I expected:
>I would think that we would not want to combine stop and crossing nodes when connecting ways since the stop is specific to the car navigable highway and should not be on the pedestrian highway because there is no stop on the pedestrian way. I think that a new node should have been created at the intersection with highway=crossing and then the stop node should have been left alone.

cc  @atiannicelli

jleedev:
Unfortunately, this was desirable behavior when connecting a `footway=crossing` way to a `highway=crossing` node.

![image](https://user-images.githubusercontent.com/23022/111473682-afdac800-8701-11eb-9d75-3c6dc9486bc0.png)

bhousel:
> Unfortunately, this was desirable behavior when connecting a `footway=crossing` way to a `highway=crossing` node.

good point - I can revisit it and maybe do something a bit better.

PR #8625, opened 2021-08-05
Crossing_ways fixing - specially handle crossing nodes

mbrzakovic:
[This change](https://github.com/openstreetmap/iD/commit/4c1d51348da76fc7afc3052ad1fafb10c12ade0b) introduced the current behavior of crossing_ways issue fixing which is following:
- Repurpose the existing nearby node if it has no interesting tags
- Create new midpoint othervise.

This was correctly done in response to raised [rapid issue](https://github.com/facebookincubator/RapiD/issues/194) duplicated as #8326 

However as @jleedev [cleverly noticed ](https://github.com/openstreetmap/iD/issues/8326#issuecomment-801073422) repurposing should be done for crossing nodes.
This PR simply covers this case as well.

After researching 'crossing' vertexes in [presets](https://github.com/openstreetmap/id-tagging-schema/blob/main/dist/presets.json) I've concluded that there are only few cases of these:
- Few types of ['Highway' crossings](https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcrossing) always defined with tag: highway = crossing
- Railway/Tram crossing combinations (Railway-Railway, Railway-Road, Railway-Path, Tram-Path, Tram-Road), which are defined with tag key 'railway' and values: 'railway_crossing', 'level_crossing', 'crossing', 'tram_crossing', 'tram_level_crossing'.

Both of these are handled in isCrossing method and finally used as a special case to crossing_ways issue node repurposing logic.

### DIFF
--- a/modules/osm/node.js
+++ b/modules/osm/node.js
@@ -142,8 +142,8 @@ Object.assign(osmNode.prototype, {
     },
 
     isCrossing: function(){
-        return this.tags.highway === 'crossing' || 
-               this.tags.railway && this.tags.railway.indexOf('crossing') !== -1;  
+        return this.tags.highway === 'crossing' ||
+               this.tags.railway && this.tags.railway.indexOf('crossing') !== -1;
     },
 
     isEndpoint: function(resolver) {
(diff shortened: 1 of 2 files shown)


bhousel (approved):
Looks good 👍

Issue #8612, opened 2021-08-01
iD performance issues since version 2.20.0


dafadllyn:
iD has become very laggy and resource-hungry since version 2.20.0 (using Firefox, Chrome, Edge and Safari on Mac and Windows).

Unfortunately, iD is not usable anymore on an older computer with a 2.4 GHz Intel Core i5-2435M processor, on which the previous version ran quite smoothly. On a newer computer with a 1.7 GHz Intel Core i5-8350U processor, iD uses about 50% CPU and is laggy but usable during about the first two minutes after launch, but then iD's CPU usage drops to about 10% and it becomes more responsive.

bryceco:
How much RAM does your computer have, and how much RAM does it report your browser using?

dafadllyn:
The older machine has 4 GB of RAM (of which 3.0 GB is in use), the newer 16 GB (of which 4.2 GB is in use). The browsers use about 700–800 MB loading [this map section](https://www.openstreetmap.org/edit#map=19/46.94801/7.44586) in iD.

bhousel:
> The older machine has 4 GB of RAM (of which 3.0 GB is in use), the newer 16 GB (of which 4.2 GB is in use). The browsers use about 700–800 MB loading [this map section](https://www.openstreetmap.org/edit#map=19/46.94801/7.44586) in iD.

Yes, that's Bern, which is pretty heavily micromapped.  The CPU usage is because iD is validating all those features as they stream in from the OSM API.

I'd like to move the validation code from a RenderIdleCallback to a [workerpool](https://www.npmjs.com/package/workerpool), so it's not bogging down the main thread so much.  This will help a bunch.

For what it's worth, it is catching a ton of real issues:
<img width="1173" alt="Screen Shot 2021-08-01 at 3 22 55 PM" src="https://user-images.githubusercontent.com/38784/127782926-11a13956-9e78-4d4a-8f4f-9d2705352d07.png">

dafadllyn:
Doesn't iD just checks my edits? At least that's what the "Issues" settings tab says.

(As for the screenshot, this likely isn't an issue. The buildings in the old town of Bern are [very old](https://en.wikipedia.org/wiki/Old_City_(Bern)) and thus quite irregular. They have been carefully mapped by a user using a map with the [official ground plans](https://wiki.openstreetmap.org/wiki/Canton_of_Bern_Address_Import#Background) we have the right to use.)

bhousel:
[...]

PR #8626, opened 2021-08-05
Validator improvements

### LINKED ISSUE #8603: "Fixme" requests are again being counted as outstanding issues
Ver 2.20 brought back "fixme" requests counting as outstanding issues that need to be resolved. 
Please change this back - these are requests, and not always actionable.

### LINKED ISSUE #8613: outstanding issues viewed are being counted towards ignored outstanding issues
With the recent iD update on live, it seems that it's all around easier to trigger an ignored outstanding issue, even if the user did not touch a feature, perhaps even just viewing (sometimes not even clicking on) a feature that has some issue in iD seems to now be sufficient to trigger the warning which gets tracked with the check-in.  Recently I've been touching turn restriction relations (without touching ways) and iD decides to include me as ignoring bus route warnings.  These are unrelated features/relations as I didn't split ways, etc., and perhaps this issue is the same as the FIXME issue #8603 .

This "ignored warning" count that gets tagged with each check-in seems quite silly and also untenable.  For those people doing user metrics, perhaps it's just HDYC, this number will become ridiculously large and impossible to reduce to zero.  For each warning ignored, everyone ignores them, and keeps on going up.  Every time we fix an error whether in iD or not, only ONE person is credited for fixing an error (and it could even be fixed in another editor) and the ignored warning count becomes an ever increasing, meaningless number.  These warning/issues are already being checked with OSMI, KeepRight, Osmose, etc. and these at least will clear credits for everyone when something gets fixed.  Is there any specific reason a problem warning ignored in iD needs to be tracked but not through JOSM/Potlatch/Merkaartor/or any other editor?  Suggestion is to stop including this count with the checkins.  In my opinion the QA should be done at the check-in server so it doesn't matter which editor software is used, not any specific editor software.

Just my observation/opinion/rant/2 cents/...

bhousel:
- a46a34564 Only expand a validation set to include parent multipolygon relations 
The previous code was grabbing _all_ parent relations, which is too much.
For example: if a user changed a road, the validator was treating it like
the user had changed bus and highway routes along that road.
(closes #8613)
(improves #8612 performance issues a lot too)

- 3b0a85040 If undo'd back to the base graph, don't show head issues as user issues
This was miscounting the "issues caused by the user" in some cases

- 9f58f1fb5 Improve code for focusing a validation issue on a relation 
The "center" of the issue might be a spot of map that doesn't contain the relation. 
This code chooses a piece of the relation that has been downloaded and focuses on that instead.

- 0085c4187 Store whether a result is provisional before filtering it
Filtering returns a new array, which was clobbering the "provisional" flag.
This was causing provisionally results to not be reprocessed later,
which meant that certain "outdated_tags" results would not be in the baseCache.
(cache of issues _before_ user edits), which would count them towards issues _caused_ by the user.

- 857b9c9ad Exclude 'fixme'/'help_request' warnings from changeset tags. 
They still appear in the issue list and in the entity editor. 
(closes #8603)


modules/core/validator.js
```diff
@@ -262,17 +261,47 @@ export function coreValidator(context) {
...
+    let focusCenter;
+
+    // Try to focus the map at the center of the issue..
+    const issueExtent = issue.extent(graph);
+    if (issueExtent) {
+      focusCenter = issueExtent.center();
+    }
 
-    // select the first entity
+    // Try to select the first entity in the issue..
```
mbrzakovic, line 274: Oh, I didn't notice this relation case before, good improvement.

What would make this perfect though, would be to focus on a node/way that is geometrically closest to the current viewport. This is low-priority so if it poses some substantial work feel free to disregard it.

modules/core/validator.js
```diff
@@ -176,24 +176,23 @@ export function coreValidator(context) {
...
+    const baseGraph = context.history().base();
     let issues = [];
     let seen = new Set();
 
     // collect head issues - caused by user edits
-    let cache = _headCache;
-    if (_headGraph) {
-      Object.values(cache.issuesByIssueID).forEach(issue => {
-        if (!filter(issue, _headGraph, cache)) return;
+    if (_headGraph && _headGraph !== baseGraph) {
```
mbrzakovic, line 184: I have a couple of questions on this:
1) In general, If some entity with already existing issue was edited, should this issue be treated as a new/'head' issue and be previewed in warnings tab and on commit page? I've observed that iD v2.19.6's answer to this was 'no'.
2) If we go with 'yes', then we should be carefull in cases such as these. I don't think that the current logic handles the following:
0. Position near an entity (e1) that has an already existing issue + open warnings tab (tab: no issues)
1. Make a change on an entity that does not resolve an issue (tab: e1's issues)
2. Undo (tab: no issues)
3. Make any other change (add new entity etc.) (tab: e1's issues) - [expected (no issues)]

In this case most likely additional logic would have to be added that would unload/load head_cache on undo/redo.

mbrzakovic (changes requested):

modules/core/validator.js
bhousel: I did consider that, but I think because relations can be very large it's ok to just find the first reasonable node or way, rather than looking at all of them to decide which one is closest (which could take a long time).

modules/core/validator.js
bhousel: > In general, If some entity with already existing issue was edited, should this issue be treated as a new/'head' issue and be previewed in warnings tab and on commit page? I've observed that iD v2.19.6's answer to this was 'no'.

I think yes, if a user edits a feature and creates and issue with it, it should count as a new issue.  iD 2.19.6 did have trouble counting issues accurately because iirc it just ran all the validations twice on every edit.

> If we go with 'yes', then we should be carefull in cases such as these. I don't think that the current logic handles the following:
> 0. Position near an entity (e1) that has an already existing issue + open warnings tab (tab: no issues)
> 1. Make a change on an entity that does not resolve an issue (tab: e1's issues)
> 2. Undo (tab: no issues)
> 3. Make any other change (add new entity etc.) (tab: e1's issues) - [expected (no issues)]
> 4. In this case most likely additional logic would have to be added that would unload/load head_cache on undo/redo.

I think `validateEntitiesAsync()` is supposed to uncache the detected issues before validation, but I do see that across undo it doesn't completely remove them all.  I'll investigate and try to fix, but would prefer for this not to be a blocker to these other improvements.

mbrzakovic: Ok, I won't block this PR as there are several good changes done.
It is possible you didn't fully catch the case I am describing, so I will create a separate ticket and link it here to track this.

mbrzakovic (approved):
Approved but will raise a separate ticket on one case.

modules/core/validator.js
mbrzakovic: https://github.com/openstreetmap/iD/issues/8632

Issue #6577, opened 2019-06-25
Handle tag to disallow automatic brand tagging


bikeoid:
I have seen several local restaurants with same or similar names to chains being incorrectly upgraded to the chain branding.   I would like to mark these so they don't get re-tagged with branding later.

  The tag not:brand:wikidata has been suggested for this.

bhousel:
> The tag `not:brand:wikidata` has been suggested for this.

This sounds good to me, but can you link a discussion about it?  Just saying "it's been suggested" doesn't mean much.

BjornRasmussen:
It's been discussed on the #poi channel of the osmus slack.

bikeoid:
Also https://osmus.slack.com/archives/C029HV951/p1561427656403000

jnicho02:
We've used the 'not:' prefix elsewhere for 'not:addr:postcode' to flag a common mistake, and i'm sure there must be others. It is probably a fairly acceptable thing to do.

quincylvania:
@bikeoid @jnicho02 I agree 100% that introducing this tag would be acceptable based on historical precedent. However, a few members of the OSM community are dead-set against iD introducing any tag without consulting them, resulting in trivial tag introductions having an outsized negative impact on the health of the iD project (see #6332). As such, a new tag like this needs to receive lots of real-world usage and/or go through the "approval" process before we can support it 🙁

1ec5:
[`not:*`](https://taginfo.openstreetmap.org/search?q=not%3A) is already a well-used tagging scheme, just not `not:brand:wikidata` specifically yet. For example, #6411 is about [`not:name`](https://wiki.openstreetmap.org/wiki/Key:not:name). But no usage of `not:brand:wikidata` yet, so to kick things off, I tagged a couple independent restaurants that happen to match the name of chains in the same country:

https://www.openstreetmap.org/changeset/71606044
https://www.openstreetmap.org/changeset/71606149

If the validator warning about the missing brand tags could have an option to permanently ignore the warning (adding the `not:brand:wikidata` or `not:brand` tag), then we could potentially remove some disambiguating suffixes from name-suggestion-index that are there just to account for independent stores.

1ec5:
[Added some additional documentation to the wiki](https://wiki.openstreetmap.org/wiki/Key:not:), for what it’s worth.

SK53:
[...]

PR #8628, opened 2021-08-06
Add a validator fix option to tag as `not:` a matched item

bhousel:
This adds an option to the outdated_tags.js validator to add a `not:` tag for a suggestion that is wrong.  This was discussed on #6577, and while iD did get the ability to recognize and ignore `not:` tags, never actually implemented the option to let users add them.

For example, there is an independent "Noah's Bagels" shop near me in New Jersey that is not affiliated with the chain of Noah's Bagels [in California](https://locations.noahs.com/us/ca). 

Screen Shot 2021-08-06 at 11 10 43 AM

Clicking the `not` option adds the tag `not:brand:wikidata=Q64517373`, which will exempt this POI from the tag suggestion going forward:

Screen Shot 2021-08-06 at 11 16 31 AM

I'm not completely happy with the wording "Tag as **not:** Noah's Bagels".   
Something like "Tag as **not:** Q64517373" would be maybe more correct, but maybe more confusing.

cc @1ec5, who has wanted this for a while 😁


bhousel:
Here's what it looks like with "Tag as **not:** Q6889938".

Screen Shot 2021-08-06 at 12 39 39 PM

data/core.yaml
```diff
@@ -1941,6 +1941,9 @@ en:
       tag_as_disconnected:
         title: Tag as disconnected
         annotation: Tagged very close features as disconnected.
+      tag_as_not:
+        title: "Tag as <strong>not:</strong> {name}"
```
1ec5: Translators will need to know not to translate “strong” and “not” literally. I’m also not sure every language will find it convenient to use a word for “not” this way. (Maybe there’s a language that has to stick the negation somewhere else in the sentence.) The use of the `not:` namespace is essentially an implementation detail for those familiar with raw tags, so I think we could remove the semicolon without sacrificing clarity.

modules/validations/outdated_tags.js
```diff
@@ -116,6 +117,19 @@ export function validationOutdatedTags() {
             }
           })
         ];
+
+        if (nsiResult && nsiResult.matched) {
+          fixes.push(
+            new validationIssueFix({
+              title: t.html('issues.fix.tag_as_not.title', { name: nsiResult.matched.displayName }),
```
1ec5: A QID in this spot would be inscrutable to ordinary mappers, especially since there’s no link or even a way to select the text. On the other hand, if the correct and incorrect brands happen to share the same name, “Tag as **not** Noah’s Bagels” wouldn’t look like the right option either. How about “Tag as a different Noah’s Bagels” or “Tag as **not the same** Noah’s Bagels”?

1ec5 (commented):
Very nice. 🍔🥯🍜

data/core.yaml
bhousel: I agree, good points on translation 👍

modules/validations/outdated_tags.js
bhousel: > “Tag as not the same Noah’s Bagels”

cool, yes this is much clearer - I'll go with this

bhousel: Screen Shot 2021-08-06 at 5 12 32 PM

1ec5 (approved):

mbrzakovic (approved):

PR #8637, opened 2021-08-13
More validator fixes

### LINKED ISSUE #8632: Validator - Existing issues handling
Question:
- In general, If some entity with already existing issues was edited, should the pre-existing issues be treated as a new/'head' issue and be previewed in warnings tab and on commit page (step 1)? I've observed that iD v2.19.6's answer to this was 'no'.
- If we go with (current) 'yes', then I've observed the following inconsistency:

Repro (dev endpoint, dev osm connection):

0. Position near an entity (e1) that has an already existing issue + open warnings tab (tab: no issues)
image

1. Make a change on an entity that does not resolve an issue (tab: e1's issues)
image

2. Undo (tab: no issues)
image

3. Make any other change (add new entity etc.) (tab: e1's issues) - [expected (no issues)] 
image

bhousel:
This includes a fix for #8632, and a bunch of other miscellaneous things.. main stuff listed below:

- 1f172d562 Make the crossing_ways hash less strict 
Previously it was including a lot of data about the edge, and a very specific
crossing location.  This meant that any tiny perturbation in the crossing ways
would generate a new issue hash, effectively "fixing" the old crossing issue and
creating a new one.

- 34c3ea472 / 93b868d95 Improvements to the code that credits a user with fixing an issue (re: #8632)

- bb0b5786d Use `context.graph()`/`context.hasEntity()` in code that collect the issues to display to the user, not `cache.graph`, because that is the graph that the calling code will be using.  (this was causing "Entity not found" errors)

- 2434e5eda In the head cache, only validate features that the user is responsible for (closes #8632)
For example, a user can undo some work and an issue will still present in the
head graph, but we don't want to credit the user for causing that issue.

- 3e8d33a66 Use `coreDifference.complete()` instead of `entityIDsToValidate()` 
From what I can tell, this code is nearly the same as what the "complete" difference already gives us - combined nodes from both previous and current, multipolygon members, parents of nodes/relations

- 96c5dd1c7 Store graph with validation cache, give them names, es6 some things

### DIFF
--- a/modules/core/validator.js
+++ b/modules/core/validator.js
@@ -492,20 +492,24 @@ export function coreValidator(context) {
 
     // If we get here, its time to start validating stuff.
     _headCache.graph = currGraph;  // take snapshot
-    const difference = coreDifference(prevGraph, currGraph);
+    const incrementalDiff = coreDifference(prevGraph, currGraph);
+    const entityIDs = Object.keys(incrementalDiff.complete());
 
-    // Gather all entities related to this difference..
-    // For created/modified, use the head graph
-    let entityIDs = difference.extantIDs(true);   // created/modified (true = w/relation members)
-    entityIDs = entityIDsToValidate(entityIDs, currGraph);  // expand set
+    // const difference = coreDifference(prevGraph, currGraph);
 
-    // For modified/deleted, use the previous graph
-    // (e.g. deleting the only highway connected to a road should create a disconnected highway issue)
-    let previousEntityIDs = difference.deleted().concat(difference.modified()).map(entity => entity.id);
-    previousEntityIDs = entityIDsToValidate(previousEntityIDs, prevGraph);
-    previousEntityIDs.forEach(entityIDs.add, entityIDs);   // concat the sets
+    // // Gather all entities related to this difference..
+    // // For created/modified, use the head graph
+    // let entityIDs = difference.extantIDs(true);   // created/modified (true = w/relation members)
+    // entityIDs = entityIDsToValidate(entityIDs, currGraph);  // expand set
 
-    if (!entityIDs.size) {
+    // // For modified/deleted, use the previous graph
+    // // (e.g. deleting the only highway connected to a road should create a disconnected highway issue)
+    // let previousEntityIDs = difference.deleted().concat(difference.modified()).map(entity => entity.id);
+    // previousEntityIDs = entityIDsToValidate(previousEntityIDs, prevGraph);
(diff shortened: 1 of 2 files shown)


modules/core/validator.js
```diff
@@ -481,26 +484,19 @@ export function coreValidator(context) {
...
-    // For modified/deleted, use the previous graph
-    // (e.g. deleting the only highway connected to a road should create a disconnected highway issue)
-    let previousEntityIDs = difference.deleted().concat(difference.modified()).map(entity => entity.id);
-    previousEntityIDs = entityIDsToValidate(previousEntityIDs, prevGraph);
-    previousEntityIDs.forEach(entityIDs.add, entityIDs);   // concat the sets
+    // If we get here, its time to start validating stuff.
+    _headCache.graph = currGraph;  // take snapshot
+    _completeDiff = context.history().difference().complete();
+    const incrementalDiff = coreDifference(prevGraph, currGraph);
+    const entityIDs = Object.keys(incrementalDiff.complete());
```
mbrzakovic, line 491: @bhousel Will this gather entire hierarchy to be validated? On touching a node (e.g. that belongs to a way street, that belongs to a Bus route that belongs to a Route master super relation), all of these would be added for validation.
Previous code would just add node and a way, which I think we established is enough.

mbrzakovic (changes requested):

modules/core/validator.js
bhousel: I don't think it gathers full hierarchies - `difference#complete` only recurses down for Multipolygon relations
https://github.com/openstreetmap/iD/blob/develop/modules/core/difference.js#L202

The point of this "complete" diff is to determine "if the user touches something, what needs to be redrawn" and I think we can reuse that same code to determine "what needs to be revalidated" rather than having a separate logic in the validator.  There's a risk that we might be missing some things that the old code caught, but I'm ok with that for now.

mbrzakovic: Oh yes, but perhaps you misunderstand - I was talking about recurse **up**. The code you linked will go 'up' in parenting chain until it reaches the end.
Actually I think re-using this method is a good move, I am just double checking on how will this impact performance (now modifying one node that I provided as example would queue Route master super relation for validation). 
@bhousel

mbrzakovic: I just did a quick look and I didn't find any red-flags that this can hurt perf. Still I want to hear your opinion and if needed we can improve this in additional change.

mbrzakovic (approved):

PR #8642, opened 2021-08-16
Test improvements

bhousel:
A bunch of fixes for the tests.  

- 4f1d5f085 Similar to e7ad3845f, avoid using finally(done) as it swallows errors

- e7ad3845f Fix file fetcher tests that touch the intro_graph 
  - Before it wasn't actually loading the intro graph because assetPath not set 
  - The test for Three Rivers City Hall was using the wrong nodeid
  - All these test errors were being ignored because of the use of .finally(done) 
  - To actually make an error happen, we can call `done(err)` with the Error

- 1282d3b05 Don't use a util function for a thing that exists 1x in the code

- 746417e1f Fix coreLocations tests access of locationSetID property

- 5aee120a8 Prefill the locale_tagging_en with fake string data 
  This avoids the numerous "missing translation" errors filling up the console

- bb84962e5 Only insert sources into the fileMap if not already there
  This allows us to setup the fileMap correctly for testing so that it never tries to fetch assets from the CDN.

### DIFF
--- a/modules/core/file_fetcher.js
+++ b/modules/core/file_fetcher.js
@@ -1,4 +1,3 @@
-import { utilFetchJson } from '../util/util';
 import parseVersion from 'vparse';
 // Double check this resolves to iD's `package.json`
 import packageJSON from '../../package.json';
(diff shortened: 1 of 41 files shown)


mbrzakovic (approved):
All good.

PR #8650, opened 2021-08-19
add link button next to website tag

k-yle:
This PR adds a button next to the [`website=`](https://wiki.osm.org/Key:website) tag, which opens the URL in a new tab when clicked.

This avoids the need to copy-paste the URL into a new tab, which is an easy way to accidently edit or delete the tag value. 

website

In the screenshot above, compare the [`website=`](https://wiki.osm.org/Key:website) tag to the [`operator=`](https://wiki.osm.org/Key:operator) tag above, which has no button.


css/80_app.css
```diff
@@ -1706,9 +1706,12 @@ a.hide-toggle {
 .form-field-input-text > input:only-of-type,
 .form-field-input-tel > input:only-of-type,
 .form-field-input-email > input:only-of-type,
-.form-field-input-url > input:only-of-type {
+.form-field-input-url > input:only-child {
```
mbrzakovic, line 1709: I can't hit this case. Can you please just quickly post why did you add this .css change?

data/core.yaml
```diff
@@ -7,6 +7,7 @@ en:
     zoom_to: zoom to
```
mbrzakovic, line 7: Usually when some change is done in core.yaml it's good to also push in the en.min.json  (you can do: npm run all) just to have this reflected immediately in en translation.

mbrzakovic (approved):

css/80_app.css
k-yle: the DOM looks like this:
```html
<div class="form-field-input-url">
	<input type="text" />
</div>

<!-- or -->

<div class="form-field-input-url">
	<input type="text" />
    <button></button>
</div>
```

`input:only-of-type` won't work to identify the second case, because the button is a different element. I think the rule was written assuming that `<input type=button />` would be used instead of `<button> </button>`. 
The existing code (line 106 & 131 in that line) uses `<button>` so I wanted to be consistent with that. I can change it to `<input type=button />` and remove the css change if you prefer.

mbrzakovic: Got it, the current version is ok, and thanks for paying attention to the details such as eliminating border radius for the corner next to the button.

PR #8663, opened 2021-08-28
More validator fixes

### LINKED ISSUE #8655: Bug - Use of 'connect the features' multiple times
This issue can be recreated by adding a marked crosswalk that crosses a dual carriageway. An issue should pop up with iD suggesting to 'connect the features'. This function works the first time, but when clicking on it again, the marked crosswalk node is only connected to the crosswalk way and not with the roadway.

bhousel:
closes #8655 
also includes a better fix for #8632

- 78961072a79e66789dadc0486d5979105f7bbe49 - Do need to actually validate the head entities
    re: the fix for #8632 - we can't actually skip validation on these.
    The better solution is to move the check to `getIssues()` so the user isn't
    credited for causing the issues unless it's something they actually touched

- 113f0794492f8bb88c0b625067c3705863e11def - Uncache the entity before starting the work
    This fixes a situation where several entities in the queue are involved in a
    disconnected crossing, and the first one detects the disconnection, but a
    later one clears out that first detection from the cache.  Now we clear
    caches one time before starting the validation work.

- b94151396d43c46bca3b7ce9761a3b2845ab4f1d - Expand set of entities to validate to include _related_ to issues
    This will catch the situation where an edit or undo affects something related without
    actually touching the item - for example an undo can cause a connected way to disconnect
    from the main graph.

- 1b5bd4d9c39c061fbc3431cfaff39747c64de28e - Actions performed (e.g. adding midpoint) must trigger validation
   Adding a midpoint was actually not triggering the re validation, so the existing fix 
   would not add the crossing point where it was supposed to (because the "edge" was wrong)
   (re: #8655)

-   9e3df2c4aa342967addf1daf9585fd7f3eea6b8c - Create an issue `key` property that changes when data needs refresh
    see [comment](https://github.com/openstreetmap/iD/issues/8655#issuecomment-906654447)
    (closes #8655)

### DIFF
--- a/modules/core/validator.js
+++ b/modules/core/validator.js
@@ -189,16 +189,22 @@ export function coreValidator(context) {
     let seen = new Set();
     let results = [];
 
-    // collect head issues - caused by user edits
+    // collect head issues - present in the user edits
     if (_headCache.graph && _headCache.graph !== _baseCache.graph) {
       Object.values(_headCache.issuesByIssueID).forEach(issue => {
+        // In the head cache, only count features that the user is responsible for - #8632
+        // For example, a user can undo some work and an issue will still present in the
+        // head graph, but we don't want to credit the user for causing that issue.
+        const userModified = (issue.entityIds || []).some(id => _completeDiff.hasOwnProperty(id));
+        if (opts.what === 'edited' && !userModified) return;   // present in head but user didn't touch it
+
         if (!filter(issue)) return;
         seen.add(issue.id);
         results.push(issue);
       });
     }
 
-    // collect base issues - not caused by user edits
+    // collect base issues - present before user edits
     if (opts.what === 'all') {
       Object.values(_baseCache.issuesByIssueID).forEach(issue => {
         if (!filter(issue)) return;
@@ -682,11 +688,6 @@ export function coreValidator(context) {
         const entity = graph.hasEntity(entityID);   // Sanity check: don't validate deleted entities
         if (!entity) return;
 
-        // In the head cache, only validate features that the user is responsible for - #8632
-        // For example, a user can undo some work and an issue will still present in the
-        // head graph, but we don't want to credit the user for causing that issue.
-        if (cache.which === 'head' && !_completeDiff.hasOwnProperty(entityID)) return;
-
         // detect new issues and update caches
         const result = validateEntity(entity, graph);
         if (result.provisional) {                       // provisional result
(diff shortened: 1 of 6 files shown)


modules/core/validator.js
```diff
@@ -189,16 +189,22 @@ export function coreValidator(context) {
...
     let results = [];
 
-    // collect head issues - caused by user edits
+    // collect head issues - present in the user edits
     if (_headCache.graph && _headCache.graph !== _baseCache.graph) {
       Object.values(_headCache.issuesByIssueID).forEach(issue => {
+        // In the head cache, only count features that the user is responsible for - #8632
+        // For example, a user can undo some work and an issue will still present in the
+        // head graph, but we don't want to credit the user for causing that issue.
+        const userModified = (issue.entityIds || []).some(id => _completeDiff.hasOwnProperty(id));
```
mbrzakovic, line 198: It's possible that we can create a version that doesn't even need _completeDiff.

Here we could just do: if (opts.what === 'edited' && _baseCache.issuesByIssueID[issue.id]) return;
and in updateResolvedIssues method I think we could remove userModified condition.

In this case we are doing 'no' answer for the first bullet question on #8632 which I am fine with,
and we achieve simplicity and some perf improvement as we don't need to do context.history().difference().complete() on each validate().

@bhousel opinion and can you please double check if this will work?

modules/core/validator.js
```diff
@@ -795,50 +797,54 @@ function validationCache(which) {
...
-    if (issueIDs) {
-      issueIDs.forEach(issueID => {
+    const entityIssueIDs = cache.issuesByEntityID[entityID];
+    if (entityIssueIDs) {
+      entityIssueIDs.forEach(issueID => {
         const issue = cache.issuesByIssueID[issueID];
         if (issue) {
           cache.uncacheIssue(issue);
-        } else {
+        } else {  // shouldnt happen, clean up
```
mbrzakovic, line 847: Please fix few spelling errors reported

modules/core/validator.js
bhousel: Comparing the difference between two graphs with `coreDifference.complete()` is already really fast - for example `map.js` uses this function to decide what features to redraw.

I think it's important to correctly credit the user with what issues their edits fixed (in `updateResolveIssues()`) and what issues their edits caused (in `getIssues()`) and to have this work reliably across undo/redo.

modules/core/validator.js
bhousel: 🙄

modules/core/validator.js
mbrzakovic: `I think it's important to correctly credit the user with what issues their edits fixed (in updateResolveIssues()) and what issues their edits caused (in getIssues()) and to have this work reliably across undo/redo.`
- of course, I completely agree, this should be validator's top priority no doubts there.

My points was, that from a quick look, I think we could achieve the abovementioned with additional simplicity.
I don't have a strong opinion on this, if it's not much of a perf improvement then it's a low impact.

mbrzakovic (approved):

PR #8671, opened 2021-09-02
create a quick method of drawing ways

### LINKED ISSUE #1614: Follow a path

k-yle:
Closes #1614 

When drawing an new way/area that abuts an existing way/area, it becomes quite boring clicking on each node to join your new area to the existing one. 

This PR allows you to press <kbd>F</kbd> while drawing a way/area to automatically keep drawing the new way along the edge of the existing way. 

Example video: After drawing 3 nodes manually, I press <kbd>F</kbd> repeatedly to continue drawing

https://user-images.githubusercontent.com/16009897/131825498-adf2627b-c1a5-4903-86e5-71787aa4fa2e.mp4

This is really helpful when mapping landuse/vegetation.

Outstanding issues: (~~I need help/guidance with these please~~)
 - [x] ~~Not sure if <kbd>7</kbd> is the right key for this. Just a random choice~~ Changed to <kbd>F</kbd>, so the key to open the settings pane is now <kbd>U</kbd>.
 - [x] ~~I think I've registered the keybindings wrong, because it should only work while drawing a way.~~ fixed 
 - [x] ~~Not sure what the techincal name for this feature is. I've called it 'race mode' for now.~~ Renamed it to 'Follow'
 - [x] ~~There's a bug when you pass the first/last node of a closed way~~ fixed

Limitations:
 - It's not smart enough to work in [this](https://user-images.githubusercontent.com/16009897/131825741-88eadda8-1d45-4e0e-b58b-36f8c2f616c0.png) situation - if there are already two ways sharing the same nodes, it can't help you join a third way to those nodes


bhousel:
nice! this is #1614

jjiglesiasg:
Worth more than a try. Thanks

k-yle:
thanks @bhousel, ironically I 👍'd that issue and then forgot about it - happy to close this PR if #7649 is a more robust solution

bhousel:
> thanks @bhousel, ironically I 👍'd that issue and then forgot about it - happy to close this PR if #7649 is a more robust solution

I think what you built is perfect, though it would be nice to use whatever key Potlatch or JOSM uses for this feature (if possible), instead of <kbd>7</kbd>.

pnorman:
I think JOSM's default is `f` for "follow".

k-yle:
I've updated it to use <kbd>F</kbd>. Currently <kbd>F</kbd> is being used to open the settings pane, so we'll need a new shortcut for settings (I've changed it to <kbd>U</kbd> for now). 

I've been mapping landuse and vegetation from this branch over the past 2 days and it's been working quite well

modules/behavior/draw_way.js
```diff
@@ -395,6 +399,9 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
             return;
         }
 
+        _secondLastAddedNode = _lastAddedNode;
```
mbrzakovic: I've noticed two things here:
1) The behavior is a bit buggy when we use it in combination with undo/redo. Repro: Draw 2 nodes -> F -> Ctrl+Z -> F
2) If the first node that is placed is an existing one, then we actually need 3 vertexes in order to get started...

Allow me to propose the following: Instead of persisting the _last and _secondLast, try to derive this information from the current way you are modifying (_origWay,nodes). This state should follow editting history by default.
If this works its a 1min fix and would nail both 1) and 2).

mbrzakovic (changes requested):

data/core.yaml
```diff
@@ -345,6 +345,12 @@ en:
       not_downloaded:
         single: This feature can't be moved because parts of it have not yet been downloaded.
         multiple: These features can't be moved because parts of them have not yet been downloaded.
+    follow:
```
mbrzakovic, line 349: Lets also include this capability in 'Help'[keyboards] menu.

modules/behavior/draw_way.js
k-yle: thanks, your suggestion works very smoothly and fixes both these issues

data/core.yaml
k-yle: 👍 done

k-yle:
not sure why tests for `iD.modeAddNote` are failing, it seems to be unrelated to this PR

modules/behavior/draw_way.js
```diff
@@ -412,6 +413,81 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
+
+            if (!lastNodeId || !secondLastNodeId) {
+                context.ui().flash
+                    .duration(4000)
+                    .iconName('#iD-icon-no')
+                    .label(t('operations.follow.error.needs_more_initial_nodes'))();
+                return;
+            }
+
+            const lastNodesParents = startGraph.parentWays(startGraph.entity(lastNodeId));
```
mbrzakovic, line 451: Apologies if being picky here, but I think we could improve the 'error' handling here and better inform user which actions should he take to achieve the follow behavior.
Currently if any of the last 2 nodes is not following the pre-existing path, the exception will happen in startGraph.entity() function and will be caught as unknown_error. I think it's better we handle these cases with error.needs_more_initial_nodes info.
Additionally, there is an edge case where node is pre-existing but it's not vertex (has no parent way).

To sum it up I propose the following:
Throw needs_more_initial_nodes if:
- <2 nodes have been drawn
- Any of the last 2 nodes are not matching pre-existing node (startGraph.entity would throw error, though I propose .hasEntity be used for this)
- Any of the last 2 nodes has 0 parents

modules/behavior/draw_way.js
```diff
@@ -412,6 +413,81 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
+            const secondLastNodesParents = startGraph.parentWays(startGraph.entity(secondLastNodeId));
+
+            if (lastNodesParents.length !== 1 || secondLastNodesParents.length !== 1) {
```
mbrzakovic: The current behavior is: code is able to deduce which path to follow if last two nodes have a single parent.

I think we could also deduce which path to follow in situation where only one of the last two nodes has a single parent,

mbrzakovic (changes requested):

mbrzakovic:
> not sure why tests for `iD.modeAddNote` are failing, it seems to be unrelated to this PR

It is unrelated. I thought I solved this in https://github.com/openstreetmap/iD/pull/8635 I will take a second look into this, please ignore it for now.

modules/behavior/draw_way.js
k-yle: good points, I've made this change and tested it quite a bit

modules/behavior/draw_way.js
k-yle: :+1: fixed

data/core.yaml
```diff
@@ -346,6 +346,13 @@ en:
...
+      key: F
+      error:
+        intersection_of_mutiple_ways: The last two nodes are intersections of multiple ways
+        intersection_of_different_ways: The last two nodes must be part of the same way
+        needs_more_initial_nodes: Draw at least two nodes in the same way first
+        unknown: Failed to continue way
```
1ec5: Explicitly state that the “follow” operation is disabled and suggest a recovery method:

```suggestion
        intersection_of_mutiple_ways: This line can't follow a way because multiple ways are connected to the line's last two points. Add another vertex manually to continue.
        intersection_of_different_ways: This line can't follow a way because the line is only connected to the way at a single point. Add another point manually to continue.
        needs_more_initial_nodes: This line can't follow a way because it isn't connected to enough consecutive points along a way. Add another point manually to continue.
        unknown: This line can't follow a way.
```

data/core.yaml
```diff
@@ -346,6 +346,13 @@ en:
       not_downloaded:
         single: This feature can't be moved because parts of it have not yet been downloaded.
         multiple: These features can't be moved because parts of them have not yet been downloaded.
+    follow:
+      key: F
+      error:
+        intersection_of_mutiple_ways: The last two nodes are intersections of multiple ways
```
1ec5: Some other operations’ messages distinguish between lines and areas, which would be clearer to the user than a message about “ways”:

https://github.com/openstreetmap/iD/blob/be737e6acaecf810c88621c6a1f4be6a0a16992a/data/core.yaml#L286-L287

data/core.yaml
```diff
@@ -346,6 +346,13 @@ en:
       not_downloaded:
         single: This feature can't be moved because parts of it have not yet been downloaded.
         multiple: These features can't be moved because parts of them have not yet been downloaded.
+    follow:
+      key: F
```
1ec5, line 350: On the one hand, it’s convenient that the <kbd>F</kbd> accelerator key matches the <ins>f</ins>ollow operation in Potlatch (miss you) and JOSM. On the other hand, we’re starting to run out of unmodified accelerator keys, especially ones that are reachable by the left hand.[^1] It’s forcing localizations that localize keys to play successive rounds of musical chairs. But that’s a broader issue and shouldn’t hold up this PR.

Perhaps a future enhancement could also assign the ⌘[ and ⌘] keys to this operation. It would be elegant to “zip” a way together with another way using the normal keyboard shortcuts for navigating along a way.

[^1]: If I remember correctly, some of the accelerator keys were chosen to be reachable by the left hand, assuming the right hand is operating a mouse in tandem. But I can’t remember where I saw that stated.

modules/behavior/draw_way.js
```diff
@@ -412,6 +413,84 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
+                geometry: { type: 'Point', coordinates: nextNode.loc },
+                id: nextNode.id,
+                properties: { target: true, entity: nextNode },
+            });
+        } catch (ex) {
+            context.ui().flash
+                .duration(4000)
+                .iconName('#iD-icon-no')
+                .label(t('operations.follow.error.unknown'))();
+        }
```
1ec5, line 501: Is this error handling meant to be defensive, or do we know of any specific situations where an error would be thrown and caught here?

k-yle: originally there was a case where `graph.entity` would throw, but we now prevent that from happening. So it's just defensive

data/core.yaml
k-yle: what if that last node is an intersection of a line and an area?

1ec5: Some other operations have a third message that says “features”. In general, though, the UI messages tend to favor [semantic elements](https://wiki.openstreetmap.org/wiki/Semantic_elements) over raw elements.

modules/behavior/draw_way.js
1ec5: OK, that’s good to know. I wasn’t sure if there was maybe more to say in the error message. There aren’t too many generic error messages like that in iD to imitate.

data/core.yaml
k-yle: okay, updated

mbrzakovic (approved):

PR #8675, opened 2021-09-04
Disable merge operation when it would damage relations

### LINKED ISSUE #1512: disable merges that would damage relations
Certain merge operations are unambiguously wrong, e.g.:
- merging ways participating in the 'from' and 'to' roles of a restriction
- merging ways that are members of different multipolygon relations

We should investigate if it is possible to reliably distinguish such merges from safe merges, and disable the operation in those cases while providing a message in the tooltip explaining why it's disabled.

### LINKED ISSUE #3825: Merging lines with certain relations should be warned/forbidden
When merging lines (roads, rivers) with differing relations of some types (borders!), the user should be at least warned of the change.

### LINKED ISSUE #8645: Merging of ways with different membership
Hello,

Merging ways without same membership should be either:
* forbidden - as it's the case when the tags are not matching;
* given a warning - JOSM gives the user the option to select if the relation membership will remain or not or not make the merge in the first place.

Currently if one would merge a road that is member of some route (usually hike/bike/bus) with one that isn't, the entire new segment will be part of the route. This created big problems as it damages the relation by introducing branches. 
The editor will likely not spot the error (only if they would be verifying the relation integrity by chance)

Repairing this is really hard even for an experienced editor. 
Reverting the changeset is not really an option due to not being able to blame easy (it doesn't appear as an edit to the relation)
Seeing the original state of the relation can be done only after a lot of analysis and history exploration.

This can be really easily be avoided by implementing this additional check.

### LINKED ISSUE #8674: [Feature Requst] Warn user when combining two highways with different relations
One of the commons sources of errors with relations seem to come from people combining highways where one of the highways belong to a relation when the other doesn't, as it's often not immediately clear why two highways with the exact same properties shouldn't be combined. At least it's something that happened to me a few times. Therefore I would really appreciate if iD could issue a warning to prevent these cases.

jfirebaugh:
The operation is disabled when attempting to join ways which don't belong to identical sets of relations. Restriction relations are excluded, because they are already handled with slightly different logic.

Fixes #8674
Fixes #8645
Fixes #3825
Fixes #1512


modules/actions/join.js
```diff
@@ -129,9 +129,27 @@ export function actionJoin(ids) {
...
         }
 
+        var i;
+
+        // All joined ways must belong to the same set of (non-restriction) relations.
+        // Restriction relations have different logic, below, which allows some cases
+        // this prohibits, and prohibits some cases this allows.
+        var sortedParentRelations = function (id) {
+            return graph.parentRelations(graph.entity(id))
+                .filter((rel) => !rel.isRestriction())
```
1ec5: I assume this is this code that already handles restriction relations?

https://github.com/openstreetmap/iD/blob/1ee45ee1f03f0fe4d452012c65ac6ff7649e229f/modules/actions/join.js#L162

I’m not sure if it’s a problem, but [connectivity relations](https://wiki.openstreetmap.org/wiki/Relation:connectivity) are intended to be modeled the same way as restriction relations. (Connectivity relations have a preset but not an editor, so currently you have to create one manually or morph a restriction relation into one using the raw tag editor.) Maybe `hasFromViaTo` would cover both relation types reasonably well?

data/core.yaml
```diff
@@ -316,6 +316,7 @@ en:
       relation: These features can't be merged because they have conflicting relation roles.
```
1ec5, line 316: This string could be renamed to `conflicting_relation_roles` since there’s now another one about overall relation membership.

1ec5 (commented):
Thanks, this will be quite useful, especially since there isn’t currently anything in the UI to inform the user upfront that the multiple selection has mixed relation membership.

modules/actions/join.js
jfirebaugh: I think you're right, but to do that right would require more changes than I want to get into in this PR.

data/core.yaml
jfirebaugh: I don't really want to touch connect.js or invalidate existing translations with this PR.

modules/actions/join.js
jfirebaugh: To elaborate a bit: I want to avoid getting into modifying the code you linked to so that it prohibits joining connectivity relations in the same situations as restriction relations, due to i18n nuances. But it would be easy enough to just exempt them from the code I added. That would preserve the existing behavior for connectivity relations, while disabling problematic merges for the much more common relation types like route.

jfirebaugh: Okay, okay, you convinced me to just implement full connectivity support. 😄 

Screen Shot 2021-09-08 at 8 28 47 PM

bhousel: Next @1ec5 is going to ask for support for [type=manoeuvre](https://wiki.openstreetmap.org/wiki/Relation:manoeuvre) relations. 😆

1ec5:
Thank you!

Issue #4379, opened 2017-09-27
Don't export all of d3


bhousel:
followup from https://github.com/openstreetmap/iD/pull/4372#issuecomment-332292559

We're currently doing this in our iD [`index.js`](https://github.com/openstreetmap/iD/blob/master/modules/index.js):
```js
import * as d3 from 'd3';
export { d3 };
```

So that later we can use it from our [`spec_helper.js`](https://github.com/openstreetmap/iD/blob/master/test/spec/spec_helpers.js) so that tests can use it.
```js
window.d3 = iD.d3;   // TODO: remove
```

Instead our test harness should be able to require d3 separately in [`index.html`](https://github.com/openstreetmap/iD/blob/master/test/index.html):
```html
<script src='../node_modules/d3/build/d3.js'></script> 
```

... and that _mostly_ works, except some of the tests that use events don't like it at all.  `d3.event` is special because it's like a global event reference.  The [d3 docs](https://github.com/d3/d3-selection#event) have this to say:

>If you use Babel, Webpack, or another ES6-to-ES5 bundler, be aware that the value of d3.event changes during an event! An import of d3.event must be a live binding, so you may need to configure the bundler to import from D3’s ES6 modules rather than from the generated UMD bundle; not all bundlers observe jsnext:main. Also beware of conflicts with the window.event global.

☝️ So I think this is an actual problem for how the tests are written, and we should figure out what this means and how to do the thing.  I did a bit of debugging, but it involves digging into what the bundled d3 code is doing, because the sourcemaps lie.

Spensaur-K:
I'll have a crack at it

darthgera123:
ill work on this issue

darthgera123:
when i make the following changes obviously by removing the import command, the entire site goes blank. So what exactly do i need to do, could you please guide me

bhousel:
> when i make the following changes obviously by removing the import command, the entire site goes blank. So what exactly do i need to do, could you please guide me

I would expect that removing the
```
import * as d3 from 'd3';
export { d3 };
```
...should not make the whole site go blank - but it should make several of the tests fail.

What kinds of errors do you see in the JavaScript console?

darthgera123:
Its going blank unfortunately, atleast on my local host. Please guide

bhousel:
I spent some time on this today and yesterday.
What's happening when we try to include d3 in a `<script>` tag is that the tests end up with a separate d3 from the iD internally bundled one.  

[...]

PR #8683, opened 2021-09-08
Switch all pure math/vector methods to use the iD SDK.

Bonkles:
This is the first in a potential series of PRs to cut iD over to use the iD-SDK math library (https://www.npmjs.com/package/@id-sdk/math). 

The iD-SDK is an effort to reduce the amount of code that needs to be cloned/forked for downstream projects from iD to do their work effectively. 

This PR does the following: 
* Removes the geo/vector code and unit tests (they're now embedded within the SDK, so no need to keep them in iD's codebase
* Replaces those methods / imports with SDK referenced methods


mmd-osm:
Assuming I need to log an issue with iD-SDK,  how would I do that? Right now GitHub tells me that’s limited to collaborators.

_An owner of this repository has limited the ability to open an issue to users that are collaborators on this repository._

Bonkles:
> Assuming I need to log an issue with iD-SDK, how would I do that? Right now GitHub tells me that’s limited to collaborators.
> 
> _An owner of this repository has limited the ability to open an issue to users that are collaborators on this repository._

Good point! As this is the first time the code will be 'public' and in use, we should change that policy and allow folks to submit issues. 

UPDATE:  You should now have the option in the repo to submit issues.

test/spec/spec_helpers.js
```diff
@@ -47,6 +47,7 @@ mocha.setup({
 expect = chai.expect;
 
 window.d3 = iD.d3;   // Remove this if we can avoid exporting all of d3.js
+window.sdk = iD.sdk;
```
mbrzakovic, line 50: Was reading this issue https://github.com/openstreetmap/iD/issues/4379 and my understanding is that re-export of d3 is done due to d3.event having some internal states => doesn't mean that this should be a common practice for all dependencies.
Curious: Could we just include id-sdk scripts in test/index.html directly (via <script src... )?

mbrzakovic (commented):
@Bonkles please resolve the code merge conflicts

PR #8707, opened 2021-09-21
Show ref in the name of route relations

### LINKED ISSUE #8559: Show ref in the name of route relations
from https://github.com/openstreetmap/iD/pull/8276#issuecomment-869580303:

image

it would be much more helpful if the ref (the bus route number) is shown

k-yle:
Closes #8559, cc @1ec5 

Editing bus routes with similar names is very confusing. Especially when the `ref` is what you use conversationally, not the `name`.

This PR changes `utilDisplayName` to use both the `name` and the `ref` tag, but only for routes.

<table>
<tr>
	<td><strong>Before
	<td><strong>After
<tr>
	<td>
	<td>
<tr>
	<td>
	<td>
<tr>
	<td>
	<td>
</table>


data/core.yaml
```diff
@@ -735,6 +735,7 @@ en:
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
     display_name:
+      ref_name: "{ref}: {name}"
```
1ec5, line 739: Should there be a `network_ref_name` string in case the `network` is also tagged? This could be especially useful in a transit hub served by multiple bus systems.

/ref https://github.com/openstreetmap/iD/issues/8559#issuecomment-871020386

test/spec/util/util.js
```diff
@@ -230,6 +230,13 @@ describe('iD.util', function() {
         it('returns the name if tagged with a name', function() {
             expect(iD.utilDisplayName({tags: {name: 'East Coast Greenway'}})).to.eql('East Coast Greenway');
         });
+        it('returns just the name for non-routes', function() {
+            expect(iD.utilDisplayName({tags: { name: 'Abyssinian Room', ref: '260-115' }})).to.eql('Abyssinian Room');
+        });
+        it('returns the name and the ref for routes', function() {
+            expect(iD.utilDisplayName({tags: { name: 'Lynfield to Midtown', ref: '25L', route: 'bus' }})).to.eql('25L: Lynfield to Midtown');
```
1ec5: We shouldn’t encourage the use of `name` for route descriptions anymore, now that #8276 has landed. Here’s an example of a name that would remain after all the route descriptions are removed from `name` tags:

```suggestion
            expect(iD.utilDisplayName({tags: { name: 'Lynfield Express', ref: '25L', route: 'bus' }})).to.eql('25L: Lynfield Express');
```

data/core.yaml
k-yle: I wasn't sure if your comment still applied, since the network name is sometimes shown twice in the latest version of iD.

If this is intentional then I'll make the change to also have `network_ref_name`

image

1ec5: Ah, that’s because Auckland Transport has an entry in the name suggestion index. Whenever NSI has an entry, it would make sense to hide the redundant, unpolished `network` value. However, omitting `network` would lead a regression of #3301 for any network that isn’t included in NSI.

This function is called from a couple different places. The relation membership editor happens to prepend the preset name. But `utilDisplayNameForPath()` calls this function without prepending the preset name, and some other places only use the preset name as a fallback if there’s no display name:

https://github.com/openstreetmap/iD/blob/fa5e7bca20ed38da3ae0ef87f7d3d7e7bdfd4827/modules/ui/keepRight_details.js#L105-L110

This function could take an additional parameter indicating whether to display the network if available. In the relation membership editor, that flag could be `false` if `matched.name()` is defined:

https://github.com/openstreetmap/iD/blob/fa5e7bca20ed38da3ae0ef87f7d3d7e7bdfd4827/modules/ui/sections/raw_membership_editor.js#L345-L346

k-yle: > In the relation membership editor, that flag could be `false` if `matched.name()` is defined:

That won't work because `matched.name()` seems to always be defined, it's either a name from NSI like `AT (Auckland Transport)` or a generic name like `Bus Route`. I used `matched.reference().qid` instead, is that reliable?

Here's the result now: the bottom one has `network:wikidata`, the top one doesn't:
image

1ec5: `reference()` is too broad, potentially returning a QID if the route happens to be tagged with `wikidata` or `operator:wikidata`, whereas directly referencing `matched.tags['network:wikidata']` would be more accurate:

https://github.com/openstreetmap/iD/blob/fa5e7bca20ed38da3ae0ef87f7d3d7e7bdfd4827/modules/presets/preset.js#L156-L164

Unfortunately, NSI has quite a few transit route entries that lack `network:wikidata` tags ([example](https://nsi.guide/index.html?t=transit&k=route&v=bus#aleopenvendee-ed7458)).

How about checking for `matched.tags.network` instead? After all, we’re trying to avoid displaying the same `network` twice, once from the preset and once from the actual entity. Every transit route entry in NSI should have a `network` tag. (You can verify that’s the case using `jq '.items | map(select(.tags | has("network") | not))' data/transit/route/*.json` in a clone of NSI.) @bhousel, would this be a safe assumption?

bhousel: > Unfortunately, NSI has quite a few transit route entries that lack `network:wikidata` tags ([example](https://nsi.guide/index.html?t=transit&k=route&v=bus#aleopenvendee-ed7458)).

I'm pretty sure that the ones without a wikidata tag won't get presets generated for them and then won't ever match a preset in this code.

You can also check whether `matched.suggestion = true`, that will be set if it's a name-suggestion-index preset.

k-yle: Thanks, `matched.suggestion` works well, I've changed it to that

papac25 (changes requested):
i just told

papac25:
not ready to merge

k-yle:
hello @papac25, could you please explain why this is "not ready to merge"?

HermanLeeZh:
Most of bus routes have already have "ref" in the name tag. It is a needless duplication, I think.

HermanLeeZh:
@k-yle

k-yle:
> Most of bus routes have already have "ref" in the name tag. It is a needless duplication, I think.

There are some mappers who insist that the `name` tag is for the real name, and isn't meant to contain the `ref`. If that advice is wrong or out-of-date I'm happy to close this PR and update my local bus routes instead. 

@HermanLeeZh Is there a recommended format for the name when it contains the ref? `name=[50X] Eastern Connector` or `name=(50X) Eastern Connector` or something else?

HermanLeeZh:
> > Most of bus routes have already have "ref" in the name tag. It is a needless duplication, I think.
> 
> There are some mappers who insist that the `name` tag is for the real name, and isn't meant to contain the `ref`. If that advice is wrong or out-of-date I'm happy to close this PR and update my local bus routes instead.
> 
> @HermanLeeZh Is there a recommended format for the name when it contains the ref? `name=[50X] Eastern Connector` or `name=(50X) Eastern Connector` or something else?

@k-yle Maybe you can look up this [wiki page](https://wiki.openstreetmap.org/wiki/Public_transport#Service_routes). Here provide a
recommended format `name=<type of transport><reference number>: <from> → <to>` eg :`Bus 201: Uitikon Waldegg, Bahnhof → Uitikon, Wängi` and `Trolleybus B103: Daguanyuan -> Yan Spring`. This change of display can mislead mappers.

LaoshuBaby:
Yes, I agree with @HermanLeeZh 's opinion, if we follow PTv2 name style, the display will make `<ref>` duplicate.

The `<ref>+<name>` display is our compromise to other name schema.

I suggest this new display schema should detect whether `public_transport:version` is defined before apply the display schema, if the value is `2`, it means the relation follow the wiki page.

---

BTW, the tagging schema truly have a lot of version, such as PTv1, PTv2, [Oxomoa schema](https://wiki.openstreetmap.org/wiki/User:Oxomoa/Public_transport_schema), [Zverik schema](https://wiki.openstreetmap.org/wiki/Proposed_features/Refined_Public_Transport)

The name style of Zverik schema are the same as PTv2

HermanLeeZh:
> 

Most of relations follow the PTv2 now.[page](https://taginfo.openstreetmap.org/keys/public_transport:version#values)

k-yle:
Great, thanks @HermanLeeZh and @LaoshuBaby ! I think we can close this PR then. I'll send [this link](https://wiki.openstreetmap.org/wiki/Public_transport#Service_routes) to anyone who complains about sticking the `ref` and destinations in the `name` tag.

1ec5:
Did you all miss [the footnote](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2226177#cite_note-name-description-1) on that wiki page? 😄 The strict name format was indeed part of the original proposal, but it is now widely discredited as a prominent example of “tagging for the editor”, as seen in the links I listed in [this discussion](https://wiki.openstreetmap.org/wiki/Talk:Buses#Bus_Route_Naming). Reducing iD’s dependency on the legacy name format was also the original motivation for #8276.

@k-yle, if you’re still interested in pushing this change to completion, maybe there’s a way to accommodate the legacy format in the interim, while other editors sort out their relation support. For example, we could make it so that the ref only appears for disambiguation purposes if the name doesn’t already contain it. That would avoid the potential confusion of a duplicated number and avoid unduly influencing mapper behavior.

HermanLeeZh:
> Did you all miss [the footnote](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2226177cite_note-name-description-1) on that wiki page? 😄 The strict name format was indeed part of the original proposal, but it is now widely discredited as a prominent example of “tagging for the editor”, as seen in the links I listed in [this discussion](https://wiki.openstreetmap.org/wiki/Talk:Buses#Bus_Route_Naming). Reducing iD’s dependency on the legacy name format was also the original motivation for #8276.
> 
> @k-yle, if you’re still interested in pushing this change to completion, maybe there’s a way to accommodate the legacy format in the interim, while other editors sort out their relation support. For example, we could make it so that the ref only appears for disambiguation purposes if the name doesn’t already contain it. That would avoid the potential confusion of a duplicated number and avoid unduly influencing mapper behavior.

@1ec5 Emm, I have noticed the footnote. I agree. I find the state is that most routes have `ref` on `name=`, at least in big cities. We should have a recommended format for tagging.

LaoshuBaby:
> I find the state is that most routes have `ref` on `name=`, at least in big cities.

The main idea of that footnote (and the [email](https://lists.openstreetmap.org/pipermail/tagging/2019-May/045180.html)) is we are wrong use the `name=*` tag to do a dirty information board. (But the situation will last for long time until we meet with a global consensus of PT `name=*` tagging and all major editor use a new method display combined value of `ref=*`, `from=*`, `to=*`. I think, it will be long future, it looks impossible.)

Like @1ec5 said in #8276 's Rationale part:

> Better editor support for the structured data will mitigate the pain that mappers might experience from deprecating the naming convention.

---

For @HermanLeeZh :

> We should have a recommended format for tagging.

If one day most data have comprehensive metadata tag and our major data consumer started to use structured data first, I agree with the assembled name like `Bus 88: Beijing Station -> Tiananmen Square`(and its all localized varient) can be deprecated because we already have many assembled-name schema, and looks like many country don't follow it.

https://github.com/openstreetmap/iD/pull/8276#issuecomment-869580303 

This situation is because the assembled-name schema haven't allow a ref, the best way is to deprecated current naming process , and let all editor changed to combine different part of metadata in a specific schema (whatever it is).

It need we unite all community to review all current relation, and all editor such as JOSM, iD, Potlatch and so on announced they change to this way. It is so hard so I think this is impossible

So, I retreat to the second best, I personally encourage and strongly support naming `name=*` according to a strict (but allow keyword localization) PTv2 schema, which at least takes away [the confusion](https://github.com/openstreetmap/iD/pull/8276#issuecomment-869580303 ) (in the comment I shown above) and provides a way to display important information such as from/to/ref at the same time, and no matter how long the name is, ref is easy to see and distinguish.

Thanks you all for read this.

:)

1ec5:
It is not impossible to make editors less reliant on the descriptive name format. It’s circular logic to claim it’s impossible to make this change as a reason to prevent iD from making this change. Coordination is not a bad idea, but @k-yle and I have been very careful not to rush any changes in iD that would lead to relations going unlabeled in JOSM overnight. This PR was only about displaying something more informative when a route happens to be named like it is in the real world.

The descriptive names are purely an editor concern. It’s quite unlikely that any data consumer goes out of their way to parse route numbers, origins, and destinations out of `name` when they can more easily and reliably get the values from `ref`, `from`, and `to`, respectively.

As important as PTv2 schema may be, it’s even more important to avoid [tagging for the ~renderer~ editor](https://wiki.openstreetmap.org/wiki/Tagging_for_the_renderer). The descriptive name format prevents `name` from being used in the manner it’s used for [all other features in OSM](https://wiki.openstreetmap.org/wiki/Homonymous_keys#Differing_formats), including other kinds of routes. In some places, routes have real names that we couldn’t use `name` for, and some routes have both common names and official names that we didn’t have enough keys for. That’s a bigger problem for data consumers than any theoretical concern about backwards compatibility.

PR #8708, opened 2021-09-21
keep the oldest Way when merging

### LINKED ISSUE #8687: When merging ways, older ways should be kept
When merging two ways (e.g. two parts of a highway), the older object should be kept to preserve as much history as possible.

For example: https://www.openstreetmap.org/changeset/111077685

Way ID 12305929 with a history of 14 years should have been kept. 
Way ID 928506251 with a history of 5 months should have been deleted.

Alternatively, ID could ask which way to keep.

k-yle:
Closes #8687

When you merge two ways, it now keeps the oldest one.

### DIFF
--- a/modules/actions/join.js
+++ b/modules/actions/join.js
@@ -27,7 +27,6 @@ export function actionJoin(ids) {
 
     var action = function(graph) {
         var ways = ids.map(graph.entity, graph);
-        var survivorID = ways[0].id;
 
         // if any of the ways are sided (e.g. coastline, cliff, kerb)
         // sort them first so they establish the overall order - #6033
@@ -40,12 +39,15 @@ export function actionJoin(ids) {
         });
 
         // Prefer to keep an existing way.
-        for (var i = 0; i < ways.length; i++) {
-            if (!ways[i].isNew()) {
-                survivorID = ways[i].id;
-                break;
-            }
-        }
+        // if there are mulitple existing ways, keep the oldest one
+        // the oldest way is determined by the ID of the way
+        const survivorID = (
+            ways
+                .filter((way) => !way.isNew())
+                .sort((a, b) => +a.osmId() - +b.osmId())[0] || ways[0]
+        ).id;
+
+
 
         var sequences = osmJoinWays(ways, graph);
         var joined = sequences[0];
--- a/test/spec/actions/join.js
+++ b/test/spec/actions/join.js
@@ -424,6 +424,29 @@ describe('iD.actionJoin', function () {
         expect(graph.hasEntity('w-2')).to.be.undefined;
     });
 
+    it('prefers to keep the oldest way', function () {
+        // n1 ==> n2 ++> n3 --> n4
+        // ==> is existing, ++> is existing, --> is new
+        // Expected result:
+        // n1 ==> n2 ==> n3 ==> n4
+        var graph = iD.coreGraph([
+            iD.osmNode({ id: 'n1', loc: [0,0] }),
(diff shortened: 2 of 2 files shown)


modules/actions/join.js
```diff
@@ -40,12 +39,14 @@ export function actionJoin(ids) {
...
-            if (!ways[i].isNew()) {
-                survivorID = ways[i].id;
-                break;
-            }
-        }
+        // if there are mulitple existing ways, keep the oldest one
+        const survivorID = (
+            ways
+                .filter((way) => !way.isNew())
+                .sort((a, b) => +new Date(a.timestamp) - +new Date(b.timestamp))[0] || ways[0]
```
1ec5: If I’m not mistaken, [`timestamp`](https://wiki.openstreetmap.org/wiki/Elements#Common_attributes) is the timestamp of the changeset that last modified the way, so this heuristic could very well keep the way with _less_ history in some cases.

https://github.com/openstreetmap/iD/blob/1ee45ee1f03f0fe4d452012c65ac6ff7649e229f/modules/services/osm.js#L215

The way’s ID is a more reliable indicator of its age, assuming the way hasn’t ever been repurposed for something completely unrelated, so choose the way with the lowest nonnegative ID.

k-yle: good point, fixed

modules/actions/join.js
```diff
@@ -40,12 +39,15 @@ export function actionJoin(ids) {
...
-                survivorID = ways[i].id;
-                break;
-            }
-        }
+        // if there are mulitple existing ways, keep the oldest one
+        // the oldest way is determined by the ID of the way
+        const survivorID = (
+            ways
+                .filter((way) => !way.isNew())
+                .sort((a, b) => +a.id.substr(1) - +b.id.substr(1))[0] || ways[0]
```
mbrzakovic: A bit better is to use existing entity.osmId() method.
So: a.osmId(), b.osmId()

mbrzakovic (approved):
Small but great fix.

test/spec/actions/join.js
```diff
@@ -424,6 +424,29 @@ describe('iD.actionJoin', function () {
         expect(graph.hasEntity('w-2')).to.be.undefined;
     });
 
+    it('prefers to keep the oldest way', function () {
```
mbrzakovic, line 427: Hmm is it just me or I can't get this entire test suite (actions/join) to work?
I did a short debugging and it seems that `['restriction', 'connectivity'].forEach` section is creating a problem.
@k-yle  can you please confirm this for me?

k-yle: Yea, I took a quick look at this, if you add the following code block to `test/index.html`:

```js
window.onerror = function (message, source, line, col) {
  console.error(
    "\n\n\nFailed to execute test file.\n",
    message + "\n",
    "  at " + source + ":" + line + ":" + col
  );
};
```

Then this error is printed:
```
SyntaxError: Invalid character: '`'`
  at file:///.../iD/test/spec/actions/join.js:100:0
```

Replacing the template-strings with a normal strings fixes that error, but the test suite still doesn't run...

_(in any case, if you remove the dodgy tests, then the one I added in this PR passes)_

modules/actions/join.js
k-yle: fixed

test/spec/actions/join.js
mbrzakovic: https://github.com/openstreetmap/iD/pull/8762 resolved this

PR #8718, opened 2021-09-24
interim maintainers

mikelmaron:


bhousel:
I'm still on Slack and happy to answer questions about iD.  Given that I wrote most of the code and maintained the project for many years, I don't see how removing me from this list helps anyone.

mikelmaron:
My impression was that this section listed the maintainers of iD, not all the people who could help answer questions, which are many. I suppose this needs more of a refresh then just replacing the names of the current maintainers.

PR #8727, opened 2021-09-30
show an error if local storage is full

k-yle:
There is a limit to how much data can be stored in `localStorage` ([you can see the limit here](https://arty.name/localstorage.html)). It's about 5 MB in chrome.

As you edit the map, the size of the history in localStorage grows very fast. Once it exceeds the limit, your changes aren't saved to localStorage anymore. So if you refresh the page, you will lose your progress :(

This PR warns people when the localStorage size limit is exceeded:

error

To test this PR: draw lots of circles, select them all, and use <kbd>Shift</kbd> + <kbd>+</kbd>/<kbd>-</kbd> to scale the circles. This quickly uses up a lot of storage space.


jjiglesiasg:
I do Agree, it's not the first time that I loose by accident 3/4 of what I have done

mmd-osm:
Small nitpick: Consider *saving* your changes may be a bit unclear, as there's no way  to save your changes to a local file or similar.
I'd suggest to rephrase it to "Consider uploading your changes...".

PR #8741, opened 2021-10-08
Turn off RTL text patch on Chromium >=96

mapmeld:
Way back in #3923 and #6018 I had a fix for right-to-left line labels in Chromium-based browsers. Long story short, iD is one of the few places on the web which uses SVG textPath elements.  A text shaping bug is making complex text (such as Indic languages , #5342) or RTL text broken for a long time - [bug reported in 2014](https://bugs.chromium.org/p/chromium/issues/detail?id=374526).  We were able to use a JS library, sort of a hack really, to show corrected letters for Arabic and other RTL languages.

This week, the bug was fixed in Chrome Canary.  In this PR I suggest that we detect the browser version and skip the right-to-left label fix in Chromium >= 96.0

Should the code and alif-toolkit dependency just be removed? Chrome, Brave, etc. are currently on Chromium 94. Once the fix is more mainstream (mid-November), we could start a grace period of 6?-18? months and then remove it.

### DIFF
--- a/modules/util/svg_paths_rtl_fix.js
+++ b/modules/util/svg_paths_rtl_fix.js
@@ -1,5 +1,6 @@
 // see https://github.com/openstreetmap/iD/pull/3707
 // https://gist.github.com/mapmeld/556b09ddec07a2044c76e1ef45f01c60
+// fixed in Chromium 96.0 https://bugs.chromium.org/p/chromium/issues/detail?id=374526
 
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Thanks @mapmeld!

>  Once the fix is more mainstream (mid-November), we could start a grace period of 6?-18? months and then remove it.

Sounds good to me!

PR #8751, opened 2021-10-14
Remove `-alt-long` name match for S/T Chinese

### LINKED ISSUE #8739: Wrong translation and can't found in Transifex
When browser using locale `zh-CN`, iDeditor will show `zh-Hans` and `zh-Hant` as this, it is wrong

"现代标准汉语" don't means Simplified Chinese, and "台湾国语" don't means all Traditional Chinese, because Hong Kong, Macau, also use their kind of T-Chinese

图片

But when I set my browser locale to `zh-TW`, iDeditor show `zh-Hans` and `zh-Hant` as this, this is right

图片

However, when I want to find those string in Transfix, I can't found.

图片

I only can found this part json in this file, but it is a dist, so I must found where is source of text

```json
"zh-Hans":"现代标准汉语","zh-Hant":"台湾国语"
```
in

https://github.com/openstreetmap/iD/blob/060a73998783b6f7229ba809ed7793ffc35c3bc5/dist/locales/zh.min.json

LaoshuBaby:
According to #8739 , this part of code import unproper translation for S/T Chinese.

Remove those part of code, because only 2 kind of Chinese have `-alt-long` , other language only have `-alt-menu` so I think this will not affected other language in the future.

Close #8739

### DIFF
--- a/scripts/language_names.js
+++ b/scripts/language_names.js
@@ -92,11 +92,6 @@ exports.languageNamesInLanguageOf = function(code) {
   });
 
   for (let langCode in translatedLangsByCode) {
-    let altLongIndex = langCode.indexOf('-alt-long');
-    if (altLongIndex !== -1) {    // prefer long names (e.g. Chinese -> Mandarin Chinese)
-      let base = langCode.substring(0, altLongIndex);
-      translatedLangsByCode[base] = translatedLangsByCode[langCode];
-    }
 
     if (langCode.includes('-alt-')) {
       // remove alternative names


scripts/language_names.js
```diff
@@ -92,11 +92,6 @@ exports.languageNamesInLanguageOf = function(code) {
   });
 
   for (let langCode in translatedLangsByCode) {
-    let altLongIndex = langCode.indexOf('-alt-long');
-    if (altLongIndex !== -1) {    // prefer long names (e.g. Chinese -> Mandarin Chinese)
```
1ec5: Looks like this logic was introduced in #6702. I don’t know the reasoning, but I agree that it’s incorrect in the context of OSM tagging. There are cases where it’s necessary to distinguish between written Mandarin Chinese and written Cantonese (Yue), but it’s a lot rarer than in text-to-speech, and `name:zh` would probably contain the non-Cantonese variant. Technically, OSM could make this distinction using `name:cmn` and `name:yue`, respectively, but `name:cmn` is very rare.

1ec5 (approved):

scripts/language_names.js
LaoshuBaby: Actually we Chinese seldom distinguish what the written form called, and "zh-Hans" is a script. Not only China use Simplified Chinese, Singapore also use this varient script.
If must say some thing is "Mandarin" in pronunciation, let this distinguish in "cmn" or "yue" extlang. That‘s where “Mandarin” can be use.

So even if we don't care about the translated name of CLDR is what, it is still unnecessary to choose a name with "-long" when display "name:xx" in OSM ( Especially the zh in OSM data have different means defined by the location)

Further:
http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

LaoshuBaby: So I am more concerned about the difficulty of this PR being merged/rejected 😭

tyrasd (approved):

PR #8757, opened 2021-10-18
Add a tailing space for the semicolon

cicku:
The issue is: image

I'm not sure whether this one will make it better or not. 🤷

### DIFF
--- a/modules/ui/commit.js
+++ b/modules/ui/commit.js
@@ -124,7 +124,7 @@ export function uiCommit(context) {
                 }
             });
 
-            tags.source = context.cleanTagValue(sources.join(';'));
+            tags.source = context.cleanTagValue(sources.join('; '));
         }
 
         context.changeset = new osmChangeset({ tags: tags });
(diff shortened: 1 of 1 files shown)


1ec5:
> I'm not sure whether this one will make it better or not. 🤷

I tried previewing the syntax before and after in a diary post on osm.org; the whitespace seems to help there. However, this isn’t the same parser used for tag values and changeset comments.

diary

bhousel:
Agree with @1ec5, I don't think adding a space really helps - the url will be broken whether we delimit with <kbd>;</kbd> (as is the "standard") or <kbd>;</kbd><kbd>space</kbd> (as suggested here)

1ec5:
Oh, I didn’t mean that it wouldn’t help, just that my initial experiment using the diary entry page didn’t prove anything.

In fact, I think the space will do the right thing based on how [this way](https://www.openstreetmap.org/way/899711965)’s `website` tag renders. iD already avoids trimming spaces after semicolons for a few tags like `opening_hours`:

https://github.com/openstreetmap/iD/blob/49f5d03e55a230ce8f04289c818ff36fb0d08569/modules/util/clean_tags.js#L15-L17

If we expect `source` on a changeset to frequently contain a URL, then keeping a space makes sense, but let’s only do that if the semicolon follows a URL, as opposed to one of the keyword-like values. It might be better to add this logic to `cleanValue()` itself instead of at the call site, so that other URL-typed keys can get the same treatment.

PR #8762, opened 2021-10-21
Testing improvement

mbrzakovic:
1) Observed that addNote test fails again on gh actions. I implemented a smarter solution then https://github.com/openstreetmap/iD/pull/8635 that should catch this.
- Edit: Although the test works locally it does not work on node10 in gh action. Disabled the test for now.
2) Observed that testing silently skips execution of test (e.g. actions/join) if the test is not written correctly. [Commented here.](https://github.com/openstreetmap/iD/pull/8708/files#r719458028)
- Added more strictness in testing infrastructure that should disable such silent errors and fail testing instead.
e.g. of expected log:
[Execution errors exist] Please resolve the following:
[Execution error] messageReferenceError: Can't find variable: a sourcefile:///C:/Work/iD/test/spec/actions/add_entity.js line7 col5
[Execution error] messageSyntaxError: Unexpected token '}' sourcefile:///C:/Work/iD/test/spec/actions/join.js line190 col0

- Fixed actions/join test

cc @k-yle

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -172,6 +172,12 @@
       'spec/validations/suspicious_name.js'
     ];
 
+    window.executionErrors = [];
+
+    window.onerror = function (message, source, line, col) {
+      window.executionErrors.push("message" + message + " source" + source + " line" + line + " col" + col);
+    };
+
     var isPhantomJS = !!(navigator.userAgent.match(/PhantomJS/));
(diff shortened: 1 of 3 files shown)


bhousel:
Thanks! this will be helpful 👍

Issue #7697, opened 2020-06-11
Replace PhantomJS for headless browser testing


quincylvania:
PhantomJS was [abandoned in 2018](https://github.com/ariya/phantomjs/issues/15344) and the mocha-phantomjs-core project used by iD for testing [followed suit](https://github.com/nathanboktae/mocha-phantomjs-core/commit/562125f41e4248a9ae19b91ed2e23a95f92262e9). We've been coasting until now, but mocha just [deprecated](https://github.com/mochajs/mocha/pull/4178) some of the functions used by mocha-phantomjs-core (#7695). This means we can't keep up-to-date with mocha with our current setup, which is an unsustainable proposition. Testing is fundamental.

There's been [talk in the past](https://github.com/openstreetmap/iD/issues/5246#issuecomment-415636003) of switching away from PhantomJS, but now it's become a more immediate priority. I'm sure there are good, modern solutions out there, but we need to make sure we're still testing legacy browsers where possible. PhantomJS uses the same polyfills as Internet Explorer and as a result, tests for compatibility issues that might otherwise be difficult to find. See #6087 for related work.

HolgerJeromin:
Can this issue be closed?

PR #8764, opened 2021-10-24
Switch to karma

wvanderp:
TLDR: I replaced phanton.js with karma and fixed the tests.

In this PR, I changed phatom.js for karma and fixed the newly failing tests.

What follows is a long list of changes:

- Added karma test runner
- Removed phantom.js
- Removed the polyfills for phantom.js
- Replace the homegrown fake fetch for a module
- Implemented code coverage and ignored them in git and npm
- Fixed the issues caused by replacing the fake fetch
- updated all test-related packages

There are also some remarks and issues:

- some of the tests fail when the tests are run in a windowed environment. but not in a headless environment 
- This PR makes it difficult to run tests on IE11 because it does away with some of the polyfills. 
   - On the other hand, you can select IE11 as a runner, so we can always add that back later.
- This is a large change, and I am new to developing for ID, so this change should be tested both in development environments and CI to gain confidence in the change.
- I didn't edit any of the documentation or code added for phantom.js

sorry for all the white space changes it seems that these files were not touched for a long time.

I am happy to answer any questions or make any changes to this commit.


wvanderp:
I don't know what to do about the failing build on node 10. I can reproduce because I encounter another problem while installing.

but some of our dependencies like `@ideditor/country-coder`, `@ideditor/location-conflation`, or `osm-community-index` don't support 10

jfirebaugh (commented):
Thanks @wvanderp!

Could you please revert the stylistic and whitespace changes? Those make it more difficult to review and would clutter git history/git blame.

wvanderp:
Sorry for the potential spam. I had to remove the formatting by hand and it didn't want to be removed.

But it is gone now.

I would advise to introduce some rules for committing the changes.

mbrzakovic:
@wvanderp big thanks for this contribution!
Could you please just shortly state what are the core gains for replacing phantom with karma?

wvanderp:
TLDR: maintainability and flexibility

**Gains in maintainability:**

The main reason is that PhantomJS is depricated.
See https://github.com/openstreetmap/iD/issues/7697

Karma is a large ecosystem, and because of this, we gain a more stable future for our test runner and a lot of support online.

After some tweaking of the config files, it is possible to debug karma test with the chrome dev tools.

**Gains in flexibility:**

Because of Karmas large ecosystem, we can get a lot of new functionality out of the box. 

There are plugins for [most browsers under the sun](https://www.npmjs.com/search?q=keywords:karma-launcher). I currently only choose chrome because it is the most straightforward. But with some modifications (mainly removing es6 and adding polyfills), we can use IE11 or even PhantomJS as our runner.

I have configured karma to collect test coverage. I think this is new for the project. Also a bit difficult to add to the old setup.

tyrasd:
> This PR makes it difficult to run tests on IE11 because it does away with some of the polyfills.

See https://github.com/openstreetmap/iD/pull/8774#issuecomment-988080734, this is not really an issue for us anymore. :wink:

tyrasd:
Thanks for this very useful switch @wvanderp!

> I don't know what to do about the failing build on node 10

this was resolved in the meantime with https://github.com/openstreetmap/iD/pull/8766

> I didn't edit any of the documentation or code added for phantom.js

I merged this more or less as is, such that work on #8774 (or #8811) isn't blocked anymore. I only removed the old "test runner" index.html for phantomjs for now, but there are quite a few remaining bits of codes referencing phantomjs around and, as you said, the documentation probably also needs to be updated. Let's track that in a separate ticket -> #8843.

PR #8768, opened 2021-10-25
Perf improvement - matchTags - improve indexing data structure

### LINKED ISSUE #8612: iD performance issues since version 2.20.0

mbrzakovic:
This change goes one step further in improving iD performance than https://github.com/openstreetmap/iD/pull/8761, and strongly closes https://github.com/openstreetmap/iD/issues/8612.
Again the focus is on speeding-up the matchTags which takes ~25% of scripting as shown in Bern benchmark https://github.com/openstreetmap/iD/issues/8612#issuecomment-947929484.

### Motivation
Current indexing data structure for presets (which include ~60k items) is the following:
[geometry] (area, line, point, vertex) -> [tagKey] (amenity, highway, ...) -> presetList (items contain tagKey)

Widely adopted tags in presets (e.g: amenity, shop, brand:wikidata) have ~5k items in presetList, so matching would have to calculate score for each item in list.

If entity has a tag amenity:cafe, score-matching would be done for all presets that contains amenity key (e.g. amenity:cafe, amenity:bank, amenity:restaurant...). Score matching only accepts presets that have a match on both key and value (with an edge case of * value), therefore majority of matching score evaluations is currently redundant. (no need to consider amenity:bank, amenity:restaurant...)

### Change being made
Removes redundant match computations by reorganizing indexing data structure in following manner:
[geometry] (area, line, point, verte)  -> [tagKey] (amenity, highway, ...) -> [tagValue] (for amenity: bank, cafe, restaurant) -> presetList (items contain tagKey:tagValue).

### Testing & Result
Running https://github.com/openstreetmap/iD/tree/compare_matchTags while loading big chunk of entities showed identical results were being received for both current and improved version of matchTags method with **improved method being x8 times faster which results in ~20% faster scripting**

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -171,7 +171,9 @@ export function presetIndex() {
       (preset.geometry || []).forEach(geometry => {
         let g = _geometryIndex[geometry];
         for (let key in preset.tags) {
-          (g[key] = g[key] || []).push(preset);
+          g[key] = g[key] || {};
+          let value = preset.tags[key];
+          (g[key][value] = g[key][value] || []).push(preset);
         }
       });
     });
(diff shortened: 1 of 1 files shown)


bhousel (commented):
Makes sense to add another layer to the internal cache, as those `keyMatches` arrays were getting big and slow.
Thanks for this!

westnordost:
I think this is definitely the right approach: nested indices.

Can't say anything about the code, but approach-wise, I'll recap what I wrote in that one OSM slack thread, which boils down to:

Nest indices by tag key and values more and you will in the end arrive at just a few dictionary-lookups for a match and that's it - can't be faster than that. The index by geometry would then also not be necessary anymore. You could even do the building of an index automatically based on the number of occurances of keys/values used so that you don't need to tweak the indexing yourself manually (i.e. by what key you index).

Why could the automatic determination what keys/values to use for indexing be better than doing it manually? 

Because it may not always be `brand:wikidata` by which the presets are best indexed. In some cases (especially non-NSI-presets), it is better to group them by another key that occurs more often (for that particular key) , for example where sub-keys play a role: `key:amenity->value:bicycle_parking->key:bicycle_parking-> presetList`, e.g. the data structure could look like this in that case.

```
{
  "amenity": {
    "bicycle_parking": {
      "bicycle_parking": {
        "stands": [<bicycle stands>],
        "shed": [<bicycle shed>],
        "building": [<bicycle park>],
        ...etc
      }
    }
  }
}
```

An implementation of this kind of tree in Java is here, maybe this can inspire for a solution in typescript/javascript:

https://github.com/westnordost/osmfeatures/blob/master/library/src/main/java/de/westnordost/osmfeatures/ContainedMapTree.java

Issue #1361, opened 2013-04-22
eliminate user of 'blur' event listeners


ansis:
Most input fields have both _blur_ and _change_ event listeners. We can probably use just one, but this should be carefully tested in all browsers.

PR #8769, opened 2021-10-26
Localize numbers in numeric fields

### LINKED ISSUE #3615: Numeric preset fields should localize numbers
The “Lanes” and “Max Height” (`maxheight=*`) fields accept numbers as input, but the placeholder string for “Max Height” highlights the fact that they only accept English-style numbers (European digits and a period as the decimal point). Numeric fields like these should convert between the language’s local number format and the format expected in OSM tags. (Note that floating-point fields like “Max Height” also tend to accept units: #3614.)

Related: #3597

1ec5:
Fields of type `number`, `roadheight`, and `roadspeed` now interpret their input as a number or semicolon-delimited list of numbers according to the UI language. If the input value can’t be parsed as a localized number, the field falls back to the raw value verbatim. Regardless of how the field displays its value, it always outputs tag values containing numbers in English/POSIX format for consistency with OSM tagging conventions.

<br><i>Width field in English</i>

<br><i>Width field in Vietnamese</i>

<br><i>Width field in Arabic</i>

<br><i>Max Height in Vietnamese</i>

`roadspeed`-type fields contain combo boxes with suggested values, which are now localized:

<br><i>Speed Limit field in Arabic</i>

When using a locale that uses non-European numerals, such as Arabic, you can enter the value using European digits. The numerals are converted into localized digits on the fly as you type.

Ideally, this PR would be nothing more than changing the `<input>` tag’s `type` attribute to `number` instead of `text`. (In fact, there’s a `number` input box in the Issues panel and it works great.)

https://github.com/openstreetmap/iD/blob/810ce60c188b10ce33518470a0ad1b32f7c9d83c/modules/ui/fields/input.js#L98

However, browsers such as Firefox are a bit too strict about what can go into a `number` input box. If a feature is tagged `maxheight=below_default` or `direction=0;180`, the built-in `number` field would be blank. So instead, I’ve attempted to emulate the built-in behavior.

For languages that use non-English number formats, [translated placeholders](https://github.com/openstreetmap/id-tagging-schema/blob/42d86acc7fa8e8ff01485437889c6784f244dea1/dist/translations/ar.json#L1343) defined in id-tagging-schema should be updated to use the localized number format instead of an English/POSIX number format.

Fixes #3615.


modules/core/localizer.js
```diff
@@ -391,5 +391,35 @@ export function coreLocalizer() {
         return code;  // if not found, use the code
     };
 
+    localizer.floatFormatter = (locale) => {
+        if (!('Intl' in window && 'NumberFormat' in Intl &&
+              'formatToParts' in Intl.NumberFormat.prototype)) {
```
1ec5, line 433: Displaying and accepting localized numbers is limited to [browsers that support `Intl.NumberFormat.prototype.formatToParts`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts#browser_compatibility).

modules/core/localizer.js
```diff
@@ -391,5 +391,35 @@ export function coreLocalizer() {
...
+            return (number) => number.toString();
+        } else {
+            return (number) => number.toLocaleString(locale);
+        }
+    };
+    localizer.floatParser = (locale) => {
```
tyrasd: (minor) can you add a separating newline between these two lines?

modules/core/localizer.js
```diff
@@ -391,5 +391,35 @@ export function coreLocalizer() {
...
+        // https://stackoverflow.com/a/55366435/4585461
+        const polyfill = (string) => parseFloat(string, 10);
```
tyrasd: 1. There is no second parameter for [parseFloat](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/parseFloat).
2. Would it make sense to trim the string before parsing here, so the behaviour here is consistent with the full-fledged implementation below (line 416)?

```suggestion
        const polyfill = (string) => parseFloat(string.trim());
```

modules/ui/fields/input.js
```diff
@@ -114,8 +116,12 @@ export function uiFieldText(field, context) {
                     var raw_vals = input.node().value || '0';
                     var vals = raw_vals.split(';');
                     vals = vals.map(function(v) {
-                        var num = parseFloat(v.trim(), 10);
-                        return isFinite(num) ? clamped(num + d) : v.trim();
+                        v = v.trim();
```
tyrasd, line 138: There should be no need to perform the trimming here, since this is now done in the parser, isn't it?

modules/ui/fields/input.js
```diff
@@ -114,8 +116,12 @@ export function uiFieldText(field, context) {
...
+                        var num = parseLocaleFloat(v);
+                        if (!isFinite(num)) return v;
+                        num = parseFloat(num, 10);
+                        if (!isFinite(num)) return v;
+                        return formatFloat(clamped(num + d));
```
tyrasd: Can you example what the `num = parseFloat(num)` is supposed to do here? Isn't `num` already numeric after parsing of `v`? Can't this and the next line be just dropped?

modules/ui/fields/input.js
```diff
@@ -213,17 +219,20 @@ export function uiFieldText(field, context) {
...
+            var displayVal = val;
+            if (field.type === 'number' && val) {
+                var vals = val.split(';');
+                vals = vals.map(function(v) {
+                    v = v.trim();
+                    var num = parseLocaleFloat(v);
+                    if (!isFinite(num)) return v;
+                    num = parseFloat(num, 10);
+                    if (!isFinite(num)) return v;
+                    return clamped(num);
```
tyrasd: see above. AFAICS this could be reduced to

```suggestion
                    var num = parseLocaleFloat(v);
                    return isFinite(num)) ? clamped(num) : v;
```

modules/ui/fields/input.js
```diff
@@ -242,7 +251,18 @@ export function uiFieldText(field, context) {
 
         var isMixed = Array.isArray(tags[field.key]);
 
-        utilGetSetValue(input, !isMixed && tags[field.key] ? tags[field.key] : '')
+        var val = !isMixed && tags[field.key] ? tags[field.key] : '';
+        if (field.type === 'number' && val) {
+            var vals = val.split(';');
+            vals = vals.map(function(v) {
+                v = v.trim();
+                var num = parseFloat(v, 10);
```
tyrasd:
```suggestion
                var num = parseFloat(v);
```

test/spec/core/localizer.js
```diff
@@ -0,0 +1,40 @@
+describe('iD.coreLocalizer', function () {
+    describe('#floatParser', function () {
+        it('roundtrips English numbers', function () {
```
tyrasd, line 32: At first glance it look a bit like you took a shortcut here, by only checking that the stringify->parse chain results in correct results. Technically this wouldn't catch the error where all locales would fall back to English. Is this on purpose to make tests pass on test environments which don't support `Intl`?

modules/ui/fields/roadspeed.js
```diff
@@ -113,16 +119,20 @@ export function uiFieldRoadspeed(field, context) {
...
+        var value = rawValue;
         var isMixed = Array.isArray(value);
 
         if (!isMixed) {
-            if (value && value.indexOf('mph') >= 0) {
-                value = parseInt(value, 10).toString();
-                _isImperial = true;
-            } else if (value) {
+            if (rawValue && rawValue.indexOf('mph') >= 0) {
+                _isImperial = rawValue && rawValue.indexOf('mph') >= 0;
```
tyrasd: can't this be simplified to 
```suggestion
                _isImperial = true;
```
?

tyrasd (commented):
Thanks for this PR! It looks great! :heart_eyes: 

I've not yet looked in too much detail into `roadspeed.js` and `roadheight.js`, but please find a few comments inline below.

test/spec/core/localizer.js
1ec5: Unfortunately, none of the Node.js configurations on CI have been built with `Intl` support. 😕 This test can only test roundtripping in the most basic sense, which I guess does cover some of the older browser versions supported by iD.

tyrasd: ok, I see. Do you know if #8764 (switch from phantomjs to karma) would change anything in this regard?

1ec5: I wouldn't expect it to, since this is [about Node.js itself](https://nodejs.org/docs/latest/api/intl.html#intl_options_for_building_node_js) rather than the test harness, but I pushed openstreetmap/iD@bd67bc3f5b964f314ac44fb417df8df942606531 so you can try it out.

tyrasd: karma seems to be capable of executing tests in actual browsers. In PR #8764 it is configured to run in a (headless version of) Chrome; see https://github.com/openstreetmap/iD/pull/8764/files#diff-25bdb46a5428d318648ec9eb46a99c0268d8c96b672c46e2247a685bc3e384fbR79-R83.

I just tried it out and it passes the `formats Bengali numbers` test merged in from `bd67bc3`. However, there seems to be an issue with the Arabic numbers test, where I get the following error:

```
Chrome Headless 88.0.4324.27 (Linux x86_64) iD.coreLocalizer #floatParser roundtrips Arabic numbers FAILED
	AssertionError: expected NaN to deeply equal -0.1
	    at Context.<anonymous> (test/spec/core/localizer.js:25:54)
```

All other tests work and pass just fine.

tyrasd: ok, I could confirm the issue also on a current "regular" Chromium browser (v95), as well as with Firefox (v94). The issue seems to be that the `floatParser` isn't able to handle input of negative numbers produced for any(?) right-to-left script:

```javascript
floatParser('ar-EG')((-0.1).toLocaleString('ar-EG'))
# NaN
floatParser('he')((-0.1).toLocaleString('he'))
# NaN
```

In these cases, the result from `toLocaleString` include either a Unicode [left-to-right mark (LRM, U+200E)](https://www.fileformat.info/info/unicode/char/200e/index.htm) or a [arabic letter mark (ALM, U+061C)](https://www.fileformat.info/info/unicode/char/061c/index.htm) as the first character. Stripping these from the input string in the parser function solves the issue, but I don't know if there are any potential side effects this could have. :thinking: 

```javascript
    localizer.floatParser = (locale) => {
        // […]
        return (string) => {
            string = string.trim()
                .replace(/^[\u200E\u200F\u061C]/, '')
                .replace(group, '')
                .replace(decimal, '.')
                .replace(numeral, getIndex);
            return string ? +string : NaN;
        };
    };
```

modules/core/localizer.js
```diff
@@ -391,5 +391,36 @@ export function coreLocalizer() {
         return code;  // if not found, use the code
     };
 
+    localizer.floatFormatter = (locale) => {
+        if (!('Intl' in window && 'NumberFormat' in Intl &&
+              'formatToParts' in Intl.NumberFormat.prototype)) {
+            return (number) => number.toString();
+        } else {
+            return (number) => number.toLocaleString(locale);
```
1ec5: `Number.prototype.toLocaleString()` and `Intl.NumberFormat()` [default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#parameters) to a maximum of three decimal digits, which means this round-tripping would actually cause rounding. We can work around that behavior by setting the `maximumFractionDigits` option to 20 (the maximum value that doesn’t result in a `RangeError`).

In practice, most occurrences of the ten-thousandths place in tag values would be from poorly executed imports, and any manual editing of those values would be unlikely to require so much precision. (This code should only run due to a user-initiated edit of the value using the field, not mere viewing or manual editing of the raw value.)

test/spec/core/localizer.js
1ec5: > In these cases, the result from `toLocaleString` include either a Unicode [left-to-right mark (LRM, U+200E)](https://www.fileformat.info/info/unicode/char/200e/index.htm) or a [arabic letter mark (ALM, U+061C)](https://www.fileformat.info/info/unicode/char/061c/index.htm) as the first character.

I guess that makes sense: the Unicode Character Database gives the hyphen-minus character `-` a bidi character type of European Number Separator, which is weakly directional. The number formatter uses the ALM in Arabic and LRM in Hebrew to ensure the hyphen-minus appears on the correct side of the digits, e.g., “-3.14159” and “؜-٣٫١٤١٥٩”, and not as “‏-3.14159” and “-٣٫١٤١٥٩”.

> Stripping these from the input string in the parser function solves the issue, but I don't know if there are any potential side effects this could have. :thinking:

`Number.prototype.formatToParts` treats these bidi characters as `literal`s, so we can remove them rather generically. I don’t think this would be any more of a problem than removing the grouping symbols, since the characters are there purely to ensure correct layout; we’re parsing the characters in logical order rather than visual order.

modules/core/localizer.js
1ec5: > We can work around that behavior by setting the `maximumFractionDigits` option to 20 (the maximum value that doesn’t result in a `RangeError`).

I implemented this workaround.

test/spec/core/localizer.js
1ec5: > `Number.prototype.formatToParts` treats these bidi characters as `literal`s, so we can remove them rather generically.

I added code to strip out any `literal`s, including these bidirectional characters in the locales that use them.

1ec5:
I rebased this branch atop the latest develop branch. The conflicts were a bit tricky this time due to f573c374c29901d6a481b9efbd3c6daa2f312aab, which I think obsoleted some of the code review feedback above, but I went through the same cases as before and the functionality still seems to work correctly.

modules/ui/fields/roadspeed.js
```diff
@@ -119,16 +125,20 @@ export function uiFieldRoadspeed(field, context) {
...
+            if (rawValue && rawValue.indexOf('mph') >= 0) {
                 _isImperial = true;
-            } else if (value) {
+            } else if (rawValue) {
                 _isImperial = false;
             }
+
+            value = parseInt(value, 10);
+            if (isNaN(value)) value = rawValue;
+            value = formatFloat(value);
```
tyrasd: When the tag is not present, this crashes in the floatFormatter as the `number` is `undefined`.

Something like would fixe it:
```suggestion
            if (isNaN(value)) {
                value = rawValue;
            } else {
                value = formatFloat(value);
            }
```

modules/ui/fields/roadheight.js
```diff
@@ -156,26 +165,33 @@ export function uiFieldRoadheight(field, context) {
...
                 if (primaryValue !== null) {
-                    primaryValue = primaryValue[1];
+                    primaryValue = formatFloat(parseFloat(primaryValue[1]));
                 }
                 _isImperial = true;
             } else if (primaryValue) {
+                var rawValue = primaryValue;
+                primaryValue = parseFloat(rawValue);
+                if (isNaN(primaryValue)) primaryValue = rawValue;
+                primaryValue = formatFloat(primaryValue);
```
tyrasd: See comment in `roadspeed.js`.

But as currently no presets includes such a field by default, the crash is not triggerable with this field at the moment.

tyrasd (changes requested):
Let's get this merged!

The only small concern I currently still have with this one is what happens when users enter numbers in the "raw" (English) format, and the field is expecting localized input. This might occur regularly because mappers are likely to assume that numbers need to be entered in the same was as they are encoded in OSM tags.

For example, when entering `2.5` (for example in a road width field) when on a locale which uses `,` as the decimal delimiter (such as `vi` or `de` for example), this results in the field value to be set to `25`, which is not ideal.

I think we need to somehow special-case the situation when a "raw" decimal is entered.

<hr> 

There's also this bug when using the +/- buttons:

1ec5:
> I think we need to somehow special-case the situation when a "raw" decimal is entered.

This suggestion came up previously, but unfortunately I think it would be difficult to have it both ways, unless we assume the user won’t use their own language’s digit-grouping character for some reason: https://github.com/openstreetmap/iD/issues/3615#issuecomment-951565683. We could require the digit-grouping character to be used only to group three digits at a time, but we’d have to be careful not to disadvantage South Asian users who use crore and lakh – even in English.

#8238 introduced a monospaced font for raw OSM tags, so I suspect a new user would intuit that this field would accept localized numbers due to its variable-width font. Granted, it will require some getting used to for longtime users, but that problem will only grow over time. Do you think we should “prime” longstanding users by first styling these fields in monospaced font for a release before landing this feature?

> There's also this bug when using the +/- buttons

Haha wow! I tested this case previously, but I guess it crept back in at some point.

modules/ui/fields/input.js
```diff
@@ -153,7 +156,7 @@ export function uiFieldText(field, context) {
                         }
                         // make sure no extra decimals are introduced
                         const numDecimals = v.includes('.') ? v.split('.')[1].length : 0;
```
1ec5: This code added in f573c374c29901d6a481b9efbd3c6daa2f312aab is causing the funny behavior with the stepper buttons. `formatToParts` can identify the decimal part of the number more reliably; just need to extract that part from `floatParser`.

1ec5:
~~In English, entering <kbd>.</kbd> followed by a digit into a blank numeric field leaves just the digit in the field, since iD immediately reformats every keystroke.~~

Edit: Never mind, I accidentally switched to a locale in which I should’ve been entering <kbd>,</kbd>.

1ec5:
> For example, when entering `2.5` (for example in a road width field) when on a locale which uses `,` as the decimal delimiter (such as `vi` or `de` for example), this results in the field value to be set to `25`, which is not ideal.
> 
> I think we need to somehow special-case the situation when a "raw" decimal is entered.

@tyrasd, I have misgivings about sniffing the format from the user’s input – differing behavior based on opaque rules is a recipe for user confusion. Your example is the easy case: no language uses a digit grouping symbol to separate the tens from the ones. But an equally realistic example of “2.500” could go either way. In my opinion, the fields should be biased towards the user’s language, since the user can always enter raw values in the raw tag editor below.

Anyways, I came up with this fancy regular expression to detect likely “raw” numbers. It catches raw numbers like “1.23”, “0.123”, “1.2345”, and “1234.567” but not formatted numbers like “1.234” and “123.456”. It would be trivial to short-circuit the float parser when the input matches this regular expression:

```regex
/^-?(0\.\d*|\d*\.\d{0,2}(\d{4,})?|\d{4,}\.\d{3})$/
```

Unfortunately, this special case actually turns out to be counterproductive. Fields commit their values continuously on each keystroke rather than waiting for the user to finish typing, so a numeric field reformats the number on each keystroke. In the cases where there are two correct answers, we would instead choose a third, incorrect answer. For example, if a German speaker intends to enter “1.234” to tag `1234`, they’ll start typing <kbd>1</kbd><kbd>.</kbd><kbd>2</kbd>. As soon as they type <kbd>2</kbd>, “1.2” becomes “1,2” (`1.2`). Ultimately, the value becomes `1.234` (displayed as “1,234”) instead of the user’s intended `1234` (“1.234”).

Continuously validating a field causes other problems too, such as preventing the user from prepending more than one character at a time: #9233. But eliminating that behavior might resurrect a previously persistent issue that the field would fail to commit after the user finishes typing and navigates to another part of the UI.

/ref #1361

tyrasd (approved):

tyrasd:
FYI: I was still concerned that users might enter number in the "raw" format, which I addressed in the following way:

* implement a fix for #9233, which does not help very much here, as numbers can have different lengths in different formats
* use your regex to detect likely cases of raw format numbers
* not updating the input element while typing (a raw number which parses to the same _numeric_ value as the localized formatting of the tag value will prevent the input field from being overwritten with the localized formatting)

I think this will work quite well in almost all cases of numeric osm tag fields, as they are typically large integers (e.g. `population`), small integers (e.g. `capacity`) or decimal numbers (e.g. `width`) with few decimals (as an example: it is very unlikely that someone had actually measured the height of a "`8.849`" meters high peak to 3 millimeter resolution).

1ec5:
> implement a fix for #9233, which does not help very much here, as numbers can have different lengths in different formats

I’m very happy for this fix, though this PR wouldn’t have affected the `addr:housenumber`, `addr:unit`, and `addr:postcode` fields anyways, since they’re just text fields, not numeric fields.

> not updating the input element while typing (a raw number which parses to the same _numeric_ value as the localized formatting of the tag value will prevent the input field from being overwritten with the localized formatting)

I considered making this change too, but in the past, the downside was that some browsers would fail to fire the blur event in some cases, unexpectedly aborting the change: #1361. Are we confident that this is no longer a risk?

> it is very unlikely that someone had actually measured the height of a "`8.849`" meters high peak to 3 millimeter resolution

For a peak’s height, no, but `ele` is tagged on other kinds of features, and there are other numeric keys that could potentially be expressed to the thousands or thousandths. [My earlier example](https://www.openstreetmap.org/relation/307556) was just import cruft, but this is generic code that could potentially be reused for measurement-typed fields in the future.

I’m also concerned about users expecting to enter raw tag values in these fields. Though the changes you made _seem_ to work, I think we can reinforce the expected user behavior with a small style change: #9677.

tyrasd:
> Are we confident that this is no longer a risk?

As far as I can tell, what I changed should not interfere with this: The `change` method is still invoked on every keystroke `on('input', …)` (in addition to the blur event `on('blur', …)`). Only the "backfeeding"[^1] of the value is suppressed in the case where the _numeric_ value is the same as what is already displayed.

> […] measurement-typed fields in the future

Yeah, valid concern. A real example of a potentially ambiguous value would be something like https://www.openstreetmap.org/node/9591246259 (if the number field were to support entering numbers with a measurement unit in the future).

I'd say, let's definitely keep this potential issue in mind and reevaluate when such fields are introduced.  

> I think we can reinforce the expected user behavior with a small style change: #9677

:+1: 

[^1]: this is the part of the data binding which is responsible to update the content of the UI field when the tag value changes through some indirect means, e.g. the use of the raw tag editor, some validation routine, etc.

Issue #8750, opened 2021-10-13
Follow operation should prefer the focused way


1ec5:
The “follow” operation added in #8671 stops when encountering a node that’s shared among multiple features. iD already keeps track of the “focused” way for the purpose of reliably navigating back and forth among its vertices using the keyboard. It would be natural to extend this capability to the follow operation, so that the user doesn’t have to think as much about other connecting ways after choosing a way to follow. Invoking the follow operation would focus the way, and subsequent follow operations would choose that way from among the candidate ways.

https://github.com/openstreetmap/iD/blob/76990349bd2442a80b4c9bfb2440ea5c7a00a2d1/modules/modes/select.js#L150 https://github.com/openstreetmap/iD/blob/76990349bd2442a80b4c9bfb2440ea5c7a00a2d1/modules/modes/select.js#L48-L50

sudhanshu2:
@mbrzakovic can I take this issue?

Pho3b:
Hi and nice to meet you @mbrzakovic @1ec5 , i would like to try and work on this issue if possible.

thyrhone:
is it still possible to work on this?

1ec5:
Thanks for your interest! There’s a fix available in #8978; perhaps you could help review it?

sherryboxall:
Hello  @mbrzakovic @1ec5 ,

I would like to help work on this issue if possible. Thank you!

Dustpie:
Can I still help on this issue if possible? Thank you!

Gua00va:
Hey!! I want to work on this issue. I'm new here so I'm really excited to start contributing. Please assign it to me.

k-yle:
This issue has already been fixed in two conflicting PRs: #8978 and #9340


I'm guessing the suprising interest in this issue is because it has a _"good first issue"_ label. @tyrasd it might be good to add this label to some other issues like #9452, #9217, #9448, #9441, #9440, #9312, #9402, #9400, ...

1ec5:
Right, GitHub has a feature that suggests issues labeled as https://github.com/openstreetmap/iD/labels/good%20first%20issue to users, but it doesn’t consider whether there are outstanding PRs already linked to them.

DujaOSM:
Objection, your honor!
Implementing this will break workflow in a very common use case where we fill up the space with landuse polygons. Let me illustrate: 
[...]

PR #8773, opened 2021-10-28
Consider unsaved vertices when following a way

### LINKED ISSUE #8749: Unable to follow another way based on a new node
The “follow” operation added in #8671 only works if the current way is connected to enough consecutive _saved_ nodes on another way, but if one or more of those nodes is being added as part of the same changeset, the `needs_more_initial_nodes` message appears. (“This line can't follow a way because it isn't connected to enough consecutive points along a way.”)

To reproduce, draw a line with two points:

line

then draw another line that connects to it at two new points in the middle:

follow

Finally, press <kbd>F</kbd> to follow the original line.

I think this is because of these `hasEntity()` checks:

https://github.com/openstreetmap/iD/blob/76990349bd2442a80b4c9bfb2440ea5c7a00a2d1/modules/behavior/draw_way.js#L443

1ec5:
The follow operation now looks for the last two nodes and their parents in the full history graph, which includes unsaved nodes. It also includes the way being drawn, so that way needs to be filtered out.

As a side effect, if the current way has already begun to double back on itself, then the follow operation avoids the current way in favor of some other way. I’m unsure if this side effect will persist if the operation is modified to prefer following the focused way (#8750), but either behavior is reasonable in my opinion.

https://user-images.githubusercontent.com/1231218/139174761-93b3c514-4832-4826-80c4-012367fb0e32.mov

Am I overlooking any adverse effects of looking up the nodes in the full graph?

Fixes #8749.

### DIFF
--- a/modules/behavior/draw_way.js
+++ b/modules/behavior/draw_way.js
@@ -440,7 +440,10 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
 
             const [secondLastNodeId, lastNodeId] = _origWay.nodes.slice(isDrawingArea ? -3 : -2);
 
-            if (!lastNodeId || !secondLastNodeId || !startGraph.hasEntity(lastNodeId) || !startGraph.hasEntity(secondLastNodeId)) {
+            // Unlike startGraph, the full history graph may contain unsaved vertices to follow.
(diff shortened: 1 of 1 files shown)


mbrzakovic (approved):
Looks good. Thanks

PR #8776, opened 2021-10-28
Sync up RapiD and iD esbuild config

bhousel:
This PR is a best attempt to combine
https://github.com/facebookincubator/RapiD/pull/246 , https://github.com/facebookincubator/RapiD/commit/f415cce02e6ab221c4e56042f280bfc0ed2d159a

... into a branch that _should_ mostly merge into @mbrzakovic's branch on  #8774 

Some things that are different are:
* iD is still not switched over to `type: module` so the config scripts need a `.mjs` extention for node to treat them as ES6 
* I didn't include the commands to enter watch mode.  We don't have those yet on our end, but it might be useful
* Babel config is confusing and I don't completely understand it.  All the tests pass in Phantom, so I assumed we were covered, but from what you said on #8774 it sounds like IE11 does not work (I haven't tested it!).  This is probably because I only added core-js-bundle to the test `index.html` 🤔 
* oh I also git removed the generated `en.min.json` file, because it is annoying that every time I merge anything that generated file has a conflict.


mbrzakovic:
> This PR is a best attempt to combine [facebookincubator#246](https://github.com/facebookincubator/RapiD/pull/246) , [facebookincubator@f415cce](https://github.com/facebookincubator/RapiD/commit/f415cce02e6ab221c4e56042f280bfc0ed2d159a)
> 
> ... into a branch that _should_ mostly merge into @mbrzakovic's branch on #8774
> 
> Some things that are different are:
> 
> * iD is still not switched over to `type: module` so the config scripts need a `.mjs` extention for node to treat them as ES6

_mbrzakovic: I am fine with this for now._
> * I didn't include the commands to enter watch mode.  We don't have those yet on our end, but it might be useful

_mbrzakovic: I think watch mode really speeds up development. I propose we merge this PR as-is, then I add it in devbuild_on_esbuild and after merging you can fetch in upstream. Sounds good?_

> * Babel config is confusing and I don't completely understand it.  All the tests pass in Phantom, so I assumed we were covered, but from what you said on [Devbuild on esbuild #8774](https://github.com/openstreetmap/iD/pull/8774) it sounds like IE11 does not work (I haven't tested it!).  This is probably because I only added core-js-bundle to the test `index.html` 🤔

_mbrzakovic: Yes, please look into this.  If we can't support IE users I'd rather we don't switch to esbuild for legacy build just yet._
> * oh I also git removed the generated `en.min.json` file, because it is annoying that every time I merge anything that generated file has a conflict.

_mbrzakovic: Sounds good._

babel.config.json
```diff
@@ -2,11 +2,22 @@
...
-        "modules": false
+        "modules": false,
+        "targets": {
+          "chrome": "58",
+          "ie": 11
+        }
       }
-    ]
+    ],
+    "@babel/preset-react"
```
mbrzakovic: iD does not use react, so this line seems redundant for now.

dist/index.html
```diff
@@ -32,8 +32,7 @@
       newScript.onerror = checkScript;
 
       var isIE11 = !!(navigator.userAgent.match(/Trident/) && !navigator.userAgent.match(/MSIE/));
-      // currently all minified files are legacy ES5 because of uglifyJS
```
mbrzakovic, line 35: Could you please help me understand this part by explaining why did we have this constraint (uglifyJS, es5)?

package.json
```diff
@@ -35,11 +35,10 @@
     "imagery": "node scripts/update_imagery.js",
     "lint": "eslint scripts test/spec modules",
     "lint:fix": "eslint scripts test/spec modules --fix",
-    "start": "npm-run-all -s build:dev start:server",
-    "start:extended": "npm-run-all -s build start:server",
+    "start": "run-s build:legacy start:server",
```
mbrzakovic: Generally in node web projects, the 'start' command is expected to build the .js that developers are most likely modifying and start the server in order to display the change.
Since the majority of developers use non-IE browsers for this, currently this setup could cause confusion by running the legacy build and not displaying the change for non-IE browsers.

I think it would be more clear and in sync with common practices that we drop the quickstart, and set start to do build:dev.

mbrzakovic (changes requested):

babel.config.json
bhousel: I agree, I'll remove it

dist/index.html
bhousel: I think it was just that UglifyJS never got support for full ES6 syntax, so people use other tools to minify JS nowadays.  https://github.com/mishoo/UglifyJS/issues/3443

package.json
bhousel: Sounds good to me to make `npm run start` serve a modern build, and `npm run quickstart` can go away.. 
We've all gotten into the habit of just doing `npm run quickstart`.  I don't remember why it was set up like that.

bhousel:
> > * iD is still not switched over to `type: module` so the config scripts need a `.mjs` extention for node to treat them as ES6
> 
> _mbrzakovic: I am fine with this for now._

Cool, yes it is probably a few hours work to add `type: module` and worth switching at some point, but not essential. 
Mainly all the `.js` scripts need to replace `require` with `import`.
 
> > * I didn't include the commands to enter watch mode.  We don't have those yet on our end, but it might be useful
> 
> _mbrzakovic: I think watch mode really speeds up development. I propose we merge this PR as-is, then I add it in devbuild_on_esbuild and after merging you can fetch in upstream. Sounds good?_

Yes 👍 

> > * Babel config is confusing and I don't completely understand it.  All the tests pass in Phantom, so I assumed we were covered, but from what you said on [Devbuild on esbuild #8774](https://github.com/openstreetmap/iD/pull/8774) it sounds like IE11 does not work (I haven't tested it!).  This is probably because I only added core-js-bundle to the test `index.html` 🤔
> 
> _mbrzakovic: Yes, please look into this. If we can't support IE users I'd rather we don't switch to esbuild for legacy build just yet._

I'll figure it out - I'm sure we can continue to support IE users.  My main priority before was keeping the PhantomJS tests running.

mbrzakovic:
Wait for: https://github.com/openstreetmap/iD/pull/8774#issuecomment-961103284

mbrzakovic:
Reverted only last commit which improves the modern/legacy build setup.

Reasoning: With [Drop IE11 support](https://github.com/openstreetmap/iD/issues/8811) planned, improvements in supporting legacy build are no longer needed.

mbrzakovic (approved):

PR #8782, opened 2021-11-01
add preview for colour input

### LINKED ISSUE #8289: 'false' when hovering over the address field
What is it?
When hovering over the address field which is filled in, there is "false".
false

### LINKED ISSUE #8894: Address fields show `false` as title
### URL

https://www.openstreetmap.org/edit#map=20/42.72185/23.27184

1. Click on the building that's in view
2. Hover over any of the address fields
3. The `title` of these elements is `false`

image
image

Firefox

k-yle:
This PR adds a preview box next to colour inputs. This should help prevent typos, and make it easier for people who aren't familiar with hex codes.

roof colour

---

also closes #8289 and closes #8894 , see https://github.com/openstreetmap/iD/pull/8782#discussion_r739928290

### DIFF
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -288,7 +288,7 @@ export function uiFieldAddress(field, context) {
             })
             .attr('title', function(subfield) {
                 var val = tags[field.key + ':' + subfield.id];
-                return val && Array.isArray(val) && val.filter(Boolean).join('\n');
+                return (val && Array.isArray(val)) ? val.filter(Boolean).join('\n') : undefined;
             })
             .classed('mixed', function(subfield) {
                 return Array.isArray(tags[field.key + ':' + subfield.id]);
--- a/modules/ui/fields/input.js
+++ b/modules/ui/fields/input.js
@@ -21,6 +21,7 @@ export function uiFieldText(field, context) {
     var dispatch = d3_dispatch('change');
     var input = d3_select(null);
     var outlinkButton = d3_select(null);
+    var wrap = d3_select(null);
     var _entityIDs = [];
     var _tags;
     var _phoneFormats = {};
@@ -62,7 +63,7 @@ export function uiFieldText(field, context) {
         calcLocked();
         var isLocked = field.locked();
 
-        var wrap = selection.selectAll('.form-field-input-wrap')
+        wrap = selection.selectAll('.form-field-input-wrap')
             .data([0]);
 
         wrap = wrap.enter()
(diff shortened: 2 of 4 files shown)


modules/ui/fields/address.js
```diff
@@ -288,7 +288,7 @@ export function uiFieldAddress(field, context) {
             })
             .attr('title', function(subfield) {
                 var val = tags[field.key + ':' + subfield.id];
-                return val && Array.isArray(val) && val.filter(Boolean).join('\n');
+                return (val && Array.isArray(val)) ? val.filter(Boolean).join('\n') : undefined;
```
k-yle, line 291: I fixed a minor unrelated bug here, if this condition is false, then `title="false"` is added to the DOM, which resulted in a strange tooltip:

tooltip

modules/ui/fields/input.js
```diff
@@ -167,9 +168,29 @@ export function uiFieldText(field, context) {
                     if (value) window.open(value, '_blank');
                 })
                 .merge(outlinkButton);
+        } else if (field.key.includes('colour')) {
```
k-yle, line 171: this works for the most common fields (`colour=*`, `roof:colour=*`, `building:colour=*`), but I think we should introduce a new field type to https://github.com/openstreetmap/id-tagging-schema, since this logic erroneously matches tags like `seamark:topmark:colour_pattern=*`

modules/ui/fields/input.js
```diff
@@ -167,9 +168,29 @@ export function uiFieldText(field, context) {
...
+            updateColourPreview();
         }
     }
 
+    function updateColourPreview() {
+        wrap.selectAll('.foreign-id-permalink')
+            .remove();
+
+        outlinkButton = wrap.selectAll('.foreign-id-permalink')
+            .data([utilGetSetValue(input)], d => d);
```
1ec5: Does this do any validation? [This documentation](https://wiki.openstreetmap.org/wiki/Key:colour) allows `#XXXXXX` and CSS color keywords, but other [CSS color](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) formats like `rgb()`, `hsl()`, and `rgba()` are excluded from that definition. A working color well would mislead the user into thinking an invalid value is valid.

modules/ui/fields/input.js
```diff
@@ -167,9 +168,29 @@ export function uiFieldText(field, context) {
...
+
+        outlinkButton
+            .enter()
+            .append('div')
+            .attr('class', 'form-field-button foreign-id-permalink colour-preview')
+            .append('div')
+            .style('background-color', d => d)
```
1ec5, line 198: This isn’t far from a [color input field](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color), which would make it easier for the user to choose an sRGB value. However, some browsers render this field as a color well and don’t fall back gracefully when the value is a color keyword.

modules/ui/fields/input.js
k-yle: good point, I've added validation based on what [openstreetmap-website does](https://github.com/openstreetmap/openstreetmap-website/blob/08e2a0/app/helpers/browse_tags_helper.rb#L177-L189).

They also check the colours against a list of all possible colour names, but we don't need to do that since setting an invalid background-colour value will be ignored by the browser

modules/ui/fields/input.js
k-yle: I experimented with [a colour picker](https://github.com/openstreetmap/iD/compare/3a3d97...54d922), but there are some issues:
- as you said, the colour picker doesn't work when value is a color keyword
- the default `<input type="color" />` element shows black for invalid colours or an empty input field, which would be quite confusing 
- the default `<input type="color" />` element doesn't have a submit button, so iD's text field would only update when the user clicks out of the colour picker. This would be confusing for some people

So a think the simple colour preview is the better option

modules/ui/fields/input.js
tyrasd: Let's continue this in https://github.com/openstreetmap/id-tagging-schema/issues/284

tyrasd:
Thanks @k-yle!

FYI: Before merging this PR, I've added a hidden color input field to be able to let the user choose a color using the browser's native colour picker dialog after clicking on the colour preview box.

Issue #8784, opened 2021-11-01
Improve history panel


boothym:
### Description

I recently discovered that the "OSM" link in the history panel actually links to the history of an object. However this is not very clear, so I would like to suggest an edit to the text to improve this.

At the first pink dot, it should say "OSM History" and at the second pink dot, it should say "OSM Profile". Unfortunately I wasn't able to edit the text in browser!
![osmh](https://user-images.githubusercontent.com/16965772/139749648-f8125e48-6824-45f6-ab1f-cafefd33ddd1.PNG)

mbrzakovic:
I agree that these link informations can be more user-friendly.

1ec5:
On the first row of links, each link opens the feature’s history but using different tools. On the second row, each tool opens the user’s profile but using different tools. It would be less confusing to have an unlinked “History:” and “Profile:” label right before the list of links on each row.

mbrzakovic:
What @1ec5 recommended makes sense, although I do have to point out that the tab name is 'History' so the additional History label can seem somewhat redundant. Perhaps in this case a better wording would be: 'View X on/via: '

I would recommend that a better alignment can go a long way in this case, and with the help of simple tooltips it could achieve making more obvious which item each link describes.

e.g. (sketch):
![image](https://user-images.githubusercontent.com/78906108/140306710-49fbc689-8b57-46f7-9ac6-153298a6de1c.png)

boothym:
The reason I opened this issue is because I associated the first "OSM" with just opening the node/way/relation on osm.org - like the "view on osm.org" one in the bottom left already does. It didn't occur to me that would go direct to the history, so anything that makes it clearer is good.

>It would be less confusing to have an unlinked “History:” and “Profile:” label right before the list of links on each row.

I would probably be fine with that.

LaoshuBaby:
> 
> e.g. (sketch): ![image](https://user-images.githubusercontent.com/78906108/140306710-49fbc689-8b57-46f7-9ac6-153298a6de1c.png)

Looks interesting! ♥👍

And this new 2 column can decrease the height of history panel to avoid this kind of hidden: https://github.com/openstreetmap/iD/issues/8699

(And I also have a question, why not add a OSMCha link to the latest editor.)

(Maybe deep history can added to Feature too)

tyrasd:
#8834 improves the ambiguous `OSM` link texts.

PR #8788, opened 2021-11-02
Change history panel link names

johnsciutto:
- "PeWu" has been changed to "History"
- "HDYC" has been changed to "Profile"

Addresses [this issue](https://github.com/openstreetmap/iD/issues/8784#issue-1041696152) 🙂

### DIFF
--- a/modules/ui/panels/history.js
+++ b/modules/ui/panels/history.js
@@ -49,7 +49,7 @@ export function uiPanelHistory(context) {
             .attr('href', 'https://hdyc.neis-one.org/?' + userName)
             .attr('target', '_blank')
             .attr('tabindex', -1)
-            .html('HDYC');
+            .html('Profile');
     }
 
 
@@ -200,7 +200,7 @@ export function uiPanelHistory(context) {
             .attr('href', 'https://pewu.github.io/osm-history/#/' + entity.type + '/' + entity.osmId())
             .attr('target', '_blank')
             .attr('tabindex', -1)
-            .html('PeWu');
+            .html('History');
 
         var list = selection
             .append('ul');


modules/ui/panels/history.js
```diff
@@ -200,7 +200,7 @@ export function uiPanelHistory(context) {
             .attr('href', 'https://pewu.github.io/osm-history/#/' + entity.type + '/' + entity.osmId())
             .attr('target', '_blank')
             .attr('tabindex', -1)
-            .html('PeWu');
+            .html('History');
```
1ec5, line 203: The link text is no longer the name of the tool itself, so now it’s more important to localize it into the user’s language. To make the link text localizable, call the `t()` method, passing in a string ID like `info_panels.history.pewu_link_text`. Then add an entry around here called `pewu_link_text` and set it to the string `History`:

https://github.com/openstreetmap/iD/blob/3e7fd6e0330a14319206c9c9eb47afb3d97ca8d9/data/core.yaml#L653

modules/ui/panels/history.js
```diff
@@ -49,7 +49,7 @@ export function uiPanelHistory(context) {
             .attr('href', 'https://hdyc.neis-one.org/?' + userName)
             .attr('target', '_blank')
             .attr('tabindex', -1)
-            .html('HDYC');
+            .html('Profile');
```
1ec5, line 52: As I belatedly realized in https://github.com/openstreetmap/iD/issues/8784#issuecomment-958186532, on each row, each link displays similar information using different tools. So there’s already a different profile link directly to the left of this link. To avoid confusion, we can call `links.html(t('info_panels.history.profile'))` a few lines above, before appending the first link, then leave this link as it was.

modules/ui/panels/history.js
bhousel: Can we rethink this? 
I actually do think it's better to have these link texts to be the name of the tool itself.  
These information panels are expected to be used by more skilled users.

johnsciutto: Makes sense 👍🏻. I'll close this PR

johnsciutto:
Closing this PR.
As this panel is meant for more experienced users, it makes sense to have the links be the name of the tools themselves, as mentioned by @bhousel.

PR #8791, opened 2021-11-03
fix clipped tooltips on controls bar

### LINKED ISSUE #8781: No popup tooltip on right menu
### URL

_No response_

Since last update, tooltips on the right menu are not showing. They had a brief description of the menu and a key shortcut, but that's no longer there when hovering over the icon.

_No response_

Firefox

tyrasd:
Fixes #8781. My solution was to add a full-width wrapper div (`map-controls-wrap`) which can scroll the content without clipping the area left of the buttons where tooltips would be displayed.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2958,11 +2958,10 @@ img.tag-reference-wiki-image {
     left: 0;
     right: 0;
     height: 100%;
+    z-index: 100;
     display: block;
     overflow-x: hidden;
     overflow-y: auto;
-    -ms-overflow-style: none;
-    scrollbar-width: none;
     pointer-events: none;
 }
 .map-controls-wrap::-webkit-scrollbar {
@@ -2973,7 +2972,6 @@ img.tag-reference-wiki-image {
     top: 0;
     width: 40px;
     position: absolute;
-    z-index: 100;
     bottom: 0;
     display: flex;
     flex-direction: column;
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2963,6 +2963,8 @@ img.tag-reference-wiki-image {
     overflow-x: hidden;
     overflow-y: auto;
     pointer-events: none;
+    -ms-overflow-style: none;
+    scrollbar-width: none;
 }
 .map-controls-wrap::-webkit-scrollbar {
     display: none;
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2953,6 +2953,21 @@ img.tag-reference-wiki-image {
 
 /* Map Controls
 ------------------------------------------------------- */
+.map-controls-wrap {
+    position: absolute;
+    left: 0;
+    right: 0;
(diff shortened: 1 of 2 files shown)


tyrasd:
PS: @mbrzakovic I'm not sure what the code here: https://github.com/openstreetmap/iD/blob/7e31120ec68aa161898338729e7b1d77af051748/modules/ui/init.js#L204-L208 is doing exactly. Perhaps it also needs to be adjusted?!

mbrzakovic:
@tyrasd your approach looks great, its simple and fixes the abovementioned conflict.

This code piece is responsible for moving the control items (in this case up-down) when mouse wheel event is generated giving the impression of scrolling. Please look into [this video](https://github.com/openstreetmap/iD/pull/8685#issue-993255229). 
> 
> https://github.com/openstreetmap/iD/blob/7e31120ec68aa161898338729e7b1d77af051748/modules/ui/init.js#L204-L208
> 
With the current code version this event is now attached to map-control, which does not have a capability to scroll (overflow is visible by default). I've pushed a minor adjustment that fixes the problem by attaching the event wheeling to the map-controls-wrap.
Additionally, I've observed that when some control pane is opened up (e.g. Background) the map controls are now behind map-panes. Nothing a minor z-index change can't fix :)

Please double check the new code version, and feel free to merge the changes.

mbrzakovic (approved):

css/80_app.css
```diff
@@ -2958,11 +2958,10 @@ img.tag-reference-wiki-image {
     left: 0;
     right: 0;
     height: 100%;
+    z-index: 100;
     display: block;
     overflow-x: hidden;
     overflow-y: auto;
-    -ms-overflow-style: none;
-    scrollbar-width: none;
```
tyrasd: Why did you get rid of these two lines? My intention was to emulate what is done [below](https://github.com/openstreetmap/iD/pull/8791/commits/71728d9415fa132cdd75e6b0afb70dbd1a9d3b3b#diff-3f86fb5cebd04342575e756d440d18a0b6cfb131a3c26d7897ff1793f8b6d30bR2967-R2969) with the `::-webkit-scrollbar` also for [Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width)/IE.

tyrasd:
> This code piece is responsible for moving the control items (in this case up-down) when mouse wheel event is generated giving the impression of scrolling.

Interestingly, the mouse wheel scrolling does work on Firefox also without the javascript wheel event handler. That's why I was confused. The code does seem to be necessary in Chrome. Not sure which browser is actually doing the right thing here. :wink:

css/80_app.css
mbrzakovic: Good catch. Naively thought these were redundant. Reverted them back. Thanks.

mbrzakovic (approved):

PR #8792, opened 2021-11-05
improve rendering of turn-restriction boxes in narrow sidebars

tyrasd:
With narrow sidebars, the turn-restriction min-map view glitches out (by being much more zoomed out than can be considered usable).

Before: 

This tweaks it such that even for the slimmest sidebar the rendering is still acceptable (i.e. no turn arrows to be rendered completely outside of the viewport) and at the same time allows generous margins to be applied at wider sidebars.

After: 

This also fixes a minor bug where the mini-map is not properly alligned vertically.

(the issue was originally reported by @jleedev)

### DIFF
--- a/modules/ui/fields/restrictions.js
+++ b/modules/ui/fields/restrictions.js
@@ -223,22 +223,24 @@ export function uiFieldRestrictions(field, context) {
             extent._extend(_intersection.vertices[i].extent());
         }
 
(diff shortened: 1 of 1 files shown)


mbrzakovic (approved):
Tested this locally with various sizes and confirming that turn-restriction min-map is behaving correctly now.
Well done

PR #8800, opened 2021-11-09
clarify "disconnected_way" validation text (fixes #8760)

### LINKED ISSUE #8760: validator could consider a `public_transport=platform` way as routable
Hi all

This is clearly related to the discussion in #6409, but I haven't followed the discussion in detail and I am not aware of what made the developers change their mind on the suggested addition of explicit `highway=footway` tags by the validator.

In any case, I came across the `disconnected_way` issue in [this object](https://www.openstreetmap.org/way/683341103), because its connection to the whole network goes through [a way tagged as `public_transport=platform`](https://www.openstreetmap.org/way/683341106).

I'm wondering if the validator should consider the way tagged as `public_transport=platform` as a routable way. It seems to me it would be quite common to have detailed public transport stations in which, to cross from one side to the other of the building, you would have to walk on sections of platforms. The debate might be more relevant for areas, but isn't this straight-forward for a way?

And just taking the opportunity to thank you all for the great work on the editor :+1:

stragu:
As discussed in issue #8760, clarify that the ways that are considered routable by the validator need to be well represented among major routing software.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1766,9 +1766,9 @@ en:
       tip: "Find unroutable roads, paths, and ferry routes"
       routable:
         message:
-          one: "{highway} is disconnected from other roads and paths"
-          other: "{count} routable features are connected only to each other"
-        reference: "All roads, paths, and ferry routes should connect to form a single routing network."
+          one: "{highway} is disconnected from other commonly routable features"
+          other: "{count} commonly routable features are connected only to each other"
+        reference: "Ways consistently considered routable by data users (including roads, paths and ferry routes) should connect to form a single routing network."
     fixme_tag:
       message: '{feature} has a "Fix Me" request'
       reference: 'A "fixme" tag indicates that a mapper has requested help with a feature.'
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1766,9 +1766,9 @@ en:
       tip: "Find unroutable roads, paths, and ferry routes"
       routable:
         message:
-          one: "{highway} is disconnected from other commonly routable features"
-          other: "{count} commonly routable features are connected only to each other"
-        reference: "Ways consistently considered routable by data users (including roads, paths and ferry routes) should connect to form a single routing network."
+          one: "{highway} is disconnected from other roads and paths"
+          other: "{count} routable features are connected only to each other"
+        reference: "All roads, paths, and ferry routes should connect to form a single routing network. This feature (or small collection of features) is either disconnected from the routing network or only connected to it via a feature that few applications consider routable."
     fixme_tag:
       message: '{feature} has a "Fix Me" request'
       reference: 'A "fixme" tag indicates that a mapper has requested help with a feature.'


data/core.yaml
```diff
@@ -1766,9 +1766,9 @@ en:
       tip: "Find unroutable roads, paths, and ferry routes"
       routable:
         message:
-          one: "{highway} is disconnected from other roads and paths"
-          other: "{count} routable features are connected only to each other"
-        reference: "All roads, paths, and ferry routes should connect to form a single routing network."
+          one: "{highway} is disconnected from other commonly routable features"
+          other: "{count} commonly routable features are connected only to each other"
+        reference: "Ways consistently considered routable by data users (including roads, paths and ferry routes) should connect to form a single routing network."
```
tyrasd: IMHO the validation's message itself should be short and easily comprehensible. It should not use overly technical language just for the sake of covering more possible edge cases. The reference text is meant to explain what is actually going on in more detail, so that's where I would adjust this. What about something like the following?

```suggestion
          one: "{highway} is disconnected from other roads and paths"
          other: "{count} routable features are connected only to each other"
        reference: "All roads, paths, and ferry routes should connect to form a single routing network. This validation takes only features into account which are commonly and unambiguously considered to be routable."
```

tyrasd (changes requested):

data/core.yaml
1ec5: It would be nice to focus on the selected feature(s) instead of the validation rule:

> All roads, paths, and ferry routes should connect to form a single routing network. This way is not (These ways are not) connected, or it is (they are) only connected via a feature that few applications consider routable.

tyrasd: @1ec5 sounds good. What about this:
```suggestion
          one: "{highway} is disconnected from other roads and paths"
          other: "{count} routable features are connected only to each other"
        reference: "All roads, paths, and ferry routes should connect to form a single routing network. This feature (or small collection of features) is either disconnected from the routing network or only connected to it via a feature that few applications consider routable."
```

1ec5: Yes, that’s good. Final nitpick is that “only few” should be either “few” or “only a few”. By the way, if you split the `reference` string into `one` and `other` strings, I think it’ll pluralize automatically.

tyrasd: Thanks for the grammar correction! :relaxed: I've amended the suggestion above.

> it’ll pluralize automatically.

I think that's only the case for the `message` part of the validation, since the `reference` part is inserted manually by the validator code, where there is currently no distinction between the `one` or `many` case: https://github.com/openstreetmap/iD/blob/v2.20.2/modules/validations/disconnected_way.js#L82-L89 – I guess that could be improved at some point, but it's not too important right now and for this PR.

tyrasd (approved):

stragu:
Thank you both for finding the perfect wording, @tyrasd and @1ec5 ! :+1:

PR #8803, opened 2021-11-12
explicit coloring of preset categories

### LINKED ISSUE #8799: Improve preset categories colors
### Description

It seems logic to me that the menu colours should be in accordance with what elements are depicted. As such the Menu folder for Waterways should be Blue, and Natural features should be green or brown, but not Blue wich is traditionally associated with Water Features that currently has No colours.

### Screenshots

Screenshot 2021-11-09 18 58 17

tyrasd:
fixes #8799

#### line presets

 :arrow_right: 

#### area presets
 :arrow_right:


mbrzakovic (approved):
Looks good. Not sure if this makes some other stylings (mostly from: 25_areas.css) redundant. 🤔

mbrzakovic (approved):
Looks good. Not sure if this makes some other stylings (mostly from: 25_areas.css) redundant. 🤔

PR #8804, opened 2021-11-12
introduce allowlist for disconnectable relation types

### LINKED ISSUE #8771: `associatedStreet` relation does not mean its members shouldn't be disconnected
### URL

https://www.openstreetmap.org/way/887735280

1. Select way [887735280](https://www.openstreetmap.org/way/887735280) and its node joining it to Gladstone rd (way [991574385](https://www.openstreetmap.org/way/991574385))
2. Right-click the node to see that it is not possible to disconnect it "because it connects members of a relation"

This kind of relation is not related to routable networks or a specific order of members, and therefore iD should let contributors disconnect features that are members of the same relation.

(As to the reason I am editing this: I surveyed the area and the two roads do not connect anymore and now have a physical barrier between them).

_No response_

Firefox

tyrasd:
This skips relation types which are only _grouping_ entities together and don't enforce connectivity or order which would be potentially broken by the disconnect operation.

Current list of disconnectable relation types:

* `site`
* `assosciatedStreet`

did I forget other relation types which have this property?

closes #8771

### DIFF
--- a/modules/actions/disconnect.js
+++ b/modules/actions/disconnect.js
@@ -18,6 +18,10 @@ import { osmNode } from '../osm/node';
 export function actionDisconnect(nodeId, newNodeId) {
     var wayIds;
 
+    var disconnectableRelationTypes = {
+        'site': true,
+        'assosciatedStreet': true
+    };
 
     var action = function(graph) {
         var node = graph.entity(nodeId);
@@ -88,7 +92,9 @@ export function actionDisconnect(nodeId, newNodeId) {
 
         parentWays.forEach(function(way) {
             var relations = graph.parentRelations(way);
-            relations.forEach(function(relation) {
+            relations
+            .filter(relation => !disconnectableRelationTypes[relation.tags.type])
+            .forEach(function(relation) {
                 if (relation.id in seenRelationIds) {
                     if (wayIds) {
                         if (wayIds.indexOf(way.id) !== -1 ||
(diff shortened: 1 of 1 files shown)


modules/actions/disconnect.js
```diff
@@ -18,6 +18,10 @@ import { osmNode } from '../osm/node';
 export function actionDisconnect(nodeId, newNodeId) {
     var wayIds;
 
+    var disconnectableRelationTypes = {
+        'site': true,
+        'assosciatedStreet': true
```
1ec5:
```suggestion
        'associatedStreet': true
```

modules/actions/disconnect.js
```diff
@@ -18,6 +18,10 @@ import { osmNode } from '../osm/node';
...
+    };
```
1ec5, line 25: Other potentially disconnectable relation types:

* [`enforcement`](https://wiki.openstreetmap.org/wiki/Relation:enforcement) – specifically, the `device` member
* [`building`](https://wiki.openstreetmap.org/wiki/Relation:building) – not about routing, but it should probably be possible to disconnect a building outline from something unrelated like an adjoining footway

tyrasd: > it should probably be possible to disconnect a building outline from something unrelated like an adjoining footway

yes, but this is not prevented by the disconnect action: only members of the _same_ relation are kept together, so that the existing connectivity within those relations are always conserved. Objects which are part of different relations (or situations where only one object is member of any relations) are (and were always) allowed to be disconnected.

> `enforcement`

:+1: this sound like a good addition in order to be able to unglue accidentally snapped enforcement devices from their streets

mbrzakovic (approved):

tyrasd:
merged as c85456438987715e9d243e519200bb4b9134087e

PR #8808, opened 2021-11-15
ways with more than 2000 nodes: prevent lockup and provide validator-fix

### LINKED ISSUE #7381: The editor locks up after error "You tried to add X nodes to way Y, however only 2000 are allowed."
After editing normally and trying to define a big reservoir you try to save and upload your changes.

Then, suddenly, the server replies with this:
> `You tried to add 2159 nodes to way 516982174, however only 2000 are allowed`

You, wearily click OK in the server-side error popup, then you try to fix it, try to send it again, and an authentication window pops up, and clicking next ends up in a 404 error. After that the iD interface locks up in drag mode and you can't click any buttons.

Reloading the page means losing a good chunk of the work, because half of the changes disappear for no reason. I think marking which changes are actually cached locally even after a page reload would be very useful.

---

I understand the need to limit the amount of vertices, but because you don't know beforehand if you've surpassed the limit or not it feels like playing Russian roulette with yourself. There's only one shot at trying to upload, so I sometimes add detail in some areas and purge in others that don't need it just to make sure I don't get to the fateful 2000th.

It would be great to either detect client-side that we've gone over the node limit and error out, giving the possibility of selecting the culprit and avoiding kamikaze uploads, or at least handling server-side errors correctly post-upload. Ideally both.

I use iD for almost all of my OSM changes and it's often more than enough, even as a power user.

Thanks in advance, let me know if you need more information.

tyrasd:
closes #7381:

1. prevents the locked-up state after the API returns a HTTP 400 status code after uploading a way with too many nodes
2. adds a validator which checks for such objects
3. the validator provides an easy fix by splitting the affected object into small enough pieces:

image

Maybe the exact wording could still be improved?

### DIFF
--- a/modules/validations/osm_api_limits.js
+++ b/modules/validations/osm_api_limits.js
@@ -4,15 +4,18 @@ import { operationSplit } from '../operations/split';
 
 export function validationOsmApiLimits(context) {
     var type = 'osm_api_limits';
-    var maxWayNodes = context.connection().maxWayNodes();
 
-    var validation = function checkOsmapiLimits(entity, graph) {
+    var validation = function checkOsmApiLimits(entity) {
         var issues = [];
+        var osm = context.connection();
+        if (!osm) return issues; // cannot check if there is no connection to the osm api, e.g. during unit tests
+        var maxWayNodes = osm.maxWayNodes();
 
         if (entity.type === 'way') {
             if (entity.nodes.length > maxWayNodes) {
                 issues.push(new validationIssue({
                     type: type,
+                    subtype: 'exceededMaxWayNodes',
                     severity: 'error',
                     message: function() {
                         return t.html('issues.osm_api_limits.max_way_nodes.message');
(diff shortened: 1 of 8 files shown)


modules/validations/osm_api_limits.js
```diff
@@ -0,0 +1,62 @@
...
+            icon: 'iD-operation-split',
+            title: t.html('issues.fix.split_way.title'),
+            entityIds: this.entityIds,
+            onClick: function(context) {
+                var entityId = this.entityIds[0];
+                var entity = context.graph().entities[entityId];
+                var numberOfParts = Math.ceil(entity.nodes.length / maxWayNodes);
+                var splitVertices = [...Array(numberOfParts - 1)].map((_, i) =>
+                    entity.nodes[Math.floor(entity.nodes.length * (i + 1) / numberOfParts)]);
+                var operation = operationSplit(context, splitVertices.concat(entityId));
```
bhousel: I'd try to split at junctions if you can, rather than where the math happens to divide the way.

PR #8817, opened 2021-11-18
reduce direct html injection, add escaping where necessary

### LINKED ISSUE #8813: OSM username isn't escaped properly
### URL

_No response_

1. Set your OSM username to something containing an HTML tag, like `Test <hr> McTesty`
2. Make some changes in iD and press upload
3. Notice that the username has been put directly into the html, instead of being escaped.

I realise this won't come up in many circumstances, but the unusual name on my development machine is a side-effect of some XSS testing for website that I was doing a few months ago. Although unlikely to be exploitable (since the username within iD is only shown to the user, not to anyone else, and due to limitations on the characters permitted in the OSM username) it's probably worth addressing anyway, and might indicate other places where "user supplied" information from the API isn't being escaped.

bitmap

Note how the username in the rails port (circled green, top right) is properly escaped, but within iD it is not escaped (circled red, bottom left)

Firefox

tyrasd:
fixes #8813 by:
* partially reverting 32f8274929c9cdd97fcdd2aff332586c0ff140f2 (#7998) so that `html` is only used when necessary and safe.
  * this is now almost exclusively the case when html comes from an internal source like `t.html(…)`.
* `t.html` now escapes all `replacement` strings, except when they are explicitly marked as `html` code.
* replacing many unnecessary calls to `d3_selection.html` with a new method that directly and safely appends the localized text to the dom: ~`selection.html(t.html("…"))` → `selection.call(t.append("…"))`
* adding a missing `encodeURIComponent` thanks @1ec5  
* use a more precise "language code" for the [no-language-code specified case](https://en.wikipedia.org/wiki/ISO_639-2#Special_situations) (unrelated to the fix, but also included because it touches the same code), thanks @1ec5 

todo:
* [x] add to changelog
* [x] add unit tests for new `t.append` method


data/core.yaml
```diff
@@ -1000,7 +1000,7 @@ en:
       location: 'This feature was moved by both you and {user}.'
       nodelist: 'Nodes were changed by both you and {user}.'
       memberlist: 'Relation members were changed by both you and {user}.'
-      tags: 'You changed the <b>{tag}</b> tag to "{local}" and {user} changed it to "{remote}".'
+      tags: 'You changed the {tag} tag to "{local}" and {user} changed it to "{remote}".'
```
1ec5: It looks like the code in merge_remote_changes.js hasn’t been changed to wrap `{tag}` in `<b>`; is that intentional? A fine alternative would be to use plain quotation marks here instead, just like for the `{local}` and `{remote}` placeholders.

modules/core/localizer.js
```diff
@@ -348,13 +350,24 @@ export function coreLocalizer() {
...
+        if (typeof replacements[k] === 'string') {
+          replacements[k] = escape(replacements[k]);
+        }
+        if (typeof replacements[k] === 'object' && typeof replacements[k].html === 'string') {
+          replacements[k] = replacements[k].html;
+        }
+      }
 
-    localizer.htmlForLocalizedText = function(text, localeCode) {
-        return `<span class="localized-text" lang="${localeCode || 'unknown'}">${text}</span>`;
```
1ec5, line 357: I think the intention in #7998 was to allow for each localizable string to be marked up with its language, even if it’s a string that gets inserted into another string as a replacement: #7963. That’s a reasonable idea, though certainly _much_ less important than string sanitization. If I’m not mistaken, these changes include reverting some of the tagged strings to plain text. We should reopen that issue to track a more secure way of handling mixed-language strings. It might require an alternative method that returns both the localized text and the resolved language (a “tagged string”, as it were) so that the parent string can perform the wrapping more securely if the text happens to be used in a UI context.

1ec5 (commented):
Thanks for making these extensive changes. I haven’t gotten a chance to look through the whole PR, but here’s a couple things that I noticed at a glance.

data/core.yaml
tyrasd: I changed this because this was the only string that included direct html formatting, and thought that the extra bolding was not really important in this case. I added some quotes to make it consistent with the rest of the message

modules/core/localizer.js
tyrasd: > If I’m not mistaken, these changes include reverting some of the tagged strings to plain text. […] handling mixed-language strings

That shouldn't be the case. It could be that I accidentally did that in a place or two, in which case I would be happy to fix those.
There were a few cases of where a translated text is nested in another translated text as a replacement value (like [here](https://github.com/openstreetmap/iD/blob/68098489afafea9f35fdfd339948c28224e29d71/modules/ui/sections/changes.js#L32)). You are right that those could technically be mixed languages. I changed these now to be formatted as nested localized-text spans. Or do you think that these nested structure (e.g. something like `<span lang="it">Ciao, <span lang="en">mapper</span>!</span>`) could be an issue for accessibility?

1ec5: Oh great, that nested structure would be ideal. 👍 Glad I was mistaken.

1ec5 (approved):
The changes look good as far as I can tell.

tyrasd:
Quick update on this issue: I do now consider (almost?) all usages of d3's `.html(…)` and thus also all usages of `t.html` as problematic code. Instead, all text should have been inserted using d3 natively (see the new `t.htmlDom` method). I'm starting to rewrite all affected portions of the code. Most "regular" calls to `t.html` should already be refactored. I don't yet have an implementation which works for the nested localized spans and there are a few places where the `.html` injection was used "creatively", which I will still have to look at.

I noticed that there are some glitches in various parts of the UI at the moment:
* ~walkthrough~
* ~help menu~
* ~issues/validation menu~(??)
* ~turn restrictions field~
* ~global search results~

<hr>

//edit: here's an example of a further html injection vector: via OSM tags :see_no_evil: 
image

tyrasd:
The UI glitches seem to be resolved now. Only a few small todos are now left (see top comment).

There's quite a bit more `t.html` I really would like to remove from the code base eventually, but they require deeper refactoring and don't seem to be obvious XSS / HTML injection vectors or AFAICS.

PR #8824, opened 2021-11-23
enable the openseamap overlay

k-yle:
iD currently has 4 sources of overlay imagery that cover the whole world, from editor-layer-index:

- Locator Overlay
- OpenRailwayMap Maxspeeds
- OpenRailwayMap Signalling
- OpenStreetMap GPS traces

There is currently an override to hide [OpenSeaMap](https://map.openseamap.org) from that list. 

This PR removes that override so that you can use OpenSeaMap. This is useful because pretty much all other imagery sources just show the ocean as blue, which isn't very helpful when you're mapping in the ocean.

### DIFF
--- a/scripts/update_imagery.js
+++ b/scripts/update_imagery.js
@@ -39,7 +39,6 @@ const discard = {
   'openinframap-telecoms': true,
   'openpt_map': true,
   'openrailwaymap': true,
-  'openseamap': true,
   'opensnowmap-overlay': true,
 
   'US-TIGER-Roads-2012': true,


tyrasd:
interesting timing :wink: this will be fixed as a part of a much bigger background-imagery overhaul in #8826 (see the [diff](https://github.com/openstreetmap/iD/pull/8826/files#diff-57c61358c75bf928b84ac47390bfb524db78ceac06c22ca7a6497ac033194518))

But thanks for the PR!

PR #8825, opened 2021-11-23
fix bug with numeric properties in custom map data

k-yle:
If you upload a geojson file under "Custom Map Data" with properties that are numbers, the tag editor gets confused:

image

This PR fixes that bug. It can be tested with this file:
```geojson
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": { "a": 1 },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [174.70733642578122, -36.83456911447563],
          [174.90234375, -36.83237080155681]
        ]
      }
    },
    {
      "type": "Feature",
      "properties": { "a": 2 },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [174.89273071289062, -36.80928470205938],
          [174.94491577148438, -36.890604744644705]
        ]
      }
    }
  ]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :bug: Bugfixes
 * Fix hidden tooltips on map control toolbar ([#8781])
 * Fix glitching out turn restriction minimap on narrow sidebars ([#8792])
+* Fix non-string properties of GeoJSON custom map data not being displayed correctly ([#8825], thanks [@k-yle])
 #### :earth_asia: Localization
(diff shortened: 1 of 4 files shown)


tyrasd:
I think if we address this, we should address all [data types](https://en.wikipedia.org/wiki/JSON#Data_types) that might potentially be included in a GeoJSON file. Also the raw tag editor module was not the right place to fix this. I've pushed 3373f1b53de443d5ff1b4ebe3348914a17287e71 to your branch, which should address both of these issues. Let me know if that is fine for you.

k-yle:
thanks @tyrasd !

PR #8826, opened 2021-11-23
Background Switcher Overhaul

### LINKED ISSUE #8814: Organize Backgrounds list by category
### Description

Currently we see:

Backgrounds
- Bing aerial imagery
- CyclOSM
- Esri World Imagery
- Esri World Imagery (Clarity) Beta
- Mapbox Satellite
- Maxar Premium Imagery
- OpenStreetMap (Standard)
- OpenTopoMap
- Stamen Terrain
- NLSC General Map with Contour line
- Taiwan Land-Section Data
- Taiwan Village Boundaries
- Taiwan e-Map Open Data
- None
- Custom

Better would be:

Backgrounds
- Bing aerial imagery
- Esri World Imagery
- Esri World Imagery (Clarity) Beta
- Mapbox Satellite
- Maxar Premium Imagery
----
- CyclOSM
- OpenStreetMap (Standard)
- OpenTopoMap
- Stamen Terrain
----
- NLSC General Map with Contour line
- Taiwan Land-Section Data
- Taiwan Village Boundaries
- Taiwan e-Map Open Data
---
- None
- Custom

tyrasd:
closes #8814 

## User Interface Changes:
* drop separation between background and overlay layers, to instead:
* show imagery grouped by category (aerial / map+elevation / qa / osm / other)
* sort imagery by:
  * best flag: "starred" layers on top
  * overlay flag: overlays after background layers
  * rough layer size: smaller "regional" layers come first (this is the opposite as before)
  * vintage: recent imagery first
  * layer name: alphabetically
* mark globally available layers with a globe emoji (to give them back some emphasis they lost by the new sorting approach)
* show the number of available imagery layers in each group in parenthesis after the groups' label
* show the "maps based on OSM data" and "other" section collapsed by default

## Processing Pipeline Changes:
* switch processing script to consume ELI's GeoJSON instead of the "legacy JSON" output
* much less agressively pre-filter ELI layers: we don't need to exclude many of the "maybe" useful layers anymore, since the better organized UI (see above) isn't cluttered too much anymore by a larger number of OSM and QA layers
* update discarded layers list: add recent-ish tiger-roads overlays

### After

  

The exanded "Maps based on OpenStreetMap" now includes quite some additional, potentially useful (QA) layers:

### Remarks

* fixes the issue that some relatively less useful layers such as the `CyclOSM` are shown quite prominently just because they are a global layer with an alphabetical “advantage”.
* ususally, region-specific layers are to be preferred over global ones because they typically offer higher resolution, newer vintage and/or are better georeferenced.
* similar argument as before, but for the age of the layer: if two imagery layers are otherwise equal, mappers should prefer the more recent one (note that previously, because of the alphabetical sorting, the older layers sometimes ended up on top)
* not all imagery is currently properly categorized in ELI – see for example that "USGS Topographic Maps" is falsely listed under "aerial imagery" (the category used as fallback in this PR if a layer has no category specified). Using the categories in iD might help getting more layers properly categorized over time.
* the "report an imagery problem" link (and extra buttons like the minimap-switcher etc.) were previously (frankly quite inaptly) placed in between the backgrounds and overlays sections. Now it is shown below all of the imagery options.

## Todo
* [x] update changelog
* [x] fix issue with displaying number of layers
* [x] clarify if the discard-list is fine as it is now -> leave it as is, perhaps reconsider later
  * perhaps a few other "low res" layers like the `osmim-imagicode-*` ones should be added?!
  * consider filtering by something like `!(max_zoom < 15)`
* [x] improve sorting by vintage for layers which only have a `start_date` set


tyrasd:
> clarify if the discard-list is fine as it is now (perhaps a few other "low res" layers like the `osmim-imagicode-*` ones should be added?!)

@imagico – do you have an opinion about this? IMHO most of the `osmim` layers are only borderline useful for editing work in iD (because of the zoom-level restrictions imposed by iD).

imagico:
In short:: If iD is not able to allow meaningful mapping work based on 10m/15m GSD imagery that is a deficit that should be fixed.  In the medium term (5-10 years) this kind of resolution will for large parts of the planet typically be the most up-to-date imagery available.  And right now for more than half the planet surface the highest impact mapping work (in breadth initial mapping and maintenance/updating of existing mapping) would be at this scale.  That means even if higher resolution (aerial or VHR sat) imagery is available having recent lower res images as a source for editing to see more recent changes, observe seasonality effects etc. is highly useful.  And the osmim are primarily meant to serve these uses, i.e. provide alternatives to where the standard layers are outdated or where specific seasons (like autumn colors) are of interest for mapping.

What could be important to allow mappers to better select a background layer is information on the actual resolution of images.  It is good that you prioritize recency of the images (because that is often less clear to a mapping newbie from looking at the images) but the very common case is that the highest resolution available is not the most recent and that the 'best' flag is not useful.  The maximum zoom level is not always a good indicator for actual resolution.  Neither is the layer size.  Unfortunately for the global aggregate layers this is also not a uniform value for the whole world - you would have to subset the layers with different attributes.

If resolution information was available you could present a mapper who has selected an older high resolution layer currently any more recent lower resolution layers as a valuable alternative and the other way round - indicate to mappers using a recent low resolution layer that an older high resolution alternative is available.  You could even go a step further and make mappers specific suggestions based on what they map - a mapper currently mapping buildings, footways and individual trees will likely not find the osmim of use while someone doing landcover mapping could well be pointed to them as a valuable alternative.

Regarding actual use by mappers - i don't have precise numbers but based on referrer stats use of the osmim via iD on osm.org is not insignificant compared to other editors like JOSM.  How much of that is actual mapping and how much is just lurking is a different question of course.

modules/ui/sections/background_list.js
```diff
@@ -164,14 +97,41 @@ export function uiSectionBackgroundList(context) {
...
             .filter(filter)
             .sort(function(a, b) {
-                return a.best() && !b.best() ? -1
-                    : b.best() && !a.best() ? 1
-                    : d3_descending(a.area(), b.area()) || d3_ascending(a.name(), b.name()) || 0;
+                return d3_descending(a.best ? 1 : 0, b.best ? 1 : 0) ||
+                       d3_ascending(a.overlay ? 1 : 0, b.overlay ? 1 : 0) ||
+                       d3_ascending(Math.floor(Math.log(a.area())/3), Math.floor(Math.log(b.area())/3)) ||
+                       d3_descending(a.endDate, b.endDate) ||
+                       d3_ascending(a.name(), b.name()) || 0;
```
1ec5: No sorting criteria is explicitly labeled, other than the icon for global sources, so adding more criteria could make the list less intuitive in some places. Now that the list is grouped by category, the category that these extra sorting criteria affect most would be “Maps based on OpenStreetMap”, but sources in that category not to have `best` or `endDate` set. If this becomes a problem in the future, maybe could consider a toggle to more straightforwardly sort by name versus vintage versus scope.

modules/ui/sections/background_list.js
```diff
@@ -164,14 +97,41 @@ export function uiSectionBackgroundList(context) {
...
-        var sources = context.background()
+    function categoryMapping(layer) {
+        var categoryWeights = {
+            'photo': categoryGroups.photo,
+            'historicphoto': categoryGroups.photo,
+            'map': categoryGroups.map,
+            'historicmap': categoryGroups.map,
+            'elevation': categoryGroups.map,
+            'osmbasedmap': categoryGroups.osm,
+            'qa': categoryGroups.osm,
```
1ec5: Is `qa` supposed to be only for OSM-based maps? In the U.S., the TIGER Roads 2020 overlay is also categorized as `qa`, even though it’s purely based on federal census data. This is one of the most important layers in the U.S., the main one used by mappers to name streets, so it would be unfortunate to bury it in the middle of such a long “Maps based on OpenStreetMap” category.

data/core.yaml
```diff
@@ -774,9 +774,15 @@ en:
     title: Background
     description: Background Settings
     key: B
-    backgrounds: Backgrounds
+    backgrounds:
+      photo: Aerial and Satellite Imagery
+      map: Maps and Terrain Cartography
+      osm: Maps based on OpenStreetMap
```
1ec5: A bit more concise (and in title case for consistency):

```suggestion
      osm: OpenStreetMap-Based Maps
```

data/core.yaml
```diff
@@ -774,9 +774,15 @@ en:
...
+      other: Other Imagery
+    extras: Background Extras
     none: None
     best_imagery: Best known imagery source for this location
+    global_imagery: This imagery source is available globaly
```
1ec5:
```suggestion
    global_imagery: This imagery source is available globally
```

modules/ui/sections/background_list.js
```diff
@@ -10,28 +10,50 @@ import { uiTooltip } from '../tooltip';
...
+                .disclosureExpanded(group.expanded || categoryMapping(context.background().baseLayerSource()).id === group.id)
+        );
+    })();
 
     function previousBackgroundID() {
         return prefs('background-last-used-toggle');
     }
 
-    function renderDisclosureContent(selection) {
+    function sectionLabelHtml(group, count) {
```
1ec5, line 92: Should a collapsed section indicate that it contains the currently selected or alternate source?

modules/ui/sections/background_list.js
```diff
@@ -164,14 +97,41 @@ export function uiSectionBackgroundList(context) {
         });
     }
 
-    function drawListItems(layerList, type, change, filter) {
-        var sources = context.background()
+    function categoryMapping(layer) {
+        var categoryWeights = {
+            'photo': categoryGroups.photo,
```
1ec5: > not all imagery is currently properly categorized in ELI – see for example that "USGS Topographic Maps" is falsely listed under "aerial imagery" (the category used as fallback in this PR if a layer has no category specified). Using the categories in iD might help getting more layers properly categorized over time.

Indeed, fixed in osmlab/editor-layer-index@4c1aeb39c679dc7bc6e65a45ebea63851eea6d7e. 😄

pnorman:
Can we pull custom and none out of "other". Custom might be any type of imagery, and none also seems special to me.

modules/ui/sections/background_list.js
tyrasd: I've added (890d654fe61bdcdabd6e987398dd518e05fbbbd5) a separate group for the QA layers. While they are by definition always based on OSM data (aren't they?), they are indeed not always solely based on OSM data. I agree that given their importance it would be a shame to "bury" them in the long list of regular OSM based maps.

However, I've chosen to show the qa section collapsed by default, since the usage and interpretation of these layers is, general speaking, not obvious and could be confusing to use without proper instructions and/or a dedicated map legend which we both don't have at the moment. I think users who want to use the TIGER overlay would likely search it in the "Quality Assurance" section.

tyrasd:
I've tweaked this a bit now:

* quality assurance layers get their own section
* display icons (if third-party-icons are enabled) to be able to more quickly recognize/find specific layers
* tweak styling such that instead of a few unconnected sections, the imagery selector is now displayed as a single "table" of entries with collapsible rows
* none, custom and the osm-gps and locator overlays are now outside of the collapsible sections

modules/ui/sections/background_list.js
tyrasd: sections are expanded if they contain a layer which is loaded at startup (e.g. from a hash parameter). I think it is not necessary to add an indicator for this because the user must have actively closed the section it is in.

modules/ui/sections/background_list.js
tyrasd: AFAICS, the OSM layers are basically the most unaffected layers, since most of them are available on a global scale (indicated by the icon), and then sorted just alphabetically.

I think where the new sorting is having the most impact is for the "photo" layers, where one typically only just has a handful up to a dozen or so non-global layers. I think that's still a reasonable amount and doesn't need a more sophisticated sorting IMHO. FWIW, not even JOSM has anything more powerful in this regard. :wink:

modules/ui/sections/background_list.js
1ec5: The TIGER Roads overlay is purely based on census data without any OSM data at all. I haven’t seen a crisp definition of each category anywhere; it could be that the TIGER Roads overlay is simply miscategorized, though mapping based on it could be described as “quality assurance”. I think it’s fine to put it under that section for now. If we hear of people having a hard time finding it, we can always recategorize it as “other”.

modules/ui/sections/background_list.js
```diff
@@ -146,7 +146,7 @@ export function uiSectionBackgroundList(context) {
                        d3_ascending(a.overlay ? 1 : 0, b.overlay ? 1 : 0) ||
                        d3_ascending(Math.floor(Math.log(Math.abs(a.area()))/3), Math.floor(Math.log(Math.abs(b.area()))/3)) ||
                        d3_descending(a.endDate || a.startDate, b.endDate || b.startDate) ||
-                       d3_ascending(a.name(), b.name()) || 0;
+                       d3_ascending(a.name().toLowerCase(), b.name().toLowerCase()) || 0;
```
1ec5: As a possible followup, alphabetical sorting could be pulled out into a utility function. It would be a good place to incorporate [diacritic folding](https://github.com/openstreetmap/iD/blob/19c95248702bda93f78969e6623c7232462b916b/modules/util/util.js#L451-L452) or, even better, [language-aware sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator), which I’m pretty sure `d3_ascending` doesn’t do on its own.

tordans:
Two bits of feedback based on the screenshot:

I don't think the icons add any value here but make the list harder to parse. To me, I do not recognize most icons. They are too small in more than half the cases to be recognizable. In general, those icons are not something "I know" (Like a brand in NSI), which have a "ah that is the right choice" moment. Even with the Bing Icons, which is the most recognizable for me, I associate this with the search, not a map layer.

In case you keep the icons, please consider leaving a blank space where no icon is available, so that the list still has a clean left alignment. And start the process in NSI to document what a good Icon is and how to add it (documentation like size).

Are those icons located on an external URL? In this case: They might need to be part of the same code that hides them whenever the privacy setting for NSI are deactivated (active by default AFAIK).

I noticed recently in https://github.com/osmlab/editor-layer-index/pull/1203#issuecomment-885457916, that the categories in ELI are not defined in sense of a documentation. Should iD start to rely on them, I suggest to start the process to define and maybe validate the categories in ELI.

css/80_app.css
```diff
@@ -3157,11 +3161,24 @@ div.full-screen > button:focus {
...
 
-.layer-list li.best > div.best {
+.layer-list li > div.stamp {
     padding: 5px;
     flex: 0 0 auto;
 }
 
+.layer-list li > div.imagery-icon {
+    margin: 5px;
+    max-width: 20px;
```
1ec5: This might just happen to do the right thing for [the abnormally wide icon](https://www.mcohio.org/_assets_/images/logo.png) specified by [this source](https://github.com/osmlab/editor-layer-index/blob/e1ec9cbf0146c8ca06d3fa6665901638a52ab92c/sources/north-america/us/oh/Montgomery_OH_2019.geojson?short_path=0de5cbe#L11).

tyrasd: I forgot to remove this css from a first implementation. Now ([see below](https://github.com/openstreetmap/iD/pull/8826/files/79a7e88537cb275187bd63b04e9336dea8b62d22#diff-3f86fb5cebd04342575e756d440d18a0b6cfb131a3c26d7897ff1793f8b6d30bR3216-R3221)) the non square icons are just shrunk down to fit, which would totally not work in this case. :see_no_evil:

tyrasd:
> I don't think the icons add any value here but make the list harder to parse. To me, I do not recognize most icons. They are too small in more than half the cases to be recognizable. In general, those icons are not something "I know" (Like a brand in NSI), which have a "ah that is the right choice" moment. Even with the Bing Icons, which is the most recognizable for me, I associate this with the search, not a map layer.

I was also having some concerns about whether or not to show the icons in the beginning, but in the end I think there is some benefit to it: Even though a regular user might not recognize all icons from the start, it might help them to recognize a layer in the future again. Repeated icons also provide a kind of "sub grouping" for layers which belong together as coming from the same provider; we might want to give users a hint to check out a terrain layer from the same source as a local "best" orthophoto layer. Additionally, some icons are based on the OSM logo on purpose to show that they are provided by the OSM community or the OSMF, which could be also an interesting information. And FWIW, JOSM displays the icons as well, so it might be just good to show the icons for cross-editor consistency. So, to summarize, I see a few small benefits which in their sum outweigh the additional noise in the UI.

> In case you keep the icons, please consider leaving a blank space where no icon is available, so that the list still has a clean left alignment.

:+1: 

> And start the process in NSI to document what a good Icon is and how to add it (documentation like size).

You mean in ELI, right? 

> Are those icons located on an external URL? In this case: They might need to be part of the same code that hides them whenever the privacy setting for NSI are deactivated (active by default AFAIK).

Yes, in general they are externally hosted (most either on ELI's github.io or direclty on the service provider), and they are already [only shown](https://github.com/openstreetmap/iD/pull/8826/files#diff-e37d93c86d6f7ed0f922b8bfad4c4ea1d9ef6b5c2643a06382adb535c0bf5520R177-R181) if the "third party icons" preference is enabled.

> I noticed recently in [osmlab/editor-layer-index#1203 (comment)](https://github.com/osmlab/editor-layer-index/pull/1203#issuecomment-885457916), that the categories in ELI are not defined in sense of a documentation. Should iD start to rely on them, I suggest to start the process to define and maybe validate the categories in ELI.

:+1: for better documentation on ELI's side. //edit see https://github.com/osmlab/editor-layer-index/pull/1354

modules/ui/sections/background_list.js
tyrasd: done in c9d6bf633

bhousel:
It looks like this refactor mixes in the overlays (with checkboxes) with the base maps (with radio buttons)?

tyrasd:
> It looks like this refactor mixes in the overlays (with checkboxes) with the base maps (with radio buttons)?

Yes, indeed.

tordans:
>>And start the process in ~~NSI~~ ELI to document what a good Icon is and how to add it (documentation like size).
>
>You mean in ELI, right?

Right, sry.

tyrasd:
@bhousel might I ask why you gave this PR a :-1: ? Is it because you disagree with the proposal to leverage the imagery category information in the first place (see #8814), or is more with any concrete implementation detail in this PR? I didn't see a good design to keep the strict separation of overlays and "background" layers while also showing the imagery categories as distinct groups of imagery layers. If you have a better solution, I'm all ears! Or is your reaction based on some of the other changes proposed here? Is there a better way to sort the layers? Does it not make sense to consume ELI's GeoJSON output? …? I'd be glad to hear your opinions.

bhousel:
> @bhousel might I ask why you gave this PR a 👎 ? 

Ok..

I really disagree with mixing the checkboxes for overlays with radio buttons for imagery in the same list.  I also disagree with hiding some sources under categories - users can end up in a situation where they can't find the imagery that is selected.  I think the lists are too long now.  "None", "Custom", "GPS Traces" all ended up somewhere, but they are clearly design afterthoughts.

> Is it because you disagree with the proposal to leverage the imagery category information in the first place (see #8814), 

Yes I disagree with this too - the categories really aren't meaningful to most users.  Some like "historicphoto" aren't even well named ("historic" in this context just means, "not latest". The imagery could be pretty recent.)

> or is more with any concrete implementation detail in this PR? I didn't see a good design to keep the strict separation of overlays and "background" layers while also showing the imagery categories as distinct groups of imagery layers. 

I think this list was more user friendly how it was before.  I don't think the categories add value.  I dont think a new user would see this list and feel confident to make good imagery decisions.

> If you have a better solution, I'm all ears! 

Maybe seed the list with the 10 most useful and add a "Manage Imagery.."  link below that opens another modal screen where power users can customize what appears in their list.  You can't show a list of 50 things with radio buttons and checkboxes jumbled together.

> Or is your reaction based on some of the other changes proposed here? Is there a better way to sort the layers? Does it not make sense to consume ELI's GeoJSON output? …? I'd be glad to hear your opinions.

I'd double check whether ELI's GeoJSON handles polygons with holes properly.  For years it didn't, so we have code in iD to workaround this issue.  (I submitted a PR to ELI years ago to fix this and Stereo didn't want it.)

bhousel:
I guess I should add what I do like about it..
- I like the globe icon for global sources
- I like the imagery icons, but agree with @tordans that they aren't very recognizable, nor designed to be displayed at the small scale.  Agree with his suggestion to review all of these in ELI for suitability.

modules/renderer/background_source.js
```diff
@@ -92,7 +93,9 @@ export function rendererBackgroundSource(data) {
 
     source.area = function() {
         if (!data.polygon) return Number.MAX_VALUE;  // worldwide
-        var area = d3_geoArea({ type: 'MultiPolygon', coordinates: [ data.polygon ] });
+        var area = data.polygon.reduce((acc, outerRing) => {
+            return acc + Math.abs(polygonArea({ type: 'Polygon', coordinates: [outerRing] }));
+        }, 0);
```
1ec5, line 98: > I'd double check whether ELI's GeoJSON handles polygons with holes properly. For years it didn't, so we have code in iD to workaround this issue. (I submitted a PR to ELI years ago to fix this and Stereo didn't want it.)

If I remember correctly, the oddest thing about ELI’s representation is that a polygon can have multiple outer rings. This change does seem to address the same case without affecting inner ring handling, but there should be a test about it if there isn’t already.

tyrasd:
Thanks @bhousel for your input!

> I think this list was more user friendly how it was before.

I beg to disagree on this point. Here's the current list in a region where I often map:



I've underlines the layers which are most useful to me (and I would believe to most mappers in the region). It's quite annoying to filter through the list to find the layers randomly in the middle or towards the bottom. I've also highlighted OSM-based map layers which are also seemingly "randomly" sprinkled over the list. It's also a bit irritating and incoherent that the relatively minor "OSM (Basque/Breton/Occitan Style)" are shown as options, while other more established styles (such as the French or German ones for example) don't show up. Users were actually irritated about this: see https://discordapp.com/channels/413070382636072960/782592770849505281/903363620908441661 for example.

> the categories really aren't meaningful to most users. Some like "historicphoto" aren't even well named ("historic" in this context just means, "not latest". The imagery could be pretty recent.)

I don't know why you bring this up. The grouping in the PR does conflate the "historic" categories on purpose. :woman_shrugging: 

I think that most mappers can actually understand the difference between aerial imagery and a rendered map.

> I dont think a new user would see this list and feel confident to make good imagery decisions.

Beginners are IMHO not very likely to use the background switcher too often, since they are likely happy with the default background anyway. If anything, I think the fact that the list is now grouped offers at least some basic guidance to new users (currently there is none at all – just a long list of layers with no apparent structure). Can you support your claim in any way? Are there studies that suggest that new users prefer linear lists over categorized ones? 

> hiding some sources under categories - users can end up in a situation where they can't find the imagery that is selected

This is only the case when the user actively closes the group which contained the currently active layer or overlay. This is currently also a possible situation (see the following screenshot), so I don't get why this would be a downside of this PR.



>  I think the lists are too long now.

As you should have been able to see from the screenshots above, in general the lists are actually shorter now. :shrug: 

> "None", "Custom", "GPS Traces" all ended up somewhere, but they are clearly design afterthoughts.

This is a valid point, I agree that my design is not perfect in this regard. I would argue that the benefits of the overhaul outweigh this flaw in the design. It would also be something that could be addressed at some later stage. Perhaps the "none" option should actually a toggle switch for example, and there could be an "add custom imagery" button which allows to add more than a single custom imagery layer…

> Maybe seed the list with the 10 most useful and add a "Manage Imagery.." link below that opens another modal screen where power users can customize what appears in their list.

I'm having a hard time to see how that could be an improvement. I would be concerned that this significantly hinders the discoverability of alternative layers for example.

If, in a specific place, there are 30 good layers to choose from, why should we not show them all? Realistically, this high number will be only ever the case in "extreme" situations (the [screenshot above](https://github.com/openstreetmap/iD/pull/8826#issuecomment-980114244) is from a place bordering to two regions where quite many layers are available). Still, the categorization of this PR makes it manageable: as a user I know if I'm looking for an alternative aerial imagery or perhaps a hillshade layer, or something else – so, even there I only have to look through a dozen or so options, not the full list of 36 layers (excluding overlays) which are currently offered there:

tyrasd:
> I'd double check whether ELI's GeoJSON handles polygons with holes properly. For years it didn't, so we have code in iD to workaround this issue.

Thanks for the pointer, I will double-check this for sure. :+1: 

> I submitted a PR to ELI

Can you share a link to that please? ~~I cannot find it right now: https://github.com/osmlab/editor-layer-index/pulls?q=is%3Apr+author%3A%40bhousel~~ //edit: sorry, I messed up the github search (the correct link should have been https://github.com/osmlab/editor-layer-index/pulls?q=is%3Apr+author%3Abhousel) – and the mentioned PR is probably https://github.com/osmlab/editor-layer-index/pull/490

bhousel:
> I've also highlighted OSM-based map layers which are also seemingly "randomly" sprinkled over the list. It's also a bit irritating and incoherent that the relatively minor "OSM (Basque/Breton/Occitan Style)" are shown as options, while other more established styles (such as the French or German ones for example) don't show up. Users were actually irritated about this: see https://discordapp.com/channels/413070382636072960/782592770849505281/903363620908441661 for example.

Yes, this is annoying and these shouldn't be in the list!  For years I had them filtered out.

> > the categories really aren't meaningful to most users. Some like "historicphoto" aren't even well named ("historic" in this context just means, "not latest". The imagery could be pretty recent.)
> 
> I don't know why you bring this up. The grouping in the PR does conflate the "historic" categories on purpose. 🤷‍♀️

I brought it up because - you asked me why I disagree with using the categories??  If you have to write code to conflate some categories, maybe it's a sign that they're not good.

I don't understand why you asked for my opinion and then just responded with "I'm having a hard time seeing" and "can you support your claim" and stuff like that.  You really want me to send you usability studies? 

I don't get why you linked to a GitHub search of ELI PRs I've submitted, but then botched up the search to make it look like I've never done any work on this before.  I _know_ you know how to search GitHub - why would you do that?

I mean look - you're going to do whatever you want with this background imagery list.  Merge it if you like it.  I don't have to pull this change into RapiD if I disagree.  It doesn't matter.

tyrasd:
> I don't get why you linked to a GitHub search of ELI PRs I've submitted, but then botched up the search to make it look like I've never done any work on this before. I know you know how to search GitHub - why would you do that?

I'm sorry! Honestly, I don't know why I thought one has to put the `@` in front of the username in the github search. I found the PR now: https://github.com/osmlab/editor-layer-index/pull/490

> I brought it up because - you asked me why I disagree with using the categories?? If you have to write code to conflate some categories, maybe it's a sign that they're not good.

My argument is that while the categories are probably not perfect, they are better than nothing. Sometimes you have to work with what you got. Or as someone else once said: don't let perfect be the enemy of good.

> I don't understand why you asked for my opinion and then just responded with "I'm having a hard time seeing" and "can you support your claim" and stuff like that.

Look, I started to work on this because there was feedback from the community that the current solution could be improved by leveraging the categories. I think it is fair to say that the ELI grew quite a bit since the time the background selector originally was developed. I came up with a solution which in my eyes solves the reported issue.

I was hoping that your feedback would be more than a lengthy _I don't like it_. :disappointed: I'm still new to doing this, so I am sincerely seeking constructive feedback, some guidance perhaps and maybe a piece advice from more experienced people like you.

Perhaps I'm still too green to see why _categories don't add value_ here. Please help me out! In my reply I tried to explain why I think that the grouping by category _does_ make an improvement. Please don't mistake it for ignorance. Can you point out what is wrong in my thinking?

> You really want me to send you usability studies?

Yes! If you know any which might be relevant: please share them!

1ec5:
Sorry to jump in the middle of this conversation, but can we take a step back? First of all, this change did turn out to be more wide-ranging than #8814 initially suggested. I recognize that sometimes [feeding the mouse](https://en.wikipedia.org/wiki/If_You_Give_a_Mouse_a_Cookie) leads to more churn than expected, but it might help to walk the audience through each of the changes that took place and the problem it’s trying to solve, and maybe even split things out into separate issues or PRs so we can distinguish between what’s contentious and needs iteration and what’s more straightforward.

For the future, if it seems like there’s a miscommunication or some piece of feedback that doesn’t read right, consider reaching out directly in chat or some other medium. The rest of us watching this repository don’t really benefit from back-and-forth, and it dissuades folks from pitching in with their own constructive viewpoints.

> Beginners are IMHO not very likely to use the background switcher too often, since they are likely happy with the default background anyway.

I don’t think I can agree with this sweeping statement. iD needs to be designed with the assumption that Bing is the default background layer, because most of the world doesn’t have the luxury of a local source that’s simultaneously fresher, better aligned, _and_ higher-resolution. That you do is something to be proud of! For the rest of us, Bing ends up being a compromise even in most places that have local sources – just look all the places on [this map](https://osmlab.github.io/editor-layer-index/) that have overlapping polygons but none gold-colored.

Over the years, as I’ve onboarded hundreds of people to OSM, the very first thing I show them after iD’s walkthrough is the background switcher, whether they’re mapping at home in the U.S. or in Africa for a HOT task. Whenever I forget to mention it, half the mappers ask for it anyways, or they ask if it’s OK to copy from Google Maps imagery. Bing is fine for most basic mapping, at least where the quality hasn’t regressed beyond utility, but one has to double-check with at least one other layer like Maxar just in case it’s outdated. This is one of those “tricks of the trade” that’s difficult to learn through a global UI, so local communities try to share this knowledge on the wiki or by word of mouth in chat.

Ohio

The screenshot in https://github.com/openstreetmap/iD/pull/8826#issuecomment-983589860 is quite intimidating. I sympathize with you and nearby mappers for having put up with such a long list. That said, this is quite a variable part of the UI. Where I map in Ohio and similarly where @bhousel probably maps, the list is significantly shorter but still atypical for having any subnational layers at all:

* Backgrounds
   * Global layers
   * Two national orthophoto mosaic layers that are useless in most places but essential in some places for freshness
   * Two national, non-OSM-based map layers useful for copying names and mapping features in rugged terrain
   * Two statewide orthophoto mosaic layers that have superb alignment and high resolution but varying freshness (so we can’t mark them as “best”)
* Overlays
   * Global overlays
   * National overlays useful for copying names and mapping roads

There are some usability issues with the status quo in places like Ohio. In particular, the most useful backgrounds are sorted to the top and bottom of the list, with less useful layers in the middle. A new user or someone mapping in this area for the first time would reasonably cycle through the list in order, optimistically looking for something better than the default Bing layer. But first they’ll see a parade of layers like CyclOSM and Standard that don’t seem like an improvement over Bing. Without encouragement from someone looking over the shoulder, they might give up before they get to OSIP, a state agency that’s obscure even among state government officials.[^osip]

The change you made to sorting deftly addresses this discoverability issue, and the 🌐 icons are a nice touch. What could be explained more clearly is how it relates to relaxing the filters to bring in more layers from ELI, which then necessitates categorizing and mixing in overlays. Where Ohioans had a minor discoverability problem before, they’d now have the cognitive overhead of the collapsible headings and a less apparent distinction between layers and overlays. (It makes sense that checkboxes behave differently than radio buttons, but you have to play around with it to get it. There’s a reason HTML forms rarely mix the two in a single `<fieldset>`.)

The tradeoff is clearly different in South Tyrol. There must be some very enthusiastic ELI contributors over there. How do we accommodate these different experiences? Shunting the more niche layers to a “More…” menu or manager is a great idea. Preset fields already make a distinction like that. It would probably make it unnecessary to mix overlays with backgrounds. The non-orthophoto background layers occupy a gray area where they’re only useful for certain tasks, but a new mapper wouldn’t need to be introduced to them right away. (Even the TIGER overlays might be described that way.) Would the old filter be a good cutoff between default and additional layers, or do we need more information from ELI?

[^osip]: Despite their names, the OSIP layers are verging on historical, but they’re such a joy to trace from. And new mappers have no clue whether 1-foot or 6-inch imagery is good or bad. Maybe I should rename these layers to “Our Stale Imagery Paradise” in ELI.

tyrasd:
Thank you @1ec5 for your wise words.

> this change did turn out to be more wide-ranging than #8814 initially suggested

Yes, this is true. At first I thought this would be only a relatively minor [reordering job](https://github.com/openstreetmap/iD/issues/8814#issuecomment-973998970). But during the initial development the simple just-adding-sorting-by-category approach didn't feel like a good improvement without (some of) the additional refactoring presented here. I don't seem to be the only one having such thoughts: Even you mentioned the [sorting "problem"](https://github.com/openstreetmap/iD/issues/8814#issuecomment-973264744) (_“it shifts the local sources down to where users are less likely to notice them”_) in the #8814 ticket. :wink: 

> > Beginners are IMHO not very likely to use the background switcher too often, since they are likely happy with the default background anyway.

> I don’t think I can agree with this sweeping statement. iD needs to be designed with the assumption that Bing is the default background layer, because most of the world doesn’t have the luxury of a local source that’s simultaneously fresher, better aligned, and higher-resolution.

I fully agree. Maybe I have phrased my sentence wrongly: What I simply meant to point out is that new users are not presented with the switcher neither in the walkthrough nor when they open iD for mapping.

> Over the years, as I’ve onboarded hundreds of people to OSM, the very first thing I show them after iD’s walkthrough is the background switcher, whether they’re mapping at home in the U.S. or in Africa for a HOT task. Whenever I forget to mention it, half the mappers ask for it anyways, or they ask if it’s OK to copy from Google Maps imagery. Bing is fine for most basic mapping, at least where the quality hasn’t regressed beyond utility, but one has to double-check with at least one other layer like Maxar just in case it’s outdated. This is one of those “tricks of the trade” that’s difficult to learn through a global UI, so local communities try to share this knowledge on the wiki or by word of mouth in chat.

What I would conclude from this is that the walkthrough should be enhanced to include one more step which explains how to use the backgrounds toolbar.

> The screenshot in #8826 (comment) is quite intimidating. I sympathize with you and nearby mappers for having put up with such a long list. That said, this is quite a variable part of the UI.

Sure. But given that even places like Ohio with relatively few additional ELI layers have issues: Is that not even more evidence that the current solution need to be changed/fixed?

> The change you made to sorting deftly addresses this discoverability issue, and the :globe_with_meridians: icons are a nice touch.

What I have not yet shown is that I added a second icon for layers which are in a "historic" category, informing users that these layers should be used with care.



> What could be explained more clearly is how it relates to relaxing the filters to bring in more layers from ELI, which then necessitates categorizing and mixing in overlays.

My reasons behind relaxing the discard-list were the following:

1. The editor layer index project is an inventory of layers specifically designed for mapping in OSM. If the community and maintainers of ELI decide that a source should be included in the index, we at iD should have really good reasons to discard it.
2. [Originally](https://github.com/openstreetmap/iD/blob/4e8cb0b97f1a2ed997256029bd44907852ec1d25/data/update_imagery.js) the discard list only included broken layers (which have long been fixed), duplicates and what was termed _“no clear use case”_.
3. The discard list was later extended to also [disable QA layers](https://github.com/openstreetmap/iD/commit/53d8a30032da9abe384b52126338ffbe548ddae0), which was explained in the following way: _**We need to think about how to best integrate these**; just dumping them into a huge list of overlays is not a good approach._ (emphasis by me)
4. The [discard list grew](https://github.com/openstreetmap/iD/blob/v2.20.2/scripts/update_imagery.js#L19-L63) quite a bit over time and becomes harder to maintain, up to a point where more and more layers "slipped through".

With the proposed UI in this PR I was able to solve the reasons behind points 3 and to some extent also point 2. It also greatly helps with point 4 and ultimately gives the community back a part of "ownership" of "their" imagery index (see point 1 and partly point 2).

> Where Ohioans had a minor discoverability problem before, they’d now have the cognitive overhead of the collapsible headings and a less apparent distinction between layers and overlays. (It makes sense that checkboxes behave differently than radio buttons, but you have to play around with it to get it. There’s a reason HTML forms rarely mix the two in a single <fieldset>.)

Yes, it is true that UIs rarely mix radio buttons and checkboxes, but IMHO it works in this case: The layers are all just background layers. Some happen to be partly transparent, so that they can be shown on top of a non-transparent one below. That's where the different UI elements come into play: a checkbox means that the layer can be activated independently from the others, while a radio button means that activating a layer will disable the previously selected one.

The alternative I have considered would be to allow all background layers to be stacked like in many graphics editors (like gimp/photoshop), GIS software (QGIS, ArcGIS) and JOSM: this is a more powerful, but at the same time much more complex UI. In my opinion that really is out of scope for iD because of the additional complexity.

> How do we accommodate these different experiences? Shunting the more niche layers to a “More…” menu or manager is a great idea. […] Preset fields already make a distinction like that. It would probably make it unnecessary to mix overlays with backgrounds.

Yes, but the problem I see with this approach is that someone has to define what is niche and what is mainstream for the imagery layers (the presets have been curated to have a distinction between "fields" and "more fields"). Do we only put bing (plus maybe also the other global aerial imagery sources) and any locally "best" layer in the default list and hide everything else behind "More…"? I think that would lower the discoverability of any alternative local layers quite a bit, which is exactly the opposite of what I wanted to achieve for iD in this PR.

> The non-orthophoto background layers occupy a gray area where they’re only useful for certain tasks, but a new mapper wouldn’t need to be introduced to them right away. (Even the TIGER overlays might be described that way.)

I would argue that "map" (and "elevation") layers are also quite clearly useful for every-day mapping: Road names and house numbers are a crucially important information for OSM and cannot be mapped from aerial imagery.

> Would the old filter be a good cutoff between default and additional layers, or do we need more information from ELI?

With "old filter" you mean the previous discard-list, right? I think this PR basically does more or less exactly this: Only the button is not called "More…" but split into two buttons titled "OSM based maps" and "QA layers".

bhousel:
> I was hoping that your feedback would be more than a lengthy _I don't like it_. 😞 I'm still new to doing this, so I am sincerely seeking constructive feedback, some guidance perhaps and maybe a piece advice from more experienced people like you.

@tyrasd - https://github.com/openstreetmap/iD/issues/8814#issuecomment-982893416 wasn't just a "lengthy I don't like it".   I mean, it wasn't even lengthy.    😠 

I took a day to think about it more, how/whether to respond, and I'm sorry but I'm bowing out here, I dont have the time to continue with this.  It sounds like you and @1ec5 can take it the rest of the way.  Please listen to his feedback.

scripts/update_imagery.js
```diff
@@ -16,53 +17,15 @@ cutoffDate.setFullYear(cutoffDate.getFullYear() - 20);
 
 
 const discard = {
```
1ec5, line 19: > The editor layer index project is an inventory of layers specifically designed for mapping in OSM. If the community and maintainers of ELI decide that a source should be included in the index, we at iD should have really good reasons to discard it.

I got the impression that ELI is pretty much editor-agnostic and currently serves multiple editors with quite different UIs. It focuses on providing layers and metadata, as opposed to being explicitly opinionated about what an editor should show and how. (Otherwise, the categories’ meanings would’ve been more apparent by now.) Go Map!! only includes [some of the categories](https://github.com/bryceco/GoMap/blob/6e4ddfc5455eba088c3997a5d0c6392e89a321d1/src/Shared/Tiles/TileServerList.swift#L81-L85) and excludes the others, whether from ELI or the JOSM wiki, since QA layers apparently aren’t as useful in a mobile editor. But if ELI would accept upstreaming any of the curation iD has done, then so much the better.

> The discard list grew quite a bit over time and becomes harder to maintain, up to a point where more and more layers "slipped through".

This list does look like a bit of a junk drawer; comments would’ve made it a bit more maintainable. Looking through blame, there seems to have been a variety of reasons for layers to get discarded.

scripts/update_imagery.js
```diff
@@ -16,53 +17,15 @@ cutoffDate.setFullYear(cutoffDate.getFullYear() - 20);
...
-  'osmbe': true,                        // 'OpenStreetMap (Belgian Style)'
-  'osmfr': true,                        // 'OpenStreetMap (French Style)'
-  'osm-mapnik-german_style': true,      // 'OpenStreetMap (German Style)'
-  'HDM_HOT': true,                      // 'OpenStreetMap (HOT Style)'
-  'osm-mapnik-black_and_white': true,   // 'OpenStreetMap (Standard Black & White)'
-  'osm-mapnik-no_labels': true,         // 'OpenStreetMap (Mapnik, no labels)'
-  'OpenStreetMap-turistautak': true,    // 'OpenStreetMap (turistautak)'
```
1ec5: > [Originally](https://github.com/openstreetmap/iD/blob/4e8cb0b97f1a2ed997256029bd44907852ec1d25/data/update_imagery.js) the discard list only included broken layers (which have long been fixed), duplicates and what was termed _“no clear use case”_.

I can understand why OSM-based, non-QA rendered map layers were discarded in the original list, 6a79f2692bc2e749c03ca8b2ae2cb2865a001673, and other commits. I don’t know the history of how they ended up in ELI in the first place, but it seems like they’re mostly useful for navigation in a desktop editor like JOSM. With iD, you can use the built-in locator overlay or global search bar or to open openstreetmap.org in a new tab.

Besides, they always struck me as an encouragement to map for the renderer. But maybe I’m missing an important use case for them? If we’re serious about showing “consumer-facing” maps in iD to trace from, then maybe we should think about providing a legend somehow to avoid confusion about these maps’ symbology. (As tail work, of course.)

scripts/update_imagery.js
```diff
@@ -16,53 +17,15 @@ cutoffDate.setFullYear(cutoffDate.getFullYear() - 20);
...
-  'Waymarked_Trails-Skating': true,
-  'Waymarked_Trails-Winter_Sports': true,
-
-  'OSM_Inspector-Addresses': true,
-  'OSM_Inspector-Geometry': true,
-  'OSM_Inspector-Highways': true,
-  'OSM_Inspector-Multipolygon': true,
-  'OSM_Inspector-Places': true,
-  'OSM_Inspector-Routing': true,
-  'OSM_Inspector-Tagging': true,
```
1ec5: > The discard list was later extended to also [disable QA layers](https://github.com/openstreetmap/iD/commit/53d8a30032da9abe384b52126338ffbe548ddae0), which was explained in the following way: _**We need to think about how to best integrate these**; just dumping them into a huge list of overlays is not a good approach._ (emphasis by me)

These QA layers were discarded back when iD had only a very basic validation feature and no support for vector layers. Resolving QA issues with just a static map layer, no feature inspection, and no legend can be difficult, unintuitive, and potentially misleading, so this rationale wasn’t mere handwaving.

I opened #8838 to track adding interactive support for OSM Inspector along the lines of what we have for other services. I think that would be more fruitful than enabling the WMS layers, even if the WMS layers would be more expedient.

scripts/update_imagery.js
```diff
@@ -16,53 +17,15 @@ cutoffDate.setFullYear(cutoffDate.getFullYear() - 20);
...
-  'openinframap-petroleum': true,
-  'openinframap-power': true,
-  'openinframap-telecoms': true,
-  'openpt_map': true,
-  'openrailwaymap': true,
-  'openseamap': true,
-  'opensnowmap-overlay': true,
-
-  'US-TIGER-Roads-2012': true,
-  'US-TIGER-Roads-2014': true,
```
1ec5: Good catch, these layers were removed in osmlab/editor-layer-index#1019. I agree with updating the exceptions to cover newer years other than the most recent year. For context, the original iD team reached a consensus to discard old TIGER Roads overlays in #2010. TIGER is infamous for bad data but gets incremental corrections and updates every year, so mappers should almost never use the older layers. In fact, should we delete these layers outright from ELI?

scripts/update_imagery.js
```diff
@@ -16,53 +17,15 @@ cutoffDate.setFullYear(cutoffDate.getFullYear() - 20);
...
-  'OSM_Inspector-Highways': true,
-  'OSM_Inspector-Multipolygon': true,
-  'OSM_Inspector-Places': true,
-  'OSM_Inspector-Routing': true,
-  'OSM_Inspector-Tagging': true,
+  'US-TIGER-Roads-2017': true,
+  'US-TIGER-Roads-2018': true,
+  'US-TIGER-Roads-2019': true,
 
   'EOXAT2018CLOUDLESS': true
```
1ec5, line 24: Some layers were apparently discarded due to low resolution: df699b2dc1a3f16efedb066516266ee58ee526d9. That’s related to the discussion above about osmim, right?

scripts/update_imagery.js
```diff
@@ -16,53 +17,15 @@ cutoffDate.setFullYear(cutoffDate.getFullYear() - 20);
...
-  'osm-mapnik-no_labels': true,         // 'OpenStreetMap (Mapnik, no labels)'
-  'OpenStreetMap-turistautak': true,    // 'OpenStreetMap (turistautak)'
-
-  'hike_n_bike': true,                  // 'Hike & Bike'
-  'landsat': true,                      // 'Landsat'
-  'skobbler': true,                     // 'Skobbler'
-  'public_transport_oepnv': true,       // 'Public Transport (ÖPNV)'
-  'tf-cycle': true,                     // 'Thunderforest OpenCycleMap'
-  'tf-landscape': true,                 // 'Thunderforest Landscape'
-  'tf-outdoors': true,                  // 'Thunderforest Outdoors'
```
1ec5: Did we solve the authentication issue preventing us from being able to display this layer (#7559)?

modules/ui/sections/background_list.js
```diff
@@ -164,46 +141,87 @@ export function uiSectionBackgroundList(context) {
...
-            .filter(filter)
+            .filter(source => !source.isHidden())
             .sort(function(a, b) {
-                return a.best() && !b.best() ? -1
-                    : b.best() && !a.best() ? 1
-                    : d3_descending(a.area(), b.area()) || d3_ascending(a.name(), b.name()) || 0;
+                return d3_ascending(categoryMapping(a).sort, categoryMapping(b).sort) ||
+                       d3_descending(a.best() ? 1 : 0, b.best() ? 1 : 0) ||
+                       d3_ascending(a.overlay ? 1 : 0, b.overlay ? 1 : 0) ||
+                       d3_ascending(Math.floor(Math.log(Math.abs(a.area()))/3), Math.floor(Math.log(Math.abs(b.area()))/3)) ||
```
1ec5, line 152: > Even you mentioned the [sorting "problem"](https://github.com/openstreetmap/iD/issues/8814#issuecomment-973264744) (_“it shifts the local sources down to where users are less likely to notice them”_) in the #8814 ticket. :wink: 

Yes, I agree with turning the sorting on its head to prefer local layers over global ones, especially since it doesn’t affect what gets chosen by default. It solves the usability problem we’ve both described using more or less a one-liner independent of the other changes. I suppose I should’ve opened a separate issue about it to avoid conflating it with the original request about categorization.

1ec5:
> Yes, it is true that UIs rarely mix radio buttons and checkboxes, but IMHO it works in this case: The layers are all just background layers. Some happen to be partly transparent, so that they can be shown on top of a non-transparent one below. That's where the different UI elements come into play: a checkbox means that the layer can be activated independently from the others, while a radio button means that activating a layer will disable the previously selected one.

Right, that much is clear from the shape of the buttons. But normally, if you see a series of checkboxes separating two series of radio buttons, it means there are three different fieldsets. Compounding the problem, the sections of an accordion view typically separate fieldsets as well – that’s how the rest of iD’s collapsible sections and multiple-choice fields work, after all. Again, the user can play around with the controls to discover how they interact, but a good form should be clear about how it behaves before interacting with it.

If the list weren’t broken up so much by lines and large headings, it would be more apparent at a glance that the radio buttons are connected to a single selection across categories:

no lines

However, this treatment won’t work well if anything is collapsible, and the checkboxes could still interfere with that perception.

> The alternative I have considered would be to allow all background layers to be stacked like in many graphics editors (like gimp/photoshop), GIS software (QGIS, ArcGIS) and JOSM: this is a more powerful, but at the same time much more complex UI. In my opinion that really is out of scope for iD because of the additional complexity.

There must be some middle ground. To me, the layer manager idea sounded more like this:

```
 V  Backgrounds

(x) Foo              *
( ) Bar
( ) Baz             (+)
( ) Boo             (+)
( ) Quux            (+)
( ) None
( ) Custom          ...

Add layer: [ Flip, Flop, Floop… ]

 V  Overlays

[ ] Blah
[x] Bloop
[ ] Blargh

Add overlay: [ Ack, Eep, Gah… ]
```

“Add layer” and “Add overlay” work like “Add field” in the inspector and the user could filter the list by typing a search term. If categorization is still desired, maybe they could instead be buttons that open a dialog with the fully categorized options.

If we’re able to select the initially visible layers well, then novice users would see only the minimum required to complete a basic mapping task, while more advanced users would still have access to obscure or specialized layers.

> Yes, but the problem I see with this approach is that someone has to define what is niche and what is mainstream for the imagery layers (the presets have been curated to have a distinction between "fields" and "more fields"). Do we only put bing (plus maybe also the other global aerial imagery sources) and any locally "best" layer in the default list and hide everything else behind "More…"? I think that would lower the discoverability of any alternative local layers quite a bit, which is exactly the opposite of what I wanted to achieve for iD in this PR.

The default background is a non-historic photo background, so maybe that sets a user expectation that the initial list also consists of  non-historic photo backgrounds. Historic and non-photo backgrounds can go in the more list. That leaves us with five global backgrounds with room for a few more backgrounds we determine to be locally relevant. If there are more locally relevant backgrounds, we can shunt some of the global backgrounds to the more list.

This is just one possible set of heuristics; maybe you could think of something smarter. If ELI doesn’t provide enough signals to make a good determination, why not introduce a `better` flag in ELI to complement the `best` flag, and shunt everything that’s merely good to the more list?

> I would argue that "map" (and "elevation") layers are also quite clearly useful for every-day mapping: Road names and house numbers are a crucially important information for OSM and cannot be mapped from aerial imagery.

They would remain accessible, but selecting one would add it to a single flat list of active options. If we don’t think this is discoverable enough, maybe we have a similar problem with preset fields? I would make a distinction between beginner tasks and everyday tasks: I frequently need to use the TIGER layer and my county’s address layer, but I wouldn’t walk through them in a five-minute introduction to OSM mapping.

> With "old filter" you mean the previous discard-list, right? I think this PR basically does more or less exactly this: Only the button is not called "More…" but split into two buttons titled "OSM based maps" and "QA layers".

After looking into it, I realized even the previous discard list was more targeted than I had imagined. Anyways, hopefully it’s clear what I was picturing when the suggestion come up to flatten and shorten the list but pair it with a more menu. It isn’t a fully baked idea, but allowing checkboxes and section headings to break up a radio button list does make me feel uneasy.

modules/ui/sections/background_list.js
```diff
@@ -189,6 +192,36 @@ export function uiSectionBackgroundList(context) {
...
+                    .ease(d3_easeCubicInOut)
+                    .style('background-color', '#ccc')
+                    .transition()
+                    .duration(300)
+                    .ease(d3_easeCubicInOut)
+                    .style('background-color', null);
+            })
+            .call(svgIcon('#maki-heart-11'))
+            .call(uiTooltip()
+                .title(t.html('background.add_remove_favorite'))
```
1ec5, line 221: These “tooltips” show up immediately, covering up the layer name. Sometimes I also see the tooltip misaligned so that it points to the strip of buttons along the right edge. I think we could just use normal HTML tooltips for the ❤️ 🌐 ⭐ icons, similar to the ↩️ ℹ️ 🚮 buttons in preset fields.

modules/ui/sections/background_list.js
```diff
@@ -189,6 +192,36 @@ export function uiSectionBackgroundList(context) {
...
+                    .transition()
+                    .duration(300)
+                    .ease(d3_easeCubicInOut)
+                    .style('background-color', '#ccc')
+                    .transition()
+                    .duration(300)
+                    .ease(d3_easeCubicInOut)
+                    .style('background-color', null);
+            })
+            .call(svgIcon('#maki-heart-11'))
```
1ec5, line 219: Now a layer could potentially be visually distinguished by activeness, category, favoriteness, geographic scope, quality, and (via the icon) provider. That’s a lot of facets to display, potentially making the UI look a bit busy. (But I did argue that mappers need to balance a lot of considerations when selecting imagery…) Maybe the “best” flag could be treated as a “favorite by default” flag instead, to streamline one of the facets.

data/core.yaml
```diff
@@ -785,6 +786,7 @@ en:
     best_imagery: Best known imagery source for this location
     global_imagery: This imagery source is available globally
     historic_imagery: This imagery is either considered outdated or there exists a more recent version from this provider
+    add_remove_favorite: Add or remove this imagery source to your list of favourites
```
1ec5, line 789: American English spelling, and I _think_ this preposition would be a little less awkward:

> Add or remove this imagery source <del>to</del> <ins>from</ins> your list of favo<del>u</del>rites

But consider splitting it up into two different messages, one for when clicking the icon would add and another for when it would remove.

1ec5 (commented):
Oh, I hadn’t noticed this favorites feature. ❤️ I’m not sure what this means for the counterproposals above, but I think it’s going to lead to some requests to resurrect the favorite presets feature from the v3 branch: #5873.

boothym:
Is there a way to see how it would look for example in the UK/Scotland in a preview iD version or something, or are you able to provide some screenshots if not?

tyrasd:
> Is there a way to see how it would look for example in the UK/Scotland in a preview iD version or something, or are you able to provide some screenshots if not?

Sure, I've put a temporary preview version of this branch on https://tyrasd.github.io/iD/dist/index.html

kymckay:
Minor piece of feedback, the first time I hover over the globe icon it looks like:
Screenshot from 2021-12-10 13-35-41

When I hover away and then back to it, it fixes itself:

Screenshot from 2021-12-10 13-35-49

boothym:
> Sure, I've put a temporary preview version of this branch on https://tyrasd.github.io/iD/dist/index.html

Thanks, here's what it looks like. Not read all of the text above, but the only real concern I have is with the placement of the OSMUK Cadastral Parcels, which are extremely useful for UK mappers in aligning aerial imagery. They are list under "other imagery" which is hidden by default - is there a better place for them?

Logos - the Maxar/Mapbox logo seems to be white, if so that will be a bit useless against this background? No Ordnance Survey logo for the OS layers either, I assume these need to be added in the ELI?

Stamen terrain - perhaps also a question for ELI but is there any point in having this layer, which doesn't seem particularly useful for mapping? Especially as it has a fairly prominent position here.

uk

simonpoole:
>     * sort imagery by:
>       
>       * best flag: "starred" layers on top
>       * overlay flag: overlays after background layers
>       * rough layer size: smaller "regional" layers come first (this is the opposite as before)
>       * vintage: recent imagery first
>       * layer name: alphabetically

If you need any validation of this (I suspect not): this is more or less what Vespucci has been doing for a long time. The main difference is that we handle overlays completely separate (but with the same sorting logic). In particular it compensates a bit for the best flag not really being find grained enough.

tordans:
Thanks for creating [a preview](https://tyrasd.github.io/iD/dist/index.html#background=custom:https://{switch:a,b,c}.tile.openstreetmap.org/{zoom}/{x}/{y}.png%0A&disable_features=boundaries&locale=en&map=17.89/52.47350/13.44894). Tip für non-EN-users: Add `&?locale=en` so remove translation errors.

---

This is what I am looking at 

| a | b |
|---|---|
| image | image  |

---

Unfortunately, I  don't have the time to write a proper feedback, but I wanted to share a few thoughts quickly. I also did not have the time to follow the discussion here, so if there is duplication, please consider it as a +1 to previous comments.

- I still find the new panel confusing to use. I recommend iterating on design (spacing) and ux (consider splitting things again). I suggest to strip this down to only one issue; I feel like this PR does too many things at once.
- Image layer images: In Berlin, they don't help but make the UI more complex visually (the gaps) and cut of space that we had for longer titles. 
- Image layer images: Maxar is white on white
- There are too many tooltips IMO. The tooltips on esri and such don't add any value IMO, a user can just click and see.
- There needs to be space between the grouping-element and the children vs. the next elements.
- Especially at the end, where "x None" looks like a child of "OSM-Based maps" but in fact is a section without group
- The tooltip about outdatedness feels cluttery. I hover over them, they all say the same but don't really help to figure out what I am supposed to do or know. I suggest removing them again. If the feedback comes up, lets add the date/year provided by ELI somewhere. If they need to stay, lets try making them smarter, like splitting the outdated- vs. more-recent-story.
- Star (best) vs. favorite (heart) is confusing to me. Both concepts (star, heart) are use for the concept of favorites. Here they are different things in the same place.
- I think adding favorites should be done in a separate effort. It adds to the complexity of this improvement and I don't think its needed (right away).
- I still dislike the concept of mixing checkboxes and radiobutton. I find it very confusing. And I never saw a good UI (or consumer UI) that used this UI pattern. Let's search for alternatives again, please.
- The global Icon is not needed IMO. I map locally. And I trust the layer to show what is available. There is no need telling me that somewhere else something is there. – We should really fight for every singe pixel and only add if, if it improves the interface by a lot. I suggest removing this again.
- **Could we move the QA-layers to the QA-panel "Map data"?** That is where I would look for this stuff. I consider layers like the address check in the same category as osmose, which is overthere.
  - On this topic, I cannot remember any (or many) issues in this repo requesting more QA layers. Maybe that is not the most pressing issue for the iD Userbase ATM?
  - IMO those layers need quite a bit of explanation for the iD Userbase to be understandable. — Side note: Once we have such a place, I would love to add more Layer to ELI that require such an explanation. Which I didn't since without explaining for example how to use and address layer for opendata Brandenburg or a Layer like [Berlin Straßenbefahrung 2014](https://supaplexosm.github.io/strassenraumkarte-neukoelln/mapproxy_demo_map.html?url=https://mapproxy.codefor.de/tiles/1.0.0/strassenbefahrung/mercator/{z}/{x}/{y}.png) untrained users would missunderstand how to use the data.
- In the QA-MapData-Panel the design seems to be a bit off, maybe a change happended there due to changes in the Background-settings-panel.

tordans:
Thanks for creating [a preview](https://tyrasd.github.io/iD/dist/index.html#background=custom:https://{switch:a,b,c}.tile.openstreetmap.org/{zoom}/{x}/{y}.png%0A&disable_features=boundaries&locale=en&map=17.89/52.47350/13.44894). Tip für non-EN-users: Add `&?locale=en` so remove translation errors.

---

This is what I am looking at 

| a | b |
|---|---|
| image | image  |

---

Unfortunately, I  don't have the time to write a proper feedback, but I wanted to share a few thoughts quickly. I also did not have the time to follow the discussion here, so if there is duplication, please consider it as a +1 to previous comments.

- I still find the new panel confusing to use. I recommend iterating on design (spacing) and ux (consider splitting things again). I suggest to strip this down to only one issue; I feel like this PR does too many things at once.
- Image layer images: In Berlin, they don't help but make the UI more complex visually (the gaps) and cut of space that we had for longer titles. 
- Image layer images: Maxar is white on white
- There are too many tooltips IMO. The tooltips on esri and such don't add any value IMO, a user can just click and see.
- There needs to be space between the grouping-element and the children vs. the next elements.
- Especially at the end, where "x None" looks like a child of "OSM-Based maps" but in fact is a section without group
- The tooltip about outdatedness feels cluttery. I hover over them, they all say the same but don't really help to figure out what I am supposed to do or know. I suggest removing them again. If the feedback comes up, lets add the date/year provided by ELI somewhere. If they need to stay, lets try making them smarter, like splitting the outdated- vs. more-recent-story.
- Star (best) vs. favorite (heart) is confusing to me. Both concepts (star, heart) are use for the concept of favorites. Here they are different things in the same place.
- I think adding favorites should be done in a separate effort. It adds to the complexity of this improvement and I don't think its needed (right away).
- I still dislike the concept of mixing checkboxes and radiobutton. I find it very confusing. And I never saw a good UI (or consumer UI) that used this UI pattern. Let's search for alternatives again, please.
- The global Icon is not needed IMO. I map locally. And I trust the layer to show what is available. There is no need telling me that somewhere else something is there. – We should really fight for every singe pixel and only add if, if it improves the interface by a lot. I suggest removing this again.
- **Could we move the QA-layers to the QA-panel "Map data"?** That is where I would look for this stuff. I consider layers like the address check in the same category as osmose, which is overthere.
  - On this topic, I cannot remember any (or many) issues in this repo requesting more QA layers. Maybe that is not the most pressing issue for the iD Userbase ATM?
  - IMO those layers need quite a bit of explanation for the iD Userbase to be understandable. — Side note: Once we have such a place, I would love to add more Layer to ELI that require such an explanation. Which I didn't since without explaining for example how to use and address layer for opendata Brandenburg or a Layer like [Berlin Straßenbefahrung 2014](https://supaplexosm.github.io/strassenraumkarte-neukoelln/mapproxy_demo_map.html?url=https://mapproxy.codefor.de/tiles/1.0.0/strassenbefahrung/mercator/{z}/{x}/{y}.png) untrained users would missunderstand how to use the data.
- In the QA-MapData-Panel the design seems to be a bit off, maybe a change happended there due to changes in the Background-settings-panel.

nc011:
I just wanted to say thanks @tyrasd for putting this together. I think it's so much more helpful to separate out aerial imagery and OSM-derived products. I never use the OSM derived layers (I really don't know what they're useful for but I guess some must use them) and find them cluttering the aerial backgrounds, so it's great they're in separate lists. 

I do agree with some of the others' comments though. Particularly:  

- A stronger separation between categories (e.g. increase line weighting or add spacing) would be useful
- Adding favourites: I think this is actually quite nice. I probably would just leave a favourited item where it is in its list but with a filled heart icon next to it. I don't think I'm keen on moving it to a new favourites category.
- Radio & checkboxes: I understand the mix-match but they are serving different functions. I think having an overlays category (like now) with a "+" (and conversely a "-") for  adding (removing) a layer on top, rather than the check box, might work better?
- Overlays: in the UK we have the Cadastral parcels and Public Rights of Way layers which are super useful. They are overlays, just lines on the screen, rather than "Other Imagery" though. So I think reintroducing that as a category (and include GPS traces in there) may be helpful.
- None and custom: they kind of look like they sit under "Other imagery" currently I guess because there's a lack of separation between categories. Since they aren't in a category, perhaps have them out of the main list?

etienneJr:
Hi,
Following[ this discussion](https://forum.openstreetmap.fr/t/choisir-son-fond-de-carte-par-defaut-dans-osm-id/30290) on the French forum, I wonder if this PR is outdated or if it is still active and could be merged one day?
The discussion was only about the _favorite background_ function, which is only 1 of many functions in this PR. I also noticed that there had been [another PR](https://github.com/openstreetmap/iD/pull/8058) about favorite background, which was not merged here (without comments), but was [merged in Rapid](https://github.com/facebook/Rapid/pull/179).
So I'm a bit confused: is it possible to re-submit a PR with just the favorite background function? Or is it pointless and we should wait for this PR to be completed?

danieldegroot2:
Suggestion for categories names(and order)
- Aerial Imagery - as collective term, used variably
- Maps[/Data] - as generic term
- Other - as simple term
- Basemaps - as overarching generic term
-- [/Reference] - as 'already currently(/historically) existing' / 'as-is' / 'snapshot'
-- ~~[/Thematic] - as specific to renderings~~ ambiguous, unlikely to be useful.
Quality Assurance
- ~~[/Advanced] - as not directly useful to new users~~ but 'Quality Assurance' is fine.

Furthermore
- Descriptions might be better positioned to left of the sidebar because mouse might obscure the overlay, and descriptions may be quite long and fit better if centered.
However, contrast might be slightly less, it may obscure data if mouse is over sidebar by accident but it shouldn't be much of an issue since user is expected to be focused on one or the other. 
- One or more 'OSM-based' layers could be 'favorited' once at any time, then the preference could be always included for the user (and can be given a hotkey if wanted and hotkey slot is still available) so it may be more useful category.
However, it might be confusing considering the layer might move around, but it will always come back on collapse.
- 'Report imagery issue' could be moved up (though capture date missing is a potential reason to submit a report).
Above the link a short small text could be added if needed.

otherwise changes seem okay.

PR #8828, opened 2021-11-29
Improve the checking of results from the NSI matcher..

### LINKED ISSUE #5693: Superchargers want to flip over to destination chargers
I've noticed that a lot of previously tagged superchargers want to update to destination chargers. I think we're going to lose a lot of good data as folks unintentionally update these. Here's an example of one with Supercharger in the name that really wants to become a destination charger.

bhousel:
(fixes: https://github.com/osmlab/name-suggestion-index/issues/5693)

Previously the code would accept the first useful match, however this match
might be an "alternate" match.  Now we keep iterating until we find a "primary"
match.

This bug could cause the validator to suggest upgrading well-tagged features to
a less wanted feature type.  For example a "Tesla Supercharger" would flip to
a "Tesla Destination Charger" if it had "Tesla" anywhere in its list of tags
(because `short_name=Tesla` is an "alternate" match for both destination
chargers and super chargers)


bhousel:
whoa thanks for the quick merge!

PR #8834, opened 2021-12-01
clear labels for OSM hyperlinks in history panel (OSM history & OSM Profile)

jasonghent98:
Hi! I saw that there were no accepted pull requests to this issue, so I decided to take a stab at it to make it a little clearer in the history panel. I hope this works!


tyrasd:
this is for issue #8784

tyrasd (changes requested):
Can you please also change the label of the changeset link to osm as well (currently line 79)?

I'm wondering if we should make these texts translatable. I think it would not hurt. I can change that for you if you're not feeling comfortable, but it would need additional entries in [`data/core.yaml`](https://github.com/openstreetmap/iD/blob/develop/data/core.yaml) and the translated text needs to be appended here using `.call(t.append(…))` instead of `.text(…)`. See line 24 for an example.

//edit: there are also a few minor code style issues (a missing `;` and some superfluous space) to be fixed, see: https://github.com/openstreetmap/iD/pull/8834/files

jasonghent98:
Hi,
Let me know if that looks better. In regards to making the texts translatable, are you referring to the core.yaml file itself, or rather the texts within the OSM history panel links (history, profile, changeset)? If the first, I could probably just use a third-party tool like iCanLocalize to do that. If the latter, I can also take care of that within the yaml file.

modules/ui/panels/history.js
```diff
@@ -40,7 +40,7 @@ export function uiPanelHistory(context) {
                 .attr('class', 'user-osm-link')
                 .attr('href', osm.userURL(userName))
                 .attr('target', '_blank')
-                .text('OSM');
+                .text('OSM Profile');
```
1ec5: Also call `t.append` here. The argument to `t.append` should refer to an entry in English that you’ll need to add to in data/core.yaml. Once this PR lands, translators will be able to provide translations using Transifex, so there’s no need to handle the translations yourself.

tyrasd:
> In regards to making the texts translatable, are you referring to the core.yaml file itself, or rather the texts within the OSM history panel links (history, profile, changeset)?

I was referring to the texts of the links in the history panel. I've just written a small tutorial about how to do that here: https://github.com/openstreetmap/iD/wiki/How-to-use-iD%27s-localizer-module

jasonghent98:
Thanks for the clarification. Let me know how this looks.

tyrasd (changes requested):
Looks like it's almost done! Only two more things:

1. drop the `.text('…')` lines (lines 44, 81 and 204). The text adding is done already with the `.call(t.append(…));`
2. start all new strings in capital letters perhaps?

tyrasd (approved):

Issue #7713, opened 2020-06-17
Document state of iD's accessibility and setup compatability


quincylvania:
iD's goal has always been to make mapping as approachable as possible to as many people as possible. Much work has been done to make the interface friendly and intuitive to non-experts. But we have to keep in mind that there are many dimensions to accessibility—just because iD works great for some people doesn't mean it works well for everyone. Here are a few of the varied groups we should be constantly considering:

- People with visual impairments
- People with motor control impairments
- People with learning difficulties
- Speakers of languages other than English
- People unfamiliar with US/European culture
- Right-to-left layout users
- Mobile device users
- Touchscreen users
- People with limited internet connections
- …

iD does pretty well with some of these and poorly with others. I'd like to start a living document that details the accessibility of iD over a wide range of dimensions, with the intent of identifying and addressing problem areas. It's imperative we're proactive since mappers won't tell us what's wrong if they've already been alienated by the app.

matkoniecz:
I would be happy to contribute - not very extensively but at least "speakers of languages other than English" is part where I may be useful.

I would consider adding "people completely unfamiliar with OSM" and "JOSM users" as both groups have some quirks.

Not sure is it within planned coverage but I would consider also "people with old hardware" - iD can be noticeably laggy even with relatively decent but old hardware, but at least it generally works.

Maybe also "people using browser other than Chrome"?

quincylvania:
I've been working on this in a new [ACCESSIBILITY.md](https://github.com/openstreetmap/iD/blob/develop/ACCESSIBILITY.md) file. The name is short for convenience; the document involves a lot more.

matkoniecz:
Would it be useful to mention #1646 in "Slow connection"?

Adamant36:
Re "People with old hardware." I've been thinking for a while now that it would be cool of there was a way to toggle between basic and expert modes. Whatever those would entitled. Personally, I rarely use most of iD Editors more advanced features and it tends to slow my computer down a lot sometimes. Even though I have a fairly new hardware and a good internet connection. So a basic version that could run on older hardware  would be really great. 

BTW, I'm sure a lot of the preformance issues I have are because of Chrome's horrible memory management, but id still benefit from a basic version of iD Editor regardless. I can't imagine how it runs in Chrome on older machines.

PR #8835, opened 2021-12-02
Improve UI accessibility - Focus on screen reader support

mbrzakovic:
Goes one step forward towards a vision shared in #7713.
This change implements several [accessibility principles](https://www.w3.org/WAI/fundamentals/accessibility-principles/) with the focus on improving [screen reader support](https://github.com/openstreetmap/iD/blob/develop/ACCESSIBILITY.md#-vision-accessibility).
Additionally, few UI inconsistencies have been addressed as well.

Changes:
- Ensures elements have discerniable text - Provided missing titles and aria-label to several items (mostly buttons) + missing tooltip added for add tags [fef933a2a49ddd8bb7d6675e156281d79f7ca8d9 - 094ace65fa0fccda6e0a833c89f3350573e101f9]
image

- More descriptive titles 10fad143f25d1a9d0c66b9fb27ce50b26ac77ae6 02e92d9e38d6305809267e34f4399f2a38a98b13
- Few [image alts](https://www.w3schools.com/tags/att_img_alt.asp) added a77840eb2a8f851bb603763d1b40be66d27648b9
- [Button role ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)added where appropriate 4bee356fc1c08d57b620a4bdeda61840046fd9f4 50de029efa08b10c18bc7934a04ca71c1ff64002 041332045d18c28a94c77fe6b31699205d509d6f
- Ensure states are being announced - aria-selected/disabled/expanded/pressed etc. e91dc1b4f6f7d308f66361ae7d59fbb74b39d4eb 6337e4d769354b9b31538f3b29b978fd6ef9638e
- Polish up the headings (this impacts the speed at which assistive technologies allow people to orient around locations on page) 30ef0b80e1fe2d6f177ea31fa53a8bb393f71fac 8eba5db4e3391c5583daa04bec4f32856104e57f. Additionally, keyboard shortcuts heading has been centered, and sidebar headings have been increased, making the ui better looking and more intuitive.
image


modules/ui/sections/background_offset.js
```diff
@@ -148,13 +148,15 @@ export function uiSectionBackgroundOffset(context) {
...
             .attr('type', 'text')
+            .attr('aria-label', t('background.offset_label'))
             .on('change', inputOffset);
 
         nudgeWrapEnter
             .append('div')
             .selectAll('button')
             .data(_directions).enter()
             .append('button')
+            .attr('title', function(d) { return 'nudge ' + d[0]; })
```
tyrasd: would it be worth the effort to make these 4 strings translatable?

tyrasd (commented):
Thank you for the pull request. These are certainly good improvements. I've got only one question/comment, see below.

PS: I've fixed the merge conflicts caused by other changes in the development branch. Also, I've added a line to the changelog to this branch (see 960f5768c4b9a783ee9b9def2b899f4bb0501c4d). Feel free to amend it if you think something should be mentioned more specifically.

modules/ui/sections/background_offset.js
mbrzakovic: Good catch, I think the answer is: yes.
Majority of 'title' attributes support translation, and I think this is a good practice given that it is both displayed as tooltip and is used by screen reader.

tyrasd (approved):

PR #8837, opened 2021-12-03
Corrections to Privacy Policy

tyrasd:
This updates and clarifies a few points in the privacy policy.

### todo
* [ ] clarify section about the ideditor.com domain

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -1,6 +1,6 @@
 ## iD Editor Privacy Policy
 
-iD Editor is an open source OpenStreetMap editor. You are not required to use iD in order to use or edit OpenStreetMap, and the OpenStreetMap Foundation does not operate iD. You can learn more about OpenStreetMap and the OpenStreetMap Foundation at [https://www.openstreetmap.org/about](https://www.openstreetmap.org/about) and [https://wiki.osmfoundation.org/wiki/Main_Page](https://wiki.osmfoundation.org/wiki/Main_Page). You can learn more about the iD Editor project and its maintainers at [https://github.com/openstreetmap/iD](https://github.com/openstreetmap/iD).
+iD Editor is an open source OpenStreetMap editor. You are not required to use iD in order to use or edit OpenStreetMap. You can learn more about OpenStreetMap and the OpenStreetMap Foundation at [https://www.openstreetmap.org/about](https://www.openstreetmap.org/about) and [https://wiki.osmfoundation.org/wiki/Main_Page](https://wiki.osmfoundation.org/wiki/Main_Page). You can learn more about the iD Editor project and its maintainers at [https://github.com/openstreetmap/iD](https://github.com/openstreetmap/iD).
 
 ### Source code and website ([http://ideditor.com/](http://ideditor.com/))
 
@@ -38,6 +38,7 @@ If you believe that we hold information that would allow us to correct, amend, o
 
 ### Changelog
 
+* _2021-Dec-xx_ - Remove false statement which claimed that _the OpenStreetMapFoundation does not operate iD_.
 * _2020-Dec-02_ - Add information about the JSDelivr CDN integration.
 * _2020-Apr-07_ - Update the iD GitHub repository file links to point to the `release` branch instead of the `master` branch.
 * _2019-Dec-17_ - Initial version.
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -34,7 +34,7 @@ The Name Suggestion Index is a feature in iD to aid users in consistently and ac
 
 ### Contact
 
-If you believe that we hold information that would allow us to correct, amend, or delete inaccurate information about you or if you believe that information about you has been processed in violation of this privacy policy or applicable law, please email us at privacy@ideditor.org.
+If you believe that we hold information that would allow us to correct, amend, or delete inaccurate information about you or if you believe that information about you has been processed in violation of this privacy policy or applicable law, please email us at [legal-questions@osmfoundation.org](mailto:legal-questions@osmfoundation.org).
 
 ### Changelog
 


bhousel:
These edits make it sound a lot like that OSM Foundation have taken over control of the iD project? 

Have you been asked by your bosses at the OSM Foundation directly to make this change?  Can you share this decision with the public?

Is this a decision that we can reconsider (at least until the next board is seated in a few weeks and has a chance to meet?)

PRIVACY.md
```diff
@@ -34,10 +34,11 @@ The Name Suggestion Index is a feature in iD to aid users in consistently and ac
 
 ### Contact
 
-If you believe that we hold information that would allow us to correct, amend, or delete inaccurate information about you or if you believe that information about you has been processed in violation of this privacy policy or applicable law, please email us at privacy@ideditor.org.
+If you believe that we hold information that would allow us to correct, amend, or delete inaccurate information about you or if you believe that information about you has been processed in violation of this privacy policy or applicable law, please email us at [legal-questions@osmfoundation.org](mailto:legal-questions@osmfoundation.org)</a>.
```
1ec5: The link syntax includes some stray HTML.

PRIVACY.md
```diff
@@ -34,10 +34,11 @@ The Name Suggestion Index is a feature in iD to aid users in consistently and ac
...
 
 ### Changelog
 
+* _2021-Dec-xx_ - Remove false statement which claimed that _the OpenStreetMapFoundatino does not operate iD_.
```
1ec5: Not every instance of iD is run from OSMF servers. Does the privacy policy need to account for independent deployments, or is it up to them to provide their own privacy policies even if nothing is being customized?

1ec5 (commented):
According to #7040, the original privacy policy was written in close consultation with the LWG, which is the legal point of contact indicated in this document. Can we get confirmation that the LWG has signed off on the changes to the document, including the change description that seems to imply a retroactive change?

tyrasd:
> the original privacy policy was written in close consultation with the LWG, which is the legal point of contact indicated in this document. Can we get confirmation that the LWG has signed off on the changes to the document

I forgot to mention this in my original post, but yes: These changes had already been signed off by the LWG (//cc ~~@KathleenLD~~ @kathleenlu09 //edit: sorry for the mixup of usernames here).

mmd-osm:
uh oh, you've mentioned the wrong Kathleen, should be @kathleenlu09

tyrasd:
Sorry for my mistake @kathleenlu09 and @KathleenLD – and thanks for correcting me @mmd-osm

tyrasd:
Replying to https://github.com/openstreetmap/iD/pull/8837/files#discussion_r762489477:

It is true that not every instance of iD is fully run by the OSMF. However, the iD repository makes it quite clear that when one just says "iD", the instance operated by the OSMF is meant. See for example the website link on the [top right](https://github.com/openstreetmap/iD), the various links to osm.org scattered throughout the repository and the [releasing instructions](https://github.com/openstreetmap/iD/blob/develop/RELEASING.md#update-openstreetmap-website). Further, this is the privacy policy which is linked from the actual iD editor hosted at osm.org.

> Does the privacy policy need to account for independent deployments, or is it up to them to provide their own privacy policies even if nothing is being customized?

As far as I can see, the policy is written in a way that is quite deployment-independent (except for the contact email and where it mentions the ideditor.com website perhaps).

If anyone runs their own public instance of iD they are required (by law, I would think) to display a privacy policy which matches their version of iD that they run. If they don't customize iD in any way, the needed changes would be minimal.

1ec5:
This probably isn’t the best venue to hash out any kind of text with legal ramifications, but I’ll just point out that “operate iD” is ambiguous: there’s a desire on the one hand to avoid implying that the OSMF runs iD _development_, whereas it sounds like the passage is being reworded to allow that the instance of iD currently being used is deployed on OSMF servers.

tyrasd:
> Have you been asked by […] the OSM Foundation directly to make this change?

No. These changes were initiated by myself in an attempt to correct a mistake in the privacy policy.

> Is this a decision that we can reconsider (at least until the next board is seated in a few weeks and has a chance to meet?)

I think it doesn't hurt to wait a few weeks for this, sure.

tyrasd:
> avoid implying that the OSMF runs iD OSMF runs iD _development_

Well, one could argue that the OSMF nowadays also "operates" (in the sense of: _finances_) at least some of the iD development. But this is in my opinion irrelevant here, since the sentence in question needs to be read in the context of a [privacy policy](https://en.wikipedia.org/wiki/Privacy_policy), which is by definition not about development processes at all.

PRIVACY.md
bhousel: What we probably should have done in #7040 was put the privacy policy and its datestamp in the `dist/` folder in a .json file with the other iD assets, then have the display code look for that file and display its contents only if it exists and is newer than what has been displayed previously.  This would let other sites override it easily, or omit it entirely. 
 The iD embedded in other sites like openstreetmap.org or openhistoricalmap.org are covered by their respective privacy policies.

A lot of the privacy policy discussion leading up to and on #7040 didn't make any sense for iD (a client side app that collects no data, stores no data, and sets no cookies), but we tried.  (more info on privacy notices can be found [here](https://gdpr.eu/privacy-notice/))

bhousel:
> > the original privacy policy was written in close consultation with the LWG, which is the legal point of contact indicated in this document. Can we get confirmation that the LWG has signed off on the changes to the document
> 
> I forgot to mention this in my original post, but yes: These changes had already been signed off by the LWG.

I don't see anything in the LWG minutes about revising the iD privacy policy, so we must have a different idea of what "signed off by the LWG" means. 
https://wiki.osmfoundation.org/wiki/Working_Group_Minutes#2021

manfredbrandl:
@tyrasd Can you please provide information about LWG signing off the changes?

tyrasd:
> Can you please provide information about LWG signing off the changes?

This came up in an email conversation following a question I sent to legal@osmfoundation.org. Perhaps this doesn't count as "signing off": if there is a better way, please let me know.

manfredbrandl:
@tyrasd Can you please share the relevant parts of the email conversation here?

simonpoole:
It should be noted that there is now some legal precedence, see https://www.gesetze-bayern.de/Content/Document/Y-300-Z-GRURRS-B-2022-N-612 , that providing javascript that is run on/in a client does in fact fall in the responsibility of the operators of the website that embeds the javascript. In particular any third party content loaded by such code needs either explicit consent or a DPA covering it (retrieving any content from the US even with consent would seem to be associated with legal jeopardy).

PRIVACY.md
```diff
@@ -34,10 +34,11 @@ The Name Suggestion Index is a feature in iD to aid users in consistently and ac
 
 ### Contact
 
-If you believe that we hold information that would allow us to correct, amend, or delete inaccurate information about you or if you believe that information about you has been processed in violation of this privacy policy or applicable law, please email us at privacy@ideditor.org.
+If you believe that we hold information that would allow us to correct, amend, or delete inaccurate information about you or if you believe that information about you has been processed in violation of this privacy policy or applicable law, please email us at [legal-questions@osmfoundation.org](mailto:legal-questions@osmfoundation.org).
```
simonpoole, line 37: IMHO, you need to check with the LWG, but at least historically privacy@osmfoundation,org has been the contact address for privacy related matters.

PRIVACY.md
```diff
@@ -1,6 +1,6 @@
 ## iD Editor Privacy Policy
 
-iD Editor is an open source OpenStreetMap editor. You are not required to use iD in order to use or edit OpenStreetMap, and the OpenStreetMap Foundation does not operate iD. You can learn more about OpenStreetMap and the OpenStreetMap Foundation at [https://www.openstreetmap.org/about](https://www.openstreetmap.org/about) and [https://wiki.osmfoundation.org/wiki/Main_Page](https://wiki.osmfoundation.org/wiki/Main_Page). You can learn more about the iD Editor project and its maintainers at [https://github.com/openstreetmap/iD](https://github.com/openstreetmap/iD).
+iD Editor is an open source OpenStreetMap editor. You are not required to use iD in order to use or edit OpenStreetMap. You can learn more about OpenStreetMap and the OpenStreetMap Foundation at [https://www.openstreetmap.org/about](https://www.openstreetmap.org/about) and [https://wiki.osmfoundation.org/wiki/Main_Page](https://wiki.osmfoundation.org/wiki/Main_Page). You can learn more about the iD Editor project and its maintainers at [https://github.com/openstreetmap/iD](https://github.com/openstreetmap/iD).
```
simonpoole, line 3: The legal responsibility for the instance of iD the user is running lies with the operator of the website that is embedding the code (see comment from me on the PR). With other words for the iD instance from openstreetmap.org with the OSMF, for RapiD with Facebook and so on. With other words this text should probably be configurable per installation/build.

Issue #7795, opened 2020-07-16
object history handling while split way function 


sun-geo:
![image](https://user-images.githubusercontent.com/47290585/87686117-6e65d780-c784-11ea-9818-5ca514b661b1.png)

It seems the current process and handling of the object histories (of a line/way element) is like the first part depending on the osm-way direction gets a +1 higher version number and the second part gets a new way-id with version#1. 

first-part:
![image](https://user-images.githubusercontent.com/47290585/87685889-2ba3ff80-c784-11ea-8573-bf93b2adf398.png)
second-part:
![image](https://user-images.githubusercontent.com/47290585/87685907-31014a00-c784-11ea-8953-8327b908793a.png)

So the current only ability of manual intervention of history handling of the two new way-parts is to reverse the osm-way direction of the initial way object before splitting.

![image](https://user-images.githubusercontent.com/47290585/87686487-d5838c00-c784-11ea-906e-c9a2e7b7cfc3.png)

Today a had a short trip to JOSM-editor (only because, I'm still waiting a bit for #6539 - undelete object function ;-)  ), and I was facing an interesting message while the Split-Way function: It asks me for:
![image](https://user-images.githubusercontent.com/47290585/87685236-73765700-c783-11ea-83ee-ccc0d888a480.png)

So it seems it is not needed (to work through the workaround - with potential failure producing when missing to reverse back in case of a oneway road) to reverse the osm-way direction and you can decide while splitting which part will get a new way-id.
Is this something which could please also be implemented for iD?

quincylvania:
@sun-geo Hmm this is a type of UI that sounds great for JOSM, but I don't think suits iD's usability goals. I'm assuming very few iD mappers know about entity IDs and version numbers, so asking them this question would be unhelpful. However, iD could be smarter when deciding which part to increment rather than simply picking the first part.

sun-geo:
Thanks for checking and thinking about.

> .... I'm assuming very few iD mappers know about entity IDs and version numbers, so asking them this question would be unhelpful.

I agree.

What about an idea of a bit hidden function to add instead a new context menu item (after right click) only active for the second way part (w-1) (which has no object id & version yet). Let's say a function which is called "Get history"- (get the version history information from the other split way segment). So, the average iD user would not bothered with confusing questions about history & versioning. The users, who know about the OSM history and version concept, would have the possibility to adapt the object histories and would have it quite easy to respect the achievements of other users who had previously worked on these objects.

[...]

PR #8839, opened 2021-12-05
Object history preservation improvements

### LINKED ISSUE #8519: cutted way filled in route in oposite direction
We have lots of errors in bus-route relations from iD-Editor-users. The editor seems to look sensible for the way to paste the cutted way, but then uses the oposite direction to paste.
Example: way/334914447 relation/1732028 (going east) and relation/1732029 (going west)
changeset/105406069  I cut the way in two ways way/334914447 and way/947627779
In result both relations are sorted in the wrong order of ways at way/334914447 and way/947627779.

tpetillon:
This PR extends what has been done in #7795 and #8708. It improves object ID preservation in the following cases:

* merging nodes,
* merging polygons,
* merging nodes into a way.

To help doing this, the PR adds some utility functions dealing with object IDs.

It also fixes the order of relation members in some cases when splitting ways. (Depending on which way was the longest after the split, the two ways resulting from the split action could be inserted in the wrong order in the relation.)

Tests are provided for these actions.

### DIFF
--- a/modules/actions/add_member.js
+++ b/modules/actions/add_member.js
@@ -32,7 +32,7 @@ export function actionAddMember(relationId, member, memberIndex, insertPair) {
     // Add a way member into the relation "wherever it makes sense".
     // In this situation we were not supplied a memberIndex.
     function addWayMember(relation, graph) {
-        var groups, tempWay, item, i, j, k;
+        var groups, tempWay, insertPairIsReversed, item, i, j, k;
 
         // remove PTv2 stops and platforms before doing anything.
         var PTv2members = [];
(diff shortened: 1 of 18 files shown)


tyrasd (commented):
Wow, what a great PR. Huge thanks! :bow: 

My only question/remark would be if it was really necessary to introduce the mew method `osmEntity.id.isNew(id)` in addition to `osmEntity.isNew()`. I don't immediately see why there would be a need for both methods. Wouldn't it be sufficient for this PR to simply call `isNew()` on the respective entity objects directly?

tpetillon:
You’re right, it wasn’t needed! I didn’t realise that while cleaning the code. I have removed it, and rebased the PR.

PR #8844, opened 2021-12-09
Fix incorrect Google validation warnings when viewing esri Africa buildings

Bonkles:
If viewing areas that include buildings from the ESRI africa building dataset, folks might see a validation error declaring 'Google products are proprietary and must not be used as references'.  This validation does NOT apply to the Africa buildings dataset since it is licensed correctly and approved for use in OSM.

Related RapiD issue: https://github.com/facebookincubator/RapiD/issues/403:

Image with the validation:  
image

This fix adds a small exception to the regex and a quick test to ensure it's correct.

### DIFF
--- a/modules/validations/incompatible_source.js
+++ b/modules/validations/incompatible_source.js
@@ -17,7 +17,7 @@ export function validationIncompatibleSource() {
     {
       id: 'google',
       regex: /google/i,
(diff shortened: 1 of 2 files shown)


tyrasd:
Thanks!

PR #8881, opened 2022-01-07
consider lifecycle prefixes in osmTagSuggestingArea

k-yle:
Ways with lifecycle prefixes as always rendered as a line in iD, even if the feature would normally be treated as an area (for example `disused:amenity=school` or `abandoned:leisure=stadium`).

This PR makes iD consider lifecycle prefixes when deciding whether a way is a line or an area.

Example for `proposed:building=yes`:

image


modules/osm/tags.js
```diff
@@ -13,6 +13,33 @@ export function osmSetAreaKeys(value) {
...
 
+export const osmLifecyclePrefixes = {
+    // nonexistent, might be built
+    proposed: true, planned: true,
+    // under maintentance or between groundbreaking and opening
+    construction: true,
+    // existent but not functional
+    disused: true,
+    // dilapidated to nonexistent
+    abandoned: true, was: true,
```
1ec5, line 19: There’s also [`not:`](https://wiki.openstreetmap.org/wiki/Key:not:), which is mostly used for `not:name` but could also be used on feature keys too.

modules/renderer/features.js
```diff
@@ -210,7 +199,7 @@ export function rendererFeatures(context) {
 
         for (var i = 0; i < strings.length; i++) {
             var s = strings[i];
-            if (past_futures[s] || past_futures[tags[s]]) { return true; }
+            if (osmLifecyclePrefixes[s] || osmLifecyclePrefixes[tags[s]]) return true;
```
1ec5, line 202: Does it make sense that `intermittent:*` would be filtered out by the past/future filter? (Also `not:*`, discussed above.)

modules/osm/tags.js
```diff
@@ -13,6 +13,33 @@ export function osmSetAreaKeys(value) {
...
+    // nonexistent, still may appear in imagery
+    dismantled: true, razed: true, demolished: true, destroyed: true, removed: true, obliterated: true,
+    // existent occasionally, e.g. stormwater drainage basin
+    intermittent: true
+};
+
+/** @param {string} key */
+export function osmRemoveLifecyclePrefix(key) {
+    const keySegments = key.split(':');
```
1ec5, line 28: `String.prototype.split()` accepts a `limit` parameter; passing in `1` would simplify this function a little.

1ec5 (commented):
Great idea!

modules/osm/tags.js
k-yle: does it make sense to include `not:` in this list though? 

Considering a closed way tagged with `barrier=wall` + `not:amenity=school`:

- It should be treated as a line, not an area (?)
- It should not be included in the past/future filter
- It should not be rendered with the dashed line like abandoned/disused features.

modules/renderer/features.js
k-yle: not sure what the right answer is for `intermittent:*` is, but bear in mind that [`intermittent:*` is only used once](https://taginfo.osm.org/search?q=intermittent:) in the world (cf. other prefixes like [`disused:*`](https://taginfo.osm.org/search?q=disused:) are used >200,000 times). It seems like `intermittent=yes` is the more common tag.

modules/osm/tags.js
k-yle: didn't know that :) but I don't quite understand how that would simplify this function

modules/osm/tags.js
1ec5: You’re right, I was thinking of the case of something that isn’t a building but isn’t anything else mappable either, such as [this patch of pavement](https://www.openstreetmap.org/way/658082823) that isn’t a building. It’s probably rare enough not to worry about yet.

modules/osm/tags.js
1ec5: I was thinking that you could limit the split to 1 to have the part before the first colon and everything after that first colon. Then you wouldn’t need to deal with indices. But what you have does work.

modules/renderer/features.js
1ec5: Looks like the `intermittent` entry originated from 494e247ad1ec458d0d2bed6102dd5deb90eb11aa. It was probably intended to support `intermittent=yes`, not `intermittent:*=*`.

modules/osm/tags.js
k-yle: gotcha, sadly JS's `split` function isn't as nice as languages like ruby

 `"a:b:c".split(":", 1)` just returns `["a"]`, not `["a", "b:c"]`

modules/renderer/features.js
k-yle: Thanks for finding that. I guess it would be logical for the past/future filter to match all the lifecycle prefixes except `intermittent`

aygo162:
Thanks

PR #8701, opened 2021-09-17
Add warning for some commercial mapservice in China

bhousel:
closes #8700 

This PR adds some additional checks to the `incompatible_source` validator to catch some popular map services in China.

@LaoshuBaby - please take a look at the rulesets and let me know if I caught everything (I'm not a native Chinese speaker, but I tried).

It also rewrites the file to be ES6 and simplifies the code a bunch.  The new code also embeds the actual matched string in the error message, and it does a better job picking out sources from a semicolon-delimited list and uses the `.hash` property to disambiguate the issues raised. So for example if someone did say `source=google;baidu` then we'll see 2 issues raised, not 1.

Screen Shot 2021-09-17 at 11 42 31 AM

PR #8889, opened 2022-01-09
Remove accidental mismatch of "PAMAP"

cicku:
image

### DIFF
--- a/modules/validations/incompatible_source.js
+++ b/modules/validations/incompatible_source.js
@@ -8,7 +8,7 @@ export function validationIncompatibleSource() {
   const incompatibleRules = [
     {
       id: 'amap',
-      regex: /(^amap$|amap\.at|^amap\.com|autonavi|mapabc|高德)/i
+      regex: /(^amap$|^amap\.com|autonavi|mapabc|高德)/i
     },
     {
       id: 'baidu',
--- a/modules/validations/incompatible_source.js
+++ b/modules/validations/incompatible_source.js
@@ -8,7 +8,7 @@ export function validationIncompatibleSource() {
   const incompatibleRules = [
     {
       id: 'amap',
-      regex: /(amap|autonavi|mapabc|高德)/i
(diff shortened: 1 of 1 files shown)


mmd-osm:
https://www.openstreetmap.org/way/166528135 would no longer be flagged by this change, although it might be still incompatible data (edit: to be verified!).

Adding more tests in https://github.com/openstreetmap/iD/blob/develop/test/spec/validations/incompatible_source.js might be good.

cicku:
I don’t know anything about Europe so I can’t help here. As far as I can see it’s a regression introduced by a patch to warn about proprietary Chinese data.

modules/validations/incompatible_source.js
```diff
@@ -8,7 +8,7 @@ export function validationIncompatibleSource() {
   const incompatibleRules = [
     {
       id: 'amap',
-      regex: /(amap|autonavi|mapabc|高德)/i
+      regex: /(^amap$|amap\.at|^amap\.com|autonavi|mapabc|高德)/i
```
tyrasd: Can you please drop `amap.at` from this list? It doesn't have anything to do with the data from [AutoNavi](https://en.wikipedia.org/wiki/AutoNavi), which this rule is [about](https://github.com/openstreetmap/iD/pull/8701#issuecomment-921922242). The amap.at domain is owned by an Austrian government agency.

@mmd-osm if you'd like to see this URL added to the list of incompatible sources, please open a new PR (or issue) for it. :blush: At first glance, amap.at seems to be a former, now dysfunctional, URL of what is now available at http://www.austrianmap.at/amap/, which AFAICS does contain non-open data (e.g., `ÖK 50`).

tyrasd (changes requested):

modules/validations/incompatible_source.js
cicku: Done

tyrasd (approved):

PR #8907, opened 2022-01-20
[Issue8906] Fix broken html in the Area part of the Walkthrough

### LINKED ISSUE #8906: The Area part of the walkthrough is broken
### Description

This issue is present on the [develop ](https://ideditor.netlify.app/) version only. The live version is working well. The broken part is with yellow on the below image.

### Screenshots

image

hodigabi:
This PR would fix [issue 8906](https://github.com/openstreetmap/iD/issues/8906).

I removed `isFirstSession` property from the core context.js because it was set but never used. The Walkthrough is anyway displayed based on [this](https://github.com/openstreetmap/iD/blob/97ed56bc3a93a6f06d0e373bc473e1a3d054eed1/modules/ui/init.js#L435) and [this](https://github.com/openstreetmap/iD/blob/97ed56bc3a93a6f06d0e373bc473e1a3d054eed1/modules/ui/splash.js#L25) conditions.

I tested the fix and it is working for me, even with localisation:
fix

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -8,7 +8,6 @@ import { t } from '../core/localizer';
 
 import { fileFetcher } from './file_fetcher';
 import { localizer } from './localizer';
-import { prefs } from './preferences';
 import { coreHistory } from './history';
 import { coreValidator } from './validator';
 import { coreUploader } from './uploader';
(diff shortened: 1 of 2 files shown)


tyrasd:
Thanks!

Issue #8871, opened 2022-01-03
Name fields that are disabled by having wikidata do not disable their language dropdown


RudyTheDev:
As per title.

Repro:

```
name=The 123
name:de=Das 123
wikidata=Q123
```

Then try to click dropdown arrow and select a different language

![lang](https://user-images.githubusercontent.com/3857299/147938998-85f66ec6-a34a-4f28-905a-ad779ffcda65.png)

Observe the `name:de` changing to new language.

Tried in both Firefox and Chrome.

tyrasd:
fixed in #8912, thanks @hodigabi

PR #8912, opened 2022-01-23
[Issue8871] Do not open disabled localized combobox when clicking on it

hodigabi:
This PR is a potential fix to #8871. If there is a better way to fix the problem, I'm open for suggestions.

Changing the localized.js was necessary only to fix the cursor of the small unfold 'icon' on the right side of the dropdown.


hodigabi:
The example I used is [this bridge](https://www.openstreetmap.org/edit?editor=id#map=19/47.49861/19.04404)

PR #8913, opened 2022-01-24
render paths under construction thinner than roads

### LINKED ISSUE #8743: render paths under construction thinner than roads
iD normally renders paths thinner than roads, but for roads/paths under construction the width is the same regardless of whether it is a road or path.

It would be nice if `highway=construction` + `construction=path/footway/cycleway/bridleway/steps` was rendered thinner than `highway=construction` + `construction=(any other value)`

Perhaps something like this:

<table>
<tr>
	<td><strong>Before
	<td><strong>After
<tr>
	<td>
	<td>
</table>

k-yle:
Closes #8743

This PR renders  `highway=construction` + `construction=path/footway/cycleway/bridleway/steps` thinner than `highway=construction` + `construction=(any other value)`

Example:

<table>
<tr>
	<td><strong>Before
	<td><strong>After
<tr>
	<td>
	<td>
</table>

<details>
<summary>Same Example in OSM Carto</summary>

image

</details>

### DIFF
--- a/css/50_misc.css
+++ b/css/50_misc.css
@@ -425,6 +425,32 @@ path.line.stroke.tag-highway.tag-status.tag-status-construction,
     stroke-dasharray: 8, 8;
 }
 
+/** Closed Paths */
+path.line.shadow.tag-highway.tag-status.tag-status-construction.tag-construction-path,
+path.line.shadow.tag-highway.tag-status.tag-status-construction.tag-construction-footway,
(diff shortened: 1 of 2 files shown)


tyrasd:
FYI: In 55d38eba6 I've extended this also to `highway=proposed` + `proposed=path/…` and tweaked the "shadow" width a bit to make it similar to other highlighted features.

PR #8914, opened 2022-01-24
prefer the last used imagery over the 'best' imagery

### LINKED ISSUE #8665: Remember last used imagery
Everytime I open iD, the imagery is set to default, although it wasn't always like this.
Is it possible to allow iD to load the last used imagery?

### LINKED ISSUE #11626: Selected background is not restored when starting
Hi,
in some former versions the selected background imagery always was saved (maybe in a cookie) when leaving iD and restored when logging in again and starting a new editing session. Since a few months this doesn't work anymore. I prefer to select "Esri World Imagery". But now each time I start a new session, always the first background imagery is selected (which for me is "Niedersachsen DOP20" and I don't like). Is it possible to get back the restoring feature again?
BR,
tkansgar

### LINKED ISSUE #11981: New user setting: Default Background map
### Description

Hi! Would be lovely to choose my default background map. For example I would like to have Openstreetmap as default. Currently it feels like aerial is often opened by default. 

Setting could be added to https://www.openstreetmap.org/preferences/basic

### Screenshots

_No response_

k-yle:
Closes #8665, Closes #11626, Closes #11981

Note that this is only a bug in the `iframe`'d version of iD on [openstreetmap.org](https://openstreetmap.org), so it's a bit tricky to test this PR. The bug is also only present in regions that have a 'best' imagery source.

To test:
 - go to http://localhost:8080/#map=19.50/-36.81455/174.80295
 - the 'best' imagery will be selected, so select a different aerial imagery source
 - close the tab, open http://localhost:8080/#map=19.50/-36.81455/174.80295 in a new tab
 - note how the last-used imagery is selected

Current behaviour:
 - go to https://ideditor.netlify.app/#map=19.50/-36.81455/174.80295
 - the 'best' imagery will be selected, so select a different aerial imagery source
 - close the tab, open https://ideditor.netlify.app/#map=19.50/-36.81455/174.80295 in a new tab
 - note how the best imagery is selected (this is unexpected)

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -483,8 +483,8 @@ export function rendererBackground(context) {
         } else {
           background.baseLayerSource(
             background.findSource(requested) ||
-            best ||
             background.findSource(prefs('background-last-used')) ||
+            best ||
             background.findSource('Bing') ||
             first ||
             background.findSource('none')


AntMadeira:
Any chance of prioritizing this? 
This is a big problem without resolution for such a long time...

tyrasd:
:thinking: I'm hesitating because this "fix" will have the major downside that after any (global) background layer was been chosen, from that point onward, there would never be a situation where the `best` imagery is preferred over the "last used" layer. The intention of the `best` flag is however to have layers which are known to be better than the global default layers in their respective regions. Therefore they should be actually get precedence over other layers.

Surely, there are situations where a user manually selects a non-`best` layer in a specific region for a specific reason and would prefer that this layer is loaded on further sessions in this area (but not in other areas where a different `best` layer exists). Unfortunately, it's not trivial to implement a mechanism which would remember that user's preference on a per-region basis (and/or on a per overridden `best` layer basis). For now, there exists the following workaround for users: bookmark a specific location requesting the preferred background layer via a hash parameter, e.g. `http://localhost:8080/#background=Maxar-Premium&disable_features=boundaries&map=19.50/-36.81455/174.80295` or `https://www.openstreetmap.org/edit#map=19/-36.81455/174.80295&background=Maxar-Premium`

PR #8975, opened 2022-02-09
Fix: Not visible imagery names + input proportion

furkanmutlu:
This PR fixes not visible background imagery names and not proportional radio buttons for backgrounds panel.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -3226,14 +3226,11 @@ div.full-screen > button:focus {
 .layer-list label > span {
     display: block;
     overflow: hidden;
-    white-space: nowrap;
-    text-overflow: ellipsis;
-    flex-grow: 1;
+    width: calc(100% - 20px); /* Ensures radio input width within flexbox */
 }
 
-.layer-list.layer-background-list label > span {
-    width: 100%;
-    white-space: normal;
+.layer-list label span.localized-text {
+    line-height: 0.95rem;
 }
 
 .layer-list input.list-item-input {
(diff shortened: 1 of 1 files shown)


tordans:
Hi! The way I read the CSS is, that the whole point of the span-tag is to add the css that prevents the line from breaking. Instead of adding more css on top that re-writes this, I would be cleaner to not use the span here or refactor the span into two `<span class="one-line">…` vs. `<span class="multiline">…` (or similar).

For the multiline, I suggest to also add something like `line-height: 0.95rem;` to fix the line-spacing.

---

Also: Adding a screenshot of the new look makes it easier to understand most PRs.

furkanmutlu:
Hi @tordans, thank you for your feedback. I think refactoring for two types of spans (one-line / multi-line) would be overkill for this issue. I found simpler solution by removing white-space, text-overflow and flex grow from the span. And set new width for it by removing paddings around to ensure radio input size doesn't change. Also added line-height as you mentioned. Please find the screen shot of new look below:
image

furkanmutlu:
@tordans Could you please take a look at this when you have time ? Thanks

tordans:
Hi @furkanmutlu it is likely frustrating to not get any feedback on this; it would be for me. However, the right person to respond would be the maintainer of the project or someone with merge permission. That would also be a good person to decide if more refactoring would result in a cleaner codebase (my last feedback) or if this current solution is preferred.

furkanmutlu:
@tordans thank you for your quick response

PR #8978, opened 2022-02-10
fix #8750 - Follow operation should prefer the focused way

### LINKED ISSUE #8750: Follow operation should prefer the focused way

### LINKED ISSUE #11489: Allow continuing drawing using follow function if two ways/areas have the same geometry
### Description

Example:
Image

### Screenshots

_No response_

Pho3b:
After studying the ID software architecture for a bit i came up with this solution for #8750.

I performed various test on my local machine and it seems to cover the cases described in the issue.
Unit tests were started, and no errors were found.

Hope it can be helpful, i will be available for any further updates.


tordans:
Ping https://github.com/openstreetmap/iD/issues/8750

tyrasd (changes requested):
Hey. Thanks for the pull request. 

While testing it, I noticed some strange behavior: when following another way which ends in a T-junction, the appended way jumps back to the first contact with the followed way and repeats in a circle infinitely. I think this is a bug, isn't it?



Can you please explain in more detail how you think the follow operation should work in situations like this (and in general). Thanks!

Pho3b:
hey @tyrasd thank you very much for the code review.

I executed some test and, correct me if i am wrong, it seems to me that i could reproduce the behavior you reported even without my code changes. 
What i noticed is that now, not stopping when a node that has multiple parent is encountered, lets that behavior to show up more often.

That said, i just pushed a new commit that ends the `following` action if a node that has already been followed is encountered again, this should avoid the user to end up following infinite circles.

I hope i correctly answered your question, waiting for your feedback.
Thanks!

Pho3b:
Hey guys, any news on this ? 
If there's there something i can do to simplify or speed up the review process just let me know

Thanks

tyrasd:
Sorry, I haven't had time to give it a deeper look yet. I'll try to come around it next week. :)

Pho3b:
Hi, any updates regarding this PR ?

Thank you

DujaOSM:
I objected on implementing this in #8750, since it would hamper usability in a significant number of use cases.

tyrasd:
Closing in favour of #9340. (also: Mixing keyboard and mouse hover at the same time seems also a bit counter-intuitive from a usability point of view.)

Thanks for your effort anyway! :bow:

PR #9020, opened 2022-03-08
create a 'divide' operation

### LINKED ISSUE #6988: Split building into 2 identical connected buildings
Currently mapping semi-detached houses (which are very common in the UK and some other places) is a process which could be better optimized.

Currently if you wanted to add a semi-detached building, you'd use buildings drawn from satellite imagery (and this is probably the easiest way to add buildings), then add new nodes to the half way point, delete the other half of the building, then copy and reflect the building, then try to line it up to connect with the other house and attach the connecting nodes.

It's long, annoying for a lot of buildings, and inaccurate.
My solution to deal with this would be a split vertically/split horizontally option on the right click, similar to the existing reflect options.
This would make the process as simple as add house, square, split, repeat.

k-yle:
This PR creates a "divide" operation, similar to the [JOSM gridify plugin](https://github.com/JOSM/josm-gridify). It lets you divide an area into a grid of smaller areas. This is useful for mapping [parking spaces](https://wiki.osm.org/Tag:amenity=parking_space), [terraced houses](https://wiki.osm.org/Tag:building=terrace), and [camp pitches](https://wiki.osm.org/Tag:tourism=camp_pitch). 

You can try it out [here](https://kyle.kiwi/iD). Select an area and press <kbd>K</kbd>. 

Closes #6988

https://github.com/user-attachments/assets/e9897ce2-e690-4ca3-b68b-f0c5d54de3e0

<details>
<summary>More Examples</summary>

image

<table>
<tr>
	<td>house1
	<td>house2

<tr>
	<td>camp1
	<td>camp2
</table>

</details>


1ec5:
Nifty, this will tempt me to start nanomapping the individual modules of solar panels. 😉

I wonder if it would be possible to replace the two prompts with something more interactive. Perhaps you hit the shortcut key, then move the mouse along the long and short axes to control the number of divisions. It’d be similar to how you can scrub the mouse cursor while rotating a feature, except that this gesture would snap to widths that evenly divide the area. For things like parking spaces, it’d be more convenient to eyeball the right width than to count the divisions manually.

> Since this operation is only useful for a small number of features it should possibly be hidden by default?

If we expect this operation to be used somewhat less frequently than, say, the merge operation, how about keeping it available in the context menu for all areas but avoid assigning a keyboard shortcut? It’s becoming ever more difficult to assign memorable single-key shortcuts without modifiers, especially in locales that translate them. Currently, operations’ availability depends on the selected geometry type, but it would be more difficult to communicate to the user that this operation depends on certain tags.

k-yle:
> If we expect this operation to be used somewhat less frequently than, say, the merge operation, how about keeping it available in the context menu for all areas but avoid assigning a keyboard shortcut?

That's a good idea - my concern was also that it will clutter the context menu, which already has 8 items, but maybe that's not a problem...

> I wonder if it would be possible to replace the two prompts with something more interactive.
> Perhaps you hit the shortcut key, then move the mouse along the long and short axes to control the number of divisions.

👍 I'll give this some thought, another option could be to replace the two prompts with a modal. That modal could have a preview similar to the turn-restriction editor. Both options will be quite complicated.

kymckay:
Probably out of scope for this PR, but may be worth considering as a follow up: Would be nice if this supported use of custom dividing line (something like: draw way across area, select both, perform divide operation). Useful for things like drawing a footprint and splitting into non-uniform buildings or building parts, or splitting land use to re-tag a section.

1ec5:
> Would be nice if this supported use of custom dividing line (something like: draw way across area, select both, perform divide operation). Useful for things like drawing a footprint and splitting into non-uniform buildings or building parts, or splitting land use to re-tag a section.

Vector drawing tools have a similar operation. It isn’t the most discoverable, but it makes a lot of sense once you’re made aware of it, similar to how iD supports punching holes in polygons using the merge operation. In that case, you could even hook it up to the split operation, so you can select an area and a line that cuts across it, but don’t select a vertex, then split the area along the line. It would be quite outside the scope of this PR, but it might be worth considering how to make the UX for this feature consistent with what might be possible for that other feature.

hb0nd:
This is looking superb! It'd be nice if it could split based on the bbox instead of the actual nodes, but that can be implemented after this :)

A case where splitting with bbox would be useful:
Screenshot 2022-04-01 at 21 26 57

modules/operations/divide.js
```diff
@@ -0,0 +1,114 @@
...
+            _extent = _extent.extend(entity.extent(context.graph()));
+        }
+
+        return actionDivide(entityID, context.projection);
+    }
+
+    const operation = () => {
+        if (!_action) return;
+
+        const longLength = +prompt('Long Length');
```
tyrasd: All texts should be [translatable](https://github.com/openstreetmap/iD/wiki/How-to-use-iD%27s-localizer-module).

modules/operations/divide.js
```diff
@@ -0,0 +1,114 @@
...
+        if (!longLength) return;
+        const shortLength = +prompt('Short Length');
```
tyrasd: Unfortunately, `prompt` is not great, for example because language metadata cannot be supplied (see #7963). We would need to find a better solution for this.

data/shortcuts.json
```diff
@@ -1,386 +1,391 @@
...
-            "modifiers": ["⌘", "⇧"],
-            "shortcuts": ["info_panels.measurement.key"],
-            "text": "shortcuts.tools.info.measurement"
-          }
-        ]
-      }
-    ]
-  }
-]
+    {
```
tyrasd: please revert the extra indentation in `shortcuts.json`

tyrasd (commented):
I found the following issues / corner cases while playing around with it:

1. if the input geometry is concave, the result will have intersecting and/or self-intersecting geometries: 
    
2. when applied to an area which shares an edge with a neighboring area (or touching line), the neighboring (or touching) feature does not receive a vertex at the newly introduced split position (which potentially also results in a validator warning messages about intersecting geometries) 
    image

In general, I would be slightly more inclined to add support "divide operation" which is more flexible, as outlined in https://github.com/openstreetmap/iD/pull/9020#issuecomment-1062333807, rather than specialized operation which only properly works in a narrow set of use cases.

danieldegroot2:
It would be nice if we could also do this with ways, for if we know there's windows/doors or extensions along a building wall.

xnousnow:
It's been 2 years... anything going on?

tordans:
Looking at the initial video, I wonder...
- could the map be displayed in the modal for reference
- would it be helpful to see the area size and or length of the rectangle somewhere for reference? Like with the parking spaces, any of the last two divisions look OK, but which one is right...
- does the modal create one entry in the history or one per rectangle? Ideally it where possible to revert the division with just one cmd-z to try again with different settings.

matkoniecz:
Looking at the initial video, I really think you should be able to see background while doing it.

Currently you need to count these parking spaces, and you have no preview. And this modal has styling not really fitting other parts of iD.

I guess that shortcut only trigger will be a bit mysterious, but it is better than putting it into prominent space.

data/core.yaml
```diff
@@ -135,6 +132,20 @@ en:
       already_circular:
         single: This can't be made more circular than it already is.
         multiple: These can't be made more circular than they already are.
+    divide:
+      title: Divide
+      key: K
+      annotation: Divide an area into a grid of smaller areas.
+      tooltip: Divide this area into a grid of smaller areas.
+      disabled:
```
matkoniecz, line 140: what about cases when area to be divided is a member of relation? say, multipolygon?

k-yle: when you divide an area, one of the new sub-areas is the 'survivior', meaning that this subarea inherits the original area's ID and parent relations (to preserve history)

matkoniecz: maybe it would be better to block division in such case? It sounds to me like it would in general destroy data

PR #9021, opened 2022-03-08
Fix validator reset bug - Deferred promise proper cleanup

mbrzakovic:
Special case can occur in validator, if there are any pending deferred validations waiting to be triggered by [window idle callback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) and validator.reset is called, which cancels the idle callbacks, however does not resolve pending promises which ultimately cause _headPromise = null [statement](https://github.com/openstreetmap/iD/blob/139638345f843e38756ecaa93cca3d037e7979fe/modules/core/validator.js#L510) to not be executed, therefore causing each next validate to not take effect due to [linked condition](https://github.com/openstreetmap/iD/blob/139638345f843e38756ecaa93cca3d037e7979fe/modules/core/validator.js#L488).

This PR handles such special case by implementing proper clean-up of hanging Promises.

This bug can be reproduced in iD in following way:
1) Click on Line/Area mode
2) Add one point
3) While still in drawing mode, go to Help -> Start walkthrough, or on dev instance click the 'live' switch (dev/prod) - Both would trigger validator.reset.
- From this point on, whatever is done would not get validated.
This can be double checked by adding a Line and not populating its type.

### DIFF
--- a/modules/core/validator.js
+++ b/modules/core/validator.js
@@ -22,7 +22,7 @@ export function coreValidator(context) {
   let _completeDiff = {};                     // complete diff base -> head of what the user changed
   let _headIsCurrent = false;
 
-  let _deferredRIC = new Set();   // Set( RequestIdleCallback handles )
+  let _deferredRIC = {};          // Object( RequestIdleCallback handle : rejectPromise method )
   let _deferredST = new Set();    // Set( SetTimeout handles )
   let _headPromise;               // Promise fulfilled when validation is performed up to headGraph snapshot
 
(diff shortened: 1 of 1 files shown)


bhousel (approved):
Great catch!  This looks good to me.  I was not thinking about having that Promise chain to other things, so rejecting is definitely better than resolving.

I'd probably use a `Map` to hold the RIC handle : rejectfunction, but using a plain Object is fine too.

Issue #546, opened 2013-01-28
Revise label stack


ansis:
Current label stack only has a bunch of tags I added when testing.
https://github.com/systemed/iD/blob/master/js/id/svg/labels.js#L4-L11

Format:
Labels are placed in the order of the entries. Each entry has a geometry, key, and optional value.

PR #9061, opened 2022-04-04
always use the `ref` as the label

### LINKED ISSUE #9054: Render golf=hole ref in edit mode
### Description

I have recently been auditing golf courses mainly for where folks have added 'name' tags to tees, fairways, bunkers, golf holes. By far the largest error is golf=hole having something like '18th' or '18th Hole' in the name field, i suspect so it shows in the editor interface. Just about all also had the correct hole number as ref. It seems to me folks are adding the name tag just to orientate themselves around the course whilst **editing**. This is the obvious way to orientate yourself on a golf course (I am not a player)

Would it be beneficial to display the ref tag as the label in the edit interface (same as the standard OSM render)? I think this would prevent many, many folks from also adding a name tag as well.

### Screenshots

golf_1
Render with incorrect use of names (forget the named tees etc)

golf_2
as incorrectly named objects look in editor - easy to orientate yourself

golf_3
how a correctly tagged course looks in the editor - harder to orientate around the course

golf_4
Correctly tagged holes  without 'name' tagging (forget the tees etc)

k-yle:
Closes #9054

If a feature has a `name` tag, iD will always use the `name` as the label. If there is no `name` tag, iD will use the `ref` as the label, but only [in a specific list of situations](https://github.com/openstreetmap/iD/blob/bd1836f/modules/svg/labels.js#L30-L63). 

This list was introduced 9 years ago in 26dfaf81 & 14272ef2, and creates a weird inconsistency - all the features in this screenshot have `ref=12`, but the label is only shown on some features.

image

This PR allows the `ref` to always be used as a label on any feature. It seems like that was the original intention anyway - https://github.com/openstreetmap/iD/issues/546#issuecomment-12950797

### DIFF
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -57,6 +57,9 @@ export function svgLabels(projection, context) {
         ['point', 'shop', '*', 10],
         ['point', 'tourism', '*', 10],
         ['point', 'camp_site', '*', 10],
+        ['line', 'ref', '*', 12],
+        ['area', 'ref', '*', 12],
+        ['point', 'ref', '*', 10],
         ['line', 'name', '*', 12],
         ['area', 'name', '*', 12],
         ['point', 'name', '*', 10]


tastrax:
Hi @k-yle I assume these failures will mean this change wont get into the next update (that I understand may be soon)?

k-yle:
Hey @tastrax, there have been various build issues with iD since early april (e.g. #9101), which were causing the tests to fail. Should be all good now :)

tastrax:
Many thanks, much appreciated.

tyrasd (approved):

Issue #7943, opened 2020-09-01
Show warning if changeset comment is too long


tordans:
ATM, I can write a very long changeset comment and save, but the comment will get cut off at a limit. Ideally, the Editor would warn me, once I am over the limit.

It seems to 255 Chars max.

Screenshot: No warning ATM
![image](https://user-images.githubusercontent.com/111561/91899246-2865c400-ec9d-11ea-8325-3a538aa2bc49.png)

mmd-osm:
> It seems to 255 Chars max.

Yes. As with any tag key/value pair, the same 255 unicode character limit also applies here.

mikini:
Second this, same just bit me in [changeset 90422778](https://www.openstreetmap.org/changeset/90422778).

In the old days of 90s HTML you`d just set a length/size/limit on the input/textarea and the browser would prevent anymore input than that in the UI.

A quick look seems to suggest that the [maxlength attribute on form elements](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-maxlength) is still supported in HTML5.

bhousel:
> A quick look seems to suggest that the [maxlength attribute on form elements](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-maxlength) is still supported in HTML5.

Yes, a long time ago iD did use maxlength for this: https://github.com/openstreetmap/iD/pull/2411/commits/8b896b8b84e858416b738a64cf17ca179682dd46

But that approach doesn't really work for strings that contain unicode - https://github.com/openstreetmap/iD/issues/6817 has more on why this is a hard problem.

mmd-osm:
... in particular https://github.com/openstreetmap/iD/issues/6817#issuecomment-579758275 and  https://github.com/openstreetmap/iD/issues/6817#issuecomment-642186244

quincylvania:
We might show a remaining character count as your input gets long in textarea fields, instead of some obtrusive warning. These are fairly standard nowadays.

danieldegroot2:
@tyrasd it currently
- has context only on new objects and exceeding limit (no label on hover, otherwise ambiguous 'bar', scroll bar or bug).
(also, a horizontal filling/emptying bar usually indicates ongoing completion or remaining time, not a scale as such)
- is only visible when typing (and may not be noticed by users typing in the plain tags field at the bottom).
- goes outside the input field UI element when exceeding length.

[...]

PR #9063, opened 2022-04-05
Add max length to the changeset comment

CorruptComputer:
The changeset comment will be silently truncated otherwise, causing confusion for folks writing a detailed changeset comment and finding that some of the info they added is cut off.

For an example of this issue effecting a changeset that I made, see: https://www.openstreetmap.org/changeset/119343887

### DIFF
--- a/modules/ui/changeset_editor.js
+++ b/modules/ui/changeset_editor.js
@@ -56,6 +56,10 @@ export function uiChangesetEditor(context) {
 
         if (initial) {
             var commentField = selection.select('.form-field-comment textarea');
+
+            // Give the comment field a max length of 255 characters since anything over that length would be truncated
+            commentField.attr('maxlength', 255);
+
             var commentNode = commentField.node();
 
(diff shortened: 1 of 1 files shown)


bhousel:
Looking through the backlog, I guess #7943 has more on this.  
I really thought it had been fixed, since we've been allowed to use the spread operator to split unicode strings for a while, and there are even util functions for correctly counting the length of strings now, but who knows.

sekerob:
It's hard to see from the CS comment box if one has used up the maximum number of characters also because not every character is using equal space. I've added many comments to my own Change Sets which oddly are of seemingly limitless length, so yes would love to see an extension and do away with wasting time on acronyms and abbreviations and cryptic descriptions not all know e.g. w/ meaning.

CorruptComputer:
Closing this since it looks like the problems with the HTML spec haven't been fixed with the `maxlength` attribute. See: https://github.com/whatwg/html/issues/1467

PR #9076, opened 2022-04-15
Fix iD using invalid user selected imagery

### LINKED ISSUE #8732: Use Bing when the last user selected imagery isn't available somewhere
To reproduce:

1. Open https://www.openstreetmap.org/edit#map=18/50.34364/5.15747
2. Change the imagery to Bing then back to the one marked as best (SPW).
3. Open https://www.openstreetmap.org/edit#map=17/28.75531/-97.51904

Result:
The SPW imagery is selected though it isn't available and there isn't any imagery with "best" set, so no imagery is shown.

Expected:
Bing imagery is selected.

renancleyson-dev:
Fixes #8732

The PR changes how imagery is loaded. The editor will try to use imageries available on the current location before using a fallback(e.g. `Bing`, `none`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,6 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased
 
 #### :newspaper: News
-<<<<<<< HEAD
 #### :tada: New Features
 * Consider lifecycle prefixes when deciding whether features should be rendered as a line or an area ([#8881], thanks [@k-yle])
 #### :white_check_mark: Validation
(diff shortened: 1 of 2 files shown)


tyrasd:
I've changed the logic slightly in [eee346e](https://github.com/openstreetmap/iD/pull/9076/commits/eee346e7dc2b78eebedc5f7c89421371205380cc)[^1] to check if the "last used" imagery is valid: It would be confusing if an externally requested layers is not always selected. If a user requests a layer which is not available in a region it is either the fault of the user requesting it or the extent of the layer might be wrong in the layers database. The "requested" backgrounds were also not causing the issue reported in #8732 as far as I can see.

[^1]: PS: sorry for the force-push, but I couldn't get the conflict-resolving merge to show up properly on github :shrug:

PR #9080, opened 2022-04-18
Fix bug: Wikidata field displays [object Object]

### LINKED ISSUE #9067: Wikidata field displays `[object Object]`
### URL

https://www.openstreetmap.org/edit?way=172049896#map=18/42.68426/23.33316

1. Go to the specified URL
2. Add a field for wikidata
3. Enter anything in it

image

Firefox

nontech:
the `wbsearchentities` response adds a `display` section that contains the label and description of the returned entity (if available) with their language code.

It means `display` in this case is an object and not a string.

### DIFF
--- a/modules/ui/combobox.js
+++ b/modules/ui/combobox.js
@@ -386,7 +386,17 @@ export function uiCombobox(context, klass) {
                     return 'combobox-option ' + (d.klass || '');
                 })
                 .attr('title', function(d) { return d.title; })
-                .html(function(d) { return d.display || d.value; })
+                .html(function(d) {
+                    // d.display can be an object
+                    if ( typeof d.display === 'object' &&
+                         !Array.isArray(d.display) &&
+                         d.display !== null
+                    ) {
+                        return d.display.label.value;
+                    }
+
+                    return d.display || d.value;
+                })
                 .on('mouseenter', _mouseEnterHandler)
                 .on('mouseleave', _mouseLeaveHandler)
                 .merge(options)


modules/ui/combobox.js
```diff
@@ -386,7 +386,17 @@ export function uiCombobox(context, klass) {
...
+                    // d.display can be an object
+                    if ( typeof d.display === 'object' &&
+                         !Array.isArray(d.display) &&
+                         d.display !== null
+                    ) {
+                        return d.display.label.value;
+                    }
+
+                    return d.display || d.value;
+                })
```
1ec5, line 399: The `display` object’s `language` property would be a suitable value for the element’s `lang` attribute.

modules/ui/combobox.js
```diff
@@ -386,7 +386,17 @@ export function uiCombobox(context, klass) {
                     return 'combobox-option ' + (d.klass || '');
                 })
                 .attr('title', function(d) { return d.title; })
-                .html(function(d) { return d.display || d.value; })
+                .html(function(d) {
+                    // d.display can be an object
+                    if ( typeof d.display === 'object' &&
+                         !Array.isArray(d.display) &&
+                         d.display !== null
```
1ec5, line 393: Does the Wikibase API potentially return either format? I thought it had already cut over to the new format irreversibly.

1ec5 (commented):
Thanks for looking into this issue. A couple minor points, but the PR does what it says.

modules/ui/combobox.js
nontech: True the Wikibase API returns `display` as an object, this property is newly introduced, but I think this part of the code is used by several input dropdown boxes handling other APIs as well to display dropdown values. 

In those other APIs, there is a field called `display` which is just a normal string.
Well, I just used console.log() to check `d.` Not sure if that's the right way.

modules/ui/combobox.js
nontech: True but isn't this part of code just responsible to display dropdown values and not assign anything.
Or am I understanding it completely wrong?

modules/ui/combobox.js
1ec5: Ah, I missed that this is in the shared combobox code. Do you think it would be possible to modify the code specifically for the Wikidata field instead? That will avoid any potential impact on other fields that use comboboxes.

https://github.com/openstreetmap/iD/blob/7b0127f5e7b98214782574698c5fc8c970ed8caa/modules/ui/fields/wikidata.js#L357-L371

/ref https://github.com/openstreetmap/iD/issues/9067#issuecomment-1098414653

nontech: Hey @1ec5,
Hmm. I think the above code `iD/modules/ui/fields/wikidata.js` > `Lines 357 to 371 in 7b0127f` only comes into play after an item is selected from the dropdown and not while showing the list of options in the dropdown (where our current problem is)
Additionally, it has a different data shape.
For example, `wikidataEntity` is
`
{
    "type": "item",
    "id": "Q23635",
    "labels": {
        "ms": {
            "language": "ms",
            "value": "Bermuda"
        },
        "en": {
            "language": "en",
            "value": "Bermuda"
        },
        "hi": {
            "language": "hi",
            "value": "बरमूडा"
        }
    },
    "descriptions": {...},
    "claims": {...},
    "sitelinks": {...}
}
`
`propKey `is `labels` & ultimately a string value in one of ['Bermuda', 'Bermuda', 'बरमूडा'], depending on the language, is returned.

Isn't this different for our context problem at hand here?
According to https://phabricator.wikimedia.org/T304878, the API changes `label` & `description` and not `labels` or `descriptions`
Or maybe I understood it wrong? Thoughts?

P.S. - Thanks for the comment.

bhousel:
I ended up instead fixing this in  `/ui/fields/wikidata.js` `itemsForSearchQuery` callback, as this is the code which receives the results of the query and converts them into data for the combobox.
https://github.com/facebookincubator/RapiD/commit/1d5aa805d40f4aeb994e1e1e205b43d71562f2c7

Nhattannnn (approved):

tyrasd:
> I ended up instead fixing this in `/ui/fields/wikidata.js` `itemsForSearchQuery` callback, as this is the code which receives the results of the query and converts them into data for the combobox. [facebookincubator@1d5aa80](https://github.com/facebookincubator/RapiD/commit/1d5aa805d40f4aeb994e1e1e205b43d71562f2c7)

In 4770dc8773b3aa418b200f5fa035f0ec4d602c6e, I also used this approach, but changed it such that the reported `language` is used as a HTML `lang` attribute for accessibility / screen reader support (as proposed above https://github.com/openstreetmap/iD/pull/9080#discussion_r852486669).

Issue #8419, opened 2021-03-23
Validator shows lots of "too close to" warnings for Stolpersteine


dreua:
According to the [Stolpersteine wiki page (german only)](https://wiki.openstreetmap.org/wiki/DE:Stolpersteine) each Stolperstein should be mapped as a single node, even when they are very close to each other.

Example: https://www.openstreetmap.org/edit#map=23/52.18570/9.08078

If you move around the city and select streets with adjacent Stolpersteine the issue list easily gets filled drowning the other real issues.

Fix: Disable those warnings for `memorial:type=stolperstein`. Code pointers welcome!

Workaround: Maybe I should just consider this a digital "Stolperstein" and live with it?

Find [Stolpersteine on Overpass-Turbo](https://overpass-turbo.eu/s/15ki)

OT: Thanks for developing iD, it's a really great tool and I use it a lot! OSM + OSS = :heart:

dreua:
Code is probably here: https://github.com/openstreetmap/iD/blob/develop/modules/validations/close_nodes.js

Can we just filter for `memorial:type=stolpersteine` and `memorial=stolpersteine` here?

k-yle:
This was fixed in #9089 and released over a year ago, so I think we can close this issue?

dreua:
Yes of course, thanks for fixing!

PR #9089, opened 2022-04-28
#8419 memorial stolperstein are ignored for close node check

paulklie:
Fixes issue #8419 
Stolperstein are often quite close together, as per wiki. iD should no longer flag this as an issue.


tyrasd:
Thx!

dreua:
Thank you!

PR #9097, opened 2022-05-06
Fix xss in user account display

jleedev:
Only affects currently logged in user who has put HTML in their display name.

#8813


tyrasd:
thanks!

Issue #9101, opened 2022-05-12
Unable to build locally


ZeeshanZulfiqarAli:
I am unable to build on my local machine using `npm run all` command, which throws an error related to not being able to find icon in font-awesome. It was working fine a few hours back ( if that helps ).

### How to reproduce the issue?

Run `npm run all` command and you'll see the following error:

```
zeeshan@Zeeshans-MacBook-Air iD % npm run all

> iD@2.21.0-dev all
> run-s clean build dist


> iD@2.21.0-dev clean
> shx rm -f dist/esbuild.json dist/*.js dist/*.map dist/*.css dist/img/*.svg


> iD@2.21.0-dev build
> run-s build:css build:data build:js


> iD@2.21.0-dev build:css
> node scripts/build_css.js


🏗   Building css...
👍  css built: 134.63ms


> iD@2.21.0-dev build:data
> shx mkdir -p dist/data && node scripts/build_data.js


🏗   Building data...
Error: No FontAwesome icon for fas-th-list
TypeError: Cannot read properties of undefined (reading 'html')
    at /Users/zeeshan/work/iD/scripts/build_data.js:229:75
    at Array.forEach (<anonymous>)
    at writeFaIcons (/Users/zeeshan/work/iD/scripts/build_data.js:224:23)
    at /Users/zeeshan/work/iD/scripts/build_data.js:114:7
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 9)

ERROR: "build:data" exited with 1.
ERROR: "build" exited with 1.
```

### Screenshot(s) or anything else?

Currently tested on `develop` branch

arch0345:
Build seems to work if you change the version of these four dependencies in `package.json` to `~6.1.0`:
https://github.com/openstreetmap/iD/blob/ce136d0144487bc3c8fb569910958865860559a3/package.json#L71-L74

PR #9116, opened 2022-05-20
Ignore invalid units in field input

### LINKED ISSUE #9110: Road speed unit field matches invalid value to km/h
If I enter an unrecognized value into the Speed Limit field’s unit textbox, then move the focus elsewhere, the value changes to “km/h”, regardless of the original value.

Matching the first value is a problem because, like a lot of users, I’m in the habit of hitting the Tab key to commit the value of a textbox. For example, if a road has a 55&nbsp;mph speed limit, I’ll enter `55` into the main numeric textbox then hit Tab. If I’m unaware that the focus is in the unit textbox, I’ll hit <kbd>2</kbd> to start drawing something else, but all I did was enter that number into the unit textbox. Then I’ll click on the map, see that nothing is being drawn, and press <kbd>2</kbd> again, unaware that I’ve actually tagged the roadway with `maxspeed=55`, meaning 55&nbsp;km/h (34&nbsp;mph). I’m pretty sure this is what happened to [this way](https://www.openstreetmap.org/way/393891760/history) in [this changeset](https://www.openstreetmap.org/changeset/69922501) and [this way](https://www.openstreetmap.org/way/630413677/history) in [this changeset](https://www.openstreetmap.org/changeset/101376197), despite tagging speed limits correctly on all the surrounding roads.

This is unlikely to be the main source of incorrect speed limits in the United States and United Kingdom.[^josm] However, I think iD’s behavior is easier to fix. The following code only checks whether the value is literally `mph` and defaults to `km/h` otherwise. Instead, the function should bail if the value isn’t one of `km/h` or `mph`. This would be consistent with other combo boxes that don’t accept freeform input, such as the “Add field” combo box and Wikipedia field’s language combo box.

https://github.com/openstreetmap/iD/blob/ce136d0144487bc3c8fb569910958865860559a3/modules/ui/fields/roadspeed.js#L73-L74

[^josm]: That distinction probably belongs to JOSM, which [doesn’t have a unit option to begin with](https://josm.openstreetmap.de/ticket/19077). Go Map!! similarly lacks this option despite being based on id-tagging-schema: bryceco/GoMap#600.

1ec5:
If the user enters something into the unit box of the Speed Limit or Max Height field that isn’t one of the two recognized units, revert the unit to what it was before without changing the measurement system.

Fixes #9110.

### DIFF
--- a/modules/ui/fields/roadheight.js
+++ b/modules/ui/fields/roadheight.js
@@ -101,7 +101,12 @@ export function uiFieldRoadheight(field, context) {
 
 
         function changeUnits() {
-            _isImperial = utilGetSetValue(primaryUnitInput) === 'ft';
+            var primaryUnit = utilGetSetValue(primaryUnitInput);
+            if (primaryUnit === 'm') {
+                _isImperial = false;
+            } else if (primaryUnit === 'ft') {
+                _isImperial = true;
+            }
             utilGetSetValue(primaryUnitInput, _isImperial ? 'ft' : 'm');
             setUnitSuggestions();
             change();
--- a/modules/ui/fields/roadspeed.js
+++ b/modules/ui/fields/roadspeed.js
@@ -70,7 +70,12 @@ export function uiFieldRoadspeed(field, context) {
 
 
         function changeUnits() {
-            _isImperial = utilGetSetValue(unitInput) === 'mph';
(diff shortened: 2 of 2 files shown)


1ec5:
Some builds are failing due to #9101, which is unrelated to these changes.

PR #9124, opened 2022-05-21
Add multilingual field exceptions

### LINKED ISSUE #9043: Multilingual field exceptions
### Description

The multilingual field shows up for all tags containing name: such as name:prefix , name:etymology , name:right, name:left , name:source , name:signed etc.  This can cause confusion and issues such as what's seen in https://www.openstreetmap.org/changeset/118595559

### Screenshots

image

wcedmisten:
# Summary

Closes https://github.com/openstreetmap/iD/issues/9043

Adds exceptions for `name:left`, `name:right`, `name:etymology`, `name:signed`, and `name:source` when opening the multilingual tag.

# Testing
Can be tested by adding these tags to a feature, and verifying that no multilingual field appears.

image

image

### DIFF
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -116,6 +116,9 @@ export function uiFieldLocalized(field, context) {
 
     // update _multilingual, maintaining the existing order
     function calcMultilingual(tags) {
+        // tags that use the format name:<item>, but
+        // aren't languages
+        const nonLanguageTags = ["left", "right", "source", "signed"]
         var existingLangsOrdered = _multilingual.map(function(item) {
             return item.lang;
         });
@@ -124,6 +127,10 @@ export function uiFieldLocalized(field, context) {
         for (var k in tags) {
             var m = k.match(/^(.*):(.*)$/);
             if (m && m[1] === field.key && m[2]) {
+                if (nonLanguageTags.includes(m[2])) {
+                    continue;
+                }
+
                 var item = { lang: m[2], value: tags[k] };
                 if (existingLangs.has(item.lang)) {
                     // update the value
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -125,7 +125,7 @@ export function uiFieldLocalized(field, context) {
             // matches for field:<code>, where <code> is a BCP 47 locale code
             // motivation is to avoid matching on similarly formatted tags that are
             // not for languages, e.g. name:left, name:source, etc.
-            var m = k.match(/^(.*):([A-Za-z]{2,3}(?:-[A-Za-z]{3}[A-Za-z]{2})?)$/);
+            var m = k.match(/^(.*):([A-Za-z]{2,3}(?:-[A-Za-z]{4})?)$/);
             if (m && m[1] === field.key && m[2]) {
                 var item = { lang: m[2], value: tags[k] };
                 if (existingLangs.has(item.lang)) {
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -125,7 +125,7 @@ export function uiFieldLocalized(field, context) {
             // matches for field:<code>, where <code> is a BCP 47 locale code
             // motivation is to avoid matching on similarly formatted tags that are
             // not for languages, e.g. name:left, name:source, etc.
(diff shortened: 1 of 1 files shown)


modules/ui/fields/localized.js
```diff
@@ -116,6 +116,9 @@ export function uiFieldLocalized(field, context) {
 
     // update _multilingual, maintaining the existing order
     function calcMultilingual(tags) {
+        // tags that use the format name:<item>, but
+        // aren't languages
+        const nonLanguageTags = ["left", "right", "source", "signed", "etymology"]
```
1ec5: This is a good start, though quite a few common non-language subkeys could be added to this list, either in this PR or in a followup. Off the top of my head, there’s also [`pronunciation`](https://wiki.openstreetmap.org/wiki/Key:name:pronunciation),[^fonipa] [`prefix`](https://wiki.openstreetmap.org/wiki/Key:name:prefix), and `genitive`.

To avoid a slow trickle of additions, perhaps we could match only the subkeys that follow the BCP&nbsp;47 pattern. The following regular expression is similar to the one StreetComplete introduced in streetcomplete/StreetComplete@04fd11acebcb17ef050696164ec583fecd843545, but it also covers country-specific subkeys, such as the Hong Kong and Taiwan variants of Traditional Chinese.

```regex
[a-z]{2,3}([A-Z][a-z]{3})?([A-Z]{2})?
```

Another syntax for transliteration is still in widespread use, for example `name:ja_rm` for romaji, but it’s deprecated, so we don’t need to go out of our way to support it.

/ref https://github.com/streetcomplete/StreetComplete/issues/2723#issuecomment-815057365

[^fonipa]: Technically, `name:pronunciation` should be `name:*-fonipa`, but no one is using that syntax.

1ec5 (commented):
Thanks for this contribution! I must admit I’m going to miss seeing the etymology displayed so prominently, which hopefully impressed mappers as to our thoroughness, but openstreetmap/id-tagging-schema#256 would be the more correct way to surface that information anyhow.

wcedmisten:
@1ec5 Thanks for the feedback! This is my first time contributing here, so my initial approach was attempting to be more permissive of nonstandard tags, but I do agree restricting the locale matching would be a better approach.

With your feedback in mind, I've changed the regex to match on 2-3 letters, optionally followed by a hyphen and 4 letters.

This seems to match the streetcomplete commit you referenced.

My main question is if this is too disruptive to non-standard uses of `name:<language>`, that may not comply with this format. Again, I'm fairly new here, so not sure how to survey the usage of these tags, or what the etiquette is around changing that behavior. Let me know what you think. Thanks!

modules/ui/fields/localized.js
```diff
@@ -122,7 +122,10 @@ export function uiFieldLocalized(field, context) {
         var existingLangs = new Set(existingLangsOrdered.filter(Boolean));
 
         for (var k in tags) {
-            var m = k.match(/^(.*):(.*)$/);
+            // matches for field:<code>, where <code> is a BCP 47 locale code
+            // motivation is to avoid matching on similarly formatted tags that are
+            // not for languages, e.g. name:left, name:source, etc.
+            var m = k.match(/^(.*):([A-Za-z]{2,3}(?:-[A-Za-z]{4})?)$/);
```
1ec5: Oh, right, I missed the hyphens in my earlier suggestion. This would be more correct:

```suggestion
            var m = k.match(/^(.*):([a-z]{2,3}(?:-[A-Z][a-z]{3})?(?:-[A-Z]{2})?)$/);
```

This will take care of country codes, and the capitalization matches the preferred case in the standards [as documented here](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2322836#Issues). Other cases are quite rare these days. You can get a sense for the prevalence of certain name subkeys [using taginfo](https://taginfo.openstreetmap.org/search?q=name%3A).

Nonstandard or deprecated keys will continue to show up in the raw tags section of the sidebar, but they won’t show up in the Multilingual Names field. It probably wouldn’t have been a great experience anyways, with the language textbox showing the raw nonstandard code instead of the name of the language.

StreetComplete uses the localized name keys somewhat differently than iD does, so I think a slight discrepancy in the regular expressions should be OK.

1ec5 (approved):

ramen2581:
I should've added this in the issue. You can find the usage of tags on taginfo. 
https://taginfo.openstreetmap.org/search?q=name%3A

PR #9162, opened 2022-06-14
Update issue template for Bug reports

CorruptComputer:
Changes
- Add 'Opera', 'Vivaldi', and 'Other (please specify)' as browser options
  - Vivaldi and Opera are both popular Chromium-based browsers
  - Also made this list in alphabetical order
- Add prompt for the version number of iD
- Give a clickable link to development deployment so the reporter of an issue can easily go there to check


CorruptComputer:
I wasn't sure what other browsers (if any) should be included here, also thought of just adding an 'Other Chromium-based (please specify)' option to the list instead of adding Vivaldi and Opera directly, that way they could be just input by an issue reporter as needed.

PR #9168, opened 2022-06-15
Fix tagging of localizable keys by multiCombo fields

### LINKED ISSUE #9164: Text from preset selections ends up in tags
Recently I fixed a couple dozen nodes of recycling containers with strange tags like

recycling:any_glass_(tempered_glass
recycling:etc.)
recycling:tout_type_de_verre_(cristal)
recycling:jakékoliv_sklo_(temperované_sklo
recycling:jedes_glas_(gehärtetes_glas

This is a recent phenomenon and happens only with edits from iD.

1ec5:
Use the raw subkey as the value of each option in a multiCombo field’s dropdown menu.

Fixes #9164.

### DIFF
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -222,7 +222,7 @@ export function uiFieldCombo(field, context) {
                 var label = field.t('options.' + k, { default: k });
                 return {
                     key: k,
-                    value: label,
+                    value: _isMulti ? k : label,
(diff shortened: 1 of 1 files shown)


1ec5:
Not sure if it’s related, but every time I hover over the feature, each chip in the field adds the localized value to the label redundantly:

glass glass glass glass glass glass glass glass glass glass glass glass glass glass

Edit: This is unrelated, so I opened #9169 to track it.

PR #9171, opened 2022-06-17
fix some boundaries not hidden by the filter

### LINKED ISSUE #6277: Boundaries not hidden
The boundaries at http://preview.ideditor.com/master/#background=GEOSN-DOP-RGB&disable_features=boundaries&id=r2827902&locale=en&map=18.08/51.03154/12.82461 are not hidden, even though the filter is on.

There is probably some special tagging cases at work here, but I don't know enough about boundary tagging to spot it.

PS: I saw a commit by @quincylvania that – as I understand it – should handle this case of boundaries that are only described in the relation(?) https://github.com/openstreetmap/iD/commit/e1da28bcd64bf4d1820611480bd1eb82196ffd21#diff-17d23fbfac3f1af98c721f940ad049e8R336

k-yle:
Closes #6277, specifically https://github.com/openstreetmap/iD/issues/6277#issuecomment-566752951

Sometimes boundaries are visible, even if they're disabled in the "Map Data" panel. 

To test this PR:
 - go to [example 1](http://localhost:8080/#disable_features=points,traffic_roads,service_roads,paths,buildings,building_parts,indoor,landuse,boundaries,water,rail,pistes,aerialways,power,past_future,others&map=19.25/-43.43082/172.64387
) or [example 2](http://localhost:8080/#disable_features=points,traffic_roads,service_roads,paths,buildings,building_parts,indoor,landuse,boundaries,water,rail,pistes,aerialways,power,past_future,others&map=19.25/39.97496/-75.19271). note how all features, including the boundary, are correctly hidden.
- go to the same locations on [openstreetmap.org/edit](https://openstreetmap.org/edit). Note how the boundaries are not hidden.

### DIFF
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -69,6 +69,11 @@ export function rendererFeatures(context) {
     }
 
 
+    /**
+     * @param {string} k
+     * @param {(tags: Record<string, string>, geometry: string, parents: osmRelation[]) => boolean} filter
+     * @param {?number} max
+     */
     function defineRule(k, filter, max) {
         var isEnabled = true;
 
@@ -133,9 +138,13 @@ export function rendererFeatures(context) {
             !_rules.pistes.filter(tags);
     });
 
-    defineRule('boundaries', function isBoundary(tags) {
+    defineRule('boundaries', function isBoundary(tags, geometry, parents) {
+        // This rule applies if the way has no interesting tags, and if either:
+        //   (a) the way has a `boundary=*` tag, or
+        //   (b) every parent of the line is a relation with `type=boundary`.
         return (
             !!tags.boundary
+            || (geometry === 'line' && parents && parents.length && parents.every(r => r.tags.type === 'boundary'))
         ) && !(
             traffic_roads[tags.highway] ||
             service_roads[tags.highway] ||
@@ -391,6 +400,8 @@ export function rendererFeatures(context) {
             var matches = {};
             var hasMatch = false;
 
+            const parents = features.getParents(entity, resolver, geometry);
+
             for (var i = 0; i < _keys.length; i++) {
                 if (_keys[i] === 'others') {
                     if (hasMatch) continue;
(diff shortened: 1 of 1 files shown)


tyrasd (approved):

modules/renderer/features.js
```diff
@@ -392,13 +401,13 @@ export function rendererFeatures(context) {
             var hasMatch = false;
 
             for (var i = 0; i < _keys.length; i++) {
+                const parents = features.getParents(entity, resolver, geometry);
```
tyrasd: we should move this line outside of the `for` loop.

tyrasd (changes requested):
If I see this correctly, the additional `parents.every(…)` check on line 415 is now not necessary anymore. Let's simplify the if statement to 

```javascript
                        // 2. belongs only to a single multipolygon relation
                        if (parents.length === 1 && parents[0].isMultipolygon()) {
```

Same for the `relationShouldBeChecked` method: When we check the parent relations direclty "in the loop", there's no need to additionally check the boundary relations as well.

tyrasd:
btw. I'm slightly unsure whether invoking the `getParents` method for every entity in this "hot" method could not be potentially a performance bottleneck. I assume not, but it's actually hard to tell.

An alternative fix (which doesn't require invoking `getParents` for every entity) for the bug would have been to write the `'boundaries`' filter in the following way:

```javascript
    defineRule('boundaries', function isBoundary(tags, geometry, parents) {
        // This rule applies if the way has no interesting tags, and if either:
        //   (a) is a way has a `boundary=*` tag, or
        //   (b) is a relation of `type=boundary`.
        return (
            !!tags.boundary
            || tags.type === 'boundary'
        ) && !(
            […]
```

:thinking: not sure what is better

modules/renderer/features.js
k-yle: I've made these two changes

tyrasd:
I went with the slightly more conservative fix mentioned [above](https://github.com/openstreetmap/iD/pull/9171#issuecomment-1182029457). This had the additional benefit of also properly disabling the preset for the generic `type=boundary` relation, as seen in this screenshot:

PR #9175, opened 2022-06-21
Fix the content of combobox when the display label is set

### LINKED ISSUE #9159: Combobox for "more fields" broken
Affected is the development branch. See 
image

Most likely a regression in https://github.com/openstreetmap/iD/commit/4770dc8773b3aa418b200f5fa035f0ec4d602c6e

diana-nicolae:
Fixes #9159

The display field is not a function, it contains html.

### DIFF
--- a/modules/ui/combobox.js
+++ b/modules/ui/combobox.js
@@ -7,8 +7,7 @@ import { utilGetSetValue, utilRebind, utilTriggerEvent } from '../util';
 // It is keyed on the `value` of the entry. Data should be an array of objects like:
 //   [{
 //       value:   'string value',  // required
-//       display: 'label function' // optional, if present will be called with d3 selection
-//                                              to modify/append, see localizer's t.append
+//       display: 'label html'     // optional
 //       title:   'hover text'     // optional
 //       terms:   ['search terms'] // optional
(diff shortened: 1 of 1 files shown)


modules/ui/combobox.js
```diff
@@ -7,8 +7,7 @@ import { utilGetSetValue, utilRebind, utilTriggerEvent } from '../util';
 // It is keyed on the `value` of the entry. Data should be an array of objects like:
 //   [{
 //       value:   'string value',  // required
-//       display: 'label function' // optional, if present will be called with d3 selection
-//                                              to modify/append, see localizer's t.append
+//       display: 'label html'     // optional
```
tyrasd, line 10: thanks for the pull request, but this property was actually changed from containing html to a callback function in 8f79932ab1259267560541bbef2ae290d68d42e6 in order to prevent possible html injection into UI fields. see also https://github.com/openstreetmap/iD/wiki/How-to-use-iD%27s-localizer-module#use-tappendstringid-for-adding-text-with-locale-info

tyrasd: PS: I'm working on a "proper" fix for this bug right now.

PR #9181, opened 2022-06-24
Fix connect features when crossing with pedestrian signals

### LINKED ISSUE #9176: Using `Connect the features` with Crossing With Pedestrian Signals doesn't work as expected
### URL

_No response_

1. Add a `Crossing With Pedestrian Signals` way that crosses a street
2. While selecting this new way, click on `Connect the features`
3. A generic `highway=crossing` node is added

When using `Connect the features` with a `Unmarked Crossing` or `Marked Crossing` way, it adds an `Unmarked Crossing` or `Marked Crossing` node respectively. Because of this, I think it makes sense for it to add a `Crossing with Pedestrian Signals` node instead of just a generic `highway=crossing` node

image
image

Chrome

faebebin:
Fix #9176 
Add a `marked crossing` node when [connect features] with `crossing with pedestrian signals` is triggered, instead of an unspecified higway crossing.

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -146,7 +146,7 @@ export function validationCrossingWays(context) {
                         return {};
                     }
                     var pathFeature = entity1IsPath ? entity1 : entity2;
-                    if (['marked', 'unmarked'].indexOf(pathFeature.tags.crossing) !== -1) {
+                    if (['marked', 'unmarked', 'traffic_signals'].indexOf(pathFeature.tags.crossing) !== -1) {
                         // if the path is a crossing, match the crossing type
                         return bothLines ? { highway: 'crossing', crossing: pathFeature.tags.crossing } : {};
                     }
--- a/test/spec/validations/crossing_ways.js
+++ b/test/spec/validations/crossing_ways.js
@@ -213,9 +213,19 @@ describe('iD.validations.crossing_ways', function () {
         verifySingleCrossingIssue(validate(), { highway: 'crossing' });
     });
 
-    it('flags road crossing crosswalk', function() {
+    it('flags road crossing marked crosswalk', function() {
         createWaysWithOneCrossingPoint({ highway: 'residential' }, { highway: 'footway', crossing: 'marked' });
         verifySingleCrossingIssue(validate(), { highway: 'crossing', crossing: 'marked' });
+    });
+
+    it('flags road crossing crosswalk with traffic_signals', function() {
+        createWaysWithOneCrossingPoint({ highway: 'residential' }, { highway: 'footway', crossing: 'traffic_signals' });
+        verifySingleCrossingIssue(validate(), { highway: 'crossing', crossing: 'traffic_signals' });
+    });
+
+    it('flags road crossing unmarked crosswalk', function() {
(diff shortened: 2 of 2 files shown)


faebebin:
This seems a quick fix. I am new to iD and getting into codebase step by step. Thus I have seen that I also could first update the https://github.com/openstreetmap/id-tagging-schema which I did not study yet. This would then allow a { highway: 'crossing', crossing: 'traffic_signals' } . Let me know if this is preferred.

test/spec/validations/crossing_ways.js
```diff
@@ -213,11 +213,21 @@ describe('iD.validations.crossing_ways', function () {
...
+    it('flags road crossing traffic_signals', function() {
+        createWaysWithOneCrossingPoint({ highway: 'residential' }, { highway: 'footway', crossing: 'traffic_signals' });
+        verifySingleCrossingIssue(validate(), { highway: 'crossing', crossing: 'marked' });
+    });
+
+    it('flags road crossing unmarked crosswalk', function() {
+        createWaysWithOneCrossingPoint({ highway: 'residential' }, { highway: 'footway', crossing: 'unmarked' });
+        verifySingleCrossingIssue(validate(), { highway: 'crossing', crossing: 'unmarked' });
+    });
+
```
faebebin, line 230: Maybe could try to group all tests in this file in nested blocks (if possible with karma?). Then having group header test messages instead of comments?

modules/validations/crossing_ways.js
```diff
@@ -150,6 +150,10 @@ export function validationCrossingWays(context) {
                         // if the path is a crossing, match the crossing type
                         return bothLines ? { highway: 'crossing', crossing: pathFeature.tags.crossing } : {};
                     }
+                    if (pathFeature.tags.crossing === 'traffic_signals') {
+                        // if the path is a crossing with traffic_signals assume the crossing to be marked
+                        return bothLines ? { highway: 'crossing', crossing: 'marked' } : {};
```
arch0345:
```suggestion
                        return bothLines ? { highway: 'crossing', crossing: 'traffic_signals' } : {};
```

arch0345 (commented):
Thanks for working on this @faebebin! A preset for `highway=crossing` + `crossing=traffic_signals` [was recently added to the id-tagging-schema repo](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/highway/crossing/traffic_signals.json)

modules/validations/crossing_ways.js
faebebin: @arch0345 thanks for the help. Actually the solution is now even easier :) . Attention, I forced pushed the update, but the diff is should be quite easy to get.

faebebin: @arch0345 ping (sry, not quite sure whether you noticed the change becuase I forced pushed )

arch0345: Just tested this on localhost and it seems to be working for me, thanks @faebebin

arch0345 (approved):

faebebin:
@tyrasd can you trigger workflow checks here? Thanks :)

tyrasd:
merged as 0c1d4c7c4 – not sure why github didn't pick up the merge commit :shrug:

PR #9205, opened 2022-07-15
Add ability to clone attributes of a feature to other features

### LINKED ISSUE #9191: Add ability to clone attributes of a feature to other features
### Description

To speed up edition of multiple similar features, we could select all those features, then right-click and select the action to clone some properties of the first selected feature (see screenshot, there is no icon though in this implementation).

Examples of features to clone are 

- address tags (addr:* properties)
- road attributes (lanes, sidewalks, etc)

### Screenshots

cloneAddressProperties

diana-nicolae:
Fixes #9191

Added an action that clones tags from the first selected entity to all other entities.
Created one base operation that clones a subset of tags.
Created several operations that call the base operation and pass a subset of related tags to clone.
It allows to save time by copying several related tags with one click.

Also added unit tests for those features.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -96,6 +96,46 @@ en:
       too_large:
         one: This can't be copied because not enough of it is currently visible.
         other: These can't be copied because not enough of them are currently visible.
+    clone_address:
+      title: Clone address tags
+      description: Clone address tags from the first selected entity.
+      key: "&"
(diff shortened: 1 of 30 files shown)


diana-nicolae:
This a base PR for discussion. All operations have the same icon for now. For the shortcuts, there are not enough keys, I'm not sure what to use.

Issue #6616, opened 2019-07-03
"Lens" system for alternative map styling


quincylvania:
The default map styling is fine for most users most of the time, but there are times when domain-specific styling could be useful. For example:

- Road `maxspeed` (#5662)
- Road `access` (#2068)
- Route membership (#659)
- Unnamed roads
- Unreviewed imports
 
We should build a general-purpose "lens" system that lets people quickly toggle between these styles ("lenses"). As part of #6483, lenses should be defined atomically and could be swapped in and out by downstream projects. Only some lenses would be suitable to ship with iD to everyone on osm.org, but ideally users could plug in their own lenses too.

I assume JOSM already has a feature like this. We should see how they do it.

quincylvania:
The toolbar might not be the best place for these to live, but for inspiration:

<img width="256" alt="Screen Shot 2019-07-03 at 4 31 54 PM" src="https://user-images.githubusercontent.com/2046746/60624883-64a94600-9db4-11e9-9ac3-5cabd6a8594f.png">
<img width="242" alt="Screen Shot 2019-07-03 at 4 40 58 PM" src="https://user-images.githubusercontent.com/2046746/60624882-64a94600-9db4-11e9-860d-fb7a1c371b31.png">

matkoniecz:
> I assume JOSM already has a feature like this. We should see how they do it.

It is called map paint styles in JOSM. See doc at https://josm.openstreetmap.de/wiki/Styles

![https://josm.openstreetmap.de/raw-attachment/wiki/Styles/addstyle.gif](https://josm.openstreetmap.de/raw-attachment/wiki/Styles/addstyle.gif)

quincylvania:
Looks like JSOM uses MapCSS. I wonder if iD should try to use that.

matkoniecz:
Not sure how hard to implement it, but making possible to use all styles made for JOSM sounds very nice.

sun-geo:
> .... For example:
> 
> * Road `maxspeed` (#5662)
> * Road `access` (#2068)
> * Route membership (#659)
> * Unnamed roads
> * Unreviewed imports

Sounds interesting, please add for me in the list
* Road `destination`
* Road `lanes`
* Road `turn:lanes`

1ec5:
Many of the [layers provided by the former ito! Map tool](https://wiki.openstreetmap.org/wiki/Ito_Map#Full_map_list) would be useful as lenses.

don-vip:
[...]

PR #9210, opened 2022-07-22
Show private and customers access with different casing colour

### LINKED ISSUE #9207: Rendering for private/customer/… access
### Description
To be able to quickly differentiate private and customers access for highway=service.
- red for private access
- orange for customers access

### Screenshots
Screenshot from 2022-07-19 12-22-52
Screenshot from 2022-07-19 12-23-34

diana-nicolae:
Fixes #9207

Added access to the tag classes.
For a highway=service with access=private the casing is red.
For a highway=service with access=customers the casing is orange.
I allows to quickly see the access type.

### DIFF
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -589,6 +589,13 @@ path.line.stroke.tag-highway-steps,
     stroke: #fff;
 }
 
+/* access: private, customers */
+path.line.casing.tag-access-customers.tag-highway-service {
+    stroke: rgb(255, 136, 0);
+}
+path.line.casing.tag-access-private.tag-highway-service {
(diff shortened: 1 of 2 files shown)


diana-nicolae:
This a base PR for discussion. I would like to ask if it would be ok to create a surface tag for all highways including those that are paths.

tyrasd:
I would say that #6616 / #11189 might be a better approach to solve this. Colouring the casing by `access` tag by default is probably a bit too noisy. Also, not only the `access` tag should be considered, but also `vehicle`, `motor_vehicle`, and potentially also tags like `foot`, `bicycle`, `bus`, `emegency`, etc. depending on the mapper's use case.

PR #9226, opened 2022-07-31
#9225 create a railway_crossing when connecting two railways

### LINKED ISSUE #9225: Automaticly add railway=railway_crossing to railway on railway crossings.
### Description
Currently when the user connects two roads or one road and a railway via the "connect the features" fix for detected missing connections, a crossing node is added. However this does not happen with railway on railway crossings, eventhough such a tag exists: railway=railway_crossing

paulklie:
see #9225 
Now a railway=railway_crossing is automatically created.

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -156,7 +156,9 @@ export function validationCrossingWays(context) {
                 return {};
             }
             if (featureType1 === 'waterway') return {};
-            if (featureType1 === 'railway') return {};
+            if (featureType1 === 'railway') {
+                return {railway: 'railway_crossing'};
+            }
 
         } else {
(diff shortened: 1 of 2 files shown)


k-yle (approved):
I can confirm that this works, I've been using this PR for several years now

tyrasd:
There could also be the case that such a vertex should be tagged as `railway=switch`. We should offer both tags as different fixes here

example of a 4-way switch:

paulklie:
> There could also be the case that such a vertex should be tagged as `railway=switch`. We should offer both tags as different fixes here
> 
> example of a 4-way switch:
> 

This sounds like a significantly larger task, and for little gain. These 4 way switches are significantly rarer than simple railway crossings in my experience.

PR #9228, opened 2022-08-02
Add ability to snap a way onto another way

### LINKED ISSUE #9223: Add ability to snap a way onto another way
### Description

Ability to snap 2 ways together when they need to follow each other on a long portion between 2 common nodes. This speeds up the follow process as it allows to snap the 2 ways together with only one click.

### Screenshots

Screenshot from 2022-07-29 08-49-53   
Screenshot from 2022-07-29 08-50-54

Screenshot from 2022-07-25 14-25-23 

Screenshot from 2022-07-25 14-26-02 |  Screenshot from 2022-07-25 14-25-41
:-------------------------:|:-------------------------:
Screenshot from 2022-07-25 14-26-19 | Screenshot from 2022-07-25 14-30-11

diana-nicolae:
Fixes #9223

Added an action that allows to snap 2 ways together.
Created the associated operation for this action.
Added some fonctions that are used by the snap action in the way.js file.
Snap allows to speed up the follow process as it is possible to snap the 2 ways together with only one click.

Also added unit tests for this feature.


diana-nicolae:
This a base PR for discussion. I picked a shortcut and an icon, but these can always be changed.

How it works :
- Snaps the first selected way onto the second way (the ways must have at least 2 common nodes)
- If no nodes are selected, snaps between the first and the second common nodes
- If one node is selected, snaps between the first common node and the selected node
- If 2 nodes are selected, snaps between the first selected node and the second selected node

In case we snap on a closed way, it is possible to chose the direction to snap by selecting the first and the second node. For now, it will snap between the first selected node and the second selected node. Another option would be to add a second operation to change the snap direction. It is also possible to have a different shortcut key that would change the snap direction.

tyrasd:
see https://github.com/openstreetmap/iD/issues/9223#issuecomment-4998625472

PR #9241, opened 2022-08-14
Added native labels for various South Asian languages to languages.json

bgo-eiu:
Includes Punjabi, Kalasha, Shina, Shugni, Bhariati, Brahui, Balti, Saraiki, Wakhi, and Torwali

Note that "pa-Arab" and "pa-Guru" have been removed in favor of just sticking with the standard codes "pnb" and "pa" for Shahmukhi and Gurmukhi. This matches the convention on OSM, with `name:pnb` and `name:pa` and in most software generally. `Arab` is not really a usable script subkey because it says nothing of which Arabic-based script it is for; it would be like if there was just "Brah" and we had to put both Devanagari and Hindi in it. وننجݨا, وننجنا, and وڄݨا are all valid representations of the same Punjabi word in different Perso-Arabic script orthographies (none of which could be read the same way in actual Arabic script) so it is better to just ignore "Arab" and use separate codes wherever possible. I left the other languages that use this script code alone since I do not know much about them, but for Punjabi this is the correct way to do it


data/languages.json
```diff
@@ -410,8 +413,6 @@
   "osa": {},
   "ota": {},
   "pa": {"nativeName": "ਪੰਜਾਬੀ"},
-  "pa-Arab": {"base": "pa", "script": "Arab"},
-  "pa-Guru": {"base": "pa", "script": "Guru"},
```
1ec5, line 414: Should `name:pa-Arab` and `name:pa-Guru` be marked as deprecated in [deprecated.json](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/deprecated.json) in the id-tagging-schema repository? That will make it easier for mappers to migrate these keys over to the ones you’re recommending without having to hunt for them in the raw tag editor.

data/languages.json
```diff
@@ -55,9 +55,11 @@
   "bew": {},
   "bez": {"nativeName": "Hibena"},
   "bfd": {},
+  "bft": {"nativeName": "بلتی"},
```
1ec5, line 58: This file was generated from [cldr-localenames-full](https://www.npmjs.com/package/cldr-localenames-full/) v37. Maybe upgrading [this dependency](https://github.com/openstreetmap/iD/blob/a2cacaaf2456de4c8443a7b5dab0ac3393be5deb/package.json#L81) to v41 would bring in some more language names so we wouldn’t have to maintain them manually in this repository.

1ec5 (commented):
Thanks for scrutinizing these codes. This file is automatically generated by scripts/language_names.js, so I’m not confident that your changes would persist after `npm run build:data` or `npm run translations` is run again. Can you move the changes to language_names.js?

https://github.com/openstreetmap/iD/blob/a2cacaaf2456de4c8443a7b5dab0ac3393be5deb/scripts/language_names.js#L6-L14

bgo-eiu:
Sorry for dropping the ball on this, and thanks for working out how to incorporate these! It wasn't obvious what to change to get it to work

Issue #7902, opened 2020-08-15
Use tram crossing tags when connecting trams and roads/paths


quincylvania:
_Broken out from https://github.com/openstreetmap/iD/issues/5990#issuecomment-674419342._

Tram tracks are rather different from other train tracks. There are separate crossing tags to indicate this: [`railway=tram_level_crossing`](https://wiki.openstreetmap.org/wiki/Tag%3Arailway%3Dtram_level_crossing) and [`railway=tram_crossing`](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dtram_crossing). iD should have presets for these and account for them in the validator.

maraf24:
So status=proposed doesn't actually mean anything?

quincylvania:
@maraf24 I'm not sure what you're referring to. The [`status=proposed`](https://taginfo.openstreetmap.org/tags/status=proposed) tag has only two uses and isn't supported by iD.

Issue #8463, opened 2021-04-18
Do not set highway=crossing on node when connecting service=driveway


Strubbl:
When there is highway=service + service=driveway, which crosses a footway and has no common node, iD correctly detects this as a data problem. If i use the "connect ways" option of the data problem, both ways get one node, which connects these ways. But also this node gets the tagging highway=crossing automatically. 

In case of a footway (or cycleway or path) crossing a highway=service + service=driveway i think this is not a desired behaviour. It would be better if iD would not do this in this case.

lectrician1:
Why?

Strubbl:
Because it is not a crossing in the sense of https://wiki.openstreetmap.org/wiki/Tag:highway%3Dcrossing
None of the mentioned use cases matches for a private road to a building (highway=service + service=driveway) which crosses a footway

lectrician1:
Just because there isn't a example doesn't mean you can't use it there. The usage definition is "Pedestrians can cross a street here". That definition is valid for service roads since they are also streets.

osm-ToniE:
Strubbl refers to a service=driveway road and I think this is something different than a service road w/o additional service= tags.

- driveways (according to my experience)
  - are usually separated from the other streets by a kerb, i.e. cars have to cross a kerb to enter the driveway
  - pedestrians usually to not have to cross kerbs to cross a driveway
  - footways and driveways therefore usually have the same level
  - driveways usually have the same surface as the footways, precedence is for the pedestrians, not for the cars (they have to watch out)
  - from the overall perspective driveway users cross footways, not vice versa
 
- highways=crossing is an optional tag on crossing nodes, iD makes this a default

I second Strubbl, highway=crossing at service=driveway is really a superfluous tag and shouldn't be set w/o a good reason

lectrician1:
It still functions as a "crossing" though. The primary use case of highway=crossing is for routers to tell cars where to expect pedestrians. When a driveway crosses a footway or sidewalk, vehicles should know that they should expect pedestrians there. This is not a superfluous tag.

osm-ToniE:
highway=crossing is not a superfluous tag at all. It is just the combination with service=driveway.

I just wanted to point out, that by turning into a driveway with a car, **usually crossing a lowered kerb**, the vehicle enters an area which is designated to pedestrians and cyclists.
There shouldn't be the need for routers and drivers with open eyes ((t) "open eyes 2.0") to give hints at every "crossing" on a driveway.
[...]

Issue #5990, opened 2019-02-27
Crossing highway/railway - exclude railway=tram and railway=construction + construction=tram and railway=disused + disused=tram


matkoniecz:
Due to various weird geometry issues (`railway=tram` follow tram tracks and `highway=*` follow centerline of road) point where `railway=tram` line and `highway=*` line crosses is often NOT representing point where tram crosses road.

In JOSM for this reason trams are excluded from this rule.

See https://josm.openstreetmap.de/ticket/16889 for partial JOSM equivalent

matkoniecz:
See for an example https://www.openstreetmap.org/?mlat=50.07323&mlon=19.93585#map=19/50.07323/19.93585&layers=N

quincylvania:
> `railway=tram` line and `highway=*` line crosses is often NOT representing point where tram crosses road.

Okay, but the two features are still technically connected and if you're driving north through that intersection you'll still end up driving over the tracks. If that road was mapped bidirectionally it'd be even more accurate. Traffic lights and stop signs aren't mapped where they're physically located either.

Even putting all that aside, trams can still cross roads via bridges and tunnels. People would expect iD to flag these.

bhousel:
> Okay, but the two features are still technically connected and if you're driving north through that intersection you'll still end up driving over the tracks. If that road was mapped bidirectionally it'd be even more accurate. Traffic lights and stop signs aren't mapped where they're physically located either.

Yeah I agree - the purpose of mapping the crossings isn't to make pretty maps of where the crossings are, but more to inform data consumers.  So that a cycling navigation app might flash up a message saying "watch out for tram tracks in 200m" or something.   

I'm inclined to leave this validation as-is..

matkoniecz:
See https://www.openstreetmap.org/?mlat=50.06646&mlon=19.93905#map=19/50.06646/19.93905&layers=N - here there is no self-crossing - tram tracks are embedded in the road.

![selection_001](https://user-images.githubusercontent.com/899988/53496324-1bfa1d00-3aa2-11e9-9de0-102c8c7157ad.png)

https://www.google.pl/maps/@50.0662939,19.9390377,3a,75y,15h,73.79t/data=!3m6!1e1!3m4!1sA865u1CQIVOHTxDQqdaZDQ!2e0!7i13312!8i6656

Or see for example https://www.openstreetmap.org/?mlat=50.07300&mlon=19.93584#map=19/50.07300/19.93584&layers=N

[...]

PR #9306, opened 2022-09-26
stop automatically adding railway crossing tags when adding tram-road/path intersection nodes

tyrasd:
Addresses concerns raised in https://wiki.openstreetmap.org/wiki/ID/Controversial_Decisions#Tram_crossings or https://wiki.openstreetmap.org/wiki/Talk:Tag:railway%3Dtram_level_crossing.

This essentially reverts #7902: Instead of using the undiscussed tags `railway=tram_level_crossing` and `railway=tram_crossing`, the regular values of railway crossing tags should be used at intersections of tram lines with roads or paths.

See https://github.com/openstreetmap/id-tagging-schema/pull/596 for a related change to the tagging schema repository.

~~Still to be discussed: Should there be a tag upgrade/deprecation rule be implemented to fix these `tram_*crossing` tags?~~

//edit: as noted below (https://github.com/openstreetmap/iD/pull/9306#issuecomment-1258412778), automatically adding any railway crossing tags is not good practice. I've changed the PR to not add any additional crossing tags in these situations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :tada: New Features
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
-* Use standard railway crossing tags when marking tram-road/path intersections ([#9306])
+* Don't add non-standard `railway=tram_crossing/rail_level_crossing` tags when using validator to add connection nodes at tram-road/path intersections ([#9306])
 #### :rocket: Presets
 #### :hammer: Development
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,11 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :tada: New Features
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
+* Use standard railway crossing tags when marking tram-road/path intersections ([#9306])
 #### :rocket: Presets
 #### :hammer: Development
+
+[#9306]: https://github.com/openstreetmap/iD/pull/9306
 
 
 # 2.22.0
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -164,6 +164,9 @@ export function validationCrossingWays(context) {
                 if (featureTypes.indexOf('railway') !== -1) {
                     if (!bothLines) return {};
 
+                    var isTram = entity1.tags.railway === 'tram' || entity2.tags.railway === 'tram';
+                    if (isTram) return {};
+
                     if (osmPathHighwayTagValues[entity1.tags.highway] ||
                         osmPathHighwayTagValues[entity2.tags.highway]) {
                         return { railway: 'crossing' };
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -164,21 +164,10 @@ export function validationCrossingWays(context) {
                 if (featureTypes.indexOf('railway') !== -1) {
                     if (!bothLines) return {};
 
-                    var isTram = entity1.tags.railway === 'tram' || entity2.tags.railway === 'tram';
-
                     if (osmPathHighwayTagValues[entity1.tags.highway] ||
                         osmPathHighwayTagValues[entity2.tags.highway]) {
-
-                        // path-tram connections use this tag
-                        if (isTram) return { railway: 'tram_crossing' };
-
-                        // other path-rail connections use this tag
                         return { railway: 'crossing' };
                     } else {
-                        // path-tram connections use this tag
-                        if (isTram) return { railway: 'tram_level_crossing' };
-
-                        // other road-rail connections use this tag
                         return { railway: 'level_crossing' };
                     }
                 }


matkoniecz:
> This essentially reverts https://github.com/openstreetmap/iD/issues/7902: Instead of using the undiscussed tags railway=tram_level_crossing and railway=tram_crossing, the regular values of railway crossing tags should be used at intersections of tram lines with roads or paths.

That is not solving main problem: iD asking people to mark fake crossing where none actually exists, see https://github.com/openstreetmap/iD/issues/5990#issuecomment-467876574

As long as this rule is present iD mappers will add fake data. And it is preferable to keep it contained within own tag. (obviously, better would be to remove this rule like JOSM is doing this and rely on manual mapping)

As large part of this tram crossings is misplaced or not showing actual crossing points, this tags should not be semiautomatically upgraded. It is better to keep this polluted data contained within own tag. So this change actually makes things worse.

tyrasd:
thanks for the additional information @matkoniecz. I agree that it is best to stop to automatically add any railway tags in these situations, as tram-road crossings are so diverse. Mappers need to choose whether in a particular situation it's best to use a `railway` crossing tag or no tag at all. I've changed the PR accordingly.

1ec5:
> Mappers need to choose whether in a particular situation it's best to use a railway crossing tag or no tag at all.

In that case, this PR still overcorrects by reducing the likelihood that `railway=level_crossing` or `railway=crossing` will be added by the mapper when necessary. Adding the intersection node will satisfy the warning, leading the mapper to think there’s no need for a second step to add this tag. This isn’t a theoretical issue: some routers issue warnings to users to look both ways when crossing, but it requires a crossing tag, so they may have a false sense of security when traversing an undertagged crossing.

By analogy, when a roadway crosses a waterway, we can’t be certain that there’s a ford, so the crossing_ways validation rule gives the user a choice of adding a ford or adding a bridge. Similarly, if we’re sure there should be a node but unsure if it should be tagged `railway=crossing`, wouldn’t it be more appropriate to warn but offer a choice between `railway=crossing` and an untagged intersection node, similar to my suggestion in https://github.com/openstreetmap/iD/issues/8463#issuecomment-1066069238?

If https://github.com/openstreetmap/iD/issues/5990#issuecomment-467876574 is illustrative of the issue around trams, then it’s worth noting that [street-running](https://en.wikipedia.org/wiki/Street_running_train) is not exclusive to trams or streetcars, so for consistency, this rule would need to exempt `railway=rail` from `railway=level_crossing` and `railway=crossing` too. Perhaps it would be less surprising to exempt ways tagged [`embedded=yes`](https://wiki.openstreetmap.org/wiki/Key:embedded), but you’ll probably notice that it’s merely “in use” – out of necessity rather than due to any editor suggestion.

This situation appears to be an unintended consequence of [a 2019 decision](https://wiki.openstreetmap.org/wiki/Proposed_features/Railway_tracks_on_highway) to map non-physically-separated railways in finer detail than non-physically-separated footways and cycleways. Maybe the real lesson here is that there needs to be a discussion about digitizing all three types of ways more consistently.

1ec5:
> [street-running](https://en.wikipedia.org/wiki/Street_running_train) is not exclusive to trams or streetcars, so for consistency, this rule would need to exempt `railway=rail` from `railway=level_crossing` and `railway=crossing` too

Also, for completeness, `railway=light_rail`, which occupies a poorly defined space between `rail` and `tram`. Very many light rail lines are street-running in some parts but not others. As I understand it, one of the original motivations for `railway=tram_crossing` and `railway=tram_level_crossing` was that the right-of-way rules for streetcars are very different than for heavy rail. But light rail is a mix of the two on a case-by-case basis. It’s similar to how a footway and cycleway can cross each other at either a pedestrian crossing or a cycle crossing, depending on the situation. It’s good to give the mapper a choice, but then we should put the choice in front of the mapper instead of making it easy for them to get yelled at by other mappers.

SK53:
> an unintended consequence of [a 2019 decision](https://wiki.openstreetmap.org/wiki/Proposed_features/Railway_tracks_on_highway) to map

I can't speak for others, but the Nottingham tram tracks have been fully mapped as separate ways since at least 2013, and I think from around 2011, and possibly earlier than that. It was such a long time ago I can no longer remember the exact reasons for this, but I do remember that there were several of them.

I was unaware of these tags and have reviewed them for the Nottingham tram network. `tram_crossing` looks not to cause any problems at all, but `tram_level_crossing` does cause issues where there is street running. In nearly all cases (I may have overlooked some) where the tram is not street running and all significant junctions with street running the junction is protected by tram-activated traffic signals (I'm afraid not always mapped, I'm working on it). Having additional routing penalties imposed on these may not be significant, but for minor road junctions and particularly service roads and driveways connecting to the road with the tramlines. In all these cases, drivers (& cyclists etc) would be expected to observe normal behaviour to check for traffic in either direction and these are no different from roads where there is no tram.

As you say right-of-way rules for streetcars can be vary variable. In many jurisdictions (useful reddit thread [here](https://www.reddit.com/r/AskEurope/comments/qx6i67/do_cars_have_to_give_way_to_tram_passengers_at/)) cars must not pass on the nearside of a tram picking up/discharging passengers (@matkoniecz can correct me, but I think stops on [Lubicz](https://www.openstreetmap.org/node/2420200262) in Krakow are like this).

I suspect the best behaviour is to suggest connecting the two ways, but nothing else. Other [suggested approaches](https://wiki.openstreetmap.org/wiki/Proposed_features/Railway_tracks_on_highway) for handling street running don't seem to be in use.

1ec5:
> I suspect the best behaviour is to suggest connecting the two ways, but nothing else.

Why not suggest both – let the user decide?

matkoniecz:
> ars must not pass on the nearside of a tram picking up/discharging passengers (@matkoniecz can correct me, but I think stops on [Lubicz](https://www.openstreetmap.org/node/2420200262) in Krakow are like this).

well, obviously - are there places with trams where situation is different? Even Poland when its traffic laws were really pedestrian-hostile was without such mistake.

SK53:
> Why not suggest both – let the user decide?

Because I'm just not convinced there's anything special about these intersections. They have no special signage (as a railway=level_crossing does), are no different from other intersections where we don't have special tags), and on a typical road with street-running we'd typically be missing plenty of other potential examples (every driveway, service road etc). This based on UK, FR, AT, CH, DE and PL tramways. I'm going to trawl though old photos, but foolishly I didn't get any on New Years Day when I was riding the tram.

Also I think there's a strong tendency for users to click yes to suggestions.

Of course if they don't exist they can be ignored, regarded as tagging syntactic sugar on a tram/highway intersection. I don't know if there's any mileage in considering usage of the [embedded_rails](https://taginfo.openstreetmap.org/keys/embedded_rails) key instead (not its original purpose I know, but not unadjacent)

hungerburg:
Looking at the pictures in the documentation https://wiki.openstreetmap.org/wiki/Tag:railway=tram_crossing, I'd question the correctness of the mappings in cases like the one here - four tram_crossings ahead - https://www.mapillary.com/app/?lat=47.2664718&lng=11.4012267&z=17.365181230201728&dateFrom=2023-09-01&pKey=840793570875240&focus=photo

Here how this looks in a map with tram*crossings marked. There are 16 shown, with help of the ID validator it is easy to make out a dozen more.
16_tram_crossings_12_left_to_map

PR #9307, opened 2022-09-27
Change None background to white, change color of roads

### LINKED ISSUE #9281: Make default map background color customizable
### Description

Currently, the map background is set to black which makes it difficult to view OSM data when:
- "none" background layer is selected
- a transparent custom layer with mostly dark features is imported

I believe in terms of readability and UX, it would be better to have a transparent background as default:

```
.ideditor .main-map {
  /* background: #000; */
  background: transparent;  
}
```

### Screenshots

Examples:
image
image

anniemichel:
Address Issue #9281 

Changes:
- Changed default map background to white
- Also changed residential and service roads to light grey, and driveways to a darker yellow. 

Purpose:
To improve readability and UX

Visuals:
Before:
Screen Shot 2022-09-26 at 3 51 43 PM
After:
Screen Shot 2022-09-26 at 3 45 36 PM

### DIFF
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -198,7 +198,7 @@ path.line.casing.tag-tertiary {
 }
 path.line.stroke.tag-highway-residential,
 path.line.stroke.tag-residential {
-    stroke: #fff;
+    stroke: #d3d3d3;
 }
 path.line.casing.tag-highway-residential,
 path.line.casing.tag-residential {
@@ -404,7 +404,7 @@ path.line.casing.tag-road {
 /* service roads */
 path.line.stroke.tag-highway-service,
 path.line.stroke.tag-service {
-    stroke: #fff;
+    stroke: #d3d3d3;
 }
 path.line.casing.tag-highway-service,
 path.line.casing.tag-service {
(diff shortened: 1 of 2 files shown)


tyrasd:
Thanks for your effort. However, I think this needs a bit more thought (and work) to be useful. There are more features currently rendered as white lines or areas (e.g. the areas of hospitals or man made structures) which would have low contrast against the white background. Similarly, newly created and still untagged lines are rendered white and thus would be invisible on the white background. Further, the background of the context (right click) menu is also white, which is poorly visible against a white background.

mxdanger:
I'd argue that getting flash banged does not improve readability and UX. But on a more serious note, most of the styling is made for aerial imagery, which is usually a darker color. That means using a white background actually decreases readability and UX due to less contrast, along with other issues that @tyrasd mentioned.

pnorman (changes requested):
Could you add a preview of how this looks against a typical aerial image, which is normally darker?

Of particular concern is using grey for roads, because roads are often grey on the imagery. How does this work?

tyrasd:
closing as stale

PR #9314, opened 2022-09-30
Color Feature

chos906:
Good afternoon,

I created a color feature that would change the background of the main map. I saw that there was a comment about how the "none" background map had a problem with readability. The color of the background was too dark for the users to read the labels. I created 10 buttons that would change the color: ROYGBIV, pink, gray, and white. On each button, there is a shadow feature that would appear whenever the user hovers over that button. Above the buttons, I have added "click to change the color of the background"

Here is what it looks like:
color_feature

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,70 +1,150 @@
 <!DOCTYPE html>
 <html>
-  <head>
+<head>
     <meta charset='utf-8'>
     <title>iD</title>
     <link rel='stylesheet' href='dist/iD.css'>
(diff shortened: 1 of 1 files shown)


tordans:
Where can I find some background about what problem (as in "User Story") you are trying to solve? As in, why do we need to change the background / in which situation / what is the goal to change the background …

tyrasd:
closing as stale

PR #9316, opened 2022-10-01
fix the phone input field showing the wrong placeholder

### LINKED ISSUE #8380: Phone number hint has the wrong localization
Steps to reproduce:

– Open iD
– Choose a country different from the current locale of your browser (in my case: Firefox)
– In this country, search a node with a supermarket, which has no phone number.
– The grey phone number hint displayed in the dedicated field for “phone=” is correctly localized for the country in which you are editing.
– Now go to properties, add a new tag, put “phone” as key and press the Tab key, so that the text cursor moves on to the “value” field.

Current results:
– The grey phone number hint displayed in the dedicated field for “phone=” is now wrongly localized: It hasn’t any more the locale of the country in which you are editing, but the locale of the country as provided by your browser.

Expected results:
– The grey phone number hint displayed in the dedicated field for “phone=” is still correctly localized for the country in which you are editing.

Note:
The wrong grey phone number hint persists even when you delete the incomplete “phone=” tag from the property list, make a mouse click on an empty place of the map to deselect the object, and then make a mouse click again on the very same object: Still the wrong phone number hint.

k-yle:
Closes #8380 

As soon as you type in the telephone input field, the placeholder changes to the default value instead of the localized value. 

See #8380 for reproduction steps


modules/ui/fields/input.js
```diff
@@ -322,6 +322,7 @@ export function uiFieldText(field, context) {
             .attr('title', isMixed ? tags[field.key].filter(Boolean).join('\n') : undefined)
             .attr('placeholder', isMixed ? t('inspector.multiple_values') : (field.placeholder() || t('inspector.unknown')))
             .classed('mixed', isMixed);
+        updatePhonePlaceholder();
```
1ec5: The other call to this function is conditional on the field’s type:

https://github.com/openstreetmap/iD/blob/43a9d275498d2dd1530e4ed7e9e1c90a272a380b/modules/ui/fields/input.js#L95

Should the conditional be repeated here?

k-yle: good spotting thanks, I lost a commit somehow......

PR #9317, opened 2022-10-01
support quickly adding a direction tag to traffic signs

k-yle:
Pressing <kbd>v</kbd> flips the direction of a node, but does nothing for direction-less nodes. 

This PR makes the <kbd>v</kbd> key add a `direction` tag if you try to flip a stop sign/give-way sign without a `direction` tag. 

This significantly speeds up the process of mapping stop signs and give-way signs. 

Alternative solution to #9802

### DIFF
--- a/modules/actions/reverse.js
+++ b/modules/actions/reverse.js
@@ -1,3 +1,5 @@
+import { presetManager } from '../presets';
+
 /*
 Order the nodes of a way in reverse order and reverse any direction dependent tags
 other than `oneway`. (We assume that correcting a backwards oneway is the primary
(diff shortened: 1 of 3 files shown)


modules/actions/reverse.js
```diff
@@ -126,6 +130,10 @@ export function actionReverse(entityID, options) {
             for (var key in node.tags) {
                 tags[reverseKey(key)] = reverseValue(key, node.tags[key], node.id === entityID);
             }
+
+            // for directionless traffic signs, the first flip just adds the direction tag
+            if (isDirectionlessTrafficSign(tags)) tags.direction = 'forward';
```
1ec5: This method name is a little confusing; I thought at first that it would apply to `traffic_sign=*` nodes, which could potentially benefit from this improvement when mapped as a vertex along a roadway. However, a `traffic_sign` node can also be a standalone node, in which case `direction` needs to be an angle, not `forward` or `backward`.

Besides traffic signs, the same improvement would benefit `highway=traffic_signals` (when mapped at the stop line) and `traffic_calming=bump`/`hump`/`table`.

tyrasd (changes requested):
This seems to be a nice usability enhancement. :+1: 

The test _reverses directional values on childnodes: does not assign a direction to a directionless stop sign on the way during a reverse_ [test](https://github.com/openstreetmap/iD/blob/develop/test/spec/actions/reverse.js#L456-L464) is currently failing. As far as I can tell, this is a bug in the implementation, as the `direction` tag of a traffic sign node should only be added if the node itself is "reversed", not when the parent way is reversed.

Could you also please add a unit test for this new feature? (should be quite similar as the basic tests at https://github.com/openstreetmap/iD/blob/develop/test/spec/actions/reverse.js#L18)

k-yle:
sorry, I lost track of this PR, I've updated it to use the presets to determine whether a node is flippable. So it works for any field that has a tag like `direction` or [`…:direction`](https://taginfo.openstreetmap.org/search?q=%3Adirection#keys)

I've also added tests and fixed the bug when you flip a way

tyrasd (commented):
There's an edge case for when a preset has multiple fields for different geometry types (e.g. `vertex` vs. `point`), like here: https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/traffic_sign.json#L5

tyrasd (approved):

PR #9333, opened 2022-10-10
add default implied access values for ways with "motorroad=yes"

tyrasd:
The tag `motorroad=yes` implies different access tags [depending on the region](https://wiki.openstreetmap.org/wiki/Key:motorroad#Regional_definitions), but all which are documented on the wiki include at least `foot=no`, `bicycle=no` and `horse=no`. Since that's all what iD currently displays in the _access_ field, this should be sufficient for now. This could be refined / revisited in the future if https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions#By_country were to be implemented.

See https://github.com/openstreetmap/id-tagging-schema/issues/609

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Better handling of rate limited API calls and other API errors ([#10299])
 * Fix info boxes for descriptions of brands referenced by [NSI](https://github.com/osmlab/name-suggestion-index) presets ([#10885])
 * Make features clickable when "Full Fill" rendering style is selected
+* Fix calculation of access field placeholders for multi selections ([#9333])
 #### :earth_asia: Localization
 #### :hourglass: Performance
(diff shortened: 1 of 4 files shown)


1ec5:
> Since that's all what iD currently displays in the access field

Good point. Technically it also displays `motor_vehicle`, but `motorroad` on a `highway=path` would be truly shocking.

By the way, that wiki page has [an important note](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access_restrictions#Alternative_ideas) about the concept of implied access defaults being merely proposed and not as authoritative in practice as the page presents. [Taginfo](https://taginfo.openstreetmap.org/keys/motorroad#projects) shows a concerning lack of support for this key. Valhalla does support it: valhalla/valhalla#913. However, OSRM does not, and there isn’t even an issue tracking support. (OSRM is the basis of all routing in Organic Maps and cycling routing by Mapbox, among others.)

PR #9340, opened 2022-10-22
allow follow ways if there are already two ways sharing the same nodes

### LINKED ISSUE #8750: Follow operation should prefer the focused way

### LINKED ISSUE #11489: Allow continuing drawing using follow function if two ways/areas have the same geometry

jguddas:
Resolves limitation of #8671 @k-yle 

## Case 1
Connect to the node that is not alongside an edge.

**Given** Way segments that connect the current node to more than two nodes.
**When** Exactly one of the next nodes is only connected using one way segment.
**Then** Connect the current node to the node that is only connected using one way segment.

image

Solved by:
```js
nextNodeIds = nextNodeIds.filter((id,_,nodes) => nodes.lastIndexOf(id) === nodes.indexOf(id))
```

<details><summary>Video</summary>

https://user-images.githubusercontent.com/25524993/197327057-5a7284aa-0b21-4b23-ac04-dee9090421fc.mov

</details>

## Case 2
Connect to next note, even if there are multiple ways leading there.

**Given** Multiple way segments that connect the current note to two nodes.
**And** One of the next nodes is the previous node.
**Then** Connect the current node to the node that is not the previous node.

image

Solved by:
```js
if (new Set([...nextNodeIds]).size > 1) {
```

<details><summary>Video</summary>

https://user-images.githubusercontent.com/25524993/197327095-34eb969f-2676-439f-b345-29217131220f.mov

</details>

## Case 3
Connect to next note if you are at the start or end of a way.

**Given** All ways that contain the current node are start or end of a way.
**And** Lead to the same next node.
**Then** Connect to the next node.

image


modules/behavior/draw_way.js
```diff
@@ -455,42 +455,34 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
 
-            const featureType = getFeatureType(lastNodesParents);
-
-            if (lastNodesParents.length !== 1 || secondLastNodesParents.length === 0) {
-                context.ui().flash
-                    .duration(4000)
-                    .iconName('#iD-icon-no')
-                    .label(t.append(`operations.follow.error.intersection_of_multiple_ways.${featureType}`))();
-                return;
+            if (!secondLastNodesParents.length) return;
```
jguddas: IDK if this is necessary.

modules/behavior/draw_way.js
```diff
@@ -455,42 +455,34 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
+                if (indexOfLast % (nodes.length - 1)) {
+                    return [nodes[indexOfLast-1], nodes[indexOfLast+1]]
+                }
+                if (nodes[0] === nodes[nodes.length - 1]) {
+                    return [nodes[1], nodes[nodes.length - 2]]
+                }
+                return [indexOfLast ? nodes[nodes.length - 2] : nodes[1]]
+            }).filter((id) => id !== secondLastNodeId)
+
+            if (!nextNodeIds.length) return;
```
jguddas, line 464: Is doing nothing a good idea when there are no options?

modules/behavior/draw_way.js
```diff
@@ -423,74 +423,53 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
-            // check if the ways that they're part of are the same way
-            // find index of the last two nodes, to determine the direction to travel around the existing way
-            // add the next node to the way we are drawing
-
             // if we're drawing an area, the first node = last node.
-            const isDrawingArea = _origWay.nodes[0] === _origWay.nodes.slice(-1)[0];
+            const isDrawingArea = _origWay.nodes.length > 1 && _origWay.nodes[0] === _origWay.nodes[_origWay.nodes.length - 1];
 
-            const [secondLastNodeId, lastNodeId] = _origWay.nodes.slice(isDrawingArea ? -3 : -2);
+            const secondLastNodeId = _origWay.nodes[_origWay.nodes.length + (isDrawingArea ? -3 : -2)];
```
jguddas, line 437: I would rather use `.at` but am not sure if [90%](https://caniuse.com/mdn-javascript_builtins_array_at) browser support is good enough.
```suggestion
            const secondLastNodeId = _origWay.nodes.at (isDrawingArea ? -3 : -2);
```

jguddas:
@tyrasd can you take a look at this, I would love to hear what you think?

modules/behavior/draw_way.js
tyrasd: > I would rather use `.at` but am not sure if [90%](https://caniuse.com/mdn-javascript_builtins_array_at) browser support is good enough.

90% is not quite usable enough for iD, I'd say. :thinking: But what was the reason for getting rid of the previous `.slice(-n)` approach? Was it because the nodes array can be shorter than 2 (or 3 for areas) nodes in the corner case when one just starts to draw a new line/area? To fix that, we could just pad the array before slicing it, e.g.:

```javascript
const [secondLastNodeId, lastNodeId] = Array(1).concat(_origWay.nodes).slice(isDrawingArea ? -3 : -2);
```

modules/behavior/draw_way.js
tyrasd: Yeah, I think there should be a flash message similar to the one below (`operations.follow.error.needs_more_initial_nodes`). I can fix that after this PR, though.

Btw, this is for situations where the drawn way is not connected to anything, right? Currently (iD v2.23.2), this results in the `needs_more_initial_nodes` message, which could be also a bit confusing in that situation. :wink:

modules/behavior/draw_way.js
```diff
@@ -423,74 +423,53 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
...
+                if (nodes[0] === nodes[nodes.length - 1]) {
+                    return [nodes[1], nodes[nodes.length - 2]];
+                }
+                return [indexOfLast ? nodes[nodes.length - 2] : nodes[1]];
+            }).filter((id) => id !== secondLastNodeId);
+
+            if (!nextNodeIds.length) return;
+
+            if (new Set([...nextNodeIds]).size > 1) {
+                if (!secondLastNodeId) {
```
tyrasd: This message is currently only shown when the way doesn't have a `secondLastNode` (i.e. a stub way with only 1 node). But we should also show it if the `secondLastNode` is not part of any candidate way, like this:
```suggestion
                if (!secondLastNodeId ||
                    !lastNodesParents.some(({nodes}) => nodes.includes(secondLastNodeId))) {
```

tyrasd (commented):
Hey. Sorry that I didn't respond earlier. These improvements look awesome!

The only thing I think we should still work on is to add back some inline documentation. As some of the parts of the code are not trivial, it wouldn't hurt to add a short explanation here and there. For example, the line `nextNodeIds = nextNodeIds.filter((id,_,nodes) => nodes.lastIndexOf(id) === nodes.indexOf(id))` could be annotated with a short explanation and a link back to this PR. Similarly, the different cases for the generation of `nextNodeIds` could be described such that other devs can get an idea of what it does more quickly.

modules/behavior/draw_way.js
jguddas: > nodes in the corner case when one just starts to draw a new line/area

Exactly.

I will keep it this way if you don't mind, even tho padding is a great solution as well.

modules/behavior/draw_way.js
jguddas: #### This is the case that is currently covered
image

#### This is the case that your code would cover
image

#### There still is this case that is not covered tho
image

#### My untested solution that could cover all three cases:
```diff
- }).filter((id) => id !== secondLastNodeId)
+ let nextNodeIds = unfilteredNextNodeIds.filter((id) => id !== secondLastNodeId)
+ const secondLastNodeIsUnfilteredNextNode = unfilteredNextNodeIds.length !== nextNodeIds.length
```

Then we could show the needs more nodes' error when `!secondLastNodeIsUnfilteredNextNode`.

jguddas: Tested and pushed my fix, I also moved the check after the filter that covers case 1 to cover the following two cases.

image

image

jguddas:
@tyrasd can you take a look again?

jguddas:
@tyrasd ?

tyrasd (approved):

MapSpot:
Really appreciate the progress on this! I've had similar comments on the Follow command in #11895 which have been indirectly addressed.

One last strange behavior I'm noticing is when following two overlapping ways (e.g. a nature reserve and coastline), the new Follow command will prefer to deviate from these two ways in favor of following a line with no overlap. Although this behavior is intended in Case 1, it probably isn't what the user would want or expect in such cases.

image

PR #9365, opened 2022-11-09
show labels for housenumbers

tyrasd:
this is a very quick and dirty go at #1524

It would look like this:

notes: the top house has a housenumber, but isn't rendered because it also has a name, which is too long to be shown at this zoom; the two other named houses have also a housenumber, but it's not rendered as the name gets precedence.

### to do
* [ ] only use housenumber as `displayName` for rendering, not for other uses such as validator warnings (e.g. `35 crosses 37` is a bit suboptimal for a warning when two buildings overlap)
* [ ] check how much does this (negatively) affect map readability and/or map rendering speed in densely mapped areas
* [ ] check what optimizations from #3840 can/should be reused

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -774,6 +774,7 @@ en:
       network_ref_direction: "{network} {ref} {direction}"
       network_ref_from_to: "{network} {ref} from {from} to {to}"
       network_ref_from_to_via: "{network} {ref} from {from} to {to} via {via}"
+      "addr:housenumber": "{addr:housenumber}"
     speed_unit: "Speed unit"
     roadheight:
       # symbol for meters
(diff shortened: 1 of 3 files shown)


tastrax:
At what zoom levels would this show in the editor? Maybe only display at zoom level 20? It might also be useful to add unit numbers as well as house number ie 3/41 for strata title type properties

Dimitar5555:
RapiD already has such feature. Would it be possible to reuse it's code?

1ec5:
See also the discussion in #8440.

tastrax:
Is there any further progress on this issue? I am keen for this to happen in edit mode to make address checking much easier.

hb0nd:
+1 for this, it'd be good to disable it by default, with a toggle in settings.

k-yle:
superseeded by #10970 and #8440

PR #9390, opened 2022-11-25
Show input string length indicator & warning when nearing maximum tag value length

### LINKED ISSUE #7943: Show warning if changeset comment is too long

### LINKED ISSUE #9374: Too long description value got pruned
### URL

_No response_

1. Add a description field to an object.
2. Write text in the large textbox, which is longer than the 255 characters allowed by the API for a tag value.
3. Upload the changeset.

Result: In the resulting changeset there are just the first 255 characters of the typed text, the rest got pruned. The editor noticed the user not about the issue.

Expected result: The Editor should warn the user that the text is too long. It would be nice if there were an indicator of the remaining characters (at least when you typed already more then the allowed maximum). Additionally there should be a warning or an error indicated in the upload dialog to avoid information lost unintended by the user.

_No response_

2.23.2

Firefox

tyrasd:
This is intended to solve #7943, #9374. It not only works on the changeset comment field, but most input fields (e.g. text fields for tags, combo boxes, etc.), to solve similar problems, e.g. when entering multiple long URLs into the `source` field or a particular long `description` or `note`.

Looks like this at the moment:

* text area, almost empty
  image
* text area, almost "full"
  image
* text area, exactly 255 characters
  image
* text area, over limit
  image

Looks similar for combo boxes, regular text fields and translatable text fields. For `semiCombo` fields, it takes preexisting value-parts into account:

image

Currently, the remaining field types do not implement the length indicator (e.g. the `number`, `multiCombo`, `address`, `wikidata` fields), as the likelihood is quite low that a users would want to enter a string > 255 characters there. And if it is done by mistake, the silent truncate is acceptable IMO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 
 # Unreleased
 
+#### :tada: New Features
+* Show a _remaining input length_ indicator and a warning if the maximum for OSM tags (typically, 255 characters) is exceeded ([#9390], [#9392] thanks [@alanb43], [#7943], [#9374])
 #### :sparkles: Usability & Accessibility
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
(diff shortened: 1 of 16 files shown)


tyrasd:
the latest version includes the differently styled warning from #9392 for the changeset field on the upload form, which is matching the warning we already show if the comment includes an invalid data source (see screenshot below). Thanks @alanb43!

image

alanb43:
Nice! Looks great :)

Issue #6817, opened 2019-08-29
Failure when server returns "Value has more than 255 unicode characters"


tbertels:
To reproduce:
* Edit a element property (in this case source) and add more than 255 characters (typically multiple addresses).
* Try to save

Result:
iD shows an error dialog: "Errors occurred while trying to save"
![image](https://user-images.githubusercontent.com/3265870/63961899-1d78c200-ca92-11e9-8e5d-7bddb5da4ab9.png)
Then after clicking OK and trying to save again, we get an OAuth dialog:

![image](https://user-images.githubusercontent.com/3265870/63962050-70527980-ca92-11e9-87e2-534d283a0408.png)
Granting access leads to a File not found server error:

![image](https://user-images.githubusercontent.com/3265870/63962110-9841dd00-ca92-11e9-95c2-f6412d53ae58.png)

This last error may be due to osm itself, but why do we get the OAuth dialog in the first place?

And isn't it possible to show a more helpful message to the user? Including the name of the element if not null and the name of the property would help. This may need some work on the osm API though.

mmd-osm:
Some of the input fields already have a max length of 255 characters, in case of "Source" (any maybe others), this seems to be missing, so that a user can enter much longer strings.

> And isn't it possible to show a more helpful message to the user? Including the name of the element if not null and the name of the property would help. This may need some work on the osm API though.

This can be much better handled by iD by no letting you enter longer strings in the first place. Nothing to do on the OSM API side here.

quincylvania:
> This can be much better handled by iD by no letting you enter longer strings in the first place.

Indeed. We should audit all tag inputs and operations to ensure that they adhere to the character limit.

quincylvania:
Related: #4061

mmd-osm:
I left a comment on https://github.com/openstreetmap/iD/commit/17ae12b3d727e395b8e7fb623c4417d8c24aeb29#commitcomment-36988467, hope it doesn't get lost somehow.

Unless there's some other substr implementation in place, the default one wouldn't handle unicode chars as expected. In the worst case this could lead to some very long description being cut off too early.

bhousel:
> Unless there's some other substr implementation in place, the default one wouldn't handle unicode chars as expected. In the worst case this could lead to some very long description being cut off too early.

Thanks @mmd-osm , following our conversation on Slack I'm reopening this so that we can reimplement string counting and truncation functions that match what the API code does.

quincylvania:
[...]

PR #9392, opened 2022-11-27
added warning when changeset comment length > 255 chars

alanb43:
Added a simple fix for the issue described in #9374 and #7943. I simply insert a warning message / box in the same format that the existing Google warning appears when the user exceeds 255 characters in their changeset message.

#6817 describes some edge cases that seem to work fine with this change, shown below:

"ZALGO" contains 58 javascript characters ([source](https://github.com/orling/grapheme-splitter#examples)), here we see exactly 255 characters being calculated as 'good' using 4 * 58 of the demon string + 18 normal chars. 

image

If we add one, we see the new warning:
image

I also have come to understand that there is a more in-depth fix being worked on in #9390, but this was made as a much simpler fix that follows suit of an existing warning.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -615,6 +615,7 @@ en:
     comment_needed_message: Please add a changeset comment first.
     about_changeset_comments: About changeset comments
     about_changeset_comments_link: //wiki.openstreetmap.org/wiki/Good_changeset_comments
+    changeset_comment_length_warning: "Changeset comments can have a maximum of 255 characters"
     google_warning: "You mentioned Google in this comment: remember that copying from Google Maps is strictly forbidden."
     google_warning_link: https://www.openstreetmap.org/copyright
   contributors:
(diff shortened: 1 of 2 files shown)


tyrasd:
Thanks for this, @alanb43.

The main difference to #9390 is that this PR only addresses the issue for the changeset comment field. The problem of too long strings is however also present on input fields for any OSM tag (see bug #6817). It's not as often encountered for regular tags, but there are also some somewhat common cases (e.g. a long `note`, `description` or `source` tag).

I think that your PR has the benefit that it stylistically matches the existing warnings for other changeset comment mistakes. I think it would be best for the changeset comment if we combine the length indicator widget from #9390 with the warning message from this PR. What do you think?

alanb43:
I agree with this and have been tinkering for a bit now trying to do it, without much luck. Will let you know if I make progress here though!

tyrasd:
This is now integrated into #9390. Thanks again!

PR #9403, opened 2022-12-02
Update core.yaml

### LINKED ISSUE #11726: Merging deletes instead of merging

Image

Everything is going well,

Image

and the left panel shows

Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building
Deleted Building 

jidanni:
Also mention how to select the ways!
(No guide to how to format key strokes. Hence I'm just making a guess on the markup needed.)

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1484,7 +1484,7 @@ en:
       multiselect_h: "Multiselect"
       multiselect: "Selecting multiple features together lets you quickly edit them all at once."
       multiselect_shift_click: "With a mouse and keyboard, hold `{shift}` and {leftclick} left-click features to add them to the selection. On a touchscreen, {longpress_icon} tap-and-hold one feature while {tap_icon} tapping additional features with another finger."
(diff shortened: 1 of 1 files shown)


jidanni:
(As to "how to select the points _and_ the ways?", that is too advanced to mention.)

tyrasd:
Hm… Unfortunately, the `Ctrl` + `↑` shortcut only works for a selection of only vertices. If one selects at least one "free standing" node (e.g. a POI or address node) – which is quite likely when using the lasso tool – the keyboard shortcut doesn't do anything.

> how to select the points and the ways […] is too advanced to mention

Now I'm curious. What's your workflow for this?

jidanni:
For instance, all the trails in the forest preserve have now been paved. We need to select all the trails, including all the bridge and tunnel segments, and change their tag all at once, from  unpaved to paved.

So we need to select all the ways within the lasso, not all the nodes.

So users might wonder, "How could there only be a way to lasso select the nodes, not the ways?"

(And for the case you mentioned, savvy users might,  in the data panel,  unselect the interfering data types first, as a workaround, perhaps.)

PR #9408, opened 2022-12-05
Add PCRS 5cm imagery WMTS source for metropolitan France

flacombe:
Dar maintainers,

I propose to add a new 5cm imagery source for metropolitan France
tms type has been used as it's a wmts service, didn't find any documentation to define native wmts layer.

GetCapabilities is here to check proposed layer: https://wxs.ign.fr/ortho/geoportail/wmts?SERVICE=WMTS&VERSION=1.0.0&REQUEST=GetCapabilities

Best regards


tordans:
This sounds like something that belongs in ELI (editor layer index) not ID directly. See readme for more (this is just a quick note from mobile).

flacombe:
I'm no specialist about how iD is splited between OSM, external content and core.
It was obvious to me if this git contains an imagery.json file, it should be edited here.

readme.md doesn't state anything about ELI, what is it?

tordans:
Yes, I agree the readme does not explain that well. 
The new issue flow will provide the link and context. 
image

flacombe:
The problem is I didn't open an issue but a PR, so such options weren't displayed.

I wonder what the point of the data/imagery.json file in this git is, if it's not maintained here?

tyrasd:
@flacombe Yes, additions to the imagery data belongs in to the ELI repository. Would you mind recreating the PR [there](https://github.com/osmlab/editor-layer-index/pulls) please?

> I wonder what the point of the data/imagery.json file in this git is, if it's not maintained here?

It's here for convenience during development and testing. I'm planning to eventually restructure this part of iD such that the imagery data is loaded dynamically from a CDN. When that is done, the files would not be included in this PR anymore.

PR #9432, opened 2022-12-18
address_formats.json: add addr:housename to Turkish

ddpasa:
In Turkey it's quite common for houses and buildings to have a name as well as a number.

See https://taginfo.openstreetmap.org/keys/addr%3Ahousename for the corresponding taginfo.

### DIFF
--- a/data/address_formats.json
+++ b/data/address_formats.json
@@ -122,6 +122,7 @@
     "format": [
       ["neighbourhood"],
       ["street", "housenumber"],
+      ["housename"],
       ["postcode", "district", "city"]
     ]
   },


tyrasd:
Hi. I'd like to double check / confirm one thing real quick before merging it… I'm not seeing any mention of the `addr:housename` tag on the OSM wiki's [_Turkey_ section](https://wiki.openstreetmap.org/wiki/Addresses#Turkey) on how to map addresses there. Is it just missing there? Was this proposal discussed on a regional community channel, or is this just your personal suggestion?

And as I'm not familiar with the region: Can you please share some real world example cases of where an address (e.g. of a business, POI, etc.) contains a housename? At first glance, I couldn't find too many "good" examples on OSM: situations where mappers mistakenly(?) repeat a POI's name in `addr:housename` [seem](https://www.openstreetmap.org/node/2527036032)  [to](https://www.openstreetmap.org/node/4903031194) [be](https://www.openstreetmap.org/way/175835245) [common](https://www.openstreetmap.org/node/1762077465).

PS: For reference: The current [usage of `addr:housename` in Turkey](http://taginfo.geofabrik.de/europe/turkey/keys/addr%3Ahousename#overview) is a bit higher than the global average: 2% of addresses in Turkey have `addr:housename` vs. 1% globally. It is however still significantly lower than the usage we have in Countries where we already display the `addr:housename` tag in the address field: Ireland has about 5% and the UK around 7% of such mapped addresses.

ddpasa:
Hello @tyrasd , this is somewhat of an confusing topic.  As far as I can see, the tag "name" has been hijacked because a lot of mappers are not really aware that the correct usage of "addr:housename" (I was also doing the same until someone corrected me).  I'm hoping that adding this change will lead to more correct tag usage.

matkoniecz:
Is this change considered as a good idea by local community in general?

Are these names used as a part of address?

ddpasa:
> Is this change considered as a good idea by local community in general?
> 
> Are these names used as a part of address?

We can reach the local community via the new forums: https://community.openstreetmap.org/c/communities/tr/118

I think this will be a very helpful change.

matkoniecz:
Can you post there, explain planned change and ask them for comments?

matkoniecz:
started https://community.openstreetmap.org/t/is-addr-housename-common-and-valid-in-turkey/142813

matkoniecz:
judging from response ("It had been discussed [here](https://community.openstreetmap.org/t/turkiyede-addr-etiketinin-kullanimi-ve-adresler-hakkinda-sorular/138800) (section 5), but I guess no final decision had been reached.") this PR should be closed

matkoniecz:
also, "Are these names used as a part of address?" asked in https://github.com/openstreetmap/iD/pull/9432#issuecomment-3945851129 was not answered

if building has name which is not part of address it should go to `name=`

tyrasd:
also https://www.upu.int/UPU/media/upu/PostalEntitiesFiles/addressingUnit/turEn.pdf does not mention the use of house-names as part of the address in Turkey.

PR #9434, opened 2022-12-20
Optimize oldTagIssues method

Zaczero:
By reordering the checks and shallow copies, this method can run its initial stage about 2x faster on average - reducing initial app load by about 25ms in my case.

#### Time measuring code

<details>
  <summary>Code</summary>

  ```js
    let validation = function checkOutdatedTags(entity, graph) {
      let ts = performance.now();
      let issues = oldMultipolygonIssues(entity, graph);
      if (!issues.length) issues = oldTagIssues(entity, graph);
      let tt = performance.now() - ts;
      total += tt;
      count += 1;
      console.log(total, count, total / count);
      return issues;
    };
  ```
</details>

### Original code

Notice the `return [];` which is used to focus the time measurement on the pre-checks only.

### Explanation

It looks like the `hasInterestingTags` call is very quick to execute and catches more cases than the `presetManager.match` *(~20ms save)*. Also it makes more sense to perform the shallow copy after all pre-checks are run *(~5ms save)*.

### DIFF
--- a/modules/validations/outdated_tags.js
+++ b/modules/validations/outdated_tags.js
@@ -24,11 +24,13 @@ export function validationOutdatedTags() {
 
 
   function oldTagIssues(entity, graph) {
-    const oldTags = Object.assign({}, entity.tags);  // shallow copy
-    let preset = presetManager.match(entity, graph);
-    let subtype = 'deprecated_tags';
-    if (!preset) return [];
     if (!entity.hasInterestingTags()) return [];
+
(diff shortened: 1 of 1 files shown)


tyrasd (approved):

Zaczero:
Thanks for the merge. I have some ideas on improving iD but had to work on something small first to get familiar with the code. I will hopefully contribute more in the future.

PR #9445, opened 2022-12-30
Update FAQ.md

GPSpilot:
self-explanatory

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -11,7 +11,7 @@ Please see [CONTRIBUTING.md](https://github.com/openstreetmap/iD/blob/develop/CO
 ## How can I report an issue with background imagery?
 
 To report an issue with missing or cloudy imagery:
-* _For Mapbox Satellite layer:_ Zoom into the location in [this imagery feedback tool](https://apps.mapbox.com/feedback/satellite/) and describe the problem.
+* _For Mapbox Satellite layer:_ Zoom into the location in [this imagery feedback tool](https://apps.mapbox.com/feedback/satellite/) [Broken link – Access Denied error] and describe the problem.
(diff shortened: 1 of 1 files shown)


danieldegroot2:
This doesn't solve any issues, but ok.

I initially can't find where to give feedback on imagery. The [fresh imagery page](https://labs.mapbox.com/bites/00240/) linked in [this article](https://blog.mapbox.com/a-year-of-improving-openstreetmap-with-map-feedback-598ac755abee) is still available. Not sure if they are still using it though.

GPSpilot:
Wikipedia encourages editors to notate broken links as such, in order to encourage other editors to fix the broken links.  Shouldn't the Openstreetmap help articles do the same?
In that sense, it solves the issue of a broken link not being noted as such.



-----Original Message-----

bryceco:
That paragraph should just be deleted. It implies that there is a feedback link for MapBox somewhere when in fact no such link exists.

tyrasd:
I asked the community team at Mapbox what the current solution(s) exist to provide feedback about their imagery.

> That paragraph should just be deleted.

I'd like to only do this if there really is no way to provide feedback (or I don't hear from them in a reasonable time).

1ec5:
The Mapbox Feedback tool that this FAQ links to went away at some point last year. In its place, there’s a new [Mapbox Contribute](https://www.mapbox.com/contribute/) tool – this is what the “Improve This Map” link points to. But it only solicits edits to addresses, speed limits, etc., not product feedback at the moment.

pnorman (changes requested):
Given the time without getting an updated link from Mapbox and them removing any way to report broken imagery, I'd remove the Mapbox line entirely.

PR #9458, opened 2023-01-12
fix: :bug: Fix multiple appending debug vintage unknown label

### LINKED ISSUE #9457: Redrawing tile label debug vintage with Unknown vintage
### URL

https://www.openstreetmap.org/edit#map=15/24.3273/-102.7650

1. Click background settings on the right pane.
2. Choose a background that doesn't have a vintage metadata. Eg. 'Maxar Premium Imagery'.
3. Click 'Show Background Panel' checkbox.
4. On the background panel click 'Show Tiles'.
5. Pan the map around. This will cause multiple label like: 'UnknownVintage: UnknownVintage: UnknownVintage: UnknownVintage:'

image

_No response_

Chrome

furkanmutlu-tomtom:
Fix multiple appending on debug vintage unknown label. 

Fixes: #9457

### DIFF
--- a/modules/renderer/tile_layer.js
+++ b/modules/renderer/tile_layer.js
@@ -250,9 +250,8 @@ export function rendererTileLayer(context) {
                         if (result && result.vintage && result.vintage.range) {
                           span.text(result.vintage.range);
                         } else {
-                          span.call(t.append('info_panels.background.vintage'));
-                          span.append('span').text(': ');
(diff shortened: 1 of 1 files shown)


tyrasd:
Thanks for the suggestion. I changed the fix, so that the `t.append` method is still used. The simpler `.text(t(…))` solution had the downside that the meta attributes about the language of the displayed strings is not rendered in the final html.

Issue #8816, opened 2021-11-18
Allow holding down mouse button to increment counters


jidanni:
### Description

Here if I want to use the mouse to get to 24, I need to click 24 times (I gave up halfway.)
How about allowing me to just hold down the mouse button?
Yes, I could just fill it in manually.

### Screenshots

![20211119T011341](https://user-images.githubusercontent.com/1230959/142463887-9c58e084-3363-41d0-bde6-c72f9c1a6ab0.jpg)

1ec5:
Continuous stepping is supported by every browser’s native `number` control, so it makes sense for this custom reimplementation of that control to also support the same behavior.

Deeptanshu-sankhwar:
I am interested to work on this issue. Can I please get this assigned to me.

Deeptanshu-sankhwar:
I have enabled continuous increment or decrement when holding the mouse down on the buttons by starting a `setInterval` on `mousedown` to repeatedly update the value, and clearing the interval on `mouseup` or `mouseleave` to stop the updates.

tyrasd:
In retrospect, I think we don't need this feature. Larger numbers should be easily typed in, and the nudge buttons are there mostly to adjust small values (e.g. lanes=1 to lanes=2). Which is sufficient.

PR #9462, opened 2023-01-16
Allow number inputs to be clicked and held

derekthesnake:
This allows for the up/down buttons on number inputs to increment the values continually when they are held.

I migrated the click logic to a subfunction, and made the click function call itself every 250ms until the button is no longer held. It tracks the state by adding an HTML attribute to the buttons while they are held down.

This passes all tests and linting in GitHub Actions.

Addresses #8816.


Zaczero:
Cool idea, I think the first delay should be slightly longer, e.g: 500ms, 250ms, 250ms, ….

I believe a solution using [`setInterval`](https://developer.mozilla.org/en-US/docs/Web/API/setInterval) and `clearInterval` would be much cleaner and easier to follow:
```js
setInterval(click_func, 250, d3_event, d)
```

If not, consider putting the recursive call after this line (for optimization):
https://github.com/openstreetmap/iD/pull/9462/files#diff-20bad07a0a50747d3c088cd4cfd32014b21edec39c11f9c71998bb0f46ef876cR92.

And don't make [a proxy function](https://github.com/openstreetmap/iD/pull/9462/files#diff-20bad07a0a50747d3c088cd4cfd32014b21edec39c11f9c71998bb0f46ef876cR88) just to pass the arguments, you can [do it natively](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout).

PR #9475, opened 2023-01-20
Introduce the hash URL param `presetUrlOverwrite`

tordans:
This enables us to create branch deployments that run `npm run dist` in the 
 id-editor-schema repo and preview the schema using an URL like `http://preview.ideditor.com/master/#locale=en&presetUrlOverwrote=https://branch-preview-123.netlify.app/`.

One Issue I found was, that translations don't work well. That might be due to the way I ran `npm run dist` my id-tagging-schema folder locally. Or maybe some piece is missing. One thing we will likely need to do is force the locale to `en` via URL (as I did in the example above) so the fallback values from the dist-folder are applied.

---

This is the first piece to solve https://github.com/openstreetmap/id-tagging-schema/issues/289. The other part would be to setup netlify with branch deployments in the id-editor-schema repo. — Update: My comment https://github.com/openstreetmap/id-tagging-schema/issues/289#issuecomment-1398366533 outlines who I would set that up.

### DIFF
--- a/API.md
+++ b/API.md
@@ -62,6 +62,7 @@ of iD (e.g. `https://ideditor-release.netlify.app`), the following parameters ar
 * __`validationError`__ - The issues identified by these types/subtypes will be treated as errors (i.e. Issues will be surfaced to the user but will block changeset upload). Each parameter value should contain a urlencoded, comma-separated list of type/subtype match rules.  An asterisk `*` may be used as a wildcard.<br/>
   _Example:_ `validationError=crossing_ways/highway*,crossing_ways/tunnel*`
 * __`walkthrough=true`__ - Start the walkthrough automatically
+* __`presetUrlOverwrite=https://example.com`__ - Overwrite the source for presets. Default is the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) release on jsdelivery ([Code](https://github.com/openstreetmap/iD/blob/develop/config/id.js#L2)). The URL has to serve the root of the id-tagging-schema repo after running `npm run dist`.
 
 ##### iD on openstreetmap.org (Rails Port)
 
--- a/modules/core/file_fetcher.js
+++ b/modules/core/file_fetcher.js
@@ -1,5 +1,5 @@
 import parseVersion from 'vparse';
-import { presetsCdnUrl, ociCdnUrl, wmfSitematrixCdnUrl } from '../../config/id.js';
+import { presetsCdnUrlTemplate, ociCdnUrlTemplate, wmfSitematrixCdnUrlTemplate } from '../../config/id.js';
 
 import packageJSON from '../../package.json';
 
@@ -11,10 +11,15 @@ export { _mainFileFetcher as fileFetcher };
 // coreFileFetcher asynchronously fetches data from JSON files
 //
 export function coreFileFetcher() {
+  const presetsVersion = packageJSON.devDependencies['@openstreetmap/id-tagging-schema'];
+  const presetsCdnUrl = presetsCdnUrlTemplate.replace('{presets_version}', presetsVersion);
+
   const ociVersion = packageJSON.dependencies['osm-community-index'] || packageJSON.devDependencies['osm-community-index'];
   const v = parseVersion(ociVersion);
   const ociVersionMinor = `${v.major}.${v.minor}`;
-  const presetsVersion = packageJSON.devDependencies['@openstreetmap/id-tagging-schema'];
+  const ociCdnUrl = ociCdnUrlTemplate.replace('{version}', ociVersionMinor);
+
+  const wmfSitematrixCdnUrl = wmfSitematrixCdnUrlTemplate.replace('{version}', '0.1');
 
   let _this = {};
   let _inflight = {};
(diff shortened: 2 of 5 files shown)


tordans:
Side note: The commit 1 could be sherry picked in main even if the other part is not merged. I find it quite a bit clearer and I also wonder if there is a bug in main ATM, since in one instance the cdn-url was [not parsed before using it](https://github.com/openstreetmap/iD/pull/9475/commits/1d64839b24d9c7f6d79b3810961d1f0a27c59d27#diff-849c38d4d2f9a07e9301d615da477fc56d407fca11d991fcccd50ba96a620fd5R92-R96)…

modules/core/localizer.js
```diff
@@ -88,6 +89,8 @@ export function coreLocalizer() {
             'locales'     // load the list of supported locales
         ];
 
+        const presetsVersion = packageJSON.devDependencies['@openstreetmap/id-tagging-schema'];
+        const presetsCdnUrl = presetsCdnUrlTemplate.replace('{presets_version}', presetsVersion);
```
tyrasd, line 93: actually… this was working as intended:

[`fileFetcher.get`](https://github.com/openstreetmap/iD/blob/1d64839b24d9c7f6d79b3810961d1f0a27c59d27/modules/core/file_fetcher.js#L67-L71) expands the "template" with the correct version of the loaded presets. Just using the version from `packageJson` doesn't always work because of caching: say you have `v5.0.1` of one of the presets' files cached, but the rest is loaded fresh from the cdn (say, `v5.2.1`), then these files might very likely mismatch in at least a few places, which can result in missing translations for example.

To avoid this, the fileFetcher does first load the preset's `package.json` and use the exact version it gets from that request to load all further preset files. This assures that all files are consistent with each other.

--

The url template can be passed to the fileFetcher:


```javascript

        const localeDirs = {
            general: 'locales',
            tagging: presetsCdnUrlTemplate + 'dist/translations'
        };
```

modules/core/file_fetcher.js
```diff
@@ -29,24 +34,23 @@ export function coreFileFetcher() {
...
+    'oci_defaults': ociCdnUrl + 'dist/defaults.min.json',
+    'oci_features': ociCdnUrl + 'dist/featureCollection.min.json',
+    'oci_resources': ociCdnUrl + 'dist/resources.min.json',
+    'presets_package': presetsCdnUrl + 'package.json',
     'deprecated': presetsCdnUrl + 'dist/deprecated.min.json',
     'discarded': presetsCdnUrl + 'dist/discarded.min.json',
     'preset_categories': presetsCdnUrl + 'dist/preset_categories.min.json',
     'preset_defaults': presetsCdnUrl + 'dist/preset_defaults.min.json',
     'preset_fields': presetsCdnUrl + 'dist/fields.min.json',
     'preset_presets': presetsCdnUrl + 'dist/presets.min.json',
```
tyrasd, line 53: These files (except `presets_package`) should use the `presetsCdnUrlTemplate` instead of ~~`presetsCdnUrlTemplate`~~. See my comment [above](https://github.com/openstreetmap/iD/pull/9475/commits/1d64839b24d9c7f6d79b3810961d1f0a27c59d27#diff-849c38d4d2f9a07e9301d615da477fc56d407fca11d991fcccd50ba96a620fd5R93).

API.md
```diff
@@ -62,6 +62,7 @@ of iD (e.g. `https://ideditor-release.netlify.app`), the following parameters ar
 * __`validationError`__ - The issues identified by these types/subtypes will be treated as errors (i.e. Issues will be surfaced to the user but will block changeset upload). Each parameter value should contain a urlencoded, comma-separated list of type/subtype match rules.  An asterisk `*` may be used as a wildcard.<br/>
   _Example:_ `validationError=crossing_ways/highway*,crossing_ways/tunnel*`
 * __`walkthrough=true`__ - Start the walkthrough automatically
+* __`presetUrlOverwrite=https://example.com`__ - Overwrite the source for presets. Default is the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) release on jsdelivery ([Code](https://github.com/openstreetmap/iD/blob/develop/config/id.js#L2)). The URL has to serve the root of the id-tagging-schema repo after running `npm run dist`.
```
tyrasd, line 65:
```suggestion
* __`presetsUrl=https://example.com/`__ - Overwrites the source for presets. Default is the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) release on jsDelivr ([code](https://github.com/openstreetmap/iD/blob/develop/config/id.js#L2)). The URL has to serve the root of the id-tagging-schema repo after running `npm run dist` and allow [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) requests.
```

modules/core/file_fetcher.js
```diff
@@ -11,10 +12,21 @@ export { _mainFileFetcher as fileFetcher };
...
 //
 export function coreFileFetcher() {
+  const presetsVersion = packageJSON.devDependencies['@openstreetmap/id-tagging-schema'];
+  let presetsCdnUrl = presetsCdnUrlTemplate.replace('{presets_version}', presetsVersion);
+  // Allow to overwrite the preset source to enable testing PRs for the id-editor-presets repo.
+  const presetUrlOverwrite = utilStringQs(window.location.hash)?.presetUrlOverwrote;
+  if (presetUrlOverwrite) {
+    presetsCdnUrl = presetUrlOverwrite;
+    console.info('`presetUrlOverwrite` applied. Presets are fetched from', presetsCdnUrl);
+  }
```
tyrasd, line 22:
```suggestion
  // Allow to overwrite the presets source to enable testing PRs for the id-editor-presets repo.
  const presetsUrlOverwrite = utilStringQs(window.location.hash)?.presetsUrl;
  const presetsCdnUrl = presetsUrlOverwrite || presetsCdnUrlTemplate.replace('{presets_version}', presetsVersion);
```

modules/core/file_fetcher.js
tyrasd: For example:

```suggestion
    'deprecated': (presetsUrlOverwrite || presetsCdnUrlTemplate) + 'dist/deprecated.min.json',
    'discarded': (presetsUrlOverwrite || presetsCdnUrlTemplate) + 'dist/discarded.min.json',
    'preset_categories': (presetsUrlOverwrite || presetsCdnUrlTemplate) + 'dist/preset_categories.min.json',
    'preset_defaults': (presetsUrlOverwrite || presetsCdnUrlTemplate) + 'dist/preset_defaults.min.json',
    'preset_fields': (presetsUrlOverwrite || presetsCdnUrlTemplate) + 'dist/fields.min.json',
    'preset_presets': (presetsUrlOverwrite || presetsCdnUrlTemplate) + 'dist/presets.min.json',
```

tyrasd (changes requested):
Thanks for this!

See a few remarks below:

* I'd rename the hash parameter to `presetsUrl` for simplicity
* There is some special logic for handling the `{presets_version}`, see below the comment in `localizer.js`

<hr>

On a more general note, I'm currently slightly hesitant to allow loading data from arbitrary URLs via a user-supplied parameter. This is because in the aftermath of #8813 I had to fix a gazillion of potential html injection / [xss](https://en.wikipedia.org/wiki/Cross-site_scripting) attack vectors, some of which were directly related with strings from presets' translations. While I think that I filled all holes where strings are supplied from the outside (e.g. come in through OSM tags via the OSM API, or other external services), I might not have been that thorough where it comes to translated strings originating from transifex.

So, before merging this branch, I will perform a review of the relevant parts of the code where these presets' strings are rendered. That might take some extra time, though.

tyrasd:
with https://github.com/openstreetmap/id-tagging-schema/pull/800, this is not needed anymore, as a dedicated customized instance of iD for each PR branch will be built alongside the presets to be previewed

modules/core/localizer.js
tordans: @tyrasd even thought this PR is closed, I think your feedback and the fact that I was not able to read those details from the variable names hints that the variable names could be improved. The name itself you communicate if its a "template" – as in, something of the string will need to be replaced in order to work – or a URL that will be used as is.

This might be something to refactor at some point…

Issue #9447, opened 2023-01-02
`planned:man_made=pipeline` should not be an area


Dimitar5555:
### URL

https://www.openstreetmap.org/edit?way=1057595981#map=17/42.91492/22.92276

### Screenshot(s) or anything else?

![image](https://user-images.githubusercontent.com/19364673/210243175-d4645faa-a2ae-4569-8978-ad06f578d107.png)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.23.2

### Which browsers are you seeing this problem on?

Firefox

k-yle:
this was fixed in #9483 but it hasn't been merged yet

1ec5:
Fixed in #9483.

PR #9483, opened 2023-02-01
Fix wrong variable used to access tags

### LINKED ISSUE #9436: razed:waterway=stream should not be expected as an area
**Describe the bug**
warning tells that razed:waterway=stream should be an area
In most cases waterway:stream are lines: (17 M line / 41 k area features)

**To Reproduce**
[Steps to reproduce the behavior:
1.got to way https://www.openstreetmap.org/way/1109163077)
remove area=no

**Expected behavior**
no warning for this feature

biswajit-k:
Fixes #9436. In the code, `key` is `realKey` with any prefix removed, like in the given issue `realKey="razed:waterway"` and `key="waterway"`. 
So, to access the value field of tag, `realKey` should be used and not one with removed prefix, i.e `key`.

### DIFF
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -76,11 +76,11 @@ export function osmTagSuggestingArea(tags) {
     var returnTags = {};
     for (var realKey in tags) {
         const key = osmRemoveLifecyclePrefix(realKey);
-        if (key in osmAreaKeys && !(tags[key] in osmAreaKeys[key])) {
+        if (key in osmAreaKeys && !(tags[realKey] in osmAreaKeys[key])) {
             returnTags[realKey] = tags[realKey];
             return returnTags;
         }
-        if (key in osmAreaKeysExceptions && tags[key] in osmAreaKeysExceptions[key]) {
+        if (key in osmAreaKeysExceptions && tags[realKey] in osmAreaKeysExceptions[key]) {
             returnTags[realKey] = tags[realKey];
             return returnTags;
         }


biswajit-k:
The above changes fix the bug as below-
s1

However, I found that if we change the tag to `razed:waterway="rapids"` the issue panel pops up again which should not be the case as `rapids` is [allowed to take the value](https://wiki.openstreetmap.org/wiki/Key:waterway#:~:text=waterway-,rapids,-A%20natural%20barrier) of `way` , as shown below-
s2

Why did this happen? I went deeper and found that the `matchTags()` function [here ](https://github.com/openstreetmap/iD/blob/develop/modules/presets/index.js#L209) accessed keys without removing the `osmLifecyclePrefix` at `valueIndex = keyIndex[k]`.
This issue lies with all the tags containing `osmLifecyclePrefixes` like `proposed:`, `construction:`, `razed:` etc. 

How? This is because the tags are defined [only for root keys](https://github.com/openstreetmap/iD/blob/develop/modules/osm/tags.js#L46) like `{ waterway: { ... }, ... }` and don't include keys with prefixes like `{ razed:waterway: { ... }, ... }` and are accessed without removing their `osmLifecyclePrefixes`, like [here](https://github.com/openstreetmap/iD/blob/develop/modules/osm/tags.js#L109) and [here](https://github.com/openstreetmap/iD/blob/develop/modules/presets/preset.js#L242).

I thought about it and could come up with a few solutions as listed below-
1. Whenever accessing the tag objects like `osmAreaKeys`, `osmAreaKeysExceptions` which are defined [here](https://github.com/openstreetmap/iD/blob/develop/modules/osm/tags.js), we need to make sure to access the keys with their `osmLifecyclePrefixes` removed as done [here](https://github.com/openstreetmap/iD/blob/develop/modules/osm/tags.js#L72)(_although the wrong variable is accessed, which is corrected in the above PR_).
2. We can initialize all these tag objects as instances of a factory function say `osmTags` which has methods to access tags after removing their `osmLifecyclePrefixes` as below-

```

function osmTag(tags) {

  var _tags = {};
  
  if(arguments.length) {
  	_tags = tags;
  }
  
  return {

    get: function(key) {
    	return _tags[osmRemoveLifecyclePrefix(key)];
    },
    set: function(tags) {
    	_tags = tags;
    },
    has: function(key) {
    	return osmRemoveLifecyclePrefix(key) in _tags;
    }

  }

}
```

Then objects can be accessed like `osmAreaKeys.has(k)` and `osmAreaKeys.get(k)`.

3. We can define custom `get` and `has` methods of objects so that whenever the keys are accessed we internally remove the prefixes and return the values, it would work without changing the way we access objects like `osmAreaKeys[k]` and `k in osmAreaKeys`. One way I found to do this is by using the[ `Proxy` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)(I don't know if this is feasible, but just got an idea so shared).

I believe that the above solutions could help in solving this and prevent bugs from popping up in future. @tyrasd I understand that you have been putting in long hours at work, and I truly appreciate your dedication. If it's not too much of an inconvenience, would you be able to spare some time to provide suggestions? I would greatly value your input and insights. Thank you in advance for your time and consideration.

Please don't mind if I overlooked or said something incorrectly. I'm still learning😄!

k-yle:
For reference, this PR closes #9436, #9491, #9447, https://github.com/openstreetmap/id-tagging-schema/issues/861 and https://github.com/openstreetmap/id-tagging-schema/issues/920

tyrasd (approved):

Higuys153:
If its been merged when will it be up on the website?

k-yle:
is someone who has permission (@tyrasd @nickrsan?) able to close the issues listed in this comment: https://github.com/openstreetmap/iD/pull/9483#issuecomment-1515419261 and ideally link those issues to this PR too?

nickrsan:
It looks like @1ec5 took care of the remaining open ones on this repository - I have triage permissions here, but not on id-tagging-schema, so someone else will need to take care of https://github.com/openstreetmap/id-tagging-schema/issues/920 which looks to be the only one remaining open.

Thanks for flagging these as ready to close.

PR #9492, opened 2023-02-07
add 4 tags which imply oneway=yes

k-yle:
Most of these are rarely used but I don't see any harm in adding them to the list


tyrasd (approved):

tyrasd:
Thanks!

Issue #8997, opened 2022-02-23
KartaView images not shown completely when resizing+zooming


FinixFighter:
### URL

https://www.openstreetmap.org/edit#map=19/59.40386/17.93796

### How to reproduce the issue?

Go to iD, enable KartaView photo overlay, open any KV image, change the size of the image window (from the top right corner, make it bigger), zoom in and then try to move the image: you'll see that you can reach the top of the image and the left side, but you can't reach the bottom and the right side of the images. It's like if the image gets cropped on two sides, the right and the bottom.

### Screenshot(s) or anything else?

Watch video at: https://odysee.com/@Finix:d/kvbug022022:c

### Which iD Editor versions do you see the issue on?

Released version at openstreetmap.org/edit

### Which browsers are you seeing this problem on?

Firefox

derkrasseleo:
I am also experiencing this issue. But it doesn't seem to be browser-specific. I tested with brave and firefox.

derkrasseleo:
This issue still persists. It seems that the image is cropped every time the outer frame is moved, or the image is zoomed in, which is not the expected behavior.

tyrasd:
fixed in 59734d172 (and d0e061c47)

PR #9509, opened 2023-02-22
Street level photo from the Norwegian Public Road Administration

noenandre:
This PR tries to bring street level photos from the Norwegian Public Road Administration to iD.
The service is called "Vegbilder" which simply means road photos.

The Norwegian OSM community has obtained [permission ](https://forum.openstreetmap.org/viewtopic.php?pid=712312#p712312) to use all open services from the Norwegian road administration.

Every national and every county road is photographed bi-annual, and the service contains regular images and from the year 2021 also 360° image spheres, so this will be a great addition to iD for Norwegian users.

The PR still has some bugs and kinks to straiten out, and I do hope that iD delopers have some future capacity to eventually get this all over the finish line. 
One thing I don't want to tackle in this PR is the lack of a geo-filteing for street level photos overlays. Photos from the Norwegian Public Road Administration is of course only available in Norway, and id should only show "Vegbilder" as a provider when zoomed in there, similarly to aerial photos.


noenandre:
Update:
A "Vegbilder" viewer need a viewer for both plane images and 360° images.
I have relied on the code for the Streetlevel viewer to make a viewer for 360° images, and Kartaview viewer to make my own viewer for plane photos. 
But the Kartaview viewer has some issues. It makes a new d3 zoom behaviour object on photoviewer resize but forgets to register/call it (#8997), and it uses the css property `object-fit: cover` which cut images that don't have the same ratio as the photo-viewer, when the desired behaviour is to hide the overflow, so the whole images is possible to pan over.
I have solved those issues, but also created a new one. 
Now the image won't zoom in anywhere near where the mousepointer is placed when using the scroll wheel, and it gets worse if you decrease the width of the photoviewer.

noenandre:
This PR is ready for review! I will be happy yo create a new PR if preferred.

modules/services/vegbilder.js
```diff
@@ -0,0 +1,609 @@
...
+
+export default {
+
+  init: async function () {
+    if (!_vegbilderCache) {
+      await this.reset();
+    }
+
+    this.event = utilRebind(this, dispatch, 'on');
+  },
```
tyrasd: I've noticed that this can cause an exception in [`renderer/photos.js`](https://github.com/openstreetmap/iD/blob/ddc9c6c67460e20a9531e6c759c61bebc5269b80/modules/renderer/photos.js#L209) (when opening a photo via URL parameter is requested).

I think we should not `await` the cache reset before binding the `on` method here:

 ```suggestion
  init: async function () {
    this.event = utilRebind(this, dispatch, 'on');
    if (!_vegbilderCache) {
      await this.reset();
    }
  },
```

But then `loadImages` needs to check if the cache is already populated:

```javascript
  loadImages: async function (context, margin) {
    margin ??= 1;
    if (!_vegbilderCache) {
      await this.reset();
    }
    const wfslayers = filterAvailableLayers(context.photos());
```

The same also in the `sequences` method plus some added async/await on top.

<hr>

Maybe there's a better option to fix this.

tyrasd (commented):
Hey. This is great!

Just a bit of first feedback from trying this out:

* Optimally, the layer should only show up and request/load layers if the user maps in the region this service covers. Similar to how the background layers and certain presets only show up if the user is mapping in the respective region.
* I've encountered a somewhat consistently thrown exception by `pannelum.js` (after being called by `_pannellumViewer.loadScene(key, pitch, yaw);` in `selectPhoto`)
  ````
  Uncaught TypeError: Cannot set properties of undefined (setting 'src')
    at FileReader.<anonymous> (pannellum.js:50:219)
  ````
  But only for the first 360°-pano to be displayed and it also does not seem to cause any noticeable issues. Did you encounter this as well?

noenandre:
> I've encountered a somewhat consistently thrown exception by pannelum.js (after being called by _pannellumViewer.loadScene(key, pitch, yaw); in selectPhoto)

Yes, I see that too, I don't know why I didn't catch that sooner. If I call `pannellum.addScene` without the option `autoLoad: true` the error seems to go away. `pannellum.loadScene` is called subsequently anyway.

> Optimally, the layer should only show up and request/load layers if the user maps in the region this service covers. Similar to how the background layers and certain presets only show up if the user is mapping in the respective region.

Yes, ideally it should. I am willing to work on such a feature, whenever this PR is accepted.

modules/services/vegbilder.js
noenandre: I have rearranged `init` as you suggested, and rearranged  `reset` to initialize a empty cache right away, before waiting for `fetchAvailableLayers`. Then it should not be necessary to check if the cache is there, in every function accessing the cache. Is that a better solution?

tyrasd (approved):
> the layer should only show up and request/load layers if the user maps in the region this service covers

d7902e138 and edb7c3b0f implement this.

While at it, I've also fixed a few issues I found (see additional commits in this branch), e.g. that zooming on non-360° photos was not preserving the center at the mouse location. Maybe you could double-check if I have not introduced any new bugs? :wink:

noenandre:
The "plane" photo viewer works great now. Can't find any bugs, old or new.

PR #9511, opened 2023-02-26
support multiple `prerequisiteTag`s for presets

k-yle:
`schema-builder` only allows you to define one `prerequisiteTag` per preset. This PR, along with ideditor/schema-builder#87, changes the format to allow multiple possible values.


tyrasd (approved):
Do you have an example field in mind, where this functionality would be useful?

I'm asking myself if we could need a similar extension also for `valueNot` as well?!

k-yle:
The case I had in mind was `seamark:bridge:clearance_height`, `seamark:bridge:clearance_height_open`, etc. where [the tag(s) depend on the type of bridge](https://wiki.osm.org/Key:seamark:bridge:category).

the `electrified` tag on railways already gets around this limitation by using `valueNot` `no` for `voltage`

another thought I had, but this could be a future PR: It would be nice to support regexs, so that [`seamark:*:colour_pattern`](https://wiki.osm.org/Seamarks/Colours#Colour_patterns) only shows up if `seamark:*:colour` contains multiple colours (i.e. the `:colour` field contains a `;`)

tyrasd (approved):

PR #9522, opened 2023-03-04
validate suspicious names using the user's language

### LINKED ISSUE #9455: Suspicious names in non-English languages are not removed
**Describe the bug**
When a `building=garage` has `name=garage` iD suggest to remove the suspicious name. On the other hand if the same building has `name=гараж` (meaning "garage" in Bulgarian), iD doesn't suggest to remove that name. The same applies for `building=garages + name=гаражи`. 

**To Reproduce**
1. Go to https://www.openstreetmap.org/edit?way=469468413

**Expected behavior**
There should be a warning, like the one that shows for ways tagged with `building=garage + name=garage`.

**Screenshots**
image
image

**Additional context**

k-yle:
Closes #9455. Currently the 'suspicious names' validator only considers the tag value, so it effectively only works in british english. 

This PR makes the validator also consider the name of the preset, which is localized.


modules/validations/suspicious_name.js
```diff
@@ -45,9 +45,17 @@ export function validationSuspiciousName() {
...
+    if (!presetName) return false;
+
+    return name.toLowerCase() === presetName.toLowerCase();
+  }
+
+  /** @param {string} name @param {string} presetName */
+  function isGenericName(name, tags, presetName) {
     name = name.toLowerCase();
-    return nameMatchesRawTag(name, tags) || isGenericMatchInNsi(tags);
+    return nameMatchesRawTag(name, tags) || nameMatchesPresetName(name, presetName) || isGenericMatchInNsi(tags);
```
1ec5, line 58: How does this avoid warning about NSI entries that add a `name` matching the brand name? Does the validator accept it because the NSI preset’s name has a compound “Brand – Preset” format?

k-yle: It works because the whole validator does nothing if a feature has `brand:wikidata` or `operator:wikidata`, even the  `not:name` check won't run...

https://github.com/openstreetmap/iD/blob/8bb1d39431f2f8a076c56e6686d45f32b31506fb/modules/validations/suspicious_name.js#L149-L151

1ec5: > even the `not:name` check won't run

Interesting, that sounds like a bug, worth reporting separately.

1ec5 (approved):
Thanks!

Suggested entry for the changelog after merging:

> The Suspicious Names validator warning now compares the Name field to the preset’s name in the user’s language, not just the raw tag value (typically in British English).

tyrasd (approved):

tyrasd:
FYI: I've enhanced this functionality a little bit:

* also warn if a name matches one of the preset's `aliases`
* add an ignore-list for special presets which should not be checked in this validator (for special presets which are similar to NSI presets, e.g. #10884)

PR #9532, opened 2023-03-06
Fix issue #8997.

noenandre:
Kartaview viewer should use a new zoom behaviour after viewer resize. This is a fix for #8997.

### DIFF
--- a/modules/services/kartaview.js
+++ b/modules/services/kartaview.js
@@ -298,13 +298,20 @@ export default {
             .attr('class', 'kartaview-image-wrap');
 
 
+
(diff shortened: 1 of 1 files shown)


tyrasd:
Sorry, I did overlook this fix. I've fixed this in a slightly different way in [59734d172 and d0e061c47](https://github.com/openstreetmap/iD/compare/118335b51674e8278cd009f582006933754a567f..d0e061c4725b92c1c1e534897cfc7056d4bdf053) now. Thank you nonetheless. :bow:

PR #9545, opened 2023-03-14
increase the limit for recent presets from 4 to 8

### LINKED ISSUE #8895: Allow users to customise number of recent presets
At the moment, iD shows the four most recently used preset at the top of the list. 

However sometimes this means recently used presets drop off the list if you are editing an area which needs a variety of presets. Add a car park, a building, some trees, a pitch and then you're having to type in "car" again to add another car park.

It would be useful if this could be increased - the preferences panel would be a good places to have a box with an up/down selector.

### LINKED ISSUE #11729: Request: more than four recent feature types?
### Description

When drawing features, a user must select a feature type. There seem to be four most recent features presented in the sidebar, but no more. It would be nice if it remembered more than four -- like six or eight. I think four is a bit too little. 

In my use case I find myself drawing houses, buildings (garages), grass, wood, specialized buildings like churches, scrub, ponds --- and with only four recent feature types, I often need to search for e.g. House again, rather than it just being there for me to click.

Thanks!

### Screenshots

Image

k-yle:
Closes #11729, ~~Closes~~ Partially addresses #8895. Superseeds #9454


Zaczero:
I don't think this really resolves #8895.

1ec5:
If you’re interested in implementing #8895, https://github.com/openstreetmap/iD/issues/8895#issuecomment-1258621496 has some starting points.

Crazycolbster:
Well, it's been a year since any activity has happened on here, but 8 presets would be extremely useful for my workflow on ID. I'm constantly having to research for elements because it just barely falls off the list.

Zaczero:
> Well, it's been a year since any activity has happened on here, but 8 presets would be extremely useful for my workflow on ID. I'm constantly having to research for elements because it just barely falls off the list.

The best workaround is to use Rapid which has this feature merged. It's a fork of iD.

tordans:
FYI: https://github.com/openstreetmap/iD/pull/11269 Will probably also help for cases like these.

matkoniecz:
I tested it and bump to 8 is in fact really nice. On my screen 12 entries can fit and ones at 9, 10, 11, 12 are occasionally useful. Beyond that - not useful but not harmful either.

matkoniecz (commented):
though having 30 recent ones may make folder use annoying (as now you will need to scroll through up to 30 entries)

maybe going back to 8 makes more sense?

(I am not using folder groups myself, but probably someone does)

PR #9567, opened 2023-03-30
Fix link to schema builder

riQQ:
The error was introduced with e466ad4bf86047dd56895d5315f8d13d7c19cc35 / #9546


arch0345 (approved):
Thanks for catching this!

PR #9573, opened 2023-04-01
Copy crossing:markings to generated node for crossing_ways resolution

### LINKED ISSUE #9525: crossing:markings isn't automatically set when a crossing is created
### URL

_No response_

1. Draw a way across a road and tag it as crossing=unmarked or crossing=marked
2. As expected, an issue will pop up saying that the two ways should intersect or be layer-separated
3. Select "Connect the features"
4. The crossing node that is created will have a new issue, saying that the feature has incomplete tags and should be upgraded with crossing:markings=no or crossing:markings=yes. If this tag is needed, it should be automatically added to the node upon creation. As it currently stands, the issue needs to be resolved individually and repeatedly by the user

_No response_

2.25.1

Firefox

vpzomtrrfrt:
Should resolve #9525

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -148,7 +148,9 @@ export function validationCrossingWays(context) {
                     var pathFeature = entity1IsPath ? entity1 : entity2;
                     if (['marked', 'unmarked', 'traffic_signals', 'uncontrolled'].indexOf(pathFeature.tags.crossing) !== -1) {
                         // if the path is a crossing, match the crossing type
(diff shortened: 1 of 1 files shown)


jtracey:
This has a bug: if you connect a crossing that has no `crossing:markings` tags, the resulting node renders as having the tag set to "multiple values".

(edit to add: if it wasn't clear, this is because you're setting the value even when there is no such member to copy from; see my PR above for how I resolved this)

vpzomtrrfrt:
#9586 is probably better, I'll close this in favor of that one

Issue #2676, opened 2015-06-03
Conditional editable features


franckt:
Map data tool can hide or show features based on filters but is it possible to make some features non editable or selectable ? 

i currently work on an indoor editing project with id where many features are at the same place but on differents levels

For example :
Room features on a floor are displayed but not selectable so we can draw on top it on an another floor
 
Is there a good or existing way to start adding this feature ?

bhousel:
It would be pretty straightforward to not display them by hacking up `features.js` code.  You could replace the existing feature filters with ones that filter on `level`.

But to display but not edit requires a bit more.  There are a bunch of places within iD code that prevent editing by calling [`context.hasHiddenConnections`](https://github.com/openstreetmap/iD/blob/master/js/id/id.js#L250-L254).  I don't remember if all actions check this or just certain ones.

But maybe a better way going forward would be to add something in `context` that *all* editing functions must check first to know whether editing of a given entity is allowed - and the `hasHiddenConnections` code would just register itself there as one possible reason lock an entity.

bhousel:
WIP on this feature: 8467433
It is called `iD.restrictions` and will allow you to add custom rules to prevent users from editing things.

PR #9584, opened 2023-04-12
Add custom features filter functionality

### LINKED ISSUE #2580: Custom Feature Filtering
The user should be able to filter based on tag and key with the operators */!=/=. Multiply filters should be supported at the same time.

Currently custom iD instances has to modify `renderer/features.js` to get feature filtering working. 
An option to disable the default feature filtering(/or an API) together with custom filtering powered by presents/taginfo(which can be custom), would make the setup for custom instances shorter.

Example use case:
 - Multiply elements with `start_date` is currently visible on the map, we are only interested in elements tagged with `start_date=1959`, we could filter it with "start_date != 1959".

 - Or if someone is doing IOF mapping and need to filter away a ISSOM/ISOM line based on standard value("line=101").

Ugly mockup:

ugly mockup

willemarcel:
Closes #2580

Related to #2676. It adds a new option in the Map Features section that allows the user to set a key=value filter.

You can see it in action in the video below:

<video src="https://user-images.githubusercontent.com/666291/231590706-94ba07df-93b9-4663-96e7-40148b33a55b.mov"></video>

### DIFF
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -189,8 +189,8 @@ export function rendererFeatures(context) {
 
     defineRule('aerialways', function isPiste(tags) {
         return tags.aerialway &&
-            tags.aerialway !== 'yes' &&
-            tags.aerialway !== 'station';
+        tags.aerialway !== 'yes' &&
+        tags.aerialway !== 'station';
     });
 
     defineRule('power', function isPower(tags) {
@@ -203,7 +203,7 @@ export function rendererFeatures(context) {
             traffic_roads[tags.highway] ||
             service_roads[tags.highway] ||
             paths[tags.highway]
-        ) { return false; }
+            ) {return false; }
 
         var strings = Object.keys(tags);
 
@@ -221,15 +221,20 @@ export function rendererFeatures(context) {
         return (geometry === 'line' || geometry === 'area');
     });
 
-
+    defineRule('custom', function isCustom(tags) {
+        const [key, value] = prefs('map-features-custom').split('=');
+        if (key) {
+            if (value === '*') return tags[key];
+            return tags[key] === value;
+        }
+    });
 
     features.features = function() {
         return _rules;
     };
 
-
     features.keys = function() {
-        return _keys;
+    return _keys;
     };
 
 
@@ -282,7 +287,6 @@ export function rendererFeatures(context) {
         }
         if (didEnable) update();
     };
-
 
     features.disable = function(k) {
         if (_rules[k] && _rules[k].enabled) {
@@ -380,6 +384,12 @@ export function rendererFeatures(context) {
         });
 
         _cache = {};
+    };
+
+    features.updateCustom = function() {
+        features.reset();
+        features.disable('custom');
+        features.enable('custom');
     };
 
     // only certain relations are worth checking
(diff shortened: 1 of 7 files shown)


tordans:
This looks like a nice addition. However, I think we should make sure that there are no side effects that might require additional work later … and if someone wants to invest this additional work. I could see this becoming a source of quite a few follow up feature requests and side effects with other features.

Is there a way to test this now without running it locally / is there a preview URL?

_A quick list of thoughts:_

* The code changes have a few unrelated re-formattings that are likely uninteded(?)
* Until now, the list of filter options is exclusive (more or less), so one option does not interact with another option. This changes now… – do we need to handle those cases? Show something to the user? Gray out options that are handled by the custom option already?
* Which filter I select is stored in the hash URL, right? Is this also the case for the custom filters? Does this maybe break user expectations when sharing a pre configured map URL?
* This is a power user feature and the input field nudges power user to think of it like the overpass turbo input which allows for more than just one `key=value`. To clarify, that (for now) this can only do key+value, the UI could be two inputs with an explicit label. That could also be the place where the `*` case for "any key" is explained.

willemarcel:
@tordans Thanks for your comments.

> Is there a way to test this now without running it locally / is there a preview URL?

I've deployed it on https://6446cbf640335e2652b8035f--bright-taffy-b33606.netlify.app

> * Until now, the list of filter options is exclusive (more or less), so one option does not interact with another option. This changes now… – do we need to handle those cases? Show something to the user? Gray out options that are handled by the custom option already?

This is the part that doesn't make me entirely happy. To make it easier to be understood, I think the best solution would be to disable all the other options when the custom is selected, or move it to another section of the UI and make it override the Map Features preferences.

> Which filter I select is stored in the hash URL, right? Is this also the case for the custom filters? Does this maybe break user expectations when sharing a pre configured map URL?

The custom filter is also stored in the URL and can be shared. It works the same way the custom background settings work.

modules/renderer/features.js
```diff
@@ -221,15 +221,20 @@ export function rendererFeatures(context) {
         return (geometry === 'line' || geometry === 'area');
     });
 
-
+    defineRule('custom', function isCustom(tags) {
+        const [key, value] = prefs('map-features-custom').split('=');
```
k-yle, line 225: I love this feature, I'm using it in a fork of iD. There is a small bug though, `prefs` can return `null` when iD is opened for the first time, which causes a crash here. Something like this would fix it:

```suggestion
        const [key, value] = prefs('map-features-custom')?.split('=') || [];
```

tordans:
@k-yle what is your experience with this in your fork, is it ready to be merged or are there things we should address?
Specifically the UX issue we talked about above…

I ran into the lacking filtering options again, today and it looks like this PR would be a solution for this.
In my case, I need to get rid of everything that is `level=-1` and `level=-2` on https://www.openstreetmap.org/edit?note=4139556#map=20/52.51122/13.40564

k-yle:
@tordans there are a few things I think we need to consider first:
1. [This crash](https://github.com/openstreetmap/iD/pull/9584#discussion_r1470702319)
2. It's not possible to add features in this mode, even if they match the custom filter
3. Should we allow Regular Expressions? Otherwise filters like `level`=(`1` or `2`) would not be possible
4. Currently, it does not work for vertices, since iD can't render vertices if the parent way is hidden.

I can't think of an easy way to implement the last point, but the other 3 are trivial changes

matkoniecz:
@willemarcel do you plan to fix mentioned crash?

PR #9587, opened 2023-04-17
handle cycleway:both=*

### LINKED ISSUE #743: Handle `cycleway:both` similar to `cycleway`
Usage: https://taginfo.openstreetmap.org/search?q=cycleway
Anzahl | Key
-- | --
945 788 | [cycleway](https://taginfo.openstreetmap.org/keys/cycleway)
904 942 | [cycleway:both](https://taginfo.openstreetmap.org/keys/cycleway%3Aboth)

**Describe the bug**
Right now, the directionalCombo only supports the `cycleway` field, not the "in use" `cycleway:both` field.

IMO, we should first look for a work around, like https://github.com/openstreetmap/id-tagging-schema/pull/723 tried.
Could we maybe duplicate the `cycleway` preset for the `both`-case (unsearchable) in order to get the data visible?

### LINKED ISSUE #9212: Cycleway:both is ignored
### URL

_No response_

1. Select a road with cycleway:both=*
2. The tagged cycleway is not displayed under "bike lanes"

image

This may be confusing to newcomers who may not think to look at the tags themselves.

As a side note, cycleway:both=* may be a better default tagging than cycleway=* as it specifies that a cycleway is in fact on both sides, similar to how the user fills in left and right in the iD interface.

_No response_

_No response_

k-yle:
Closes #9212 and Closes https://github.com/openstreetmap/id-tagging-schema/issues/743

This PR makes the cycle lane input consider `cycleway:both=*` if `cycleway=*` is missing

The UI will never set `cycleway:both=*`, it will continue to use `cycleway=*` - see https://github.com/openstreetmap/iD/issues/9212#issuecomment-1366655208 for background info. cc: @tordans 

To test: visit https://kyle.kiwi/iD

<details>
<summary>If a feature has both <code>cycleway:both=*</code> and <code>cycleway=*</code>, and they have different values, then that's obviously a mistake, so the UI shows "Multiple Values":</summary>

</details>


mxdanger:
Hi, are there any updates on the status of PR? It’s been more than a year.

tyrasd (commented):
Maybe we need to add a check if the `commonKey` is already a tag with a `:both` suffix?

k-yle:
done, and added unit tests. 

id-tagging-schema can now decide whether the main key should be `cycleway` or `cycleway:both`. iD will respect the preset's choice, but still recognise the other tag.

tordans:
This is great. Hope this makes it into the next release.

> id-tagging-schema can now decide whether the main key should be `cycleway` or `cycleway:both`. iD will respect the preset's choice, but still recognise the other tag.

To double check: The schema still only specifies one "both" key for a directionalCombo and then iD will recognize both variation internally. So the idea to have the schema to list both variations (in order of preference) is obsolete, right? (I think there was an issue on that that we should closed afterwards.)
I think we should extend the docs in the schema [on the directionalCombo](https://github.com/search?q=repo%3Aideditor%2Fschema-builder+directionalCombo+language%3AMarkdown&type=code&l=Markdown) to hint to data consumers that they are expected to handle this logic in app. I am not aware of any consuming app that does this kind of UI processing but I am still hopeful one will start using the schema in this way in the future.

k-yle:
Docs PR: https://github.com/ideditor/schema-builder/pull/179

---

> To double check: The schema still only specifies one "both" key for a directionalCombo and then iD will recognize both variation internally. So the idea to have the schema to list both variations (in order of preference) is obsolete, right

Yes those two statements are correct. There are now two valid ways to define the preset:

1. ```js
   { "key": "cycleway",      "keys": ["cycleway:left", "cycleway:right"] }
    ```
2. ```js
   { "key": "cycleway:both", "keys": ["cycleway:left", "cycleway:right"] }
    ```

PR #9588, opened 2023-04-17
consider other tags for labelling features

### LINKED ISSUE #8966: Route relation label should include unsigned_ref if ref is unset
`unsigned_ref` is an alternative to `ref` for when the route number isn’t signposted or is signposted too obscurely for wayfinding. A route relation tagged with `unsigned_ref` but not `ref` ([example](https://www.openstreetmap.org/relation/1613724)) should have a label that includes the `unsigned_ref` so that it remains identifiable in a list of relations.

### LINKED ISSUE #9526: Display 'distance' instead of 'ref' for highway milestones
### Description

Please, display the `distance` value instead of `ref` to better identify highway milestones on the editor.

image

If you have more than one on screen, it's hard to distinguish which is which since all of them will display the same ref number if they all belong to the same highway route.

image

Moreover, displaying the distance can be an aid so that you don't get lost while mapping long stretches of road with few reference points.

Thank you.

Igor

k-yle:
Closes #8966, Closes #9526

This PR considers other name tags when labelling features in the UI and on the map.

These tags are considered: [`alt_name`](https://wiki.osm.org/Key:alt_name), [`official_name`](https://wiki.osm.org/Key:official_name), [`loc_name`](https://wiki.osm.org/Key:loc_name), [`loc_ref`](https://wiki.osm.org/Key:loc_ref),  [`unsigned_ref`](https://wiki.osm.org/Key:unsigned_ref), [`seamark:name`](https://wiki.osm.org/Key:seamark:name), [`sector:name`](https://wiki.osm.org/Key:sector:name), [`addr:housename`](https://wiki.osm.org/Key:addr:housename), [`lock_name`](https://wiki.osm.org/Key:lock_name) and in some cases [`distance`](https://wiki.osm.org/Key:distance) and [`railway:position`](https://wiki.osm.org/Key:railway:position) (see #9526)

#8707 and #8440 turned out to be controversial but I can't think of any reason why this change would be unwelcome (?)

### DIFF
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -55,6 +55,16 @@ export function svgLabels(projection, context) {
         ['point', 'shop', '*', 10],
         ['point', 'tourism', '*', 10],
         ['point', 'camp_site', '*', 10],
+        ['*', 'alt_name', '*', 12],
+        ['*', 'official_name', '*', 12],
+        ['*', 'loc_name', '*', 12],
+        ['*', 'loc_ref', '*', 12],
+        ['*', 'unsigned_ref', '*', 12],
+        ['*', 'seamark:name', '*', 12],
+        ['*', 'sector:name', '*', 12],
+        ['*', 'lock_name', '*', 12],
+        ['*', 'distance', '*', 12],
+        ['*', 'railway:position', '*', 12],
         ['line', 'ref', '*', 12],
         ['area', 'ref', '*', 12],
         ['point', 'ref', '*', 10],
@@ -297,7 +307,7 @@ export function svgLabels(projection, context) {
                 var matchVal = labelStack[k][2];
                 var hasVal = entity.tags[matchKey];
 
-                if (geometry === matchGeom && hasVal && (matchVal === '*' || matchVal === hasVal)) {
+                if ((matchGeom === '*' || geometry === matchGeom) && hasVal && (matchVal === '*' || matchVal === hasVal)) {
                     labelable[k].push(entity);
                     break;
                 }
--- a/modules/util/util.js
+++ b/modules/util/util.js
@@ -242,10 +242,34 @@ export function utilDisplayName(entity, hideNetwork) {
     }
 
     if (keyComponents.length) {
-        name = t('inspector.display_name.' + keyComponents.join('_'), tags);
+        return t('inspector.display_name.' + keyComponents.join('_'), tags);
     }
 
-    return name;
+    const alternativeNameKeys = [
+        'alt_name',
+        'official_name',
+        'loc_name',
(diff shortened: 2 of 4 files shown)


modules/util/util.js
```diff
@@ -215,10 +215,35 @@ export function utilDisplayName(entity) {
...
 
     if (keyComponents.length) {
-        name = t('inspector.display_name.' + keyComponents.join('_'), tags);
+        return t('inspector.display_name.' + keyComponents.join('_'), tags);
     }
 
-    return name;
+    const alternativeNameKeys = [
+        'alt_name',
+        'official_name',
```
1ec5, line 250: Note that until we add fields for keys such as `official_name` in openstreetmap/id-tagging-schema#215, a user unfamiliar with the raw tag editor will perceive a phantom label if it comes from one of these keys. Not a showstopper in my opinion, but it would increase the pressure for finally landing those changes.

/ref https://github.com/openstreetmap/id-tagging-schema/issues/77#issuecomment-1165137251

modules/svg/labels.js
```diff
@@ -57,6 +57,17 @@ export function svgLabels(projection, context) {
...
+        ['*', 'official_name', '*', 12],
+        ['*', 'loc_name', '*', 12],
+        ['*', 'loc_ref', '*', 12],
+        ['*', 'unsigned_ref', '*', 12],
+        ['*', 'seamark:name', '*', 12],
+        ['*', 'sector:name', '*', 12],
+        ['*', 'addr:housename', '*', 12],
+        ['*', 'lock_name', '*', 12],
+        ['*', 'distance', '*', 12],
+        ['*', 'railway:position', '*', 12],
```
1ec5, line 67: Should this be limited to points? [The `railway:position` documentation](https://wiki.openstreetmap.org/wiki/Key:railway:position) says this key can be used on any kind of feature, but it seems to be [extremely rare](https://taginfo.openstreetmap.org/keys/railway%3Aposition#overview) on ways and relations, and all of the presets that use the Milestone Position field are for nodes only.

modules/util/util.js
1ec5: This landed, so it’s no longer a blocker.

modules/svg/labels.js
k-yle: sure, we could limit it to points, but also keep in mind that most of the geometry rules in this array have no effect, because of these 3 wildcard entries:

https://github.com/openstreetmap/iD/blob/9cf0a20592ce8792aa14f2faeb6b30f122101434/modules/svg/labels.js#L74-L76

1ec5 (approved):

tyrasd (approved):

PR #9593, opened 2023-04-23
Altering the text when an NSI suggestion is displayed.

### LINKED ISSUE #9592: Altering the text when an NSI suggestion is displayed.
At the moment, any [Name Suggestion Index](https://github.com/osmlab/name-suggestion-index/) (NSI) POI that is suggested is accompanied by some text that I don't think is informative enough to the end user, as it refers to "_common tags_" rather than about branding. It also merges the POI type and name together as one long sentence, such as "_Burger Fast Food burger king_". There's also no reference to the fact the POI suggestion comes from the NSI.

nsi-text

I'd like to suggest a change to the wording, to specifically refer to the POI being a suggestion, and highlighting that the suggestion is powered by the NSI, for example:

nsi-text-new

Aditya-13:
Fixed #9592 

This pr shows a NSI reference when any POI suggestion comes.

It also differentiates between the POI type and the name such as "The Burger Fast Food POI "burger king".

*This is my first contribution to open-source. Any feedback is welcome.*

Cheers.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1860,6 +1860,7 @@ en:
         message: "{feature} looks like a common feature with nonstandard tags"
         message_incomplete: "{feature} looks like a common feature with incomplete tags"
         reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
+        nsi_reference: "The suggestions above are powered by the Name Suggestion Index."
     point_as_area:
       message: '{feature} should be a point, not an area'
     point_as_line:
--- a/modules/util/util.js
+++ b/modules/util/util.js
@@ -260,7 +260,7 @@ export function utilDisplayLabel(entity, graphOrGeometry, verbose) {
     var presetName = preset && (preset.suggestion ? preset.subtitle() : preset.name());
 
     if (verbose) {
-        result = [presetName, displayName].filter(Boolean).join(' ');
+        result = [`The ${presetName} POI,`, `"${displayName}"`].filter(Boolean).join(' ');
     } else {
         result = displayName || presetName;
     }
--- a/modules/validations/outdated_tags.js
+++ b/modules/validations/outdated_tags.js
@@ -220,6 +220,11 @@ export function validationOutdatedTags() {
           return `tagDiff-cell tagDiff-cell-${klass}`;
         })
         .html(d => d.display);
+      
+      enter
+        .append('div')
+        .attr('class', 'issue-reference')
+        .call(t.append(`issues.outdated_tags.${prefix}nsi_reference`));
     }
   }
 


modules/util/util.js
```diff
@@ -260,7 +260,7 @@ export function utilDisplayLabel(entity, graphOrGeometry, verbose) {
     var presetName = preset && (preset.suggestion ? preset.subtitle() : preset.name());
 
     if (verbose) {
-        result = [presetName, displayName].filter(Boolean).join(' ');
+        result = [`The ${presetName} POI,`, `"${displayName}"`].filter(Boolean).join(' ');
```
1ec5, line 263: As @bhousel noted in https://github.com/openstreetmap/iD/issues/9592#issuecomment-1559651686, the word “POI” can be unintuitive to the user. Most newcomers to OSM aren’t mapping professionals who would be familiar with this acronym or “point of interest”. In fact, some OSM software and documentation uses “POI” to mean a node that isn’t part of a way. (iD calls these “points”.) Finally, the name suggestion index includes some feature types that aren’t normally described as POIs, such as flagpoles and power lines, in order to standardize how their operators are tagged.

The example given in https://github.com/openstreetmap/iD/issues/9592#issue-1679566314, “The Burger Fast Food burger king”, is awkward primarily because the preset is named “Burger Fast Food”. “Fast food” is a mass noun, so it’s weird to put “the” before it and a name after it. In everyday speech, I think you’d say something like “The burger joint ‘burger king’”, but that would be quite informal as a preset name, and it doesn’t generalize well to other presets.

I wonder if it would be possible to restructure the sentence somewhat. For instance:

> “burger king” looks like a Burger Fast Food with nonstandard tags

1ec5, line 263: The `filter` and `join` calls were a clever way to display only `presetName` if `displayName` is empty and vice versa. But now that the array’s only item is always a nonempty string, these calls have no effect. The code needs to handle these edge cases in some other way. While we’re at it, we should make the new format string localizable.

data/core.yaml
```diff
@@ -1860,6 +1860,7 @@ en:
         message: "{feature} looks like a common feature with nonstandard tags"
         message_incomplete: "{feature} looks like a common feature with incomplete tags"
         reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
+        nsi_reference: "The suggestions above are powered by the Name Suggestion Index."
```
1ec5, line 1863: As far as I can tell, this is the only place in the UI where “Name Suggestion Index” is mentioned. The user might be curious about NSI, but they would have to Google it on their own. Maybe we could provide a link?

modules/util/util.js
Aditya-13: Then what should i put here as text?

1ec5: The tricky thing is that this method is not actually about the validation warning you’re trying to fix. It’s a more widely used method that generates the name of a feature in many contexts that have nothing to do with POIs or NSI, everything from the warning about crossing ways to the confirmation message that you’ve just pasted a feature.

Let’s look at the code that currently calls this method in order to generate the message reported in #9592:

https://github.com/openstreetmap/iD/blob/43784e2eff766577b0bb39d5f2010c0d8f968903/modules/validations/outdated_tags.js#L185-L191

This is taking one of the following strings and replacing `{feature}` with the return value of `utilDisplayLabel`:

https://github.com/openstreetmap/iD/blob/43784e2eff766577b0bb39d5f2010c0d8f968903/data/core.yaml#L1856-L1862

If you agree with my suggestion above:

> “burger king” looks like a Burger Fast Food with nonstandard tags

then the string needs two placeholders, one for the display label and the other for the preset name:

* The first placeholder would be similar to what’s there now, but `verbose` should not be `true` when calling `utilDisplayLabel`. The normal behavior of displaying just the display name or the preset name is good enough.
* The second placeholder would be the preset name.

There are other messages that currently rely on this `verbose` flag; they’d need to be updated similarly. This could get a bit complicated. If you prefer a more targeted fix, maybe something like this would suffice:

> “burger king” (Burger Fast Food) looks like a common feature with nonstandard tags

Then the line you changed is the only code you need to consider.

PR #9603, opened 2023-04-28
Support addr:place tag in Address field

### LINKED ISSUE #2898: Support `addr:place` in the Address preset

tyrasd:
This adds support for the `addr:place` tag in the Address field (closes #2898).

[image](https://id.raifer.tech/addr_place/#background=Bing&disable_features=boundaries&id=w151682818&map=20.00/49.48880/8.46379)

A [demo instance](https://id.raifer.tech/addr_place) is available.

This changes the "Street" subfield of the Address field to a combined "Street / Place" input element which is little bit more clever:
* If an address has no `addr:street`, but does have an `addr:place`, it is displayed
* The dropdown offers not only nearby street names, but also nearby names of places[^1], which are made distinguishable by an icon
* When selecting an option from the dropdown, the correct tag is used `addr:street` for streets, `addr:place` for places (and in case a different tag was previously present, it is removed automatically)[^2]
* When manually entering a string into the box, the correct tag will be used if the entered name matches a nearby street or place from the dropdown.
* When manually entering a string which does not match any nearby street or place names, the `addr:street` tag is used as a fallback

[^1]: Map features with a `place=*` tag or `boundary=administrative` objects with `admin_level>8`
[^2]: Note that this implies that if an address had (for whatever reason — most probably a mapping mistake) both `addr:street` _and_ `addr:place`, selecting any option from the dropdown would remove the other tag from the type which one has selected (e.g. when selecting a street name, the `addr:place` would be removed).

Not all countries have this "enabled", as in some countries the use of place names in addresses is non-existent in OSM (e.g. [Latvia](http://taginfo.geofabrik.de/europe/latvia/keys/addr%3Aplace#map)), or use a special "non-standard" addressing scheme which allows for both `addr:street` and `addr:place` to be set at the same time (e.g. Czech Republic).

The following countries I'm currently unsure about, and would really appreciate some feedback by the local community how they would like to be handled in iD:
* **Taiwan** has currently (iD v2.25.2) both `addr:street` and `addr:place` enabled (see #3261), but I don't know whether this was just done as a hacky workaround to allow map either `addr:street` _or_ `adr:place` (but never both), or if it is actually possible for an address in Taiwan to have both tags set at the same time. I assumed that the latter is the case, as before #3261 was merged only 63 out of ~120k addresses had both tags in the country (for comparison: today the proportion is still pretty small, but the absolute number has increased significantly to 2.1k out of 1.5M). It would be nice if someone from the local community could confirm or deny this. //cc @david082321 
* In the **US** and **Canada** `addr:place` seems to be very rarely used, but one pattern I noticed is for addresses on sites which are only accessible by boat (example [1](https://www.openstreetmap.org/node/7387091670), [2](https://www.openstreetmap.org/way/1104205729)). In this PR I have currently not enabled the `addr:place` tag for these countries (my reasoning for this was that offering place names in the dropdown list of the "street field" could be confusing if it is such a rare thing to see in an actual address). However, this could be enabled easily. It would be nice to hear from the community what approach would be preferred in the US and Canada.
* The following countries also currently only the the "old" `addr:street`-only field, because they seem to not use `addr:place` in practice (and for these countries I had to make a decision because they already had localized address formats defined in iD): **Australia**, Finland, Iceland, Liechtenstein, the Netherlands, Norway, Monaco, Latvia, Peru, Bolivia, **China**, Ukraine, **India** (highlighted are the countries which I am a bit unsure about, similar to the US and Canada above, and would especially appreciate some feedback about). If further countries should be included in this list, it would be good to know what the local address format should look like (i.e. in which order the different parts of an address should be arranged into a typical list of "address rows").
* The **Czech Republic** and **Slovakia** seem to be [quite](https://github.com/openstreetmap/id-tagging-schema/issues/353#issuecomment-1290117116) [complicated](https://wiki.openstreetmap.org/wiki/Cs:%C4%8Cesko/Syst%C3%A9m_adres) in their own ways. For now, I disabled the feature in these two countries (i.e. left it untouched to the status quo) because at least in Czechia it is possible that both tags must be used at the same time. To not digress from this PR's main point: Let's discuss how the quite complicated local address schema in these countries can be better represented in iD in a separate issue.


1ec5:
> In the US and Canada addr:place seems to be very rarely used, but one pattern I noticed is for addresses on sites which are only accessible by boat (example [1](https://www.openstreetmap.org/node/7387091670), [2](https://www.openstreetmap.org/way/1104205729)). In this PR I have currently not enabled the addr:place tag for these countries (my reasoning for this was that offering place names in the dropdown list of the "street field" could be confusing if it is such a rare thing to see in an actual address). However, this could be enabled easily. It would be nice to hear from the community what approach would be preferred in the US and Canada.

@blackboxlogic [asked about using `addr:place`](https://osmus.slack.com/archives/C2VJAJCS0/p1580597986229300) before using it in that import, specifically for addresses on islands that lack streets. In my opinion, it’s quite an edge case, and I don’t personally think it’s necessary to facilitate additional usage of that tag in the U.S. Just as `addr:city` doesn’t necessarily name a city (and doesn’t even necessarily name a place), I don’t think it’s strictly necessary for `addr:street` to be a street, as long as it’s whatever would go in the “street” slot of the standard address format. After all, many addresses on college campuses have the name of a walkway where a street name would normally be. There wasn’t very much discussion about it at the time, so I’d pose the question again in Slack or the forum before enabling `addr:place` for the U.S.

The U.S. has some alternative address formats that would probably require a more significant rearchitecture of this field. For example, some rural addresses use [rural delivery route numbers](https://pe.usps.com/text/pub28/28c2_021.htm), which have no established tagging, while others use P.O. boxes (`addr:postbox`).

[Puerto Rico](https://pe.usps.com/text/pub28/28c2_041.htm) has a unique series of standard address formats. There’s [some scattered usage](http://taginfo.geofabrik.de/north-america/us/puerto-rico/keys/addr%3Aplace) of `addr:place` there, but `addr:street` is overwhelmingly common.

css/80_app.css
```diff
@@ -2100,6 +2100,22 @@ input.date-selector {
...
...cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgeD0iMCIgeT0iMCIgd2lkdGg9IjIwIiBoZWlnaHQ9IjIwIiB2aWV3Qm94PSIwIDAgMjAgMjAiPg0KICAgIDxwYXRoIGQ9Ik0xNC41NTgsMyBDMTMuMjEsMyAxMi4xMTYsNC4wOTMgMTIuMTE2LDUuNDQyIEMxMi4xMTYsNS42ODggMTIuMTYxLDUuOTIgMTIuMjI4LDYuMTQ0IEw2LjE0NCwxMi4yMjggQzUuOTIsMTIuMTYxIDUuNjg4LDEyLjExNiA1LjQ0MiwxMi4xMTYgQzQuMDkzLDEyLjExNiAzLDEzLjIxIDMsMTQuNTU4IEMzLDE1LjkwNyA0LjA5MywxNyA1LjQ0MiwxNyBDNi43OSwxNyA3Ljg4NCwxNS45MDcgNy44ODQsMTQuNTU4IEM3Ljg4NCwxNC4zMTIgNy44MzksMTQuMDggNy43NzIsMTMuODU2IEwxMy44NTYsNy43NzIgQzE0LjA4LDcuODM5IDE0LjMxMiw3Ljg4NCAxNC41NTgsNy44ODQgQzE1LjkwNyw3Ljg4NCAxNyw2Ljc5IDE3LDUuNDQyIEMxNyw0LjA5MyAxNS45MDcsMyAxNC41NTgsMyB6IE0xNC41NTgsNC40NjUgQzE1LjA5OCw0LjQ2NSAxNS41MzUsNC45MDIgMTUuNTM1LDUuNDQyIEMxNS41MzUsNS45ODEgMTUuMDk4LDYuNDE5IDE0LjU1OCw2LjQxOSBDMTQuMDE5LDYuNDE5IDEzLjU4MSw1Ljk4MSAxMy41ODEsNS40NDIgQzEzLjU4MSw0LjkwMiAxNC4wMTksNC40NjUgMTQuNTU4LDQuNDY1IHogTTUuNDQyLDEzLjU4MSBDNS45ODEsMTMuNTgxIDYuNDE5LDE0LjAxOSA2LjQxOSwxNC41NTggQzYuNDE5LDE1LjA5OCA1Ljk4MSwxNS41MzUgNS40NDIsMTUuNTM1IEM0LjkwMiwxNS41MzUgNC40NjUsMTUuMDk4IDQuNDY1LDE0LjU1OCBDNC40NjUsMTQuMDE5IDQuOTAyLDEzLjU4MSA1LjQ0MiwxMy41ODEgeiIgZmlsbD0iY3VycmVudENvbG9yIi8+DQo8L3N2Zz4=);
```
1ec5: Is this the generic way icon? I’m not sure if it’ll be recognizable enough as the icon for a street as opposed to a place.

modules/ui/fields/address.js
```diff
@@ -36,91 +36,94 @@ export function uiFieldAddress(field, context) {
...
-                return a.dist - b.dist;
-            });
+        return getNear(isAddressable, 'street', 200);
+    }
 
-        return utilArrayUniqBy(cities, 'value');
+    function getNearPlaces() {
+        function isAddressable(d) {
+            return d.tags.name && d.tags.place ||
+                d.tags.boundary === 'administrative' && d.tags.admin_level >= 8;
```
1ec5: The order of precedence here is pretty confusing without parentheses.

1ec5: For the countries that normally use `addr:place`, is it likely to be the kind of place that has an administrative boundary relation, or something less formal that’s only mapped as a point feature? Could the criteria vary by country?

css/80_app.css
tyrasd: Yes, it is (see screenshot above). Maybe we could just leave the "regular" street-name suggestions without an icon, as these are a) usually clearly identifiable as streets from their name, b) in unclear situations could still be disambiguated by their hover text, and c) represent the default case which mappers usually expect for this field. :thinking:

modules/ui/fields/address.js
tyrasd: good catch! as written `(d.tags.name && d.tags.place) || (d.tags.boundary === 'administrative' && d.tags.admin_level >= 8)` the logic is even not working as intended! it should be `d.tags.name && (d.tags.place || (d.tags.boundary === 'administrative' && d.tags.admin_level >= 8))` :bow:

tyrasd: From my experience, _small places_ (typically without a polygon, just nodes with `place=village`, `place=hamlet` or similar) are more common, but there are also cases where they reference an admin boundary (example: https://www.openstreetmap.org/node/296611194 -> https://www.openstreetmap.org/relation/436341). Perhaps referencing an `admin_level=8` boundary would be a bit too "big" of an object to reference, though. And offering these suggestions could in worst case potentially even lead to users mistakenly choosing this field instead of the intended `addr:city`, which would not be great. I'm changing it to `admin_level>8`.

PR #9633, opened 2023-05-13
fix highway=busway not supported in the turn restriction editor

k-yle:
Added [`busway`](https://wiki.osm.org/Tag:highway=busway) as a routable highway type. This fixes an issue where busways would not appear in turn restriction editor, and presumably other places that use these constants.

### DIFF
--- a/modules/ui/preset_icon.js
+++ b/modules/ui/preset_icon.js
@@ -369,6 +369,7 @@ export function uiPresetIcon() {
     subway: ['railway/subway', 'railway/subway', 'railway/subway'],
     train: ['railway/rail', 'railway/rail', 'railway/rail'],
     tram: ['railway/tram', 'railway/tram', 'railway/tram'],
+    railway: ['railway/rail', 'railway/rail', 'railway/rail'],
     waterway: ['waterway/stream', 'waterway/stream', 'waterway/stream']
   };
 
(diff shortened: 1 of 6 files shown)


tyrasd (approved):

modules/ui/preset_icon.js
```diff
@@ -369,6 +369,7 @@ export function uiPresetIcon() {
     subway: ['railway/subway', 'railway/subway', 'railway/subway'],
     train: ['railway/rail', 'railway/rail', 'railway/rail'],
     tram: ['railway/tram', 'railway/tram', 'railway/tram'],
+    railway: ['railway/rail', 'railway/rail', 'railway/rail'],
```
matkoniecz, line 372: wait, why there is `railway` added for busway?

Is it fixing mildly related problem?

Also, looking at this and https://github.com/openstreetmap/iD/blob/develop/modules/ui/preset_icon.js#L348 - what is the meaning of key and value list here?

k-yle: you're right, this is completely unrelated to the busway PR. it should have been added together with https://github.com/openstreetmap/iD-tagging-schema/issues/483

---

the purpose of this constant is to define the styles for the pretty icons used by Route relation presets:

image

e.g. the icon for road routes is a combination of the styles used for `highway=unclassified`, `=secondary`, and `=primary`.


therefore every new preset for `route=*` needs to be hardcoded in this file.

matkoniecz: opened https://github.com/openstreetmap/iD/pull/12833 in attempt to make more clear what is happening here

PR #9634, opened 2023-05-13
remove `noexit=yes` and `fixme=continue` when you continue a line

k-yle:
If you right click a node and choose "Continue" and that node has `fixme=continue`, then `fixme=continue` will be removed.

Relates to https://github.com/openstreetmap/id-tagging-schema/issues/558


tyrasd (approved):
Looks good to me: A `noexit=yes` in the middle of an extended street does not contain any information worth keeping.

PR #9635, opened 2023-05-13
allow broken (unclosed) areas to be continued

k-yle:
The continue operation cannot be used on broken areas at the moment. This is a bit frustrating, because it makes it hard to repair these areas.

The only solution right now is to delete all tags so that iD thinks it's a line, and then allows you to use the 'continue' operation.

This PR makes the operation also support unclosed area:

Screenshot 2023-05-13 141141


tyrasd (changes requested):
It's a good usability improvement. :+1: Thanks.

Let's just add a changelog entry for this.

tyrasd (approved):

PR #9636, opened 2023-05-13
add a changeset tag if a merge conflict occurred

k-yle:
If a merge conflict occured, it's quite possible that iD wasn't able to perfectly resolve the error.

Adding a tag to the changeset allows reviewers to pay more attention to edits that encountered conflicts.

Conflicts are detected after the changeset is created, so another API call is required to update the changeset tags.

Currently the tag is [`merge_conflict=yes`](https://wiki.osm.org/Tag:merge_conflict=yes). I'll document it in the wiki if this PR is accepted.


tordans:
That tag/value suggest to me that there still is a merge conflict. How about something with the term "resolved" or "auto_resolved" in key or value?

k-yle:
good point, what about `resolved_merge_conflict=yes`?

tordans:
How about …

```
merge_conflict_resolved=automatically
merge_conflict_resolved=manually
merge_conflict_resolved=automatically;manually
```

At least that would open the tagging to be used in those other cases, when things need to be done manually or partially manual. AFAIK only JOSM would be a software for this for now; but that might change.

What other software does auto merge conflict resolution? GoMap does not. Does StreetComplete?

---

Btw, it would be awesome to have this resolution code extracted to be used in other editors, see https://github.com/rapideditor/rapid-sdk/issues/166 for more.

k-yle:
Changed to `merge_conflict_resolved`=(`automatically` or `manually`). Another option could be `merge_conflict_resolution`=(`automatic` or `manual`)

Tested [here](https://api06.dev.openstreetmap.org/changeset/403889) and [here](https://api06.dev.openstreetmap.org/changeset/403887)

tyrasd (approved):

PR #9637, opened 2023-05-13
fix tag values with spaces causing css bugs

### LINKED ISSUE #9448: Fire extinguisher icon is rendered with `colour` unlike other POIs
https://www.openstreetmap.org/node/10089768614/history
is shown in red. noteably, it has the tag `colour=red`.

- When the colour is changed/removed, nothing changes (I have not saved these changes, so can't check the aftermath)
- Other fire extinguisher **without `colour`** renders fine. https://www.openstreetmap.org/node/9915072688/history
- A newly added fire extinguisher, also with colour=red, will be rendered in black. (I have not saved said non-existent fire extinguisher, so same as above.)

This rendering seems to be inconsistent with any other POIs and makes it appear as if there is an issue with this object.

image

k-yle:
Closes #9448

Some tags (like `location`) allow special CSS classes to be applied, but if the tag values has a space, then each word is added as a seperate css class. Some word like `red` or `selected` can cause expected behaviour since they're global class names. 

This PR fixes the bug by preventing tag-based class names from being added if they contain a space (which is the delimeter for css class names)


tyrasd (approved):

tyrasd:
Thx @k-yle

PR #9638, opened 2023-05-13
fix wikidata field broken for some languages

k-yle:
today is an amazing day.... for years I thought the `wikidata` field was a dumb readonly field, and I've always used the raw tag editor to set this key. 

I only realized today that it's supposed to have dropdown suggestions.... It's never worked for me, because my language is set to `en-NZ` which the wikidata API rejects, causing the field to delete everything you type into it

I'm not sure if simply stripping out the `-XX` country code suffix is the right solution, but this is an incredible QOL improvement

image

(although the suggestions are a bit weird)

### DIFF
--- a/modules/services/wikidata.js
+++ b/modules/services/wikidata.js
@@ -32,9 +32,11 @@ export default {
             search: query,
             type: 'item',
             // the language to search
-            language: lang,
+            // do not include the country-code suffix
+            language: lang.split('-')[0],
             // the language for the label and description in the result
-            uselang: lang,
+            // do not include the country-code suffix
+            uselang: lang.split('-')[0],
             limit: 10,
             origin: '*'
         });


modules/services/wikidata.js
```diff
@@ -32,9 +32,11 @@ export default {
             search: query,
             type: 'item',
             // the language to search
-            language: lang,
+            // do not include the country-code suffix
+            language: lang.split('-')[0],
```
1ec5, line 36: > I'm not sure if simply stripping out the -XX country code suffix is the right solution, but this is an incredible QOL improvement

Unfortunately, this would regress important behavior in several languages, such as Serbian and Chinese, where the country or script would be quite significant. Even in English, Wikibase supports `en-ca`, `en-gb`, and `en-us` but not `en-nz` for some reason. [This API request](https://www.wikidata.org/wiki/Special:ApiSandbox#action=query&format=json&meta=wbcontentlanguages&formatversion=2) returns the list of languages supported for labels in Wikibase.

Considering that NZ English has some terms that differ from both British and American English, it may make sense to request it as a new Wikibase language on Phabricator, where a contributor recently [worked out how to add new languages](https://phabricator.wikimedia.org/T335857#8854666).

Actually, MediaWiki requires the language codes to be fully lowercase, not with uppercase regions as is the norm. iD has a workaround for American English that appears to be unnecessary, but I guess you could use it as a simple workaround for NZ English:

https://github.com/openstreetmap/iD/blob/bd98ff904bb9498ae60fd32548cd2bbcf244bb53/modules/services/wikidata.js#L90-L94

modules/services/wikidata.js
```diff
@@ -32,9 +32,11 @@ export default {
...
             // the language for the label and description in the result
-            uselang: lang,
+            // do not include the country-code suffix
+            uselang: lang.split('-')[0],
```
1ec5, line 39: [This API query](https://www.wikidata.org/wiki/Special:ApiSandbox#action=query&format=json&meta=siteinfo&formatversion=2&siprop=languages) returns the list of interface languages supported by MediaWiki. However, MediaWiki is more lenient; you can use `en-NZ` as an interface language to get the expected results in English. (Using a more obscure interface language tends to prevent a site’s customized messages from coming through, but this is irrelevant for item queries.)

modules/services/wikidata.js
k-yle: thanks, that's really helpful!

I had a look thru [all the languages that iD supports](https://github.com/openstreetmap/iD/tree/develop/dist/locales), and of those, this issue seems to affect `en-NZ`, `en-AU`, `en-IE`, `en-IN`, `nl-NL`, `he-IL`, `fa-IR`, `fr-FR`, and `ar-AA`

I'm a bit confused though, I see [this request to support `en-IN`](https://phabricator.wikimedia.org/T212313) and [this one for `en-AU`](https://phabricator.wikimedia.org/T286862) appear to be "resolved", but neither locale is supported by the API...[^1]

[^1]: I've been testing this by opening [this URL](https://www.wikidata.org/w/api.php?action=wbsearchentities&format=json&formatversion=2&language=en-au&limit=10&origin=*&search=as&type=item&uselang=en-nz) in a new tab and tweaking `language=*`

1ec5: > I'm a bit confused though, I see [this request to support `en-IN`](https://phabricator.wikimedia.org/T212313) and [this one for `en-AU`](https://phabricator.wikimedia.org/T286862) appear to be "resolved", but neither locale is supported by the API...[1](#user-content-fn-1-cabe53c9cc00738d9d5f62f6975af1f0)

These were requests for Wikibase to accept new languages as the language of the value of a statement for which the property’s data type is “monolingual text”. For example, you can now add [a statement to say this item’s name](https://www.wikidata.org/wiki/Q1211349#P2561) is something different in Indian English or that [this item’s title](https://www.wikidata.org/wiki/Q3509128#P1476) is something different in Australian English, but this doesn’t affect either item’s label in any language.

Yes, MediaWiki/Wikibase have multiple differing lists of all the languages for hysterical reasons.

tyrasd:
I solved this slightly differently in 862ca6522a08d194e3fbb30cfd7424290eaf779f by just re-requesting the API with the simplified language code if the returned error looks like the language was the issue. This gracefully falls back to the intended behavior once the wikidata API adds supports for `en-NZ`, etc.

1ec5:
Possibly relevant MediaWiki issue:

https://phabricator.wikimedia.org/T156220

k-yle:
hey @tyrasd, your fix in 862ca6522a08d194e3fbb30cfd7424290eaf779f doesn't seem to work for me, I think it might be because there's no `return` statement after [this line](https://github.com/openstreetmap/iD/commit/862ca6522a08d194e3fbb30cfd7424290eaf779f#diff-4792335060bfa3c6e570952103d09aee92c54b5d7dd36d6b8adc31a95fa07b3aR49), so the code is calling `callback()` immediately instead of retrying the request

tyrasd:
Good catch! It was that in combination with the `this` parameter not being available in the _retry_ code path. That's now fixed in ea054a00a. Can you please double-check if it works?

k-yle:
thanks! it works perfectly now

PR #9650, opened 2023-05-25
Offer to connect sidewalk to service road without tagging crossing

### LINKED ISSUE #8463: Do not set highway=crossing on node when connecting service=driveway

1ec5:
When a driveway or other service road crosses a sidewalk, offer to either connect them without any tags or connect them with `highway=crossing`. Previously, there was only one option that implied nothing about tags but added `highway=crossing` anyways:

Driveway crosses Sidewalk Service Road crosses Sidewalk

When a service road crosses an ordinary footpath, or when a larger road crosses a sidewalk, continue to offer just the option to connect them with `highway=crossing`, but with the same icon as the crossing presets. Similarly, when a stream crosses a sidewalk, show the same icon as the Ford preset:

Foot Path crosses Service Road Residential Road crosses Sidewalk Sidewalk crosses Stream

Fixes #8463.

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -123,9 +123,8 @@ export function validationCrossingWays(context) {
         motorway: true, motorway_link: true, trunk: true, trunk_link: true,
         primary: true, primary_link: true, secondary: true, secondary_link: true
     };
-    var nonCrossingHighways = { track: true };
 
-    function tagsForConnectionNodeIfAllowed(entity1, entity2, graph) {
+    function tagsForConnectionNodeIfAllowed(entity1, entity2, graph, lessLikelyTags) {
         var featureType1 = getFeatureType(entity1, graph);
         var featureType2 = getFeatureType(entity2, graph);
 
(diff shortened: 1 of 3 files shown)


modules/validations/crossing_ways.js
```diff
@@ -435,6 +441,10 @@ export function validationCrossingWays(context) {
 
                 if (connectionTags) {
                     fixes.push(makeConnectWaysFix(this.data.connectionTags));
+                    let lessLikelyConnectionTags = tagsForConnectionNodeIfAllowed(entities[0], entities[1], graph, true);
+                    if (lessLikelyConnectionTags && JSON.stringify(connectionTags) !== JSON.stringify(lessLikelyConnectionTags)) {
```
1ec5, line 445: When a footway crosses a major road, it would be possible to offer a `crossing=no` connection as a less likely option. However, `crossing=no` doesn’t have a preset yet (openstreetmap/id-tagging-schema#548) and looks just like `highway=crossing` on the map, which could be confusing.

tyrasd: >  offer a `crossing=no` connection

I agree that the tagging-schema would need to support this first in order to make sense. Let's keep that option in mind for a later stage.

tyrasd (approved):
Looks good to me. Thanks :+1:

PR #9657, opened 2023-05-26
Continue drawing line or area at any point

### LINKED ISSUE #2272: Allow to continue drawing of area
(Sorry if there is already a thread for this, but I just couldn't find one.)

Currently there is no way how to continue drawing of a finished area. After selecting node that belongs to area, the radial menu option to continue drawing is disabled saying "No line can be continued here". This is pretty annoying when you accidentally click on a node while trying to draw a large area.

### LINKED ISSUE #10473: Refine Way tool like JOSM
### Description

In JOSM it's called the "Improve Way Accuracy Mode". This tool is very useful for realigning roads etc, it'd be great to have it in iD.

### Screenshots

image

1ec5:
The Continue operation now allows you to draw freely starting from any point along the line, not only its endpoints. It works on closed lines and areas, too. This makes the operation useful for more tasks, for example:

* Refining a river with more detailed meanders
* Updating a building with a newly constructed wing
* Continuing to draw an area after you accidentally exit drawing mode by clicking twice

In recognition of the Continue operation’s expanded abilities, it has been renamed to Draw.

https://github.com/openstreetmap/iD/assets/1231218/dffc6d46-a60d-4f09-97b8-d444a56f113c

Fixes #2272.

### DIFF
--- a/modules/modes/draw_line.js
+++ b/modules/modes/draw_line.js
@@ -2,7 +2,7 @@ import { t } from '../core/localizer';
 import { behaviorDrawWay } from '../behavior/draw_way';
 
 
-export function modeDrawLine(context, wayID, startGraph, button, affix, continuing) {
+export function modeDrawLine(context, wayID, startGraph, button, position, continuing) {
     var mode = {
         button: button,
         id: 'draw-line'
@@ -15,13 +15,21 @@ export function modeDrawLine(context, wayID, startGraph, button, affix, continui
                 .label(t.append('self_intersection.error.lines'))();
         });
 
+    // If the position is a node index rather than an affix string, the node index will be dynamic based on an offset from the end.
+    var offsetFromEndNode;
+    if (typeof position === 'number') {
+        offsetFromEndNode = context.entity(wayID).nodes.length - position - 1;
+    }
+
     mode.wayID = wayID;
 
     mode.isContinuing = continuing;
 
     mode.enter = function() {
-        behavior
-            .nodeIndex(affix === 'prefix' ? 0 : undefined);
+        behavior.offsetFromEndNode(offsetFromEndNode);
+        if (position === 'prefix') {
+            behavior.nodeIndex(0);
+        }
 
         context.install(behavior);
     };
--- a/modules/operations/continue.js
+++ b/modules/operations/continue.js
@@ -1,5 +1,6 @@
 import { t } from '../core/localizer';
 import { modeDrawLine } from '../modes/draw_line';
+import { modeDrawArea } from '../modes/draw_area';
 import { behaviorOperation } from '../behavior/operation';
 import { actionDeleteNode } from '../actions/delete_node';
 import { utilArrayGroupBy } from '../util';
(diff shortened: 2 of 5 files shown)


modules/operations/continue.js
```diff
@@ -16,20 +18,27 @@ export function operationContinue(context, selectedIDs) {
...
+    var _candidateGeometry = _candidate && _candidate.geometry(context.graph());
 
 
     var operation = function() {
-        var candidate = _candidates[0];
+        var affix = _candidate.affix(_vertex.id);
+        // Unless an endpoint is selected, delete the selected node in favor of whatever we're going to draw.
+        // This avoids a situation where it's possible to draw along all but one of the edges.
+        if (!affix) {
+            context.perform(actionDeleteNode(_vertex.id));
```
1ec5, line 36: If you select a vertex other than an endpoint, the operation replaces it with the newly drawn vertices. Otherwise, if we were to persist the selected vertex and only continue drawing along either adjacent edge, then either the first or the last edge of the way would become impossible to continue drawing along.

Ideally, this operation would detach the node from the way and only delete the node if it has no tags and no parents.

modules/modes/draw_line.js
```diff
@@ -15,13 +15,21 @@ export function modeDrawLine(context, wayID, startGraph, button, affix, continui
                 .label(t.append('self_intersection.error.lines'))();
         });
 
+    // If the position is a node index rather than an affix string, the node index will be dynamic based on an offset from the end.
+    var offsetFromEndNode;
+    if (typeof position === 'number') {
+        offsetFromEndNode = context.entity(wayID).nodes.length - position - 1;
```
1ec5, line 21: When you select a vertex other than an endpoint, you draw forward according to the way’s direction. This is simpler and more intuitive than devising a gesture to decide which direction to draw. However, it can get in the way when continuing to draw an area, since the Reverse operation is disabled for areas.

tyrasd:
Hey. Very cool! However… I had been "secretly" working on a very similar feature recently. My version has the additional functionality to allow to "fly" along a line (or area outline) without having to re-select the start-vertex of the segment to be improved (similar to how it was proposed in #760):

way-amend-tool

IMHO that functionality is quite important because in a typical use case (e.g. when improving the bends of a waterway or realigning of a path), one does typically want to work on more than just a single segment of a way.

We talked about this the last monthly community meetup, but I definitely should have announced that I had started working on it in the issue #2272 to avoid duplicate work on it. Sorry for that!

1ec5:
Oh cool, I had totally forgotten about #760. If I’m interpreting your screen recording correctly, you can enter drawing mode and click on an existing adjacent node to move it during drawing mode, correct? That satisfies the use cases I had in mind and gets closer to the request for a gesture in https://github.com/openstreetmap/iD/issues/2272#issuecomment-366485669. I’m curious about how the user can exit edit mode using the mouse if it has this additional behavior. Presumably the user won’t always want to redraw the entire line to the very end.

I’ll leave this PR open for now to serve as a reference point, but I look forward to your better design!

tuxayo:
https://github.com/openstreetmap/iD/pull/9657#issuecomment-1564188119
> I had been "secretly" working on a very similar feature recently. [...]


@tyrasd Is there a draft MR for your attempt at this?

PR #9664, opened 2023-05-26
Add mapilio data

sezerbozbiyik:
Integrates the Mapilio imagery in to the iD editor.
You're welcome.

-Mapilio


tyrasd:
Hi. Thank you for the pull request. :tada: I'm looking forward to give this a proper review in the near future, but could you in the meantime take a look at the failing test?

PS: I had seen some discussions in the OSM forum about mapilio. One [commentor was concerned](https://community.openstreetmap.org/t/mapilios-support-for-openstreetmap/99021/9) about the potential problem of the CC-BY-SA license being incompatible with OSM, but I assume the additional terms of use on https://mapilio.com/openstreetmap would allow the use as proposed in this PR for OSM mapping, right? Maybe you guys could jump in to clarify this (any maybe some other) question on the OSM forum. :smiley:

mnalis:
@channel-s while it is nice that you have separate (human-readable) page for OSM, would you please update your [Terms & Conditions](https://mapilio.com/terms) to actually include those permissions before proceeding further? That one is the legally binding text, and as it is, it currently still seems to strictly prohibit using Mapilio data to contribute to OSM (other pages on mapilio.com website notwithstanding).

See thread above: OSM is **not** compatible with CC-BY-SA, so you'll have to give special exception for OSM (or dual-license / re-license with one of the **[compatible licenses](https://wiki.openstreetmap.org/wiki/Import/ODbL_Compatibility)** like PDDL or CC0) if you want people to be able to use it as a source in OSM.

sezerbozbiyik:
Thank you for your comments. First of all, I made a commit that resolved the testing error. In the near future, we will provide a detailed explanation of the terms and conditions.

ozcan-durak:
Hello, we would like to inform you that we have recently revised our terms and conditions regarding the usage of Openstreetmap.

mnalis:
@ozcan-durak Firstly, it's great that you try to support OpenStreetMap! However, those terms are still hugely problematic, and still prohibit using it in OpenStreetMap.

Particularly, in your `Right To Use For OpenStreetMap` section you say:

> We support OpenStreetMap and allow you and others to use the Mapilio Services, and all the data and content in them, to create, edit and add content to OpenStreetMap

which is great! But then you follow that with:
> You must follow the terms of each entity when you contribute to OpenStreetMap.

which in effect nullifies that clause (as OSM terms are incompatible with rest of Mapilio terms, as noted earlier, so it is impossible to follow both at the same time). IOW, you can't have your cake and eat it too.

You should replace that second sentence with something like this instead:
> This clause always overrides any other restrictions placed, as long as You follow OpenStreetMap terms.

But I am not a lawyer (especially not your lawyer),  and this is not a legal advice. You should probably get one if you want this done correctly without legal ambiguity.

ozcan-durak:
@mnalis 

I appreciate your response. Our lawyer has prepared the terms and conditions, and we have thoroughly reviewed the licenses for Bing Streetside, Kartaview, and Mapillary. Rest assured, our license does not contradict the terms of OSM.

mnalis:
@ozcan-durak Thank you for your assurances! Can your lawyers then explicitly confirm that in particular following clauses are _null and void_ when user uses data from Mapilio in conjunction with OpenStreetMap:

> Use any Mapilio Services with any products, systems, or applications for or in connection with real-time navigation or route guidance, such as to turn-by-turn route guidance that is synchronized to the position of a sensor-enabled device, or any systems or functions for automatic or autonomous control of vehicle or device behavior;

> Copy, reproduce, distribute, publicly perform or publicly display all or portions of the Mapilio Services, except as expressly permitted by us or our licensors;

E.g. can the user use data from Mapilio to create/update OpenStreetMap data, and then use the resulting OSM data for real-time navigation or autonomous control of vehicle?

tyrasd:
@mnalis I appreciate your thoroughness on this topic. However, I think this github issue is not the right place for such intricate legal discussions. This ticket should be focused around the technical implementation of the service into iD.

@ozcan-durak Could you please contact me privately at martin@raifer.tech with the preferred contact from your side to talk about legal details? I would like to connect you with the corresponding person(s) at the OSM Foundation ([licensing working group](https://wiki.osmfoundation.org/wiki/Licensing_Working_Group) and/or the board) to double-check the compatibility of the Mapilio terms with OSM.

noenandre:
#9509 depend on Pannellum too, to view panoramic images. Is it possible to improve [this module](https://github.com/openstreetmap/iD/pull/9509/files#diff-c9df0c6b7a658cccbe333627e3364350523834b2f4b1fe34362cf6d36a0693dc) instead of duplicating the code?

sezerbozbiyik:
@noenandre I dont have much time to this, after my pr merged i will check what can i do

modules/services/mapilio.js
```diff
@@ -0,0 +1,564 @@
...
+
+const apiUrl = 'https://end.mapilio.com';
+const imageBaseUrl = 'https://cdn.mapilio.com/im';
+const baseTileUrl = 'https://geo.mapilio.com/geoserver/gwc/service/wmts?REQUEST=GetTile&SERVICE=WMTS&VERSION=1.0.0&LAYER=mapilio:';
+const pointLayer = 'map_points';
+const lineLayer = 'map_roads_line';
+const tileStyle = '&STYLE=&TILEMATRIX=EPSG:900913:{z}&TILEMATRIXSET=EPSG:900913&FORMAT=application/vnd.mapbox-vector-tile&TILECOL={x}&TILEROW={y}';
+
+const minZoom = 14;
+const dispatch = d3_dispatch('change', 'loadedImages', 'loadedLine');
```
tyrasd: probable typo and `change` event is not used?!

```suggestion
const dispatch = d3_dispatch('loadedImages', 'loadedLines');
```

modules/services/mapilio.js
```diff
@@ -0,0 +1,564 @@
...
+        .then(function(response) {
+            if (!response.ok) {
+                throw new Error(response.status + ' ' + response.statusText);
+            }
+            cache.loaded[tileId] = true;
+            delete cache.inflight[tileId];
+            return response.arrayBuffer();
+        })
+        .then(function(data) {
+            if (!data) {
```
tyrasd: As `data` is always an `ArrayBuffer` object (see 3 lines above), I assume you meant something like this to detect an empty response:

```suggestion
            if (datadata.byteLength === 0) {
```

modules/services/mapilio.js
```diff
@@ -0,0 +1,564 @@
...
+
+            if (which === 'images') {
+                dispatch.call('loadedImages');
+            } else {
+                dispatch.call('loadedLines');
+            }
+        })
+        .catch(function() {
+            cache.loaded[tileId] = true;
+            delete cache.inflight[tileId];
```
tyrasd: the `cache.inflight[tileId]` is already deleted in the first `fetch(…).then` callback (line 93), so I assume it's not necessary here.

modules/services/mapilio.js
```diff
@@ -0,0 +1,564 @@
...
+            loadTileDataToCache(data, tile, which);
+
+            if (which === 'images') {
+                dispatch.call('loadedImages');
+            } else {
+                dispatch.call('loadedLines');
+            }
+        })
+        .catch(function() {
+            cache.loaded[tileId] = true;
```
tyrasd: This is a bit problematic, as it silently ignores all kinds of exceptions, not just the "No data" case , I assume you want to handle here.

I'd say that for generic errors (e.g. a non-HTTP/200 response from the server, corrupted data which cannot be parsed, or just javascript runtime errors), there should be at least a console warning output, maybe like this:

```suggestion
        .catch(function(e) {
            if (e.message === 'No Data') {
                cache.loaded[tileId] = true;
            } else {
                console.error(e);
            }
```

modules/services/mapilio.js
```diff
@@ -0,0 +1,564 @@
...
+        layer = vectorTile.layers[pointLayer];
+
+        for (i = 0; i < layer.length; i++) {
+            feature = layer.feature(i).toGeoJSON(tile.xyz[0], tile.xyz[1], tile.xyz[2]);
+            loc = feature.geometry.coordinates;
+
+            let resolutionArr = feature.properties.resolution.split('x');
+            let sourceWidth = Math.max(resolutionArr[0], resolutionArr[1]);
+            let sourceHeight = Math.min(resolutionArr[0] ,resolutionArr[1]);
+            let isPano = sourceWidth % sourceHeight === 0;
```
tyrasd, line 146: This approach to detect panoramic images does not work (anymore?): panos appear to not come with the `resolution` property set.

This worked for me:

```javascript
            let isPano = feature.properties.resolution === undefined;
            let sourceHeight, sourceWidth;
            if (!isPano) {
                let resolutionArr = feature.properties.resolution.split('x');
                sourceWidth = Math.max(resolutionArr[0], resolutionArr[1]);
                sourceHeight = Math.min(resolutionArr[0] ,resolutionArr[1]);
            }
```

modules/services/mapilio.js
```diff
@@ -0,0 +1,564 @@
...
+                .append('span')
+                .text('|');
+        }
+
+        attribution
+            .append('a')
+            .attr('class', 'image-link')
+            .attr('target', '_blank')
+            .attr('href', `https://mapilio.com/app?lat=${d.loc[1]}&lng=${d.loc[0]}&zoom=17&pId=${d.id}`)
+            .text('mapilio.com');
```
tyrasd, line 369: I would prefer to have some kind of feedback that the next image is currently loading (when clicking on the :arrow_backward:&nbsp;:arrow_forward: buttons, for example). At least something simple like just removing the old image:

```javascript
        wrap
            .selectAll('img')
            .remove();
```

tyrasd (changes requested):
Hey. Thank you very much. This is almost ready to be merged, I'd say. See some comments inline below, and the following remarks:

* In some places, the mapilio API does not seem to return a heading of the images example location: 41.0028, 28.9639 (FWIW, these look fine on the mapilio [website](https://mapilio.com/app?lat=41.0028751296185&lng=28.963961317836862&zoom=18.429677507141115&pId=245785)). It would be best if this issue could be resolved on the API side, or otherwise that these would be handled more gracefully (e.g. use pseudo heading in the direction to the previous/next images).
* The mapilio "layer" should implement the `photo` [hash-parameter](https://github.com/openstreetmap/iD/blob/develop/API.md) (see [`updateUrlImage` in kartaview.js](https://github.com/openstreetmap/iD/blob/v2.26.2/modules/services/kartaview.js#L563-L573)).
* The non-360 images are not zoomable at the moment. That functionality is however quite important for mapping: e.g. to read details like road names, street signs, …. https://github.com/tyrasd/iD/commit/fcac05745 has an implementation that works quite the same as the kartaview layer
* It would be nice if filtering by date (and optimally also username) would work also for mapilio images. As far as I can see, this information is already present in the Mapilio vector tiles, so it should in principle be possible to implement. This could also be done after this PR.

modules/services/mapilio.js
sezerbozbiyik: This issue has been resolved on the API side, and now data without resolution values is no longer being returned.

sezerbozbiyik:
Hello @tyrasd,
With the recent changes, the issue of missing heading and resolution information in the project has been corrected by the API. Additionally, the zoom and photo ID parameters have been added to the pull request. Moreover, small errors in the project have been fixed, and a new commit has been sent. Filtering of data based on date and user will be implemented in the next PR. Thank you for your assistance.

tyrasd:
Looks great, it is merged with some minor tweaks (e.g. ad1162dd1 fixes a  bug which made panning of the 360° almost impossible because the `d3_zoom` was interfering with the pannellum code).

PS: I noticed that a few of the 360° images don't load (also on your website, example: photo [1879418](https://mapilio.com/app?lat=41.08197950170589&lng=28.80543887615204&zoom=17&pId=1879418)), maybe you can check this on your system

StyXman:
What happened to the legal issues? Have they been resolved? Aren't we exposing OSM if not?

tyrasd:
Yep, those questions have all been resolved by clarifying their [terms](https://mapilio.com/terms) after consulting OSM's licensing working group.

Issue #974, opened 2013-03-11
opening hours form field


ansis:
http://wiki.openstreetmap.org/wiki/Key:opening_hours

M1dgard:
Any progress here?
In the mean time maybe a link could be provided to simple instructions (a [page like this](https://wiki.openstreetmap.org/wiki/User:M!dgard/How_to_map_opening_hours)) for the syntax. Many contributors just fill in something in natural language, which is useless to automated interpreters like [OpenLinkMap](http://openlinkmap.org) and [OsmAnd](http://osmand.net).

bagage:
Also another great alternative not mentioned yet is [YoHours](http://github.pavie.info/yohours/). 

Maybe iD could give an hint to use one of these solutions when starting writing opening hours manually? 
It would not be a hard solution to implement right?

tmcw:
> It would not be a hard solution to implement right?

Please implement it, if you perceive it to be easy.

bhousel:
> Also another great alternative not mentioned yet is YoHours.
> 
> Maybe iD could give an hint to use one of these solutions when starting writing opening hours manually? 

I just spent about 10 minutes playing around with it and it's unfortunately kind of buggy and hard to use.  I do think it's probably a good starting effort though.  If it were a bit more polished and could look nice in a lightboxed small dimension iframe, I'd consider it.

> It would not be a hard solution to implement right?

I'm vetoing what @tmcw says above, this is too buggy to use currently and I wouldnt merge it.  For now spend time improving the tool, but not on integrating it with iD.

iandees:
Also, supporting all the complex features behind the `opening_hours` tag language is nice, but the interface on YoHours is *way* too complex for iD. Limit to full (or half maybe) hours, don't expose the seasonal options, and only allow one tab to start. If someone wants to build more complex OH strings then they could switch to some advanced mode or something.

bagage:
@tmcw It was more of a question than an assumption, I would certainly not state on unknown technologies; my apologies if it looked like it. 
I just wanted to bump the issue since I found `YoHours` a bit earlier and found it valuable to map a swimming pool with seasonal opening hours which would have take me half an hour to do manually. 
I am not `YoHours` author at all, but considered it as easy enough to use in iD. It seems I was wrong :). At least it may help on what-to-do and what-not-to-do for iD widget.

tyrasd:
[...]

PR #9678, opened 2023-06-01
Add YoHours button to recurring schedule fields

1ec5:
Fields that accept opening_hours syntax are now standard text fields with the addition of a button that links out to [YoHours](https://projets.pavie.info/yohours/). The user can use this button to compose a value or visualize an existing value, or they can compose the value manually as before.

Hours

Depends on ideditor/schema-builder#101. Itty-bitty baby steps toward #974.

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -69,7 +69,9 @@ export function presetIndex() {
         // https://github.com/ideditor/schema-builder/issues/100
         const scheduleFields = ['opening_hours', 'opening_hours/covid19', 'service_times', 'collection_times'];
         scheduleFields.forEach(id => {
-          vals[3][id].type = 'schedule';
+          if (id in vals[3]) {
+            vals[3][id].type = 'schedule';
+          }
         });
 
         _this.merge({
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -66,6 +66,12 @@ export function presetIndex() {
         fileFetcher.get('preset_fields')
       ])
       .then(vals => {
+        // https://github.com/ideditor/schema-builder/issues/100
+        const scheduleFields = ['opening_hours', 'opening_hours/covid19', 'service_times', 'collection_times'];
+        scheduleFields.forEach(id => {
+          vals[3][id].type = 'schedule';
+        });
+
         _this.merge({
           categories: vals[0],
           defaults: vals[1],
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -66,14 +66,6 @@ export function presetIndex() {
         fileFetcher.get('preset_fields')
       ])
       .then(vals => {
-        // https://github.com/ideditor/schema-builder/issues/100
-        const scheduleFields = ['opening_hours', 'opening_hours/covid19', 'service_times', 'collection_times'];
-        scheduleFields.forEach(id => {
-          if (id in vals[3]) {
-            vals[3][id].type = 'schedule';
-          }
-        });
-
         _this.merge({
           categories: vals[0],
           defaults: vals[1],
--- a/modules/ui/fields/input.js
+++ b/modules/ui/fields/input.js
@@ -18,12 +18,14 @@ export {
     uiFieldText as uiFieldEmail,
     uiFieldText as uiFieldIdentifier,
     uiFieldText as uiFieldNumber,
+    uiFieldText as uiFieldSchedule,
     uiFieldText as uiFieldTel,
     uiFieldText as uiFieldUrl,
     likelyRawNumberFormat
 };
 
 const likelyRawNumberFormat = /^-?(0\.\d*|\d*\.\d{0,2}(\d{4,})?|\d{4,}\.\d{3})$/;
+const yoHoursURLFormat = 'https://projets.pavie.info/yohours/?oh={value}';
 
 export function uiFieldText(field, context) {
     var dispatch = d3_dispatch('change');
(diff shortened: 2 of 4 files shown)


modules/ui/fields/input.js
```diff
@@ -18,12 +18,14 @@ export {
...
     uiFieldText as uiFieldIdentifier,
     uiFieldText as uiFieldNumber,
+    uiFieldText as uiFieldSchedule,
     uiFieldText as uiFieldTel,
     uiFieldText as uiFieldUrl,
     likelyRawNumberFormat
 };
 
 const likelyRawNumberFormat = /^-?(0\.\d*|\d*\.\d{0,2}(\d{4,})?|\d{4,}\.\d{3})$/;
+const yoHoursURLFormat = 'https://projets.pavie.info/yohours/?oh={value}';
```
1ec5, line 28: The spellcheck action on GitHub Actions doesn’t like this URL because French words sometimes look like misspelled English words. 😉

tordans:
I like the idea of having this as a step towards better UI inside iD!

A few things I notices:

- Ideally, on https://projets.pavie.info/yohours/ the term `opening_hours=` that is used as a label of the field could also be configured by url param so the field can specify it to create a smoother transition between the two apps.
- Ideally, the page would offer a copy button next to the input, so the UI signals that the next step will be to manually copy and paste the content back to iD.
- In fact, it would be great if this flow could be explained somewhere. Maybe with a message on yohours along the line of "you are using this tool from <app>, please copy the value back to the <app> after you modified the schedule".
- Would it be possible to add some shared logic between iD and yohours to post messages back to iD? I remember solutions that where used to resize an parent iframe based on height data from inside the iframe using [postMesssage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)…

modules/ui/fields/input.js
```diff
@@ -18,12 +18,14 @@ export {
     uiFieldText as uiFieldEmail,
     uiFieldText as uiFieldIdentifier,
     uiFieldText as uiFieldNumber,
+    uiFieldText as uiFieldSchedule,
```
1ec5, line 21: Actually it would be nice to keep the dropdown button for some common values like `24/7` that YoHours doesn’t support. But the combo field doesn’t support an external link button as far as I know.

1ec5:
> * Ideally, on https://projets.pavie.info/yohours/ the term `opening_hours=` that is used as a label of the field could also be configured by url param so the field can specify it to create a smoother transition between the two apps.
> 
> * Ideally, the page would offer a copy button next to the input, so users can easily copy the value back.

I guess this would be feedback for @PanierAvide. I’m not sure if the YoHours website itself has an issue tracker, but there is one for [the jQuery widget](https://github.com/simialbi/jquery-yohours/) that’s embedded in it.

> * Would it be possible to add some shared logic between iD and yohours to post messages back to iD? I remember solutions that where used to resize an parent iframe based on height data from inside the iframe using [postMesssage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)…

There is an iframe version of YoHours intended to be embedded inside a Web application like iD: https://github.com/openstreetmap/iD/issues/974#issuecomment-134230497. However, I didn’t use it here because embedding it into a rather space-constrained part of iD would be more difficult than simply linking out, and there’s no precedent for opening modals from the inspector.

modules/presets/index.js
```diff
@@ -66,6 +66,14 @@ export function presetIndex() {
         fileFetcher.get('preset_fields')
       ])
       .then(vals => {
+        // https://github.com/ideditor/schema-builder/issues/100
+        const scheduleFields = ['opening_hours', 'opening_hours/covid19', 'service_times', 'collection_times'];
+        scheduleFields.forEach(id => {
+          if (id in vals[3]) {
+            vals[3][id].type = 'schedule';
```
k-yle: Since https://github.com/ideditor/schema-builder/pull/101 is now merged, this temporary override can be removed, and then this PR would be ready to review&merge

k-yle (changes requested):

modules/presets/index.js
1ec5: We’re still waiting on openstreetmap/id-tagging-schema#1537, right?

k-yle: in a recent community meeting I think we decided that the order should be:
- define the new syntax in [schema-builder](https://github.com/ideditor/schema-builder)
- add backwards-compatible support in [iD](https://github.com/openstreetmap/id)
- then start using the new syntax in [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema)

the order of the last two is less strict, but this approach decouples the software changes from the tagging-schema update

1ec5: > add backwards-compatible support

Would this compatibility shim live somewhere other than in the part of the codebase where iD overrides id-tagging-schema data? Should the `combo` field code special-case these keys instead?

k-yle: I think deleting this code block is all we need to do right? That would allow this PR to be merged now, although the new code could have no effect. 

Then the last step is to merge https://github.com/openstreetmap/id-tagging-schema/pull/1537, which would activate the new code from this PR

1ec5: Ah, that’s what you meant. I removed the code that injects `schedule` into the presets.

k-yle: thanks! sorry i didn't phrase that very well

k-yle (approved):
lgtm ! and I think we can remove https://github.com/openstreetmap/iD/labels/waitfor-upstream , since this technically has no more blockers

k-yle:
I think we can also merge this one now, it won't take effect until schema-builder v7 is released

PR #9709, opened 2023-06-23
warn when mapy.cz is used in a source tag

### LINKED ISSUE #9708: Ban some (copyrighted) custom background
### Description

We have this issue in czech community: some begginers are using copyrighted custom background when editing OSM. 

In the whole database, there is around [11000 edits so far](https://osmcha.org/?aoi=0c8c392a-2104-4382-a8ed-7bed8745975c). Of course members of community are educating those users and reverting the changes, but we would like to ask, if it could be somehow banned in iD itself. It is becoming a little tedious :-)

- Proposed change: when custom background hostname contains "mapy.cz", warn the user and disable the Save button.

.

Here is our thread in czech: https://openstreetmap.cz/talkcz/c3793 ([translator](https://openstreetmap-cz.translate.goog/talkcz/c3793?_x_tr_sl=cs&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp))

### Screenshots

_No response_

Jakuje:
I went through the changesets and most of them use directly `mapserver.mapy.cz` (some older used `mX.mapserver.mapy.cz` so I hope this regex should work and would not cause any false positives.

See #9708

### DIFF
--- a/modules/validations/incompatible_source.js
+++ b/modules/validations/incompatible_source.js
@@ -18,6 +18,10 @@ export function validationIncompatibleSource() {
       id: 'google',
       regex: /google/i,
       exceptRegex: /((books|drive)\.google|google\s?(books|drive|plus))|(esri\/Google_Africa_Buildings)/i
+    },
+    {
+      id: 'mapy.cz',
+      regex: /mapy\.cz/i
     }
   ];
 


modules/validations/incompatible_source.js
```diff
@@ -18,6 +18,10 @@ export function validationIncompatibleSource() {
       id: 'google',
       regex: /google/i,
       exceptRegex: /((books|drive)\.google|google\s?(books|drive|plus))|(esri\/Google_Africa_Buildings)/i
+    },
+    {
+      id: 'mapy.cz',
+      regex: /mapy.cz/i
```
zbycz: I think this `.` should be escaped like this `\.` – see above.

Otherwise it would match any character.

Jakuje: Thanks! Fixed!

tyrasd:
Adding this validation is fine, but note that it likely does not do what you might have intended with this PR: it only shows a warning when `mapy.cz` is used in a `source` tag of a map feature, (see screenshot below). It does not prevent users from using mapy.cz as a background layer.

image

tyrasd (changes requested):
One small detail is missing here: There must be a translatable string with a description of why this source is not suitable for mapping in OSM, see https://github.com/openstreetmap/iD/blob/v2.25.2/data/core.yaml#L1810-L1813 how it was done for the other existing source tag validations.

pnorman:
Given the original issue is solved, can this PR be closed?

zbycz:
Yes, thanks. This can be closed. Solved via https://github.com/openstreetmap/openstreetmap-website/pull/4093

PR #9731, opened 2023-07-06
Allow panning with PgUp/PgDown/Home/End

### LINKED ISSUE #9452: Feature request: make good use of Page Up / Page Down buttons
I love that iD has a fair amount of keyboard support:

- **\+** to zoom in, **-** to zoom out
- **arrow keys** to pan the view left, right, up, and down
- **D** to disconnect features
- etc.

But iD's keyboard support has one glaring omission.

Editing would be far more efficient if users could use 

- the **Page Up** key to move their view of the map one whole screen up
- the **Page Down** key to move their view of the map one whole screen down

Yes, these actions can be approximated by dragging the map with the mouse, or by holding down the up/down arrow key long enough.

But accomplishing it with one quick whack of **Page Up** or **Page Down** would be soo nice and efficient!

As it stands now, sometimes nothing happens when I press **Page Up** or **Page Down** – and sometimes something happens, but certainly not anything intuitive.  I'm not sure what it's trying to do when I press those keys.  Lot of room for improvement here.

k-yle:
Closes #9452

### DIFF
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -21,6 +21,11 @@
             "separator": ","
           },
           {
+            "shortcuts": ["↖", "↘", "⇞", "⇟"],
+            "text": "shortcuts.browsing.navigation.pan_more",
+            "separator": ","
+          },
+          {
             "shortcuts": ["+", "-"],
             "text": "shortcuts.browsing.navigation.zoom",
             "separator": ","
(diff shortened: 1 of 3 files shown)


tyrasd:
As it is already possible to pan the map with [alt/option+arrow-keys](https://github.com/openstreetmap/iD/issues/9452#issuecomment-1374576909), and this functionality is not really used often enough to warrant alternative keys, I think it's better to reserve the PgUp/PgDown/Home/End keys for new functionality in the future.

PR #9732, opened 2023-07-06
render diameter when a feature is selected

### LINKED ISSUE #9189: Create empty node from empty area
### Description

Allow to extract/create an empty node from an empty area in its center, to allow the user to find the center of a circle or other shape for placing ie. a turning circle. This saves the user from typing and afterwards removing something like `amenity=yes` to make it extractable.

### Screenshots

_No response_

k-yle:
If a node has a tag like `diameter` or `radius`, iD will now render a circle when you select the node.

image

- As per the wiki, spaces are allowed between the number and the unit.
- These tags are supported: `diameter`, `diameter_crown`, `hole:diameter`, `radius`, `crown_radius`,  and `seamark:anchor_berth:radius`.
- [Many different units of measurement](https://github.com/k-yle/iD/blob/0ca1f63f/modules/core/planar.js#L34-L64) are supported, including a combination of feet and inches
- The default unit is `m` for all tags except `diameter`, where the [default unit is `mm`](http://wiki.osm.org/Key:diameter) and `seamark:anchor_berth:radius` where the unit is defined by a separate tag (`seamark:anchor_berth:units`)

---

does not exactly fix #9189, but solves the problem with an alternative solution


tordans:
I really like this idea. We are using [diameter on trees](https://www.openstreetmap.org/node/4495822552) a lot in Neukölln to render a more realistic Map like https://strassenraumkarte.osm-berlin.org/?map=micromap#19/52.47378/13.43892 so that will help at lot to get a feel for this value.

modules/svg/points.js
```diff
@@ -122,6 +123,22 @@ export function svgPoints(projection, context) {
             .append('path')
             .call(markerPath, 'shadow');
 
+        // add a ring if there's a radius/diameter tag
+        enter
+            .filter((d) => getRadiusTag(d.tags))
+            .append('circle')
+            .attr('class', 'radius')
+            .attr('data-nodeId', node => node.id);
```
tyrasd: it's probably better to avoid camelcasing in html attributes, see e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/data-*

```suggestion
            .attr('data-nodeid', node => node.id);
```

or maybe even simpler:

```suggestion
            .attr('data-id', node => node.id);
```

modules/core/planar.ts
```diff
@@ -0,0 +1,97 @@
...
+    parseDistanceWithUnit(tags.diameter_crown, 'm') ||
+    parseDistanceWithUnit(tags['hole:diameter'], 'm')
+  );
+  if (diameter) return diameter / 2;
+
+  const radius = (
+    parseDistanceWithUnit(tags.radius, 'm') ||
+    parseDistanceWithUnit(tags.crown_radius, 'm') ||
+    parseDistanceWithUnit(tags['seamark:anchor_berth:radius'], tags['seamark:anchor_berth:units'] || 'm')
+  );
```
1ec5: This is a good reminder that, even if ideditor/schema-builder#15 ends up allowing a field to specify a dimension, it may sometimes need to specify a default unit too.

modules/core/planar.ts
```diff
@@ -0,0 +1,97 @@
...
+    case 'mi': return value * 1609;
+
+    case 'nautical_miles':
+    case 'nm': return value * 1852;
+
+    case 'yard':
+    case 'yards':
+    case 'yd': return value / 1.094;
+
+    case '′':
```
1ec5: The prime marks are a nice touch, but I don’t think we’ve documented them (or `foot` or `feet`) as [allowable unit symbols](https://wiki.openstreetmap.org/wiki/Map_features/Units) or even discouraged ones. Keywords such as `nautical_miles` also aren’t documented anywhere. Is the goal to be follow Postel’s law, or do you see a potential for these keywords to become standard at some point too?

modules/core/planar.ts
```diff
@@ -0,0 +1,97 @@
...
+    const inches = +imperialCombo[3];
+    return feet / 3.281 + inches / 39.37;
+  }
+
+  // the remaining code parses normal values (number + optional unit)
+
+  const unit = tagValue.match(/[^\d.]+/)?.[0].trim() || defaultUnit;
+  const value = parseFloat(tagValue);
+
+  if (Number.isNaN(value)) return undefined; // invalid value
```
1ec5: `parseFloat()` will also return infinity if `tagValue` is `Infinity`, so check `isFinite()` too.

modules/svg/points.js
```diff
@@ -122,6 +123,22 @@ export function svgPoints(projection, context) {
...
+
+        // we need to trigger a re-render when the zoom changes
+        selection
+            .selectAll('.radius')
+            .data([zoom])
+            .attr('r', (_zoom, _zero, [self]) => {
+                const node = graph.entities[self.dataset.nodeId];
+                return getRadiusInPixels(node, context.projection);
```
1ec5: Should we constrain the radius for rendering purposes? What happens in case of a mistake such as a negative or very large value?

modules/core/planar.ts
k-yle: Yeah we should probably only include documented values. I've removed `nautical_miles` ([only 30 uses](https://taginfo.osm.org/tags/seamark:distance_mark:units=nautical_miles), no docs). 

I can't tell how widespread primes are, since taginfo's search [shows 0 results](https://taginfo.openstreetmap.org/search?q=%E2%80%B2), but this is wrong ([counter example](https://osm.org/way/928692851)). Should we remove them too?

modules/svg/points.js
k-yle: Negative numbers are handled by us now. For large values, we just get a massive SVG `<circle />` that's too big to see at an editable zoom-level. Not sure if we need an upper limit?

modules/core/planar.ts
k-yle: thanks, didn't realise that ! fixed now

modules/core/planar.ts
k-yle: the proposal from https://github.com/ideditor/schema-builder/pull/198 would support this; an empty string = the default unit for that tag. 
```jsonc
        "units": {
            "kilometer-per-hour": [""], // 👈
            "mile-per-hour": ["mph"],
            "knot": ["kt", "kn"]
        }
```

css/20_map.css
```diff
@@ -185,6 +185,17 @@ g.vertex.selected .shadow {
     fill-opacity: 0.7;
 }
 
+/* circle to visualize radius/diameter */
+g:is(.point, .vertex) .radius {
+    display: none;
+}
+g:is(.point, .vertex):is(.active, .selected) .radius {
+    display: block;
```
tyrasd: This made me wonder what `display` value is the correct one to use for SVG elements. 😅 Couldn't find anything authoritative at a first glance, but Firefox' and Chrome's dev tools say that it's `inline` if not set (which is AFAIK simple the default value of `display` in CSS). OTOH, the [`display` SVG attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/display) only really defines what `none` should do. So, maybe here it makes more sense to just unset the previously set `display: none`??


```suggestion
    display: unset;
```

//PS: setting the `display:none` is probably not even needed, see below

modules/svg/vertices.js
```diff
@@ -116,6 +117,23 @@ export function svgVertices(projection, context) {
...
+        enter
+            .filter((d) => getRadiusTag(d.tags))
+            .append('circle')
+            .attr('class', 'radius')
+            .attr('data-nodeid', node => node.id);
+
+        // we need to trigger a re-render when the zoom changes
+        selection
+            .selectAll('.radius')
+            .data([zoom])
```
tyrasd: I think this does not work as intended: it's buggy when there is more than one point/vertex to be rendered with a circle in the map view: `.data([zoom])` would always only affect the first matched `.radius` SVG element (because `[zoom]` is an array with only one element). Multiselections would also not show the diameters of all features.

I think it would be cleaner to replace this with an implementation similar to how the viewfield cones are rendered below, i.e. something like the following: 


```js
        // Highlighted vertices with a radius/diameter get a circle
        const circles = groups
            .selectAll('.radius')
            .data(d => d.id in sets.selected && getRadiusTag(d.tags) ? [d] : [], fastEntityKey);
        circles.exit().remove();
        circles.enter()
            .append('circle')
            .attr('class', 'radius')
            .merge(circles)
            .attr('r', node =>
                Math.max(0, getRadiusInPixels(node, context.projection))
            );

        // Vertices with directions get viewfields
        var dgroups = groups
        …
```

This does also only add the `<circle>` elements when needed (i.e. the selected map features), resulting in no rendering overhead and the CSS can also be simplified a little bit. 😊  And the `data-osmid` attribute is also not needed anymore when one continues to bind the OSM entity data to the selection.

tyrasd (commented):
Regarding the concrete styling: I think the red-ish color is a little bit distracting in this context. It reminiscent of the color iD uses for buildings, or could be interpreted as an error even.

What about going with a neutral styling similar to the viewport cones?

image

```css
g.point .radius,
g.vertex .radius {
    fill: #333;
    fill-opacity: 0.6;
    stroke: #fff;
    stroke-width: 0.5px;
    stroke-opacity: 0.75;
}
.fill-wireframe .radius {
    fill: none;
}
```

modules/core/planar.ts
tyrasd: According to overpass, there are currently about 250 object with a tag that ends with ′ in the value (400 more have the ′ character somewhere else in the tag value)

1ec5: There are probably more objects with U+2019 (’), which would be an obvious mistake compared to U+2032 (′).

modules/svg/vertices.js
matkoniecz: @k-yle can you look into this?

k-yle: thanks, this is much more elegant. updated now and tested with multiselections

css/20_map.css
k-yle: yeah, this is no longer applicable, deleted

modules/svg/points.js
k-yle: updated so it refuses to render a circle where the radius is > [`innerWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth) or [`innerHeight`](https://developer.mozilla.org/en-US/docs/Web/API/Window/innerHeight). and ignore negative numbers

modules/core/planar.ts
k-yle: prime marks removed. i think we'll end up deleting this file anyway after https://github.com/ideditor/schema-builder/pull/198 lands

k-yle:
styling also changed per suggestions above:
image
image

matkoniecz:
@k-yle 

> This branch has conflicts that must be resolved

k-yle:
@matkoniecz almost all my PRs will have that label, because editing the CHANGELOG file will create conflicts if a different PR is merged first

css/20_map.css
```diff
@@ -188,6 +188,25 @@ g.vertex.selected .shadow {
     fill-opacity: 0.7;
 }
 
+/* circle to visualize radius/diameter */
+g.point .radius,
+g.vertex .radius {
+    fill: #333;
```
tordans, line 194: Can we share some of this with the direction cone?

k-yle: not easily, it would end up being more lines of code

matkoniecz:
```
natural=tree
diameter=2m
diameter_crown=10m
```

this will show only the first one

maybe both can be shown?

k-yle:
> maybe both can be shown?

possibly, but the `diameter` of a tree is usually impossible to see from aerial imagery, right? 

so maybe we should only show 1 tag, and prioritise `diameter_crown` over `diameter` ?

matkoniecz:
> possibly, but the diameter of a tree is usually impossible to see from aerial imagery, right?

I thought mostly that it would 

- look pretty
- be less confusing when some trees have `diameter` and show this, while other also have `diameter_crown` and show something else instead
- be used to spot blatantly invalid data

k-yle:
conclusion:
- we should render multiple rings if there are multiple radius-like tags on the same node
- we should also render $\frac{\text{circumference}}{2\pi}$ 

that means there could be up to 3 circles rendered for a single tree, `diameter`, `diameter_crown`, `circumference`

there are 13337 nodes with both `diameter` and $\frac{\text{circumference}}{2\pi}$, which should ideally be consistent

k-yle:
updated to support the `circumference=*` tag, and to support multiple tags on the same node:

image

usually`circumference` will be too small to see, unless you zoom very far in

modules/renderer/map.js
1ec5, line 1: When I navigate to and from the vertex with the `diameter=*` tag using the <kbd>[</kbd> and <kbd>]</kbd> keys, the halo only appears or disappears about half the time.

modules/svg/vertices.js
```diff
@@ -165,6 +166,20 @@ export function svgVertices(projection, context) {
...
+            .data((d) => context.selectedIDs().includes(d.id) ? getRadiiInPixels(d, projection) : []);
+
+        circles.exit()
+            .remove();
+
+        circles.enter()
+            .insert('circle', '.shadow')
+            .attr('class', 'radius')
+            .merge(circles)
+            .attr('r', d => d);
```
1ec5, line 181: Not sure if it’s a problem, but `highway=turning_circle` `turning_circle=knuckle` etc. render the full circle. The circle is still useful as a guide, as long as the user doesn’t perceive it as an attempt to render the physical turning “circle”.

modules/renderer/map.js
k-yle: good spotting, that's an existing performance optimisation to avoid redrawing while you're zooming along with the <kbd>[</kbd> or <kbd>]</kbd> keys. 

to make it less confusing, i've changed it to immediately remove any radii if there were some selected. but it won't redraw them until the 750ms timeout is reached

modules/svg/vertices.js
k-yle: there are only 58 in the world right now with `turning_circle=knuckle` + `diameter=*`, so maybe could wait and see if users complain about any situations like this?

imo the right style is better, in which case this is not an issue. but the wiki [doesn't say](https://wiki.openstreetmap.org/wiki/Key:turning_circle) which style is preferred

image

matkoniecz (approved):
I have not managed to have complaints this time :)

PR #9765, opened 2023-07-13
Update dependabot.yml

als-alt:


pnorman:
This removes the override file and goes back to the default behavior, which is not really what the PR description says.

tyrasd:
@als-alt please tell us what you had in mind with this PR. As it stands now – without any explanation or reasoning – it can't be considered.

Issue #9233, opened 2022-08-06
addr:housename textbox skips to the end 'onkeypress'


UKChris-osm:
### URL

https://www.openstreetmap.org/#map=19/51.04120/-3.06323

### How to reproduce the issue?

When editing an **addr:housename** tag, iD will skip to the end of the text area if you attempt to edit a character that isn't at the end of the text.

1: Attempted to edit any instance of  **addr:housename**
2: Select anywhere in the text box and type something (or press delete).
3: The text pointer will skip to the end of the text box upon a keypress event.

### Screenshot(s) or anything else?

https://user-images.githubusercontent.com/59393694/183243211-3fc4b622-8a8b-4ce8-bf87-869cae9f056f.mp4

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app, RapiD version at mapwith.ai/rapid

### What version numbers does this issue effect?

2.21.1

### Which browsers are you seeing this problem on?

Chrome, Edge

UKChris-osm:
I just tried using the Developer Tools to remove an event listener from this element that's triggered upon "input" and this seems to remove the problem.

1ec5:
This happens on a lot of fields that iD uses to label features on the map or format in some way. The `addr:housenumber` and `addr:unit` fields are particularly problematic, because I often need to edit the ranges in non-sequential order, but then the live update and validation causes my input to get mangled. The live updating is a cool effect, but I don’t think much would be lost by making the map update onblur. Unfortunately, we’ve had a lot of issues in the past with edits not taking or being misapplied because of elements getting detached from the DOM before they’ve had a chance to lose focus, such as in #5449 #5824 #6028.

tyrasd:
this was fixed in 14c752548

PR #9766, opened 2023-07-13
fix the raw tag editor broken and strange cursor behaviour

### LINKED ISSUE #9762: Tag raw editor is empty
As of 2.26 or 2.26.1 the raw tag editor list is empty:

empty tag list

You can add text to it, but it generally leads to broken tagging, as it won't preserve existing ones.

This happens on any selected element - existing, new, modified.

No errors in console.

Win 10 Enterprise x64, Firefox and Opera.

### LINKED ISSUE #9764: addr:housenumber is glitchy during typing
### URL

https://www.openstreetmap.org/edit#map=19/51.74659/20.77788

1. Select an address
2. Clear the house number field
3. Start typing at least 2 char long number: 123
4. Result is 231

_No response_

2.26.1

Chrome

### LINKED ISSUE #9767: Changeset comment dropdown does not work
### URL

_No response_

Clicking the changeset comment box produces a dropdown list of recent comments, but clicking those comments no longer works - the dropdown disappears but the comment box is still empty

2023-07-13 09_52_06-OpenStreetMap — Mozilla Firefox

2.26.1,2.27.0-dev

Chrome, Firefox

### LINKED ISSUE #9773: Cannot select recent changeset comment
### URL

Anywhere

1. Make changes on OSM
2. Click save
3. Click the textbox where the changeset comment should go to see all the recent changeset comments you can select
4. Click on a changeset comment you'd like to select
5. The changeset comment did not get added to the textbox

This is the area I'm talking about
image

Furthermore, the textbox was already supposed to be filled because I am doing a Maproulette challenge that auto fills the textbox. 
Also, whenever I type one character of the top suggestion, so in this case #, it will suggest the top comment and fill it in if I click tab.

iD 2.26.1

Microsoft Edge

### LINKED ISSUE #9774: Text mode editing for Tags does not work with iD 2.26.1
### URL

_No response_

1. Switch the tag list to text mode
2. Select any object on the map
3. Observe the tag list contents

Nothing happens, i.e. it is empty (although it does accept input, but it still does not change with selection). This is a horrible regression introduced in 2.26.0 or 2.26.1, so much I'd prefer it rolled back until fixed, since it makes any work with tags extremely cumbersome. 

TagsBox

_No response_

2.26.1 (perhaps 2.26.0)

Firefox

### LINKED ISSUE #9775: Issue with Tag Conversion and Auto-modification of Elements
### URL

No response

1. Convert a list of tags to text.
2. Observe that the tags within the list disappear and the original elements lose their attributes.
3. Copy the tag text.
4. Notice that certain elements in the viewport are automatically modified to match the copied tag without any editing being performed.

Uploading id 2.26.1.mp4…

2.26.1

Chrome

k-yle:
Closes #9762, Closes #9764, Closes #9767, Closes #9773, Closes #9775, Closes #9774

to test:
 - go to https://osm.org/edit
 - select any feature with tags
 - note how the raw tag editor works again

edit: this PR indirectly fixes another bug reported on the OSM-US slack, where the changeset comment dropdown doesn't work

### DIFF
--- a/modules/util/get_set_value.js
+++ b/modules/util/get_set_value.js
@@ -32,11 +32,12 @@ export function utilGetSetValue(selection, value, shouldUpdate) {
         // see https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
         const supportedTypes = ['text', 'search', 'url', 'tel', 'password'];
         return function() {
+            func.apply(this, arguments);
+
             if (!supportedTypes.includes(this.type)) {
                 return;
             }
             const cursor = { start: this.selectionStart, end: this.selectionEnd };
-            func.apply(this, arguments);
             this.setSelectionRange(cursor.start, cursor.end);
         };
     }
--- a/modules/util/get_set_value.js
+++ b/modules/util/get_set_value.js
@@ -28,16 +28,9 @@ export function utilGetSetValue(selection, value, shouldUpdate) {
     }
 
     function stickyCursor(func) {
-        // only certain input element types allow manipulating the cursor
-        // see https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
-        const supportedTypes = ['text', 'search', 'url', 'tel', 'password'];
         return function() {
-            func.apply(this, arguments);
-
-            if (!supportedTypes.includes(this.type)) {
-                return;
-            }
             const cursor = { start: this.selectionStart, end: this.selectionEnd };
+            func.apply(this, arguments);
             this.setSelectionRange(cursor.start, cursor.end);
         };
     }
@@ -48,6 +41,13 @@ export function utilGetSetValue(selection, value, shouldUpdate) {
 
     if (shouldUpdate === undefined) {
         shouldUpdate = (a, b) => a !== b;
+    }
+
+    // only certain input element types allow manipulating the cursor
+    // see https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
+    const supportedTypes = ['text', 'search', 'url', 'tel', 'password'];
+    if (!supportedTypes.includes(this.type)) {
+        return selection.each(setValue(value, shouldUpdate));
     }
 
     return selection.each(stickyCursor(setValue(value, shouldUpdate)));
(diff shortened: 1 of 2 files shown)


modules/util/get_set_value.js
```diff
@@ -32,11 +32,12 @@ export function utilGetSetValue(selection, value, shouldUpdate) {
         // see https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
         const supportedTypes = ['text', 'search', 'url', 'tel', 'password'];
         return function() {
+            func.apply(this, arguments);
+
             if (!supportedTypes.includes(this.type)) {
                 return;
             }
             const cursor = { start: this.selectionStart, end: this.selectionEnd };
```
1ec5, line 32: Based on the code that changed, these regressions were probably introduced in 14c752548d2899b7411fcc8480ce3e2784df1253 (related to #8769 and 2b64d703523a2856bd41b894dff9fcd5dc4bdc75). I believe the intention was for this line to memoize the cursor position before setting the value in order to restore that position afterwards. It needs to go to the very top of this function before the call to `apply()`.

1ec5 (commented):
Good catch, but this introduces another subtle regression that should hopefully be simple to avoid.

modules/util/get_set_value.js
k-yle: thanks, it seems like this bug was present even before this PR (#9764) 

I've reworked the logic now so it also fixes #9764. are there any specific cases that are still broken for you?

1ec5: Sorry, I was a bit unclear. 14c752548d2899b7411fcc8480ce3e2784df1253 attempted to fix #9233 but barely avoided fixing #9764 while creating other regressions. As far as I can tell, these changes do fix both issues.

1ec5 (approved):

modules/util/get_set_value.js
```diff
@@ -49,5 +43,12 @@ export function utilGetSetValue(selection, value, shouldUpdate) {
         shouldUpdate = (a, b) => a !== b;
     }
 
+    // only certain input element types allow manipulating the cursor
+    // see https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
+    const supportedTypes = ['text', 'search', 'url', 'tel', 'password'];
+    if (!supportedTypes.includes(this.type)) {
```
tyrasd, line 49: This does not quite work as probably intended: `this` is here the global (`window`) object, and not the input element we'd want to check (d3's `selection.each' binds the `this` variable to the DOM element in the callback function).

So, in practice, this condition is always fulfilled, even for input element types which would support cursor manipulation. As it still fixes the textarea/raw tag editor bug, I think we can keep it for the moment (even though #9233 is now an issue again), but I'll see if I can fix this a little bit better. :blush:

Issue #8651, opened 2021-08-19
Allow using TypeScript in iD's source?


til-schneider:
I'd like to start a discussion whether iD should allow to use TypeScript in its source.

Using TypeScript in iD's source [was rejected by @bhousel in 2018](https://github.com/openstreetmap/iD/issues/3744#issuecomment-373035353). But it seems as he changed his opinion about TypeScript since then as he is using TypeScript in [id-sdk](https://github.com/ideditor/id-sdk) and [RapiD - Facebook's fork of iD](https://github.com/facebookincubator/RapiD) (as he pointed out in [a comment](https://github.com/openstreetmap/iD/pull/7861#issuecomment-895221588) and in [his presentation at SOTM-US in 2019](https://youtu.be/cnML0w0qFCw?t=880)).

In my experience, using TypeScript does nothing less than taking the coding experience to a whole new level: When using an editor with TypeScript support (like [VS Code](https://code.visualstudio.com/) for instance) you get code assistance as you type. This makes coding much more fluid and less error-prone than the indirect and delayed console outputs of the build or - even worse - errors which occur at runtime.

In my private fork of iD I added type information to iD (in the form of `.d.ts` files), which prevents me from looking up internal APIs all the time. If the iD team decides to allow TypeScript, I would be happy to contribute this work (by migrating the according `.js` files to `.ts` files, instead of adding `.d.ts` files).

In order to allow TypeScript, iD's build has to be changed in order to support this. I already contributed TypeScript build support (using the TypeScript compiler) in [my PR "Add support for ReactJS"](https://github.com/openstreetmap/iD/pull/7861). As [I wrote in a comment](https://github.com/openstreetmap/iD/pull/7861#issuecomment-895006514), I would also contribute TypeScript build support using [esbuild](https://github.com/evanw/esbuild) if the iD team decides rather to go this way, which would dramatically speed up build time.

matkoniecz:
How large is the increase of complexity in setup, for new contributors and in toolchain?

til-schneider:
The setup is just the same: Once `npm i`, then the usual build commands `npm run …`.

tyrasd:
[...]

PR #9805, opened 2023-08-01
DX improvements for the unit tests

k-yle:
Installing `@types/chai` and declaring `expect` as a global means that IDEs like VS Code can provide auto-complete suggestions as you type, which makes it much easier to write tests for people who aren't familiar with chai

likewise, declaring the type of `window.iD` means that IDEs can show suggestions and allow you to use a keyboard shortcut like <kbd>F12</kbd> to jump to the definition of any function. I think this would pretty helpful for occasional contributors like myself


k-yle:
superseeded by #10452

PR #9816, opened 2023-08-06
optionally ignore relations in the extract operation

### LINKED ISSUE #6700: When a node gets detached from way, the relations the node belonged to stay on the way node, rather than on the detached node
When a node gets detached from a way, the relations the node belonged to stay on the way node, rather than on the detached node. The detached node should mantain the relations instead of the way node.

### LINKED ISSUE #6807: Relations break on node unglue
I am reporting for the forum user aixbrick following this thread: 
https://forum.openstreetmap.org/viewtopic.php?pid=760219#p760219

# What did they do
The user unglued a node which was part of a road and also part of 10 relations.

# What happened
The node was detached from the road as expected. Then it was also removed from all 10 relations. 
A new node was automatically created and added to the relations, this node was at the position of the other node before unglueing (it was also part of the highway/road).

# What should have happened
There are different possibilities, including asking the user for feedback, adopt the automatic actions according to the relation type etc., but in the most simple version it should:
just unglue the node like it did, but NOT remove it from any relations, NOT create a duplicate node of it. Just do nothing more than unglue the node from the way (remove the node from the way so it can be freely moved independently of the way geometry to which it belonged before).

My personal comments:
IMHO, maybe in most cases it would make more sense to create indeed a duplicate node at the former position, e.g. to keep the highway geometry the same, but I agree the relations should not be changed (remove old node, add new node), currently this is creating a lot of new object versions by trying to do actively "nothing". JOSM is asking the user which node should get a new id / keep the old id. Maybe you can convert this to a question that sounds less technical and still allows for user feedback about the desired behaviour.

### LINKED ISSUE #8929: Extract tool removes relations from node
### URL

https://www.openstreetmap.org/edit?node=4940530185#map=20/-23.56026/-46.65741

Hi,

The extract tool is removing relations from node.

How to reproduce (testing with this bus stop https://www.openstreetmap.org/node/4940530185):

1. Select a bus stop, which is member of bus route, stop area relations, etc and is attached to a building;
2. Use "Extract" to extract the bus stop from the building;
3. Now the bus stop has no relations, and all related relations are affected.

The same happens with other nodes with relations.

image
image
image

Chrome

### LINKED ISSUE #9229: Node gets removed from relations when taken out of way
### URL

https://www.openstreetmap.org/edit?editor=id&node=6813583920#map=19/48.16067/16.30284

- Select/right click a node that is part of a way (but not the start or end point) and is part of a relation (eg. stop_position of a bus line)
- select the function to take the point out of the way (image)
--> The removed node is no longer member of any relation it used to be a member of.

**- Expected behavior: Like in JOSM, the node keeps it's membership of all relations it is a part of.**

image
image

2.21.1 ; 2.22.0-dev

Chrome, Firefox, Microsoft Edge, Safari

### LINKED ISSUE #9359: Easier workflow for adding members to a relation
### Description

Copying an item should not only take tags, but also relations with it.
E.g. if I have to make a new bus/tram stop it takes a lot of work.

### Screenshots

_No response_

### LINKED ISSUE #9394: Relations not copied along with the node
Let's say a user is having great success using control+C to copy things, until one day he discovers that relations are not copied when copying the node.

In other words we create a node, add it to a relation, and then we copy it, and then we paste the copy somewhere else  and all the tags are there, except that relation stuff at the bottom of the list. That got dropped.

So then we got to find all the nodes that we copied like that and then add them to the relation, doubling our work. Yes we multi-select them one by one and then one more step to add them all to the relation. But still we're wondering isn't there a way just to copy that **"complete node"?**

Yes you might say well we are supposed to add them to the relation in their certain order, but sometimes order doesn't matter...

Or, our control+V's would indeed place them in the right order along the side of a road, so why can't they be automatically added to a relation?

### LINKED ISSUE #10234: Extrating a bus stop (highway=bus_stop) node doesn't keep relation
### URL

https://www.openstreetmap.org/edit?node=304828058#map=19/46.81263/-71.21529

1. Select a bus stop node that is part of a road
2. Extract the bus stop node
3. Notice that the extracted node wasn't retained in the relation

_No response_

2.29

Firefox

k-yle:
Closes #6700, Closes #6807, Closes #8929, Closes #9229, ~~Closes #9359~~, Closes #9394, Closes #10234

If you hold down <kbd>⇧ Shift</kbd> while triggering the extract operation (<kbd>E</kbd> by default), then the operation will not extract relations from the node. 

There's a good example in the original issue #8929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Fix flickering when switching between background imagery layers, make switching backgrounds snappier
 * Prevent password managers from autofilling tag fields ([#10508], thanks [@michaelabon])
 #### :scissors: Operations
+* When extracting a node from a way (shortcut: <kbd>E</kbd>), the relations are now preserved by default. Extracting a node without its relations is still possible using <kbd>⇧ Shift</kbd> <kbd>E</kbd> ([#9816], thanks [@k-yle])
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
@@ -61,6 +62,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Migrate unit tests from karma to vitest ([#10452])
 
 [#9013]: https://github.com/openstreetmap/iD/issues/9013
+[#9816]: https://github.com/openstreetmap/iD/issues/9816
 [#10452]: https://github.com/openstreetmap/iD/pull/10452
 [#10459]: https://github.com/openstreetmap/iD/pull/10459
 [#10488]: https://github.com/openstreetmap/iD/pull/10488
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2413,6 +2413,7 @@ en:
         merge: "Combine (merge) selected features"
         disconnect: "Disconnect selected features"
         extract: "Extract a point from a feature"
+        extract_without_relations: "Extract a point from a feature, without extracting its relations"
         split: "Split features at the selected points"
         reverse: "Reverse selected features"
         move: "Move selected features"
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -272,6 +272,11 @@
             "text": "shortcuts.editing.operations.extract"
           },
           {
+            "modifiers": ["⇧"],
+            "shortcuts": ["operations.extract.key"],
+            "text": "shortcuts.editing.operations.extract_without_relations"
+          },
+          {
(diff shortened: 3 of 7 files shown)


tyrasd (commented):
Maybe we should swap the two operations around: Without `shift` keeps the relations intact, and with `shift` it drops them?

And perhaps the edit menu should be adapted to show the appropriate description whether the shift key is pressed or not?!

tyrasd (approved):

Dimitar5555:
I think that this PR should also close #6700.

PR #9847, opened 2023-08-16
Adds nycdoitt:bin tag to list of tags to keep on building

### LINKED ISSUE #8539: nycdoitt:bin building tags stripped on "Extract" and "Downgrade" operations
This issue concerns the `nycdoitt:bin` tag, which is part of the New York City building footprint import. This tag holds the building id number (BIN) of the imported building. It's associated with the physical structure, not with the building's use or contents.

**Expected behavior:** `nycdoitt:bin` tags will be considered building tags and will remain on a building way (or multipolygon) after "Downgrade" or "Extract" operations. They will not be copied to POI nodes created during "Extract" operations.

**Observed behavior:** `nycdoitt:bin` tags are removed from buildings during "Downgrade" and "Extract" operations, and are copied to the newly created POI node during "Extract."

**Steps to reproduce:**
1. Right-click on an imported New York City building footprint that also has POI tags (eg [https://www.openstreetmap.org/way/265796841](https://www.openstreetmap.org/way/265796841))
2. Select "Downgrade" from context menu
3. Observe tags of building way -- `nycdoitt:bin` has been removed
4. Undo to restore building way tags
5. Right-click and select "Extract" from context menu
6. Observe tags of building way and newly created POI node -- `nycdoitt:bin` has been removed from way and added to node

mxxcon:
Closes #8539
When extracting a POI from a building or downgrading a building, this update keeps `nycdoitt:bin=` tag with the building, and not extracts it to the node or deletes it.


mxxcon:
PR sitting for 4 month with no action.
It was literally 2 lines of changes to preserve data from getting destroyed!
Ridiculous.

riQQ:
Integrated via 87819ed53f67f4ba3999d1bffd7145454d5c0826

PR #9857, opened 2023-08-18
Add a vector layer by TileJSON

### LINKED ISSUE #6021: Overzoom vector tiles
Vector tile support is amazing \o/

Vector tiles are typically provided up to (say) z14, with all data for lower zoom levels in the z14 tiles. Currently, iD only requests the vector tile for the current zoom level. It would be great to "overzoom" earlier zoom levels if a 404 is received.

Possible ideas for doing it:

- provide a "max zoom" field in the Custom Map Data dialogue
- attempt to fetch z14 if 404 received for a higher level (z14 seems to be a fairly wide standard)
- iterate up zoom levels until no 404 received (may be a bit messy to implement)

jleedev:
Can supply a URL pointing to a TileJSON resource, from which we can extract the xyz mvt URL, minzoom, and maxzoom. Fixes #6021.

Accept any URL that returns the appropriate content-type rather than requiring a URL that ends in ‘.json’, but still fall back to inspecting the extension if the content-type is not an expected one. (This might be from a misconfigured web server, but browsers seem to be flaky when loading a local file.)

FYI, local file uploads are sort of broken on Chrome Android: https://crbug.com/1466960

The settings page seems to forget to show the current URL, possibly as of https://github.com/openstreetmap/iD/commit/16280638b5a6bfaa941a1bee18d57e79dea8e65c

The changeset’s `imagery_used` continues to be the underlying template URL.

The format argument to setFile is unused except for when the editor is loaded with `#gpx=(url)`. This could perhaps be generalized to accept the URL of any data layer we would like to load, not just gpx. Some data formats really aren't self-describing and need some sort of type identifier, and the URL alone which is stored in the settings might need to be changed to something more structured. (For example, FlatGeobuf, PMTiles, and CSV can perhaps be trusted to always have a file extension in the URL; TopoJSON is often just `.json` but can be identified by the `"type"` key … there's some design space to be explored here.)


matkoniecz:
It is quite sad that it has not been reviewed so far.

Are you interested in following up on this one if review would finally arrive? I have some capacity for reviewing and testing, but there is noticeable risk that this PR may be (understandably) abandoned.

1ec5:
The conflicts so far are just in the NPM dependency files. This PR adds a few new dependencies, some of which have also been added in develop, so we’d just need to adopt the latest version of the dependency between the two branches.

PR #9883, opened 2023-09-04
add a validator for `lanes=1` + `oneway=no`

k-yle:
This PR adds a new validator warning when a road has [`lanes`](https://osm.wiki/Key:lanes)=`1` but is does not have [`oneway`](https://osm.wiki/Key:oneway)=[`yes`](https://osm.wiki/Tag:oneway=yes)/[`alternating`](https://osm.wiki/Tag:oneway=alternating)/[`reversible`](https://osm.wiki/Tag:oneway=reversible)

image

### DIFF
--- a/modules/validations/oneway_onelane.js
+++ b/modules/validations/oneway_onelane.js
@@ -1,0 +1,76 @@
+import { actionChangeTags } from '../actions/change_tags';
+import { t } from '../core/localizer';
+import { validationIssue, validationIssueFix } from '../core/validation';
+import { utilDisplayLabel } from '../util';
+
+
+export function validationOneLaneWithNoOneway() {
+  const type = 'oneway_onelane';
+
+  /** @param {string} onewayValue  @param {osmEntity} entity */
+  function createSuggestion(onewayValue, entity) {
+    return new validationIssueFix({
+      icon: 'iD-icon-data',
+      title: t.append(`issues.fix.tag_as_oneway_${onewayValue}.title`),
+      onClick: (context) => {
+        const newTags = { ...entity.tags, oneway: onewayValue };
+        context.perform(
+          actionChangeTags(entity.id, newTags), t(`issues.fix.tag_as_oneway_${onewayValue}.annotation`)
+        );
+      }
+    });
+  }
+
+  /** @param {osmEntity} entity */
+  function makeOneLaneIssue(entity) {
+    return new validationIssue({
+      type,
+      subtype: type,
+      severity: 'warning',
+      message: (context) => t.append(
+        'issues.oneway_onelane.message',
+        { feature: utilDisplayLabel(entity, context.graph()) }
+      ),
+      reference: showReference,
+      entityIds: [entity.id],
+      hash: type,
+      dynamicFixes: () => [
+        // suggest adding `oneway` tag (automatic)
+        ...['yes', 'alternating'].map(onewayValue => createSuggestion(onewayValue, entity)),
+
+        // or changing the `lanes` tag (manual)
+        new validationIssueFix({ title: t.append('issues.fix.change_lane_tag.title') })
+      ]
+    });
+  }
+
+
+  function showReference(selection) {
+    selection.selectAll('.issue-reference')
+      .data([0])
+      .enter()
+      .append('div')
+      .attr('class', 'issue-reference')
+      .call(t.append(`issues.${type}.reference`));
+  }
+
+  /** @param {osmEntity} entity */
+  const validation = (entity) => {
+    const isOneLane = entity.tags.lanes === '1';
+
+    if (
+      entity.type !== 'way' ||
+      entity.isOneWay() ||
+      !isOneLane ||
+      entity.tags.leisure === 'slipway'
+    ) return [];
+
+    // this line has lanes=1, but is not oneway. This makes no sense.
+    return [makeOneLaneIssue(entity)];
+  };
+
+
+  validation.type = type;
+
+  return validation;
+}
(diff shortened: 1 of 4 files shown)


test/spec/validations/oneway_onelane.js
```diff
@@ -0,0 +1,88 @@
...
+            expect(issue.type).to.eql('oneway_onelane');
+            expect(issue.subtype).to.eql('oneway_onelane');
+            expect(issue.severity).to.eql('warning');
+            expect(issue.entityIds).to.have.lengthOf(1);
+            expect(issue.entityIds[0]).to.eql('w-1');
+            done();
+        }, 20);
+    });
+
+    it('flags roads with lanes=1 no oneway tag', (done) => {
```
1ec5, line 83: Can you add a test that a roundabout (or any other way that’s implied to be `oneway=yes`) can have `lanes=1`?

modules/validations/oneway_onelane.js
```diff
@@ -0,0 +1,71 @@
...
+      .call(t.append(`issues.${type}.reference`));
+  }
+
+  /** @param {osmEntity} entity */
+  const validation = (entity) => {
+    const isOneLane = entity.tags.lanes === '1';
+
+    if (entity.type !== 'way' || entity.isOneWay() || !isOneLane) return [];
+
+    // this line has lanes=1, but is not oneway. This makes no sense.
```
1ec5, line 81: Just to make sure we’re covering all our bases:

* One-lane bridges are very common in some rural areas where I map. Traffic can enter from either direction, but not at the same time. Technically the lane is reversible/alternating, but not on a schedule and not controlled by a stoplight as illustrated on the `oneway=alternating` page. Would this situation still be `oneway=alternating`?
* Are there any situations where we’d need to consider conditional tags?

modules/validations/oneway_onelane.js
```diff
@@ -0,0 +1,71 @@
...
+        { feature: utilDisplayLabel(entity, context.graph()) }
+      ),
+      reference: showReference,
+      entityIds: [entity.id],
+      hash: type,
+      dynamicFixes: () => [
+        // suggest adding `oneway` tag (automatic)
+        ...['yes', 'alternating'].map(onewayValue => createSuggestion(onewayValue, entity)),
+
+        // or changing the `lanes` tag (manual)
```
1ec5, line 41: Should we add a suggestion of adding `lane_markings=no`, for unstriped two-way roads?

pnorman (changes requested):
Only about 58% of lanes=1 roads have one of these oneway values. In fact, 7.7% of them have an explicit oneway=no value, so this PR seems like a bad idea.

modules/validations/oneway_onelane.js
bhousel: My first thought was also one-lane bridges - they are very common.

test/spec/validations/oneway_onelane.js
k-yle: 👍 done

modules/validations/oneway_onelane.js
k-yle: sounds good, I've made this change

modules/validations/oneway_onelane.js
k-yle: > - [...] Technically the lane is reversible/alternating, but not on a schedule and not controlled by a stoplight as illustrated on the `oneway=alternating` page. Would this situation still be `oneway=alternating`?

My understanding is that this is a perfect example of `oneway=alternating`. If there are no traffic lights then the [`priority`](https://osm.wiki/Key:priority) tag is useful, which [the wiki page](https://osm.wiki/Tag:oneway=alternating) mentions.

> - Are there any situations where we’d need to consider conditional tags?

I can't think of any situatations because in this case, `oneway:conditional` should be combined with a valid default `oneway` tag (`yes`/`-1`/`alternating`/`reversible`), like [this example from Brisbane](https://goo.gl/maps/o9TDeoLW2YD6Y9fG6)

k-yle:
> Only about 58% of lanes=1 roads have one of these oneway values. In fact, 7.7% of them have an explicit oneway=no value, so this PR seems like a bad idea.

I had a look at some of these and most seem to be mistakes... The ones that I reviewed should either have `oneway=alternating` or have `lane_markings=no` instead of `lanes=1` based on [the example](https://osm.wiki/Key:lane_markings)[s in the wiki](https://osm.wiki/Key:lanes).

So my hope is that this PR will reduce some of the confusion that led to 7.7% having `oneway=no`

riQQ:
At the moment StreetComplete tags `oneway=no` on a way with `lanes=1` or no `lanes` tag after answering *No / both directions* to the oneway quest. If this gets merged, StreetComplete should be changed, too.

Relevant Code:
https://github.com/streetcomplete/StreetComplete/blob/3b074e1d077aad25071e9773f6471b2490077869/app/src/main/java/de/westnordost/streetcomplete/quests/oneway/AddOneway.kt#L26-L31

pnorman:
I think this can be closed.

Given these are being added and a large percentage of them are present, I don't think there should be a validation rule in place. If these are an error, existing software should stop adding them and most of the mistakes fixed, so user's don't get tons of validation errors when editing an existing area.

I'm not convinced this is a mistake, but that's a discussion out of scope of iD.

tordans:
@k-yle are you still in favor of adding this to iD?

k-yle:
> @k-yle are you still in favor of adding this to iD?

Since there are editors like StreetComplete who are promoting `lanes=1` + `oneway=no`, I don't think iD can show a warning (yet), even if it contradicts the wiki and makes no sense... There are just too many examples which are wrongly tagged. 

So we should probably close this PR, maybe it can be revisited one day

tordans:
Agreed. Thanks für the productive discussion!

PR #9888, opened 2023-09-06
avoid ambiguous wording in the `crossing_ways` validator

### LINKED ISSUE #7793: Say which item is to be tagged as higher / lower
Maybe say:
- Tag 老人活動中心 as higher
- Tag 老人活動中心 as lower...

or

- Tag stream as higher
- Tag stream as lower ...

else the wording is ambiguous, and to some users
it seems you are talking about the stream, despite,
yes, 老人活動中心 blinking, etc. so they are not 100% sure
of your meaning.

### LINKED ISSUE #8657: Item crossing other item warning isn't as descriptive as it could be
Screenshot 2021-08-23 222817

This warning is mostly helpful, but it could be a lot clearer!  Currently (based on this text alone) it is unclear which object is to be made higher or lower.  Selecting either object will change the outcome of the warning and while the end result works as intended it's not terribly user friendly.

This is really just a copy fix, instead of telling users to "Tag _this_ as higher" substitute "this" for the name of the object in question, eg: "Tag _Path_ as higher".  This should give users a clearer idea of which object they are changing to be on top or below as the new copy gives more insight into the action they are performing.

### LINKED ISSUE #8728: "This" ambiguous
20211002T221912a

### LINKED ISSUE #9728: tag_this_as_higher|lower ambiguous
### URL

_No response_

- Create a new path that intersects with a building.
- A warning concerning the intersection will appear, asking you to specify the vertical ordering:
screenshot-2023-07-04T222445+0200
- Selecting `tag_this_as_higher` ("Das ist darüber" in DE) or `tag_this_as_lower` ("Das ist darunter" in DE) will set the layer on the new path.

### Problem description
At least in the German translation, the wording is very ambiguous. "This" can refer to either the path or the building. I have to guess from the `layer` tag this adds to know what it actually does.

Changing this to say "Footpath is higher" ("Fußweg ist darüber) would be the best solution, i think.

2.25.2

Firefox

k-yle:
Closes #7793, Closes #8657, Closes #8728, Closes #9728

Replaces "this" with the feature's name to avoid ambiguous wording. 

I have added 2 new translatable strings instead of repurposing the existing ones, so that translators are prompted to create more helpful translations.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>

 <td>

</table>


Dimitar5555:
It should also close #8728.

k-yle:
thanks, description updated

modules/core/validation/models.js
```diff
@@ -75,28 +76,28 @@ export function validationIssue(attrs) {
 
         fixes.forEach(function(fix) {
             // the id doesn't matter as long as it's unique to this issue/fix
-            fix.id = fix.title.stringId;
+            // the id does matter if the fix depends on the currently selected feature.
```
matkoniecz:
```suggestion
            // except cases where fix depends on the currently selected feature.
```

attempt to make it easier to parse

I think removing self-contradiction makes easier to read it

matkoniecz (commented):
tested on local branch where latest master was merged in

for code I have only minor comment text suggestion

PR #9891, opened 2023-09-07
Add rendering for roller_coaster=track and roller_coaster=support

### LINKED ISSUE #9839: Add rendering for roller_coaster=track
### Description

Rollercoaster tracks are [soon being rendered in Carto](https://github.com/gravitystorm/openstreetmap-carto/pull/4666) and I think it would be appropriate to do the same in iD. 

I already prototyped this change and could open a PR, if you think that this is okay. The rendering is similar to the one used in Carto.
This change would go along nicely with https://github.com/openstreetmap/id-tagging-schema/issues/566 (for which I also could open a PR).

Current:
Screenshot 2023-08-14 at 19 34 22

Prototype:
Screenshot 2023-08-14 at 19 34 47

I am unsure about roller_coaster=support. Should they be rendered as well? If so, I would think of using the same dark gray background color. What do you think?

### Screenshots

_No response_

mangerlahn:
Closes #9839 by adding rendering to `roller_coaster=track` and `roller_coaster=support`. 
Fits the [preset PR](https://github.com/openstreetmap/id-tagging-schema/pull/985) and [the rendering used in OpenStreetMap Carto](https://github.com/gravitystorm/openstreetmap-carto/pull/4666).

I am a bit unsure on the rendering of the support beams. Are they being rendered too thick, is the contrast too low?

Before:
Screenshot 2023-09-07 at 14 36 15

After:
Screenshot 2023-09-07 at 14 35 47

### DIFF
--- a/css/50_misc.css
+++ b/css/50_misc.css
@@ -143,7 +143,7 @@ path.line.stroke.tag-roller_coaster-support {
     stroke: #707070;
 }
 path.line.casing.tag-roller_coaster-support {
-    stroke: #707070;
+    visibility: hidden;
 }
 
 
--- a/css/50_misc.css
+++ b/css/50_misc.css
@@ -6,6 +6,8 @@ path.line.shadow.tag-attraction-water_slide,
 path.line.shadow.tag-golf-cartpath,
 path.line.shadow.tag-man_made-pipeline,
 path.line.shadow.tag-natural-tree_row,
+path.line.shadow.tag-roller_coaster-track,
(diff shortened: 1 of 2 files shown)


paganella31:
At this point in time both Carto and OsmAnd do render roller coasters, so it would be good to merge this PR.
Regarding the support beams, indeed they seem slightly thick to me. I would personally make them a little smaller.

tyrasd (approved):

Issue #9884, opened 2023-09-04
Mapilio zoom-in & move bug


sun-geo:
### URL

https://www.openstreetmap.org/edit#map=20/53.45913/-6.26780

### How to reproduce the issue?

Thanks for the implementation of the new Mapilio service!

But there is a bug:

Zoom-in and afterwards try to move the new image-area all the way to the right hand side, which isn't possible.

Seems this is the same issue like with Kartaview images, see [#8997](https://github.com/openstreetmap/iD/issues/8997) which also seems not fixed yet completely.

For mapillary images it works well.

Btw. anybody else can confirm this behavior?

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

current iD v2.27.1

### Which browsers are you seeing this problem on?

Firefox

arch0345:
I'm also able to reproduce this on chrome

tordans:
Mapilio was added in https://github.com/openstreetmap/iD/pull/9664 by @sezerbozbiyik. @sezerbozbiyik  are you able to help with this issue? (It is from last year, so this might have resolved itself in the meantime…)

matkoniecz:
> afterwards try to move the new image-area all the way to the right hand side, which isn't possible.

how can I do this? It is not obvious to me that it is even a possible and how to achieve it

EDIT: it appears to be about moving with street imagery, not about moving entire street imagery subwindow

PR #9903, opened 2023-09-15
[Mapilio] Add username field

### LINKED ISSUE #10400: Mapilio: Add author attribution
### Description

Mapilio is missing attribution for the image's author. Other providers already have this.
Currently only the date and the image link are shown.

### Screenshots

image

sezerbozbiyik:
This PR enables adding the username to Mapilio images.

image


modules/services/mapilio.js
```diff
@@ -348,25 +362,39 @@ export default {
         if (!d) return this;
 
         let wrap = context.container().select('.photoviewer .mapilio-wrapper');
-        let attribution = wrap.selectAll('.photo-attribution').text('');
-
-        if (d.capture_time) {
+        let attribution = wrap.selectAll('.photo-attribution').text('\u00A0');
+      
+        getUserData(d.created_by_id).then((username) => {
```
k-yle, line 368: If the username fails to load, everything else in the footer will be hidden too. 

I think this could be optimised by appending the DOM element during the first render, but leaving the element empty until the API response is received.

css/60_photos.css
```diff
@@ -397,6 +399,13 @@ label.panoramax-hd {
 }
 
 
+#ideditor-viewer-mapilio-simple img {
```
k-yle: this block will duplicate the styles added in #9849, so it should probably be removed

modules/services/mapilio.js
```diff
@@ -348,25 +362,39 @@ export default {
...
+          if (username) {
             attribution
-                .append('span')
-                .attr('class', 'captured_at')
-                .text(localeDateString(d.capture_time));
-
+              .append('a')
+              .attr('class', 'captured_by')
+              .attr('target', '_blank')
+              .text('@' + username);
```
k-yle, line 373: this is an `<a>` element, but it has no `href`. Is there a way to link to a mapillio user profile? If not, we should remove this link

k-yle (dismissed):
thanks for the PR! I've left some suggestions below

modules/services/mapilio.js
ozdmreren: I removed the a tag because there is no user profile in mapilio.

css/60_photos.css
ozdmreren: I have removed that block.

modules/services/mapilio.js
ozdmreren: It's been corrected.

modules/services/mapilio.js
k-yle: thanks!

k-yle:
20260105220623_rec_

I tried to test this with [the PR preview (here)](https://pr-9903--ideditor.netlify.app/#disable_features=boundaries&map=3.20/47.59/-3.85&background=MAPNIK&photo_overlay=mapilio), but due to #9884 the integration seems to be very broken and I can't zoom in.

However, the code changes itself seems okay to me. Should we go ahead and merge this?

matkoniecz:
screen-2026-03-06-20-04-25

css/60_photos.css
```diff
@@ -413,6 +415,7 @@ label.panoramax-hd {
     display: block !important;
 }
 
+
```
matkoniecz, line 418: this ideally would not be randomly changed, but lets not expect PR author to fix it given how long this PR was rotting

matkoniecz:
Also, I see

```
  367:1   warning  Trailing spaces not allowed  no-trailing-spaces
  374:1   warning  Trailing spaces not allowed  no-trailing-spaces
  379:1   warning  Trailing spaces not allowed  no-trailing-spaces
  380:21  warning  Missing space before =>      arrow-spacing
  380:24  warning  Missing space after =>       arrow-spacing
  386:1   warning  Trailing spaces not allowed  no-trailing-spaces
  391:1   warning  Trailing spaces not allowed  no-trailing-spaces

✖ 7 problems (0 errors, 7 warnings)
  0 errors and 7 warnings potentially fixable with the `--fix` option.
```

matkoniecz:
Maybe I should take PR as base, make this cosmetic changes and open a new one? Replacing this one?

PR #9906, opened 2023-09-19
Fix regression in user account display (#8813)

jleedev:


pnorman:
A test would be good, given this has regressed multiple times.

tyrasd:
a safeguard test for this is now added in fdcf6369f

PR #9912, opened 2023-09-21
Hide horizontal scrollbars in hover mode for fields

### LINKED ISSUE #10551: Hover inspector UI overflow
### URL

https://www.openstreetmap.org/edit#map=18/48.129455/7.827063

Hover over any element.

I found this was caused by `overflow-y: auto;` introduce in #9074 to fix #9070.

current `overflow-y: auto` vs `overflow-y: visible`
overflow no overflow

I'm not sure what is the best way to fix both, maybe only add scrollbar to notes like this?
```css
.inspector-hover div {
    overflow-x: visible;
    overflow-y: visible;
}
.inspector-hover .comment div {
    overflow-y: auto;
}
```

2.30.4

Chrome, Firefox

Dimitar5555:
Likely resolves #10551 

Horizontal scrollbars in fields are useless since you can't hover over the element and move a specific scrollbar at the same time.

Before:
image
After:
image

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4137,7 +4137,7 @@ li.issue-fix-item button:not(.actionable) .fix-icon {
 
 /* scrollbars only when necessary*/
 .inspector-hover div {
-    overflow-x: visible;
+    overflow-x: hidden;
     overflow-y: auto;
 }
 


Nekzuris:
This doesn't take care of the border.
image

PR #9914, opened 2023-09-22
Added warning message if not enough features are selected

AviralSingh-code:
Fixes #9896

PR #9917, opened 2023-09-25
Collapse sidebar narrow screen

### LINKED ISSUE #9815: Collapse sidebar on narrow screens
### Description

When opening iD in a device with a narrow screen, the sidebar appears automatically even though it is empty, occupying precious real estate:

Screen Shot 2023-08-05 at 02 08 28

IMO it should automatically be loaded in the collapsed state when the container is narrow:

Screen Shot 2023-08-05 at 02 10 13

As an example of existing behavior that matches the expectation, the toolbar buttons already  adopt a more compact display in narrower screens by omitting the labels and showing just the icons, which IMO is sensible behavior:

| Wide screen | Narrow screen |
| ------ | ----- |
| image | image |

I would gladly contribute a PR making the sidebar also auto-collapse in narrow screens, if there's agreement that this is desirable, and if I can get some guidance on how to do it — I suppose one needs to do something like, say `selection.classed('collapsed', containerWidth < 600);` (or whatever pixel width we consider narrow enough) somewhere when [initializing the sidebar](https://github.com/openstreetmap/iD/blob/5a14edfe368e14e7adf359d036ce5fe3fb1140be/modules/ui/sidebar.js#L38)?

AviralSingh-code:
Fixes #9815

### DIFF
--- a/modules/ui/sidebar.js
+++ b/modules/ui/sidebar.js
@@ -57,9 +57,9 @@ export function uiSidebar(context) {
         var flagCollapse = false;
         var flagExpand = false;
 
-        var holdWindowSize = 0;
+        var previousWindowSize = 0;
 
-        d3_select(window).on('mouseover', function(){checkWindowWidth()}); //event listener
+        d3_select(window).on('resize', function(){checkWindowWidth()}); //event listener
 
         function pointerdown(d3_event) {
             if (downPointerId) return;
@@ -425,32 +425,31 @@ export function uiSidebar(context) {
             the side bar when the window width is below 700 then it will automatically collapse the side bar
             which is not what we want.
         */ 
-            function checkWindowWidth()
+        function checkWindowWidth()
+        {
+            containerWidth = container.node().getBoundingClientRect().width;
+
+            /*
+                Reset `flagCollapse` every time we resize the window
+                otherwise the behavior would only apply the first time iD is loaded.
+            */ 
+            if(containerWidth != previousWindowSize)
             {
-                containerWidth = container.node().getBoundingClientRect().width;
-    
-                /*
-                    this is to reset the flag every time we resize the window
-                    so that the functionality offered by the function is not just for the first time the
-                    iD is loaded.
-                */ 
-                if(containerWidth != holdWindowSize)
-                {
-                    holdWindowSize = containerWidth;
-                    flagCollapse = false;
-                    flagExpand = false;
-                }
-                if(containerWidth < 700 && !flagCollapse)
-                {
-                    flagCollapse = true;
-                    sidebar.collapse();
-                }
-                else if(containerWidth >= 700 && !flagExpand)
-                {
-                    flagExpand = true;
-                    sidebar.expand();
-                }
+                previousWindowSize = containerWidth;
+                flagCollapse = false;
+                flagExpand = false;
             }
+            if(containerWidth < 750 && !flagCollapse)
+            {
+                flagCollapse = true;
+                sidebar.collapse();
+            }
+            else if(containerWidth >= 750 && !flagExpand)
+            {
+                flagExpand = true;
+                sidebar.expand();
+            }
+        }
     
         // toggle the sidebar collapse when double-clicking the resizer
         resizer.on('dblclick', function(d3_event) {
--- a/modules/ui/sidebar.js
+++ b/modules/ui/sidebar.js
@@ -54,6 +54,12 @@ export function uiSidebar(context) {
             .on(_pointerPrefix + 'down.sidebar-resizer', pointerdown);
 
         var downPointerId, lastClientX, containerLocGetter;
+        var flagCollapse = false;
+        var flagExpand = false;
+
+        var holdWindowSize = 0;
+
+        d3_select(window).on('mouseover', function(){checkWindowWidth()}); //event listener
 
         function pointerdown(d3_event) {
             if (downPointerId) return;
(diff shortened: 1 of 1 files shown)


waldyrious:
It looks like this PR contains a commit from #9914, so either that PR should be merged before this one, or this PR's branch should be rebased to start off from `develop`.

AviralSingh-code:
Ya I will do the needful.

AviralSingh-code:
Is it ok now ??

modules/ui/sidebar.js
```diff
@@ -412,6 +418,40 @@ export function uiSidebar(context) {
...
+            which is not what we want.
+        */ 
+            function checkWindowWidth()
+            {
+                containerWidth = container.node().getBoundingClientRect().width;
+    
+                /*
+                    this is to reset the flag every time we resize the window
+                    so that the functionality offered by the function is not just for the first time the
+                    iD is loaded.
```
waldyrious:
```suggestion
                    Reset `flagCollapse` every time we resize the window
                    otherwise the behavior would only apply the first time iD is loaded.
```

modules/ui/sidebar.js
```diff
@@ -412,6 +418,40 @@ export function uiSidebar(context) {
...
         };
 
+        /*
+            The checkWindowWidth function checks the width of the working window when the mouse is over the
+            working area and flagCollapse and flagExpand is to prevent the re-triggering action after the window
+            is resized if the window is resized and the flags are not used then whenever we will try to expand
+            the side bar when the window width is below 700 then it will automatically collapse the side bar
+            which is not what we want.
+        */ 
+            function checkWindowWidth()
```
waldyrious: Why is this entire function indented in relation to the comment above (and surrounding code)? IIUC it should have 4 fewer spaces of indentation.

modules/ui/sidebar.js
```diff
@@ -54,6 +54,12 @@ export function uiSidebar(context) {
             .on(_pointerPrefix + 'down.sidebar-resizer', pointerdown);
 
         var downPointerId, lastClientX, containerLocGetter;
+        var flagCollapse = false;
+        var flagExpand = false;
+
+        var holdWindowSize = 0;
```
waldyrious: Perhaps a different name would make the logic clearer:

```suggestion
        var previousWindowSize = 0;
```

modules/ui/sidebar.js
```diff
@@ -54,6 +54,12 @@ export function uiSidebar(context) {
...
+
+        d3_select(window).on('mouseover', function(){checkWindowWidth()}); //event listener
```
waldyrious: Why not trigger this in the [resize](https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event) event?

waldyrious:
@AviralSingh-code I managed to test this branch locally by following the [instructions in the README](https://github.com/openstreetmap/iD#building-id). It seems to work as expected! :tada: however, I'd like to try out the behavior of the sidebar when objects are selected/deselected, but my locally running instance doesn't seem to load any geometry or background imagery. Is this the case for you as well? If not, do you have any hints as to how to get it to work locally?

Edit: I've [asked](https://community.openstreetmap.org/t/how-to-run-id-locally-with-background-imagery-and-osm-geometry/104279) in the OSM Discourse forum, let's see if anyone has hints there.

waldyrious:
Also, you probably should run `npm run lint:fix` to correct some code style issues with the way the code is currently formatted.

AviralSingh-code:
Can you elaborate on the problem a bit more like the issue that you are facing while trying out the behavior of the of the sidebar  when objects are selected/deselected because for me, everything seems fine on my local machine !! But maybe I am missing on to the details so if you could elaborate more on this.

AviralSingh-code:
And yes there is one more question that the PR is showing the message "2 workflows awaiting approval", so do you have any idea on how to resolve that ?

waldyrious:
> Can you elaborate on the problem a bit more like the issue that you are facing while trying out the behavior of the of the sidebar when objects are selected/deselected because for me, everything seems fine on my local machine !! 

Cool, glad to hear it works well locally for you. In my case I can't see any OSM geometry when I run iD locally (with `npm start`), nor any background (satellite) imagery. All I see is black in the area of the editor where the map should appear. See the screenshot I added to my post in the Discourse forum linked above:



Do you see something different when you do `npm start`? Or perhaps you are using a different command?

> And yes there is one more question that the PR is showing the message "2 workflows awaiting approval", so do you have any idea on how to resolve that ?

That's because you haven't had any contributions merged into this repository yet. So for now, you'll need a maintainer to explicitly trigger the checks. Once one of your PRs is merged, actions will start triggering automatically in further PRs.

AviralSingh-code:
The first time I started the environment I had to do :
1. npm install
2. npm run all
3. npm start
Then whenever I made changes I always did :
1. npm run all
2. npm start

So I think maybe you could try to run npm run all before npm start. I am not sure whether that would help but no harm in trying !!

waldyrious:
Thanks for the suggestion, but running `npm run all` before `npm start` does not make a difference: I still see the iD interface, and it is functional, but nothing appears in the actual map area.

AviralSingh-code:
What I feel is that you should try to reset your code with the upstream because I think that you might have altered the code that could have broken down the code because sometimes just to check the flow of the code I used to comment out the sidebar element or other elements and then the iD used to behave similar to what you are facing !!

waldyrious:
OMG, sorry, the issue was on my side — it looks like running iD locally doesn't work on Firefox. I can get the site to work normally if I use a Chromium-based browser, like Vivaldi or Brave. Apologies for the noise! :sweat_smile:

So, getting back on track: it seems like resizing the browser window to trigger the sidebar collapse makes the background imagery shift. Here's the current behavior in the develop branch:

[Screencast from 02-10-2023 22:26:44.webm](https://github.com/openstreetmap/iD/assets/478237/45dd7a6c-8cf1-4c6a-a00c-4c2fceab69fc)

And here is the behavior with this PR's branch:

[Screencast from 02-10-2023 22:43:42.webm](https://github.com/openstreetmap/iD/assets/478237/e9f0afef-8bd8-4cdd-ab0a-065130adda93)

Apologies for the frame skips — my graphics card isn't the best. But do try resizing the browser window yourself and you should notice the issue. IMO the background should remain static, i.e. "anchored" to the right edge of the window, rather than to the border of the sidebar.

AviralSingh-code:
Do you have any suggestions on how should I try that ? Like what will be the best fix for this because I haven't added any code that affects the frame of the window but then also this shift is happening !!

waldyrious:
Yeah, indeed this behavior does not seem like it was introduced by your code, but it is there nevertheless. Unfortunately I don't know enough about this codebase to guide you towards what changes need to be done to avoid the issue, but hopefully once a maintainer of this project reviews this PR they will be able to point out the needed changes. In any case, that is a relatively minor issue since it's not expectable that people would be manually resizing their browsers frequently.

I _did_ notice a separate issue, though: if you resize the browser to a small enough width that the sidebar is collapsed, then refreshing the page loads iD with the sidebar expanded. It seems like we need to perform the check not just in the resize event but also in the initial load event.

AviralSingh-code:
Yes I will fix the initial load of the sidebar also !!

tyrasd:
closing because this behavior could be confusing for new users. see https://github.com/openstreetmap/iD/issues/9815#issuecomment-2934144750 for further details.

PR #9931, opened 2023-10-07
disable the locator overlay when switching to osm carto

k-yle:
This PR automatically disables `Locator Overlay` when you select `OpenStreetMap (Standard)`.

If you use the `OpenStreetMap (Standard)` background imagery, the first thing you have to do after selecting `OpenStreetMap (Standard)` is disable `Locator Overlay`, because the two layers create an unreadable mess of labels:

image

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -337,6 +337,18 @@ export function rendererBackground(context) {
     baseLayer.source(!fail ? d : background.findSource('none'));
     dispatch.call('change');
     background.updateImagery();
+
+    // if switching to a background layer that contains labels,
+    // disable the locator overlay to avoid overlapping labels.
+    if (d.id === 'MAPNIK') {
+      const locatorOverlay = background.overlayLayerSources().find(
+        overlay => overlay.isLocatorOverlay()
+      );
+      const isLocatorOverlayActive = !!locatorOverlay;
+      if (isLocatorOverlayActive) {
+        setTimeout(() => background.toggleOverlayLayer(locatorOverlay), 0);
+      }
+    }
     return background;
   };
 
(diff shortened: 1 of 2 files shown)


tordans:
Does it enable again automatically, when switching back?

k-yle:
At the moment, no, it just stays off. Do you think users would expect it to automatiaclly re-enable?

tordans:
> Do you think users would expect it to automatiaclly re-enable?

I think yes. My thinking is: I click something and the system does something else automatically. I don't know that it did this thing. So once I go back to the regular map, something is missing and I don't know why or how to add it again…

matkoniecz:
@k-yle Do you plan to somehow prevent user accidentally losing overlay without having an idea how to get it back?

tyrasd:
I would agree with @tordans and @matkoniecz: it would be confusing if the locator overlay would be completely missing after selecting the `mapnik` layer once…

Maybe the code could be enhanced to just skip rendering the locator overlay if a background layer that is of [category `map`](https://github.com/osmlab/editor-layer-index/blob/gh-pages/CONTRIBUTING.md#categories) is currently rendered (and keep the overlay selection untouched, so that it is displayed again after switching back to a aerial imagery)?

tyrasd (changes requested):
(see comment above)

k-yle:
done, now the state is restored when you switch back to a different imagery source

matkoniecz (changes requested):
I tested it, and with locator overlay and Bing as background layer - locator overlay is not appearing

I have not looked at code (at least for now)

k-yle:
> I tested it, and with locator overlay and Bing as background layer - locator overlay is not appearing

the locator overlay includes an API key which is restricted to [openstreetmap.org](https://openstreetmap.org), so it can't be tested from the PR preview nor from localhost.

You'll either need to use a command like [`--disable-web-security`](https://groups.google.com/a/chromium.org/g/chromium-discuss/c/Wms-ZaAwA1Q), or test it based on the presence of the API calls in the devtools' network tab

PR #9935, opened 2023-10-08
Removed the escape sequence characters form the string

### LINKED ISSUE #9886: iD display double quote values with the escaping \ character in text mode
### URL

https://www.openstreetmap.org/edit?way=193112329#map=18/-22.91583/-43.62896

1. Some OSM elements have double quote in their names. Consider this way: https://www.openstreetmap.org/way/193112329. It's `old_name` value is: `Viaduto José Ramos de Oliveira "Zé Cocada"`. The way this data is retrieved/saved from/to OSM database, the double quotes gets "escaped" and thus the character `"` in the middle of strings is stored as `\"`.
2. On list mode, the double quote character is shown correctly:
image
3. However on text mode, the double quote character is shown with the escaping character:
image
4. Since `key=value` pairs are not quoted in this text list (i.e. they are not `key="value"`) I think it does not make sense to display the double quotes with `\`.

_No response_

2.27.1

Chrome

AviralSingh-code:
Fixes #9886

### DIFF
--- a/modules/ui/sections/raw_tag_editor.js
+++ b/modules/ui/sections/raw_tag_editor.js
@@ -133,6 +133,9 @@ export function uiSectionRawTagEditor(id, context) {
         var textarea = wrap.selectAll('.tag-text')
             .data([0]);
 
+        //this removes the escape sequence from the string
+        textData = textData.replace(/\\/g, '');
+
         textarea = textarea.enter()
             .append('textarea')
             .attr('class', 'tag-text' + (_tagView !== 'text' ? ' hide' : ''))


k-yle (commented):
Hey, unfortunately this doesn’t quite solve the problem.

Internally, the raw editor represents key-value pairs as values wrapped in quotation marks (`"`), but these quotation marks are not shown to the user.

For example, if you see:

```yaml
my key=my value

another key=a different value \" with a quotation mark
```

then the internal representation is:

```scala
"my key"="my value"

"another key"="a different value \" with a quotation mark"
```

If you simply remove the backslash, the second example will be parsed as:

```json
"another key"="a different value " with a quotation mark"
```

which will cause everything after the first quotation mark to be silently deleted.

There is no straightforward solution to this issue, because we will always have to escape either `"` or `=`.

Maybe the real solution would be to encourage the use of [smart quotes](https://smartquotesforsmartpeople.com) in tag values (?), since it’s easier for software to normalise smart quotes than the reverse process. Then there would relatively few examples of `"` in the database.

I assume languages that use other quotation marks (e.g. `« »`) already use their own language's typography, rather than straight quotes. [exa](https://osm.org/way/976768177)[mples](https://osm.org/way/462453685)

tyrasd:
closing as stale

PR #9938, opened 2023-10-12
migrate to ELI's modern json format

k-yle:
In a fork of iD I was trying to access the [`category` attribute](https://github.com/osmlab/editor-layer-index/blob/ba0c4ff764f587e49419205ef5526e63cdb47939/schema.json#L46-L59) from https://github.com/osmlab/editor-layer-index, but that attribute is not available because we uses the "legacy" json format from ELI. This PR updates iD to use ELI's geojson format, which contains an extended set of fields.

Test plan:
 - run `npm run imagery`
 - check that there is no change to `data/imagery.json` apart from the new [`category`](https://github.com/osmlab/editor-layer-index/blob/ba0c4ff764f587e49419205ef5526e63cdb47939/schema.json#L46-L59) attribute

### DIFF
--- a/scripts/update_imagery.js
+++ b/scripts/update_imagery.js
@@ -1,14 +1,31 @@
 /* eslint-disable no-console */
 const fs = require('fs');
-let sources = require('editor-layer-index/imagery.json');
 const prettyStringify = require('json-stringify-pretty-compact');
 
+/** @type {import("geojson").FeatureCollection} */
+const sources = JSON.parse(
+  fs.readFileSync(require.resolve('editor-layer-index/imagery.geojson'), 'utf8')
+);
+
 if (fs.existsSync('./data/manual_imagery.json')) {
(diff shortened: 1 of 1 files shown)


scripts/update_imagery.js
```diff
@@ -1,14 +1,31 @@
 /* eslint-disable no-console */
 const fs = require('fs');
-let sources = require('editor-layer-index/imagery.json');
 const prettyStringify = require('json-stringify-pretty-compact');
 
+/** @type {import("geojson").FeatureCollection} */
+const sources = JSON.parse(
+  fs.readFileSync(require.resolve('editor-layer-index/imagery.geojson'), 'utf8')
```
k-yle, line 7: I had to change `require` to `JSON.parse(fs.readFile(...))` because ELI uses the file extension `.geojson` instead of `.geo.json`, so it can't be imported

PR #9945, opened 2023-10-19
Add eox.at 2022 to discard list in update_imagery script

kmpoppe:
Following from #9807, this adds the newly published 2022 eox cloudless imagery to the discard list (osmlab/editor-layer-index#2112)


arch0345:
It might be worth using regex for the discard list like Rapid so that new eox.at ids won't have to be added every year.
https://github.com/facebook/Rapid/blob/f9ff458c96d3857ad90be4f864739c175dda7af2/scripts/update_imagery.js#L28-L56

I'll open a separate issue for this

PR #9949, opened 2023-10-23
Warn when toll_gantry on highway without toll=yes #9885 

### LINKED ISSUE #9885: Warn when toll_gantry on highway without toll=yes and toll:*=yes
### Description

From the [OSM wiki on toll_gantry](https://wiki.openstreetmap.org/wiki/Tag%3Ahighway%3Dtoll_gantry):
"The way that the node is placed on must also be tagged with [toll](https://wiki.openstreetmap.org/wiki/Key:toll)=yes on any applicable road segments."

Could iD issue a warning whenever this is not the case?

### Screenshots

_No response_

gitgutgit:
Warn when toll_gantry on highway without toll=yes #9885
https://github.com/openstreetmap/iD/issues/9885

 This feature ensures that if there's a gantry on a highway, the toll attribute should be set to "yes", otherwise, a warning is given.

image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1903,6 +1903,9 @@ en:
     unknown_road:
       message: "{feature} has no classification"
       reference: "Roads without a specific type may not appear in maps or routing."
+    no_toll:
+      message: "Warn when toll_gantry on {feature} without toll=yes."
+      reference: "Ensure that highways with toll_gantry are properly tagged with toll=yes for accurate mapping and routing."
     impossible_oneway:
       title: Impossible One-Ways
       tip: "Find route issues with one-way features"
--- a/modules/validations/missing_tag.js
+++ b/modules/validations/missing_tag.js
@@ -42,7 +42,7 @@ export function validationMissingTag(context) {
         return entity.type === 'relation' && !entity.tags.type;
     }
 
-    function isWayWithTollNoConnectedToGantry(way, context) {
+    function isWayWithNoTollConnectedToGantry(way, context) {
         if (way.type !== 'way' || way.tags.toll !== 'no') return false;
     
         var childNodes = context.graph().childNodes(way);
@@ -72,8 +72,8 @@ export function validationMissingTag(context) {
             }
         }
 
-        if (isWayWithTollNoConnectedToGantry(entity, context)) {
-            subtype = 'toll_gantry_without_toll';
+        if (isWayWithNoTollConnectedToGantry(entity, context)) {
+            subtype = 'way_with_no_toll_connected_to_gantry';
         }
         
         // flag an unknown road even if it's a member of a relation
@@ -85,12 +85,12 @@ export function validationMissingTag(context) {
 
         if (!subtype) return [];
 
-        var messageID = subtype === 'highway_classification' ? 'unknown_road' : 'missing_tag.' + subtype;
-        var referenceID = subtype === 'highway_classification' ? 'unknown_road' : 'missing_tag';
+        var messageID = subtype === 'highway_classification' ? 'unknown_road' : subtype === 'way_with_no_toll_connected_to_gantry'? 'no_toll' : 'missing_tag.' + subtype;
+        var referenceID = subtype === 'highway_classification' ? 'unknown_road'  : subtype === 'way_with_no_toll_connected_to_gantry' ? 'no_toll' : 'missing_tag';
 
         // can always delete if the user created it in the first place..
         var canDelete = (entity.version === undefined || entity.v !== undefined);
-        var severity = (canDelete && subtype !== 'highway_classification') ? 'error' : 'warning';
+        var severity = (canDelete && (subtype !== 'highway_classification' && subtype !== 'way_with_no_toll_connected_to_gantry')) ? 'error' : 'warning';
 
         return [new validationIssue({
             type: type,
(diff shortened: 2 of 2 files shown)


matkoniecz (commented):
what about cases where toll gantries collect toll only from say HGV or only from buses?

in such case toll=no [toll:hgv](https://wiki.openstreetmap.org/wiki/Key:toll:hgv?uselang=en)=yes may be entirely correct tagging

and missing toll and toll:hgv=yes is even more likely to be valid

(Poland is one of cases where this applies)

data/core.yaml
```diff
@@ -1946,6 +1903,9 @@ en:
     unknown_road:
       message: "{feature} has no classification"
       reference: "Roads without a specific type may not appear in maps or routing."
+    no_toll:
+      message: "Warn when toll_gantry on {feature} without toll=yes."
```
matkoniecz, line 1907: this message also should be rewritten

not a native speaker but maybe these would be better?

```suggestion
      message: "Toll gantry on {feature} without toll=yes."
```

KronosTheLate:
As an editor, ID can afford to be biased. I surely hope that this PR does not get bogged down for another year or two on the corner-case of HGV vehicles, when it is very usefull for normal people as it is.

With that said, what about this: if a node is tagged as a toll on a way that is not, we display the warning as proposed, but add a line along the lines of "Note that some tolls apply only to specific vehicles. Please verify the vehicle types". As it only displays when the warning is active, it should not really be that obtrusive in general.

matkoniecz:
> As an editor, ID can afford to be biased. I surely hope that this PR does not get bogged down for another year or two on the corner-case of HGV vehicles, when it is very usefull for normal people as it is.

the problem is that in the current state it will emit useless warnings for every single major road in Poland, and maybe elsewhere, making iD validator utterly useless there

in the current form merging is not viable at all

mass scale false positives from validator are not OK

KronosTheLate:
~Oh! I just though we were talking about a message whilst editing the node. Yes, agree this is a major problem for large scale validation~

See my new comment below.

KronosTheLate:
From https://wiki.openstreetmap.org/wiki/Key:toll:

> When the general condition is that vehicles are not charged then use toll=no together with over-rides for classes that are charged. For example: toll=no, toll:hgv=yes

So there we have an authoratative source specifying how it should be done in Poland.

Also, as stated in the original issue #9885:
> From the [OSM wiki on toll_gantry](https://wiki.openstreetmap.org/wiki/Tag%3Ahighway%3Dtoll_gantry):
"The way that the node is placed on must also be tagged with [toll](https://wiki.openstreetmap.org/wiki/Key:toll)=yes on any applicable road segments."

So toll gantry nodes should indeed have a corresponding toll-tag on the way on which they are placed. This is what this PR warns about when it is not the case, as far as I understand it.

If this results in a warning on every major Polish road, I believe that just means that every major Polish road is incorrectly mapped. If we can agree upon that, then this fact is an argument _for_ the usefullness of this warning, not against.

matkoniecz:
I would rather see it as indicator that second article has incomplete advise.

And if it is correct, then how this toll gantries are supposed to be tagged?

I opened https://wiki.openstreetmap.org/wiki/Talk:Tag:highway%3Dtoll_gantry#HGV_tolls

If you are convinced this wiki advise makes sense, feel free to open thread at https://community.openstreetmap.org/c/communities/pl/40 with suggestion how to change tagging

matkoniecz:
I think this PR should be closed without merge given its current state

KronosTheLate:
I strongly disagree that this should be closed based on the argumentation above. If a road segment has a toll booth/gantry, then it makes perfect sense that the way should have a toll=yes tag. Just because it does not apply to most vehicles, that does not change the fact that it is a toll road. Then, the tags should make clear which vehicles it applies to.

If the road has a toll booth/gantry which does not apply to most vehicles, then the appropriate tagging is toll=no with appropriate overrides for different types of vehicles, as I quoted in my previous comment. This PR could make it so that toll=no + toll:*=yes is a viable alternative to toll=yes for ways that have a toll booth/gantry for "killing the warning".

I honestly feel like @matkoniecz is looking to change the wiki page and this warning, which both make sense in most countries (I believe), just because it is less convenient for Poland. I do believe that we should just tag Poland's roads with toll=no and toll:hgv=yes where applicable, which is something that the editor warning implemented in this PR would greatly help with. 

I mean, I hope we can all agree that if there is a toll for HGV vehicles, then tagging it with toll=no and toll:hgv=yes is more representative of the underlying truth than no toll tags at all, right??

matkoniecz:
> then the appropriate tagging is toll=no with appropriate overrides for different types of vehicles

I agree, that is why I think that "Warn when toll_gantry on highway without toll=yes" would be giving bad warnings

> This PR could make it so that toll=no + toll:*=yes is a viable alternative to toll=yes for ways that have a toll booth/gantry for "killing the warning".

It is possible that it would work. As it has not happened and it does not look likely it will happen, I think this PR should be closed.

If someone is interested in making working one, maybe taking these commits as a base it would be welcome.

Note that closing abandoned PR to do X is not the same as rejection of doing X.

KronosTheLate:
Okay okay so it seems there is a solution we can agree on: Provide a warning if there is a way with a toll_gantry on a way without toll=yes OR without toll=no and toll:*=yes.

So the only change to this PR would be adding the OR clause to the logic, so adding that to this PR makes the most sense to me. But implementing this in a new PR also achieves the same goal, so that is ofc fine too.

tyrasd:
I think it is fine to leave this QA up to external QA tools like osmose.

Closing this PR as stale.

KronosTheLate:
Turns out the osmose already has an issue for this: https://github.com/osmose-qa/osmose-backend/issues/2520. Agreed that this validation can be done over there.

PR #9950, opened 2023-10-23
Reduced map pane width and added a sidebar to it 

### LINKED ISSUE #9872: A slimmer, movable Background Pane
### Description

For users that like switching back and forth between backgrounds a lot, the Background Panel takes a lot of screen real estate.

We notice:
- The area I marked in green could be trimmed off, thus only needing to be as wide as there are letters.

And:
- The whole panel could be allowed to be slid around by the user, so they could still at least just click the buttons.
(but then they would also need the different parts to show up: 2022, 2021... , not just Cook, Cook...)

### Screenshots

m

navyn13:
Fixes #9872 

- [x] div. Reduced the width of map pane for better view of map while editing.
- [x] Made a function to add sidebar of any div.

https://github.com/openstreetmap/iD/assets/121725696/34051449-fef3-4488-a584-9e37a72447bf


modules/ui/pane_features.js
```diff
@@ -0,0 +1,55 @@
...
+    let originalWidth = getOriginalWidth(resizeDiv);
+    if (isResizingAllowed(originalWidth, e.clientX)) {
+        isResizing = true;
+
+        if(e.stopPropagation) e.stopPropagation();
+            if(e.preventDefault) e.preventDefault();
+            e.cancelBubble=true;
+            e.returnValue=false;
+
+        document.addEventListener('mousemove', (evnt) => {
```
lefuturiste, line 19: evnt -> event

modules/ui/pane_features.js
```diff
@@ -0,0 +1,55 @@
...
+export function isResizingAllowed(originalWidth, clientX) {
+    return (clientX > window.innerWidth - originalWidth  && clientX < window.innerWidth - originalWidth+10);
+}
+ 
+export function startResizing(resizeDiv, e) {
+    let originalWidth = getOriginalWidth(resizeDiv);
+    if (isResizingAllowed(originalWidth, e.clientX)) {
+        isResizing = true;
+
+        if(e.stopPropagation) e.stopPropagation();
```
lefuturiste, line 14: lint: missing space between if and start of the condition (

modules/ui/pane_features.js
lefuturiste, line 1: New file: so you probably want to add a short documentation comment at the top on what's this file is about.

modules/ui/pane_features.js
```diff
@@ -0,0 +1,55 @@
...
+    const originalWidth = getOriginalWidth(resizeDiv);
+    if (e.clientX > window.innerWidth - originalWidth && e.clientX < window.innerWidth - originalWidth + 6) {
+        resizeDiv.style.cursor = 'ew-resize';
+    } else {
+        resizeDiv.style.cursor = 'auto';
+    }
+}
+
+export function initializeResizePane(resizeDiv) {
+
```
lefuturiste, line 42: too many blank line

lefuturiste (changes requested):
I just did a review of the code lint. But the logic seem fine to me. You probably want to add a small integration test case for this (it's a new feature, so I think asking for a test case is not far fetched).

PR #9952, opened 2023-10-26
 enhanced field form=surface

### LINKED ISSUE #9898: Autocomplete changes its mind after 0.3s
**Reproduce:** Type "g", "gr", or "gra" in the `surface=` field

**Video:** https://github.com/openstreetmap/iD/assets/5626710/3c398fcb-a0c6-4c1b-8e3f-e35b5c4c9899

**User story:**

Earlier, I typed "gr" and saw it already completed to grass, so I tabbed out of the field, only for it to last-second change to ground!

Now at the current road, I was aware of the problem so I typed "gra", see it complete to grass, and tab out of the field :)  ...... wait .... gravel?! That's not what I selected!

Someone on Telegram suggested arrow keys: I'm using my left hand to type (qwerty) and right for the mouse, so arrow keys are far out of reach. I like the autocomplete a lot but please don't show it before it has made up its mind about what value it'll use! :D

**iD version:** Released version at openstreetmap.org/edit

**Browser:** Firefox 116.0.3 on Linux

navyn13:
Fixes #9898 
Before: On entering surface type it was changing its suggestion in combobox.
This PR updates the [combobox](https://github.com/openstreetmap/iD/blob/develop/modules/ui/combobox.js) where _suggestions was populated at [change](https://github.com/openstreetmap/iD/blob/develop/modules/ui/combobox.js?plain=1#L231-L258) and [fetchComboData](https://github.com/openstreetmap/iD/blob/develop/modules/ui/combobox.js?plain=1#L319-L333). So changes are made so it doesn't get populated twice.

https://github.com/openstreetmap/iD/assets/121725696/0f75ed2f-ec7e-4267-a6af-13dfb312fb22

### DIFF
--- a/modules/ui/combobox.js
+++ b/modules/ui/combobox.js
@@ -230,32 +230,32 @@ export function uiCombobox(context, klass) {
         // Called whenever the input value is changed (e.g. on typing)
         function change(doAutoComplete) {
             if (doAutoComplete === undefined) doAutoComplete = true;
-            fetchComboData(value(), function() {
-                _selected = null;
-                var val = input.property('value');
-
-                if (_suggestions.length) {
-                    if (doAutoComplete && input.property('selectionEnd') === val.length) {
-                        _selected = tryAutocomplete();
-                    }
-
-                    if (!_selected) {
-                        _selected = val;
-                    }
+        
+            _selected = null;
+            var val = input.property('value');
+        
+            if (_suggestions.length) {
+                if (doAutoComplete && input.property('selectionEnd') === val.length) {
+                    _selected = tryAutocomplete();
                 }
-
-                if (val.length) {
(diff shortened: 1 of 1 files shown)


tyrasd:
Hey. Thanks for the suggested fix. :relaxed:  However, I noticed that this solution to the issue did break the autocomplete feature on combo fields which don't have `options` available in the respective preset field. And it also prevents the fetching of additional options from taginfo for combo fields that do have such options.

I did come up with a solution which just suppresses the second autocomplete when the taginfo results come in (shortly) after the initial autocomplete has already been done using the static field options. see dabbd89e5

navyn13:
I am glad to know it helped you to solve this issue, I am still learning and I will come up with some better solutions😄

PR #9974, opened 2023-11-12
feat(fields/combo): add navigate button to web links in values

lefuturiste:
The common use case is for the `sources` key, often you have multiple URLs separated with semicolon. This allow to navigate to URL and check them more easily instead of doing manual work to select the URL copy and paste.

Created with the help of @tyrasd at the sotmeu2023!


lefuturiste:
may be we can stack the buttons?
image

Issue #9946, opened 2023-10-19
Use regex for discard list in imagery update script


arch0345:
### Description

Rapid currently uses regex for the discard list in the update imagery script:
https://github.com/facebook/Rapid/blob/f9ff458c96d3857ad90be4f864739c175dda7af2/scripts/update_imagery.js#L28-L56

As mentioned in https://github.com/openstreetmap/iD/pull/9945#issuecomment-1770225887, it might be worth implementing this so that, for example, new eox.at sources won't have to be added every year.

0xAnon0602:
Hey, I would like to work on this issue.
Can you assign this to me ?

0xAnon0602:
@tyrasd  @arch0345 
Hey, I would like to work on this issue.
Can you assign this to me ?

arch0345:
Unfortunately don't have perms to assign people, however you're more than welcome to work on this issue!

0xAnon0602:
@tyrasd 
Added a pull request 
Let me know if anything else needs to be added here

arch0345:
Closed via #9975

PR #9975, opened 2023-11-12
Regex added

0xAnon0602:
Added Regex code instead of hardcoding it each year as discussed  here https://github.com/openstreetmap/iD/issues/9946

### DIFF
--- a/scripts/update_imagery.js
+++ b/scripts/update_imagery.js
@@ -42,10 +42,10 @@ const discard = [
   /^openseamap$/,
   /^opensnowmap-overlay$/,
 
-  /^US-TIGER-Roads-201\d/,     
-  /^Waymarked_Trails/,         
-  /^OSM_Inspector/,           
-  /^EOXAT/                     
+  /^US-TIGER-Roads-201\d/,
+  /^Waymarked_Trails/,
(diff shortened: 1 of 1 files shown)


scripts/update_imagery.js
```diff
@@ -17,63 +17,36 @@ let imagery = [];
...
+  /^openinframap-(petroleum|power|telecoms)$/,
+  /^openpt_map$/,
+  /^openrailwaymap$/,
+  /^openseamap$/,
+  /^opensnowmap-overlay$/,
+
+  /^US-TIGER-Roads-201\d/,     
+  /^Waymarked_Trails/,         
+  /^OSM_Inspector/,           
+  /^EOXAT/
```
arch0345:
```suggestion
  /^US-TIGER-Roads-201\d/,
  /^Waymarked_Trails/,
  /^OSM_Inspector/,
  /^EOXAT/
```

arch0345 (approved):
Tested it locally and it seems to be working! Just noticed some warnings about trailing spaces

0xAnon0602:
I think this should fix that

tyrasd (approved):

PR #9979, opened 2023-11-13
Use preferred tile.openstreetmap.org URL

danieldegroot2:
I would like to suggest using the now preferred https://tile.openstreetmap.org URL instead of the current one, see

https://github.com/openstreetmap/operations/issues/737


arch0345:
FYI this just changes the example shown in Custom Background settings, which shows users how to use the supported TMS tokens:
image

All background layers, including OpenStreetMap Carto, are sourced from ELI:
https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/world/OpenStreetMap-Mapnik.geojson

tyrasd (approved):

tyrasd:
@arch0345 correct, but I think even the examples should follow the recommended syntax, and as dns multiplexing is becoming obsolete in the grand scheme of things, it is also not too important to show this feature in the example.

PR #9988, opened 2023-11-21
add a link next to foreign keys

### LINKED ISSUE #5838: Use URL formatter for tag values
There is a growing list of tags whose values represent links to a site, such as `wikipedia` or `contact:yelp`.  Some of them have special handling in the top section _all fields_ (wikipedia, wikidata), and as a result have a well defined link:

image

While other tags do not shown as links at all -- _all tags_ section:
image

We should add a small external link icon next to all linkable entries in the _all tags_ section, using the formatter URL (P8) property of the data item.  For example see [contact:yelp formatter URL](https://wiki.openstreetmap.org/wiki/Item:Q8244#P8) (set to `https://www.yelp.com/biz/$1`)

### LINKED ISSUE #9804: Automatic linking of Wikidata/Wikipedia entries
### Description

Essentially a duplicate of #5838. I'm trying to verify `operator:wikidata`/`operator:wikipedia`, and not being able to click the links makes the checking a lot slower. I instead open the normal OSM viewer for the object which links to Wikidata/Wikipedia for the relevant tags.

The URL generation could be obtained from the OpenStreetMap wiki, e.g. [Property P8 "formatter URL"](https://wiki.openstreetmap.org/wiki/Property:P8) of [Q99 "brand:wikidata"](https://wiki.openstreetmap.org/wiki/Item:Q99) is "http://www.wikidata.org/entity/$1".

### Screenshots

By @nyurik, in https://github.com/openstreetmap/iD/issues/5838#issuecomment-45989162:

### LINKED ISSUE #11120: Handle url->id mapping, id->url click linking for key:contact=*
### Description

Image

Image

### Screenshots

_No response_

k-yle:
There is a now a button next to foreign-key fields like `operator:wikidata`, `mapillary`, `gnis:feature_id`, which will open the feature in a new tab.

image

The URL templates are fetched from https://github.com/JOSM/tag2link, which gets its data from [`P1630`](https://www.wikidata.org/wiki/Property:P1630) on wikidata or [`P8`](https://osm.wiki/Property:P8) in the osm wiki.

Closes #5838, Closes #9804

### DIFF
--- a/modules/services/tag2Link.js
+++ b/modules/services/tag2Link.js
@@ -1,0 +1,25 @@
+// @ts-check
+import tag2linkRaw from 'tag2link';
+
+const RANKS = ['deprecated', 'normal', 'preferred'];
+
+/** @param {import('tag2link').Tag2Link[]} input */
+function convertSourceData(input) {
+  /** @type {Record<string, string>} */
+  const output = {};
+
+  const allKeys = new Set(input.map(item => item.key));
+
+  for (const key of allKeys) {
+    // find the item with the best rank
+    const bestDefinition = input
+      .filter(item => item.key === key)
+      .sort((a, b) => RANKS.indexOf(b.rank) - RANKS.indexOf(a.rank))[0];
+
+    output[key.replace('Key:', '')] = bestDefinition.url;
+  }
+
+  return output;
+}
+
+export const TAG2LINK = convertSourceData(tag2linkRaw);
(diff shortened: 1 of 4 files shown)


matkoniecz:
I am not convinced that raw tag table should get UI. Maybe these should stay in main user interface part?

Admittedly you still have one big text field for raw tag purists.

k-yle:
yeah, I think we can close this. The main OSM website [supports tag2link now](https://github.com/openstreetmap/openstreetmap-website/pull/6197), so it's only 1 extra click to open the feature on OSM.org, then click the foreign key.

PR #9990, opened 2023-11-21
show date/time for georeferenced photos

k-yle:
The new georeferenced photos feature is pretty cool, but the popup only shows the filename, which is often auto-generated by the camera and not very meaningful. 

This PR updates the popup to also show the date & time of the photo, which is a lot more helpful when you have similar photos.

image


tyrasd (approved):

tyrasd:
That's a very useful enhancement, thank you!

PR #9999, opened 2023-11-25
remove rarely-used keyboard shortcut L

### LINKED ISSUE #8474: Accidentally clicked on show my location, no way to return to previous location
I was editing a fair distance away from my home on the map, when I accidentally clicked on the "show my location" button (I was aiming for the background settings button). This jumped to my current location, but there was no easy way to get back to where I was editing. 

I hadn't edited anything at that point (otherwise I could click save then click on something I'd edited), so I basically had to press the back button in the browser then go back in and edit again (also easier than trying to search for my previous location).

My suggestions would be:

Add a confirmation step before jumping to the user's current location and/or 

Move the show my location button to somewhere that it won't be clicked accidentally.

### LINKED ISSUE #8534: Add a confirmation dialog to the L keyboard shortcut
The keyboard shortcut `L` is assigned to finding your location (which might take some time) and then moving the visible viewport to that position. Sometimes when you are typing tags on the keyboard you might miss the `TAB` and if you type `L`, then the editor would be stuck into determining your position for a while then jump for a position different from the one you are editing. So it would be great to add a confirmation dialog, or another measure such as disabling the shortcut for a little bit after filing tags manually to avoid such hassle.

### LINKED ISSUE #9128: Remove L shortcut key to locate me
### URL

_No response_

1. Start drawing an area, creating a few points
2. While still incomplete, press **L** on the keyboard (Locate me)
As a result, the map is moved to my current location, but my unfinished area is gone for good. Undo does not help.

Rather than investigating why the area is gone, I suggest removing the L shortcut key altogether. It's too easy to hit accidentally, e.g. while searching for presets or objects (the above was result of a brain fart, I wanted an **F** instead). Locating myself while drawing is seldom necessary, and there's a dedicated side button for that anyway.

_No response_

_No response_

_No response_

k-yle:
Closes #9128. related to but does not fix #8474

It's easy to accidentally click <kbd>L</kbd> which messes up what you're doing. So changing it to <kbd>⌘</kbd> + <kbd>L</kbd> makes it harder to accidently trigger the shortcut.

This makes it slightly less convinient for people who genuinely use this feature. But I don't think it's a very popular shortcut, since iD doesn't work well on touch screens. Maybe it's useful if you're mapping live from a moving bus or ferry? or wandering down the road with your laptop....


tyrasd (changes requested):
Ctrl+L is already used as a system shortcut to get to the address bar of the browser.

As this is not a functionality that people use often, and at most once per session: Let's just remove the shortcut `L` completely.

k-yle:
done, this PR now completely removes the shortcut

tyrasd (approved):

PR #10002, opened 2023-11-25
show empty tag values in multi-selections

### LINKED ISSUE #9876: Null hiding among "multiple values"
### Description

Here there are multiple values.
But we only see one value.
Then how could there be multiple values?
Because the other value is the null value.

What would the use of also displaying a second line
"(null)" or "(empty)"?

Well besides being able to click on the one named value, in order to
change all items to use that single value,
we could also click "(null)" to make them all null.

But OK, if we were smart we could also just click on the trash can.

But at least showing "(null)" would reveal the mystery behind the single
"multiple" value.

Also consider the case where there are three values, A, B, and null.

Well when the user clicks, s/he will just see A and B, and go home thinking s/he knows the whole story.

But that isn't the whole story at all!!

So one should see
- A
- B
- (null)

### Screenshots

20230830T044303

k-yle:
Closes #9876

If you select a node with `natural=*` and another node with no `natural` tag, the multiselect dropdown now includes `(empty)` as an option, to avoid confusion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :sparkles: Usability & Accessibility
 * Show the color of (route) relations in the form of small colored circles in relation membership section and feature search results ([#9424])
 * Hide tag suggestions for tags like `name_1` in raw tag editor autocomplete ([#9422])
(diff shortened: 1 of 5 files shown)


tyrasd (approved):
Thanks. I've made a few minor tweaks:

* use italic font style to make option visibly different from regular tag values
* switch back to use the raw tag value as the displayed option label in the combobox dropdown: using `d.title` was confusing for tags where taginfo returns rather long descriptions (these are still displayed as the respective option's hover text, btw)

PR #10003, opened 2023-11-25
prevent degenerate ways caused by deleting a corner of a triangle

### LINKED ISSUE #7946: Disallow undoing to a degenerate area

### LINKED ISSUE #9329: Create duplicate line segments by deleting node on line.
The URL is not great because I fixed the issue there, but if you follow the steps to reproduce you will see the issue.

### URL

1. Click on "line"
2. create a triangle so that the last node is the same as the first node.
3. delete one of the nodes.
4) Choose a type for the feature (e.g. "tree row").

At this point you will have a line that references the same node twice.

The change file shows a the two new nodes and one node being referenced twice.

```
    <create>
        <node id="-24" lon="116.36601429790252" lat="39.98403808057159" version="0"/>
        <node id="-25" lon="116.36629195628717" lat="39.98402975391612" version="0"/>
        <way id="-4" version="0"><nd ref="-24"/><nd ref="-25"/><nd ref="-24"/>
            <tag k="natural" v="tree_row"/>
        </way>
    </create>
    <modify/>
    <delete if-unused="true"/>
</osmChange>
```

2.22.0

Chrome

k-yle:
Closes #7946, Closes #9329

@tyrasd had some alternative ideas in https://github.com/openstreetmap/iD/issues/9329#issuecomment-1274984052, this PR implements option 1: you can no longer create a degenerate way by deleting the third-last node.


k-yle:
_**edit:** unit test fixed by making it more realistic (circularising a triangle)_

There's one test failing:

https://github.com/openstreetmap/iD/blob/407360943ce5db9d98e857f6148c7876d789ff37/test/spec/actions/circularize.js#L281

It's a bit of a weird test, because if you try to use the circularlise operation on a way that loops back on itself (`A -> B -> A`), nothing happens anyway. I think it's better to change update this test case.

matkoniecz:
#7946 was not fixed by it and got reopened

PR #10007, opened 2023-11-28
Sanitize Amazon CloudFront signature in imagery_used

excitablesnowball:
Similar to https://github.com/openstreetmap/iD/commit/47aaec0db6e73f56d4bb14447155de1b4abd400b and https://github.com/openstreetmap/iD/pull/8976, this removes Amazon CloudFront signatures from the URL in `imagery_used`. This comes up when using the [Strava global heatmap](https://wiki.openstreetmap.org/wiki/Strava#Global_Heatmap_in_High_Resolution) in iD.


tyrasd (approved):
Makes sense, thank you!

PR #10013, opened 2023-12-01
Fix geo-images UX issue #9918

### LINKED ISSUE #9918: new geo-images has buggy UX
### URL

_No response_

Should work like this:
- click "edit" to launch iD
- open "Map data" (shortcut U) from the right-hand panel
- go down to "georeferenced photos"
- click the three dots to open the "load photos" dialogue
- select a random NOT geo-tagged image.

The result is the attached images.

Problems:

* the actual image filename (IMG_etc) is in the same color as the frame. It is not black, like all other text.
* the alert icon is too easy to miss. There is nothing else in the app to show that there are issues so you create a UX problem with people missing there is a detected issue because your alert image is light gray. Even less conspicuous than plain text.
* hovering over the warning icon shows a warning that is cut off.

Screenshot_20230927_103451

While hovering the 'alert' icon:

Screenshot_20230927_102756

_No response_

Firefox

ramith-kulal:
This pull request addresses UX issues related to the georeferenced photos feature. The identified problems included:

- The image filename (IMG_etc) displaying in the same color as the frame.
- The alert icon being too easy to miss due to its light gray color.
- A warning message being cut off when hovering over the alert icon.

https://github.com/openstreetmap/iD/assets/142984776/4910b06b-3043-4e9f-9ec7-ada6ddec9a95

Closes #9918

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -497,8 +497,8 @@ label.streetside-hires {
 
 .list-local-photos {
     max-height: 40vh;
-    overflow-y: scroll;
-    overflow-x: auto;
+    overflow-y: visible;
+    overflow-x: visible;
     /* workaround for something like "overflow-x: visible"
        see https://stackoverflow.com/a/39554003 */
     margin-left: -100px;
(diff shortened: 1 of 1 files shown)


tyrasd (approved):
Hey. Thank you for the PR.
There was a reason for the overflow=scroll in the css, though: when many (e.g. 20) photos are loaded, the list would actually overflow onto the confirm button area in the message dialog. To ix this, the intention was the list to be scrollable. I just changed the fix to keep the scroll behavior, but also leave enough space for the tooltip to not be clipped (see fd88a265bffc3f9518e7877699abf449b3f62cbe).

Issue #6411, opened 2019-05-22
Warn when changing a name to the wrong name


1ec5:
The title sounds amazingly infeasible, but hear me out.

Every once in a while, a local mapper gets frustrated that armchair mappers from elsewhere keep changing a road name to something it shouldn’t be. Perhaps the street-level imagery is outdated or misleading or an official source like TIGER is simply incorrect. The local mapper can tag the road with [`not:name`](https://wiki.openstreetmap.org/wiki/Key:not:name) (optionally with [`not:name:note`](https://taginfo.openstreetmap.org/keys/not:name:note) or [`source:not:name`](https://taginfo.openstreetmap.org/keys/source%3Anot%3Aname)). If an armchair mapper later comes along changing `name` to the same value as `not:name`, iD should issue a validation warning.

There are [some other `not:*` tags](https://taginfo.openstreetmap.org/search?q=not%3A) too.

1ec5:
@jleedev gave an example of [this road](https://www.openstreetmap.org/way/456793503), which was changed to the same misspelling on [two](https://www.openstreetmap.org/changeset/65119981) [separate](https://www.openstreetmap.org/changeset/65113240) occasions.

1ec5:
[This POI](https://www.openstreetmap.org/node/3648453571) has two names in `not:name`. Changing `name` to either or both of them should result in a warning.

quincylvania:
@1ec5 Interesting. I do think iD should honor semicolon-separated `not:name` values for this validation.

quincylvania:
@1ec5 Is there consensus on how to tag incorrect names in different languages? I see a little usage of `not:name:cy` and `not:name:en`, but are these supposed to map only to `name:cy` and `name:en`? Or if a value in `not:name` equals `name:de`, should iD issue a warning?

I think I'd prefer the latter, so all incorrect names in all languages go in `not:name`. This could lead to some edge cases, but iD already has a field for `not:name`.

1ec5:
In general, `not:x` should only apply to `x`, and subkeys `not:x:y` should only apply to `x:y`, not `x:z` or `x`. A namespace is just that, not a specific variation of the namespaced key.

Imagine a situation where mappers feel the need to clarify that the Spanish name of a place is not the same as the local name (which happens to be in English). This seems like quite a corner case, so I guess we can wait and see if iD’s use of `not:name` for all languages happens to cause any problems.

quincylvania:
[...]

PR #10035, opened 2023-12-10
New validator: mutually exclusive tags

### LINKED ISSUE #7589: name and noname tags together
When `noname=yes` tag is present, could we make the Name field grey like it is when it has a wikidata tag?
We can additionally change the placeholder to, e.g, "This feature has a noname tag".

### LINKED ISSUE #8306: iD should suggest removing noname=yes from roads with a name
I've come across a number of named roads that still have the noname=yes tag set. This is invalid and should be marked so when editing.

image

mtmail:
Closes #8306, Closes #7589

About 20.000 places currently have `name` and `noname` set. https://overpass-turbo.eu/s/1Eyn (map screenshot and discussion at https://community.openstreetmap.org/t/places-with-name-but-also-noname-yes/106858)

I chose 'mutually exclusive' but 'incompatible' or 'conflicting' could also work. I'm open to wording suggestions.

Future versions might need to handle values, too. For example `["wheelchair=yes", "ladder=yes"]`. Or patterns `["noname", "name:*"]`

In the future I'd like to see the list of pairs maintained by the community in https://github.com/openstreetmap/id-tagging-schema/ like the list of deprecated tags.

Thank you @tyrasd for the iD development workshop at SotM Antwerp.

image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1839,9 +1839,6 @@ en:
         amap: "Amap products are proprietary and must not be used as references."
         baidu: "Baidu products are proprietary and must not be used as references."
         google: "Google products are proprietary and must not be used as references."
-    incorrect_name:
-      message: '{feature} has the mistaken name "{name}"'
-      message_language: '{feature} has the mistaken name "{name}" in {language}'
     invalid_format:
       title: Invalid Formatting
       tip: Find tags with unexpected formats
@@ -1873,6 +1870,15 @@ en:
         message: "{feature} has no descriptive tags"
       relation_type:
         message: "{feature} is a relation without a type"
+    mutually_exclusive_tags:
+      title: Contradictory Tags
+      tip: "Find features that have contradictory tags."
+      default:
+        message: "{feature} has both {tag1} and {tag2}"
+        reference: 'The tags "{tag1}" and "{tag2}" contradict each other. Only one tag should be set.'
+      same_value:
+        message: "{feature}: {tag1} and {tag2} have same value"
+        reference: 'The tags "{tag1}" and "{tag2}" have the same value. Only one tag should be set.'
     old_multipolygon:
       message: "{multipolygon} has misplaced tags"
       reference: "Multipolygons should be tagged on their relation, not their outer way."
@@ -1998,6 +2004,9 @@ en:
       remove_tag:
         title: Remove the tag
         annotation: Removed tag.
+      remove_named_tag:
+        title: 'Remove the "{tag}" tag'
+        annotation: 'Removed "{tag}" tag.'
       remove_tags:
         title: Remove the tags
       remove_the_name:
--- a/modules/validations/mutually_exclusive_tags.js
+++ b/modules/validations/mutually_exclusive_tags.js
@@ -1,0 +1,93 @@
+import { actionChangeTags } from '../actions/change_tags';
+import { t } from '../core/localizer';
+import { utilDisplayLabel } from '../util';
+import { validationIssue, validationIssueFix } from '../core/validation';
+import { osmMutuallyExclusiveTagPairs } from '../osm/tags';
+
+export function validationMutuallyExclusiveTags(/* context */) {
+    const type = 'mutually_exclusive_tags';
+
+    // https://wiki.openstreetmap.org/wiki/Special:WhatLinksHere/Property:P44
+    const tagKeyPairs = osmMutuallyExclusiveTagPairs;
+
+    const validation = function checkMutuallyExclusiveTags(entity /*, graph */) {
+
+        let pairsFounds = tagKeyPairs.filter((pair) => {
+            return (pair[0] in entity.tags && pair[1] in entity.tags);
+        }).filter((pair) => {
+            // noname=no is double-negation, thus positive and not conflicting. We'll ignore those
+            return !((pair[0].match(/^(addr:)?no[a-z]/) && entity.tags[pair[0]] === 'no') ||
+                     (pair[1].match(/^(addr:)?no[a-z]/) && entity.tags[pair[1]] === 'no'));
+        });
+
+        // Additional:
+        // Check if name and not:name (and similar) are set and both have the same value
+        // not:name can actually have multiple values, separate by ;
+        // https://taginfo.openstreetmap.org/search?q=not%3A#keys
+        Object.keys(entity.tags).forEach((key) => {
+            let negative_key = 'not:' + key;
+            if (negative_key in entity.tags && entity.tags[negative_key].split(';').includes(entity.tags[key])) {
+                pairsFounds.push([negative_key, key, 'same_value']);
+            }
+            // For name:xx we also compare against the not:name tag
+            if (key.match(/^name:[a-z]+/)) {
+                negative_key = 'not:name';
+                if (negative_key in entity.tags && entity.tags[negative_key].split(';').includes(entity.tags[key])) {
+                    pairsFounds.push([negative_key, key, 'same_value']);
+                }
+            }
+        });
+
+        let issues = pairsFounds.map((pair) => {
+            const subtype = pair[2] || 'default';
+            return new validationIssue({
+                type: type,
+                subtype: subtype,
+                severity: 'warning',
+                message: function(context) {
+                    let entity = context.hasEntity(this.entityIds[0]);
+                    return entity ? t.append(`issues.${type}.${subtype}.message`, {
+                        feature: utilDisplayLabel(entity, context.graph()),
+                        tag1: pair[0],
+                        tag2: pair[1]
+                    }) : '';
+                },
+                reference: (selection) => showReference(selection, pair, subtype),
+                entityIds: [entity.id],
+                dynamicFixes: () => pair.slice(0,2).map((tagToRemove) => createIssueFix(tagToRemove))
+            });
+        });
+
+        function createIssueFix(tagToRemove) {
+            return new validationIssueFix({
+                icon: 'iD-operation-delete',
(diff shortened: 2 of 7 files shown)


modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,46 @@
+import { t } from '../core/localizer';
+import { utilDisplayLabel } from '../util';
+import { validationIssue } from '../core/validation';
+
+export function validationMutuallyExclusiveTags(/* context */) {
+    const type = 'mutually_exclusive_tags';
+
+    // https://wiki.openstreetmap.org/wiki/Special:WhatLinksHere/Property:P44
```
1ec5, line 10: Good idea to consult this property. 👍

modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,46 @@
...
+    const tagPairs = [
+        ['noname', 'name'],
```
1ec5: #6411 added a check for `name` matching `not:name`. It isn’t quite the same thing as checking whether two keys are merely present, but I wonder if it should be integrated into this validator rule somehow:

https://github.com/openstreetmap/iD/blob/bb17d5c24eb6c58724b8be83c646bfe0aa9cf3ad/modules/validations/suspicious_name.js#L154-L170

In that issue, we considered it important to add a field for `not:name` so that the user wouldn’t have to dig through the raw tags to resolve the issue. I don’t think any of the negative keys here have fields yet. Maybe it wouldn’t block this PR at all, but we should consider adding fields to id-tagging-schema separately.

data/core.yaml
```diff
@@ -1867,6 +1867,10 @@ en:
         message: "{feature} has no descriptive tags"
       relation_type:
         message: "{feature} is a relation without a type"
+    mutually_exclusive_tags:
+      title: Mutually Exclusive Tags
+      tip: "Find features having conflicting tags set."
+      message: "{feature} has conflicting tags: {tag1}, {tag2}. Only one should be set."
```
1ec5: This warning message should be brief; provide more details in a `reference` property:

```suggestion
      title: Contradictory Tags
      tip: "Find features that have contradictory tags."
      message: "{feature} has both {tag1} and {tag2}"
      reference: "The keys {tag1} and {tag2} contradict each other. Only one should be set."
```

modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,46 @@
...
+        let issues = pairsFounds.map((pair) => {
+            return new validationIssue({
+                type: type,
+                severity: 'warning',
+                message: function(context) {
+                    let entity = context.hasEntity(this.entityIds[0]);
+                    return entity ? t.append('issues.' + type + '.message', {
+                        feature: utilDisplayLabel(entity, context.graph()),
+                        tag1: pair[0],
+                        tag2: pair[1]
```
1ec5, line 52: Since these are raw key names, how about wrapping them in `<code>` elements so they look monospaced?

modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,46 @@
...
+                    }) : '';
+                },
```
1ec5, line 54: In addition to specifying a `reference`, consider supplying a `dynamicFixes` with two fixes that remove either of the contradictory tags.

data/core.yaml
mtmail: I updated the titles

modules/validations/mutually_exclusive_tags.js
mtmail: I added fixes now.

mtmail:
I added dynamic fixes and better titles, thanks @1ec5.

image

modules/validations/mutually_exclusive_tags.js
mtmail: I couldn't find any other translation strings using HTML and when supplying values for the `{}` replacements any HTML gets stripped. I added quotes around the name.

modules/validations/mutually_exclusive_tags.js
mtmail: So basically whenever a "tagname" and "not:tagname" is found and both have the same value then validation should fail? I think it makes sense to move that logic into this new validator.

pnorman:
> I added dynamic fixes and better titles, thanks @1ec5.
> 
> image

There are two quotation marks after name in the text, there should only be one.

mtmail:
@pnorman Thank you, fixed now.

modules/validations/mutually_exclusive_tags.js
1ec5: Yes, that’s what I was thinking. However, another difference between this validator rule and the existing `not:name` rule is that the latter validates `name:*` subkeys against each value in `not:name`. We could justify keeping the two rules separate as long we gracefully handle the `noname=yes` `not:name=*` combination that currently occurs on about 100 features.

modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,75 @@
...
+        ['noaddress', 'addr:housenumber'],
+        ['noaddress', 'addr:housename'],
+        ['noaddress', 'addr:unit'],
+        ['addr:nostreet', 'addr:street']
+    ];
+
+    const validation = function checkMutuallyExclusiveTags(entity /*, graph */) {
+
+        const pairsFounds = tagPairs.filter((pair) => {
+            return (pair[0] in entity.tags && pair[1] in entity.tags);
```
1ec5, line 16: It isn’t as common, but a negative key can technically be set to the value `no` as well. The double negative would be redundant to the positive tag but wouldn’t contradict it. I guess it’s OK if we still show the warning in this case, but it’s a bit confusing if you know what’s in the raw tags.

modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,75 @@
...
+import { validationIssue, validationIssueFix } from '../core/validation';
+
+export function validationMutuallyExclusiveTags(/* context */) {
+    const type = 'mutually_exclusive_tags';
+
+    // https://wiki.openstreetmap.org/wiki/Special:WhatLinksHere/Property:P44
+    const tagPairs = [
+        ['noname', 'name'],
+        ['noref', 'ref'],
+        ['noaddress', 'addr:housenumber'],
```
1ec5: [`nohousenumber`](https://wiki.openstreetmap.org/wiki/Key:nohousenumber) also contradicts `addr:housenumber`.

data/core.yaml
```diff
@@ -1992,6 +1997,9 @@ en:
       remove_tag:
         title: Remove the tag
         annotation: Removed tag.
+      remove_named_tag:
+        title: Remove the '{tag}' tag
+        annotation: Removed '{tag}' tag.
```
1ec5: Use double quotation marks for consistency with this validator rule’s `reference` and with other validator rules such as `tag_as_not`:

```suggestion
        title: 'Remove the "{tag}" tag'
        annotation: 'Removed "{tag}" tag.'
```

mtmail: good idea

modules/validations/mutually_exclusive_tags.js
mtmail: I added logic now

modules/validations/mutually_exclusive_tags.js
mtmail: Added to the list now

modules/validations/mutually_exclusive_tags.js
mtmail: I added logic to look for "tagname" and "not:tagname" and if both have the same value. And if tagname is "name:*" then it's also checked against "not:name". I'd appreciate if you could double-check.

modules/validations/mutually_exclusive_tags.js
```diff
@@ -0,0 +1,97 @@
...
+        let pairsFounds = tagKeyPairs.filter((pair) => {
+            return (pair[0] in entity.tags && pair[1] in entity.tags);
+        }).filter((pair) => {
+            // noname=no is double-negation, thus positive and not conflicting. We'll ignore those
+            return !((pair[0].match(/^(addr:)?no[a-z]/) && entity.tags[pair[0]] === 'no') ||
+                     (pair[1].match(/^(addr:)?no[a-z]/) && entity.tags[pair[1]] === 'no'));
+        });
+
+        // Addtional:
+        // Check if name and not:name (and similar) are set and both have the same value
```
1ec5, line 24: Thanks for humoring my suggestion. This implementation looks pretty good, but the message for this particular situation should be about the _value_ of `not:name` matching the _value_ of `name`, not that both keys exist. After all, the combination of `name=Mi Casa` and `not:name=Su Casa` is valid. You’re currently checking that correctly, but the warning message is a bit misleading.

Moreover, there are two redundant warnings every time `name` matches `not:name`. Either we should keep the preexisting `not:name` validation rule and special-case this code to skip over `not:name`, or we should remove the preexisting `not:name` rule and clarify the messages similar to the original.

ABC has both not:name and name, Building has the mistaken name “ABC”

mtmail: I moved logic and tests from the `suspicious_name` validator now. You're right I overlooked that `not:name` can be a list of values, semicolon separated. That's also covered now.

mtmail:
Updated screenshots

image

mtmail:
@tyrasd I removed the commented out code, fixed a codespell warning and moved the list of tags to `modules/osm/tags.js`. The branch is rebased with `develop` branch and test pass for me.

tyrasd (approved):

PR #10045, opened 2023-12-19
Adds feature to discard all the changes on the map

### LINKED ISSUE #10033: Allow throwing away all changes instead of uploading them
### Description

Vespucci allows throwing away all changes instead of uploading them. iD should too. Yes, after a "are you sure?" screen.

You might say
> Just undo undo undo. Voila, no more changes to upload.

But as we see in #10029, there might be no full undo list.
So the user is stuck with changes they can't get rid of unless they upload them, or make a new OSM account.

> Wait until #10031 is implemented.

But that is hard to implement, whereas throwing away all changes at once is easy to implement.

So at the Save > Upload / Cancel screen, a third choice "Throw away all these edits" etc. should appear.

Asif-Sheriff:
### Description 
This PR addresses the issue #10033 which concerns the need for a discard button to discard all the changes on the map. Previously this was done by reloading the map and then discarding all the changes but this PR will create a discard button to dump all the changes without having to reload the page, also unlike the reload method this discard action can be undone.

### Changes made
Presently this PR has 3 commits 

Commit1: creates the discard feature.
Commit2: merges changes made into the original repo.
Commit3: fixes lint issues in the code.

Here's the rundown of changes made in Commit 1:

1. Creates a discard() function in history.js which sets the current state of the map to the original state.
2. Created a discard.js file in /tools to render the discard button and write the logic for the same.
3. Added UI label in helper.js
4. Added the discard button to top_toolbar.js

### Proof of concept
[Screencast from 12-12-23 10:52:08 PM IST.webm](https://github.com/openstreetmap/iD/assets/130328580/1b8a7132-b6d7-41d4-8590-86d83c8accc8)

### Final thoughts
I understand that this PR makes some major changes to iD therefore I would request the admins to take a look and this and if there are any changes to be made feel free to let me know as I am more than willing to work further on this issue.


dypa:
The demo looks amazing!  Сan you add improvements?
- the "discard" button has replaced the save button location, this may confuse users - you need to move the "discard" button near the "undo" button
- add a confirmation window
- check will the translation mechanism work for other languages?

Asif-Sheriff:
@dypa 
Sure Ill work on it

Asif-Sheriff:
@dypa 
Is this better?
[Screencast from 08-01-24 10:42:25 AM IST.webm](https://github.com/openstreetmap/iD/assets/130328580/d7a5b509-6367-4e72-8849-d79d80879d90)

tyrasd (commented):
This seems to be useful in some niche cases. However, I'm not convinced a button in the main edit toolbar is needed for this. I would have put it into the "Map Data" panel, perhaps as in this mockup: 



Alternatively, there could be a dropdown/long-click option of the undo button to _"Undo everything"_ (with slightly different semantics: instead of creating a new state which could be undone to get one's edits back in a single click of the _Undo_ button, this would correspond to undo all changes which could be redone one by one using the _Redo_ button).

tyrasd:
PS: If changes can be undone, there is really no need to add a confirmation dialog. I'd remove it.

PSS: the `dist` directory should never be manually changed, as it is recreated automatically by the build scripts. New translatable strings should be added to the `data/core.yaml` file ([read more](https://github.com/openstreetmap/iD/wiki/How-to-use-iD%27s-localizer-module#basic-usage)).

Asif-Sheriff:
@tyrasd 
Hey I liked the long press functionality for the undo button and here's the implementation

[Screencast from 21-01-24 03:38:15 PM IST.webm](https://github.com/openstreetmap/iD/assets/130328580/58edfff0-45aa-4a12-87cd-d807fe9aa84e)


 Like you suggested, long pressing the undo button will discard all the changes and set the state of the map to the way it was initially also the user can redo the changes one by one if they wish.

If this is alright then I will commit the changes.

mxdanger:
Since there hasn’t been any other feedback, please reconsider the use of hold undo to discard. This is unintuitive and unexpected user interface design. There’s no need to reinvent the wheel here.

Similar to JOSM, having the delete button alongside the OSM data layer is the most straightforward option. This keeps it tucked away out of the way from new users but located in a spot where it would be expected by power users. The mockup @tyrasd made would be ideal.

For the question of delete confirmation, here are three options we could go with:

1. No confirmation. You can use the redo button to revert (This applies to all three)
2. Click delete twice. Trashcan icon changes into a check icon requiring a second click to confirm.
3. Confirmation dialog.

tordans:
I still just reload the browser window (via browser shortcut) and discard all previous changes in the dialog that shows whenever you open iD with previous changes in cache. To me, that gives all I need: Very quick access, a confirmation … and no change for anyone to click it accidentally when submitting the work.

matkoniecz:
why this is even needed?

that seems to increase complexity of user interface for no good reason

tyrasd:
I think, @matkoniecz is right: this is such a rare thing one would need in regular mapping that the workarounds of e.g. just refreshing the page an selecting the discard option is sufficient.

Issue #9339, opened 2022-10-20
Add a button to the Mapillary field to set the value to the currently selected Mapillary photo id


RudyTheDev:
In addition, always show the [`mapillary`](https://wiki.openstreetmap.org/wiki/Key:mapillary) field when a Mapillary photo is selected.

This would significantly simplify this action. Currently, one has to either (A) open the photo link, copy the id from the browser's URL bar, and close the tab, add the field, and paste the id into the field or (B) copy the photo link, add the field, paste the id into the field, and trim the non-id part from the URL.

This should encourage more use and promote verifiability. It would also make mapping with Mapillary in iD a little more usable. Without going into tangentials, it's not the easiest workflow at the moment.

It would also make sense to have a button to activate Mapillary layer and select the photo. Not sure how viable this is though - in general, one would have to query Mapillary API for the location of the photo, so this wouldn't be an instant action. Perhaps it may need to be restricted to those Mapillary photos that are currently loaded/visible.

timothywashere:
I also find it quite annoying that there is no easy way to add a Mapillary image tag to something when I am using the Mapillary layer inside the iD editor. I have to click on the mapillary icon to open the image in another window, then find the id and close the window again. It's such a waste of time and bandwidth.
So it would be very nice if there was a quick way of getting the iD of the currently displayed Mapillary image within the iD editor

Ni-vik:
@tyrasd Is this issue still open? It's my first time contributing in openstreetmap as this issue is labelled with good first issue , I would like to work on this.

tordans:
FYI, there is a comment of mine from Dez 2000 with some Wireframe-like illustrations of this feature idea https://github.com/openstreetmap/iD/issues/6196#issuecomment-751460516

jdabapo:
@Ni-vik are you working on this issue already? I would like to contribute as well, but I do not want to take an issue away if you are already working on it.

Ni-vik:
@jdabapo yep I am working on it

kunjbosamia:
Hey is this issue still open if yes then I would like to work on this

Gua00va:
Hey!! I want to work on this issue. I'm new here so I'm really excited to start contributing. Please assign it to me.

KishorKumarParoi:
Hello Friends and folks, is this issue already fixed? If not I want to work in this issue

tyrasd:
> Hello Friends and folks, is this issue already fixed? If not I want to work in this issue

[...]

Issue #6196, opened 2019-04-17
Support for Photo linking


famo:
It would be nice if iD could support Photo linking:
https://wiki.openstreetmap.org/wiki/Photo_linking

However, it shouldn't show the image directly in the editor, because this can introduce unwanted data traffic (and the image would probably be to small anyways).

Rather the editor should provide a link with the complete URL, which upon clicking would open in a new tab.

Ideally this link is directly available in the fields section or should be visible after clicking the `i` next to the tag.

bhousel:
Thanks @fmao - have you tried iD's Mapillary / Streetside / OpenStreetCam integrations?

famo:
No, so far I encountered only links which referred to wikimedia commons...

Bibi56:
> I'd suggest this for OpenStreetCam and Bing Streetside too but people don't seem to be linking those with tags right now.
(from @quincylvania in #7064)

"right now" as the coverage is not as good and as the editors were not as tuned for those sources. If you improve the usability, you'll be rewarded by the growth of usage.

quincylvania:
> "right now" as the coverage is not as good and as the editors were not as tuned for those sources. If you improve the usability, you'll be rewarded by the growth of usage.

@Bibi56 I agree. When we add photo linking I don't see why it wouldn't work the same way for all four photo overlay sources.

quincylvania:
Re-reading this issue, I see it just specifies the sort of field functionality added for #7064. I've been considering "adding support for photo linking" to involve the actual linking of images from iD's street level viewer to OSM features with a UI element, so I'd say let's keep this open until we add that.

tordans:
> I've been considering "adding support for photo linking" to involve the actual linking of images from iD's street level viewer to OSM features with a UI element, so I'd say let's keep this open until we add that.

I talked to @cbeddow about this by mail and wanted to add a few thoughts.

Possible usecases:

a. Case: OSM object selected, does not have a mapillary-key, street level viewer shows image
  => Button on street level viewer to add the current image to the object as mapillary=<ID>

![add-tag-when-no-key Kopie-2](https://user-images.githubusercontent.com/111561/103170545-3e4b5600-4845-11eb-8be4-fbc539701093.jpg)

b. Case: OSM object selected, mapillary-key field was added manually via "add field", street level viewer shows image
  => Button on the key-form-element to copy the ID

[...]

PR #10046, opened 2023-12-20
Add a button in the mapillary image photo viewer to "add the current image id as field to feature"

laigyu:
https://github.com/openstreetmap/iD/issues/9339
https://github.com/openstreetmap/iD/issues/6196#issuecomment-751460516

edit : this pr only focus on button in viewer


data/core.yaml
```diff
@@ -786,6 +786,8 @@ en:
       inch: in
     max_length_reached: "This string is longer than the maximum length of {maxChars} characters. Anything exceeding that length will be truncated."
     set_today: "Sets the value to today."
+    set_mapillary: "Sets the current ID of mapillary"
+    show_mapillary_from_field: "Show image on viewer"
```
tyrasd:
```suggestion
    set_photo_from_viewer: "Store this photo on the currently selected map object"
    set_photo_from_field: "Use the currently displayed photo"
    show_photo_from_field: "Open image in viewer"
```

(the string references would of course need to be adjusted accordingly as well)

tyrasd (dismissed):
Hey. This is awesome!

There are only a few minor-ish things I would change:

* The icon used inside of the mapillary viewer is not all too intuitive: One might assume that it refreshes the image or something like that. Maybe the [`load` icon](https://github.com/openstreetmap/iD/blob/develop/svg/iD-sprite/icons/icon-load.svg) would be better? Or the [`operation-extract`](https://github.com/openstreetmap/iD/blob/develop/svg/iD-sprite/operations/operation-extract.svg) one? Or the [`+` from `operations-merge`](https://github.com/openstreetmap/iD/blob/develop/svg/iD-sprite/operations/operation-merge.svg)?
* The used strings should be made independent from the photo service provider to be able to be used them in the future with the other street level photo providers as well. See suggestions below. 
* The button to set the mapillary tag should be disabled or hidden if no OSM entity is selected.
* The button to open the viewer should be disabled if the photo is already currently displayed
* If the mapillary id of an OSM object is invalid or does not exist, the viewer is opened showing the last selected photo (or remains open if it was already showing a photo). It would be better if the viewer would show some kind of error state in this situation.

laigyu:
set button in field and photoviewer are both use to set current id to input, 
not sure about how to do
1. is fine to have both and use same logic as field set button  to disable photoviewer set button 
2. if mapillary id field not show,show photoviewer set button and after click it ,remove the button ,and use only field set button to update id
3.  only use photoviewer button and not use set button in field
4.  or other better way ? :pray:

tordans:
@laigyu why did you close this PR? I was looking forward to seeing it as part of iD at some point?
Would you be up to polish it and bring it over the finish line?
(This does take a while … but that is how it is with projects like iD…)

tordans:
@laigyu I tested the UI and I suggest you split of a few things just to make this easier to merge an iterate on.

The part that is most clear to me is his use case: 

## UseCase A: Add the Mapillary field. And if an mapillary image is selected, provide an action to copy the iD into the field.

This part of the UI and interaction looks good to me with some adjustments

1. Change the icon of the button
1. Fix the border radius of the button

### icon

I assume we are using font awesome free icons here, right?

I looked at some options, here are my picks

| icon | link | comment |
|--------|--------|--------|
| image | https://fontawesome.com/icons/eye-dropper?f=classic&s=solid | my favorite. it tells the "pick this" story well |
| image | https://fontawesome.com/icons/right-to-bracket?f=classic&s=solid |  |
| image | https://fontawesome.com/icons/arrow-right-to-bracket?f=classic&s=solid | |
| Cell | https://fontawesome.com/icons/file-import?f=classic&s=solid | not really |
| Cell | https://fontawesome.com/icons/file-arrow-down?f=classic&s=solid | not really | 

### border radius

It looks like, this CSS needs to change:

```diff
.ideditor[dir='ltr'] .form-field-input-identifier > input:last-child,
.ideditor[dir='rtl'] .form-field-input-identifier > input:first-child,
- .ideditor[dir='ltr'] .form-field-input-identifier > button {
+ .ideditor[dir='ltr'] .form-field-input-identifier > button:last-child {
    border-bottom-right-radius: 4px;
}
```
so the radius is only applied to the last button

---

I suggest to make a PR that only includes this change, which should be much easier to merge.

tordans:
About the other usecases…

### UseCase B: Show current mapillary image from the sidebar

I think this is a nice idea, but it has a lot of edge cases that are hard to get right. I suggest to work on this last, because there is a feature to show the current image already. It is not as ideal, because it opens it in an new tab, but the other use cases are more important IMO.

This is what it looks like now 
image

The issues I see is, that one can show an image but that show click does not activate the mapillary overlay, it only shows the current image. Apart from that, I think the Icon and interaction is good.

There is also this issue from https://github.com/openstreetmap/iD/pull/10046#pullrequestreview-1827207017
> If the mapillary id of an OSM object is invalid or does not exist, the viewer is opened showing the last selected photo (or remains open if it was already showing a photo). It would be better if the viewer would show some kind of error state in this situation.

tordans:
About the other usecases

### UseCase C: Add a `mapillary` field with the current image id to the selected feature from the current image UI

This feature is super nice! 

I think those things should change…

1. Use the same icon as for the field, eg. the "picker" icon. I think we should reuse the icon because it does the same thing, it copies the id inside the field. The context of the icon placement clarifies the direction of the action well enough, so no need for a different icon IMO. (Also, those actions are super abstract, so it's hard to find a good icon…)
2. Add a tooltip. It looks like iD has a tooltip system and I hope that can be reused in this case, so users see the icon title attribute as a tooltip which will explain the feature a lot better. 
   Example tooltip image
3. Move the button into the panel and change the style.
   Style: I suggest to make it look like any button in the sidebar: image (with full rounded corners)
   Position: I suggest here Bildschirmfoto 2024-04-05 um 09 21 27
3. Only one button per feature. Right now there are situations when multiple buttons are present: 
    [Example](http://127.0.0.1:8080/#background=Bing&disable_features=boundaries&id=w41391577&locale=en-GB&map=19.56/54.20783/10.42413&photo=mapillary/249298364284942&photo_overlay=mapillary):  image
    Both buttons seem to update the same tag, however. 
    There should be one button only per selected feature. When more than one feature is selected (multiselect) the button should be either inactive or it should add the same image id to both features.

data/core.yaml
```diff
@@ -788,6 +788,10 @@ en:
       inch: in
     max_length_reached: "This string is longer than the maximum length of {maxChars} characters. Anything exceeding that length will be truncated."
     set_today: "Sets the value to today."
+    set_photo_from_viewer: "Store this photo on the currently selected map object"
```
tordans:
```suggestion
    set_photo_from_viewer: "Tag this photo id on the currently selected map object"
```

"Store" suggest an upload / file. But it is only the id that is referenced as a tag. Maybe like this?

data/core.yaml
```diff
@@ -788,6 +788,10 @@ en:
...
+    set_photo_from_field: "Use the currently displayed photo"
```
tordans:
```suggestion
    set_photo_from_field: "Use the id of the currently displayed photo"
```

data/core.yaml
```diff
@@ -788,6 +788,10 @@ en:
...
+    show_photo_from_field: "Open image in viewer"
```
tordans:
```suggestion
    show_photo_from_field: "Open image in street-level photo viewer"
```

Just to make it a bit more precise.

laigyu:
so there should be only two button, one is  at field the other at photo


| field (leftside)  | photo (right) | before--after | field (leftside)  | photo (right) |
| ------------- | ------------- | --------| --| --|
Show current mapillary image from the sidebar
| 123  | 456 or (no photo and open it)  ||123 |123 |
||Add a mapillary field with the current image id to the selected feature from the current image UI
| 123 or (no field and add field)    | 456  ||456 |456 |


still confused

laigyu:
so there should be only two button right ?

tordans:
> so there should be only two button right ?

- UseCase A adds a button in the sidebar mapillary field to "fetch the current image id"
- UseCase C adds a button in the mapillary image panel to "add the current image id as field to feature"
- UseCase B adds another button in the sidebar mapillary field to "open current image in preview" (that would be three buttons on this field now) – However, as suggest I would extract this into a separate PR to get UseCase A+C merged faster.

laigyu:
I remove a and b part , only left C part 

ccc

> 1. Use the same icon as for the field, eg. the "picker" icon. I think we should reuse the icon because it does the same thing, it copies the id inside the field. The context of the icon placement clarifies the direction of the action well enough, so no need for a different icon IMO. (Also, those actions are super abstract, so it's hard to find a good icon…)

I think the svg is from [svg](https://github.com/openstreetmap/iD/tree/develop/svg)
I dont know is it possiable to add your suggestion

> 2. Add a tooltip. 

I think already has tooltip. currently is `.attr('title', t('inspector.set_photo_from_viewer'));`  ( no triangle )
the other is `.call(uiTooltip().title(() => t.append('inspector.set_photo_from_viewer')))` just like your picture show 
first type is match the right side close button (X) 

> 3. Move the button into the panel and change the style 

I think position is already at the top left corner. but I dont know why your pic is not as same as mine. plz check if still wrong
as for round  style I copy the style from close button right. shall we change it ?

> 4. Only one button per feature. Right now there are situations when multiple buttons are present: 

I see your pic show two button + + , could you give me step to reproduce it 


functionality : currently the button is listen to `imagechanged` event.
so button will not show at the beginning .
to trigger the imagechange . need to click < > at the top black bar or white arrow first.
as for other , I think is good . 

Any suggestions are welcome.Thank you

tordans:
I think reducing the footprint of this feature is good thing. Some feedback:


> I think the svg is from [svg](https://github.com/openstreetmap/iD/tree/develop/svg) I dont know is it possiable to add your suggestion

I thing this svg is the wrong folder it is https://github.com/openstreetmap/iD/tree/develop/svg/iD-sprite which tyrasd referenced in https://github.com/openstreetmap/iD/pull/10046#pullrequestreview-1827207017
The readme there suggests that adding files is possbile.
**But … lets stick with the `+` icon for now** an see what tyrasd thinks about it.

The `+` is used in other places in the sidebar like the name-tag-translation feature, so that does fit.

> 
> > 2. Add a tooltip.
> 
> I think already has tooltip. currently is `.attr('title', t('inspector.set_photo_from_viewer'));` ( no triangle ) the other is `.call(uiTooltip().title(() => t.append('inspector.set_photo_from_viewer')))` just like your picture show first type is match the right side close button (X)

If possible, I really think we should use the black tooltip for this.
For the `X` it really is more of an accessibility feature. But here we need it for people to understand what the feature is (the `x` is clear)


> 
> > 3. Move the button into the panel and change the style
> 
> I think position is already at the top left corner. but I dont know why your pic is not as same as mine. plz check if still wrong as for round style I copy the style from close button right. shall we change it ?

It is working now. I opened https://github.com/openstreetmap/iD/issues/10193 about the technical issue behind it.

> 
> > 4. Only one button per feature. Right now there are situations when multiple buttons are present:
> 
> I see your pic show two button + + , could you give me step to reproduce it

This is gone now, as well. No idea why the `npm run build` fixed that…

Btw, selecting multiple feature and using the feature also works like it should.

---

### open issue 1: show button once feature is selected
> 
> functionality : currently the button is listen to `imagechanged` event. so button will not show at the beginning . to trigger the imagechange . need to click < > at the top black bar or white arrow first. as for other , I think is good .

Yes, that is an issue that needs to be fixed somehow.

### open issue 2: hide button once no feature is selected (anymore)

The other issue is, that the `+` does not go away when I unselect a feature. 
This sounds to me like you will also need to listen to some "feature un/selected" event. But I don't know enough about this part of the code.
Should you not find anything, I suggest to ping tyrasd.

### open issue 3: disable, not hide button

In https://github.com/openstreetmap/iD/pull/10046#pullrequestreview-1827207017 the suggestion is to disable the buttons, not hide them. (They need a visual disabled state for that; maybe a `cursor: not-allowed;`?)

> * The button to set the mapillary tag should be disabled or hidden if no OSM entity is selected.
> * The button to open the viewer should be disabled if the photo is already currently displayed

laigyu:
besides button svg, I think is good.

tordans (commented):
I looks like I never clicked "send" on those wording suggestions.

tordans:
@laigyu this looks great, thanks. And @tyrasd I think this is ready for another review from you.

Maybe @tyrasd can polish off the wordings / resolve the wording suggestions? 

---

There is a minor thing that gave me pause: Whenever I select an image which was already added as tag, the image PLUS is disabled. However, there is no help in the UI that explains why it is disabled.

> image

What do do…
* I think it is OK to leave it as is
* I think it would be OK to activate the button which would just add the same tag again (change nothing, but allow clicking)
* More complex solution would be a separate error message in the tooltip "This image is already tagged on this feature."

laigyu:
@tordans I think we dont need usecase a button,which is offer the same functionanality as the usecase c button,
the diff is that a is in field, c is in photoviewer, or did I miss something?

tyrasd:
Sorry for the long wait on this PR. I've in the meantime fixed one last(?) bug: The "add mapillary tag" button was not shown if a map feature was newly created while a street level photo was open[^1].

Also, I refactored some parts of the code: It's not good practice to directly rely on other modules' DOM elements, as the implementation details could change at any moment and would break the functionality here. The implementation now directly manipulates the respective OSM entities[^2] instead.

There is now also a limit of (an arbitrary) 100m to the distance between the map feature and the photo to prevent accidentally adding a mapillary tag to a far away map feature.

Finally, I've made minor tweaks to the icon (slightly smaller `+` which matches the styling used on other locations in the UI where we use a plus to indicate that a tag is going to be added) and tooltip texts.

@tordans maybe you would like to give it a final look-over to check whether this didn't introduce any new bugs? :bow: 

[^2]: read more about [iD's data model](https://github.com/openstreetmap/iD/wiki/How-to-get-started#data-model-and-concepts)
[^1]: use case: when mapping a missing business from a mapillary image one would like to immediately set the photo as a "source" of the newly created osm object.

tordans (approved):
@tyrasd thanks for the update <3. I checked the frontend / UX just now and all works great.

There are things to optimize, but I am not sure they are relevant enough as to hold this back…

- when I add an image with `+`, then move up/down the image-path (AKA next/prev image), I cannot overwrite the image (anymore?) – this is something I see myself doing. However, I can still remove the image from the field first, then reselect the feature, then add with `+`
   - note that the same does work, when I change to a different image, so both cases (the one below) might actually be the same

- add image with `+`, then use the undo or trash icon on the mapillary field to remove the image id again. Now the plus will be disabled with the "already_set" reason until I un-and-reselect the feature. — I don't think this is a relevant use case, so I would leave it as is.

- The same issue is true for the undo action: use `+` to add the image id to a building, use cmd+z to undo the change => now i cannot re-add the image, or the next/prev image (which would be my use case) but instead have to reselect the building first.

tyrasd:
> undo or trash icon […] undo action

good catch, this should now work as expected

> next/prev image

this works for me (already before today's fix commit)?! maybe I'm overlooking an edge case where this is not updating properly? If this still exists: could you maybe open a new issue and list the individual steps to reproduce it?

data/core.yaml
```diff
@@ -788,6 +788,11 @@ en:
       inch: in
     max_length_reached: "This string is longer than the maximum length of {maxChars} characters. Anything exceeding that length will be truncated."
     set_today: "Sets the value to today."
+    set_photo_from_viewer:
+      enable: "Tag this photo id on the currently selected map feature"
```
1ec5, line 792: “ID” is misspelled as “id”.

tordans: https://github.com/openstreetmap/iD/pull/11016

Issue #10038, opened 2023-12-13
Hashtags entered into the hashtag field get lowercased


mxmeinhold:
### How to reproduce the issue?

1. Write a Camel Case hashtag into the Hashtags entry box (e.g. `#ThisIsATest`)
2. Click outside the Hashtags entry box, and see that the hashtag has been lowercased (e.g. now reading `#thisisatest`)

### Screenshot(s) or anything else?

![A screenshot of the iD editor changeset submission page, with a hashtag reading "thisisatest"](https://github.com/openstreetmap/iD/assets/12436574/0f677f00-3de5-4c75-b1d5-dc21db3a4ddb)

This stands out as an issue for me as I'm used to formatting hashtags elsewhere in camel case, so that they're parsable by screen readers

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app, RapiD version at mapwith.ai/rapid

### What version numbers does this issue effect?

2.27.3, 2.1.1, 2.28.0-dev

### Which browsers are you seeing this problem on?

Chrome

1ec5:
The issue lies upstream in the id-tagging-schema repository, which provides the data about tagging conventions that this repository turns into a UI.

The fix is to modify [the file for the `hashtag` field](https://github.com/openstreetmap/id-tagging-schema/blob/c4c57f701c8e11f1d065f6a48ece8adc8e058d37/data/fields/hashtags.json) to set [`caseSensitive`](https://github.com/ideditor/schema-builder#casesensitive) to `true`.

bhousel:
I dunno, when we built this feature, we normalized them to lowercase on purpose. The hashtags field was supposed be for leaderboards like missing maps to connect statistics.

1ec5:
It looks like [osm-stats](https://github.com/AmericanRedCross/osm-stats-workers/blob/de6219edaf5b0ba80b0b71588ff4aace23d92364/src/stats.js#L213-L217) already does case-folding, but I don’t know about OSMCha or Pascal Neis’s tools.

PR #10047, opened 2023-12-23
Modified combo.js file to change restrictTagValueSpelling function to prevent lowercasing the hashtags #10038

sonisumit7904:
Apparently, This function was changing tags according to their Casing,
Removing the toLowerCase() , makes the issue go away!
combo.js
line 94
inside restrictTagValueSpelling

Removed toLowerCase() function

### DIFF
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -91,7 +91,7 @@ export function uiFieldCombo(field, context) {
         }
 
         if (!field.caseSensitive) {
-            dval = dval.toLowerCase();
+            dval = dval;
         }
 
         return dval;


modules/ui/fields/combo.js
```diff
@@ -91,7 +91,7 @@ export function uiFieldCombo(field, context) {
         }
 
         if (!field.caseSensitive) {
-            dval = dval.toLowerCase();
+            dval = dval;
```
1ec5, line 94: Thanks for looking into this issue. This fix happens to address #10038 as described, but it goes too far: most keys that use this combo field do need to be lowercased. I left a note in https://github.com/openstreetmap/iD/issues/10038#issuecomment-1868331583 about a more correct fix.

1ec5 (changes requested):

modules/ui/fields/combo.js
sonisumit7904: @1ec5 Thank you for pointing that out, Since this 1line fix requires changes in another repo, 
can I open a PR in id-tagging-schema?

PR #10062, opened 2024-01-08
Search OSM notes by ID

### LINKED ISSUE #7618: Search OSM notes by ID
Similar to how you can search for nodes, ways and relation ids in the search box, it would be good to have this option for OSM notes as well.

Screenshot_2020-05-19 OpenStreetMap

I don't know if it would even be possible to fetch the highest current note id and use that as an upper limit (plus a margin), but there could be a way for iD to recognise that a note doesn't exist for a given id number?

Same could be done for ways or relations where their id number is much lower than nodes.

laigyu:
Closes #7618

only support number without prefix,
I wonder if needed to treat note as osmEntity, and support one letter shortcut,
but 'n' is taken by node.

noteSearchId

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -176,6 +176,13 @@ export function coreContext() {
       _connection.loadEntityRelations(entityID, afterLoad(cid, callback));
     }
   };
+  // Download single note
+  context.loadNote = (entityID, callback) => {
+    if (_connection) {
+      const cid = _connection.getConnectionId();
+      _connection.loadEntityNote(entityID, afterLoad(cid, callback));
+    }
+  };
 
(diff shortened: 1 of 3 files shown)


tyrasd:
Thanks :+1:  It totally makes sense to be able to search for notes by id as well :+1:

> support one letter shortcut, but 'n' is taken by node.

Yeah, it's a bit unfortunate that `n` is clashing. But I'd say we could support `note` in the id regex in the following line to be able to search for example for `note123` or `note/123`:

https://github.com/openstreetmap/iD/blob/04733726365e9c8e148538e3ac408e00760c1f2d/modules/ui/feature_list.js#L142

tyrasd (approved):

PR #10066, opened 2024-01-13
Expand DMS format and fix sec rounding logic

### LINKED ISSUE #5144: DMS rounding errors, display truncation; expand DMS syntax
Much to my (pleasant) surprise, I discovered that it is in fact possible to paste a DMS lat/lon into the iD search box, because I dug deep enough to find this change, https://github.com/openstreetmap/iD/commit/702e3d2a90e67b7e6ff314f4fd840ef0e2a89881 .

The trouble is as it is now, it seems to accept exactly one (undocumemted) format for such a cut/paste. It is:

[D]DD°MM'SS.[S]"[NS],[D]DD°MM'SS.[S]"[EW]

There are a few issues I've identified:

1. If you enter a location with a negative rather than NSEW, it is actually converted and centered incorrectly:

45°,-90°0'0.5" (to be 45,89°59'59.? W, so it's subtracting wrong somewhere)

2. Location display rounding is performed around certain cases that shouldn't be, as implied by the above:
45°,90°0'0.9" shows as 45°N,90° (seems that < 1 second is just dropped from the display)
45°,90°0'1" shows as 45°N,90°0'1" (that's actually correct!)
45°,90°0'1.1" still autocompletes as 45°,90°0'1" (display not correct because decimals are dropped, but I think the centering is right)

(1) and (2) make it difficult to assume that the entry of a dms style coordinate pair 'works' correctly, either because it's displayed with bad rounding, or it actually centers on the wrong place.

3. I guess this is technically an RFE, but I'll toss it in if people are working on the above as defects anyway:
It would be really awesome if one could paste a dms that looks like this, because it's really common to see done like this:
45 0 18.229944, -90 19 0.4445 (or) 45 0 18.229944 N, 90 19 0.4445 W

laigyu:
re:#5144

try to let  45°,90°0'0.5" display as 45°,90°0'1"

precision error explain: 
if the sec input is 0.5 ,after calculating it will not equal 0.5 but 0.499999999829015,
my soloution is  round it  by 10e-8 to fix it to 0.5
and do the origin display round 0.5=>1

new format
|35  11  10.1 , 136  49  53.8|| D M SS|
|35  11.168 , 136  49.896||D MM|

zoom center loc always use the user input ,
dont matter how display rounding and show,
always center correct.


modules/util/units.js
```diff
@@ -168,3 +171,44 @@ export function decimalCoordinatePair(coord) {
         longitude: wrap(coord[0], -180, 180).toFixed(OSM_PRECISION)
     });
 }
+
+// Return the parsed value  that @mapbox/sexagesimal can't parse
+// return value format : [D, D]  ex:[ 35.1861, 136.83161 ]
+export function dmsMatcher(q) {
+    const matchers = [
+        // D M SS , D M SS  ex: 35 11 10.1 , 136 49 53.8
```
1ec5, line 193: These examples would make for great test cases. Would you mind creating a test/spec/util/units.js that tests this and other edge cases? Thanks!

modules/util/units.js
```diff
@@ -108,31 +108,34 @@ function clamp(x, min, max) {
...
+    var degreesFloor = Math.floor(Math.abs(deg));
+    var min = (Math.abs(deg) - degreesFloor) * 60;
+    var minFloor = Math.floor(min);
+    var sec = (min - minFloor) * 60;
 
-    if (Math.floor(sec) > 0) {
-        displayCoordinate = displayDegrees +
+    var displayCoordinate;
+    var secToFixedDisplayPrecision = 1;
+    var secFixed = Number(sec.toFixed(secToFixedDisplayPrecision));
```
1ec5: The previous code rounded the seconds, so this should similarly round the seconds to the nearest tenth instead of flooring it to the lower tenth.

1ec5 (commented):
Thanks! This is looking pretty good at a glance. Perhaps some tests would be a good idea, since there are several edge cases when we do the rounding manually.

modules/util/units.js
laigyu: It seems tofixed is not a proper way , so I change it to Math.round.
sec is always not precise after calculate from deg, ex: 0.05 after calculate became 0.04999999 , 
so I just round it two time ,try to fix it at the first round,

I will try to write test to test this and dmsMatcher case

modules/util/units.js
```diff
@@ -106,33 +106,51 @@ function clamp(x, min, max) {
...
+    var locale = localizer.localeCode();
 
-    if (Math.floor(sec) > 0) {
-        displayCoordinate = displayDegrees +
+    var degreesFloor = Math.floor(Math.abs(deg));
+    var min = (Math.abs(deg) - degreesFloor) * 60;
+    var minFloor = Math.floor(min);
+    var sec = (min - minFloor) * 60;
+
+    // console.log(sec);
```
tyrasd: please drop commented out debug code :nerd_face:

modules/ui/feature_list.js
```diff
@@ -5,7 +5,7 @@ import * as sexagesimal from '@mapbox/sexagesimal';
 
 import { presetManager } from '../presets';
 import { t } from '../core/localizer';
-import { dmsCoordinatePair } from '../util/units';
+import { dmsCoordinatePair ,dmsMatcher } from '../util/units';
```
tyrasd:
```suggestion
import { dmsCoordinatePair, dmsMatcher } from '../util/units';
```

tyrasd (commented):
Could you please explain the rationale behind adding the additional decimal precision? IMHO, the resolution of the single minutes is already far sufficient for the intended use case (i.e. navigating to a specified location). 0.1 seconds would only translate to a few additional pixels on the screen… A small negative side effect of showing the decimal is that in some locales (e.g. German) this would introduce the output of a decimal comma in the displayed coordinate which would not be supported as a valid input in the search box (e.g. `12°N, 12°3'4,5"W`).

Are the `D M SS` / `D MM` formats really found in practice? I guess it would not hurt to support it additionally for convenience (e.g. when copy-pasting from spreadsheets/csv files with dedicated columns for each coordinate part). I'd agree that it would be required to add some unit tests for these routines. And maybe also add support for it with N/S and E/W suffixes in that format (e.g. `45 0 18 N, 90 19 0 W`)?

modules/util/units.js
laigyu: thanks the review.I didn't consider that. I agree with you , I will change it to no decimal in seconds.

I add D M S because the orgin post want it. and DMM I found it at [google](https://support.google.com/maps/answer/18539?hl=en&co=GENIE.Platform%3DDesktop). I really dont know is common in practice or not.

tyrasd (approved):

tyrasd:
Thanks for the additional test cases!

PR #10067, opened 2024-01-16
Add support  for multiple layers

### LINKED ISSUE #10055: Possibility to use more that one custom layer on ID editor
Hello
I wanted to propose a new functionality that would be very useful for me when mapping, and I found this old post with my same idea. Are there any news about this ?
It's not fundemental for me to save the custom layers in user's profile: it would be enough to me to have 2 or 3 custom layers options, to set up each time I log in pasting the URL. Then, during the mapping work, it woule let me to switch very fast from a layer to another.
Thanks !

            More generally, it would be convenient if we could add as many custom layers as we wish: in France there are at least 3 rendering that might be useful at the same time (IGN aerial photos, cf #3420, Strava heatmap and [BANO](http://tile.openstreetmap.fr/~cquest/leaflet/bano.html)). 
It would mean a bit of UI reworking, but maybe something like that?

capture d ecran de 2017-02-22 19-07-22

_Originally posted by @bagage in https://github.com/openstreetmap/iD/issues/2095#issuecomment-281755531_

Asif-Sheriff:
Fixes #10055
How it works:

1. You have multiple layers in the custom background section
2. you can prefix the layer with '#' to not select it
3. if a layer does not have the '#' prefix then it is selected

Below I have demonstrated how it works.

[Screencast from 15-01-24 05:15:31 PM IST.webm](https://github.com/openstreetmap/iD/assets/130328580/770cd6f0-c4a1-4332-a317-6cc19d7aa918)

This way editors can quickly switch between layers rather than having to copy paste manually every time.

Also added instructions for the same.
image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -965,6 +965,13 @@ en:
             switch: "`{switch:a,b,c}`: DNS server multiplexing"
             quadtile: "`{u}`: quadtile (Bing) scheme"
             scale_factor: "`{@2x}` or `{r}`: resolution scale factor"
+        multiple:
+          multiple_label: "Adding multiple layers"
+          steps:
+            step1: "Type or paste the URLs in the input field, each on a new line."
+            step2: "To include a layer, leave it without any prefix."
+            step3: "To exclude a layer, prefix it with `#`."
+            example_step: "In this case the 2nd layer i.e `https://example-2` will be selected"
         example: "Example:"
       template:
         placeholder: Enter a url template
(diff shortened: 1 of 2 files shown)


tordans:
Personally, I was hoping for a solution where we can have multiple custom layer and store them in the osm user settings.

I am not sure if your proposed changes are really the way to go. Using the `#` is a hacky workaround which already works for people who feel comfortable using it. I don't consider it a good enough workaround to make it the main way to do things.

---

Side note: It is always possible to add layer to [ELI](https://github.com/osmlab/editor-layer-index) so fewer custom layers are required…

Asif-Sheriff:
Hmm ok I'll close this PR then.
I will try to implement this the way you have suggested
also, can you take a look at this PR #10045 ?
Thanks

PR #10089, opened 2024-02-03
Fixed a bug which didn't allow proper comparison of objects

### LINKED ISSUE #9502: Cannot merge lines with same relations in it when the relations have different order
### URL

_No response_

When adding a line with relations (e.g. separate multipoly away from some feature like footpath) I cannot merge the new line with another line containing same relations, when those relations are in different order.
This should normally work as the relation list contains same elements.

example: selected are 2 connected lines (sharing point on top-left):
image

line1 relations:
image

line2 relations:
image

Workaround: Edit the relations so they are in the same order then merge option is active and merge works.
Workaround2: Commit eventual changes (or restart editor) -> the merge option is available (also the relations are now in same order)

To reproduce (simple version): just split any line with 2+ relations, select one of the lines, delete first relation, add the same relation (will be added at the end). Merge the lines (will not work).

_No response_

version 2.24.2

_No response_

_No response_

### LINKED ISSUE #9788: Objects cannot be merged if they belong to same relations but in different order
Here's another one which has irked me for a while: when two adjacent ways are being merged, iD validation would not let me do that if they belong to different relations (so far, so good, although I don't quite like this). However, for the two sets relations to match, iD requires them to be **in the same order** on the object as well, which is simply wrong: this order is basically arbitrary and should not matter anywhere. 

As a simplest example, let's create two consecutive paths and assign them to two hiking route relations: the first one to routes {A, B} and the second one to routes {B, A}. When I select them and try to [C]ombine them, I get:
CantBeMerged
"These features can't be merged since they belong to conflicting relations"

For simple cases it's just annoying (need to reassign relations in the same order), but for highways featuring a high number of e.g. bus route relations this makes merging nigh impossible. 

The fix seems to be simple: internally set the relation sets by name or ID, so that they will match when compared. I wouldn't mind having that sorted in the UI as well -- at least it would make the relation order predictable.

Dimitar5555:
Resolves #9502
Resolves #9788

I know this is hack, but it's the simplest solution to this problem. You could compare key by key, but if you happen to have nested objects/nested arrays of objects, it will quickly turn into recursion hell.

### DIFF
--- a/modules/util/array.js
+++ b/modules/util/array.js
@@ -7,7 +7,8 @@ export function utilArrayIdentical(a, b) {
     var i = a.length;
     if (i !== b.length) return false;
     while (i--) {
-        if (a[i] !== b[i]) return false;
+        // When a and b contain objects, the !== operator will compare their references which may be diffrent even if the object is the same, related #9502
+        if (JSON.stringify(a[i]) !== JSON.stringify(b[i])) return false;
     }
     return true;
 }


Dimitar5555:
Superceeded by #10648

Issue #10074, opened 2024-01-24
Bing Streetside data not showing 


elevenchars:
### URL

https://www.openstreetmap.org/edit#map=17/30.42430/-97.76190

### How to reproduce the issue?

1. Open iD in an area where Bing Streetside data should be visible
2. Enable "Bing Streetside" in map data tab
3. Observe that nothing loads

### Screenshot(s) or anything else?

![image](https://github.com/openstreetmap/iD/assets/11055681/953afcfa-fb98-489b-9c5a-b191f1a70231)

It looks like all of the requests for the bubbles are returning 401.

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.27.3

### Which browsers are you seeing this problem on?

Firefox

elevenchars:
See https://github.com/facebook/Rapid/issues/1305

jjiglesiasg:
![Screenshot (159)](https://github.com/openstreetmap/iD/assets/22849578/61be5101-d091-41ca-8184-c95bcfc1445a)

OSM Data not loading in ID or if load a portion and then move the screen area, no data is shown.

https://www.openstreetmap.org/edit#map=19/-16.50479/-68.13016

k-yle:
@jjiglesiasg that's because the OSM API [has been intermittently unavailable](https://prometheus.openstreetmap.org/d/5rTT87FMk/web-site?orgId=1&refresh=1m&from=1705987783000&to=now&viewPanel=23) over the past 24 hours

dankarran:
I've been seeing this issue recently as well, often getting 401 errors in the console when trying to enable the Streetside layer. It still works occasionally, but generally not.

FinixFighter:
I still have the issue (Brave browser)

dankarran:
I tried creating myself a test key in the [Bing Maps Dev Center](https://www.microsoft.com/en-us/maps/bing-maps/create-a-bing-maps-key), and I still get a 401 when trying to access the API, suggesting this may be an issue with the Bing Maps API as a whole rather than an issue specific to the key that iD (and Rapid) is using.

dankarran:
I was lucky enough to get some StreetSide bubbles on one editing session this morning, but then it stopped working again.

The Bing Maps website is using a different URL to get the StreetSide bubble metadata, so I wonder if they've made some updates to this (apparently undocumented) API:

[...]

PR #10100, opened 2024-02-08
Switch to new (undocumented) Bing Streetside bubble metadata URL

dankarran:
Bing Streetside bubbles have often been missing recently, with the bubble metadata requests coming back with 401 errors. Occasionally, enabling the Streetside bubbles does work, but more often than not they don't show up.

This issue was first reported in the iD issue queue about two weeks ago (#10074) and also in the Rapid editor issue queue around the same time.

Per my comment in the issue, I noted that the Bing Maps website was now using a different URL, so I went poking around. I can't find either the old URL or the new URL documented anywhere, but switching to the new URL seems to work fine, and so far feels more reliable. The JSONP callback doesn't seem to be available with the new URL, so I dropped that and replaced it with a request method similar to the one in the Mapillary service.

Whether or not this is the right fix, I don't know. Probably needs discussing with someone at Microsoft?

### DIFF
--- a/modules/services/streetside.js
+++ b/modules/services/streetside.js
@@ -7,7 +7,6 @@ import {
 
 import RBush from 'rbush';
 import { t, localizer } from '../core/localizer';
-import { jsonpRequest } from '../util/jsonp_request';
 
 import {
   geoExtent, geoMetersToLat, geoMetersToLon, geoPointInPolygon,
@@ -17,7 +16,7 @@ import {
 import { utilArrayUnion, utilQsString, utilRebind, utilStringQs, utilTiler, utilUniqueDomId } from '../util';
 
 
-const bubbleApi = 'https://dev.virtualearth.net/mapcontrol/HumanScaleServices/GetBubbles.ashx?';
+const bubbleApi = 'https://t.ssl.ak.tiles.virtualearth.net/tiles/cmd/StreetSideBubbleMetaData?';
 const streetsideImagesApi = 'https://t.ssl.ak.tiles.virtualearth.net/tiles/';
 const bubbleAppKey = 'AuftgJsO0Xs8Ts4M1xZUQJQXJNsvmh3IV8DkNieCiy3tCwCUMq76-WpkrBtNAuEm';
 const pannellumViewerCSS = 'pannellum/pannellum.css';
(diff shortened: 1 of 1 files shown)


tyrasd:
Hey, thanks for digging into this! :bow:  However, I'm not quite sure if switching from one undocumented broken API to a slightly different undocumented endpoint is really going to fix this in a stable way.

Did you hear from Microsoft that using the undocumented API at `https://t.ssl.ak.tiles.virtualearth.net/tiles/cmd/StreetSideBubbleMetaData` URL would be fine for OSM editors?

When I last talked to someone at Microsoft about this topic, it was recommended that we should not use these undocumented APIs and instead migrate to the official, documented [Bing Maps streetside imagery API](https://learn.microsoft.com/en-us/bingmaps/rest-services/imagery/get-imagery-metadata#get-streetside-metadata-centered-at-a-point) (cc @mbrzakovic). I just looked at my notes and it seemed that the only outstanding issue was whether we'd need a dedicated API key for that, as the one which we're using for the undocumented API does not work for the "proper" API.

As Bing's streetside imagery is currently now working in iD, I'm going to see if this can be done in a relatively quick way. Closing here for now.

dankarran:
Thanks for that @tyrasd. I haven't heard back from Microsoft since their original acknowledgment last week. If there's an official API that works for this though, that sounds like a better solution anyway.

By the way, I've just tried the streetside bubbles in the live iD editor, and got 401 errors again, so it seems like it's not working reliably, at least for me. Maybe it depends which IP the requests come from, so works for some but not for others?

streetside-401s

tyrasd:
oh, I had accidentally used a non-https URL for the API endpoint (copy-pasted it and didn't get the error locally as my local dev setup is not running on https). this should be fixed in 17d1c27bec0cc44b53e450532434297849e818a

PS: by "live iD editor": do you mean the one on osm.org (your screenshot still has the old URL from before the fix…)? That instance does not have the fix yet (it requires to do a full release to updated that instance). But you should be able to test the development version on https://ideditor.netlify.app/.

dankarran:
> PS: by "live iD editor": do you mean the one on osm.org (your screenshot still has the old URL from before the fix…)? That instance does not have the fix yet (it requires to do a full release to updated that instance). But you should be able to test the development version on https://ideditor.netlify.app/.

Ah, I misunderstood, sorry - I thought the old version was still working for you, even without your fix. Your new version looks good, and the bubbles work for me there 👍

tsmock:
@tyrasd : What was the process like for getting a new API key (or is it one for OSM editors in general)? I was looking at fixing a similar problem in JOSM Microsoft Streetside plugin, but it appears that the old API key only works with the undocumented API.

tyrasd:
@tsmock It was recommended to me to re-use the key we use to fetch the Bing aerial imagery metadata (`https://dev.virtualearth.net/REST/v1/Imagery/Metadata/AerialOSM?…`). I would assume that this should also work for the key you have in JOSM for that endpoint. @mbrzakovic maybe you can confirm this for us?

mbrzakovic:
Apologies for late reply,

@tyrasd you can re-use the existing iD key for Bing streetside imagery as well.
Great job on transitioning to new api - it was the right and recommended move.
 
@tsmock there is no specific treatment to account/key that josm is using (if it works for Aerial imagery, it should work for Streetside as well). Please write us at osm@microsoft.com with some more technical specifics and we will try to resolve the issue you are experiencing.

tsmock:
Sorry, I forgot to reply to this issue when I fixed Microsoft Streetside ~2 weeks ago.

As of [r36228](https://josm.openstreetmap.de/changeset/36228/osm), the plugin is now using the official API (I have no clue what API it was using previously, probably an "internal" API or something), and the key for background imagery works with that.

PR #10117, opened 2024-02-21
Fractional Zoom Added

### LINKED ISSUE #10091: Use fractional zoom levels in the URL bar
We see at zoom 19 points are present:

20240205T123954

Now we roll the mouse wheel back and forth a little,
and we see at zoom 19 points are *not* present,

20240205T123920

When we rolled the mouse wheel, the URL's coordinates updated, but not the zoom level.

Now if we use the + and - buttons on the side panel, we get more concrete results, the official 19.0 zoom level view.

So what is the problem?

Well if we share one of the above URLs, we might see something that the person we shared it to can't, or vice versa.

I think Google Maps solves this by using 19.3 etc. zoom levels.

SharanRP:
closes #10091


mxdanger:
I recommend limiting the scope of the pull solely to the fractional zoom rather than loader icon.

PR #10119, opened 2024-02-22
fix bug causing a spam of network requests

### LINKED ISSUE #9516: Constant requests for the same file when zooming/panning
### URL

Any url

1. Go to any place using iD
2. Open the developer tools
3. Select the network tab
4. Add a filter so that only XHR requests are shown
5. Start panning or zooming in and out on the map
6. Notice the repeating requests for `imagery.min-(random string).json`

It may not be a big problem since the file is being fetched from the cache but it's definitely something that can "clog" the main thread on lower end PCs (keeping in mind that this file is 2.55 MB and needs to be parsed every time). 

2.24.2

Firefox

k-yle:
Closes #9516

This bug meant that the cache was never used, so whenever you pan the map, several assets were constantly being re-requested

### DIFF
--- a/modules/core/file_fetcher.js
+++ b/modules/core/file_fetcher.js
@@ -67,7 +67,7 @@ export function coreFileFetcher() {
           return getUrl(url.replace('{presets_version}', presetsVersion), which);
         });
     } else {
-      return getUrl(url);
+      return getUrl(url, which);
     }
   };
 


tyrasd (approved):
Thanks! This was a regression from 18b50ccd6da3830abc4ad8c829d9c0ca43ebdfec :see_no_evil:

PR #10120, opened 2024-02-22
fix multipolygon nodes unable to be moved

### LINKED ISSUE #9777: Allow to adjusting nodes of multipolygon objects
### Description

Hi, 

I would propose to adjust some of the settings that make working with multipolygon buildings easier. 

Let's say that you have two areas and you combined them into multipolygon and assigned a type to this new object. In my case I've created two building-parts and then combined them into the building using multipoylgon:

object 1:
```
building=yes
type=multipolygon
```
object 2:
`building:part=yes`
object 3:
`building:part=yes`

Here is an image of my object:
image

Now when I try to move any free nodes or linked nodes in the same relation you will get an error: 
`Areas cannot cross over themselves.`
Picture below illustrate which edges of the multipolygon I was trying to move:
image

Desired behaviour:
- You should be able to move free nodes of the areas even if they are part of the bigger relation (at least for buildings). 
- You should be able to move linked nodes if they are part of the same relation. And if both objects are only in single relation. I.e. in my case node is shared between two children of single multipologyon builidng.
- 
image

### Screenshots

Screenshot 2023-07-14 at 13 32 02
Screenshot 2023-07-14 at 13 26 22
Screenshot 2023-07-14 at 13 33 36

### LINKED ISSUE #9951: Spurious "Areas cannot cross over themselves."
### URL

https://www.openstreetmap.org/edit#map=20/47.49719/9.74947

iD bars the user from moving the vertex of an area by "Areas cannot cross over themselves". The area does not seem to exhibit any particular issues from visual inspection. If there is some sort of obscure topological artefact somehow which causes it to cross into itself, then it's already there and hasn't been prevented earlier.

Since this blocks all edits, the situation can not be resolved without deleting the entire polygon.

There should be a way to bypass the (faulty?) detection.

image

None of the selected vertices can be moved. Others can be moved.

2.27.1

Firefox

### LINKED ISSUE #10043: iD does not allow this node to be moved and does not give an indication as to why
### URL

https://www.openstreetmap.org/edit?node=5693640329#map=19/29.92386/-95.67725

1. Open the link provided, try to move the selected node.
2. Notice that it snaps back to its original position every time you do so, and doesn't tell you why.

image

_No response_

2.27.3

Firefox

k-yle:
Closes #9777, Closes #9951, Closes #10043

Fixes an edge case where iD was preventing nodes from being moved, even though the geometry is officially valid.

image
screenshot courtesy of @Sirorezka


tyrasd:
> added https://github.com/openstreetmap/iD/labels/waitfor-consensus because of https://github.com/openstreetmap/iD/issues/9777#issuecomment-2155787414.

:thinking: 

From what I can see, according to the [wiki](https://wiki.openstreetmap.org/wiki/Relation:multipolygon), touching outer polygons should not be considered valid:

> Valid multipolygon conditions
> Generally, the multipolygon relation can be used to build multipolygons in compliance with the OGC Simple Feature standard ([Graphical examples of OGC validity](http://postgis.refractions.net/docs/using_postgis_dbmanagement.html#OGC_Validity)). **Anything that is not a valid multipolygon according to this standard** (e.g., polygons with intersecting rings) **should also be considered an invalid multipolygon relation**, with the notable exception of touching inner rings

And the [OGC Simple features specifications](https://portal.ogc.org/files/?artifact_id=25355) do not allow outer rings to have common segments (section 6.1.14):

> The boundaries of any 2 Polygons that are elements of a MultiPolygon may not “cross” and may touch at only a finite number of Points.

tyrasd:
That said, I think in the editor we should probably always strive towards:

* being strict when it comes to the creation of an invalid data as the result of an edit/operation
* being lenient when editing existing data that is already invalid; especially we should not disallow edits that are not related to how the respective feature is "broken"
* offering means to fix features that are existing in an invalid state

In this particular case, this would mean: 

* we should not allow the snapping operation of different outer rings (except at singular points?)
* allow to drag vertices of a multipolygon, even if there are touching outers or a self-overlap somewhere, as long as the drag does not introduce an additional invalid new touching/crossing rings
* allow to disconnect outer (and inner) rings of a multipolygon from each other

modules/modes/drag_node.js
```diff
@@ -327,6 +327,10 @@ export function modeDragNode(context) {
                 for (k = 0; k < rings.length; k++) {
                     if (k === activeIndex) continue;
 
+                    // if the node being dragged belongs to both rings,
```
tyrasd, line 330: btw, this does not cover all cases, as also the vertices before/after the node that belongs to both rings cannot bot moved with the current implementation (as they belong to the same ring segment).

matkoniecz: @k-yle Can you followup on this?

Should code be changed? Comment noted that some cases are excluded?

matkoniecz:
> added waitfor-consensus because of https://github.com/openstreetmap/iD/issues/9777#issuecomment-2155787414.

commented there

matkoniecz:
In this case, I think, we should not care about whether geometry is valid.

If it invalid allowing edits is fine to enable user to make it less invalid.

If it is valid then we should not lock editing and claiming it is invalid.

PR #10121, opened 2024-02-22
set the locale of the login popup

### LINKED ISSUE #3595: localized authentication
If iD is loaded using the .locale('fr') Context option, the authentication should also happen in the same language. Currently, if ID is loaded in French, as an example, when doing edit->save->authenticate, the login OSM popup window is in English.

In the context of Governmental use of iD in jurisdictions with more than one official language, this is a big issue. Is there a way to modify this behaviour?

k-yle:
Closes #3595

We now send the locale code that iD is using to the OAuth popup, so that the user always sees their preferred language. Same idea as #9501

This will have no effect until https://github.com/osmlab/osm-auth/pull/130 is merged and released


k-yle:
https://github.com/openstreetmap/iD/labels/waitfor-upstream can be removed , see #10223

Issue #9695, opened 2023-06-09
Add a new layer for survey points


Dimitar5555:
### Description

It would be nice if there is a new data layer with survey points which is disabled by default. It would also be nice if survey points are excluded from the general "Points" data layer. That way new mappers won't change/move/delete them.

RAytoun:
@Dimitar5555   **That would be ideal.** There are no fixed anchor points on OpenStreetMap so nothing can be positioned with certainty.

Most international boundaries use coordinates to legally identify their position. 

Other fixed points are [triangulation stations](https://en.wikipedia.org/wiki/Triangulation_station) identified by a solid structure.

In cartography the coordinate system has been the basis for fixing positional data accurately on official documents and maps.

The very notion that anyone can move an international boundary if they so wish makes me cringe.

jidanni:
All I know is I spent a lot of effort researching where
https://www.openstreetmap.org/node/10171213029
belongs, and intend to nail down the rest of them
dragging the border with it, in the OSM inaccurately demarcated
sections in that river, based on https://www.internationalboundarycommission.org/ data,
one day. Somebody's got to fix that border.

tyrasd:
see also #9255 and https://github.com/openstreetmap/iD/pull/10122#issuecomment-1984069497

Ideally, there would be a dedicated feature layer for features such as survey points (or admin boundaries) that is deactivated by default, but still visible on the map in a grayed out state (see also https://github.com/openstreetmap/iD/issues/9714#issuecomment-3101811495). That way they would still be visible, but users could not accidentally move them or connect them without actively enabling the layer.

PR #10122, opened 2024-02-22
prevent survey markers from being moved

### LINKED ISSUE #9255: survey_point shouldn't be moved
### URL

https://www.openstreetmap.org/edit?editor=id&node=669983383#map=19/49.75438/4.87604

1. Open the link
2. Read the misleading proposed "correction"

As stated in the note (for survey points in France): _Ne pas déplacer ce point, cf. - **Do not move this node**, see - http://wiki.openstreetmap.org/wiki/WikiProject_France/Repères_Géodésiques#Permanence_des_rep.C3.A8res_
(...)_two geodetic points at the same position (latitude, longitude) should not be merged, their altitudes (ele) being different._

More generally, two points at the same geographical position (latitude, longitude) but different elevation (ele) should not be candidate for a move: they are at the same geographical position by construction/design.

2.21.1

Firefox

### LINKED ISSUE #9867: Hide/lock survey points
### Description

I've seen cases where mappers liked a survey point and decided to use it for some other object (like a church)... It would be nice if these points are always hidden (with no option to see them) or if their location and tags were immutable in iD.

### Screenshots

_No response_

k-yle:
Closes #9255, Closes #9867

### DIFF
--- a/modules/modes/drag_node.js
+++ b/modules/modes/drag_node.js
@@ -232,6 +232,9 @@ export function modeDragNode(context) {
             isInvalid = hasInvalidGeometry(entity, context.graph());
         }
 
+        if (entity.type === 'node' && entity.isLocked()) {
+            isInvalid ||= 'locked';
+        }
 
         var nope = context.surface().classed('nope');
         if (isInvalid === 'relation' || isInvalid === 'restriction') {
@@ -243,6 +246,11 @@ export function modeDragNode(context) {
                         { relation: presetManager.item('type/restriction').name() }
                     ))();
             }
+        } else if (isInvalid === 'locked') {
+            context.ui().flash
+                .duration(3000)
+                .iconName('#iD-icon-no')
+                .label(t.append('operations.move.locked.single'))();
         } else if (isInvalid) {
             var errorID = isInvalid === 'line' ? 'lines' : 'areas';
             context.ui().flash
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -65,6 +65,15 @@ export var osmAreaKeysExceptions = {
     }
 };
 
+/** iD will not allow nodes with this tag to be moved */
+export const osmTagsPreventingMove = {
+    man_made: {
+        survey_point: true
+    },
+    latitude: true,
+    longitude: true,
+};
+
 // returns an object with the tag from `tags` that implies an area geometry, if any
 export function osmTagSuggestingArea(tags) {
     if (tags.area === 'yes') return { area: 'yes' };
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -340,6 +340,9 @@ en:
         feature:
           one: "Moved a feature."
           other: "Moved {n} features."
+      locked:
+        single: This type of feature has tags that prevent it from being moved.
+        multiple: Some of these features have tags that prevent them from being moved.
       incomplete_relation:
         single: This feature can't be moved because it hasn't been fully downloaded.
         multiple: These features can't be moved because they haven't been fully downloaded.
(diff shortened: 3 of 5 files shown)


modules/osm/tags.js
```diff
@@ -65,6 +65,15 @@ export var osmAreaKeysExceptions = {
     }
 };
 
+/** iD will not allow nodes with this tag to be moved */
+export const osmTagsPreventingMove = {
+    man_made: {
+        survey_point: true
+    },
+    latitude: true,
+    longitude: true,
```
k-yle, line 74: I'm not sure if encouraging these two keys is a good idea, since most appear to be introduced by bodged imports

modules/modes/drag_node.js
```diff
@@ -232,6 +232,9 @@ export function modeDragNode(context) {
             isInvalid = hasInvalidGeometry(entity, context.graph());
         }
 
+        if (entity.type === 'node' && entity.isLocked()) {
```
1ec5, line 235: Would it be feasible to allow the drag when some modifier key is held down, like Alt/Option, when the mapper is sure they know what they’re doing? If not, another approach would be to create a filter for things like survey points and hide it by default, as we do with boundaries.

/ref https://github.com/openstreetmap/iD/issues/9255#issuecomment-1224954446

modules/osm/tags.js
1ec5: Yeah, I could see this legitimizing the two keys as a “one weird trick” for locking a feature. Imagine someone pranking other mappers by adding a `longitude` to all the POIs they map. It would never show up in the fields, only the raw tags, hidden by default.

1ec5 (commented):
Should we apply any styling to the point on the map, so that it looks locked before the user tries to futz with it?

https://github.com/openstreetmap/iD/blob/b5f1f78e8701e92b7b726c3ef70de9242f681da3/css/20_map.css#L296-L304

Will mappers get annoyed that they can’t move the point and delete it instead? A `wikidata=*` tag locks a feature from getting deleted too.

k-yle:
All your suggestions seem like good ideas, I'll wait and see if there are any other comments before making changes.

> Would it be feasible to allow the drag when some modifier key is held down, like Alt/Option, when the mapper is sure they know what they’re doing?

I'm pretty sure it's possible since we have access to [`MouseEvent#altKey`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/altKey) in both contexts

tyrasd (commented):
This looks like a reasonable behavior for iD. Yes, not every survey point mapped in OSM was imported from a source with authoritative precision, but I would say that accidentally moved survey points are a bigger issue than misplaced survey points which are not correctable easily in iD. An override for advanced users to still edit the location of such points would be nice to have, but a relatively easy workaround is to just change the feature temporarily to a different preset while moving.

I'd also exclude `latitude/longitude` from the tags for this feature. :+1: 

As far as I see, it does not fully fix the issue reported in #9255, as the validation warning is still shown. Perhaps the validation should not be triggered when all of the affected entities are _locked_.

1ec5:
> Yes, not every survey point mapped in OSM was imported from a source with authoritative precision, but I would say that accidentally moved survey points are a bigger issue than misplaced survey points which are not correctable easily in iD.

True, it isn’t a big inconvenience in the grand scheme of things. However, the lock design seems superficially like the Wikidata lock, except that it locks something other than a field that can actually get in the way when mapping – and it doesn’t protect the feature from getting deleted. Will we consider using the same affordance for anything else?

Conceptually, I don’t see a difference between these survey points and other things that non-experts should think twice before touching, such as boundaries and (takes a deep breath) abandoned railways. A filter would be a more robust solution to this problem, because the user has an opportunity to keep immovable features from cluttering the map as they map other types of features around them. A filter wasn’t an option for Wikidata-tagged features because that’s a secondary attribute.

tyrasd:
As mentioned above: making specific points read-only could be to extreme of a solution for this problem.

> A filter would be a more robust solution to this problem

:+1: See #9695 for that suggestion.

PR #10124, opened 2024-02-22
tel no. validation

### LINKED ISSUE #9559: When pasting telephone links, drop tel: prefix
### Description

It'd be nice if, when pasting a telephone number copied from a telephone link, which starts with `tel://`, that initial part be already dropped off so you don't have to manually erase it.

### Screenshots

Telephone pasted with tel:// prefix

SharanRP:
Closes #9559 - Implemented a regular expression to restrict the pasting of alphabetical characters into the input field.


pnorman:
There seem to be changes unrelated to telephone numbers here. Could you also title your PR and add a comment so that it's obvious what it's for?

SharanRP:
I have updated the files and body of message

SharanRP:
Is this correct now ? @pnorman

modules/ui/fields/input.js
```diff
@@ -404,16 +407,19 @@ export function uiFieldText(field, context) {
...
-
     function change(onInput) {
         return function() {
             var t = {};
             var val = utilGetSetValue(input);
-            if (!onInput) val = context.cleanTagValue(val);
 
-            // don't override multiple values with blank string
+            if (field.type === 'tel') {
+                val = val.replace(/[^0-9+]/g, '');
```
1ec5: This appears to be stripping anything that isn’t a digit or plus sign from the tag value. I don’t think this is consistent with what the `phone` key is supposed to be set to in most regions, at least according to [this documentation](https://wiki.openstreetmap.org/wiki/Key:phone). The linked issue appears to be more limited in scope: it would only take effect if the field’s value resembles a tel: URI. But tel: URIs can certainly contain non-digits.

SharanRP: Yes, I wasn't aware of that previously. So, what specifically should be disregarded?

1ec5: If I understand correctly, https://github.com/openstreetmap/iD/issues/9559#issue-1634183847 was only asking to strip the `tel:` or `tel://` prefix if present.

SharanRP:
code updated to only neglect tel: or tel:// @1ec5

modules/ui/fields/input.js
```diff
@@ -413,7 +413,8 @@ export function uiFieldText(field, context) {
             var val = utilGetSetValue(input);
 
             if (field.type === 'tel') {
-                val = val.replace(/[^0-9+]/g, '');
+                val = val.replace(/^tel:\/\//, ''); // Remove 'tel://' prefix if present
+                val = val.replace(/^tel:/, ''); // Remove 'tel:' prefix if present
```
1ec5, line 417: Since these are regular expressions, you could combine these two steps by surrounding the slashes in `(\/\/)?`. (Technically, RFC&nbsp;3966 doesn’t allow slashes after the colon, but as https://github.com/openstreetmap/iD/issues/9559#issuecomment-1489139026 demonstrates, there’s probably enough confusion about that to go ahead and strip out those slashes too.)

modules/ui/fields/input.js
```diff
@@ -404,16 +407,20 @@ export function uiFieldText(field, context) {
...
+            if (field.type === 'tel') {
+                val = val.replace(/^tel:\/\//, ''); // Remove 'tel://' prefix if present
+                val = val.replace(/^tel:/, ''); // Remove 'tel:' prefix if present
+            }
+    
+            if (!onInput) val = context.cleanTagValue(val);
+ 
             if (!val && getVals(_tags).size > 1) return;
-
+
```
1ec5, line 423: We should probably perform the `tel:`-stripping after cleaning the tag value. `cleanTagValue()` removes stuff like leading and trailing whitespace, something that can easily occur if someone is copy-pasting from another website.

modules/ui/fields/input.js
```diff
@@ -50,6 +50,9 @@ export function uiFieldText(field, context) {
     }
 
 
+    
+
+
```
matkoniecz, line 55: why some random new lines are added?
```suggestion
```

matkoniecz:
It looks like it is an abandoned PR and should be closed as such.

Issue #9439, opened 2022-12-25
Walkthrough bug: prompt suggests searching for 'cafe' feature type after selecting non-point shape during walkthrough


jayaddison:
### URL

https://www.openstreetmap.org/edit

### How to reproduce the issue?

1. Open the 'Help' menu (shortcut: 'H')
2. Begin the editing walkthrough
3. Select the 'Points' walkthrough section
4. Click the 'Point' icon to create a new point
5. Click on the border of the building that appears in the middle of the transparent area
6. Search for the word 'Cafe' as instructed by the tutorial

Bug: 'Cafe' does not appear in the feature type selection dropdown (the top result is 'Cape')

### Screenshot(s) or anything else?

![cafe](https://user-images.githubusercontent.com/55152140/209480040-2358db6e-4253-45ac-8f29-d9c9eea6b511.png)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app, RapiD version at mapwith.ai/rapid

### What version numbers does this issue effect?

2.23.2 (iD), 2.24.0-dev (iD), 1.1.9 (RapiD)

### Which browsers are you seeing this problem on?

Firefox

jayaddison:
> Click on the border of the building that appears in the middle of the transparent area

It seems like this was the problem here.  Creating a point within an unfilled region of the transparent area (where no border or other feature exists) results in a point being created, and `Cafe` can, as expected, be found and selected as the type for that feature.

If a bug remains here, then it is that the tutorial continues following a click on a border segment (way?).

nickrsan:
I'm able to reproduce this issue. Adding the point to a way renders "Cafe" unfindable in the presets and the tutorial can't proceed, but adding it elsewhere makes it findable.

This issue should be a quick fix by changing tutorial text to tell people not to click on an existing line (maybe more complex if we want it to detect that the point is on a line). I'm going to mark this as a good first-time contribution issue.

StillAbeginnerr:
Is it open? I want to contribute. @jayaddison

riQQ:
@tyrasd this can be closed, right? It wasn't closed automatically as the PR wasn't linked properly.

PR #10129, opened 2024-02-25
Fix for Point Tutorial

mattiapezzotti:
Fix for Issue #9439:
- user now cannot place a point on an edge
- a warning is shown if done so


tyrasd (approved):
Great, thank you!

jayaddison:
@mattiapezzotti I can confirm from using the recent v2.29.0 release of the iD editor that this fixes the problem encountered in #9439.  Thank you very much / grazie!

PR #10138, opened 2024-02-28
show more relevant presets when switching preset of an object

tyrasd:
When changing the preset of an existing object, the preset list currently contains a few of the recently used presets and a relatively long "default" list of presets which match the selected feature's geometry. Listing the completely context-less list of "defaults" is in general not very useful: it is pretty unlikely that one wants to change a `building` feature to something completely different like a park for example.[^1] Usually one wants to change an object's preset to either refine it (choosing to a more specific sub-preset, e.g. from a Doctor to a General Practitioner), or update the map by choosing a closely related preset (e.g. when switching from a `shop=shoes` to `shop=clothes` if a POI changed in reality or the POI was previously mapped incorrectly).

[^1]: There is a little bit of merit to not completely omit showing these defaults, tough: It teaches the mapper that there are many very different thing one can map in OSM: e.g. from natural features, landuse and water features, to points of interest, etc.

In this PR, the suggestions are based on the object's previous preset:
* if it was part of a category, show that category again (e.g. editing a `highway=primary` will offer the "Major Roads" category)
* if a preset has sub-presets: offer up to 5 of these (e.g. `barrier=kerb` will offer presets like Raised Curb, Lowered Curb, etc.)
* show the "partent" i.e. more generic preset if there is a matching one
* show up to 5 "sibling" presets

(currently this also increases the number of shown previously used presets from 4 to 5, see #9545)

Some unfinished business:
* [ ] should these be sorted differently, e.g. before the previously used presets)?
* [ ] should there be a visual separation between these suggestions, the recents and the defaults?
* [ ] investigate if a better sorting of candidates could be done when many siblings exist (e.g. `building=yes` might more likely be changed to a common building type like `building=house`, compared to something rare like `building=yurt` for example)
* [ ] add mechanism to offer more related presets (and/or more recently used presets)
* [ ] add test cases

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -9,7 +9,7 @@ import { presetCategory } from './category';
 import { presetCollection } from './collection';
 import { presetField } from './field';
 import { presetPreset } from './preset';
-import { utilArrayUniq, utilRebind } from '../util';
+import { utilArrayUniq, utilEditDistance, utilRebind } from '../util';
 
 export { presetCategory };
 export { presetCollection };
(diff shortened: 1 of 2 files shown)


matkoniecz:
>  should there be a visual separation between these suggestions, the recents and the defaults?

I see no reason to do this, I expect that it would just add noise

>  should these be sorted differently, e.g. before the previously used presets)?

maybe intermixed with them? And treat it as a boost? For example if something is both matching this new ranking and was used recently it can go to top

Issue #10135, opened 2024-02-27
Mapillary viewer link to viewed image


tordans:
### URL

https://github.com/facebook/Rapid/issues/1337

### How to reproduce the issue?

Something somewhere changed, and now there is no direct link to the Mapillary Image on mapillary.com anymore from the image viewer inside iD/Rapid.

<img width="360" alt="image" src="https://github.com/openstreetmap/iD/assets/111561/a6d52d5e-3f07-4cff-a68f-4da1301e82bc">

(The mouse hovers the Mapillary Logo which AFAIK was the direct link once but now opens the Mapillary homepage.)

There is an issue at https://github.com/facebook/Rapid/issues/1337 which tracks the same topic.

Sushil642:
@tordans 
 This issue was happening just because of anchor tag link.I made changes and now the mapillary logo work as direct link to 
mapillary image on mapillary.com.
I submitted the PR #10141 .Please approve my Pr.

tyrasd:
Turns out this is a bug in the upstream library. iD _might_ be able to hot-patch the output produced by the mapillary-js library to work around this, but it would be better if this could be fixed at the root.

tordans:
@tyrasd It might not be a bug but the feature changed upstream. @eneerhut wo created https://github.com/facebook/Rapid/issues/1337 is from Mapillary (https://github.com/eneerhut) so I assumed by the fact that he reported it the issue should be solved by the editor(s).

Update: Having read https://github.com/openstreetmap/iD/pull/10141#issuecomment-1972783620 more carefully I agree that this should be fixed in https://github.com/mapillary/mapillary-js – FYI @eneerhut.

tyrasd:
Hmm, wait… You might be right, actually. :see_no_evil:  To me this looked like an oversight, as the library does produce the "seemingly correct" image link on the mapillary logo when it is using the code path for the case where the username is not present ([here](https://github.com/mapillary/mapillary-js/blob/a1df915/src/component/attribution/AttributionComponent.ts#L146)). But what I had missed is that the function that is called with the username present, there is an additional (deep) link produced. But for some reason that link is however currently [hidden](https://github.com/openstreetmap/iD/commit/7a0d8adb7b074cbf4a40cfbfe47a3f0cfc09ddaf#diff-56fc787113948ee3ddbf3d4979def5dbe8a49fc58cd98ebfd25c8ef364d4fa93R319) in iD.

[...]

PR #10141, opened 2024-02-29
Fix: Mapillary viewer link to viewed image

Sushil642:
Fix for issue:-#10135

  In this issue, I have made changes in a function  **_makeCreatorBy( )**  of dist/mapillary-js.
  I changed the anchor tag link of class  "mapillary-attribution-icon-container"  
  to **this.makeImageUrl(exploreUrl,imageId)**. Now it is working properly.
id


Sushil642:
@tyrasd  please approve my PR.

tyrasd:
Thanks for digging into this. :bow:

Note that the `dist` directory is not meant to be modified (as it contains the automatically generated "compiled" files produced by iD's build scripts, whose contents are overridden every time one runs `npm run dist`).

~~It appears that the bug is an upstream issue in the mapillary-js npm package at https://github.com/mapillary/mapillary-js/blob/e62378b6bc03cfbbe26c5f6349a39f0b250dc534/src/component/attribution/AttributionComponent.ts#L95 – @Sushil642 could you please opening a  pull request with your change on that repository? Thanks!~~

tyrasd:
ok, after giving this a second look (see https://github.com/openstreetmap/iD/issues/10135#issuecomment-1972877066), it seems that my initial evaluation that this is an upstream issue was incorrect: mapillary-js does produce also a deep link to the image already, but that link is currently hidden by iD (https://github.com/openstreetmap/iD/commit/7a0d8adb7b074cbf4a40cfbfe47a3f0cfc09ddaf#diff-56fc787113948ee3ddbf3d4979def5dbe8a49fc58cd98ebfd25c8ef364d4fa93R319). It would be best to just fix this in the css such that the link to the mapillary image does show up again (and is properly styled).

tyrasd (changes requested):
ok, after giving this a second look (see https://github.com/openstreetmap/iD/issues/10135#issuecomment-1972877066), it seems that my initial evaluation of this being an upstream issue was incorrect: mapillary-js does produce also a deep link to the image already, but that link is currently hidden by iD (https://github.com/openstreetmap/iD/commit/7a0d8adb7b074cbf4a40cfbfe47a3f0cfc09ddaf#diff-56fc787113948ee3ddbf3d4979def5dbe8a49fc58cd98ebfd25c8ef364d4fa93R319). It would be best to just fix this in the css such that the link to the mapillary image does show up again (and is properly styled).

Sushil642:
@tyrasd 
Both Mappilary-logo and username should serve as direct link to Mapillary image in mapillary.com?

tordans:
My understand is this is now about removing the css rule that hide the link.

Sushil642:
@tyrasd 
  I made the changes in css file "60_update.css"  and now username is visible.
🥑please review my PR.

tordans:
Please restore the unrelated file changes and then squash and force push so we have a clean git history. 
Can you also share a screenshot?

Sushil642:
@tordans 
I have restored all changes✅
Here what changes i have made ⬇️
css-issue

tyrasd (approved):

tyrasd:
@Sushil642 I think @tordans meant a screenshot of the mapillary viewer after the fix. There's no need to post screenshots of your code, as everyone can see that in the code tab of the pull request here on github already. :wink: 

Here's how it looks now:

small version <-> expanded version

css/60_photos.css
```diff
@@ -381,7 +381,9 @@ label.streetside-hires {
 }
 
 .mly-wrapper .mapillary-attribution-container .mapillary-attribution-username {
-    display: none;
+    display: block;
```
tordans, line 384:
```suggestion
```

A DIV ist always 'block' unless changed somewhere (which it is not).

css/60_photos.css
```diff
@@ -381,7 +381,9 @@ label.streetside-hires {
...
+    padding: 0px 8px 0 6px;
+    margin-bottom: -2px;
```
tordans, line 386: It looks like the `margin-bottom` does nothing?

css/60_photos.css
```diff
@@ -381,7 +381,9 @@ label.streetside-hires {
 }
 
 .mly-wrapper .mapillary-attribution-container .mapillary-attribution-username {
```
tordans, line 383: It looks like this for me image

We need to also set the following, so that the height for the name has the same rules that `.mapillary-attribution-date` uses.

With the height-reset it looks like this image (I did not add the padding in this example, yet)

```suggestion
.mapillary-attribution-compact .mapillary-attribution-image-container {
  height: auto;
}

.mly-wrapper .mapillary-attribution-container .mapillary-attribution-username {
```

tordans (changes requested):

css/60_photos.css
tyrasd: Thanks, that works much better! updated bf184e21c

PR #10146, opened 2024-03-03
Add Tracestrack Localized Topo Universal imagery

quinncnl:
Hi,

I'm from Tracestrack Maps and I'd like to provide the Topo map as an imagery to help mappers.


LaoshuBaby:
Wow, thank you for geneous grant your tiles' access to iD, but I guess iD's imagery index are build from editor-layer-index project?

https://github.com/openstreetmap/iD/blob/cc920d3fbade3603be679f83566adc5551287a33/FAQ.md?plain=1#L17-L19

https://github.com/osmlab/editor-layer-index/blob/9d718bed470f65b17d49b45c0e2f09093f013626/README.md?plain=1#L24

(I'm not the developer of iD so the above is just guessed)

tordans:
> iD's imagery index are build from editor-layer-index project

That is true. That will also be the place where licensing will be verified.

quinncnl:
Thank you both. A quick question before I go there, will the API key be shared across all projects? I only want the key to be used in iD to start with.

bryceco:
It will be shared across projects, but there are only a few projects other than iD that use ELI. JOSM uses its own layer index. 

You can also add it without specifying an api key, and issue keys separately to client projects (but there's no guarantee they will special-case your layer to support the key).

tyrasd:
> I only want the key to be used in iD to start with.

As already pointed out above, this is not quite the preferred way by the community. But if you would like to continue that approach, you would add the layer using an `{apikey}` placeholder to ELI, and simultaneously add the layer also to iD in `data/manual_imagery.json`, where the api key can also be stored in an obfuscated way such that it is not picked up by search engines and/or "malicious" scrapers.

> as an imagery to help mappers

Again, thank you for this kind offer. For mapping, I'm seeing the layer's biggest benefit in the very high quality contour lines. That could be quite useful. On the other hand, there's little benefit in having rendered OSM data in the background map while mapping, and it could potentially be even distracting. Just out of curiosity: Would there be the possibility of you creating a dedicated rendering based on this map, but which excludes the OSM data? That would be more useful for mapping, specifically.

quinncnl:
> Again, thank you for this kind offer. For mapping, I'm seeing the layer's biggest benefit in the very high quality contour lines. That could be quite useful. On the other hand, there's little benefit in having rendered OSM data in the background map while mapping, and it could potentially be even distracting. Just out of curiosity: Would there be the possibility of you creating a dedicated rendering based on this map, but which excludes the OSM data? That would be more useful for mapping, specifically.

Indeed the most useful part for mappers is LiDAR based contour lines for Europe. We don't have dedicated contour lines at the moment. But that's definitely something worth looking into.

quinncnl:
For the record I opened a PR to ELI: https://github.com/osmlab/editor-layer-index/pull/2264 For JOSM it's already been added.

Issue #8415, opened 2021-03-19
Route relations are broken by iD


Yogurt4:
When someone edits ways that are part of a route relation (like cutting a way in two, or to delete a piece and draw a new way instead), iD most often corrupts the order of relation members. Even though the order would be very important for routes. Both OSMI, Osmose and JOSM display a "relation contains a gap" error for such relations.

The situation with iD is so bad that you may check the history of FlixBus routes, for example. A great deal of the versions is about "fix the gap" or "correct order" - after someone's edits with iD. Here we are in a continuous daily work to check and fix public transport routes. And it's an endless job as they get corrupted again and again.

a.) When having a route of A->B->C and the middle way is split in two, the proper order could be A->B'->New->C, or A->New->B'->C. So far I could not deduce a sure way of telling which one will happen.
- It might depend on the order of the nodes in the way: the first node stays in B', the last node gets into New.
- It might depend on the number of nodes: most nodes stay in B', the less nodes get into New.
- It might depend on distance: the longer part stays in B', the shorter one gets into New.
- It might be completely random, depending on node IDs or some internal map/tree/hash.

Algorithm-wise it is a simple task to decide the correct order so that chain remains intact.

b.) When adding a new way to a (route) relation, it's often added to the end of the members. Even if that way was added to mend a gap.
iD should walk along the existing relation members and check if the new way is connected to one of them. Of course, things get much more complicated in case of multiple connections (when having loops in the graph, like the two sides of a roundabout) but iD doesn't have to solve all these tricky situations perfectly.

c.) iD could also check the order of route member ways and give the user a warning.

1ec5:
> iD should walk along the existing relation members and check if the new way is connected to one of them. Of course, things get much more complicated in case of multiple connections (when having loops in the graph, like the two sides of a roundabout) but iD doesn't have to solve all these tricky situations perfectly.

Apparently iD does attempt to do this, as seen in [some tests](https://github.com/openstreetmap/iD/blob/6cfd09cb66c51c41c0334ebf30666c2a166b64b0/test/spec/actions/split.js#L445), but perhaps there are some cases that are being handled suboptimally.

matkoniecz:
Can you give example of changeset where breakage was caused by iD?

Yogurt4:
1. For example changeset [#101007031](https://www.openstreetmap.org/changeset/101007031).
If I see it right, the user split way 348179105 into two. The new way is 917172308.
[...]

Issue #8578, opened 2021-07-13
iD damages relations when splitting ways


robhubi:
### My System
Windows 10, 64 Bit
Firefox 89.0.2, 64 Bit
MS Edge 91.0.864.67, 64 Bit
iD 2.19.6

### Description
Splitting ways that are members of relations destroys the order of the members. See also [Order matters](https://wiki.openstreetmap.org/wiki/Relation:route#Order_matters).

### Steps to Reproduce
**Starting point**
The [Rabenfeldstraße ](https://www.openstreetmap.org/way/152695507) is, among other things, part of the Relation [Weingarten-Tour](https://www.openstreetmap.org/relation/109727). It has 58 members:
![iDbR_01](https://user-images.githubusercontent.com/38103976/125509061-48fbbac5-4f9a-48bd-9b71-6da9db01046a.jpg)

**Change**
The "Rabenfeldstraße" is split with the scissors:
![iDbR_02](https://user-images.githubusercontent.com/38103976/125509231-123cff3c-936c-44b6-9d35-50fd2854e12b.jpg)

### Actual Result
As expected, the relation Weingarten-Tour has 59 members afterwards, but the order is destroyed:
![iDbR_03](https://user-images.githubusercontent.com/38103976/125509446-e82e3f63-2823-4f87-ac3c-684da405d62c.jpg)

The effect is not always reproducible. On the "Panorama Tour (Eibiswald)" it no longer occurred the third time I wanted to take screenshots. With the "Weingarten Tour" I took the screenshots on the first repetition.
Of course, I did not save the changes.

### Expected Result
Unchanged order of members as done by JOSM.

### Other, possibly relevant issues
https://github.com/openstreetmap/iD/issues/4978
https://github.com/openstreetmap/iD/issues/4876
https://github.com/openstreetmap/iD/issues/4861

### PLEASE
Do not allow splitting of ways that are part of a relation as long as the issue is unresolved.

hungerburg:
Did you try with release version 2.20? Reading #8612 makes it look like iD started to validate and therefore fetch route relations in their entirety, which might be promising?

On the expected result: Of course order will be changed, there is a new member in town! Then, behaviour does not depend on how JOSM performs, the error is an error, regardless of that.

But most of all: How is order destroyed? Does only the new member end up in the wrong place? Or are members completely shuffled?

robhubi:
Version 2.20.1 shows the error in the same way.

[...]

Issue #6656, opened 2019-07-16
Download all members of relations in the background


quincylvania:
It's useful to have all members of a relation readily accessible, such as when zooming out to show the full extent of the relation. Rather than making the user manually download members (#5420), iD should automatically start downloading the full relation in the background once part of it is loaded for a tile. We might have to limit this by the number of members a relation has.

slhh:
@quincylvania 
Downloading the members of selected relations should have priority (part of #4903). In case automatic download of the selected relations needs to be limited, the user should be able to request the download manually (#5420).
Downloading any other relations can be limited, without generating such need for #5420.

BjornRasmussen:
I don't think this is a good idea, because the advantage is small, and it would lead to iD being slower for many people, especially those with older computers.

quincylvania:
> the advantage is small

iD could do things like validate the continuity of large multipolygons, boundaries, and routes, as well as completely render relations. The usefulness of this depends on each individual mapper's needs.

> it would lead to iD being slower for many people, especially those with older computers.

We wouldn't ship this if it makes iD slower. We'd have to figure out how to download and merge data without stalling iD on the main thread.

I'm thinking about this issue as more of a long-term goal than an immediate concern. It makes sense that a modern editing tool should have all relevant data available to the editor instead of them needing to worry about what's downloaded and what's not.

pnorman:
https://github.com/openstreetmap/iD/issues/1201#issuecomment-24951533 suggests 

> Issue a /relation/id#/full query for incomplete relations after all map? queries have finished. This is equivalent to doing JOSM's "Download Members" for each incomplete relation.

quincylvania:
> Issue a /relation/id#/full query for incomplete relations after all map? queries have finished.

@pnorman As you also mentioned in the comment, this call can run into issues with very large relations. For performance, iD really needs to exercise more granular control over how many members are getting fetched at once.

Re: https://github.com/openstreetmap/openstreetmap-website/issues/2348

pnorman:
> @pnorman As you also mentioned in the comment, this call can run into issues with very large relations. For performance, iD really needs to exercise more granular control over how many members are getting fetched at once.

Is it how many are getting fetched at once, or how many are added to iD's graph? The former would surprise me - is there much difference between requesting 500 ways 4 times or requesting 2000 ways once?

quincylvania:
> is there much difference between requesting 500 ways 4 times or requesting 2000 ways once?

[...]

Issue #5420, opened 2018-10-20
Add button for complete member download.


slhh:
The "all members" section has recently got buttons to download individual members. This is a nice improvement, but doesn't cover many usecases well, which need downloading all members.

Full member download buttons should be added to 

- the  "all members" section headline of incomplete relations.
- the items of the multiselection list, which are representing incompletely loaded relations.
This one might be the most important case. E.g. potential polygon operations like splitting or merging require multiple operands, and completely downloaded relations.

To prevent blocking the UI a long time, limit using the full relation API call to reasonable low member counts. For larger relations start downloading in small portions consecutively, and stop downloading when the user clicks elsewhere.

To prevent overloading the graph:
Short term solution: stop downloading when a fixed limit of loaded nodes is reached.
Long term solution: Do not load the elemnts into the graph, but a graph extension data structure.
The graph needs to get a second (sub)set of functions, which do also work with the elements of the graph extension. Some operations have to use these extended functions, but rendering  etc. can continue to use the normal graph.

mmd-osm:
Calling way/full for each way member can in fact be much slower than calling relation/full for the whole relation, though it is very difficult to estimate the expected runtime. I used https://www.openstreetmap.org/relation/148838 as an example, which has some 700+ way members:


#### Download full relation via relation/full

```
time curl -s --compressed https://api.openstreetmap.org/api/0.6/relation/148838/full > /dev/null
```

real	0m14,220s
user	0m0,295s
sys	0m0,069s


#### Download each member way via way/full

```
[...]

PR #10149, opened 2024-03-04
Robust local splitting of ways which are part of (route) relations

### LINKED ISSUE #7653: Incorrect reorder of relation members when not all members are loaded
Initially I thought that issue is appearing when route is made circular, after discussion on Slack and looking into code I was able to narrow it down to case when relation is not fully loaded and new member is added.

How to reproduce:
https://preview.ideditor.com/release/#background=Bing&disable_features=boundaries&map=16.00/43.8434/22.3387
Select path going from centre to north-east, add it to hiking route "Вршка чука - Влашки до - кружна стаза".  It will cause iD to add new member and reorder ways, even changing direction of route. 
Revert back, before adding path to relation download missing members of relation, add path to relation and everything goes ok.

Maybe useful, I'm attaching v5 and v6 preview of relation. Relation is now reverted to v5 ( v7 = v5 ).

[11164146.zip](https://github.com/openstreetmap/iD/files/4725109/11164146.zip)

Relation in question
https://www.openstreetmap.org/relation/11164146

tyrasd:
(should) fix #7653, #8415

This reworks the handling of relation-membership of ways during a split operation:

Instead of fully re-sorting the whole relation every time a member is split (which is inherently susceptible to faults when encountering edge cases, partially incomplete or partially broken data), this performs the split as a local operation:
* The new way is inserted into the relation before or after the existing member, depending only[^1] on how the old/new way connect to their neighboring members.
* Disable the split operation when the neighboring members have not been loaded (i.e. because they are too far out of the viewport)
* Disable the split operation on closed roundabout ways, where the split operation would potentially result in invalid routes

Adding this has the pleasant side effect that the complexity in `add_member` action could be reduced significantly, as it now does not have to keep track of `insertPairs` anymore.

The new algorithm works fine for:
* existing test cases
* situations where a route relation is loaded partially and the loaded members form a different, shorter route than the full set of relation members do
* cases with unsplit roundabouts (e.g. [relation/13092746](https://www.openstreetmap.org/relation/13092746))

[^1]: for cases where the split way and its neighbors form a loop which turns back to itself (i.e. situations like `way-1 -> way-to-split -> way-1`), an additional look-ahead is implemented to disambiguate the proper order


modules/actions/split.js
```diff
@@ -428,6 +428,9 @@ export function actionSplit(nodeIds, newWayIds) {
                             }
                         }
                     }
+                    if (way.tags.junction === 'roundabout' && way.isClosed()) {
```
1ec5: If the roundabout is a member of a `type=restriction`, `type=enforcement`, `type=junction` `junction=roundabout`, or `type=destination_sign` relation, but not a `type=route` relation, then it should be perfectly safe to split the roundabout. I’m pretty sure the code that handles restriction-like relation roles can handle closed ways just fine, but it would be a good case to test.

data/core.yaml
```diff
@@ -499,6 +499,8 @@ en:
           one: "Split a feature."
           other: "Split {n} features."
       not_eligible: Lines can't be split at their beginning or end.
+      parent_incomplete: This line cannot be split because it is part of a larger relation which has only been partially loaded. Make sure that all ways connected to this way are present on the map.
```
1ec5, line 502: I’m not sure if a user would understand what it means for the ways to be “present on the map”. This isn’t requiring the relation members to all be visible in the current viewport, is it?

How about adding a button to the top or bottom of the relation member list to load the full relation on demand? Currently, you can only do that by entering the relation ID into the main search bar, but it’s far from discoverable.

Since iD would be sorting the relation less aggressively, how about adding another button that does a full sort on demand, like JOSM has? This would be especially handy if the user has accidentally sorted the members by manually dragging them around. If some members haven’t downloaded yet, the button could be disabled.

1ec5 (commented):
I haven’t tested the local sorting approach yet, but hopefully that really cuts down on the issues people are seeing.

Some things that would benefit from a unit test:

* [Dual carriageways](https://wiki.openstreetmap.org/wiki/Route_directions#Cardinal_directions_as_roles) mapped as a single route, dividing and coming back together
* Forks in [recreational routes](https://wiki.openstreetmap.org/wiki/Roles_for_recreational_route_relations)
* A route that begins or ends in a closed way roundabout (probably no ambiguity there)

data/core.yaml
tyrasd: Yeah, this one was tricky to write a concise and understandable description for.

And you're right: the entities don't need to be actually _visible_ in the viewport. I thought about automatically downloading the missing prev/next members in such cases (it's really only up to two members, which is not a lot), but unfortunately the edit operations don't support async workflows at the moment (which would be required here to do the downloading).

> How about adding a button to the top or bottom of the relation member list to load the full relation on demand?

That sounds useful. But maybe in a different issue to not overload this one too much.

> full sort on demand, like JOSM has

I thought about that as well, but left it for another day/issue/pr. :wink: I think such a feature should come with some kind of (very) basic visualization of the order too.

modules/actions/split.js
tyrasd: Good point. `type=junction` and `type=enforcement` should work as is.

I have to test this again, but I thought that the turn restriction editor would not properly handle this right now… I mean… does it even make sense to use a closed loop as a from/to/via member of a turn restriction or destination sign? :thinking:

data/core.yaml
1ec5: iD already has a good enough visualization for this basic purpose (hovering over members when zoomed to fit). I just wouldn’t want the outcome of this genuine improvement to be that people still tell new mappers to only map route relations in JOSM because it doesn’t clean up after new or existing scrambling.

I think #5420 is more tractable than #6656 because we won’t wastefully spam the API for unwanted data, and because the code basically already exists. Some other messages warn about a way not being fully visible, but at least there’s a button and keyboard shortcut for that.

If you’d like to keep the scope to a minimum, that’s fine, but then I have no idea how to make this message genuinely helpful. It might turn out to be easier to implement one of the two buttons. Is the new implementation still capable of a full relation sort, or would that functionality have to be reimplemented?

modules/actions/split.js
```diff
@@ -313,12 +402,37 @@ export function actionSplit(nodeIds, newWayIds) {
...
+                            return 'parent_incomplete';
+                        }
+                    } else {
+                        // other relations (e.g. route relations): at least one members before or after way must be present
+                        for (let i = 0; i < parentRelation.members.length; i++) {
+                            if (parentRelation.members[i].id === way.id) {
+                                const memberBeforePresent = i > 0 && graph.hasEntity(parentRelation.members[i - 1].id);
+                                const memberAfterPresent = i < parentRelation.members.length - 1 && graph.hasEntity(parentRelation.members[i + 1].id);
+                                if (!memberBeforePresent && !memberAfterPresent) {
+                                    return 'parent_incomplete';
```
tyrasd, line 475: not to self: maybe add `…&& !parentRelation.members.length > 1` (otherwise it will fail for newly created routes with only 1 member to start with).

modules/actions/split.js
1ec5: > does it even make sense to use a closed loop as a from/to/via member of a turn restriction or destination sign?

I don’t know how common it is, but picture a turning loop mapped as a closed way. Some turning loops allow you to go all the way around, while others do not. If not, the loop could be a member of a `no_u_turn` relation as both the `from` and `to`.

I’ve encountered this situation plenty of times at schools, but these tend to be parking lot roads that would at most be the beginning or end of a school bus route relation. Perhaps it would be more likely at a bus station?

A roundabout could be a likely `via` member, come to think of it. For example, if there’s a slip lane for going through, there might be a restriction prohibiting the user from using the roundabout itself to go through. I would always split a roundabout in such a scenario, but some mappers might prefer to keep the roundabout as a single closed way, for the same reason they’d do so in a route relation.

1ec5:
Here’s a pathological case that I had been meaning to use as a test in my own experiments toward a solution: https://www.openstreetmap.org/relation/15504667 traverses https://www.openstreetmap.org/relation/15508044 more than once, going over a different portion each time. It’s also a one-way relation, so the order definitely matters. It’s a very long relation,[^unmapped] so it would be impractical to expect the user to load its members one by one. It backtracks many times, sometimes doubling back immediately on a given roadway.

Would a split in this roundabout or one of the backtracks preserve the order? What if the roundabout had originally been mapped as a single closed way and added to the relation twice?

[^unmapped]: Actually, only a third of the actual length has been mapped so far.

modules/actions/split.js
tyrasd: I'm not sure I can quite follow your turning loop example. Do you have an example at hand how this looks like?

The roundabout-as-via situation is what I had in mind of not being properly handled currently: it's more or less the same issue as with route relations: the splitting would need to be more clever than what iD currently does [^1], as after the split the from -> via [--> via]* --> to path should "remain" contiguous. As far as I know it was never documented that closed roundabouts are allowed to be used in that way for turn restrictions, but examples [do exist (rarely)](https://overpass-turbo.eu/s/1Icb) (e.g., a closed roundabout as `to`: [relation/2996734](https://www.openstreetmap.org/relation/2996734), roundabouts as `via`: [relation/5378197](https://www.openstreetmap.org/relation/5378197), [relation/13437147](https://www.openstreetmap.org/relation/13437147)). FWIW, not even JOSM handles splitting those gray edge cases correctly :shrug: , but for iD I'd rather gravitate towards the more conservative approach of disallowing potentially data-breaking operations.

[^1]: Closed ways are split at the selected split point and the first/last node of the closed way.

data/core.yaml
tyrasd: d7111e2 adds a button to do the `/full` download of a particular relation.

modules/actions/split.js
1ec5: > I'm not sure I can quite follow your turning loop example. Do you have an example at hand how this looks like?

I’m not sure how to search for this, even though I encounter it from time to time. It would look something like this, but obviously in a more built-up environment. There would be a traffic sign indicating a No U Turn or No Left Turn restriction as you finish looping around:

Turning circle

Currently, creating a `restriction=no_u_turn` relation at the highlighted point would automatically split the closed way in two, which is fine. But if the roadway happens to be part of a bus route relation (not necessarily at the beginning or end of it), I don’t know if that would still happen automatically or if it would conflict with what you’ve written.

That said, I just noticed that it’s an and, not an or, so no turning loop would be affected.

I agree with taking a conservative approach. It’s just that iD unavoidably has a preference against closed roundabouts due to the ambiguity, so I’m a bit reticent about the possibility that it would become more difficult to resolve that ambiguity by splitting the way. Thinking through this is helping me feel better about it though.

modules/actions/split.js
```diff
@@ -313,12 +403,38 @@ export function actionSplit(nodeIds, newWayIds) {
...
+                                const memberBeforePresent = i > 0 && graph.hasEntity(parentRelation.members[i - 1].id);
+                                const memberAfterPresent = i < parentRelation.members.length - 1 && graph.hasEntity(parentRelation.members[i + 1].id);
+                                if (!memberBeforePresent && !memberAfterPresent && parentRelation.members.length > 1) {
+                                    return 'parent_incomplete';
+                                }
+                            }
+                        }
+                    }
+                    const relTypesExceptions = ['junction', 'enforcement']; // some relation types should not prehibit a member from being split
+                    if (way.tags.junction === 'roundabout' && way.isClosed() && !relTypesExceptions.includes(parentRelation.tags.type)) {
```
1ec5: `junction=circular` would be equivalent to `junction=roundabout` for the purpose of this check. That said, the scrambling misbehavior was never sensitive to whether the loop was tagged as a roundabout or circular intersection.

modules/ui/sections/raw_membership_editor.js
```diff
@@ -376,6 +388,13 @@ export function uiSectionRawMembershipEditor(context) {
             .style('border-color', d => d.relation.tags.colour)
             .text(function(d) { return utilDisplayName(d.relation); });
 
+        labelEnter
+            .append('button')
+            .attr('class', 'members-download')
+            .attr('title', t('icons.download'))
+            .call(svgIcon('#iD-icon-load'))
+            .on('click', downloadMembers);
```
1ec5, line 396: Thanks for adding this button! This placement works, but I wonder if it would be less confusing to put it in the raw member editor, at the bottom next to the ➕ button. After all, that’ll be right where the user is seeing that some of the members haven’t downloaded yet. (Not a blocker for this PR.)

modules/actions/split.js
```diff
@@ -253,6 +231,118 @@ export function actionSplit(nodeIds, newWayIds) {
         return graph;
     }
 
+    function splitWayMember(graph, relationId, wayA, wayB) {
```
1ec5, line 252: This function could use some additional commenting.

modules/ui/sections/raw_membership_editor.js
tyrasd: My idea behind this was that for the use cases of this issues, going to the relation to do the full members download would be an extra click, so having the button in the membership of the way to be split would be more efficient UX. But we can of course also thing about adding a similar button to the relation entity view to make them idempotent. The download members feature would especially useful when the "sort relation members" functionality is going to be implemented in that component.

tyrasd:
Thanks for the feedback on this @1ec5 ! :bow: 

I forgot to ask whether you have a suggestion to improve the "cannot split because of missing members" message now that we have the download button for the relations. Perhaps something like the following?

> This line cannot be split because it is part of a larger relation which has only been partially loaded. Make sure that the way's relations are fully downloaded.

1ec5:
Yes, something along those lines:

> This line cannot be split because a parent relation isn’t fully downloaded. Download the full relation.

Hopefully the user would be able to figure out where to download the members. Would it be too verbose to read before the message goes away?

danieldegroot2:
Fixes also #8578

PR #10154, opened 2024-03-11
fixed #10153 Moving source in front of external source inherits exter…

### LINKED ISSUE #10153: Moving source in front of external source inherits external url
### URL

Any

1. Add article url
2. Add source
3. Move source in front of article url
4. Source inherits same article url.
5. Moving it all the way to the end transfers the new link to the first source after the original link.
6. Etc., seems to be a visual bug.

image

2.28.1

Chrome

critophice:
fixed #10153


tordans:
Can you provide a screencast of the issue and how this change fixes it. I don't think it will...

tyrasd:
yeah, something is missing here, as only the documentation page was changed (and that was not really necessary to fix the bug).

PR #10165, opened 2024-03-17
Field: fix wikiURL by URL encoding it - and slight refactors

### LINKED ISSUE #10065: Wikipedia link with question mark
### URL

https://www.openstreetmap.org/edit?way=394705975#map=19/44.81767/20.45303

1. Clear the wikipedia tag.
2. Select language "српски / srpski".
3. Start typing "Кафана „?”". Select such value from the drop-down.
4. Click on the button "view on wikipedia.org".
5. You will get the error page in the browser.

Tag "wikipedia=sr:Кафана_„%3F”" works, tag "wikipedia=sr:Кафана „?”" doesn't.
If you save tag "wikipedia=sr:Кафана „?”" problem will stay in the view (not the edit) mode too.

image

2.27.3

Chrome

hirako2000:
### Description 

closes #10065 by 

-  URL encoding wikiURL tag on change

#### Refacrtor

- Define wikipedia domain URL and scheme for reuse
- Extract toUrlAnchor as its own function for reuse 

### How was this tested?

- Executed the full unit test suite
- Added a unit tests to cover my changes
- ~Would do with complete assertion on the unit test to check the set wikiURL but was unable to expose it to test suite~
- Verified by running the application locally and auto filling various wikipedia URLs containing characters requiring URL encoding, such as `https://en.wikipedia.org/wiki/%3F_(film)` _and followed by some section header as anchor, directly_.

### DIFF
--- a/modules/ui/fields/wikipedia.js
+++ b/modules/ui/fields/wikipedia.js
@@ -308,9 +308,9 @@ export function uiFieldWikipedia(field, context) {
   }
 
   function updateEncodedWikiUrl(tagLang, tagArticleTitle, anchor) {
-    const underscored = tagArticleTitle.replace(/ /g, '_');
+    const underscoredUrlEncodedTitle = encodeURIComponent(tagArticleTitle.replace(/ /g, '_'));
     const urlAnchored = toUrlAnchor(anchor);
-    _wikiURL = encodeURI(`${scheme}${tagLang}.${domain}/wiki/${underscored}${urlAnchored}`);
+    _wikiURL = `${scheme}${tagLang}.${domain}/wiki/${underscoredUrlEncodedTitle}${urlAnchored}`;
   }
 
   function toUrlAnchor(anchor) {
--- a/modules/ui/fields/wikipedia.js
+++ b/modules/ui/fields/wikipedia.js
@@ -11,6 +11,8 @@ import { utilGetSetValue, utilNoAuto, utilRebind } from '../../util';
 
 
 export function uiFieldWikipedia(field, context) {
+  const scheme = 'https://';
+  const domain = 'wikipedia.org';
   const dispatch = d3_dispatch('change');
   const wikipedia = services.wikipedia;
   const wikidata = services.wikidata;
@@ -132,7 +134,7 @@ export function uiFieldWikipedia(field, context) {
     link = link.enter()
       .append('button')
       .attr('class', 'form-field-button wiki-link')
-      .attr('title', t('icons.view_on', { domain: 'wikipedia.org' }))
+      .attr('title', t('icons.view_on', { domain }))
       .call(svgIcon('#iD-icon-out-link'))
       .merge(link);
 
@@ -279,7 +281,7 @@ export function uiFieldWikipedia(field, context) {
     if (tagLangInfo) {
       const nativeLangName = tagLangInfo[1];
       utilGetSetValue(_langInput, nativeLangName);
-      utilGetSetValue(_titleInput, tagArticleTitle + (anchor ? ('#' + anchor) : ''));
+      utilGetSetValue(_titleInput, tagArticleTitle + toUrlAnchor(anchor));
       if (anchor) {
         try {
           // Best-effort `anchorencode:` implementation
@@ -288,8 +290,7 @@ export function uiFieldWikipedia(field, context) {
           anchor = anchor.replace(/ /g, '_');
         }
       }
-      _wikiURL = 'https://' + tagLang + '.wikipedia.org/wiki/' +
-        tagArticleTitle.replace(/ /g, '_') + (anchor ? ('#' + anchor) : '');
+      updateEncodedWikiUrl(tagLang, tagArticleTitle, anchor);
 
     // unrecognized value format
     } else {
@@ -297,7 +298,7 @@ export function uiFieldWikipedia(field, context) {
       if (value && value !== '') {
         utilGetSetValue(_langInput, '');
         const defaultLangInfo = defaultLanguageInfo();
-        _wikiURL = `https://${defaultLangInfo[2]}.wikipedia.org/w/index.php?fulltext=1&search=${value}`;
+        _wikiURL = `${scheme}${defaultLangInfo[2]}.${domain}/w/index.php?fulltext=1&search=${value}`;
       } else {
         const shownOrDefaultLangInfo = language(true /* skipEnglishFallback */);
         utilGetSetValue(_langInput, shownOrDefaultLangInfo[1]);
@@ -306,6 +307,15 @@ export function uiFieldWikipedia(field, context) {
     }
   }
 
+  function updateEncodedWikiUrl(tagLang, tagArticleTitle, anchor) {
+    const underscored = tagArticleTitle.replace(/ /g, '_');
+    const urlAnchored = toUrlAnchor(anchor);
+    _wikiURL = encodeURI(`${scheme}${tagLang}.${domain}/wiki/${underscored}${urlAnchored}`);
+  }
+
+  function toUrlAnchor(anchor) {
+    return anchor ? '#' + anchor : '';
+  }
 
   wiki.entityIDs = (val) => {
     if (!arguments.length) return _entityIDs;
--- a/modules/ui/fields/wikipedia.js
+++ b/modules/ui/fields/wikipedia.js
@@ -299,22 +299,15 @@ export function uiFieldWikipedia(field, context) {
 
   wiki.encodePath = (tagArticleTitle, anchor) => {
     const underscoredTitle = tagArticleTitle.replace(/ /g, '_');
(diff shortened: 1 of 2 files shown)


modules/ui/fields/wikipedia.js
```diff
@@ -306,6 +307,15 @@ export function uiFieldWikipedia(field, context) {
     }
   }
 
+  function updateEncodedWikiUrl(tagLang, tagArticleTitle, anchor) {
+    const underscoredUrlEncodedTitle = encodeURIComponent(tagArticleTitle.replace(/ /g, '_'));
+    const urlAnchored = toUrlAnchor(anchor);
+    _wikiURL = `${scheme}${tagLang}.${domain}/wiki/${underscoredUrlEncodedTitle}${urlAnchored}`;
```
1ec5: Perhaps this would be more testable if it returns the URL instead of setting `_wikiURL`.

modules/ui/fields/wikipedia.js
```diff
@@ -11,6 +11,8 @@ import { utilGetSetValue, utilNoAuto, utilRebind } from '../../util';
 
 
 export function uiFieldWikipedia(field, context) {
+  const scheme = 'https://';
+  const domain = 'wikipedia.org';
```
1ec5, line 15: This is reasonable, though FYI the _languageCode_.wikipedia.org subdomain pattern, `w/index.php` path, and `wiki/` rewrite rule are all specific to Wikipedia and [configurable by each wiki](https://www.mediawiki.org/wiki/Manual:Short_URL).

modules/ui/fields/wikipedia.js
```diff
@@ -306,6 +307,15 @@ export function uiFieldWikipedia(field, context) {
...
 
+  function updateEncodedWikiUrl(tagLang, tagArticleTitle, anchor) {
+    const underscoredUrlEncodedTitle = encodeURIComponent(tagArticleTitle.replace(/ /g, '_'));
+    const urlAnchored = toUrlAnchor(anchor);
+    _wikiURL = `${scheme}${tagLang}.${domain}/wiki/${underscoredUrlEncodedTitle}${urlAnchored}`;
+  }
+
+  function toUrlAnchor(anchor) {
+    return anchor ? '#' + anchor : '';
+  }
```
1ec5: This ternary is simple enough to inline. While we’re at it, the anchor should also be encoded. You can apply the same encoding to both the page title and the named anchor, replacing spaces with underscores and calling `encodeURIComponent()`. (There are some [nuanced edge cases](https://www.mediawiki.org/wiki/Manual:PAGENAMEE_encoding) that you probably don’t need to spend time worrying about.)

modules/ui/fields/wikipedia.js
hirako2000: Good idea!

modules/ui/fields/wikipedia.js
hirako2000: the lang as subdomain is already being taken care by pre-existing logic. I believe.

Good catch about the specificities. Is this code meant to deal with non wikipedia rewrite rules? I would happily make it wiki agnostic, but it doesn't seem trivial to retrieve the used rewrite rules. Not present in result search metadata, no idea from where to determine it.

modules/ui/fields/wikipedia.js
hirako2000: sure thing, will refactor. 

reading through the nuanced edge cases, xin cảm ơn.

modules/ui/fields/wikipedia.js
hirako2000: Made the function return, and assign to wiki so that the tests can invoke it.

modules/ui/fields/wikipedia.js
hirako2000: I came across the fact URI encoding the whole path (that is including the hash/ref of the anchor) causes an unnecessary redirect. e.g [https://en.wikipedia.org/wiki/%3F_(film)%23Themes_and_style](https://en.wikipedia.org/wiki/%3F_(film)%23Themes_and_style)

I have refactored the function to keep inline use, and dealt with encoding each part, the title and anchor fragment separately.

Note: I'm unable to manually test via the UI assignment of an anchored wiki URI.

modules/ui/fields/wikipedia.js
1ec5: You’re right, this feature is 100% specific to Wikipedia. I just pointed it out since the code is becoming a bit more generic in this PR, but I wouldn’t expect any change as a result.

modules/ui/fields/wikipedia.js
```diff
@@ -307,15 +297,25 @@ export function uiFieldWikipedia(field, context) {
...
-  }
+  wiki.anchorFragment = (anchor) => {
+    if (!anchor) {
+      return '';
+    }
+
+    try {
+      // Best-effort `anchorencode:` implementation
+      anchor = anchor.replace(/ /g, '_').replace(/%/g, '.');
+    } catch (e) {
```
1ec5: Good idea to emulate `anchorencode:`. If it fails for some reason, would it result in an exception that you can catch here, or can we omit the try-catch safely?

modules/ui/fields/wikipedia.js
```diff
@@ -307,15 +297,25 @@ export function uiFieldWikipedia(field, context) {
...
   }
 
-  function updateEncodedWikiUrl(tagLang, tagArticleTitle, anchor) {
-    const underscoredUrlEncodedTitle = encodeURIComponent(tagArticleTitle.replace(/ /g, '_'));
-    const urlAnchored = toUrlAnchor(anchor);
-    _wikiURL = `${scheme}${tagLang}.${domain}/wiki/${underscoredUrlEncodedTitle}${urlAnchored}`;
-  }
+  wiki.encodePath = (tagArticleTitle, anchor) => {
+    const underscoredTitle = tagArticleTitle.replace(/ /g, '_');
+    return `${encodeURIComponent(underscoredTitle)}${anchor ? '#' + encodeURIComponent(wiki.anchorFragment(anchor)): ''}`;
```
1ec5: Nit: This line is getting a little unwieldy. Inlining `toUrlAnchor()` was good, but we don’t have to put everything on a single line. It’s OK to declare a `let` variable, for instance, and append to it when necessary.

1ec5 (commented):
Thanks for iterating – this is getting close.

modules/ui/fields/wikipedia.js
hirako2000: agreed, and fixed. 

I didn't make use of a `let` declaration though, for clear variable naming.

modules/ui/fields/wikipedia.js
hirako2000: I refactored it and yes I believe it could go away, so it is now gone. Please note that _try-catch_ `anchorencode` was in there before this fix. It got extracted into a separate function without  affecting the flow.

I dug further into it, and couldn't find any reason* for replacing `%` with `.`

I concluded that it was a Best-effort at trying to deal with URI encoded looking characters that happen to be contained in sections then made anchors, without resorting to `encodeURIComponent()`, here also for reasons I could not explain. Also, could not think of any possible string content that would make `replace()` throw an error, it may throw error on `null`, `undefined` value, and invalid regex though. But then it would throw an error within the catch block.

Anyhow I have added further unit test cases, which hopefully clarifies intents of those functions now. If the test cases aren't correct, happy to iterate again.


*or was it perhaps an attempt to defend against [Double Encoding Attacks](https://owasp.org/www-community/Double_Encoding) ? If that could be the case perhaps it should be put back where it was, which was the rational of keeping it in there initially, but in any case, it does not seem like a solid way to prevent URL tempering  attacks so I settled with eliminating it.

hirako2000:
> getting close.

Getting close, I don't know. But closer, **yes**. Thanks for your valuable review comments.

modules/ui/fields/wikipedia.js
hirako2000: I was finally able to manually test with anchored URL, figured out how.

modules/ui/fields/wikipedia.js
1ec5: It looks like my recollection of MediaWiki’s behavior may be somewhat outdated. If a section is titled something like “[#time](https://www.mediawiki.org/wiki/Help:Extension:ParserFunctions##time)”, there will be two named anchors, `#time` and `.23time`. Some other reserved characters behave similarly. It used to be the case that the page would contain only `.23time`, but now that it contains `#time` as well, there’s no need for anything special, other than replacing spaces with underscores.

1ec5 (approved):

tyrasd:
Right! This should be fixed in iD with https://github.com/openstreetmap/iD/commit/ee56b9c8ae1ec1e5f7544b9644409b82e4e63e3a

PR #10175, opened 2024-03-22
Fix roadheight field width

### LINKED ISSUE #10174: Height Limit field horizontally overflows sidebar
### URL

_No response_

Add the Height Limit field to the inspector.

The two textboxes each take up so much room that the rightmost textbox pushes off the right edge of the containing sidebar.

This issue has also affected the OpenHistoricalMap fork’s date field type, which is styled just like the Height Limit field. It appears to be limited to Firefox 123 and above on macOS and Firefox 124 and above on Windows: https://github.com/OpenHistoricalMap/issues/issues/714#issuecomment-2013888949.

_No response_

Firefox

1ec5:
Recent versions of Firefox give input fields a greater intrinsic width that causes `roadheight` fields to overflow the container. Constrain these input fields to an initial width of zero; they’ll grow proportionally as part of the flexbox.

The four text fields that make up the Road Height field no longer overflow horizontally.

Fixes #10174.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Upgrade closed ways with `traffic_calming=island` to `area:highway=traffic_calming` ([id-tagging-schema#1162])
 #### :bug: Bugfixes
 * Prevent (route) relations from getting corrupted while splitting their way members in certain conditions ([#7653], [#8415])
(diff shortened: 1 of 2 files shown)


1ec5:
I tested this fix in Chrome 123, Firefox 125.0b3, and Safari 17.3.1 on macOS.

tyrasd (approved):
I noticed that setting `width: inherit` also fixes the problem, but I guess keeping the width in sync with the `flex-basis` is slightly better as it is a more explicit fix/workaround…

PR #10181, opened 2024-03-29
Sort "more fields" dropdown items

zbycz:
The more fields combo dropdown offers 1) `moreFields` from the current preset and 2) all fields with `universal:true`. Currently they are all mixed together.

This PR sorts the dropdown so the moreFields are displayed first, and the universal below.

The idea is, that user would like to know which fields are relevant to the specific preset first, and not to look for them among all the universal ones.

###  Screenshot 

eg. for **train track** (w166662057)

| old | this PR |
| -- | -- |
| image | image |

### DIFF
--- a/modules/ui/form_fields.js
+++ b/modules/ui/form_fields.js
@@ -16,7 +16,8 @@ export function uiFormFields(context) {
     function formFields(selection) {
         var allowedFields = _fieldsArr.filter(function(field) { return field.isAllowed(); });
         var shown = allowedFields.filter(function(field) { return field.isShown(); });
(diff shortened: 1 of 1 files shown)


1ec5:
Not a bad idea. 👍 Since the user may not intuit the difference between a universal field and a non-universal one, maybe there should be a horizontal divider or some other visual distinction.

zbycz:
> maybe there should be a horizontal divider or some other visual distinction.

Could be - i wanted to submit the simplest mergable solution. Lets see if it will be approved 🙂

tyrasd (approved):
looks good to me, thanks

PR #10189, opened 2024-04-05
do not remove spaces in inscription after ; sign

### LINKED ISSUE #10188: Don't remove spaces after the semicolon in the [inscription] field
Currently when I type
```
description=sth; sth; sth
inscription=sth; sth; sth
note=sth; sth; sth

cuisine=sth; sth; sth
source=sth; sth; sth
sport=sth;   sth;   sth
```
iD changes it to

```
description=sth; sth; sth
inscription=sth;sth;sth
note=sth; sth; sth

cuisine=sth;sth;sth
source=sth;sth;sth
sport=sth;sth;sth
```

It is correct to remove spaces for `cuisine`, `source` and `sport` (there should be no spaces).
It is correct to not remove spaces for `description` and `note`.
But for `inscription` there should be the same behavior as for `description` and `note` becuase it is the similar type of key.

matkoniecz:
fixes #10188

note: this is untested, at least for now (I will see is CI deployment will appear)

### DIFF
--- a/modules/util/clean_tags.js
+++ b/modules/util/clean_tags.js
@@ -17,7 +17,7 @@ export function utilCleanTags(tags) {
         }
 
         function skip(k) {
-            return /^(description|note|fixme)$/.test(k);
+            return /^(description|note|fixme|inscription)$/.test(k);
         }
 
         if (skip(k)) return v;


tyrasd (approved):
Thanks, this looks good to me.

Are there more tags that could be included here? Maybe a `name` in some special situations could have a semicolon?

1ec5:
> Are there more tags that could be included here? Maybe a `name` in some special situations could have a semicolon?

A literal semicolon needs to be [escaped by another semicolon](https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator#Escaping_with_';;'). A plain text field can’t do anything smart about that, because it displays value-delimiting semicolons verbatim. A multicombo field, such as the one for `destination=*`, could interpret a user-entered <kbd>;</kbd> as `;;`, but it hasn’t come up because such cases are so rare in practice. If we ever enhance the `name=*` field with multivalue functionality, we could consider supporting `;;` for [this one café in New Zealand](https://www.openstreetmap.org/node/9725500420). 🙃

PR #10190, opened 2024-04-05
Doc/API: Add default `en-GB` to examples

tordans:
I was looking at a local build (`npm run start`) and wanted to switch to the EN version to better compare to what others see. It looks like, `en-GB` does use the translations that is the default for iD, is that correct?
`en-US` did show missing translations locally, so that cannot be right, so does just `en`.

### DIFF
--- a/API.md
+++ b/API.md
@@ -32,7 +32,7 @@ of iD (e.g. `https://ideditor-release.netlify.app`), the following parameters ar
 * __`id`__ - The character 'n', 'w', or 'r', followed by the OSM ID of a node, way or relation, respectively. Selects the specified entity, and, unless a `map` parameter is also provided, centers the map on it.<br/>
   _Example:_ `id=n1207480649`
 * __`locale`__ - A code specifying the localization to use, affecting the language, layout, and keyboard shortcuts. Multiple codes may be specified in order of preference. The first valid code will be the locale, while the rest will be used as fallbacks if certain text hasn't been translated. The default locale preferences are set by the browser.<br/>
-  _Example:_ `locale=ja`, `locale=pt-BR`, `locale=nl,fr,de`<br/>
+  _Example:_ `locale=en-GB` (default), `locale=ja`, `locale=pt-BR`, `locale=nl,fr,de`<br/>
   _Available values:_ Any of the [supported locales](https://github.com/openstreetmap/iD/tree/develop/dist/locales).
 * __`map`__ - A slash-separated `zoom/latitude/longitude`.<br/>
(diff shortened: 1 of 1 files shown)


tordans:
Mh. It looks like in Dev mode `en-GB` is not the right translation. @tyrasd do you know from the top of your head what the default locale string would be? The code references `en` and `en-US` but both don't work. Do I need to do something else than `npm i && npm run start`?


[Testlink](http://127.0.0.1:8080/#background=Bing&disable_features=boundaries&id=w150634670&locale=en-US&map=18.89/52.52067/13.39044)

1ec5:
`en` implicitly means American English, so there’s no `en-US` locale. When a browser requests `en-US`, it gets `en`. See the lengthy discussion in openstreetmap/id-tagging-schema#357.

tyrasd:
Yeah… the default locale (if a browser does not specify any, which is unlikely afaik) is `en-US`, which is internally "shortened" to `en` (for iD, `en` is considered a synonym of `en-US`).

> Do I need to do something else than npm i && npm run start?

You are likely missing the file `dist/locales/en.min.json`, which is omitted in the development branch on purpose (see `.gitignore`, was removed with #8776). This file is generated from the strings in `core.yaml` by [running](https://github.com/openstreetmap/iD/wiki/How-to-get-started#build-and-test-instructions for more) `npm run all`[^2].

It's arguably a bit confusing that all the other locales (e.g. `en-GB`) are included in the development branch. Perhaps even unnecessary (except maybe when working on the localization code)… :thinking: 

[^2]: If you're on a fresh clone and do not run this, you'll also be missing most icons for example.

tordans:
Thanks @tyrasd, I skipped the "run all" part in deed. TBH, I would expect that to be part of "npm run dev" which could run "npm run all && npm run start". Because when would I _not_ want to run all?

> This file is generated from the strings in `core.yaml` by [running](https://github.com/openstreetmap/iD/wiki/How-to-get-started#build-and-test-instructions for more) `npm run all`[1](#user-content-fn-2-92cdc663d6b02f25c7f736ad0cef40f2).

I suggest removing this section in a similar way to https://github.com/openstreetmap/iD/pull/10203 (AKA have a headline but link to some other parts of the docs). I find it super confusing when Information is scattered in different areas of the project. The wiki in general is something that is confusing for me, because we already a markdown-docs-"wiki" inside the main repo. Whenever I find some info in the readme or contribution MD, I never look anywhere else but expect that to have all the info (which it does not ATM in the contribution MD, which is https://github.com/openstreetmap/iD/pull/10203).

PR #10203, opened 2024-04-15
Docs/CONTRIBUTING: Reference readme on how to build

tordans:
The readme has more information about how to build and run iD than this doc. Instead of having some info duplicated, lets reference it so we only need to keep one up to date.


tyrasd (approved):
FYI: I've consolidated this all into the wiki page: https://github.com/openstreetmap/iD/wiki/How-to-get-started#build-and-test-instructions

I think the main project readme was not really the best place for this information, and the wiki can also be more easily be updated.

PR #10205, opened 2024-04-16
fix double click bug in note comment

### LINKED ISSUE #8994: Notes: Resolve with comment requires double click
### URL

edit*

1. Open a note.
2. Write a comment.
3. Click "resolve with comment". Nothing happens.
4. Click it again.
(Optionally, exit the note between step 3 and 4, this action is still saved)

This does work in one click on the `https://mapwith.ai/rapid` version.

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.com

_No response_

laigyu:
fix: #8994 

I tested it, it should fixed , plz test

``` <button><span></span></button> ```
the bug is when you click span part it bug but if click button part it work, 
see: https://css-tricks.com/slightly-careful-sub-elements-clickable-things/ 

There are also has same``` <button><span></span></button> ```structure in context ,
maybe consider fix it

### DIFF
--- a/modules/ui/note_editor.js
+++ b/modules/ui/note_editor.js
@@ -371,7 +371,9 @@ export function uiNoteEditor(context) {
                 var andComment = (d.newComment ? '_comment' : '');
                 return t.html('note.' + action + andComment);
             })
-            .on('click.status', clickStatus);
+            .on('click.status', clickStatus)
+            .select('span')       // fix double click bug - #8994
+            .attr('style', 'pointer-events: none;');
 
         buttonSection.select('.comment-button')   // select and propagate data
             .attr('disabled', isSaveDisabled)
--- a/modules/ui/note_editor.js
+++ b/modules/ui/note_editor.js
@@ -370,10 +370,9 @@ export function uiNoteEditor(context) {
                 var action = (d.status === 'open' ? 'close' : 'open');
                 var andComment = (d.newComment ? '_comment' : '');
                 return t.html('note.' + action + andComment);
-            })
-            .on('click.status', clickStatus)
-            .select('span')       // fix double click bug - #8994
-            .attr('style', 'pointer-events: none;');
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by tyrasd — APPROVED** [review]
Thank your for pin-pointing the issue down to this.

I found that there's a slightly more elegant way to apply the event handler properly to the button including its child elements (e.g. the localized text `<span>` here). See my added commit.

**Comment by laigyu** [comment]
I believe the left button still has double click bug
it happen when you input the text, and the ```<span>``` localized text is changed to 'Reopen and Comment' or 'Close and Comment'

plz add console.log in clickStatus 
```
function clickStatus(d3_event, d) {
        console.log(d3_event)
        ... ohter code
    }
```
when ```target: span.localized-text``` it will need double click ,
but if is ```target: button.button.status-button.action``` is only need one. 

env :2.29.0-dev
chrome : Version 124.0.6367.60 (Official Build) (64-bit)   
firefox : 125.0.1 (64-bit)

could someone plz confirm

tyrasd:
odd :thinking:  It seems to work fine for me (tested only on Chrome for now). Do you have the issue also when using https://ideditor.netlify.app/?

laigyu:
@tyrasd yes  on chrome Version 124.0.6367.79 (Official Build) (64-bit)  https://ideditor.netlify.app/  still has the issue

tyrasd:
Sorry, you're right. ~~I had only tested it in this branch, but it seems like my follow-up change in fc4e05601a30afaf0ed0dbbbefac26f53574d0f1 immediately broke the fix again.~~ //edit: now I cannot reproduce the original fix anymore. :see_no_evil:

tyrasd:
5251ee6b65937382da963f6969e2435064d83ce4 should finally fix it properly. :bow:

laigyu:
its good :+1:

PR #10207, opened 2024-04-16
Add background blur to semitransparent elements

linfindel:
This adds a small 2px backdrop blur to the top and bottom bars, and the map controls. It improves readability and personally I think it just looks more stylish.


angiekins:
Niceeeee

tyrasd:
Hey! @linfindel Would you mind sharing a few example screenshots about how this looks before/after this change? And what your motivation was to introduce this. Thank you!

PR #10210, opened 2024-04-17
Add border radius to minimap

linfindel:
This makes the minimap fit in more with the rest of iD by having a small border radius


tyrasd:
sure, why not. :wink:

PR #10228, opened 2024-05-02
Move `demolished:building` to Future/Past and Buildings layers

### LINKED ISSUE #7582: Change category for Recently Demolished Building
I think `demolished:building=yes` preset should be in the category Future/Past or Buildings, not Landuse.

I think Buildings will be better because it will prevent from adding a building again from aerial imagery.

Dimitar5555:
Resolves https://github.com/openstreetmap/iD/issues/7582

Note: I haven't tested it but looking at the other code, it should work.


tyrasd:
> it should work

Unfortunately, it didn't quite. :stuck_out_tongue_winking_eye: 

In addition to what @bhousel mentioned in https://github.com/facebook/Rapid/issues/1403, the past/futures filter was also not quite doing it's job properly (it was only matching features with e.g. `demolished=*` or `<tag>=demolished`, but not features with `demolished:*=*`).

I've also had to change the _Landuse_ feature class to be slightly more specific. The previous catch-all implementation was not very clever to be honest, especially giving that there exists also an _Other_ class for exactly that. Hopefully this isn't breaking some people's workflow. :crossed_fingers:

maro-21:
Thanks!

PR #10254, opened 2024-05-25
support 3rd-party icons for preset fields

k-yle:
iD already supports third-party icons for presets, but not for fields.

This PR resolves the inconsistency, which is very useful for some presets.

### DIFF
--- a/modules/svg/icon.js
+++ b/modules/svg/icon.js
@@ -10,3 +10,16 @@ export function svgIcon(name, svgklass, useklass) {
             .attr('class', useklass);
     };
 }
+
+/** @param {string} url */
+export function svgIconExternal(url) {
+    /** @param {import("d3-selection").Selection} selection */
+    return function drawIcon(selection) {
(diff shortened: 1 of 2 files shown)


tyrasd (commented):
I'm not sure if this is a good idea: "Regular" presets don't use external [icons](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) in general. This has the benefit that consumers of the presets can predict which icons might appear. External icons were introduced for NSI because it was deemed unfeasible to maintain an icon set for the large amount of brands in that data set. This is however generally not the case for regular presets from the `id-tagging-schema`. The icons in the example screenshot could simply be added to temaki, couldn't they?

tordans:
> "Regular" presets don't use external [icons](https://github.com/ideditor/schema-builder/blob/main/ICONS.md) in general.

I agree and added https://github.com/ideditor/schema-builder/issues/161 so we can clarify this in the schema repo at some point.

PR #10256, opened 2024-05-25
render shared bicycle/foot paths with a dedicated colour

### LINKED ISSUE #2327: take into account access tags for path rendering
See for example http://www.openstreetmap.org/way/162926405 - new user is retagging it to highway=footway despite messages (see http://osmlab.github.io/osm-deep-history/#/way/162926405 ) as it looks clearly wrong in iD.

[highway=path, bicycle=designated, foot=designated, segregated=yes/no] is a standard way to tag way that is both cycleway and footway.

See for example http://wiki.openstreetmap.org/wiki/Key:segregated

note that JOSM renders differently:
highway=path (equivalent to iD style)

following are rendered like highway=cycleway/highway=footway with varying proportion of used color in line.
highway=path; bicycle=designated
highway=path; foot=designated
highway=path; bicycle=designated; foot=yes
highway=path; foot=designated; bicycle=yes
highway=path; foot=designated; bicycle=designated

Can iD at least stop encouraging bad edits?

### LINKED ISSUE #8485: Display of cycleways in map view
There are several ways to mark up cycleways. They are considered equal, see  #6836. Yet, the iD editor does not show them the same! Recent discussion in the Austrian forum about useless re-tagging of cycleways circled around opting out of the preset, but the preset was considered valuable.

If correctly mapped cycleways would appear as what they are (like in OSM-Carto e.g.), this re-tagging might not happen at all.

Pictures: User sees a path; Searches for bicycle; Creates a cycleway - which it was, from the start

bikepath-1
bikepath-2

### LINKED ISSUE #10139: Styling of `hw=path + bicycle=designated` as Path-Line with blue (cycleway) dashes
### Description

There is this tagging schema for shared foot-bike-path that uses `hw=path + bicycle=designated`.

It would be great, if we found a way to separate those path visually from regular `highway=path`.

My idea: Lets use the light brown base color of the path, but change the dark brown dashes to the blue of cycleways whenever the tag combination is `hw=path + bicycle=designated`.

Would that work? Can the styling system do this kind of matching?

Example: https://www.openstreetmap.org/edit?way=387515184#map=19/52.44572/13.44526

### Screenshots

image

k-yle:
Closes #8485, Closes #10139, closes #2327

Currently, it's impossible to tell the difference between the `Cycleway` and `Cycle + Foot Path` presets.

There are [6 special countries](https://github.com/openstreetmap/id-tagging-schema/blob/e11a6d2c7468267ea4e579a8dc8a7e4487f34729/data/presets/highway/cycleway/bicycle_foot.json#L4-L9) that use a different preset, which is simiarly confusing, since it renders exactly the same as `crossing=*`.

This PR makes the styling consistent for `bicycle=designated` + `foot=designated`, regardless of whether you use `highway=path` or `highway=cycleway`

<table>
<tr>
 <td><strong>Before
 <td>
<tr>
 <td><strong>After
 <td><br />(only the bottom 2 have changed)
</table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased (2.40.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* Render shared bicycle/foot paths with a dedicated colour ([#10256], thanks [@k-yle])
 #### :scissors: Operations
 * Correctly split a way on all selected vertices when three or more nodes are selected ([#12120])
 * Create fewer points when circularizing small features (and vice versa): the number of vertices in the resulting circle is now dynamic between a 12 and 32 depending on the radius of the circle (instead of always resulting in 19 nodes) ([#12139])
@@ -63,6 +64,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Add type annotations to `context.js` module ([#11589], thanks [@k-yle])
 
 [#9406]: https://github.com/openstreetmap/iD/issues/9406
+[#10256]: https://github.com/openstreetmap/iD/pull/10256
 [#11589]: https://github.com/openstreetmap/iD/pull/11589
 [#12010]: https://github.com/openstreetmap/iD/pull/12010
 [#12050]: https://github.com/openstreetmap/iD/issues/12050
@@ -75,6 +77,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#12110]: https://github.com/openstreetmap/iD/issues/12110
 [#12120]: https://github.com/openstreetmap/iD/issues/12120
 [#12139]: https://github.com/openstreetmap/iD/pull/12139
+
 
 
 # 2.39.5
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -588,6 +588,27 @@ path.line.stroke.tag-highway-cycleway,
     stroke: #fff;
 }
 
+/* shared paths */
+path.line.casing.tag-highway-cycleway.tag-foot-designated:not(.tag-crossing),
+path.line.casing.tag-highway-path.tag-bicycle-designated.tag-foot-designated:not(.tag-crossing) {
+    stroke: #fff;
+}
+.preset-icon .icon.tag-highway-cycleway.tag-foot-designated,
+.preset-icon .icon.tag-highway-path.tag-bicycle-designated.tag-foot-designated {
+    color: #b458ed;
+    fill: #fff;
+}
+path.line.stroke.tag-highway-cycleway.tag-foot-designated,
+path.line.stroke.tag-highway-path.tag-bicycle-designated.tag-foot-designated,
+.preset-icon-container path.casing.tag-highway-cycleway.tag-foot-designated,
+.preset-icon-container path.casing.tag-highway-path.tag-bicycle-designated.tag-foot-designated {
+    stroke: #b458ed;
+}
+.preset-icon-container path.stroke.tag-highway-cycleway.tag-foot-designated:not(.tag-crossing),
+.preset-icon-container path.stroke.tag-highway-path.tag-bicycle-designated.tag-foot-designated:not(.tag-crossing) {
+    stroke: #fff;
+}
+
 /* bridleways */
 .preset-icon .icon.tag-route-horse,
 .preset-icon .icon.tag-highway-bridleway {
--- a/modules/svg/tag_classes.js
+++ b/modules/svg/tag_classes.js
@@ -14,7 +14,7 @@ export function svgTagClasses() {
         'oneway', 'bridge', 'tunnel', 'embankment', 'cutting', 'barrier',
         'surface', 'tracktype', 'footway', 'crossing', 'service', 'sport',
         'public_transport', 'location', 'parking', 'golf', 'type', 'leisure',
-        'man_made', 'indoor', 'construction', 'proposed'
+        'man_made', 'indoor', 'construction', 'proposed', 'bicycle', 'foot'
     ];
     var _tags = function(entity) { return entity.tags; };
 
@@ -61,15 +61,9 @@ export function svgTagClasses() {
 
         // pick at most one primary classification tag..
         for (i = 0; i < primaries.length; i++) {
-            k = primaries[i];
+            k = primaries[i].replace(':', '_');
             v = t[k];
             if (!v || v === 'no') continue;
-
-            if (k === 'piste:type') {  // avoid a ':' in the class name
-                k = 'piste';
-            } else if (k === 'building:part') {  // avoid a ':' in the class name
-                k = 'building_part';
-            }
 
             primary = k;
             if (statuses.indexOf(v) !== -1) {   // e.g. `railway=abandoned`


tordans:
@k-yle thanks for kicking this off! I am wondering, if we can find a system in how we use the background vs. dashed line.

- Background for highway class
- Dashed line for modifiers and access
- _pathy_ things are brown
- foot stuff is white
- bicycle stuff is blue

My mental model has a hierarchy like this on this topic:
image

We could translate this in something like:
image

Document: https://docs.google.com/presentation/d/1zQ6mqbGgyRRwq9Zakof_kPfAzX1JeRVplxMr4SLKMsQ/edit?usp=sharing

Notes:
* hw=cycleway + foot is [used a million times](https://taginfo.openstreetmap.org/tags/highway=cycleway#combinations)
* hw=cycleway should probably also trigger on foot=yes because any non-cycle traffic is the modifier
* the background colors need to be slightly different when modifiers are applied to make them look the same (the colors interact which makes them look different)
* The new blue for cycleways would give a nudge to the rendering on osm carto which I think is a plus
* on `footway=sidewalk`: I wonder if we should also change the dash style a bit and not just the color. the color change is super subtle. hw=stairs has a smaller dash, maybe something in between? (but I cannot simulate that in powerpoint :-D) – however: see below on crossings

tordans:
**And for the crossings…**

Note: In https://github.com/openstreetmap/id-tagging-schema/pull/1201#issue-2261598786 I added crossing presets to match all the main way presets.

Given that both background and dash color are now reserved what else can we change?
- **Change the dash style** (shorter, longer)
- **Add a solid border** like on highway hw=service+service=alley image
- **Add a dashed border** like on hw=path image
- **Modify the background color** to something darker/lighter

Thoughts:
- A different dash style is probably the easiest. We use it for hw=strairs but those are green dashes and distinct. It kind of indicates "slow down", which is OK for a crossing.
- Dashed borders collide with the markings presets; it could look like they are a markings-visualization
- Solid borders would work. They could collide with other service styles but should never directly connect so no real issue here. However they kind of signal a higher class of road or signal safety which does not fit for a crossing.
- Modified backgrounds is hard when we use white for footways. We could try something like a "neutral dark gray" for the white, a "warm dark gray" for the path and a "blue dark gray" for the bicycle ways.

k-yle:
Sorry for the delayed reply, this comment suprised me: 
> My mental model has a hierarchy like this on this topic:
> [image](https://github.com/openstreetmap/iD/assets/111561/87790a2a-d1e4-42ff-89e9-65fcbf806848)

because I thought that (`highway=cycleway` + `foot=designated` + `cycleway=designated`) was exactly the same as (`highway=path` + `foot=designated` + `cycleway=designated`). Is there a wiki page which explains the difference? I can't contribute much to this discussion since my understanding of this situation seems to be flawed

tordans:
> > My mental model has a hierarchy like this on this topic:
> > [image](https://github.com/openstreetmap/iD/assets/111561/87790a2a-d1e4-42ff-89e9-65fcbf806848)
> 
> because I thought that (`highway=cycleway` + `foot=designated` + `cycleway=designated`) was exactly the same as (`highway=path` + `foot=designated` + `cycleway=designated`). Is there a wiki page which explains the difference? I can't contribute much to this discussion since my understanding of this situation seems to be flawed

You are right. Illustrating stuff is always hard and I mixed two things which did not work…

Looking at it some more, I think the main issue is: You are introducing the new pink color for foot+bike traffic ways.
But hw=path is also foot+bike traffic, just not explicitly.

Does this make sense: The miss-match of Column D vs. I? 

image

[Here is the doc…](https://docs.google.com/spreadsheets/d/1dJH3nNLIjr6TNSC-gPvRDxHHJL2AOd32G_AQ-Tv1zXI/edit?gid=0#gid=0)

I don't have a solution ATM.

I really like to separate the crossing styles. But maybe we should look more at options from https://github.com/openstreetmap/iD/pull/10256#issuecomment-2210151779 – when we introduce a new modifier to the style, we make it easier to work with what we have for the rest.

I am not sure yet how to handle the access. vs. highway class in the styles.
The main issue is IMO, that hw=path is open for bike+foot traffic by default. And that usually we show the highway class in our style, not the access. 

Have to stop now, but 🤔 …

Docs
- https://wiki.openstreetmap.org/wiki/Tag:highway%3Dpath

---

Updated: I added an "Another try" Column. 
- The two Cycle & Foot Path presets use the same background  as path, because all three are open for bike+foot
- But that system breaks again with 740k footway+bicycle in line 12
- So this 🤔…

k-yle:
> Change the dash style

I think this might be a step backwards from the current situation, where the colours make it easy to distinguish most presets, so I haven't considered it below

> Add a solid border
> Add a dashed border

These two are already used to distinguish paved/unpaved, so we can't really change the line casing

> Modify the background color

I think this is the only option...


> The main issue is IMO, that hw=path is open for bike+foot traffic by default.

I overlooked that until now... maybe we need a separate category for unspecified access?

Here's what this could look like, using a modified version of [your first diagram](https://github.com/openstreetmap/iD/pull/10256#issuecomment-2210124911), including some of the changes from [this table](https://github.com/openstreetmap/iD/pull/10256#issuecomment-2227296876):

_(upper line = normal, lower line = crossing)_


- this assumes 4 colours for the 4 categories
  - explicit foot = $`{\color{#938888} \blacksquare}`$ brown
  - explicit both = $`{\color{#a267ea} \blacksquare}`$ purple ([`DE:240`](https://osm.wiki/File:Zeichen_240_-_Gemeinsamer_Fu%C3%9F-_und_Radweg,_StVO_1992.svg) / [`AU:R8-2`](https://osm.wiki/File:Australia_road_sign_R8-2.svg))
  - explicit cycle = $`{\color{#a3c9f2} \blacksquare}`$ blue
  - implicit access = $`{\color{#ddd8c5} \blacksquare}`$ beige 
- crossings could use the same colours, but the foregound/background is inverted
- I don't know how to handle the 2 situations in $`{\color{#e29999} \blacksquare}`$ red. Do people use these tags to indicate that one mode has more priority?

matkoniecz (commented):
I admit that I braced myself to writing comment of "sorry, but this does not work well..."

While in testing on area known to me it looks and works really well!

I am not opposed to trying some of tweak proposed in discussion, if anyone want to implement them.

But this PR (proposed in May 2024!) seems a clear improvement to me and I would suggest merging it as-is.

matkoniecz (commented):
is it intentional that `highway=cycleway foot=designated` is blue, not as new purple?

And `highway=footway bicycle=designated` continues to render as footway, not as new purple?

Note that https://github.com/openstreetmap/id-tagging-schema/pull/1792 if merged would make it more acute

matkoniecz:
BTW, it also fixes #2327 - can you update description so it will autoclose it on merge?

k-yle:
> is it intentional that `highway=cycleway` `foot=designated` is blue, not as new purple?
> 
> And `highway=footway` `bicycle=designated` continues to render as footway, not as new purple?

yes, that's intentional (for now), so that the rendering exactly matches the presets. 

If https://github.com/openstreetmap/id-tagging-schema/pull/1792 goes ahead, then I suggest updating the rendering at the same time as the presets.

matkoniecz:
> so that the rendering exactly matches the presets.

Note that "Cycle & Foot Path" in UK or Australia will produce

```
bicycle=designated
foot=designated
highway=cycleway
```

tagging. It seems preferable to support also that.

Note: the same preset name will produce different effect when used in Poland.

As both "Cycle & Foot Path" (with `highway=cycleway` as primary tag) and "Cycle & Foot Path" (with highway=path` as primary tag) are effectively the same also rendering should be the same.

k-yle:
> It seems preferable to support also that

currently both options are supported (`=path` and `=cycleway`), they both get the purple colour if `bicycle=designated` and `foot=designated`

matkoniecz:
Interesting. It should not be region specific.

Now in my testing only stuff like 

```
foot=designated
highway=cycleway
```

without bicycle=designated breaks it and it was discussed above. I will look again at it.

matkoniecz:
I think that at least

```
bicycle=designated
foot=designated
highway=cycleway
```
and

```
foot=designated
highway=cycleway
```

and

```
bicycle=designated
foot=designated
highway=path
```

should get this purple styling, as meaning is the same in each and they fall into cycle-foot path in iD presets. So different colours would be confusing.

Ideally also

```
bicycle=designated
foot=designated
highway=footway
```

would get it, but it is actually listed as Foot Path by iD presets so I understand special treatment here.

EDIT: highway=cycleway foot=designated in Germany get cycleway preset, but this is more of fixable iD tagging schema bug

matkoniecz:
Sorry that it is change from earlier comments and sorry for earlier confused review.

k-yle:
done, `highway=cycleway` + `foot=designated` now gets the same purple colour, so that the colour always matches the presets

matkoniecz (commented):
thanks! Now I am entirely happy with it, I hope we will see it soon

tyrasd (approved):
Thanks a lot @k-yle and @matkoniecz for completing this great improvement!

tordans:
I am happy to see this merged and sorry for being late with sending this message (I had it pending for a few days).

I would appreciate it if someone double-checked that I put this into words the right way.
(I wrote this message a few times and had a hard time finding the right pattern and words…)

> The new style indicates a way that is designated for bike and foot traffic in cases where we have established tagging for this kind of infrastructure.
>
> `highway=path`: When `bicycle+foot=designated` are explicitly tagged, the way is displayed in our new, shared infrastructure style.
> `highway=cycleway`: When `foot=designated` is added, the way is displayed in our new, shared infrastructure style (`bicycle=designated` can but does not have to be tagged; it is indicated by the highway class)
> `highway=footway` does not change; regardless of ` bicycle+foot=designated` because there is not established tagging that fits this new style (which only visualizes `designated`).
>
> Crossings are updated as well to follow this system.

Test screenshot ([L](https://pr-10256--ideditor.netlify.app/#disable_features=boundaries&map=18.24/52.36147/13.61916&background=Bing)):

> grafik

<details><summary>Tags for the first row of the screenshot</summary>

Second row `path`, last row `cycleway`.



<p>


```
highway=footway
```

```
foot=designated
highway=footway
```


```
bicycle=designated
highway=footway
```


```
bicycle=designated
foot=designated
highway=footway
```

**Crossings**

```
bicycle=designated
crossing=unmarked
footway=crossing
highway=footway
```

```
bicycle=designated
crossing=uncontrolled
footway=crossing
highway=footway
```

```
bicycle=designated
crossing=traffic_signals
footway=crossing
highway=footway
```

</p>
</details>

matkoniecz:
>  highway=footway does not change; regardless of  bicycle+foot=designated because there is not established tagging that fits this new style (which only visualizes designated).

and if that in fact is also established tagging schema: it should be added to iD tagging schema (currently `highway=footway bicycle=designated` is not classified as shared foot-cycle path, though it is listed in access tagging)

matkoniecz:
BTW, I opened https://github.com/openstreetmap/id-tagging-schema/issues/2169

You have edge case of 

```
highway=cycleway
bicycle=designated
foot=designated
```

in Germany/etc and

```
highway=path
bicycle=designated
foot=designated
```

in Netherlands etc.

k-yle:
> [^](https://github.com/openstreetmap/iD/pull/10256#issuecomment-4211687070) [...] (`bicycle=designated` can but does not have to be tagged; it is indicated by the `highway` class)

correct, but imo that situation isn't worth highlighting, because iD encourages people to explicitly add `bicycle=designated` in that situation: 

image

this is existing logic from [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) which was not changed in this PR.

matkoniecz:
> correct, but imo that situation isn't worth highlighting, because iD encourages people to explicitly add bicycle=designated in that situation:

though note https://github.com/openstreetmap/id-tagging-schema/pull/1792

PR #10258, opened 2024-05-25
fix wikidata styling not applying to vertices

k-yle:
Nodes with a `*:wikidata` tag normally render in a special grey colour. However, this does not work for vertices.

This PR fixes the bug.

<table>
<tr>
 <td><strong>Before
 <td><strong>After
<tr>
 <td>
 <td>
</table>
only the bottom left example has changed

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -294,12 +294,14 @@ text.point {
 
 
 /* Wikidata-tagged */
-g.point.tag-wikidata path.stroke {
+g.point.tag-wikidata path.stroke,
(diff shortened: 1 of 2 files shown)


tyrasd (approved):

tyrasd:
I agree that if there is a dedicated styling, it should be the same for nodes and vertices. But to be fair, I don't know how useful the special styling is in the first place. #6389 had initially introduced this styling, but does not really give an argument about _why_ this was introduced. Am I overlooking a use case here?

PR #10264, opened 2024-05-28
Fix typo

linfindel:
repectively -> respectively


linfindel:
Just to clarify, the old changes I removed was the backdrop blur. That is not included in this PR.

tordans (approved):

Issue #10249, opened 2024-05-21
Adding installation instructions in Chinese


brianh-osu:
Hiya!

I can see how this project can benefit from a Chinese version for the README and or installation instructions. The repo is used by international corporations and international users worldwide. Although English is primarily recognized, there may be users who will better benefit from a Chinese translation.

### **URL**
https://github.com/openstreetmap/iD/blob/develop/README.md

### **How to reproduce the issue?**
n/a

### **Screenshot(s) or anything else?**
n/a

### **Which deployed environments do you see the issue in?**
The most up to date version of iD

### **What version numbers does this issue effect?**
The most up to date version of iD

### **Which browsers are you seeing this problem on?**
Chrome

pnorman:
What languages would the readme need too be translated into, and who would both translate it and keep it up to date?

kartik-212004:
I'm interested in working on this issue and would appreciate being assigned to it.

tyrasd:
Closing as this would not be maintainable by the devs here.

k-yle:
This has already been implemented here: https://github.com/openstreetmap/iD/wiki/Build-Instructions-(Chinese)

PR #10270, opened 2024-06-06
Chinese translation for Build/installation instructions. 

brianh-osu:
Hi there!

This is related to issue: https://github.com/openstreetmap/iD/issues/10249

The language is translated to Chinese, and would have to be maintained by contributors. I've added the "Updated date" at the bottom of the document so that the reader is aware that the document can be updated occasionally.

The document in concern is: https://github.com/openstreetmap/iD/wiki/Build-Instructions-(Chinese) ; I'm not certain if a pull request needs to be created for this type of request (as it is for a wiki upload instead of actual code repo). 

Thank you in advance.


tordans:
Hey, this PR does not have any changes, so I am closing it.

PR #10278, opened 2024-06-11
Added `railway=ventilation_shaft` as area key

### LINKED ISSUE #1263: Don't render `railway=ventilation_shaft` as rails
**Describe the bug**
Features tagged with `railway=ventilation_shaft` and without other "area" tags are rendered as rails.

**To Reproduce**
1. https://www.openstreetmap.org/edit?way=685774055#map=19/40.80283/-73.95643

**Expected behavior**
Render them as areas?

**Screenshots**
image

**Additional context**
It would be nice to have a preset but there is also the competing tag `man_made=venttilation_shaft`...

Dimitar5555:
Resolves https://github.com/openstreetmap/id-tagging-schema/issues/1263

### DIFF
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -58,7 +58,8 @@ export var osmAreaKeysExceptions = {
         station: true,
         traverser: true,
         turntable: true,
-        wash: true
+        wash: true,
+        ventilation_shaft: true
     },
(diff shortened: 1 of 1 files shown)


k-yle (approved):
image

I can confirm that this works

PR #10283, opened 2024-06-19
Add waterway=flowline to waterway tags representing directional water flows

waldyrious:
Currently, lines with `waterway=flowline` are represented as generic waterway features, and don't get the arrows pointing in the water flow direction along their length (as is the case with rivers, streams, canals, etc.). This change adds the tag so the representation is consistent with the other flowing waterway tags.

### DIFF
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -167,6 +167,7 @@ export var osmOneWayTags = {
         'ditch': true,
         'drain': true,
         'fish_pass': true,
+        'flowline': true,
         'pressurised': true,
         'river': true,
         'spillway': true,
@@ -245,7 +246,7 @@ export var osmRailwayTrackTagValues = {
 
 // "waterway" tag values for line features representing water flow
 export var osmFlowingWaterwayTagValues = {
-    canal: true, ditch: true, drain: true, fish_pass: true, river: true, stream: true, tidal_channel: true
+    canal: true, ditch: true, drain: true, fish_pass: true, flowline: true, river: true, stream: true, tidal_channel: true
 };
 
 // Tags which values should be considered case sensitive when offering tag suggestions


JesseWeinstein:
It might be worth making them slightly visually different, just so it's easier to tell if a waterway thru a water body has been converted yet. Maybe a slightly different shade of blue?

tordans:
For easy reference: https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dflowline , ~800 usage globally.

tordans:
@waldyrious what about https://wiki.openstreetmap.org/wiki/Tag:waterway%3Dlink with 3k usage. The wiki sounds like this is something that indicates a direction as well?

waldyrious:
> The wiki sounds like this is something that indicates a direction as well?

Does it? I read the opposite:

> Unlike most [waterway](https://wiki.openstreetmap.org/wiki/Key:waterway)=* values, the tag does not model a discrete physical feature and **does not imply a flow direction**.

(emphasis mine)

waldyrious:
> It might be worth making them slightly visually different, just so it's easier to tell if a waterway thru a water body has been converted yet. Maybe a slightly different shade of blue?

Sounds good to me, but I didn't want to do too dramatic a change. Happy to add the change if maintainers agree it makes sense.

tordans:
> Does it? 

You are right. I think I was thinking about it because of the routing part of the wiki but that has nothing to do with the direction.

waldyrious:
@tordans @JesseWeinstein any suggestions on who I should ping (here or elsewhere) for additional feedback on the suggestion of using a different visual style?

tordans (approved):

tordans:
@waldyrious ATM Martin is the only one who can merge. We are working on adding community contributors that can help with easy PRs like this one. Once that is set up, I will be able to merge as well…

waldyrious:
That's great to hear! But I was asking specifically about making additional changes to the rendering of `waterway=flowline` ways, as @JesseWeinstein proposed [above](https://github.com/openstreetmap/iD/pull/10283#issuecomment-2178396819), so that I can know whether there's more work for me to do in this PR as opposed to just wait for the merge. Is this something that Martin alone can decide? Do you or others have thoughts on the matter?

tyrasd (approved):
this change is fine as is, thanks :+1: 

> additional changes to the rendering of `waterway=flowline` ways

Feel free to propose additional rendering changes as a follow-up PR. :blush:  If it were me, I'd think about increasing the width to the same as `waterway=river` (because these are meant to be used for large bodies of water), and perhaps experiment with some kind of dash pattern to represent that these "waterways" are virtual lines to connect the network?

image

The mentioned approach of a slightly different shade of blue seems to also work fine and has the benefit of not clashing with the dashed rendering of intermittent waterways (for a feature tagged as `waterway=flowline` and `intermittent=yes`)

image

waldyrious:
Thanks for merging and for the suggestions! I like both of them, actually :sweat_smile: @JesseWeinstein do you have any preference between the two rendering options?

JesseWeinstein:
I still like the different shade of blue better. Glad to get this PR merged though, in any case! Thank you both.

tyrasd:
ok, these features a rendered in a slightly different shade of blue now.

JesseWeinstein:
Ah, it looks really nice: https://ideditor.netlify.app/#background=MAPNIK&disable_features=boundaries&map=16.62/43.24503/-79.21668

The next thing is to add a preset, so flowlines don't just show up as "Waterway feature". If no-one else has done this, I'll try and pull together a PR soon.

(and done so: https://github.com/openstreetmap/id-tagging-schema/pull/1300 )

JesseWeinstein:
I opened a formal proposal, here: https://wiki.openstreetmap.org/wiki/Proposal:Flowlines

Issue #10128, opened 2024-02-25
`highway=cyclist_waiting_aid` `side` gets reversed with the way


RudyTheDev:
When a way is reversed and there is a `highway=cyclist_waiting_aid` (openstreetmap/id-tagging-schema#1066 / openstreetmap/id-tagging-schema#1069) with a `side` value, then this value also gets "reversed". However, `side` represents the side of the waiting aid from the _cyclist's_ perspective, not from the way's perspective, so this actually sets an incorrect value.

For example, add `highway=cyclist_waiting_aid` + `direction=forward` + `side=right` to a way's node. Then reverse the parent way. Observe that `direction=backward` got swapped correctly, but also `side=left` got swapped, which is now on the incorrect side.

SupaplexOSM:
To avoid making things too easy:
- If it's a oneway cycleway/path, the side flip would be OK, because iD assumes that the direction of travel also changes in this case (and does not become `oneway=-1`).
- However, if it is a oneway road that can also be used by cyclists in the opposite direction, the `side` flip would be wrong, as cyclists can continue to ride in the previous direction.

I'm sorry that the tagging has such a pitfall in this case. (This issue is also mentioned in [this forum thread](https://community.openstreetmap.org/t/approved-feature-proposal-cyclist-footrests-and-handrails-cyclist-waiting-aid/105762/3))

tyrasd:
oh… I had not thought of that.

This might be pretty hard to handle 100% correctly. :grimacing: But how likely is it that this will actually be a real problem in practice? I mean, given that these features are by definition always placed in front of certain obstacles like intersections: When the direction of travel of, say, a oneway bikepath is flipped in reality, then that would mean that previously existing waiting aids are now all placed in locations where they are not useful anymore (e.g. right after intersections instead of right before them). So… realistically, the waiting aids would have been moved in reality as well, meaning that they would have to be fully remapped accordingly anyway.  :thinking: Right?

RudyTheDev:
Is this the only feature that uses `side` this way? May be it's better to either change the tag or make it way-dependent before "it's too late". If all editors are flipping `side` by default, may be it's a bad idea to set a precedent for exceptions. I had assumed tag flipping was hard-coded per feature, depending on what it means. But I guess it had always meant "side from way perspective".

tyrasd:
[...]

PR #10303, opened 2024-07-01
render the `side` tag

k-yle:
This PR renders the [`side=*`](https://osm.wiki/Key:side) tag, similar to how [`direction=*`](https://osm.wiki/Key:direction) is rendered. 

`side=*` is used with [`highway=cyclist_waiting_aid`](https://osm.wiki/Tag:highway=cyclist_waiting_aid), but it is currently slow and tricky to use this tag. This PR aims to make it much more intuitive to map the `side`. It can be tested [from here](https://kyle.kiwi/iD).


modules/osm/node.js
```diff
@@ -130,14 +149,15 @@ Object.assign(osmNode.prototype, {
 
             Object.keys(nodeIds).forEach(function(nodeId) {
                 // +90 because geoAngle returns angle from X axis, not Y (north)
-                results.push(
-                    (geoAngle(this, resolver.entity(nodeId), projection) * (180 / Math.PI)) + 90
-                );
+                results.push({
+                    type: isSide ? 'side' : 'direction',
+                    angle: (geoAngle(this, resolver.entity(nodeId), projection) * (180 / Math.PI)) + (isSide ? 0 : 90)
```
1ec5, line 179: Good idea, though I wonder if reusing the “cone” would risk confusing some users. Some mappers are also using `side=*` on `traffic_signals=*` vertices. This change will cause these vertices’ cones to point away from the street, which would be inconsistent with the cones for standalone `traffic_signals=*` nodes. Plus, it might be necessary to combine `side=*` with `direction=*` in some cases. In the screenshot above, this would imply four different “faces” rather than two. Maybe `side=*` could act as more of a mask, chopping the directional cone in half?

tordans: I was wondering about how to visualize the `side` as well. 

The `direction` is more of a "looking direction" or "affected area". 

But the `side` is more of "an anchor" or a "connection point" or a "this belongs to over there" indicator. However I don't see a way to visualize those well. Especially when the actual "anchor point" is unknown (we only now "it is to the right").

The other UI we use to show direction of elements is on `barrier=retaining_wall` and such with the small arrows next to the line. Maybe this is something we can build upon for this UI? Like one of those arrows next to the node?

tordans: > Like one of those arrows next to the node?

Something like … image?

k-yle: > Good idea, though I wonder if reusing the “cone” would risk confusing some users.
> [...]
> Maybe `side=*` could act as more of a mask, chopping the directional cone in half?

That's a cool idea, although it might be confusing if `side`+`direction` renders differently to `side` and `direction` on their own? Here's what it could look like:

`⊿`: 
`◭`: 

> Like one of those arrows next to the node?

Something like this maybe?

`△`: 
`→`: 
`⏢`: 
`▭`: 
`T`: 

Any other ideas would be greatly appreciated...

tyrasd:
For cyclist waiting aids the `side` tag is working a bit differently than normal `side` tags. :see_no_evil:  We need to check if this rendering is working for them as documented on the wiki.

modules/osm/node.js
tyrasd: I like the `△`  :+1:

tyrasd: Another way to do it could be to shift the vertex icon itself to the side, like this:

<br>


But I don't think that looks good and could be even confusing (e.g. it's unclear where other objects would snap to, etc.).

k-yle:
Following a discussion last week, we will not render the `side` tag for [`highway=cyclict_waiting_aid`](https://osm.wiki/Tag:highway=cyclist_waiting_aid) on a bidirectional cycleway, because the wording in [the proposal](https://osm.wiki/Proposal:Highway%3Dcyclist_waiting_aid) is too ambiguous 😕 I wanted to [fix the definition](https://github.com/openstreetmap/iD/issues/10128#issuecomment-1965268713) a year ago, but [now it's too late](https://overpass-turbo.eu/s/27eq)...

PR #10324, opened 2024-07-13
Improve types

tordans:
The optional params need to be added as `[varName]` not (or not only) with `?number` (Questionmark).

### DIFF
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -201,7 +201,7 @@ export function rendererFeatures(context) {
     });
 
     defineRule('aerialways', function isAerialways(tags) {
-        return tags?.aerialway !== 'yes' && tags?.aerialway !== 'station';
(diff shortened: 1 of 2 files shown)


k-yle (approved):
oops, one of these was my fault 🤭

modules/renderer/features.js
```diff
@@ -194,9 +201,7 @@ export function rendererFeatures(context) {
     });
 
     defineRule('aerialways', function isAerialways(tags) {
-        return tags.aerialway &&
-            tags.aerialway !== 'yes' &&
-            tags.aerialway !== 'station';
+        return tags?.aerialway !== 'yes' && tags?.aerialway !== 'station';
```
k-yle: This has changed the logic, if `tags = {}`, then the rule now returns `true`. If the goal is to keep typescript happy, then this would work:

```suggestion
        return !!tags?.aerialway && tags.aerialway !== 'yes' && tags.aerialway !== 'station';
```

Issue #10322, opened 2024-07-13
Add filter for things under ground, level=-*


tordans:
### Description

When mapping in areas like https://www.openstreetmap.org/edit?note=4139556#map=20/52.51122/13.40564, I want to be able to hide the features that are underground. Especially to prevent my edits to snap and glue themselves to those underground features.

AFAIK there is no way to do this today.

https://github.com/openstreetmap/iD/pull/9584 could be a solution but will be more complex to finish, it looks like. So in the meantime, maybe we can add another filter for things "Underground"?

GoMap has such a filter and that helps a lot.

kjonosm:
See also #7261

tyrasd:
closing in favor of #7261

Issue #7261, opened 2020-01-17
Hide underground features


forteller:
I appreciate the fact that it is possible to see and add important underground features to OSM! But I personally have no idea where any of it is, and I can't help in that matter. I'm convinced the same is true for at least 95% of all contributors. 

This means that the underground features are just in the way, and sometimes really annoying, a great majority of the time that the iD editor is in use. You might find yourself attach other features to them, or they are just confusing or in the way when you try to interpret the background photo. Especially in areas with low quality satelite photos. 

I tried to find an option to disable all underground features, but could not find it.

I would appreciate if this would be possible to add. And actually I think the most logical thing to do would be to have this disabled by default, given the fact that I'm quite certain that almost no one can actually help with these things (and the few that can probably has an easier time enabling things than many others have disabling them, if that makes sense).

Thanks!

quincylvania:
Hmm interesting idea. Do you have an area in mind where lots of underground features are causing issues? I'd support adding toggle but I'm not sure whether or not it should be disabled by default.

forteller:
Thanks for your fast and positive replies! ;)

Here's the most in one place that I found in one place with just a little bit of searching: https://preview.ideditor.com/master/#background=geovekst-nib&disable_features=boundaries&map=18.30/59.90626/10.74338

It's not that there's such a huge quantity at one place that's bothersome. It's more that almost everywhere I go in my city there's _something_ underground, and it's never useful to me, it's always only in the way.

forteller:
Here's a better example. This road is clearly drawn to the side of the actual road. Why? I don't know, but it could have something to do with the underground subway platform being on to of the actual road. And anyway, that makes it just a little bit harder for me to adjust the road to be where it should be.

![Screenshot from 2020-01-19 17-19-32](https://user-images.githubusercontent.com/343561/72684358-4bfa4680-3ae0-11ea-8b57-a77e894b2475.png)

Underground cables are worse, btw, because their lines are so thin, you often don't know their there before whatever you're drawing gets stuck to it. And I'm guessing most volunteer editors don't know about the trick with holding down the Alt key.

Mannshoch:
I would propose a Vertical Slider with two buttons to drag. Every level between those two could be seen So you could say "show me only elements between +2 and 0".

forteller:
That's a very cool idea :)

quincylvania:
Related: #4780. Indoor and underground toggles should probably be part of the same control.

[...]

PR #10325, opened 2024-07-13
Filter underground

tordans:
Attempt at fixing ~~https://github.com/openstreetmap/iD/issues/10322~~ #7261 by adding another filter.

However, this does not work ATM. It looks like a feature is only hidden when all filters hide it?
In my example
- http://127.0.0.1:8080/#background=Berlin-2023&disable_features=traffic_roads,underground,boundaries&id=n-1&map=21.41/52.51124/13.40554
- https://www.openstreetmap.org/edit?way=199847350#map=20/52.51122/13.40564

… the way is still visible, because it is also handled by the railway filter.

Is there a way around this? Like telling the `defineRule` to overwrite other rules @tyrasd?

---

Links: 
- https://wiki.openstreetmap.org/wiki/DE:Tag:location%3Dunderground

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -929,6 +929,9 @@ en:
     past_future:
       description: Past/Future Features
       tooltip: "Proposed, Construction, Abandoned, Demolished, etc."
+    underground:
+      description: Unterground
+      tooltip: "Unterground features or negative level"
     others:
       description: Other Features
       tooltip: "Everything Else"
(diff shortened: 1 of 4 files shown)


tyrasd:
Yeah, the problem here is that the current feature type selection feature expects that a feature can only ever be in one of the defined types. Additional filters that work on a per-property (e.g. the _is underground_ here) don't really fit into that… What we probably need instead is a separate filter to enable the visibility of features on a per _level_ manner, as one would also need for indoor mapping. That would need to be implemented in a separate module / UI component, though.

PR #10333, opened 2024-07-16
Fix localized field hidden despite nonempty subkeys

### LINKED ISSUE #10323: Localized field hidden despite nonempty subkeys
The Name field is a `localized` field that displays not only `name=*` but also a very large number of potential `name:xyz=*` subkeys. The Name field is universal, so it’s enabled for every preset. However, some presets don’t display it by default. For these presets, the field stays hidden even if the feature has `name:xyz=*` set.

## Steps to reproduce

To reproduce this issue, create a Windpump (`man_made=windpump`) and manually tag it with `name:es=El Gran Molino`. Observe that the Name field doesn’t appear.

No Name field despite name:es being set

## Impact

In general, it’s incorrect to set `name:xyz=*` without setting `name=*`. However, this is a subject of debate for some very international features, such as [oceans and seas](https://community.openstreetmap.org/t/atlantic-ocean-repeated-name-removal/101196). Even if it’s a tagging error, the field should automatically show because it isn’t empty. This will be a bigger problem for the name fields being added in openstreetmap/id-tagging-schema#215. Those fields are universal but aren’t included by default in any preset, because the local language isn’t guaranteed to need them all, even as some other language may need one of them.

## Diagnosis

The logic for showing a field needs a special case for fields of type `localized` that tests for any key beginning with the field’s key and a colon:

https://github.com/openstreetmap/iD/blob/aaa144191c5eea5d0155b262d89f88c924672431/modules/ui/field.js#L90-L102

We should use the same regular expression added in #9124 to test whether the field would show the subkey:

https://github.com/openstreetmap/iD/blob/aaa144191c5eea5d0155b262d89f88c924672431/modules/ui/fields/localized.js#L127-L131

samhoooo:
Fix https://github.com/openstreetmap/iD/issues/10323

- Add logic to show a field of type `localized` that has any key beginning with the field’s key and a colon
- Use the same regular expression added in https://github.com/openstreetmap/iD/pull/9124 to test whether the field would show the subkey

https://github.com/user-attachments/assets/1c8e6be9-ef55-4b80-9090-8f141b843825

### DIFF
--- a/modules/ui/field.js
+++ b/modules/ui/field.js
@@ -97,6 +97,15 @@ export function uiField(context, presetField, entityIDs, options) {
                 }
                 return false;
             }
+            if (field.type === 'localized') {
+                for (let tagKey in _tags) {
+                    // matches for field:<code>, where <code> is a BCP 47 locale code
+                    let match = tagKey.match(/^(.*):([a-z]{2,3}(?:-[A-Z][a-z]{3})?(?:-[A-Z]{2})?)$/);
+                    if (match && match[1] === field.key && match[2]) {
+                        return true;
+                    }
+                }
+            }
             return _tags[key] !== undefined;
         });
     }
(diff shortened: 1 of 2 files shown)


tyrasd (approved):
thanks, this fix looks good to me :+1:

PR #10371, opened 2024-07-31
allow to enter strings with quote marks in raw tag editor's text mode

### LINKED ISSUE #10369: quotation marks in the text field are ignored
- Find any point with tags
- Swtich from "list" of tags to "text"
- Modify tags, add tags or delete tags
- Add manualy any key with quotation marks, but not as a first character, e.g. `note = this is "TEST"`
- Switch back to "list" view - no tags have changed
- Switch back to "text" - they are still there

tyrasd:
…and use fully quoted strings for cases which have characters that need to be encoded

old: `tag=foo\"bar\"` (or `tag="foo\"bar\""`) -> new: `tag=foo"bar"` or `tag="foo\"bar\""`
old: `tag=foo\nbar` (or `tag="foo\"bar\""`)   -> new: `tag="foo\nbar"`

closes #10369

### DIFF
--- a/modules/ui/sections/raw_tag_editor.js
+++ b/modules/ui/sections/raw_tag_editor.js
@@ -332,7 +332,11 @@ export function uiSectionRawTagEditor(id, context) {
     function unstringify(s) {
         const isQuoted = s.length > 1 && s.charAt(0) === '"' && s.charAt(s.length - 1) === '"';
         if (isQuoted) {
-            return JSON.parse(s);
+            try {
+                return JSON.parse(s);
+            } catch {
+                return s;
+            }
         } else {
             return s;
         }
--- a/modules/ui/sections/raw_tag_editor.js
+++ b/modules/ui/sections/raw_tag_editor.js
@@ -321,21 +321,21 @@ export function uiSectionRawTagEditor(id, context) {
     }
 
     function stringify(s) {
-        return JSON.stringify(s).slice(1, -1);   // without leading/trailing "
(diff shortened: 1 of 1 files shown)


modules/ui/sections/raw_tag_editor.js
```diff
@@ -321,21 +321,21 @@ export function uiSectionRawTagEditor(id, context) {
...
-        if (s.length < 1 || s.charAt(0) !== '"') {
-            leading = '"';
-        }
-        if (s.length < 2 || s.charAt(s.length - 1) !== '"' ||
-            (s.charAt(s.length - 1) === '"' && s.charAt(s.length - 2) === '\\')
-        ) {
-            trailing = '"';
+        const isQuoted = s.length > 1 && s.charAt(0) === '"' && s.charAt(s.length - 1) === '"';
+        if (isQuoted) {
+            return JSON.parse(`"${s}"`);
```
1ec5: Should we catch an exception here if the JSON string ends up being invalid JSON? Or is this guaranteed to be working with valid JSON?

1ec5 (approved):

modules/ui/sections/raw_tag_editor.js
tyrasd: Good point. There could still be cases where a user enters an invalid quoted string, say something like `key="boo"far"`. Catching the exception here would allow us to only skip that particular offending tag, and still set the remaining tags properly. Which I think is better than the current situation where the full input field is ignored like reported in #10369

Issue #10366, opened 2024-07-30
Panoramax: Highlight selected sequence


danieldegroot2:
### How to reproduce the issue?

Ref #9941 

When selecting a sequence, the sequence is not highlighted (trace and bubbles get brighter colour), like for other providers.
This can be useful for areas with multiple sequences, when you _do_ want to use multiple dates (no date filter).
~~(and troubleshooting for issue linked below, where I had to use the date filter to know if it was actually the same sequence~~ sorry, fyi this metadata is available on panoramax website if you need to troubleshoot such issue)
However, I assume this is not implemented yet.

(Sidenote: there are no sequences in Bing Streetside, they are not connected traces eventhough they allow to move between them and between sequences)

Chrome 127.0.6533.73 on Windows 10

### Screenshot(s) or anything else?

Panoramax
Ref #10365

![image](https://github.com/user-attachments/assets/0679dd63-5c76-4250-a161-5ec9e029d0c2)

Kartaview
![image](https://github.com/user-attachments/assets/a0cdcc4e-5323-458c-bad7-588d3f2227c7)

### Which deployed environments do you see the issue in?

Development version at ideditor.netlify.app

### What version numbers does this issue effect?

2.30.0-dev

### Which browsers are you seeing this problem on?

Chrome

mattiapezzotti:
Thank you for noticing, this will be fixed in the next update

PR #10374, opened 2024-08-02
Fix panoramax highlight selected sequence

mattiapezzotti:
fix for #10366


tyrasd (approved):
Thanks!

FYI: I've added another condition to the "is photo highlighted" check to make it the same as the mapillary layer (see https://github.com/openstreetmap/iD/blob/2325de4/modules/services/mapillary.js#L662).

Issue #10377, opened 2024-08-03
Panoramax: Bubbles stay loaded below minimum zoom


danieldegroot2:
### How to reproduce the issue?

Ref #9941 

Bubbles start appearing at zoom 15 on fresh start / restart, but zooming out disappear only at z10 together with the layer.

Let's zoom in to this roundabout. (but let's skip to the proper zoom level)

https://ideditor.netlify.app/#background=fr.ign.bdortho&disable_features=boundaries&map=15.00/-20.9173/55.5424&photo_overlay=panoramax
Zoom 15: There is the bubbles

Zoom 15 => 10: If you zoom out now bubbles stay loaded in the already visited area eventhough you are below min. zoom. They only disappear at z10 with the rest of the layer.

Chrome 127.0.6533.89 on Windows 10

### Screenshot(s) or anything else?

![image](https://github.com/user-attachments/assets/384347c9-0c61-43c3-a318-dca0069fafec)

### Which deployed environments do you see the issue in?

Development version at ideditor.netlify.app

### What version numbers does this issue effect?

2.30.0-dev

### Which browsers are you seeing this problem on?

Chrome

PR #10386, opened 2024-08-05
Fix panoramax bubbles stay loaded below minimum zoom

mattiapezzotti:
fix for https://github.com/openstreetmap/iD/issues/10377


tyrasd (approved):
Thanks!

PR #10396, opened 2024-08-09
Docs: Small wording update

tordans:
Minor changes to the wording, IMO. Mainly punctuation, spacings. And "the first thing" to "anything" which ChatGPT liked more.


k-yle (approved):
👍 I agree that the new version makes more sense

PR #10397, opened 2024-08-09
Docs: Remove reference to mobile

tordans:
The readme has very prominent wording that makes it sound like iD is not usable one mobile at all. However, Quincy did a lot of work in this area a while back and IMO iD works well enough to look at data and make minor changes.

I suggest we remove this reference now, because it is not true anymore. At least not to this extent.


tyrasd (commented):
I've added Opera back to the list of supported browsers. It's still [reasonably popular](https://gs.statcounter.com/).

PR #10399, opened 2024-08-09
Docs: Remove gif and debugbrowser reference

tordans:
- Gifs: Github now supports videos, which are the better way to share animations.
- `debugbrowser`: This side is not what is used to be anymore (open in private browsing…)


k-yle (approved):
I think this deserves the https://github.com/openstreetmap/iD/labels/priority label since the URL now leads to some sort of gambling scam

PR #10406, opened 2024-08-12
fix eslint not working in IDEs

### LINKED ISSUE #10225: ESLint 9 invalid options
It seems that ESLint 9 has deprecated/changed some configuration stuff. My IDE (JetBrains Rider) reports:

````
Error: Invalid Options:
- Unknown options: reportUnusedDisableDirectives, resolvePluginsRelativeTo, rulePaths, useEslintrc, extensions, ignorePath
- 'extensions' has been removed.
- 'resolvePluginsRelativeTo' has been removed.
- 'ignorePath' has been removed.
- 'rulePaths' has been removed. Please define your rules using plugins.
- 'reportUnusedDisableDirectives' has been removed. Please use the 'overrideConfig.linterOptions.reportUnusedDisableDirectives' option instead.
````

These come from `processOptions (node_modules\eslint\lib\eslint\eslint-helpers.js:839:15)`

I am not familiar enough with any of this lint stuff to say more. These errors are not present on an earlier v8.57.0. So I am assuming some config files need to be updated? It's not saying where it's reading stuff from, but the IDE says it looks for `.eslintrc.*` files. These also seem to be more like warnings than errors because they don't stop the project being built/tested. I am not 100% sure this isn't just my IDE doing something weird.

### PR DESCRIPTION
**Pull request by k-yle**
Closes #10225

I just rebased onto the lastest `develop` branch, and it seems like eslint has been broken since 211b6dde (when running in an IDE), because the new flat config has to be at the root of the repo, not in a subfolder[^1]

This PR fixes the bug. Test plan:
 - check that `npm run lint` passes in the CI
 - check that eslint warnings now appear in our IDE

[^1]: https://eslint.org/docs/latest/use/configure/configuration-files

### CODE REVIEW

File: eslint.config.mjs [hunk 1]
```diff
@@ -10,7 +10,6 @@ export default [
       sourceType: 'module',
       globals: {
         ...globals.browser,
-        ...globals.es6
```
k-yle, line 13: I deleted this because `globals.es6` does not exist, it's `undefined`. es6 globals are already handled by this line:

https://github.com/openstreetmap/iD/blob/ba8d52e3c9ce7bf674693ef5144f7cf4a3c9bff1/config/eslint.config.mjs#L9

tyrasd (approved):

PR #10407, opened 2024-08-12
Support importing osmChange files

### LINKED ISSUE #7109: Feature request: "Open osmChange" / "Import CS"
Currently there is an option to save a changeset as an osmChange-file, e.g. before uploading the the osm-db, which is a kind of an CS-"export".

It would be also nice to have the ability the other way around to open an osmChange-file, for loading cs-data into iD, which means for prepared to upload the data to the osm-db as next step, the whole procedure could maybe also called as a CS-"import".

### LINKED ISSUE #8036: Save changes locks up, how to import changes.osc
I was editing ok this morning in my lightweight browser (Seamonkey), so I think the servers are up. But it would not show the trace I uploaded, so I switched to Firefox (68.12.0esr) to edit ways with that trace as reference through the tree cover.
Now when I press save and Upload, it freezes. 
The browser tells me the script is slow or stuck. I press Wait... 
Half an hour later I try Stop. Now the gif spinner moves again, but still will not upload or finish.

I did save the changes.osc but how do I import that into my working browser? 
When I log in there, it does not show I have unsaved changes, so I guess that change set is linked to my browser and not my account?

k-yle:
Closes #7109, Closes #8036

This PR adds a new button which allows you to import an [osmChange](https://osm.wiki/osmChange) file into the editor. [Preview it here](https://kyle.kiwi/iD).

~~This PR is cherrypicked from my fork of iD, so it also supports importing a [geojson](https://geojson.org)/[osmPatch](https://github.com/osm-nz/linz-address-import/blob/main/SPEC.md) file. Presumably these features would be removed if this PR were to be accepted. But I wanted to make a PR so there's a place for discussion.~~

Based on my use cases, the main benefits are:
 - you can backup and restore your changes from iD
 - you can start an edit on your phone using OsmAnd / EveryDoor, and continue it in iD
 - you can use a tool like osm-revert and then tidy up the revert in iD before uploading it[^1]
 - ~~the feature to import a [geojson](https://geojson.org)/[osmPatch](https://github.com/osm-nz/linz-address-import/blob/main/SPEC.md) file is what I use for many import projects~~
 - _(niche use case)_ during long editing sessions, if iD gets too slow, you can export your changes, reload the page, discard your edits, and re-import the osmChange file. 

[^1]: this is still quite buggy


dabreegster:
Super awesome timing, I was just looking for something like this. I've got a few tools in progress that help find/fix tagging problems. The focus of those tools doesn't include checking tags of nearby objects, having a list of edited objects, uploading changesets, etc, and I don't want to reinvent parts of iD. So this feature will be perfect, thanks for implementing it!

I tried out your demo with an example (**sample edit, don't actually upload it**):

```
<osmChange version="0.6" generator="osm2streets">
<create/>
<modify>
<way id="436794680" version="12">
  <nd ref="4346605157"/>
  <nd ref="2860653272"/>
  <nd ref="5707494413"/>
  <nd ref="2860653268"/>
  <nd ref="2941419917"/>
  <tag k="cycleway:right" v="lane"/>
  <tag k="cycleway:right:lane" v="exclusive"/>
  <tag k="highway" v="secondary"/>
  <tag k="lanes" v="3"/>
  <tag k="lit" v="yes"/>
  <tag k="maxspeed" v="35 mph"/>
  <tag k="maxspeed:type" v="sign"/>
  <tag k="name" v="East University Drive"/>
  <tag k="oneway" v="yes"/>
  <tag k="smoothness" v="good"/>
  <tag k="surface" v="asphalt"/>
  <tag k="turn:lanes" v="||right"/>
</way>
</modify>
</osmChange>
```

but I just got "Failed to import file" at the bottom with no further details in the console or elsewhere. I'd suggest:

1) Opening a modal to display this error; the status at the bottom is tiny and disappeared quickly
2) Surfacing some info on what went wrong

k-yle:
thanks for the feedback @dabreegster :) I've added an error modal, and the errors are now logged to the console 

In your case, it's because of a conflict, the xml has `version="12"` but [v13](https://osm.org/way/436794680) is the latest (published 5 months ago)

tordans:
I think adding this is a great enhancement for advanced use cases! 

The only reason I can see not to add it is if it would create too much additional maintenance burden or technical debt. However, we haven't really discussed this aspect for other features… :-).

On the UI side:

> This PR adds a new button which allows you to import an [osmChange](https://osm.wiki/osmChange) file into the editor. [Preview it here](https://kyle.kiwi/iD).
> 

I am thinking more in the direction of the custom image overlay which has two feautres…
1. a may to toggle the layer on, which think is useful
2. a link to a modal that explains the feature

Actually, isn't that more or less the same as …?
> image
(Sorry, German UI)

> * the feature to import a [geojson](https://geojson.org)/[osmPatch](https://github.com/osm-nz/linz-address-import/blob/main/SPEC.md) file is what I use for many import projects

I really like this use case, but I am not sure if we want to open up this area of features for iD. In my mind, all those import use cases are the domain of Rapid which has more feature in this area. There is https://github.com/facebook/Rapid/issues/585 which touches on some use cases that could use this feature. — There is more to talk about here which we will not resolve in this PR IMO. But I think adding a focus on just re-uploading saved OSM changes session would make this PR easier to handle.

tordans:
@dabreegster This is off topic here, but FYI:

> Super awesome timing, I was just looking for something like this. I've got a few tools in progress that help find/fix tagging problems. The focus of those tools doesn't include checking tags of nearby objects, having a list of edited objects, uploading changesets, etc, and I don't want to reinvent parts of iD. So this feature will be perfect, thanks for implementing it!

The fix-and-improve-tagging workflow that I am working on in for https://radverkehrsatlas.de/ is to have some place (our backend) create the issues as a geojson endpoint. Then setup a MapRoulette Challenge which gives room for explanation and guidance. And then – for some types of challenges – add "fix tag" attributes that allow a one click modification in MapRoulette. This workflow has become more relevant IMO since Rapid now has a MapRoulette integration. And they might add the "fix tag" challenges as well https://github.com/facebook/Rapid/issues/1340 which would make updating OSM data super easy right inside the editor. — The advantage of adding a raw file like here is IMO, that the process is a lot more guided and allows for more mapper control. Which is always important to prevent accidental miss edits and negative community feedback.

k-yle:
> >the feature to import a [geojson](https://geojson.org/)/[osmPatch](https://github.com/osm-nz/linz-address-import/blob/main/SPEC.md?rgh-link-date=2024-08-14T05%3A18%3A09Z) file is what I use for many import projects
>
> I really like this use case, but I am not sure if we want to open up this area of features for iD

I agree, I don't think this part should be added to iD itself. It would be too easy for beginners to make an enourmous mess. https://github.com/facebook/Rapid/issues/585 would be the best solution, but the RapiD team were quite hestitant when I asked about it 4 years ago (hence why I built a different [fork of RapiD for tag fixes](https://osm-nz.github.io/RapiD), but now we're getting off topic 😆)

---

> On the UI side:

Your suggestion is possible, althought there would be no checkbox, because you can't "turn off" the data once it's imported. So it would look something like this:
image

A modal with more info is also a good idea. How detailed do you think the explaination should be? The only similar example is [this popup](https://github.com/user-attachments/assets/5cd6f67f-5afc-4b29-a6fa-d956602d5489), which has very little info and somewhat confusing jargon

bhousel:
> [facebook/Rapid#585](https://github.com/facebook/Rapid/issues/585) would be the best solution, but the RapiD team were quite hestitant when I asked about it 4 years ago

That issue is 2 years old, and I don't remember being "hesitant" about it at all.  
It's something we want too, but there isn't really a consensus on how to turn custom data into OSM tags yet.
I agree it would be great to add support to Rapid to ingest osmchange files.

k-yle:
@bhousel that's great to hear, I think I asked on the OSM US slack back in 2020 and was told that there were no plans to support custom data using the new esri feature. Shall we the continue the import-related discussion in facebook/rapid#585 ?

Nekzuris:
I tested this today and it seems to work fine when there’s no conflict.

On the UI side, I was expecting drag and drop to work, the file explorer picker workflow is slow and feels outdated now.
Drag and drop is already supported for GPX and GeoJSON files, I don’t see why we still need a button for this one.

ikkemaniac:
tested this over the past few weeks and seems to be working fine!

vshashenko:
Any chance this pr will get finalized and merged? What's stopping it?

k-yle:
we first need @tyrasd to confirm if this idea would be accepted.

and it would be good to get some more opinions on [where to put the 'import' button](https://github.com/openstreetmap/iD/pull/10407#issuecomment-2287875278). Currently it's here, but it feels a bit out of place:

vshashenko:
I think this feature of importing osmChange files is very useful and should be finalized. It makes interoperability with other tools possible and thus enables the workflow where you can get data from somewhere else in whatever format, convert them to osmChange (via other tools), load in iD, check how it aligns with existing data and save to OSM.

Importing osmChange files alone is already useful enough, so It doesn't need additional convenience features (geojson/osmPatch) right away.

As for the location of button, it would be logical to place it right next to the "Download osmChange file" button, but that requires some changes to the behavior of the "Upload to OpenStreetMap" panel. Right now it appears only when you press the "Save" button (enabled only if there are modifications). The required changes would be to rename the "Save" button to smth like "View changeset", make it enabled even with zero changes, it opens the same panel where you can have the new "Upload osmChange file" button.

On the other hand, current location of the button is fine as well. It least it should not be a block for adding this feature at all. It can always be changed later if needed. The priority is to get the feature released.

tyrasd:
> supports importing a geojson

We should limit it to only support _osmChange_ files for now. GeoJSON files could be confusing as they are already supported as background data layers, and also it could look like we're inviting data imports if geojson data can be directly added in this way.

From a UI perspective: as this is a rather rarely needed feature, the button could be potentially moved into a "hamburger" menu of the _data layers_ section.

modules/actions/import_osmPatch.js
```diff
@@ -0,0 +1,268 @@
...
+  }
+  return newMembersList;
+}
+
+/**
+ * @param {import('geojson').Geometry} geom
+ * @param {Osm.Tags} tags
+ * @param {Osm.OsmRelation['members']} relationMembers
+ */
+function geojsonToOsmGeometry(geom, tags, relationMembers) {
```
matkoniecz, line 75: this should be removed, right?

matkoniecz (changes requested):

modules/actions/import_osmPatch.js
matkoniecz: and other geojson code

PR #10409, opened 2024-08-15
configure vscode i18n extension

k-yle:
Added a few configuration files so that vscode can show you a preview of localised text within the code (where possible)

image


tyrasd (approved):
This is very cool :heart_eyes:  Thanks for the suggestion!

Issue #7123, opened 2019-12-15
Country should be a combo field


1ec5:
The `country` tag is supposed to be an ISO&nbsp;3166 alpha-2 code, but not everyone is familiar with the ISO code for their own country (dependent territory, etc.), let alone the country of a consulate or embassy. Meanwhile, tags like [`country=Turkey`](https://taginfo.openstreetmap.org/tags/country=Turkey) and [`country=United_States_of_America`](https://taginfo.openstreetmap.org/tags/country=United_States_of_America) crop up because it looks a lot like a free-form text field.

Mappers would be more likely to enter consistent values if we turn the Country field into a combo field with each choice being a spelled-out country name. As with the multilingual name field, CLDR would be a natural source for country names localized into every supported language, to avoid additional burden on translators.

Maybe in the future, this work could also come in handy for other tagging schemes that incorporate ISO&nbsp;3166 codes, like `network` and `cycle_network` tags.

lukf:
I’ve spent a lot of time cross-referencing ISO 3166 tables when tagging diplomatic offices. This would be a welcome improvement!

lukf:
Please note that both `country=*` and `target=*` may be a semicolon-delimited list, not just a single country code. [See wiki](https://wiki.openstreetmap.org/wiki/Tag:office%3Ddiplomatic)

1ec5:
[This MapRoulette challenge](https://maproulette.org/browse/challenges/13851) includes hundreds of flagpoles to retag where a mapper had written a full country name in the Country field. That could’ve been due to a lack of awareness of the convention to use ISO codes, the inconvenience of looking up an ISO code, or the taginfo-based suggestions that currently show “Tanzania” as the most common value for some reason.

1ec5:
openstreetmap/id-tagging-schema#799 was closed in favor of the CLDR-based approach described above. We already use CLDR for the list of languages in the Multilingual Name field, but at build time, we throw away the rest of the CLDR data, including country names:

https://github.com/openstreetmap/iD/blob/a38d7a1b8d50a3e358f4e0f0a5ae5a42b3d7cf58/scripts/language_names.js#L69

1ec5:
[...]

PR #10432, opened 2024-08-24
Update localized.js

harshendram:
Country should be a combo field #7123 has been solved. As mentioned in the issue " Since there are so many potential options to list for this field, we would want to generate them dynamically at runtime instead of hard-coding them in id-tagging-schema, but maybe not directly in the UI code, because other modules like the validator might eventually need the same data." This has been solved increasing the performance too.

### DIFF
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -521,16 +521,52 @@ export function uiFieldLocalized(field, context) {
         loadCountryCode();
         return localized;
     };
+async function fetchCountryData() {
+    const response = await fetch('https://restcountries.com/v3.1/all');
+    const countries = await response.json();
+    const countryToISO = {};
 
-    function loadCountryCode() {
-        var extent = combinedEntityExtent();
-        var countryCode = extent && countryCoder.iso1A2Code(extent.center());
-        _countryCode = countryCode && countryCode.toLowerCase();
+    countries.forEach(country => {
+        const countryName = country.name.common.toLowerCase();
+        const isoCode = country.cca2.toLowerCase();
+        countryToISO[countryName] = isoCode;
+    });
+
+    return countryToISO;
+}
+
+let countryToISO = {};
+
+async function getISOCode(input) {
+    if (Object.keys(countryToISO).length === 0) {
+        countryToISO = await fetchCountryData();
     }
 
-    function combinedEntityExtent() {
-        return _entityIDs && _entityIDs.length && utilTotalExtent(_entityIDs, context.graph());
+    const lowerInput = input.toLowerCase();
+
+    if (Object.values(countryToISO).includes(lowerInput)) {
+        return lowerInput;
     }
 
-    return utilRebind(localized, dispatch, 'on');
+    const isoCode = countryToISO[lowerInput];
+    if (isoCode) {
+        return isoCode;
+    } else {
+        return "ISO code or country name does not exist.";
+    }
+}
+
+async function loadCountryCode() {
+    var extent = combinedEntityExtent();
+    var countryCode = extent && countryCoder.iso1A2Code(extent.center());
+    _countryCode = await getISOCode(countryCode) || (countryCode && countryCode.toLowerCase());
+}
+
+function combinedEntityExtent() {
+    return _entityIDs && _entityIDs.length && utilTotalExtent(_entityIDs, context.graph());
+}
+
+return utilRebind(localized, dispatch, 'on');
+    
+  
 }


modules/ui/fields/localized.js
```diff
@@ -521,16 +521,52 @@ export function uiFieldLocalized(field, context) {
         loadCountryCode();
         return localized;
     };
+async function fetchCountryData() {
+    const response = await fetch('https://restcountries.com/v3.1/all');
```
1ec5, line 525: Loading a third-party resource on demand raises privacy concerns. Anyways, the list of countries doesn’t change often enough to require fetching a new list every time the user uses this field. In the worst case, we could hard-code the list of valid country codes, but we don’t have to do that either: we already depend on [country-coder](https://github.com/rapideditor/country-coder/); the code you replaced refers to this package in order to get the ISO code of the country that the user is currently looking at. You can use this same package to enumerate all the valid country codes.

modules/ui/fields/localized.js
1ec5, line 1: This file and the `localized` field type have nothing to do with the Country field. Rather, it’s for the Name field, which lets you choose from a list of languages, not countries. The reason it deals with country codes at all is that some language codes include country codes to indicate a dialect, for example, `zh-CN` for Chinese in Hong Kong.

Contrary to the title of #7123, the Country field is [already of type `combo`](https://github.com/openstreetmap/id-tagging-schema/blob/e25f91e8c0590024cfa6e23fe2a425594f0985b5/data/fields/country.json#L3), but the issue is that it doesn’t provide any suggestions other than what comes from taginfo by default. So we need to special-case the combo field type to do something different for the key `country`. The place to do that is in modules/ui/fields/combo.js, which corresponds to that field type.

The point I made before is that, if we need to provide a similar list of countries in multiple fields, somewhere in modules/presets/index.js might be a good place to centrally calculate that list so it can be reused. But let’s start with just combo.js and the Country field specifically for now.

modules/ui/fields/localized.js
```diff
@@ -521,16 +521,52 @@ export function uiFieldLocalized(field, context) {
...
+        return isoCode;
+    } else {
+        return "ISO code or country name does not exist.";
     }
+}
+
+async function loadCountryCode() {
+    var extent = combinedEntityExtent();
+    var countryCode = extent && countryCoder.iso1A2Code(extent.center());
+    _countryCode = await getISOCode(countryCode) || (countryCode && countryCode.toLowerCase());
```
1ec5, line 562: As a reminder, the task is to display a list of countries by name and apply a country code in response to the user selecting a country name from the list. [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) is a useful API for converting country codes to names in the user’s language (only for display purposes).

harshendram: import { countryCoder } from '@rapideditor/country-coder';
// ... (other existing imports)

export function uiFieldCombo(field, context) {
    // ... (existing code)

    let _countryList;

    function getCountryList() {
        if (!_countryList) {
            const allCountries = countryCoder.countryList;
            _countryList = allCountries.map(country => ({
                value: country.code.toLowerCase(),
                label: new Intl.DisplayNames([context.locale()], { type: 'region' }).of(country.code)
            }));
            _countryList.sort((a, b) => a.label.localeCompare(b.label, context.locale()));
        }
        return _countryList;
    }

    function setStaticValues(callback, filter) {
        if (field.key === 'country') {
            _comboData = getCountryList();
        } else {
            _comboData = getOptions();
        }

        if (filter) {
            _comboData = _comboData.filter(filter);
        }

        _comboData = objectDifference(_comboData, _multiData);
        _combobox.data(_comboData);
        if (callback) callback(_comboData);
    }

    function combo(selection) {
        // ... (existing code)

        if (field.key === 'country') {
            _combobox
                .fetcher(function(value, callback) {
                    const filtered = getCountryList().filter(d => 
                        d.label.toLowerCase().includes(value.toLowerCase()) || 
                        d.value.includes(value.toLowerCase())
                    );
                    callback(filtered);
                });
        } else if (_showTagInfoSuggestions && services.taginfo) {
            selection.call(_combobox.fetcher(setTaginfoValues), attachTo);
            setTaginfoValues('', setPlaceholder);
        } else {
            selection.call(_combobox, attachTo);
            setStaticValues(setPlaceholder);
        }

        // ... (rest of existing code)
    }

    // ... (rest of existing code)
}

PR #10442, opened 2024-08-28
Render parking space partial fill thinner

### LINKED ISSUE #10320: Decrease fill width of parking spaces
### Description

Decrease fill width of parking spaces to resemble building part width.
It is unlikely having same fill width as building parts would make mapping them harder.

Similar to building parts, the parking (like buildings), should still be selectable if parking spaces (partts) are present.

### Screenshots

_No response_

### LINKED ISSUE #10439: Too difficut to select parent parking of parking space, add drop-down menu to select between all overlapping features nearby
### Description

I'm aware of #2225 and #8264, but I was stuck for a long time trying to select the main parking area [here](https://www.openstreetmap.org/edit?way=849745943#map=21/53.2146861/5.8025116) because it's covered with parking spaces and the `Ctrl` + `↑` shortcut doesn’t work when the parking space is selected. You need to select the right point first, only then the shortcut goes to the parking area.
Also, the shortcut is hard to find and not intuitive.

I would suggest adding a drop-down menu to select between all overlapping features nearby. This menu could be opened with a right-click, middle-click, or simply by hovering the mouse over the area for 1 second (like in Autodesk Inventor).

### Screenshots

_No response_

Nekzuris:
Similarly to dcd9876, I reduced the partial fill to make it possible to select parking covered with parking spaces.

Closes #10320, closes #10439

### DIFF
--- a/css/70_fills.css
+++ b/css/70_fills.css
@@ -46,6 +46,9 @@
 .fill-partial path.area.fill.tag-indoor {
     stroke-width: 20px;
 }
+.fill-partial path.area.fill.tag-amenity-parking_space {
+    stroke-width: 30px;
+}
 .ideditor.mode-browse .fill-partial path.area.fill,
 .ideditor.mode-select .fill-partial path.area.fill,
 .ideditor.mode-select-data .fill-partial path.area.fill,


tyrasd:
Hi. Thanks for this suggestion. It seems fine at first glance, but could you please post one or two before/after screenshots for this change?

Nekzuris:
parking

PR #10452, opened 2024-09-03
migrate unit tests from karma to vitest

### LINKED ISSUE #10408: Speed up unit tests
There are a lot of annoying aspects to our current unit test system:
 - Currently we run all tests in a headless chrome browser. This is painfully slow, and also redundant for the majority of tests cases which don't need a real browser environment
 - We use karma (see #8764), but it's deprecated and has performance problems
 - We run our tests using the compiled JS bundle (`dist/iD.js`), instead of importing the source code itself. This means it takes forever to test your changes. After every change, you have to run `npm run build:js`, restart `npm run test:spec` (which restarts chrome), then reattach the debugger, and then wait for ages for the results. 
 - There is no autocomplete for the chai assertions (see #9805)

I think there are some simple changes that would make it much easier to write tests.

Based on a quick test, 1733 of our 2354 tests (73%) pass when running with [vitest](https://github.com/vitest-dev/vitest)+[jsdom](https://github.com/jsdom/jsdom), with barely any changes required. And it's about 10-20x faster 🎉

Therefore, I propose we introduce a lightweight test framework, and rename each test file based on which test framework it uses:

 - 73% of tests would keep the file extension `*.test.js`, an run with a lightweight framework, e.g. vitest or jest
 - 27% of test files could be renamed to `*.e2e.js`, and continue to use karma + a headless browser.

This would be a major improvement for the majority of tests which are not UI-related. There would be no improvement for the tests related to UI/rendering code.

k-yle:
>  ~~this is blocked by #10431, and needs to be rebased after that PR is merged~~

---

Closes #10408, Closes #9805, Closes #10382

This PR migrates _all_ our unit tests from karma to vitest. See issue #10408 for more details about why this beneficial. In short: the development cycle is now 10-20x faster, which makes it significantly easier and more pleasant to write tests.

This was suprisingly easy, and only a few test files had to be modified - these changes have been split into a separate commit.


modules/index.js
```diff
@@ -23,6 +23,9 @@ export * from './validations/index';
 // This is only done in testing because of the performance penalty.
 export let debug = false;
 
+/** @param {boolean} newValue */
+export const setDebug = (newValue) => { debug = newValue; };
```
k-yle, line 27: this is required because mutable export bindings (`export let ...`) have a lot of caveats in ESM, and don't work well with our setup

modules/services/vegbilder.js
```diff
@@ -34,18 +34,9 @@ async function fetchAvailableLayers() {
...
-  const l = response.evaluate(
-    xPathSelector,
-    response,
-    NSResolver,
-    XPathResult.ANY_TYPE
-    );
-  let node;
   const availableLayers = [];
-  while ( (node = l.iterateNext()) !== null ) {
+  for (const node of response.querySelectorAll('FeatureType > Name')) {
```
k-yle, line 39: jsdom doesn't support [`document.evaluate`](https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate) perfectly, and tbh I think this is more readable than 10 lines of XPath wizardry

test/spec/actions/delete_way.js
```diff
@@ -8,7 +8,7 @@ describe('iD.actionDeleteWay', function() {
 
     it('removes a way from parent relations', function() {
         var way      = iD.osmWay(),
-            relation = iD.osmRelation({members: [{ id: way.id }, { id: 'w-2' }]}),
+            relation = iD.osmRelation({members: [{ id: way.id }, { id: 'w-99' }]}),
```
k-yle, line 11: this test sporadically fails if `way.id` is also allocated ID #&shy;2. Changing it to `w-99` prevents this edge case.

test/spec/services/maprules.js
```diff
@@ -163,7 +163,17 @@ describe('maprules', function() {
...
         it('clears _validationRules array', function() {
+            iD.serviceMapRules.clearRules();
+            expect(iD.serviceMapRules.validationRules()).to.be.empty;
+
+            iD.serviceMapRules.addRule({
+                geometry:'node',
+                equals: {amenity:'marketplace'},
+                absence:'name',
+                warning:'\'Marketplace\' preset must be coupled with name'
+            });
```
k-yle, line 174: previously, this test case only passed because state from the previous test was leaking into this one. Now it explicitly sets up the environment.

k-yle (commented):
I've left some code review comments explaining why certain test cases had to be changed

tyrasd:
#10431 is merged now. There are no merge conflicts, but can you please double-check that everything works as expected?

k-yle:
thanks @tyrasd :) it looks like github automatically changed the target branch which is a neat feature. I've rebased it anyway so that the history is cleaner, and I can confirm that it still works

tyrasd (approved):
Perfect. :100: 

Only one comment: I noticed that on my system, a full run of the new test suite is significantly slower then previously:

>   Duration  158.62s (transform 3.53s, setup 204.50s, collect 5.84s, tests 9.83s, environment 63.09s, prepare 13.09s)

For comparison, the old code runs `npm test:spec` in approximately a minute here.

Also the CI on github seems to be a bit slower now: [new](https://github.com/openstreetmap/iD/actions/runs/10682107510/job/29607179917): `1m 51s`, [old](https://github.com/openstreetmap/iD/actions/runs/10681911117/job/29606554447) `1m 8s`. :thinking: 

Is that expected because of jsdom overhead?

modules/globals.d.ts
```diff
@@ -0,0 +1,12 @@
+import type { FetchMockStatic } from 'fetch-mock';
+
+declare global {
+  declare var iD: typeof import('.');
+  declare var d3: import('d3');
```
tordans:
```suggestion
  declare var d3: typeof import('d3');
```

Probably like this?

test/spec/modes/add_note.js
```diff
@@ -1,4 +1,4 @@
-describe('iD.modeAddNote', function() {
+describe.skip('iD.modeAddNote', function() {
```
tordans, line 1: Just checking, is this by design? It does skip the whole block, right? (https://vitest.dev/api/#describe-skip)

tyrasd: The actual test code in this file is all commented out (see line 28 and following), so skipping the whole thing saves a little bit of setup time. I haven't looked into why the tests are commented out for now, but that would be something for a follow-up PR anyway.

modules/globals.d.ts
k-yle: done, and I just resolved the git conflicts

k-yle:
> Only one comment: I noticed that on my system, a full run of the new test suite is significantly slower then previously: [...]

interesting, I hadn't checked the overall execution time since it felt similar, but it is indeed slower :(

For me, the more important metric is the time between saving a file, and seeing the test results. That's now reduced to 0-1 seconds, previously it was 10-20 seconds or longer. (since we don't need to run `npm run build` after every change, and there is a 'watch' mode)

But it's a shame that this is a trade-off... We could try to use the [`--no-isolate` flag](https://vitest.dev/guide/improving-performance), but this will require tweaking a few more tests. In GitHub Actions, [it only takes 75 seconds](https://github.com/k-yle/iD/actions/runs/10682937150/job/29609887626#step:6:336) with `--no-isolate`!

tyrasd:
> For me, the more important metric is the time between saving a file, and seeing the test results. 

Agree. :+1: 

> We could try to use the --no-isolate flag

Didn't the test previously essentially run in something equivalent to that? :wink:

Also, as the iD modules should not interfere with each other via global states anyway, it might be even preferred to run them like this?!

~~I just found out that running it with `--no-isolate --maxConcurrency 1` seems to work. Without limiting it to a single thread, there are odd errors like `ReferenceError: window is not defined`, which I don't immediately see where they might originate from. :shrug: But as it's also a lot faster than the isolated approach, it might be worth considering…~~

//edit: turns out this was unfortunately just a fluke. It now also fails for me even with `--maxConcurrency 1`. Strange.

k-yle:
I figured out what was causing the flaky tests and enabled `--no-isolate` in 156f6a8, which has slightly sped up the process

tyrasd (approved):

tyrasd:
Merged as e839b6307 + 108893c62 (where I activated `--no-isolate` also for the default `npm test` command) and b4e55c202 + e7f56cf7f because I was stupid and accidentally used an outdated branch while merging this PR. Sorry!

PR #10459, opened 2024-09-09
fix unsolvable validator error for presets with locationSet

k-yle:
This was observed in https://github.com/openstreetmap/id-tagging-schema/pull/1335#pullrequestreview-2279573101

`presetManager.matchTags(...)` did not consider the `locationSet` of each preset, so it produced an error which was impossible to fix.

Regardless of whether https://github.com/openstreetmap/id-tagging-schema/pull/1335 is accepted, this PR should be merged to prevent errors in other situations (for example, [this node](https://osm.org/edit?node=7503247785))

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -56,8 +56,9 @@ export function presetIndex() {
   let _loadPromise;
 
 
-  _this.ensureLoaded = () => {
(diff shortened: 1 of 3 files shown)


tyrasd (approved):
Thanks!

PR #10478, opened 2024-09-24
Include route number in label of named route relations

### LINKED ISSUE #8559: Show ref in the name of route relations

1ec5:
When labeling a route relation that has a `name=*`, include its `network=*` and route number (`ref=*`) for disambiguation. This is identical to 4856eef32b4b04ef40a86b64c98bb0e23cbb3305 by @k-yle. The only additional change is that we continue to show the `name=*` verbatim, without appending other details, if it looks like a [PTv2-formatted description](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2741977#cite_ref-name-description_1-0) of the route. This should address the criticisms in #8707.

Fixes #8559.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -763,6 +763,8 @@ en:
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
     display_name:
+      network_ref_name: "{network} {ref}: {name}"
+      ref_name: "{ref}: {name}"
       direction: "{direction}"
       network: "{network}"
       from_to: "from {from} to {to}"
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -763,8 +763,6 @@ en:
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
     display_name:
-      network_ref_name: "{network} {ref}: {name}"
-      ref_name: "{ref}: {name}"
       direction: "{direction}"
       network: "{network}"
       from_to: "from {from} to {to}"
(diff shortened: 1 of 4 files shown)


data/core.yaml
```diff
@@ -763,6 +763,8 @@ en:
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
     display_name:
+      network_ref_name: "{network} {ref}: {name}"
+      ref_name: "{ref}: {name}"
```
1ec5: I sometimes encounter situations where it would be nice to also append `from=*`, `to=*`, and `via=*` for disambiguation – just like when the route has a `ref=*` instead of a `name=*`. Of course, the name might push all these extra details off out of view, but at least they’d appear in a tooltip as a last resort.

k-yle: 👍 to adding a variant of `ref_from_to_via` which also includes the name at the end

k-yle (approved):
👍 I think this is a nice solution

data/core.yaml
1ec5: I’ve modified this PR to replace the short-circuit on `name` with something similar to what we do with `ref`. Now the label can combine the ref, name, and other details to avoid ambiguity. It only does this for routes; other things with names and refs, like indoor rooms, keep the same labels as before. Unfortunately, this change results in a combinatorial explosion of localizable strings, but hopefully they’re straightforward enough for translators to figure out.

k-yle (approved):

PR #10488, opened 2024-10-09
Update language data for Sinitic languages

winstonsung:
Bug: #10488
Bug: https://github.com/openstreetmap/id-tagging-schema/pull/1356
Change-Id: I0ca3cd8b2a82a7edeff70b4be7678e0699214ccc


1ec5 (commented):
This looks like a good set of changes to me at a glance. Have you verified that the associated `name:*` tags are well-established, so that we don’t end up promoting tags that the language community is unaware of? Also, we may need to update our copy of [wmf-sitematrix](https://github.com/osmlab/wmf-sitematrix/) so that the Wikipedia field’s language selector and Wikidata field’s label functionality reflects recent changes to Wikimedia’s language list.

winstonsung:
> Have you verified that the associated `name:*` tags are well-established, so that we don’t end up promoting tags that the language community is unaware of?

The language tags were well-established and being used in MediaWiki and on Wikimedia wikis.

- The variant subtag `pehoeji`, `tailo` were supported by the Hokkien community, proposed and added to *Language Subtag Registry* on 2024-03-04.
- The variant subtag `pinyin` was already existed in *Language Subtag Registry* since 2008-10-14, but not using the standard BCP 47 language tag on OpenStreetMap (so we need to fix it now).

<https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry>

> Also, we may need to update our copy of [wmf-sitematrix](https://github.com/osmlab/wmf-sitematrix/) so that the Wikipedia field’s language selector and Wikidata field’s label functionality reflects recent changes to Wikimedia’s language list.

I'll take a look into it.

winstonsung:
> Also, we may need to update our copy of [wmf-sitematrix](https://github.com/osmlab/wmf-sitematrix/) so that the Wikipedia field’s language selector and Wikidata field’s label functionality reflects recent changes to Wikimedia’s language list.

The list is up-to-date as far as I can see.

1ec5 (approved):
> The language tags were well-established and being used in MediaWiki and on Wikimedia wikis.

I was referring to the OSM database. In theory, the `name:*` tagging scheme relies on BCP&nbsp;47, but there have been exceptions in the past, not necessarily the same exceptions as in Wikimedia’s wiki farm. I see that `name:zh_pinyin=*` has already been mass-replaced with [`name:zh-Latn-pinyin=*`](https://wiki.openstreetmap.org/wiki/Key:name:zh-Latn-pinyin), and `name:zh-Latn-pinyin=*` is now documented. So that’s good – the momentum is already on the side of the correct code, and iD is just responding to that change. openstreetmap/id-tagging-schema#1356 is another good step.

> The list is up-to-date as far as I can see.

It looks like iD is currently using wmf-sitematrix v0.1._x_, which means it’ll pull in [v0.1.5](https://github.com/osmlab/wmf-sitematrix/releases/tag/v0.1.5). We’re overdue for a new release of that package, which will probably be versioned as v0.2.0 due to build system changes, so a small tweak to this code will be necessary once that happens:

https://github.com/openstreetmap/iD/blob/72643ba10bed96d785a8b33b767b38bfe4eb9304/modules/core/file_fetcher.js#L42

It looks like iD is no longer using wmf-sitematrix for anything that also uses the built-in language data, so we can take care of that change in a separate PR.

1ec5:
> It looks like iD is no longer using wmf-sitematrix for anything that also uses the built-in language data, so we can take care of that change in a separate PR.

#10489

PR #10489, opened 2024-10-09
wmf-sitematrix v0.2.0

1ec5:
Upgraded to [wmf-sitematrix v0.2.0](https://github.com/osmlab/wmf-sitematrix/releases/tag/v0.2.0), which adds 34 new languages to the Wikipedia field. Also updated the changelog to mention the changes in #10488.


tyrasd (approved):
Thanks!

PR #10494, opened 2024-10-11
fixed the overflowing combobox

### LINKED ISSUE #10482: The combobox options are truncated when displayed near the edge of the viewport
### URL

https://www.openstreetmap.org/edit#map=19/42.361060/-71.061190

Steps to replicate the issue:
1. Locate a feature with tags or relations containing fields that have options (example: https://www.openstreetmap.org/edit#map=19/42.361060/-71.061190)
2. Adjust the viewport if necessary so that a field with options is positioned at the bottom of the viewport
3. Select the field

Selecting a combobox near the bottom of the viewport can cause the field's options to be truncated if there is not enough space relative to the options list height. While users can use up and down keys to select from the available options in these cases, clicking truncated options is not possible. This can lead to unnecessary scrolling and/or collapsing of sections to accommodate option list height to make selections with the pointer.

If there is not sufficient space beneath a field to display all options, the options list could be displayed above the field.

The issue is observed in Firefox version 130.0.1 and Chrome version 129.0.6668.60 on Windows 11.

Example of truncated tag options:
OpenStreetMap_ComboboxOptionsTruncated

2.30.2

Chrome, Firefox

Prasanth-S7:
Fixed: #10482
the current comboBox box css styles makes the content overflow out of the screen which creates a bad user experience .
Solved this by tweaking the marginTop of the comboBox
Previous 
image
Modified
image

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2370,7 +2370,7 @@ div.combobox {
     z-index: 9999;
     display: none;
     box-shadow: 0 4px 10px 1px rgba(0,0,0,.2);
-    margin-top: -1px;
+    margin-top: -240px;
     background: #fff;
     max-height: 245px;
     overflow-y: auto;


tyrasd (changes requested):
Hi, thanks for the effort. However, this fix is just moving the problem to a different situation: Now, the combobox will be cut off when the content would have to be displayed above the scroll:

screenshot showing how the PR would behave in a case where the combobox need to be rendered towards the top of the content


A proper fix would need to dynamically choose to display the dropdown above the input element only if the content would not fit on the remaining space on the page . See `tooltip.js` for a module which already implements this logic.

tyrasd:
closing as stale

Issue #10465, opened 2024-09-19
Allow multiselection of search results


jidanni:
### Description

Here the user wishes to multiselect all these search results.

So perhaps add checkboxes, or allow SHIFT+CLICK or CTRL+CLICK. 

### Screenshots

![Screenshot 2024-09-20 03 49 45](https://github.com/user-attachments/assets/fa6f537f-59f6-4018-bc1d-2fc34aad1dc1)

draunger:
@jidanni can i do start work on this issue, I am new here and I want to do this issue?

draunger:
@jidanni  I think I can solve this issue

jidanni:
That's great.
I'm sure the team will love your fix.

draunger:
@jidanni you are asking for multiselect for highlighting the map ?? means when hover only one search result some part of the map become highlighted so you want this that on multiselect the result the item which are selected must be highlighted ??

draunger:
@jidanni i have raised a PR https://github.com/openstreetmap/iD/pull/10497 against this issue as described in it , you can review this PR as soon as possible thank you.

1ec5:
> I'm sure the team will love your fix.

@jidanni, please don’t speak for the development team. The maintainers of this repository haven’t weighed in on your feature request, which means it isn’t necessarily something that will be accepted. You’re doing @draunger a disservice by prematurely asserting that it will be.

draunger:
@1ec5 what I do now? Is my PR is now waste of time?

1ec5:
Hi @draunger, thanks for submitting a PR, and sorry that you were misled. Please be patient while the maintainers work through the very large backlog of issues and PRs. It’s possible that they may agree with the feature request, but there’s no guarantee. In the meantime, the issues in https://github.com/openstreetmap/iD/labels/good%20first%20issue have been identified by others as good issues for familiarizing yourself with the iD codebase, hopefully without getting into any controversy.

jidanni:
> @jidanni you are asking for multiselect for highlighting the map ?? means when hover only one search result some part of the map become highlighted so you want this that on multiselect the result the item which are selected must be highlighted ??
> 
> 

All I know is I wanted to change the name of all the items I saw in that search result. I wanted to correct it by adding the same missing character to every single name. 
So multi-select is what I needed. 

Sorry for being too optimistic about team policies. Okay next time I won't make the same mistake.

tyrasd:
[...]

PR #10497, opened 2024-10-13
feat: multi-select checkboxes added

draunger:
# Description
# Issue
https://github.com/openstreetmap/iD/issues/10465

## Before this PR
image

## After this PR
image

### DIFF
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -292,17 +292,45 @@ export function uiFeatureList(context) {
             var items = list.selectAll('.feature-list-item')
                 .data(results, function(d) { return d.id; });
 
+            // var enter = items.enter()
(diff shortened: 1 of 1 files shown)


tyrasd (changes requested):
I don't think simply highlighting the selected search result features is what the intention behind #10465 was. For example, can you explain how this would help with a use case like the following:

> I wanted to change the name of all the items I saw in that search result

draunger:
@tyrasd I think it is beneficial if we want to show someone highlighted multiple things on the map from the search result..

tyrasd:
What I don't understand is how this highlighting would be helpful for a particular mapping task?

draunger:
@tyrasd ok sir I understand, I Mistankly understood this issue wrongly and makes this I am closing this PR

Issue #10495, opened 2024-10-11
Street level imagery: Viewer overlaps right click menu


danieldegroot2:
### How to reproduce the issue?

**Observed behaviour**
1. Enable street level imagery
2. Select an image to view
3. Create an object
4. Move the map manually so the object is above the street level imagery viewer [1]
5. Right click on the object
6. The right click / edit menu and associated hover tooltips are partially obscured by the street level imagery viewer

**Expected behaviour**
The right click / edit menu and tooltips are moved in front of the viewer.
Additionally the orientation of the menu could be changed to horizontal so it does not overlap the viewer.

Related #7368 

---

[1]; can be useful, as it is closer to the user's reading direction as well as being closer to the 'Tag this photo id on the currently selected map feature'.

### Screenshot(s) or anything else?

![image](https://github.com/user-attachments/assets/9118a9d7-fee8-418b-a3c3-b58ecf8b7f44)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.30.4

### Which browsers are you seeing this problem on?

Chrome

FrankOHara43:
hi, can you please specify how to open street level imagery? As a new contributor I am finding it difficult to find that option. Also, I am interested in working on this issue.

1ec5:
Hi @FrankOHara43, thanks for your interest. The screenshot above shows the Mapillary image overlay, which you can enable from the Photo Overlays section of the Map Data pane (accessible from the strip of buttons on the right side of the map). You’ll need to zoom into a location that has Mapillary imagery, such as https://ideditor.netlify.app/#map=20.43/39.28200/-84.28412&photo=mapillary/172552988091232.

If you haven’t used iD before, please go to https://ideditor.netlify.app/ (or go to https://www.openstreetmap.org/, create a free account, zoom in to your neighborhood, and click Edit), and follow the on-screen tutorial.

danieldegroot2:
Ref #10504.

tyrasd:
This can also happen with the info panels on the bottom right part of the screen. I guess the easiest solution would be to render the right click menu over all UI elements (as one would expect from regular OS context menus). https://github.com/openstreetmap/iD/pull/10504#issuecomment-2419906458 has some pointers for how to approach this issue should @FrankOHara43 not like to continue working on this.

FrankOHara43:
i am interested to continue working on this.

haresahani:
[...]

PR #10504, opened 2024-10-17
Ref issue #10495: Viewer overlaps right click menu fixed.

FrankOHara43:
Screenshot 2024-10-17 174659

changed the orientation of the menu to appear horizontally.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -5706,6 +5706,7 @@ li.hide + li.version .badge .tooltip .popover-arrow {
     flex-direction: column;
     background: #fff;
     border-radius: 4px;
+    flex-direction: row;
     /* padding is set in edit_menu.js */
 }
 


css/80_app.css
```diff
@@ -5706,6 +5706,7 @@ li.hide + li.version .badge .tooltip .popover-arrow {
     flex-direction: column;
     background: #fff;
     border-radius: 4px;
+    flex-direction: row;
```
1ec5, line 5709: This would unconditionally change the menu to be horizontal, whereas https://github.com/openstreetmap/iD/issues/10495#issue-2581547423 suggested only doing so when showing the street-level imagery pane. Most likely, this would clash with the tooltip that appears when hovering over each menu item. I think changing the orientation _could_ be an improvement, but it needs more thought. I don’t know yet whether users would be surprised by a menu that changes from horizontal to vertical depending on what’s around it. This reorientation also doesn’t really solve the problem stated by @danieldegroot2, because you could right-click on a location close to the pane and it would still underlap the pane.

A fix for #10495 should focus on bringing the context menu and its tooltip higher in the z order than the street-level imagery pane.

1ec5 (commented):
Hi, thanks for your contribution! Please try to revise the fix to solve the stated problem, even if it differs somewhat from the proposed solution. Thanks!

tyrasd:
As @1ec5 explained above, this simple approach is not going to properly fix #10495. @FrankOHara43: if you're interested in giving a proper try to fix this, here's some pointers to get you started:

* currently, the edit menu (`uiEditMenu`) is attached to the html div `supersurface` (`context.map().supersurface.call(_editMenu)`)
* this is the container of the "map" in iD (background tiles and map features).
* most other map UI elements (street level image viewer, info panels, control panels, etc.) are attached to the `over-map` div
* it might make sense to attach to the `over-map` instead of the map directly, then it can be properly ordered to be displayed on top of the clashing UI elements
* when doing so, there are some corner cases to consider: for example, the `over-map` div does currently not extend to the top of the page, so any absolute positioning will be slightly off when using mouse coordinates relative to the absolutely positioned map div


Let us know, if you're interested in working on this. :blush:  For now however, I'm going to close this PR for now, because it's not really in a state that can be considered to be merged.

PR #10508, opened 2024-10-18
Stop password managers from autofilling address fields

### LINKED ISSUE #10507: Stop password managers from autofilling address fields
### Description

I am a 1Password user (and employee). Whenever I am editing a local business's details in iD, I am frustrated that the 1Password browser extension offers to fill out my personal address instead of the business's address.

I don't work on our browser extension, I could not tell you why every password manager has their own custom data attribute
when `autocomplete=off` exists.

I was very pleased to find out that there was already a well-used utility function
to disable autocomplete/autocorrect/autocapitalize in modules/util/util.js under the `utilNoAuto` function. So, I already have a working commit that would address this for 1Password, Bitwarden, Dashlane, and LastPass. It extends what was already built in #3839 and #5818.

### Screenshots

Screenshot 2024-10-18 at 12 16 32

michaelabon:
I am a 1Password user (and employee). Whenever I am editing a local business's details in iD, I am frustrated that the 1Password browser extension offers to fill out my personal address instead of the business's address.

| **Before** | **After** |
|--------|--------|
| Screenshot 2024-10-18 at 12 16 32  |  Screenshot 2024-10-18 at 12 17 24 | 

I don't work on our browser extension, I could not tell you why every password manager has their own custom data attribute
when `autocomplete=off` exists. But, for fairness, I am including other major password managers' custom data attributes to stop them as well from autofilling address fields:

- [1Password](https://developer.1password.com/docs/web/compatible-website-design/)
- [Bitwarden](https://bitwarden.com/help/releasenotes/#2021-06-29)
- [Dashlane](https://www.dashlane.com/blog/dashlane-integration-forms#h-other-miscellaneous-musings)
- [LastPass](https://support.lastpass.com/s/document-item?language=en_US&bundleId=lastpass&topicId=LastPass/c_lp_prevent_fields_from_being_filled_automatically.html&_LANG=enus)

Fixes #10507

### DIFF
--- a/modules/util/util.js
+++ b/modules/util/util.js
@@ -509,6 +509,10 @@ export function utilNoAuto(selection) {
         .attr('autocomplete', 'new-password')
         .attr('autocorrect', 'off')
         .attr('autocapitalize', 'off')
+        .attr('data-1p-ignore', 'true')  // 1Password
+        .attr('data-bwignore', 'true')   // Bitwarden
+        .attr('data-form-type', 'other') // Dashlane
+        .attr('data-lpignore', 'true')   // LastPass
         .attr('spellcheck', isText ? 'true' : 'false');
 }
 


tyrasd (approved):
Thanks. Looks good to me (I did not check with all of the mentioned password managers, but assume that the attributes are set correctly).

PR #10510, opened 2024-10-22
update temaki to v5.10.0

jake-low:
Bump the `@rapideditor/temaki` dependency version to `^5.10.0`. This version of temaki has a new icon in it that I made, indended for a preset that I'm currently [working on adding to id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/pull/1352). 🙂

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "@mapbox/maki": "^8.0.1",
         "@openstreetmap/id-tagging-schema": "^6.8.1",
         "@rapideditor/mapillary_sprite_source": "^1.8.0",
-        "@rapideditor/temaki": "^5.9.0",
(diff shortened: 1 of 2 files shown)


k-yle (commented):
as far as I understand, this PR should not be necessary, because the lockfile is already using v5.10.0 [as of 1 week ago](https://github.com/openstreetmap/iD/blame/develop/package-lock.json#L834):

https://github.com/openstreetmap/iD/blob/5024feafe83db6b698098d65996cd272e894e60e/package-lock.json#L834

So I think you just need to push a new commit to https://github.com/openstreetmap/id-tagging-schema/pull/1352 to trigger the CI to re-run, and the icon should work

jake-low:
Gotcha, thanks! I'll close this PR then.

Issue #7485, opened 2020-03-28
Don't show value suggestions for full_name, loc_name, reg_name


maro-21:
Do not show value suggestions for keys: `full_name`, `loc_name`, `reg_name`, `short_name`, `sorting_name`, `artist_name`, `nat_name`, `long_name`, `bridge:name` in All tags.

quincylvania:
Sorry, we don't support tag-specific functionality in the raw tag editor.

maro-21:
What do you mean? I don't understand.

quincylvania:
The "All tags" section is for editing plain keys and values. We don't have code like "if this is the `name` tag, do something special". That's what the fields are for.

maro-21:
I meant this: if you add manually `alt_name` and click an arrow - there are no value suggestions - this is ok. The same with for example `name` and `old_name`.
But when a key is one of the above, if you click an arrow, there are random value suggestions from different languages, which is not ok. Because names are unique and it doesn't matter what values were used on the other side of the world.
There should be some code which prevents the editor from listing these values for `alt_name`.

![full_name](https://user-images.githubusercontent.com/57260466/77856575-5f204580-71f8-11ea-9d55-d03e15d5831b.png)

quincylvania:
> I meant this: if you add manually alt_name and click an arrow - there are no value suggestions - this is ok. The same with for example name and old_name.

@maro-21 Yes, iD automatically ignores the keys with the greatest number of unique values, including `name` and `alt_name` but not `full_name`. Though I suppose this behavior is actually enforced at the service level and not the UI level, so we can ignore these keys in order to be less taxing on TagInfo if nothing else.

1ec5:
@joto (taginfo maintainer) has previously recommended that we avoid making this sort of query for common freeform keys like `name`, for which each feature generally has a different value. #3975 disabled value suggestions for the 100 keys with the most distinct values, plus `opening_hours`:

https://github.com/openstreetmap/iD/blob/a7ce956e1f2f12c10608f630bbe4505dc96141f6/modules/services/taginfo.js#L201-L218

(You can see the same list by sorting [this table](https://taginfo.openstreetmap.org/keys) by the “Values” column.)

`alt_name` is 30th on the list, so suggestions are disabled. But the other keys mentioned above are much further down in the list.

[...]

PR #10521, opened 2024-11-01
remove taginfo suggestions for `name:*` tags

### LINKED ISSUE #10287: Don't suggest values from Taginfo for `name:xx` tags
### URL

_No response_

1. Create an element
2. Go to it's tags
3. Add `name:xx`
4. Switch to the value box
5. Start typing
6. (notice the suggestions below)

image

2.29.0

Firefox

draunger:
# Description:
## approach
i have used slice method which slice the key which are starting with name and checks the condition after slicing

## attached screenshot after the fix 
image

## before this fix 
image

 ## attached issue
 https://github.com/openstreetmap/iD/issues/10287

### DIFF
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -215,7 +215,7 @@ export default {
         this.keys(params, function(err, data) {
             if (err) return;
             data.forEach(function(d) {
-                if (d.value === 'opening_hours') return;  //exception
+                if (d.value === 'opening_hours') return;  // exception
                 _popularKeys[d.value] = true;
             });
         });
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -192,8 +192,6 @@ function checkCache(url, params, exactMatch, callback) {
 
 export default {
 
-    
-    
     init: function() {
         _inflight = {};
         _taginfoCache = {};
@@ -226,7 +224,7 @@ export default {
         this.keys(params, function(err, data) {
             if (err) return;
             data.forEach(function(d) {
-                if (d.value === 'opening_hours') return;
+                if (d.value === 'opening_hours') return;  //exception
                 _popularKeys[d.value] = true;
             });
         });
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -11,6 +11,10 @@ import { taginfoApiUrl } from '../../config/id.js';
 var apibase = taginfoApiUrl;
 var _inflight = {};
 var _popularKeys = {};
+// manually exclude some additional keys – #5377, #7485, #10287
+// these will be returned by keys(), but taginfo will not be queried for values() requests
+var _extraExcludedKeys = /^(postal_code|via|((int_|loc_|nat_|official_|old_|ref_|reg_|short_|full_|sorting_|alt_|artist_|long_|bridge:|tunnel:)?name(:left|:right)?(:[a-z]+)?))$/;
+
 var _taginfoCache = {};
 
 var tag_sorts = {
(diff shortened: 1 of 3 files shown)


Dimitar5555:
I've just did a quick test and it seems like you can add `name: true` to this list and it will catch all `name:xx` variants, without all of the extra code you added. It is likely using some sort of regex under the hood, but I haven't looked for it.

> https://github.com/openstreetmap/iD/blob/49492a0e06aef71f90e96ab3ed50098df0f4ce93/modules/services/taginfo.js#L196-L209

draunger:
@Dimitar5555 i did this before but it does not working

draunger:
@1ec5 please review my PR so that i can make changes

modules/services/taginfo.js
```diff
@@ -222,7 +222,7 @@ export default {
         this.keys(params, function(err, data) {
             if (err) return;
             data.forEach(function(d) {
-                if (d.value === 'opening_hours') return;  // exception
+                if (d.value === 'opening_hours') return;  //exception
                 _popularKeys[d.value] = true;
```
1ec5: I was wondering how this service excludes `name`, since it isn’t explicitly listed in `_popularKeys`. It turns out that `name` happens to be one of the 100 most popular keys that we add to the `_popularKeys` here. (There’s some additional explanation at https://github.com/openstreetmap/iD/issues/7485#issuecomment-605681756 that I had forgotten.)

modules/services/taginfo.js
```diff
@@ -289,7 +289,7 @@ export default {
     values: function(params, callback) {
         // Exclude popular keys from values lookups.. see #3955
         var key = params.key;
-        if (key && _popularKeys[key]) {
+        if (key && _popularKeys[key] || key.slice(0,4)==='name') {
```
1ec5: [`String.prototype.startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) is less error-prone than comparing a substring manually. A more thorough fix would catch any key if it’s just a localized version of any of the popular keys. A localized key looks like `baseKey:xy`, where _xy_ is a language code (not necessarily two characters long). I think what you’ve got is OK, but I’d recommend switching to `startsWith` and also requiring a colon, like `name:`, in case someone types in a key that doesn’t have it like [`name_disaster`](https://taginfo.openstreetmap.org/keys/name_disaster#overview) or [`name_base`](https://taginfo.openstreetmap.org/keys/name_base#overview).

1ec5 (commented):
Thanks for your patience. In addition to the following comments, please consider writing a unit test for the change. You can find tests related to this file at test/spec/services/taginfo.js.

modules/services/taginfo.js
draunger: @1ec5 as you said that start with name: will cause to name that starting without colon  shows suggestion,  so do i change with this  key.slice(0,4)==='name'  to this =>  key.slice(0,5)==='name:'  ??

tyrasd (changes requested):
In addition to what @1ec5 already suggested, I think it would be preferred if this would not be a hardcoded case inside an `if` statement deep inside the code, especially given that there is already a list of tags for this. For example, one might want to also add `loc_name:xx`, or `description:xx` to this. One possible solution would be to change `_popularKeys` from a dictionary to a (list of) regular expressions:

```
_popularKeys = [
    /^postal_code$/,
    …
    /^name:../
];

…

if (key && _popularKeys.some(regex => regex.test(key)) {
    …
```

As this should not be a performance bottleneck, the lost efficiency compared to the dictionary lookup should be negligible. What do you think?

modules/services/taginfo.js
draunger: > [`String.prototype.startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) is less error-prone than comparing a substring manually. A more thorough fix would catch any key if it’s just a localized version of any of the popular keys. A localized key looks like `baseKey:xy`, where _xy_ is a language code (not necessarily two characters long). I think what you’ve got is OK, but I’d recommend switching to `startsWith` and also requiring a colon, like `name:`, in case someone types in a key that doesn’t have it like [`name_disaster`](https://taginfo.openstreetmap.org/keys/name_disaster#overview) or [`name_base`](https://taginfo.openstreetmap.org/keys/name_base#overview).

Now sir what I do :
1.can i just make the testcase for it and send PR for it and remaining thing should be the same??

draunger:
hey @tyrasd i have done the commited the changes as you said means used regex ...

draunger:
hey @tyrasd i think you are right so i have done changes as you said means used regex and write the test case regarding this :)

draunger:
@1ec5 can you re review my this PR as I have made some changes as Martin sir suggest and also made a testcase for this issue :)

PR #10523, opened 2024-11-02
fix erroneous `mismatched_geometry` warning for preset-less tags

### LINKED ISSUE #10522: wrong error message?
### URL

https://www.openstreetmap.org/edit#map=21/49.3531020/11.3518268

If I add disused:waterway=security_lock as a line (way), I get the following error message when uploading (de): Fläche sollte eine geschlossene Fläche sein mit den Eigenschaft "disused:waterway=security_lock“.
According to the wiki, the tag is used as a line. That's why I think it's a wrong error message.

_No response_

2.30.4

Firefox

k-yle:
Closes #10522

There is no preset for `disused:waterway=security_lock`, so it matches the fallback presets (called ["`Line`"](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/line.json) and ["`Area`"](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/area.json)). But these two presets have different tags ([`{}`](https://github.com/openstreetmap/id-tagging-schema/blob/bf95421/data/presets/line.json#L8) vs [`{ area: 'yes' }`](https://github.com/openstreetmap/id-tagging-schema/blob/bf95421/data/presets/area.json#L8-L10)) which caused the bug.

This edge-case only occurs when the tag has a lifecycle prefix, because the preset for `waterway=*` does not affect features with a lifecycle preset.

### DIFF
--- a/modules/validations/mismatched_geometry.js
+++ b/modules/validations/mismatched_geometry.js
@@ -34,6 +34,12 @@ export function validationMismatchedGeometry() {
             return null;
         }
 
+        if (asLine.isFallback() && asArea.isFallback() && !deepEqual(tagSuggestingArea, { area: 'yes' })) {
+            // if the entity matches the fallback preset, regardless of the
+            // geometry, then changing the geometry will not help.
+            return null;
+        }
+
         return tagSuggestingArea;
     }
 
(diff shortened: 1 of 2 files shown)


tyrasd (approved):
Looks good, thanks a lot!

PR #10524, opened 2024-11-02
automatically create pedestrian crossing line when joining sidewalk/road

### LINKED ISSUE #7385: Support automatically creating pedestrian crossing when sidewalk crosses roads
One of the major PITAs of helping map for OSM is when you have to stop one line you're working on just to insert a short line of something else before continuing with the original line.

The functionality of automatically adding a bridge or tunnel when a road crosses a river was really great, and made mapping a lot easier. But much, much more often than a road crossing a river, a sidewalk crosses a road. And every time that happens you have to stop, select pedestrian crossing, draw that, and then go back to the sidewalk to keep drawing that.

If the automatic bridge feature could be adjusted to also make pedestrian crossings for sidewalks automatically that would be such a time saver!

Screenshot from 2020-02-26 20-30-10

(In this screenshot there's not actually a pedestrian crossing. It's just an illustration.)

Thank you!

k-yle:
Closes #7385

When a sidewalk crosses a road, the validator already adds a crossing node, but it does not split the sidewalk line.

This PR makes the validator automatically split the sidewalk and create a [`crossing`](https://osm.wiki/Tag:footway=crossing) line.

image

after clicking <kbd>Connect using a crossing</kbd>:
image

Most of the code is re-used from the <kbd>Connect using a bridge/tunnel</kbd> action. This means the length of the automatically-generated crossing also uses the existing code.

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -476,6 +476,28 @@ export function validationCrossingWays(context) {
                     if (allowsTunnel(selectedFeatureType) && !skipTunnelFix) {
                         fixes.push(makeAddBridgeOrTunnelFix('add_a_tunnel', 'temaki-tunnel', 'tunnel'));
                     }
+
+                    // special case: if
+                    //  (1) we're about to join these lines with a highway=crossing node; and
+                    //  (2) one of the lines is a sidewalk
+                    // then we will split the sidewalk and create a highway=crossing way
+                    const isSidewalk = (
+                        entities[0].tags.footway === 'sidewalk' ||
+                        entities[1].tags.footway === 'sidewalk' ||
+                        entities[0].tags.cycleway === 'sidewalk' ||
+                        entities[1].tags.cycleway === 'sidewalk'
+                    );
+                    if (connectionTags.highway === 'crossing' && isSidewalk) {
+                        const newAction = makeAddBridgeOrTunnelFix(
+                            'connect_using_crossing',
+                            'temaki-pedestrian',
+                            'crossing',
+                            connectionTags
+                        );
+                        // replace the default action with this action
+                        fixes = fixes.filter(action => action.id !== newAction.id);
+                        fixes.unshift(newAction);
+                    }
                 }
 
                 // repositioning the features is always an option
@@ -498,7 +520,13 @@ export function validationCrossingWays(context) {
         }
     }
 
-    function makeAddBridgeOrTunnelFix(fixTitleID, iconName, bridgeOrTunnel){
+    /**
+     * @param {string} fixTitleID
+     * @param {string} iconName
+     * @param {'bridge' | 'tunnel' | 'crossing'} crossingType
+     * @param {Tags=} connectionTags
+     */
+    function makeAddBridgeOrTunnelFix(fixTitleID, iconName, crossingType, connectionTags){
         return new validationIssueFix({
             icon: iconName,
             title: t.append('issues.fix.' + fixTitleID + '.title'),
@@ -678,10 +706,10 @@ export function validationCrossingWays(context) {
                     });
 
                     var tags = Object.assign({}, structureWay.tags); // copy tags
-                    if (bridgeOrTunnel === 'bridge'){
+                    if (crossingType === 'bridge'){
                         tags.bridge = 'yes';
                         tags.layer = '1';
-                    } else {
+                    } else if (crossingType === 'tunnel') {
                         var tunnelValue = 'yes';
                         if (getFeatureType(structureWay, graph) === 'waterway') {
                             // use `tunnel=culvert` for waterways by default
@@ -689,9 +717,24 @@ export function validationCrossingWays(context) {
                         }
                         tags.tunnel = tunnelValue;
                         tags.layer = '-1';
+                    } else if (crossingType === 'crossing') {
+                        // we know that the line will already have
+                        // `footway=sidewalk` or `cycleway=sidewalk`
+                        tags[tags.footway ? 'footway' : 'cycleway'] = 'crossing';
                     }
+
                     // apply the structure tags to the way
                     graph = actionChangeTags(structureWay.id, tags)(graph);
+
+                    // for crossing, we also need to join the two lines
+                    if (crossingType === 'crossing') {
+                        const edgesToJoin = [
+                            [structEndNode1.id, structEndNode2.id],
+                            crossedEdge,
+                        ];
+                        graph = actionConnectCrossingWays(crossingLoc, edgesToJoin, connectionTags)(graph);
+                    }
+
                     return graph;
                 };
 
@@ -699,6 +742,42 @@ export function validationCrossingWays(context) {
                 context.enter(modeSelect(context, resultWayIDs));
             }
         });
+    }
+
+    /**
+     * @param {[number, number]} loc
+     * @param {[string, string][]} edges
+     * @param {Tags} connectionTags
+     */
+    function actionConnectCrossingWays(loc, edges, connectionTags) {
+        return (graph) => {
+            // create the new node for the points
+            var node = osmNode({ loc: loc, tags: connectionTags });
+            graph = graph.replace(node);
+
+            var nodesToMerge = [node.id];
+            var mergeThresholdInMeters = 0.75;
+
+            edges.forEach(function(edge) {
+                var edgeNodes = [graph.entity(edge[0]), graph.entity(edge[1])];
+                var nearby = geoSphericalClosestNode(edgeNodes, loc);
+                // if there is already a suitable node nearby, use that
+                // use the node if node has no interesting tags or if it is a crossing node #8326
+                if ((!nearby.node.hasInterestingTags() || nearby.node.isCrossing()) && nearby.distance < mergeThresholdInMeters) {
+                    nodesToMerge.push(nearby.node.id);
+                // else add the new node to the way
+                } else {
+                    graph = actionAddMidpoint({loc: loc, edge: edge}, node)(graph);
+                }
+            });
+
+            if (nodesToMerge.length > 1) {
+                // if we're using nearby nodes, merge them with the new node
+                graph = actionMergeNodes(nodesToMerge, loc)(graph);
+            }
+
+            return graph;
+        };
     }
 
     function makeConnectWaysFix(connectionTags) {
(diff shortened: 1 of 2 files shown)


modules/validations/crossing_ways.js
```diff
@@ -476,6 +476,28 @@ export function validationCrossingWays(context) {
...
+
+                    // special case: if
+                    //  (1) we're about to join these lines with a highway=crossing node; and
+                    //  (2) one of the lines is a sidewalk
+                    // then we will split the sidewalk and create a highway=crossing way
+                    const isSidewalk = (
+                        entities[0].tags.footway === 'sidewalk' ||
+                        entities[1].tags.footway === 'sidewalk' ||
+                        entities[0].tags.cycleway === 'sidewalk' ||
+                        entities[1].tags.cycleway === 'sidewalk'
```
k-yle, line 488: this part needs more discussion, currently it only applies to `footway=sidewalk` and `cycleway=sidewalk`. 

I think it would make sense to apply this autofix to any `highway=X` where `X=crossing` is a valid tag. This means it would apply to [`footway`](https://taginfo.osm.org/tags/footway=crossing), [`cycleway`](https://taginfo.osm.org/tags/cycleway=crossing), [`bridleway`](https://taginfo.osm.org/tags/bridleway=crossing), and [`path`](https://taginfo.osm.org/tags/path=crossing)

modules/validations/crossing_ways.js
```diff
@@ -701,6 +744,42 @@ export function validationCrossingWays(context) {
         });
     }
 
+    /**
+     * @param {[number, number]} loc
+     * @param {[string, string][]} edges
+     * @param {Tags} connectionTags
+     */
+    function actionConnectCrossingWays(loc, edges, connectionTags) {
```
k-yle, line 752: this function is not new, if you click [hide whitespace changes](https://github.com/openstreetmap/iD/pull/10524/files?w=1), then the github UI understands that this function was simply deindented & moved 20 lines up.

tordans:
@k-yle This looks great. JFYI there are things happening at https://github.com/facebook/Rapid/pull/1583 and https://github.com/facebook/Rapid/pull/1599 which are related to crossings but I did not look into it more, yet.

k-yle:
https://github.com/facebook/Rapid/pull/1583 seems pretty cool, if it were backported to iD, then we should probably auto-create the kerb nodes, otherwise every new crossing will immediately cause a validator warning

forteller:
As the creator of the original issue I just have to say thank you so much for doing this! <3 I hope it can be merged soon :D

tordans (changes requested):
I tested this out and it works great. Its a huge improvement and time saver.

**Possible follow up for id-tagging-schema:**
We should look into giving the unspecified crossing point and way a way that signals that it is unspecified. This will help explain that users still have something to do.
Once we have a signal from Martin that this might get merged I can create the issues to handle this over there.

**For this PR:**
I have three commits in https://github.com/tordans/iD/commits/crossing-connection-addons/

1. https://github.com/openstreetmap/iD/commit/2606f07dbd6a7959a03eb0b96ef430aa754ad456


> When creating a pedestrian crossing, the code was splitting the
> selected way instead of always splitting the sidewalk way. This
> meant if the road was selected (and then the validation fix triggered) instead of the sidewalk, the wrong
> way would be split.
> 
> Now the code identifies which way is the sidewalk and ensures that
> way is split, regardless of which way is selected.
> 
> Test case:
> 1. Create a sidewalk way with tags: footway=sidewalk, highway=footway
> 2. Create an overlapping/crossing highway way with tags: highway=tertiary
> 4. Select the HIGHWAY way (not the sidewalk)
> 5. Trigger the "Connect using a crossing" fix
> 6. Expected: Sidewalk way is split (not the highway), crossing node created
> 7. Before fix: Highway way was split incorrectly
> 8. After fix: Sidewalk way is correctly identified and split

2. https://github.com/openstreetmap/iD/commit/1cb2c25999005c12f6de6efa399958d43c440ce8

> When creating a pedestrian crossing between a sidewalk and a road,
> if both ways had bridge tags (bridge=yes, layer=1), the crossing
> fix was not being created because the code was inside an else-if
> block that only executed when isCrossingBridges was false.
> 
> Changed the else-if to a separate if statement so the crossing fix
> code runs regardless of bridge/tunnel tags.
> 
> Test case:
> 1. Create a sidewalk way with tags: bridge=yes, footway=sidewalk, highway=footway, layer=1
> 2. Create an overlapping/crossing highway way with tags: bridge=yes, highway=tertiary, layer=1
> 4. Select the sidewalk way
> 5. Trigger the "Connect using a crossing" fix
> 6. Expected: Crossing node is created AND sidewalk way is split with footway=crossing tags
> 7. Before fix: Only crossing node created, no split
> 8. After fix: Both crossing node and split work correctly

3. https://github.com/tordans/iD/commit/8cd089de87d9c5d0d50e11ce92e6d854c45bb9ef as an answer to you comment https://github.com/openstreetmap/iD/pull/10524/changes#r1826530119

I think we should apply this new feature to all crossings that have dedicated crossing presets right now, which is what this commit does.

Update/Fix: This now work for path, cycleway, footway

tordans:
Ah, and then I tried to go a step further:

At places like https://www.openstreetmap.org/way/1000285274/history#map=19/52.481189/13.452202&layers=N **it would be great if the split happens on the `way(barrier=kerb)` and those new nodes had a `barrier=kerb`**. Easy enough to find the right intersecting kerb, right? Unfortunately not (for me). I fiddled with this for more than an hour. WIP is at https://github.com/tordans/iD/commit/a86703d51e026936d295ad3088ad7a19692f5f4a
The only useful part of this commit is probably the test case at https://github.com/tordans/iD/commit/a86703d51e026936d295ad3088ad7a19692f5f4a#diff-eb48a4c6f83b98aaf09f65841ef030279805649a578b68ddf8d2403eeb469e41R21-R56 and the idea to find the right intersecting kerb-line on each side of the road by first splitting the orthogonal footway and then find the right intersection per side https://github.com/tordans/iD/commit/a86703d51e026936d295ad3088ad7a19692f5f4a#diff-30223f93411f392dcf5558fae40993174f2364bf0e7f120790d2931fd15dfb2bR824-R829.
However, all this only looks at the most simple setup, yet, so dual carriageways, multiple kerb ways and such need to be tested as well or the code should just skip this magic in those cases…

So this is definitely something for a follow up. Maybe I should have looked at https://github.com/openstreetmap/iD/issues/10757 more closely first…

---

**One question we should answer explicitly for this PR though:** 
Are we sure we don't want to add `barrier=kerb` on those split nodes that we create?
In my mind, a crossing always also has some kind of kerb… and be that the kerb is flush. If not, is it really a crossing (way)?
The advantage of adding the unspecified `barrier=kerb` is that StreetComplete users have a nice data point that triggers a review of those on the ground.

forteller:
Thank you, @tordans! This would be incredible to finally have, and would for sure make me willing to do more work on OSM after being implemented! 

> Are we sure we don't want to add barrier=kerb on those split nodes that we create?
> In my mind, a crossing always also has some kind of kerb… and be that the kerb is flush.

Seeing as almost all crossings does have a kerb, and seeing as [the kerb key has a 'no' value](https://wiki.openstreetmap.org/wiki/Key:kerb), to let people know that even if it looks like there should be a kerb here there's not, so you don't have to check… based on that I think the ideal thing would be for a kerb key to be added to start and end nodes.

But! The most important thing is for the main part of this to be implemented. If deciding on adding a kerb key or not, how to do it, etc, might delay this feature, then I say: Drop it for now, implement the important part of the feature, and think about adding kerb nodes later.

Thank you!

matkoniecz:
>  In my mind, a crossing always also has some kind of kerb… and be that the kerb is flush. If not, is it really a crossing (way)?

why not?

screen-2026-08-10-11-35-37

no kerb on one end (if area get torturously micromapped) and clearly crossing

https://www.openstreetmap.org/#map=19/50.074454/19.929813

https://www.google.pl/maps/@50.0743224,19.9291338,3a,47.2y,344.86h,78.17t/data=!3m7!1e1!3m5!1s5qNTtzTJpiob94hchYxCdg!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fcb_client%3Dmaps_sv.tactile%26w%3D900%26h%3D600%26pitch%3D11.829941404105512%26panoid%3D5qNTtzTJpiob94hchYxCdg%26yaw%3D344.8634776712857!7i16384!8i8192?entry=ttu&g_ep=EgoyMDI2MDgwNS4xIKXMDSoASAFQAw%3D%3D

matkoniecz:
https://wiki.openstreetmap.org/wiki/Key:crossing:continuous more broadly

modules/validations/crossing_ways.js
```diff
@@ -478,12 +474,29 @@ export function validationCrossingWays(context) {
...
-                        } else {
-                            fixes.push(makeAddBridgeOrTunnelFix('add_a_tunnel', 'temaki-tunnel', 'tunnel'));
-                        }
+                        fixes.push(makeAddBridgeOrTunnelFix('add_a_tunnel', 'temaki-tunnel', 'tunnel'));
+                    }
+
+                    // special case: if
+                    //  (1) we're about to join these lines with a highway=crossing node; and
+                    //  (2) one of the lines is a sidewalk
+                    // then we will split the sidewalk and create a highway=crossing way
```
matkoniecz, line 483:
```suggestion
                    // then we will split the sidewalk and create a way marked as a crossing
```

way does not get `highway=crossing`

modules/validations/crossing_ways.js
matkoniecz: definitely makes sense for path but note that `footway=sidewalk` is fine on path

and I think I have seen `highway=path bicycle=designated foot=designated segregated=yes` with both `footway=sidewalk` and `cycleway=sidewalk`

modules/validations/crossing_ways.js
```diff
@@ -476,6 +476,28 @@ export function validationCrossingWays(context) {
...
+                    const isSidewalk = (
+                        entities[0].tags.footway === 'sidewalk' ||
+                        entities[1].tags.footway === 'sidewalk' ||
+                        entities[0].tags.cycleway === 'sidewalk' ||
+                        entities[1].tags.cycleway === 'sidewalk'
+                    );
```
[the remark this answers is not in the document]
matkoniecz: this does not sound like a good idea to me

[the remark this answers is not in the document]
k-yle: yeah tobias' solution from https://github.com/openstreetmap/iD/pull/10524#pullrequestreview-3673558298 is a better approach.

I haven't addressed any feedback yet because I believe this PR is still pending an inital review, to confirm if the idea would be accepted

PR #10552, opened 2024-11-23
Fix hover inspector UI overflow & scrollbars

### LINKED ISSUE #9070: When hovering over a note, text goes beyond the border
### URL

https://www.openstreetmap.org/edit#map=20/32.16334/-110.99066

1. Go to a note with a large amount of text in the first message. (Varies by formatting, etc.)
2. Hover over the note in the editor.
The issue should be present by going through these two steps.

I was not able to test the issue on Safari. 

Screenshot

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.com, RapiD version at mapwith.ai/rapid

Firefox, Chrome, Microsoft Edge

### LINKED ISSUE #9910: When hovering over an object with a field icon, the field icon goes over the border
Displays scroll bars, indicating icon is too large / box is too small to be displayed by default.
Example: tree node with leaf type set.

It does not seem resolution or zoom dependent.

See also #9070

1. Open the editor with the "Edit" button (or manually with the down arrow menu to the right of it).
image
2. Create (or select) an object with a 'sub-type' field which usually has an icon, i.e. a tree with "leaf type" can be "broadleaved".

Let's add a node
image
...make this node a regular tree by selecting from the presets
image
...and then select a sub-type. Here it is for example "leaf type". Let's select "Broadleaved tree".
image

I have now specified a sub-type for the object. It looks like this when it is selected.

image

6. Deselect the object by clicking anywhere on the map. The UI will change to a search bar as usual.
image

8. **Hover over** the object, i.e. a tree node. The UI will change to a preview of the object, as usual.

The "Broadleaved tree" icon is used for the preset "Broadleaved tree (deciduous)" itself at the top, in this case.
It is also present in the "leaf type" field to give users a visual indication of the subtype regardless.
If you look at the field "leaf type" it will have the "broadleaved tree" icon, but there *it is obscured by two sliders*, indicating the icon or the box it is in may not fit inside the UI properly when you hover over it. (*incorrect behaviour*)

image

9. **Select** the object

If we actually select it again, it looks fine. There are no sliders or similar obscuring the icon and the icon itself is fully visible and fits inside the box in the UI. (*expected behaviour*)

image

image

(Rapid does not have hover UI.)

2.27.1, 2.28.0-dev

Chrome, Microsoft Edge

### LINKED ISSUE #10551: Hover inspector UI overflow

Nekzuris:
This small CSS change fixes #10551, fixes #9910 and closes #9912
before & after
overflow no overflow

These extra scrollbars were added by `overflow-y: auto;` in #9074 to fix #9070
I reverted it to its previous value of `visible` and made sure not to reintroduce the previous overflow issue in notes with [this](https://github.com/openstreetmap/iD/commit/42190dfb8d52f3127f3aa4b8267a898a9b7ffa54#diff-3f86fb5cebd04342575e756d440d18a0b6cfb131a3c26d7897ff1793f8b6d30bR4241)
You can test it here https://www.openstreetmap.org/edit#map=18/48.129455/7.827063

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4232,9 +4232,12 @@ li.issue-fix-item button:not(.actionable) .fix-icon {
     border: 1px solid #ccc;
 }
 
-/* scrollbars only when necessary*/
+/* scrollbars only when necessary */
 .inspector-hover div {
     overflow-x: visible;
+    overflow-y: visible;
+}
+.inspector-hover .comment div {
     overflow-y: auto;
 }
 


Nekzuris:
I updated my description to make it clearer for reviewers.

Nekzuris:
What's wrong with this PR @1ec5 ? You're taking time to review other PR but ignore this one, it goes with #10554.

tyrasd:
Thanks for giving this one a try and for the patience. I've fixed the issue slightly differently, as there was already a css rule for `.comment-text` which already had the required overflow property, but was just not in effect in the hovered state because the other rule did override it. See 18de00a7bfd26155cb5fe25789c926a6aff27ec5

PR #10554, opened 2024-11-24
Thin scrollbars for Firefox

Nekzuris:
Since Firefox doesn't use `::-webkit-`, we should use `scrollbar-width: thin` to better match the style on Chrome. 
This doesn't override other `scrollbar-width: none`

Chrome scrollbars:
image
Firefox scrollbars:
ff
Firefox with thin scrollbars:
ff_

Also I’m quite sure the rule `body { scrollbar-width: 10px; }` doesn’t do anything because it’s unsupported and should be deleted.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -5774,9 +5774,6 @@ li.hide + li.version .badge .tooltip .popover-arrow {
         background-color: rgba(0,0,0,.05);
     }
 }
-body {
-    scrollbar-width: 10px;
-}
 /* Firefox */
 @-moz-document url-prefix() {
 	* {
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -5777,6 +5777,12 @@ li.hide + li.version .badge .tooltip .popover-arrow {
 body {
     scrollbar-width: 10px;
 }
+/* Firefox */
+@-moz-document url-prefix() {
+	* {
+	    scrollbar-width: thin;
+	}
+}
 
 
 /* Intro walkthrough


css/80_app.css
```diff
@@ -5777,6 +5777,12 @@ li.hide + li.version .badge .tooltip .popover-arrow {
 body {
     scrollbar-width: 10px;
```
1ec5: Yeah, I can’t find anything that suggests this was ever a valid or supported syntax in CSS. Feel free to remove it.

css/80_app.css
```diff
@@ -5777,6 +5777,12 @@ li.hide + li.version .badge .tooltip .popover-arrow {
...
 }
+/* Firefox */
+@-moz-document url-prefix() {
```
1ec5, line 5778: [This documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/@document) says `@document` is deprecated and the prefixed version no longer works in author-defined stylesheets. Can we count on it to work on Firefox going forward?

Nekzuris: I'm not sure, the browser compatibility table says it is supported for Firefox browser detection.

Maybe I'm missing something, but I don't see any other way. If we set `scrollbar-width: thin;` for everyone, I don't think we can override it with `::-webkit-scrollbar`.

1ec5: I see, is the intention to avoid clobbering the `width` here with whatever WebKit would use as `thin` in its (currently experimental) implementation?

https://github.com/openstreetmap/iD/blob/d04c3cdb73bc4661232d453ff749208ca7aedcba/css/80_app.css#L5749-L5754

I suppose this specific `url-prefix` hack does still work, so it’s an improvement over the status quo.

1ec5 (approved):
Thanks for this contribution. As a longtime Firefox user, I had no idea the scrollbars were supposed to be thin!

css/80_app.css
Nekzuris: Yes, I think the current WebKit scrollbar looks better, especially with the rounded borders, but I couldn’t find any way to achieve the same on Firefox.

PR #10556, opened 2024-11-25
fixed the disable and enable button position

### LINKED ISSUE #8878: Move "Deactivate/Activate all" to top of list
### Description

When choosing which objects to show there's a long list of objects to choose from, and at the very end there's the options "Deactivate all" and "Activate all". Good options to have, but only if you know they're available. And having them at the very end of that long list makes them very easy to overlook. I think it would be much better to have them at the top instead, so they're discoverable by everyone.

Thanks!

### Screenshots

Screenshot from 2022-01-05 20-07-06

Here you can see that the options to activate or deactivate all is hidden until you scroll down

draunger:
# Description
## Approach
I have used position absolute property and margin property to set buttons.
## Before the fix 
image
## After the fix
image

# Attached issue 
Closes #8878

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -3723,6 +3723,9 @@ button.autofix.action.active {
     flex-flow: row nowrap;
     justify-content: flex-end;
     height: 30px;
+    position: absolute;  /*fixed the disable and enable all button position*/ 
+    margin-top: -532px;
+    margin-left: 182px;
 }
 .section-footer a {
     padding: 5px;
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -3723,9 +3723,6 @@ button.autofix.action.active {
     flex-flow: row nowrap;
     justify-content: flex-end;
     height: 30px;
-    position: absolute;  /*fixed the disable and enable all button position*/ 
-    margin-top: -532px;
-    margin-left: 182px;
 }
 .section-footer a {
     padding: 5px;
--- a/modules/ui/sections/map_features.js
+++ b/modules/ui/sections/map_features.js
@@ -11,52 +11,51 @@ export function uiSectionMapFeatures(context) {
         .disclosureContent(renderDisclosureContent)
         .expandedByDefault(false);
 
-    function renderDisclosureContent(selection) {
-
-        var container = selection.selectAll('.layer-feature-list-container')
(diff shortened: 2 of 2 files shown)


draunger:
@bhousel can you review my PR and if you seems it has some issue please tell me regarding this

bhousel:
> @bhousel can you review my PR and if you seems it has some issue please tell me regarding this

Using absolute positioning and a -532px margin is certainly one way to move the text, but it's not how I would do it.  
What if later someone wants to change the checkboxes or the font or something.

Screenshot 2024-12-09 at 3 47 50 PM

draunger:
@bhousel ok i understand then please give me some time i can change my approach 
but sir one question ,  without applying css how we can change the position of the buttons

bhousel:
> @bhousel ok i understand then please give me some time i can change my approach
> but sir one question , without applying css how we can change the position of the buttons

insert them into the document where you want them to be

draunger:
hey @bhousel i have changed the position of the footer without css but now it look like this but in this image map feature text and both button i.e(disable all and enable all) are not aligning in the same line.
image
if i use absolute property in map feature text css then it looks like this :
means i have apply css on the text map feature not on the buttons :
so which i use above or this ??
image

draunger:
hey @bhousel what do you think which method i use as stated above ?

k-yle (changes requested):
Hi @draunger, as Bryan mentioned, using a margin of `-532px` isn't a particularly robust solution. It also has consqeuences for other components that use the same class name.

It would be more idiomatic to move `div.feature-list-links` above `ul.layer-feature-list`. You can find the relevant code here:

https://github.com/openstreetmap/iD/blob/ff8f044994ef418b614dd18f73c3e81e2946b381/modules/ui/sections/map_features.js#L27-L29

draunger:
@k-yle ok sending its commit using moving div.feature-list-links above ul.layer-feature-list

draunger:
@k-yle i have made the changes as you said , you can now review it :)

draunger:
@tyrasd in this PR i have changed the "disable all" and "enable all" buttons position as you can see in this:
image

tyrasd:
closing in favor of #11268

PR #10560, opened 2024-11-27
Ensure correct this context for addMembership function to call the blur method correctly. Fixes #10555

### LINKED ISSUE #10555: Error adding way to new relation — can't access property "blur", this is undefined
### URL

https://ideditor.netlify.app/#background=Mapbox&disable_features=boundaries&id=w168825325&locale=en&map=20.43/39.28230/-84.28460

1. Select [this footpath](https://www.openstreetmap.org/way/168825325).
2. Click the ➕ button in the Relations section of the inspector and select “New relation…”.

Nothing happens. The dropdown menu doesn’t close.

```
TypeError: can't access property "blur", this is undefined
```

https://github.com/openstreetmap/iD/blob/3bda6c78ee9a5952145ce07ea7e60246f9ac4107/modules/ui/sections/raw_membership_editor.js#L560 https://github.com/openstreetmap/iD/blob/3bda6c78ee9a5952145ce07ea7e60246f9ac4107/modules/ui/sections/raw_membership_editor.js#L174

2.31.0-dev

Firefox

Deeptanshu-sankhwar:
Fixes issue #10555 

- Pass the `this` reference explicitly to the `addMembership` function to ensure the correct context is available.
- Resolved an issue where `this` was undefined due to implicit context binding in event handlers.
- Updated `acceptEntity` to pass the triggering DOM element explicitly as a parameter.

Screen Recording 2024-11-27 at 6 25 10 AM

### DIFF
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -170,8 +170,10 @@ export function uiSectionRawMembershipEditor(context) {
     }
 
 
-    function addMembership(d, role) {
-        this.blur();           // avoid keeping focus on the button
+    function addMembership(d, role, domElement) {
+        if (domElement && typeof domElement.blur === 'function')   {
+            domElement.blur(); // avoid keeping focus on the button
+        }
         _showBlank = false;
 
         function actionAddMembers(relationId, ids, role) {
@@ -557,7 +559,7 @@ export function uiSectionRawMembershipEditor(context) {
             if (d.relation) utilHighlightEntities([d.relation.id], false, context);
 
             var role = context.cleanRelationRole(list.selectAll('.member-row-new .member-role').property('value'));
-            addMembership(d, role);
+            addMembership(d, role, this);
         }
 
 
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -170,10 +170,7 @@ export function uiSectionRawMembershipEditor(context) {
     }
 
 
-    function addMembership(d, role, domElement) {
-        if (domElement && typeof domElement.blur === 'function')   {
-            domElement.blur(); // avoid keeping focus on the button
-        }
+    function addMembership(d, role) {
         _showBlank = false;
 
         function actionAddMembers(relationId, ids, role) {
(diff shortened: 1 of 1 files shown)


modules/ui/sections/raw_membership_editor.js
```diff
@@ -170,8 +170,10 @@ export function uiSectionRawMembershipEditor(context) {
     }
 
 
-    function addMembership(d, role) {
-        this.blur();           // avoid keeping focus on the button
+    function addMembership(d, role, domElement) {
+        if (domElement && typeof domElement.blur === 'function')   {
+            domElement.blur(); // avoid keeping focus on the button
+        }
```
1ec5: This is a defensive fix, but since `addMembership` is only called from one place, I think it would be cleaner to move this `blur` call to the caller, or even replace this with a closure that blurs the element, which would explicitly be in scope:

https://github.com/openstreetmap/iD/blob/d46dcea320616c5cca253599f4712dfa5a04346e/modules/ui/sections/raw_membership_editor.js#L507

As it is, if `acceptEntity` has to bail for some reason, it won’t blur the element, which might result in the menu getting stuck on screen.

Deeptanshu-sankhwar: Thank you for the suggestion @1ec5 ! I completely agree that handling the blur() in a closure at the .on('accept') level would ensure the element is blurred reliably and keeps the UI logic separate from addMembership. I will updat the code to use the closure approach, which explicitly handles the blurring before calling acceptEntity.

Deeptanshu-sankhwar:
I've updated the PR with the updated code to use the closure approach, which explicitly handles the blurring before calling acceptEntity. This change aligns with the feedback and avoids potential issues with the menu getting stuck on screen. Let me know if there are any further adjustments you'd recommend! 🙌

1ec5 (approved):
Thanks, and congratulations on your first contribution!

Deeptanshu-sankhwar:
Thank you very much @1ec5 ! Hoping to be more involved.

tyrasd:
Sorry to jump in a bit late on this here. But was there a reason for this particular approach to fix this, compared to just passing the expected `this` object to the `addMembership` method like the following ([here](https://github.com/Deeptanshu-sankhwar/iD/blob/78cc893bd7ef2cde075da82db5fa22a71c38ff80/modules/ui/sections/raw_membership_editor.js#L562)):

```js
            addMembership.call(this, d, role);
```

At first glance, the fix from this PR seems unnecessarily convoluted. :sweat_smile:

Deeptanshu-sankhwar:
Thanks for the feedback, @tyrasd ! The reasoning behind this approach was to keep UI logic (like `blur()`) separate from the business logic in `addMembership`. Handling `blur()` in the closure ensures it always executes reliably, even if `acceptEntity` exits early, and avoids tightly coupling `addMembership` to UI-specific behavior. Let me know if you think this could still be simplified further! 🙂

PR #10561, opened 2024-11-27
Refactor Address Suggestion Logic to Improve Readability and Align with Addressing Standards. Fixes #10541

### LINKED ISSUE #10541: Address field dropdowns should choose a name in the relevant language
As far as I know, most postal services and other addressing authorities use only one name at a time in each part of an address (street, city, etc.), even if OSM would give those features multiple values in `name=*` as a linguistic compromise.

For example, I recently had to [retag](https://www.openstreetmap.org/changeset/159141323) a bunch of POIs where a mapper had previously accepted iD’s suggestion of `addr:street=Bellaire Boulevard;Đại Lộ Sàigòn` based on [this nearby street](https://www.openstreetmap.org/way/1067465473), which is tagged `name=Bellaire Boulevard;Đại Lộ Sàigòn`. This street has two values in `name=*` because of [dual wayfinding signs](https://www.mapillary.com/app/?pKey=314599710013458). But neither the U.S. Postal Service nor the county records manager recognizes “Đại Lộ Sàigòn” as a street name for addressing purposes. Even if they did, they still wouldn’t recognize a dual name. (Do other countries’ postal systems work like this? Please correct me if I’m wrong.)

Ideally, iD would somehow know that it should use `name:en=Bellaire Boulevard` instead, because that’s the standard language for addressing in most of the U.S. We might be able to use data/territory_languages.json for this purpose, but it gets tricky in multilingual countries where the countrywide default language may not be a good fit. Since iD also consults Nominatim, maybe it can access the `default_language=*` from the surrounding boundaries, or it could base the decision on the interface language. (Not great, but this is a fallback.) Or maybe iD could simply use the `name=*` but truncate it at the first delimiter (a semicolon in this case).

https://github.com/openstreetmap/iD/blob/4ffc0bcd824b98ac69ca8ce4c9ab83ac2f309db9/modules/ui/fields/address.js#L58

Deeptanshu-sankhwar:
Fixes #10541 

This PR refactors the logic for determining address suggestions to improve readability and ensure compliance with addressing standards. The updated logic falls back to `name:en` for English names, and truncates multi-value name tags at the first delimiter (;) to handle dual names appropriately. 

Screenshot 2024-11-27 at 10 24 44 AM

### DIFF
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -55,7 +55,14 @@ export function uiFieldAddress(field, context) {
                     dist = geoSphericalDistance(choice.loc, l);
                 }
 
-                const value = resultProp && d.tags[resultProp] ? d.tags[resultProp] : d.tags.name;
+                const value = resultProp && d.tags[resultProp]
+                    ? d.tags[resultProp]
+                    : d.tags['name:en']
+                        ? d.tags['name:en'] // Fallback to English name
+                        : d.tags.name
+                            ? d.tags.name.split(';')[0].trim() // Use the first part of name (split by delimiter `;`)
+                            : null; // Default to null if no valid value exists
+
                 let title = value;
                 if (type === 'street') {
                     title = `${addrField.t('placeholders.street')}: ${title}`;


modules/ui/fields/address.js
```diff
@@ -55,7 +55,14 @@ export function uiFieldAddress(field, context) {
                     dist = geoSphericalDistance(choice.loc, l);
                 }
 
-                const value = resultProp && d.tags[resultProp] ? d.tags[resultProp] : d.tags.name;
+                const value = resultProp && d.tags[resultProp]
+                    ? d.tags[resultProp]
+                    : d.tags['name:en']
+                        ? d.tags['name:en'] // Fallback to English name
```
1ec5, line 61: If I’m reading this code correctly, it doesn’t fall back to `name:en`, it prioritizes `name:en` over `name`. But not every region speaks English; that was just an example I gave. I gave some more details about how you might solve this problem in https://github.com/openstreetmap/iD/issues/10541#issuecomment-2504476250. However, the simplest solution may just be to truncate on the first semicolon; we could leave #10541 open to address the linguistic problem separately.

modules/ui/fields/address.js
```diff
@@ -55,7 +55,14 @@ export function uiFieldAddress(field, context) {
...
+                        : d.tags.name
+                            ? d.tags.name.split(';')[0].trim() // Use the first part of name (split by delimiter `;`)
+                            : null; // Default to null if no valid value exists
```
1ec5, line 64: I think this code would be clearer if you first refactor the line to use if/else statements instead of a ternary. Nested ternaries are very difficult to reason about. (Besides, without parentheses, I wouldn’t know at a glance whether the order of precedence is actually what you’ve implied using whitespace.)

matkoniecz:
I guess that it should be closed as abandoned - and if author is interested it can be reopened

tyrasd:
I would also argue that the logic of using `name:*` for address parts is not necessarily correct: the address is typically given in the local language, and AFAIK geocoders like nominatim rely on the match between the `addr:` tags and the `name` of nearby features.

Issue #10558, opened 2024-11-26
Street level imagery: Limit maximum viewer size


danieldegroot2:
### How to reproduce the issue?

Limit the maximum viewer size to stay inside the interactable frame, at least in height.
While the resize button can still -barely- be interacted with at this size, the viewer itself gets cut off vertically.

Alternatively, the viewer could be visible through the top toolbar UI (as it is already transparent), but this is likely not desirable.

Chrome 131.0.6778.86 on Windows 10

### Screenshot(s) or anything else?

![image](https://github.com/user-attachments/assets/5fc5068f-daa1-4d61-971d-5774b0f9d570)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app

### What version numbers does this issue effect?

2.30.4, 2.31.0-dev

### Which browsers are you seeing this problem on?

Chrome

draunger:
@danieldegroot2 i will do this issue

draunger:
@danieldegroot2 i have make PR of this issue as https://github.com/openstreetmap/iD/pull/10563

PR #10563, opened 2024-11-29
fixed limit maximum viewer size in height

draunger:
# Description
## Approach
i have used transform property in css to limit the maximum viewer size .
# Before the fix:
image
# After the fix:
image

# Attached issue
https://github.com/openstreetmap/iD/issues/10558

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -25,10 +25,12 @@ li.list-item-photos.active:after {
 .ideditor[dir='ltr'] .photoviewer {
     margin-left: 10px;
     margin-right: 2px;
+    transform: translate(10px,21px);
 }
 .ideditor[dir='rtl'] .photoviewer {
     margin-right: 10px;
     margin-left: 2px;
+    transform: translate(10px,21px);
 }
 
 @media screen and (min-width: 1600px) {


tyrasd:
Thanks for giving this a shot @draunger, but this solution had an issue, as the viewer would clip the map footer, making the photo attribution, username, etc. hard to read:

image

It would have been better to adjust the limits in the javascript resizing code (e.g. here: https://github.com/openstreetmap/iD/blob/34bb14/modules/ui/photoviewer.js#L219), but I've now replaced the hardcoded limits with dynamically calculated correct margins in 404e085.

draunger:
@tyrasd can you tell me which problem had you seen in my PR as I have applied transfrom property in css to correct the problem which is described in the issue??

draunger:
> Thanks for giving this a shot @draunger, but this solution had an issue, as the viewer would clip the map footer, making the photo attribution, username, etc. hard to read:
> 
> image
> 
> It would have been better to adjust the limits in the javascript resizing code (e.g. here: https://github.com/openstreetmap/iD/blob/34bb14/modules/ui/photoviewer.js#L219), but I've now replaced the hardcoded limits with dynamically calculated correct margins in 404e085.
> 
Okay but if you tell me this problem I will fix it

Issue #10531, opened 2024-11-07
KartaView: Only highlight sequence on hover


danieldegroot2:
### Description

Minor visual remark

For KartaView, only highlight sequence and cone in focus on hover, not also all individual image cones.
This is inconcistent with all other providers. This behaviour has existed for a long time.

### Screenshots

![image](https://github.com/user-attachments/assets/e77bfcd9-300a-4da3-aca7-81240928bace)

draunger:
@danieldegroot2 i have done this issue as https://github.com/openstreetmap/iD/pull/10565

PR #10565, opened 2024-11-30
Highlight

draunger:
# Description
# Before the fix:
image
# After the fix:
image
# Attached issue :
https://github.com/openstreetmap/iD/issues/10531

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -149,8 +149,6 @@ li.list-item-photos.active:after {
 }
 .viewfield-group.highlighted circle {
     stroke: #222;
-    /* stroke-opacity: 0.9; */
-    /* fill-opacity: 0.9; */
 }
 .viewfield-group.highlighted.hovered circle {
     stroke: #222;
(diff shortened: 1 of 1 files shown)


tyrasd:
Hi @draunger. Really appreciate your effort on this PR!

However, I don't necessarily understand how your solution was meant to address the issue. Maybe a bit of explanation and a better PR title would have helped. :wink: 

See [https://github.com/openstreetmap/iD/commit/f38bc5fc789cd0ebb7d0beff1c4427b15e50fe95 line 530](https://github.com/openstreetmap/iD/commit/f38bc5fc789cd0ebb7d0beff1c4427b15e50fe95#diff-a1012da97d7e83701133fcac732015b2968a41ce9992646f2eb8640fb4236915R530) for where the inconsistency actually originated (compare against the corresponding line in the panoramax module for example).

Cheers!

draunger:
@tyrasd i have changes some css so that  on hover only one triangle highlight not the all triangles

As described in the issue

PR #10568, opened 2024-11-30
add a short reminder for data being objective during the walkthrough, fixes #10566

### LINKED ISSUE #10566: Walkthrough: Add short reminder that data should be objective
### Description

The iD walkthrough is the first point of contact with OSM for many new users who are not yet familiar with our customs and mapping practices. That includes business owners wanting to add their business to OSM, who might be tempted to embellish the information they are adding. Subjective descriptions, 'reviews' or SEO optimization fluff text have no place in OSM, but new users unfortunately might not be fully aware of that.

Adding a short sentence about the OSM subjectivity standards to the walkthrough could help to cut down on the amount of SEO spam descriptions that are added by new users.
(Obviously some people will unfortunately still choose to do it anyway, but ignorance about the topic would be less of a factor.)

A good spot for this reminder would be during the walkthrough section where users are guided to add a new cafe POI. A simple short sentence along the lines of _"All information should be objective and not a personal opinion."_ would be helpful.

### Screenshots

image
#

Deeptanshu-sankhwar:
Fixes #10566 

I have added a reminder statement in `core.yaml` and rendered it in `point.js` to make it visible during the walkthrough.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2213,6 +2213,7 @@ en:
       feature_editor: "The point is now marked as a cafe. Using the feature editor, we can add more information about the cafe."
       fields_info: "In OpenStreetMap, all of the fields are optional, and it's OK to leave a field blank if you are unsure."
       add_name: "Let's pretend that you have local knowledge of this cafe, and you know its name. **Add a name for the cafe.**"
+      add_reminder: "All information should be objective and not a personal opinion."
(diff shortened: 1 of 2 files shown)


Deeptanshu-sankhwar:
I've updated the PR accordingly over here @tyrasd . Please take a look whenever you can. 

Thanks!

tyrasd (approved):

PR #10570, opened 2024-11-30
Fix for issue #4090: cmd.js line 20-26 replaced with OS-specific keyboard shortcut mappings with localization.

### LINKED ISSUE #4090: Localize keyboard button texts  (e.g. "Ctrl" -> "Strg")
There is one untranslatable item at Browsing "Enter full screen mode".
Maybe you can make some button-texts translatable:
"Ctrl", "Space", "PgUp", "PgDn", "Home", "End", "Backspace"
maybe even those I would use the same Text in German as in English:
"Shift", "Pause", "Alt", "Enter", "Esc"

image

StillAbeginnerr:
Description of Code Change
The proposed change enhances keyboard shortcut mappings to dynamically adapt based on the operating system.

Original Code

The replacements object statically maps symbols (e.g., ⌘, ⇧) to generic labels (e.g., Ctrl, Shift), with no differentiation for macOS.

image

This shows CTRL + Shift + C on German language.

---

Proposed Change :

OS Detection
Introduces a mac variable to determine if the operating system is macOS (detected.os === 'mac').

Dynamic Mapping
For macOS:
Uses macOS-specific symbols and labels (e.g., ⌘ Cmd, ⇧ Shift).
Prepends symbols to enhance clarity.

For Others:
Defaults to generic labels like Ctrl, Alt, and Shift.

Localization:
Integrates t() to fetch OS-specific, human-readable labels, supporting multiple languages.

image

This newly proposed shows STRG + Umschalt + C on German language.

### DIFF
--- a/modules/ui/cmd.js
+++ b/modules/ui/cmd.js
@@ -17,11 +17,11 @@ export var uiCmd = function (code) {
     var mac = (detected.os === 'mac');
     var result = '',
         replacements = {
-            '⌘': mac ?  t('shortcuts.key.cmd')    : t('shortcuts.key.ctrl'),
-            '⇧': mac ? t('shortcuts.key.shift')  : t('shortcuts.key.shift'),
-            '⌥': mac ?  t('shortcuts.key.option') : t('shortcuts.key.alt'),
-            '⌫': mac ?  t('shortcuts.key.delete') : t('shortcuts.key.backspace'),
-            '⌦': mac ?  t('shortcuts.key.del')    : t('shortcuts.key.del'),
+            '⌘': mac ? '⌘ '  : t('shortcuts.key.ctrl'),
+            '⇧': mac ? '⇧ '  : t('shortcuts.key.shift'),
+            '⌥': mac ? '⌥ '  : t('shortcuts.key.alt'),
+            '⌫': mac ? '⌫ '  : t('shortcuts.key.backspace'),
+            '⌦': mac ? '⌦ '  : t('shortcuts.key.del'),
         };
 
     for (var i = 0; i < code.length; i++) {
--- a/modules/ui/cmd.js
+++ b/modules/ui/cmd.js
@@ -14,14 +14,15 @@ export var uiCmd = function (code) {
     if (detected.os === 'win') {
         if (code === '⌘⇧Z') return 'Ctrl+Y';
     }
-
+    var mac = (detected.os === 'mac');
     var result = '',
         replacements = {
-            '⌘': 'Ctrl',
-            '⇧': 'Shift',
-            '⌥': 'Alt',
-            '⌫': 'Backspace',
-            '⌦': 'Delete'
+            '⌘': mac ? '⌘ ' + t('shortcuts.key.cmd')    : t('shortcuts.key.ctrl'),
+            '⇧': mac ? '⇧ ' + t('shortcuts.key.shift')  : t('shortcuts.key.shift'),
+            '⌥': mac ? '⌥ ' + t('shortcuts.key.option') : t('shortcuts.key.alt'),
+            '⌃': mac ? '⌃ ' + t('shortcuts.key.ctrl')   : t('shortcuts.key.ctrl'),
+            '⌫': mac ? '⌫ ' + t('shortcuts.key.delete') : t('shortcuts.key.backspace'),
+            '⌦': mac ? '⌦ ' + t('shortcuts.key.del')    : t('shortcuts.key.del'),
         };
 
     for (var i = 0; i < code.length; i++) {
@@ -36,12 +37,14 @@ export var uiCmd = function (code) {
 };
 
 
+
 // return a display-focused string for a given keyboard code
 uiCmd.display = function(code) {
     if (code.length !== 1) return code;
 
     var detected = utilDetect();
     var mac = (detected.os === 'mac');
+    
     var replacements = {
         '⌘': mac ? '⌘ ' + t('shortcuts.key.cmd')    : t('shortcuts.key.ctrl'),
         '⇧': mac ? '⇧ ' + t('shortcuts.key.shift')  : t('shortcuts.key.shift'),
--- a/modules/ui/cmd.js
+++ b/modules/ui/cmd.js
@@ -17,12 +17,11 @@ export var uiCmd = function (code) {
     var mac = (detected.os === 'mac');
     var result = '',
         replacements = {
-            '⌘': mac ? '⌘ ' + t('shortcuts.key.cmd')    : t('shortcuts.key.ctrl'),
-            '⇧': mac ? '⇧ ' + t('shortcuts.key.shift')  : t('shortcuts.key.shift'),
-            '⌥': mac ? '⌥ ' + t('shortcuts.key.option') : t('shortcuts.key.alt'),
-            '⌃': mac ? '⌃ ' + t('shortcuts.key.ctrl')   : t('shortcuts.key.ctrl'),
-            '⌫': mac ? '⌫ ' + t('shortcuts.key.delete') : t('shortcuts.key.backspace'),
-            '⌦': mac ? '⌦ ' + t('shortcuts.key.del')    : t('shortcuts.key.del'),
+            '⌘': mac ?  t('shortcuts.key.cmd')    : t('shortcuts.key.ctrl'),
+            '⇧': mac ? t('shortcuts.key.shift')  : t('shortcuts.key.shift'),
+            '⌥': mac ?  t('shortcuts.key.option') : t('shortcuts.key.alt'),
+            '⌫': mac ?  t('shortcuts.key.delete') : t('shortcuts.key.backspace'),
+            '⌦': mac ?  t('shortcuts.key.del')    : t('shortcuts.key.del'),
         };
 
     for (var i = 0; i < code.length; i++) {
@@ -44,7 +43,7 @@ uiCmd.display = function(code) {
 
     var detected = utilDetect();
     var mac = (detected.os === 'mac');
-    
+
     var replacements = {
         '⌘': mac ? '⌘ ' + t('shortcuts.key.cmd')    : t('shortcuts.key.ctrl'),
         '⇧': mac ? '⇧ ' + t('shortcuts.key.shift')  : t('shortcuts.key.shift'),
--- a/modules/ui/sections/data_layers.js
+++ b/modules/ui/sections/data_layers.js
@@ -405,6 +405,7 @@ export function uiSectionDataLayers(context) {
                 .placement('top')
             );
 
+
         historyPanelLabelEnter
             .append('input')
             .attr('type', 'checkbox')


modules/ui/cmd.js
```diff
@@ -14,14 +14,14 @@ export var uiCmd = function (code) {
...
-            '⌘': 'Ctrl',
-            '⇧': 'Shift',
-            '⌥': 'Alt',
-            '⌫': 'Backspace',
-            '⌦': 'Delete'
+            '⌘': mac ? '⌘ '  : t('shortcuts.key.ctrl'),
+            '⇧': mac ? '⇧ '  : t('shortcuts.key.shift'),
+            '⌥': mac ? '⌥ '  : t('shortcuts.key.alt'),
+            '⌫': mac ? '⌫ '  : t('shortcuts.key.backspace'),
+            '⌦': mac ? '⌦ '  : t('shortcuts.key.del'),
```
1ec5, line 24: This is essentially the same as what c22fa60b8a33fdfd9dc668619a797f8c2889f9ad did in `uiCmd.display` down below to fix the Keyboard Shortcuts screen. I suspect c22fa60b8a33fdfd9dc668619a797f8c2889f9ad didn’t touch this function because it’s also being used in places where raw key identifiers are needed rather than human-readable labels. For example, the following code uses the return value to register the actual keyboard shortcut:

https://github.com/openstreetmap/iD/blob/3025d4f4217238ecd57dbb09734c9d9e93124569/modules/ui/full_screen.js#L72-L73 https://github.com/openstreetmap/iD/blob/3025d4f4217238ecd57dbb09734c9d9e93124569/modules/ui/zoom.js#L102-L110

If you run this branch on macOS, or on Windows with the German localization, will the shortcuts function correctly when you press them? If not, then maybe we need to revert this function to what it was previously. If we look back at the code I pointed to earlier, notice how it’s only being used to create the tooltip – so it’s probably only for display purposes:

https://github.com/openstreetmap/iD/blob/b5d45e3f28ccb9a39d243e7504415152e1ce8ec2/modules/ui/sections/data_layers.js#L404

If that’s the case, then changing that code to call `uiCmd.display` instead might fix the issue for that particular tooltip. So instead of implementing anything new, you’d be reviewing each of the calls to `uiCmd`, deciding whether it should be replaced with a call to `uiCmd.display`.

StillAbeginnerr: Thank very much for the detailed explanation, it's very helpful and I now understood with more clarity, gotta read more about this codebase, and I've marked the calls where uiCmd been called and now checking how useful, the uiCmd.display would be for a particular tooltip...will submit a pull request again after high detailed review of each call.

StillAbeginnerr: @1ec5 

While working, I found these 2 issues/bug

1. Issue with Key Functionality.

It seems that the keyboard shortcut for ? isn’t functioning as expected. Based on testing, it appears the shortcut should be labeled as Shift + ? instead of just ?, as pressing ? alone triggers the black overlay below the inspect button in the top-left corner.

<br>

Image 1 :

image
<br>

In Image 2, The Keyboard Shortcut should also be labeled as Shift + ?.

Image 2 : 

image



---

2. Interaction Limitation with Quality Assurance Tab
When the Quality Assurance tab is open in the Help section, the Keyboard Shortcuts option becomes unclickable. Interestingly, this behavior is not observed with other tabs, which allow seamless access to the keyboard shortcuts.

Image 1 : For example, the Keyboard Shortcuts options is clickable in case of points tab.
<br>
image


Image 2 : When the Quality Assurance tab is open in the Help section, the Keyboard Shortcuts option becomes unclickable. 
<br>
WhatsApp Image 2024-12-03 at 11 03 13


---

Checked with latest changes on both local host and online deployment.

StillAbeginnerr: image

Help/Keyboard Shortcuts.

I’ve resolved the labeling issue mentioned in the previous comment (Issue 1, Image 2) by replacing `?` with `Shift + ?` in the relevant file for "Show Keyboard Shortcuts". Should I create a separate issue and raise a PR for this fix, or is it okay to include it as part of this current work?

1ec5: > It seems that the keyboard shortcut for ? isn’t functioning as expected. Based on testing, it appears the shortcut should be labeled as Shift + ? instead of just ?, as pressing ? alone triggers the black overlay below the inspect button in the top-left corner.

I think spelling it as just “?” is a standard convention, at least on macOS. If you’re using an English keyboard layout, pressing that key alone would be <kbd>/</kbd>, not <kbd>?</kbd>. A user-selectable keyboard layout could conceivably have <kbd>?</kbd> as the primary function of a key, or it could require a modifier key other than Shift. All that matters to a Web application like iD is whether the user has pressed the some key combination that would input “?” into a text field.

1ec5: > When the Quality Assurance tab is open in the Help section, the Keyboard Shortcuts option becomes unclickable.

I can reproduce this issue – you can only click on a tiny sliver at the bottom of the button. This is probably unrelated to displaying localized keyboard shortcuts in the tooltip, so please open a separate issue about it. Thanks!

StillAbeginnerr: Thanks for reviewing! I’ll go ahead and create a separate issue for this fix and then focus on the `uiCmd` functionality.

StillAbeginnerr: > > It seems that the keyboard shortcut for ? isn’t functioning as expected. Based on testing, it appears the shortcut should be labeled as Shift + ? instead of just ?, as pressing ? alone triggers the black overlay below the inspect button in the top-left corner.
> 
> I think spelling it as just “?” is a standard convention, at least on macOS. If you’re using an English keyboard layout, pressing that key alone would be /, not ?. A user-selectable keyboard layout could conceivably have ? as the primary function of a key, or it could require a modifier key other than Shift. All that matters to a Web application like iD is whether the user has pressed the some key combination that would input “?” into a text field.

You're absolutely right—using "?" as the shortcut makes sense, especially since it's about the input rather than the exact key combination. It’s a good reminder that different keyboard layouts might handle this differently, Thanks for pointing that out. 😊

StillAbeginnerr: _reference undo_redo.js_

image

WhatsApp Image 2024-12-04 at 18 10 59


Line 4 : 
if we check by using console.log(uiCmd('⌘Z'));  // Outputs either '⌘Z' on Mac or 'Ctrl+Z' on Windows/Linux

and now if we replace that with console.log(uiCmd.display('⌘Z')); // Outputs either '⌘Z' on Mac or '⌘Z' on Windows/Linux

Now, I observed that uiCmd outputs 'CTRL+Z' on German localization or any other regional language, like in case of german it should be STRG+Z when using uiCmd.display but this function shows Outputs either '⌘Z' on Mac or '⌘Z' on Windows/Linux.

It looks like there's some inconsistency with uiCmd.display when handling localization for keyboard shortcuts. While uiCmd correctly adapts to regional differences (e.g., showing "STRG+Z" in German locales) on applying t('shortcuts.key.ctrl') there, uiCmd.display seems to bypass this behavior and hardcodes "⌘Z" for both Mac and Windows/Linux.

I tried to fix this bug using the localization function t(), and it worked with uiCmd but not with uiCmd.display. I'm a bit unsure whether using uiCmd.display for this undo action is a bad idea, or if it's an exception due to hardcoding, or if there's something else I'm missing.

@1ec5 I'm kind of stuck on this, the problem I'm facing is like whether there's need for a smaller fix in the uiCmd function, not sure if its a exceptional case or I should think from a new implementation perspective which can be very challenging for a new contributor like me as I should be reviewing calls to these and deciding which function to be used ( uiCmd or uiCmd.display )

1ec5: How are you testing Windows or Linux? Are you testing on an actual Windows or Linux machine, spoofing the user agent string, or something else?

StillAbeginnerr: I use macOS 15.1.1 to test how things work on Windows and Linux by spoofing the user agent string. When I need to double-check or confirm any issues or assumptions, I rely on a real Windows 10 PC of my friend.

1ec5: The Keyboard Shortcuts pane has already been using `uiCmd.display`; is it also showing macOS key names on Windows or Linux?

StillAbeginnerr: Checked throughly, there are no signs of macOS key names on Windows and Linux.

But since, the keyboard shortcuts pane been using uiCmd.display and there's no problems with it, 

---

I see, it could be just a unique case for undo_redo.js because German localization is not happening on this part,

As we can see there's CTRL + Z on Linux System(I used Parallels Desktop to set up Debian OS) on Firefox Browser which should have been STRG + Z.

image


   console.log('Detected OS:', mac ? 'mac' : 'not mac');   // Works Correctly
  console.log('Detected OS:', utilDetect().os);        // Works Correctly
    console.log('Current Language:', detected.language);  
    console.log('Translated CTRL:', t('shortcuts.key.ctrl'));   // Works Correctly
    console.log('uiCmd: ⌘Z is', uiCmd('⌘Z'));   // Works Correctly except Translation.
    console.log('uiCmd.display: ⌘Z is', uiCmd.display('⌘Z'));   // not working properly because of code.length issues otherwise would have displayed STRG + Z in German localization.
    
  In Cmd.js 
    
   So I think that  this line `if (code.length !== 1) return code;` returns the non-translated value because for ('⌘Z') length is 2 and replacement doesn't take place in uiCmd.display in cmd.js
   
   But If I change it to like  `if (code.length !== 2) return code;` 
   [the stack calls exceed](init.js:501 RangeError: Maximum call stack size exceeded) and crashes the webpage.

---

It uses the uiCmd Function right now in undo_redo.js, but if I want to localise it using uiCmd.display,  For Mac/Windows/Linux, ⌘Z appears.

My LocalHost :

image

StillAbeginnerr: So I concluded that, 

I encountered a problem caused by  `if (code.length !== 1) return code; `  being inefficient for Shortcut key translation of Toolbars like Undo, redo if we're to replace uiCmd by uiCmd.Display for translation since uiCmd don't support t() as some function don't have to require t() and uiCmd.display is not good for these as they don't return length in 1.

These toolbars like undo, redo and else which show Mac commands in Windows/Linux have to be hardcoded like 
`  if (detected.os === 'win') {
        if (code === '⌘Z') return t('Ctrl') + 'Z';
    }` in uiCmd function otherwise, I believe as per my experience, a lot of structural changes are to be done. 

@1ec5

StillAbeginnerr: image

I thought something like this little change may help but yeah didn't.

matkoniecz:
This PR seems abandoned, right? With "a lot of structural changes are to be done" conclusion?

matkoniecz:
I will close it as abandoned - unless it is not?

pinging @StillAbeginnerr

matkoniecz:
sadly, closing as abandoned

matkoniecz:
if not abandoned: ping me and I will reopen, or make a new PR linking this one in description

StillAbeginnerr:
I had to abandon it because I was not sure how to proceed forward and maybe asked on slack for help as well that time, although loved contributing here.

PR #10581, opened 2024-12-04
Update help panel navigation buttons

### LINKED ISSUE #10573: Keyboard Shortcuts option becomes unclickable when the Quality Assurance tab is open in the Help section.
### URL

https://ideditor.netlify.app/#background=Bing&disable_features=boundaries&map=2.00/0.0/0.0

### Description
When the "Quality Assurance" tab is open in the Help section, the "Keyboard Shortcuts" option becomes unclickable. This issue does not occur with other tabs in the Help section, where the "Keyboard Shortcuts" option remains accessible without any problems.

### Steps to Reproduce
Open the Help section.
Navigate to the "Quality Assurance" tab.
Attempt to click the "Keyboard Shortcuts" option.

### Expected Behavior
The "Keyboard Shortcuts" option should remain clickable and open the keyboard shortcuts panel regardless of the tab currently open in the Help section.

### Actual Behavior
The "Keyboard Shortcuts" option becomes unclickable when the "Quality Assurance" tab is active.

Image 1 : For example, the Keyboard Shortcuts options is clickable in case of points tab.
<br>
image

Image 2 : When the Quality Assurance tab is open in the Help section, the Keyboard Shortcuts option becomes unclickable. 
<br>
WhatsApp Image 2024-12-03 at 11 03 13

Image 3 :
image

_No response_

Chrome, Safari

ForgottenHero:
Fixes #10573  where the empty nav button for 'next' was overlapping the keyboard shortcuts button when the user was currently view the Quality Assurance page.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4184,12 +4184,6 @@ li.issue-fix-item button:not(.actionable) .fix-icon {
     width: 100%;
 }
 
-.ideditor .help-pane .nav::after {
-    content: '';
-    display: table;
-    clear: both;
-}
-
 .help-pane .nav a {
     flex: 0 0 50%;
     text-align: center;
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4179,11 +4179,19 @@ li.issue-fix-item button:not(.actionable) .fix-icon {
 .help-pane .nav {
     position: relative;
     padding-bottom: 30px;
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+}
+
+.ideditor .help-pane .nav::after {
+    content: '';
+    display: table;
+    clear: both;
 }
 
 .help-pane .nav a {
-    float: left;
-    width: 50%;
+    flex: 0 0 50%;
     text-align: center;
 }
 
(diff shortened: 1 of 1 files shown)


css/80_app.css
```diff
@@ -4179,11 +4179,19 @@ li.issue-fix-item button:not(.actionable) .fix-icon {
...
     padding-bottom: 30px;
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+}
+
+.ideditor .help-pane .nav::after {
+    content: '';
+    display: table;
+    clear: both;
```
tyrasd: What is the purpose of this `::after` block? In a quick test, for me the fix seems to work also without it.

tyrasd (commented):
Nice catch! The fix looks good, and using a flex solution instead of floating elements is definitely an improvement in general for this ui element. Only one question below.

css/80_app.css
ForgottenHero: My mistake. I tried a clearfix hack first, but don't need it now the elements are floated instead. It's redundant now.

ForgottenHero (commented):
Removed redundant clearfix hack after I changed to flex layout instead of float-based layout.

tyrasd (approved):

Issue #10567, opened 2024-11-30
Improve search for languages with accents


Nekzuris:
### Description

Currently if a word is written with accents in Transifex, the user must include the accents to find it.
I know I could just add a synonym without accents, but that’s redundant. There are probably hundreds of missing synonyms like that.
The word should be written only once with the correct accents, and the search function should ignore them.

### Screenshots

example with French (fr):
![image](https://github.com/user-attachments/assets/5d3da06b-9e9f-4d07-ad24-53deec8cf861) ![image](https://github.com/user-attachments/assets/365c40bd-6066-45d5-8170-171ea9f0bfcf)
![image](https://github.com/user-attachments/assets/87b3342a-5751-41cf-a2eb-3b20757f38b1)

1ec5:
Diacritic-insensitive search was implemented in #8242. You may manually provide diacritic-folded synonyms, as I’ve continued to do in Vietnamese. However, the results you’re seeing have a different cause: the word “pétanque” is such a small part of the overall preset name “Terrain de boules (Pétanque, Bocce…)” that other, shorter presets have a much shorter edit distance. If you search for “boules petanque” or even “petanque b”, this preset appears on top.

iD doesn’t expect a preset name to contain a gloss with a list of synonyms. The synonym string in your screenshot should also contain “pétanque” in the list, since iD boosts exact matches. Also note the instructions for the “Boules/Bocce Court” string:

> Translate the primary name. Optionally, add equivalent synonyms on newlines in order of preference (press the Return key).

So you can also add “Pétanque” and “Bocce” on a separate lines in the same string to ensure that iD will find and prioritize these search terms. #6139 remains open for labeling the preset with the matching synonym instead of the main preset name.

Nekzuris:
Okay, thanks for the explanation.
I don’t want to add diacritic-folded synonyms as they’re pointless in French (I’d even like to remove them all, is that a good idea?)
I didn’t write the current translation with synonyms in parentheses, but I think it’s quite useful for the user. Can I leave them and write it like this:?
```
Terrain de boules (Pétanque, Bocce...)
Terrain de Pétanque
Terrain de Bocce
```
and remove "Terrain de" from the synonym string, so only `Pétanque, Bocce, Lyonnaise`?

1ec5:
Yes, I think it’s worth a try. If I’m reading the code correctly, it’ll consider the aliases when matching; all that’s missing is showing the alias as the search result.

[...]

PR #10583, opened 2024-12-05
Fix: Remove Disambiguators from Relation Names in Membership Editor. Fixes #8759

### LINKED ISSUE #8759: Omit disambiguator from relation type in member/membership editor
[This way](https://www.openstreetmap.org/way/28119544) is a member of a few [bus routes](https://www.openstreetmap.org/relation/5366547) that are part of the [VTA](https://nsi.guide/index.html?t=transit&k=route&v=bus&tt=vta#vta-f0e5b1) network. Since there are multiple VTA networks, this network is glossed as “Santa Clara Valley Transportation Agency”, which is very long. iD lists the relation membership with a relation type so long that the relation’s actual ref and name get crowded out:

vta

NSI has a convention of putting disambiguating glosses in parentheses, and there doesn’t seem to be any mechanism for a localization to use anything other than parentheses for the disambiguator. The raw member and membership editors could strip out the disambiguator to save space. In this context, the user is more likely to know which VTA it is anyways.

https://github.com/openstreetmap/iD/blob/19c95248702bda93f78969e6623c7232462b916b/modules/ui/sections/raw_membership_editor.js#L345-L346

Deeptanshu-sankhwar:
This fixes #8759 

This PR improves the usability of the raw member/membership editor by removing disambiguators (text in parentheses, e.g., (Santa Clara Valley Transportation Authority)) from relation names. The change simplifies the display, making it less cluttered and more focused on relevant information like the relation's ref and primary name.

### DIFF
--- a/modules/ui/sections/raw_member_editor.js
+++ b/modules/ui/sections/raw_member_editor.js
@@ -193,7 +193,12 @@ export function uiSectionRawMemberEditor(context) {
                         .attr('class', 'member-entity-type')
                         .text(function(d) {
                             var matched = presetManager.match(d.member, context.graph());
-                            return (matched && matched.name()) || utilDisplayType(d.member.id);
+
+                            var fullName = (matched && matched.name()) || utilDisplayType(d.member.id);
+
+                            // Remove the disambiguator (text in parentheses) from the full name
+                            var simplifiedName = fullName.replace(/\s*\(.*?\)$/, ''); // Removes " (…)" at the end
+                            return simplifiedName;
                         });
 
                     labelLink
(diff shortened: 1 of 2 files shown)


1ec5 (commented):
This looks pretty reasonable for the case where a preset comes from NSI, which has a convention of using a disambiguator like this. We’ve recently also seen an example where the French localization uses a parenthetical for explanatory text in an ordinary preset name, but I think it should be fine in that case too: #10567.

In addition to the _membership_ editor, there’s also a _member_ editor that might need similar treatment. Consider the case where a bus stop is part of a route relation and you inspect the route relation; the list of members should label the bus stop by the truncated name too.

Deeptanshu-sankhwar:
Thanks for your feedback @1ec5 . The example of french localization shall fall within the scope of regex, so it should be fine. 

For the member editor, I have made the similar regex changes such that disambiguators are now removed from both membership editor and member editor. Kindly take a look whenever you can.

Thanks!

PR #10586, opened 2024-12-06
Render edit menu to the over-map layer. Fixes #10495

### LINKED ISSUE #10495: Street level imagery: Viewer overlaps right click menu

fahnestd:
This Fixes #10495 by rendering the menu to the `over-map` layer rather than the `supersurface`. I had to make a ref to `overMap` available which i'm not sure how to get around currently. If this is not ideal, please let me know if there are any tips on how to achieve this otherwise.

image

### DIFF
--- a/modules/ui/edit_menu.js
+++ b/modules/ui/edit_menu.js
@@ -36,7 +36,7 @@ export function uiEditMenu(context) {
 
     // offset the menu slightly from the target location
     var _menuSideMargin = 10;
-    
+
     // hardcoded values for repositioning the menu to account for the top menu height
     var _verticalOffset = 70;
 
--- a/modules/ui/edit_menu.js
+++ b/modules/ui/edit_menu.js
@@ -36,6 +36,9 @@ export function uiEditMenu(context) {
 
     // offset the menu slightly from the target location
     var _menuSideMargin = 10;
+    
+    // hardcoded values for repositioning the menu to account for the top menu height
+    var _verticalOffset = 70;
 
     var _tooltips = [];
 
@@ -226,7 +229,8 @@ export function uiEditMenu(context) {
         }
 
         var origin = geoVecAdd(anchorLoc, offset);
-
+        // adjusts for top menu height
+        origin[1] -= _verticalOffset;
         _menu
             .style('left', origin[0] + 'px')
             .style('top', origin[1] + 'px');
--- a/modules/ui/edit_menu.js
+++ b/modules/ui/edit_menu.js
@@ -6,6 +6,7 @@ import { localizer } from '../core/localizer';
 import { uiTooltip } from './tooltip';
 import { utilRebind } from '../util/rebind';
 import { utilHighlightEntities } from '../util/util';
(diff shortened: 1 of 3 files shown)


modules/ui/init.js
```diff
@@ -665,13 +666,16 @@ export function uiInit(context) {
             .triggerType(triggerType)
             .operations(operations);
 
-        // render the menu
-        context.map().supersurface.call(_editMenu);
+        // render the menu onto the overmap
+        overMap
+            // .append('div')
+            // .attr('class', 'menu-topbar-offset')
```
tyrasd: please remove the commented out lines. :relaxed:

modules/ui/edit_menu.js
```diff
@@ -226,7 +229,8 @@ export function uiEditMenu(context) {
         }
 
         var origin = geoVecAdd(anchorLoc, offset);
-
+        // adjusts for top menu height
+        origin[1] -= _verticalOffset;
```
tyrasd: instead of the hardcoded offset, we could instead read the rendered height of the toolbar like this:

```suggestion
        // adjusts for top menu height
        var _verticalOffset = parseFloat(utilGetDimensions(d3_select('.top-toolbar-wrap'))[1]);
        origin[1] -= _verticalOffset;
```

This would then automatically continue to work if the toolbar size would change in the future.

tyrasd (changes requested):
Thank you, this PR looks quite good already.

It's fine to store a local reference to the `overMap` selection inside of a module.

Please see the two comments line below with some suggestions:

modules/ui/edit_menu.js
fahnestd: Great! that's changed!

tyrasd (approved):

modules/ui/init.js
```diff
@@ -665,13 +666,14 @@ export function uiInit(context) {
...
-        context.map().supersurface.call(_editMenu);
+        // render the menu onto the overmap
+        overMap
+            .call(_editMenu);
     };
 
     ui.closeEditMenu = function() {
         // remove any existing menu no matter how it was added
-        context.map().supersurface
+       overMap
```
tyrasd: to fix unit tests (where `closeEditMenu` might be classed without a properly initialized `overMap`):

```suggestion
        if (overMap !== undefined) {
            overMap.select('.edit-menu').remove();
        }
```

Issue #3053, opened 2016-04-01
Lines around background imagery tiles


pnorman:
#1976 was about this issue when zoomed in/out, but it now happens when at 100% zoom.

![image](https://cloud.githubusercontent.com/assets/1190866/14198679/db872dc2-f791-11e5-9765-403d5d0ecc62.png)

Browser is 49.0.2623.110 (Official Build) m (64-bit) on Windows 8.1.

I'm not yet sure if this is is a regression in iD or a Chrome change.

bhousel:
Yes, this is a regression related to #2849 

The seams between tiles could happen before but it was very unlikely because of the image dimensions being rounded to whole pixels.  But now it happens a lot more visibly.

I can close the seams by scaling up the image tiles very slightly.

Issue #2849, opened 2015-11-27
iD.geo.roundCoords drops too much precision on retina


bhousel:
On retina screens ["a pixel is not a pixel is not a pixel"](http://www.quirksmode.org/blog/archives/2010/04/a_pixel_is_not.html) so our use of `iD.geo.roundCoords` to truncate all the svg pixels to integer values should be reevaluated.   

I know that in the past `iD.geo.roundCoords` has been attributed to significant performance speedup, so I want re-benchmark that and decide what to do.  We probably want to round to tenths instead of integers - but not sure whether to do that always or only if HiDPI detected.

e.g.: rounding to integer css pixels makes square buildings look weird while zooming by single mousewheel clicks:
![roundcoordinates2](https://cloud.githubusercontent.com/assets/38784/11446420/a0a35b30-9504-11e5-92dc-a0ae1570b280.gif)

pnorman:
> all the svg pixels to integer values should be reevaluated ... but not sure whether to do that always or only if HiDPI detected

Is there a difference with sub-pixel rendering and anti-aliasing on non-HiDPI devices?

bhousel:
> Is there a difference with sub-pixel rendering and anti-aliasing on non-HiDPI devices?

I did some digging on this today and made this codepen to understand better for myself.
http://codepen.io/bhousel/pen/MaNBbJ

2 triangles drawn with 1px stroke width.  The left triangle is drawn with [shape-rendering](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) `auto` (the default).  The right is drawn with `crispEdges`, which is the svg way of saying "not antialiased".   We draw everything in iD with `auto` aliasing, (except for the scale bar, which is drawn with `crispEdges`).  Then I translate these triangles around in 0.1px increments.

The left triangle slides around smoothly, and the right triangle jumps at device pixel boundaries.  I have a retina macbook pro with a non-retina secondary monitor.  So I when I run this codepen on the non-retina monitor, I see jumps at 1px, and on the retina side, I see several smallish jumps around 0.3px and 0.7px.

The important takeaway is that the loss of precision is very noticeable on both the normal and the HiDPI screens, and we should really to avoid it.  Benchmarks next...

pnorman:
> right triangle jumps at device pixel boundaries.

On my Windows 8.1 desktop with conventional monitors and Chrome, I see the straight lines jump, but the angled line continually progress.

[...]

PR #10594, opened 2024-12-11
Fix visible grid around tiles

### LINKED ISSUE #10589: Borders around tiles at specific zoom levels
### URL

https://www.openstreetmap.org/edit#map=15/48.24613/7.00700

Use this custom layer over France and change the zoom level in small steps using the mouse wheel.
`https://data.geopf.fr/wmts?service=WMTS&format=image/png&version=1.0.0&request=GetTile&style=normal&tilematrixset=PM&tilematrix={z}&tilerow={y}&tilecol={x}&layer=IGNF_ELEVATION.ELEVATIONGRIDCOVERAGE.SHADOW`

Some zoom levels are fine, some show only horizontal or vertical lines, and some show a full grid.

I’ve experienced this issue for a long time on my phone with Chrome when I enabled "Show desktop version", Firefox is fine.
But now I also experience it on my computer (on Chrome and Firefox), but only with this custom layer. Note that the layer has transparency.

image
Screenshot_20241207-012830

2.30.4

Chrome, Firefox

Nekzuris:
This properly fixes #10589 by removing the tile overlap that was causing white borders with transparent tiles.

The tile overlap was added as a dirty fix to #3053, it doesn't even fully resolved the issue anyway.
This fix uses the CSS rule `mix-blend-mode: plus-lighter`, applied only to the background layer and is only needed to fix [this bug](https://issues.chromium.org/issues/40084005) in Chrome.
A future version of Chrome might make this CSS rule unnecessary (but leaving it shouldn't cause the borders to reappear).

Tested on Chrome and Firefox with normal and fractional zoom levels, with regular and transparent layers.
Not tested on macOS Safari or mobile.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4369,9 +4369,9 @@ img.tile {
 }
 
 /* Workaround to remove visible grid around tile borders on Chrome */
-/* with 3 exceptions for 110%, 80% and 67% browser zoom */
+/* with exceptions for 110%, 90%, 80% and 67% browser zoom */
 /* https://issues.chromium.org/issues/40084005  https://github.com/Leaflet/Leaflet/pull/8891  https://github.com/openstreetmap/iD/pull/10594 */
-@media not ((1.09 < -webkit-device-pixel-ratio < 1.11) or (0.79 < -webkit-device-pixel-ratio < 0.81) or (0.66 < -webkit-device-pixel-ratio < 0.68)) {
+@media not ((1.09 < -webkit-device-pixel-ratio < 1.11) or (0.89 < -webkit-device-pixel-ratio < 0.91) or (0.79 < -webkit-device-pixel-ratio < 0.81) or (0.66 < -webkit-device-pixel-ratio < 0.68)) {
     .layer-background img.tile {
         mix-blend-mode: plus-lighter;
     }
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4367,12 +4367,16 @@ img.tile {
     opacity: 1;
     transition: opacity 120ms linear;
 }
-/* Fix visible grid around tiles borders on Chrome */
+
+/* Workaround to remove visible grid around tile borders on Chrome */
+/* with 3 exceptions for 110%, 80% and 67% browser zoom */
+/* https://issues.chromium.org/issues/40084005  https://github.com/Leaflet/Leaflet/pull/8891  https://github.com/openstreetmap/iD/pull/10594 */
 @media not ((1.09 < -webkit-device-pixel-ratio < 1.11) or (0.79 < -webkit-device-pixel-ratio < 0.81) or (0.66 < -webkit-device-pixel-ratio < 0.68)) {
     .layer-background img.tile {
         mix-blend-mode: plus-lighter;
     }
 }
+
 img.tile-removing {
     opacity: 0;
     z-index: 1;
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4367,7 +4367,10 @@ img.tile {
     opacity: 1;
     transition: opacity 120ms linear;
 }
-
+.layer-background img.tile {
+    /* Fix visible grid around tiles borders on Chrome */
+    mix-blend-mode: plus-lighter;
+}
 img.tile-removing {
     opacity: 0;
     z-index: 1;
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4369,9 +4369,13 @@ img.tile {
 }
 
 /* Workaround to remove visible grid around tile borders on Chrome */
-/* with exceptions for 110%, 90%, 80% and 67% browser zoom */
-/* https://issues.chromium.org/issues/40084005  https://github.com/Leaflet/Leaflet/pull/8891  https://github.com/openstreetmap/iD/pull/10594 */
-@media not ((1.09 < -webkit-device-pixel-ratio < 1.11) or (0.89 < -webkit-device-pixel-ratio < 0.91) or (0.79 < -webkit-device-pixel-ratio < 0.81) or (0.66 < -webkit-device-pixel-ratio < 0.68)) {
+/* Only works with browser zoom multiple of 25 (75%, 100%, 125%...) */
+/* Should be removed when https://issues.chromium.org/issues/40084005 is resolved */
+@media (-webkit-device-pixel-ratio = 1) or (-webkit-device-pixel-ratio = 1.25) or (-webkit-device-pixel-ratio = 1.5) or (-webkit-device-pixel-ratio = 1.75)
+    or (-webkit-device-pixel-ratio = 2) or (-webkit-device-pixel-ratio = 2.25) or (-webkit-device-pixel-ratio = 2.5) or (-webkit-device-pixel-ratio = 2.75)
+    or (-webkit-device-pixel-ratio = 3) or (-webkit-device-pixel-ratio = 3.25) or (-webkit-device-pixel-ratio = 3.5) or (-webkit-device-pixel-ratio = 3.75)
+    or (-webkit-device-pixel-ratio = 4) or (-webkit-device-pixel-ratio = 4.25) or (-webkit-device-pixel-ratio = 4.5) or (-webkit-device-pixel-ratio = 4.75) 
+    or (-webkit-device-pixel-ratio = 5) or (-webkit-device-pixel-ratio = 0.25) or (-webkit-device-pixel-ratio = 0.5) or (-webkit-device-pixel-ratio = 0.75) {
     .layer-background img.tile {
         mix-blend-mode: plus-lighter;
     }
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4368,9 +4368,10 @@ img.tile {
     transition: opacity 120ms linear;
 }
 
-/* Workaround to remove visible grid around tile borders on Chrome */
-/* Only works with browser zoom multiple of 25 (75%, 100%, 125%...) */
-/* Should be removed when https://issues.chromium.org/issues/40084005 is resolved */
+/* Workaround to remove visible grid around tile borders on Chrome
+   Only works with browser zoom multiple of 25 (75%, 100%, 125%...) 
+   Should be removed when https://issues.chromium.org/issues/40084005 is resolved.
+   See https://github.com/openstreetmap/iD/pull/10594 */
 @media (-webkit-device-pixel-ratio = 1) or (-webkit-device-pixel-ratio = 1.25) or (-webkit-device-pixel-ratio = 1.5) or (-webkit-device-pixel-ratio = 1.75)
     or (-webkit-device-pixel-ratio = 2) or (-webkit-device-pixel-ratio = 2.25) or (-webkit-device-pixel-ratio = 2.5) or (-webkit-device-pixel-ratio = 2.75)
     or (-webkit-device-pixel-ratio = 3) or (-webkit-device-pixel-ratio = 3.25) or (-webkit-device-pixel-ratio = 3.5) or (-webkit-device-pixel-ratio = 3.75)
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4367,9 +4367,11 @@ img.tile {
     opacity: 1;
     transition: opacity 120ms linear;
 }
-.layer-background img.tile {
-    /* Fix visible grid around tiles borders on Chrome */
-    mix-blend-mode: plus-lighter;
+/* Fix visible grid around tiles borders on Chrome */
+@media not ((1.09 < -webkit-device-pixel-ratio < 1.11) or (0.79 < -webkit-device-pixel-ratio < 0.81) or (0.66 < -webkit-device-pixel-ratio < 0.68)) {
+    .layer-background img.tile {
+        mix-blend-mode: plus-lighter;
+    }
 }
 img.tile-removing {
     opacity: 0;
--- a/modules/renderer/tile_layer.js
+++ b/modules/renderer/tile_layer.js
@@ -19,6 +19,7 @@ export function rendererTileLayer(context) {
 
     // Workaround to remove visible grid around tile borders on Chrome with dynamic epsilon for specific browser zoom levels
     // Should be removed when https://issues.chromium.org/issues/40084005 is resolved
+    // See https://github.com/openstreetmap/iD/pull/10594
     if (window.chrome) {
         updateEpsilon();
         window.addEventListener('resize', updateEpsilon);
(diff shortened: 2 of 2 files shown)


tyrasd (dismissed):
Hmm… Nice idea, but there are still corner cases where it does not work: To reproduce on Chrome 129 on Ubuntu, set browser zoom to 110%:

image

And, more importantly, now even non-transparent tile layers show the seam under the above mentioned conditions (this works fine on the development branch):

image

<hr>

I think the +epsilon workaround is still needed to account for rounding errors associated with fractional browser zoom settings.

To improve the rendering of the small overlap for tile layers with transparency, I tried to play around with a few other combinations of CSS blend modes, background values, etc. but could unfortunately not find a solution that preserves the rendering of the non transparent tiles. :unamused: 

As transparent imagery layers are a rather rare occurence[^1], I'm leaning towards closing this as an unfortunate _wontfix / known bug_. Except if you or someone is able to find a better solution?!

[^1]: some overlays use an alpha channel, but as long as it is only used as a mask ("1 bit transparency"), these also work fine with the current solution.

Nekzuris:
It is unfortunate that it doesn't work for 110% zoom in Chrome but it's fine with 125%, 150%, 175%, 200%.

The current solution with +epsilon to scale up by 1px and create an overlap is fundamentally flawed because it messes with tiles sharpness and will never work for transparent layers.
It's also only needed to fix a rendering bug in Chrome and doesn't even completely fix the issue at regular zoom levels and 100% browser zoom: https://www.openstreetmap.org/edit#map=16/48.36450/7.14242&background=MAPNIK
image

Value 90% and 80% are also problematic with both epsilon or plus-lighter, but 75% is fine.
So with my testing, the +epsilon only improves the situation for 110% in Chrome.
While removing it in favor of plus-lighter :
- fixes everything at 100%
- makes perfect zooming transitions
- supports transparent layers
- but slightly degrades experience at 110% in Chrome

The solution `mix-blend-mode: plus-lighter` was also considered the best by [Leaflet#8891](https://github.com/Leaflet/Leaflet/pull/8891)

I found that not applying plus-lighter at 110% improves the situation, as well as 80% and 67%.
90% is still bad in Chrome and might be improved with specific epsilon value in JS, but I don't think that's a good idea.

Nekzuris:
I tried to compile in a comprehensive way everything I've tested to remove the grid:
`0.002 normal` is the current solution
`0 plus-lighter` is the Leaflet solution
`0 normal` is an exception (only included if plus-lighter isn't perfect)

### Firefox
| Browser zoom | epsilon | mix-blend-mode | normal zoom | fractional zoom | zoom transition | transparent layer |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| any | 0.002 | normal | ✔️ | ✔️ | ❌ | ❌ |
| any | 0 | normal or plus-lighter | ✔️ | ✔️ | ✔️ | ✔️ |

### Chrome
<table><thead>
  <tr>
    <th>Browser zoom</th>
    <th>epsilon</th>
    <th>mix-blend-mode</th>
    <th>normal zoom</th>
    <th>fractional zoom</th>
    <th>zoom transition</th>
    <th>transparent layer</th>
  </tr></thead>
<tbody>
  <tr>
    <td rowspan="3">50%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.006</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
  <tr>
    <td rowspan="3">67%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>✔️</td>
    <td>〰️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.003</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter or normal *</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td rowspan="3">75%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>〰️</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.004</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
  <tr>
    <td rowspan="4">80%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.003</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>normal</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>〰️</td>
  </tr>
  <tr>
    <td rowspan="3">90%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.005</td>
    <td>normal</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter or normal</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td rowspan="3">100%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.003</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
  <tr>
    <td rowspan="3">110%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>normal</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>〰️</td>
  </tr>
  <tr>
    <td rowspan="3">125%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>〰️</td>
    <td>〰️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0.003</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
  <tr>
    <td rowspan="2">150-400%</td>
    <td>0.002</td>
    <td>normal</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>❌</td>
    <td>❌</td>
  </tr>
  <tr>
    <td>0</td>
    <td>plus-lighter</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
</tbody></table>
* normal is a bit better 
〰️ means okayish

#
TLDR: This PR with 3 CSS exceptions is a clear improvement over the current solution, the only annoying regression is 110% in Chrome but it's minor. We could go further with more logic in JS to keep epsilon in some cases but I don't like it.

Nekzuris:
I tried to compile in a comprehensive way everything I've tested to remove the grid:
`0.002 normal` is the current solution
`0 plus-lighter` is the Leaflet solution
`0 normal` is an exception (only included if plus-lighter isn't perfect)

### Firefox
| Browser zoom | epsilon | mix-blend-mode | normal zoom | fractional zoom | zoom transition | transparent layer |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| any | 0.002 | normal | ✔️ | ✔️ | ❌ | ❌ |
| any | 0 | normal or plus-lighter | ✔️ | ✔️ | ✔️ | ✔️ |

### Chrome
image

\* normal is a bit better 〰️ means okayish
(image because the html table broke the markdown, message wouldn't send)

#
TLDR: This PR with 3 CSS exceptions is a clear improvement over the current solution, the only annoying regression is 110% in Chrome but it's minor. We could go further with more logic in JS to keep epsilon in some cases but I don't like it.

Nekzuris:
### Visual before vs after

current zoom 100%

https://github.com/user-attachments/assets/ee6b6985-19d1-4dca-bf47-b5e3b0cef19f

new zoom 100%

https://github.com/user-attachments/assets/38562223-9b32-47af-bd0d-cc8d01687640

current zoom 110%

https://github.com/user-attachments/assets/932d8c99-e981-4dbf-a02c-1dfdf4e294e5

new zoom 110%

https://github.com/user-attachments/assets/85e079e1-ee51-4210-b871-9f71e907cf38

I chose Carto because it's where the grid is most noticeable.

### Here is what it looks like with real imagery at zoom 18

current vs new 100%

(click for full resolution)

current vs new 110%

The thin black lines currently visible at 100% are now at 110%.

I think most users will only really notice and appreciate the new flawless zoom transition.

tyrasd:
Thanks for the in-depth analysis and detailed tests!

> The current solution with +epsilon to scale up by 1px and create an overlap is fundamentally flawed because it messes with tiles sharpness and will never work for transparent layers.

Good point about the loss in sharpness with the existing workaround. I'm tending towards accepting the `0 plus-lighter` implementation, but what do you think about using a dynamic epsilon? That is only set to a non-zero value when the workaround is really "needed": i.e. Chromium-based browser and `window.devicePixelRatio` is one of the "problematic" values.

Nekzuris:
I did my initial test with all the logic in the `tileSizeAtZoom` function, but it had a noticeable performance impact. I thought this function was only called for rendering tiles, but it's called hundreds of times even when the data layer is hidden, not sure if that's normal.

This is a better implementation, but maybe you have an even better idea.
And unlike the exceptions in CSS, this will still prevent good zooming transitions and transparent layers when the [chromium bug](https://issues.chromium.org/issues/40084005) is fixed.

Nekzuris:
Is the failed test caused by the EventListener ?
It's fine on my end.

Nekzuris:
While testing for Leaflet [here](https://github.com/Leaflet/Leaflet/pull/9560#issuecomment-2543372689), I found that _plus-lighter_ can introduce some blur on Chrome, but it's nothing to worry about because iD is already blurry all the time due to sub-pixel translations.
Sub-pixel precision is apparently needed (see #2849), and trying to fix it for the map would create a noticeable sharpness difference when switching between normal and fractional zoom levels.

Again, this is only an issue on Chrome, Firefox handles both sub-pixel translations and plus-lighter just fine.

Nekzuris:
After further testing with non-standard pixel ratios, I found that the _plus-lighter_ trick only works with zoom levels that are multiples of 25 (75%, 100%, 125%...).

So instead of a few exceptions, I changed it to be applied only at those specific pixel ratios where it’s been confirmed to work well.
And for all the other case we still use the dynamic epsilon with 0.003 by default.

tordans:
Just to put this out there: We could also decide to not support zoomed in UIs and leave it at that. 
Or, consider adding a notice for zoomed in UIs for those users that zoom in but forgot about it.
TBH I am not sure it the added complexity of the workarounds and their added maintenance burden are worth it compared to the issue and how many users are likely affected (not many).

Nekzuris:
I originally made this PR to support transparent tiles and improve zoom transitions (which are currently awful for everyone), and fix the thin black lines visible with light backgrounds on Chrome.

The situation seems complex due to all the incremental improvements, but I think the final solution isn't that hard to understand.

If you ignore all other pixel ratios, simply remove _epsilon_ and apply _plus-lighter_ with no condition. That would improve things for the majority, but it would certainly be a regression for some.

There are many valid reasons to not have your zoom set to 100%, especially with high DPI screens, which are much more common now.

This mess is entirely caused by poor rendering in Chrome, and the only maintenance I foresee is removing all this when the [bug](https://issues.chromium.org/issues/40084005) is fixed.

css/80_app.css
```diff
@@ -4368,6 +4368,19 @@ img.tile {
     transition: opacity 120ms linear;
 }
 
+/* Workaround to remove visible grid around tile borders on Chrome */
+/* Only works with browser zoom multiple of 25 (75%, 100%, 125%...) */
+/* Should be removed when https://issues.chromium.org/issues/40084005 is resolved */
```
tyrasd:
```suggestion
/* Workaround to remove visible grid around tile borders on Chrome
   Only works with browser zoom multiple of 25 (75%, 100%, 125%...) 
   Should be removed when https://issues.chromium.org/issues/40084005 is resolved.
   See https://github.com/openstreetmap/iD/pull/10594 */
```

modules/renderer/tile_layer.js
```diff
@@ -15,11 +15,29 @@ export function rendererTileLayer(context) {
     var _tileOrigin;
     var _zoom;
     var _source;
+    var _epsilon = 0;
 
+    // Workaround to remove visible grid around tile borders on Chrome with dynamic epsilon for specific browser zoom levels
+    // Should be removed when https://issues.chromium.org/issues/40084005 is resolved
```
tyrasd, line 21:
```suggestion
    // Should be removed when https://issues.chromium.org/issues/40084005 is resolved
    // See https://github.com/openstreetmap/iD/pull/10594
```

tyrasd (approved):
Thanks again! This seems to be the best we can do for now.

Below, I've added a few inline links to this PR for future reference.

PR #10595, opened 2024-12-11
#10516  added ellipsis if more than 5 warnings occurs

### LINKED ISSUE #10516: Too many warnings hinders save
### Description

OK, we made an edit, pressed Save, entered our changeset comment, and...
Let's say there are 1000 warnings. I.e., the scrollbar indicator "A" is even smaller.
And say those warnings are all for bus routes and we have never rode a bus in our lives and they certainly are not our fault.

So we want to hurry up and find the final Upload button, but we must scroll past all those warnings first.

Yes we could grab the scrollbar and force it down.

But we don't edit often and are not sure that there is even a final Upload button at the end of all those warnings.

We try the X (B) but that just cancels our edit!

One solution would be a floating final Upload button, always visible.

### Screenshots

Screenshot 2024-10-26 17 31 38

AkashNegi1:
added ellipsis if more than 5 warnings occurs it will create an ellipsis which hold the additional warnings and we can use the final upload button without any warnings disturbances
Screenshot 2024-12-11 094011
Screenshot 2024-12-11 094028

fixes https://github.com/openstreetmap/iD/issues/10516

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
         "@rapideditor/mapillary_sprite_source": "^1.8.0",
         "@rapideditor/temaki": "^5.9.0",
         "@transifex/api": "^7.1.3",
-        "@types/chai": "^5.2.0",
+        "@types/chai": "^5.2.1",
         "@types/d3": "^7.4.3",
         "@types/geojson": "^7946.0.16",
         "@types/happen": "^0.3.0",
@@ -1668,9 +1668,9 @@
       }
     },
     "node_modules/@types/chai": {
-      "version": "5.2.0",
-      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-5.2.0.tgz",
-      "integrity": "sha512-FWnQYdrG9FAC8KgPVhDFfrPL1FBsL3NtIt2WsxKvwu/61K6HiuDF3xAb7c7w/k9ML2QOUHcwTgU7dKLFPK6sBg==",
+      "version": "5.2.1",
+      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-5.2.1.tgz",
+      "integrity": "sha512-iu1JLYmGmITRzUgNiLMZD3WCoFzpYtueuyAgHTXqgwSRAMIlFTnZqG6/xenkpUGRJEzSfklUTI4GNSzks/dc0w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "@rapideditor/mapillary_sprite_source": "^1.8.0",
     "@rapideditor/temaki": "^5.9.0",
     "@transifex/api": "^7.1.3",
-    "@types/chai": "^5.2.0",
+    "@types/chai": "^5.2.1",
     "@types/d3": "^7.4.3",
     "@types/geojson": "^7946.0.16",
     "@types/happen": "^0.3.0",


modules/ui/commit_warnings.js
```diff
@@ -92,12 +97,72 @@ export function uiCommitWarnings(context) {
...
+                (function(localContainer) {
+
+                    var ellipsisItem = localContainer.select('ul').selectAll('.ellipsis')
+                        .data([0]);
+
+                    ellipsisItem.enter()
+                        .append('div')
+                        .attr('class', 'ellipsis')
+                        .append('button')
+                        .text('...More warnings')
```
tyrasd: Labels should be translatable. See https://github.com/openstreetmap/iD/wiki/How-to-Use-iD%27s-Localizer-Module for a tutorial for how the localizer module is used.

AkashNegi1:
@tyrasd can you tell me what I should do next or guide me further

tyrasd (changes requested):
Generally, iD uses chevron icons for collapsed/collapsible UI sections. It would be good to reuse this design pattern here.

image

In the iD code this is called a _disclosure_ (see `modules/ui/disclosure.js`).

Maybe it would be most consistent to just use a similar collapsible list of warnings in the commit section as in the validation panel:

image

See `modules/ui/sections/validation_issues.js` for how that was implemented.

What do you think?

--

PS: can you please exclude the unrelated changes in the `data` directory from this PR?

modules/ui/commit_warnings.js
AkashNegi1: yup! i think that dropdown kind of UI looks better I'll try to fix that and also try to fix those unrelated changes

AkashNegi1: image
I've made the changes which you told me to

tyrasd (changes requested):
Thanks! A few more things need to be adjusted from a styling perspective:

* to restore the whitespace/margins around the section: please wrap the selection in a div with css class `modal-section` (see how this `container` was set up in the previous implementation)
* also, the yellow/red background color should be still applied for the respective warning/error sections (css class `.warning-section`/`.error-section`). Alternatively, we could change the styling to how it looks like in the validation sidebar (colored individual warning lines, see screenshot above). I don't have a strong preference between these two options.

AkashNegi1:
image
I've made the warning section look like this -^
@tyrasd need your review

tyrasd (changes requested):
I don't like that the styling of the list is now suddenly more squished together. Please restore the previous css class of the `<ul>` to `changeset-list`, then it should be fine to merge.

AkashNegi1:
restored the css class of ul as it was before (i.e. to change-list)
Screenshot 2024-12-18 222018
@tyrasd I've updated the CSS class Could you please check it out

modules/ui/commit_warnings.js
```diff
@@ -1,103 +1,113 @@
...
-                    return d.message(context)(d3_select(this));
-                });
+            // Add the appropriate class for styling based on severity
+            selection
+                .call(section.render)
+                .classed(severity + '-section', true);
         }
     }
 
+    function renderIssuesList(selection, severity, issues) {
```
tyrasd: The function argument `severity` does not seem to be used inside the function. Please drop it.

modules/ui/commit_warnings.js
```diff
@@ -1,103 +1,113 @@
...
+        selection.selectAll('.issues-list').remove();
```
tyrasd: While this could work (if the correct css class was referenced in the d3 select :sweat_smile: ), the approach to recreate the list from scratch every time the list needs to be rendered is not the best. See the suggestion in the next comment for a solution that does actually use a d3 data binding to make sure the `ul` is only created if needed, and an existing `ul` is reused if already present.

modules/ui/commit_warnings.js
```diff
@@ -1,103 +1,113 @@
...
+        var container = selection
+            .append('ul')
+            .attr('class', 'changeset-list');
+
+        container.exit().remove();
```
tyrasd:
```suggestion
        let container = selection
            .selectAll('.changeset-list')
            .data([0]);

        container = container.enter()
            .append('ul')
            .attr('class', 'changeset-list')
            .merge(container);
```

Notice the `.data([0])` here: this assigns a constant `0` as a kind of _dummy data_ to the list: this results in the list only to only be created once (in `.enter().append(…)`) and only if the selection is empty.

modules/ui/commit_warnings.js
```diff
@@ -1,103 +1,113 @@
...
+        selection = selection.append('div').attr('class', 'modal-section');
+
+        // Load issues by severity
+        _issuesBySeverity = context.validator()
             .getIssuesBySeverity({ what: 'edited', where: 'all', includeDisabledRules: true });
 
-        for (var severity in issuesBySeverity) {
-            var issues = issuesBySeverity[severity];
+        for (let severity in _issuesBySeverity) {
+            let issues = _issuesBySeverity[severity];
```
tyrasd: good idea to refactor some of the old code using `var`. Could you please also use `const/let` instead  of `var` in the remaining new code of this PR?

modules/ui/commit_warnings.js
```diff
@@ -1,103 +1,113 @@
...
+                        'inspector.title_count',
+                        { title: t('issues.' + severity + 's.list_title'), count: count }
+                    );
                 })
-                .on('mouseout', function() {
-                    context.surface().selectAll('.hover')
-                        .classed('hover', false);
+                .disclosureContent(function(selection) {
+                    return renderIssuesList(selection, severity, issues);
                 })
```
tyrasd: this could be an arrow function to be more concise and consistent with the `.label()` property a few lines above 

```suggestion
                .disclosureContent(selection => renderIssuesList(selection, ~~severity,~~ issues))
```

same also for the `shouldDisplay` callback and several ones called in `renderIssuesList`.

tyrasd (changes requested):
Thanks, optically it looks fine now. There is a small regression caused by the changed css class in the last commit: the list is not duplicated when collapsing the section and un-collapsing it again. See below for a suggestion to fix the bug and at the same time improve the implementation slightly.

Additionally, I included some general code style suggestions below.

modules/ui/commit_warnings.js
AkashNegi1: i've tried all of the above things but there's problem occuring 
image
when i hit save commitWarnings() function is called twice and when click anywhere in the screen again it called and new sections of warnings (duplicate sections) are being added at the bottom 
@tyrasd can you help me with this

AkashNegi1: applied all the requested changes
image
@tyrasd need to review

AkashNegi1: @tyrasd if you are not busy could you please review it

PR #10615, opened 2024-12-20
automatically download adjacent ways when splitting a line

### LINKED ISSUE #10316: "Download the full relation" error message not actionable
### URL

https://www.openstreetmap.org/edit?node=5290940#map=19/51.78760/-1.28748

Try and split way (which is part of a large relation)
The little icon that would allow you to split the way is greyed out.  Some transient text flashes on and off the screen saying "this line cannot be split".  It also says something like "Download the full relation".  No indication how to do that is provided.

Screenshot_20240711_094127

2.29.0

Firefox

### LINKED ISSUE #10469: Very long Road Relation Split NOT Possible if not all elements are downloaded
### URL

https://www.openstreetmap.org/edit?node=12187466479#map=19/-17.606368/-65.670977

I tryed to split a road to include a bridge but could not do it because ID requires to have all elements of the road (trunk) downloaded. How all elements of a road relation could be downloaded if they cover dozens of kilometers?

This have NOT SENSE AT ALL. Who invented this Limitation?

Refer to the links above

_No response_

2.30.2

Microsoft Edge

### LINKED ISSUE #11601: iD is very slow after downloading big relations
### URL

https://www.openstreetmap.org/edit#map=18/-3.762197/12.134522

1. follow the link above
2. select the way then move a node in different directions : the movement is very fluid
3. at a node, try to cut the way : you cannot because relations are not loaded
4. download both relations (wood and grassland)
5. select a node and try to move it : the movement is jerky, the point has difficulty following the mouse. It means that all next changes need a lot of time to do them, iD is very slow and almost impossible to use.

Could you improve that ?

My suggestion : after the download of all relations that we have to download to do an action (to cut a way for example), it would be nice to be able to unload relations (I mean to clean the memory except our changes and what is visible on the screen as before the downloads).

_No response_

2.37.3

Firefox

k-yle:
Closes #10469, Closes #10316, Closes #11601

This error message has been removed: 

image

and replaced with a popup that allows you to download the adjacent relation members and then continue the split operation:

image


k-yle:
I think a nicer solution would be to make the popup appear instead of this error, so that splitting a line is a more streamlined process.

image

But I thought I'd open a PR first to get feedback

modules/ui/sections/raw_membership_editor.js
```diff
@@ -205,15 +206,64 @@ export function uiSectionRawMembershipEditor(context) {
...
 
-    function downloadMembers(d3_event, d) {
+    async function downloadMembers(d3_event, d) {
         d3_event.preventDefault();
         const button = d3_select(this);
 
-        // display the loading indicator
-        button.classed('loading', true);
-        context.loadEntity(d.relation.id, function() {
+        if (d.relation.members.length > 250) {
```
1ec5: > If the relation has more than ~300 members, clicking the button will crash iD.

In my experience, 300 members isn’t that extreme and definitely wouldn’t risk a crash in general. [A mid-sized U.S. state](https://www.openstreetmap.org/relation/162061) or [a statewide bike trail](https://www.openstreetmap.org/relation/118953) would easily surpass this threshold. A confirmation dialog wouldn’t hurt, but the wording doesn’t need to be so certain and alarming. If this prevents the user from downloading the full relation, that could be unfortunate. Is there any way we could offer a third option to download the full relation at your own risk?

[Mainland Australia](https://www.openstreetmap.org/relation/4095122)[^boundary] has 3,583 members, which is a lot, but downloading it works for me without crashing in Firefox 134. It’s just very slow. I regularly work with [large boundary relations](https://www.openhistoricalmap.org/relation/2798718) in OpenHistoricalMap’s fork of iD that don’t incur any noticeable performance penalty. The performance issue is probably a function of the total number of nodes rather than the number of members, and maybe other factors too. For example, [this multipolygon](https://www.openhistoricalmap.org/relation/2746142), which represents some marshes on the outskirts of San José, California, has only three members, but drawing an area that intersects it slows down iD even more than mainland Australia for me.

[^boundary]: Why the heck is there a `boundary` relation for a continent?

k-yle: yeah, I wonder if this is the total wrong approach. #10316 is the root of the problem, which this doesn't even't solve. This is more of a hack so I can edit my local beach 😅

I'm starting to think that we need a different popup which interrupts the split operation, and offers to download the prev+next members. And then we leave this download button alone.

k-yle: okay, I've completely the changed the approach, so it directly solves #10469 by interrupting the split operation, rather than the <kbd>Download Full Relation</kbd> button

k-yle: I'm going to mark this thread as resolved since most of the discussion is now out-of-date, the approach of this PR has been completely changed

tyrasd:
in the future this functionality could potentially made more generic and reused for things like #9020

k-yle:
all the conflicts from 4dc4e7d and e25af7c should be fixed now

Mahabarata:
I don't know if it is a new issue or if it is ok to show a strange case in this one.

- follow this link : https://www.openstreetmap.org/edit?editor=id#map=20/43.0796653/5.7613228
- select the way at the bottom : you can see that there are 16 relations, no one is loaded
- right click on a node of this way :  you can split the way
- now select the way at the top of the screen : you can see there are 16 relations, no one is loaded
- right click on a node of this way :  you cannot split the way

Why is it different?

matkoniecz:
> I don't know if it is a new issue or if it is ok to show a strange case in this one.

have you tested it with code on this branch?

matkoniecz:
> Closes https://github.com/openstreetmap/iD/issues/11601

maybe other issue was supposed to be linked?

matkoniecz:
I am confused, see https://github.com/openstreetmap/iD/issues/10469#issuecomment-4010836797

It can be split in unmodified iD without downloaded relations.

With this PR it is also split, without relations being downloaded.

k-yle:
> It can be split in unmodified iD without downloaded relations.

Nope, it's still broken and still [frustratin](https://discord.com/channels/413070382636072960/1028917673315017891/14380633121149607869)[g people](https://github.com/openstreetmap/iD/issues/10316#issuecomment-2363039101) who have been unable to edit `natural=coastline` for over 2 years.


This issue only affects ways where the next relation member is so far away that it hasn't been downloaded yet. That's why it affects `natural=coastline` often, but rarely affects route relations.


For example, try and split at [this node](https://www.openstreetmap.org/edit?node=1899717674#map=19/-36.075437/173.850642):

image

You can't. And if you follow iD's instruction and try to download [r6571744](https://osm.org/relation/6571744), it will crash your browser because the relation is so big

k-yle:
> > Closes https://github.com/openstreetmap/iD/issues/11601
>
> maybe other issue was supposed to be linked?

True, this doens't directly solve that issue, but this PR means that you hardly ever need to click <kbd>Download full relation</kbd> button, so iD will feel faster for people who are currently forced to click the button

tyrasd (approved):

PR #10618, opened 2024-12-20
enable IDE auto-completion for the main classes

k-yle:
With only a few lines of changes + some JSDoc comments, we can significantly improve the DX in this codebase by enabling autocompletion/"intellisense" in most IDEs.

Currently, common classes like `context` and `osmEntity` are typed as `any`, so your IDE can't help you. With this PR, we get autocomplete for most of the class methods:

image

&nbsp;

We can also reference common classes from JSDoc comments, such as `iD.Context` or `iD.Graph`:

```js
/** @param {iD.Context} context */
function whatever(context) { 
  // `context` now has auto-complete 
}
```


modules/core/context.js
```diff
@@ -25,7 +25,7 @@ import { utilKeybinding, utilRebind, utilStringQs, utilCleanOsmString } from '..
 
 export function coreContext() {
   const dispatch = d3_dispatch('enter', 'exit', 'change');
-  let context = utilRebind({}, dispatch, 'on');
+  const context = {};
```
k-yle, line 28: This works because JS autocompletion is powered by TypeScript, which is smart enough to infer the type of `context` when defined as an empty object

[proof here](https://www.typescriptlang.org/play/?filetype=js#code/MYewdgzgLgBAHjAvDA3gXwNwCgtwHQCGSMAFAJRIB8MAjLngEbHlUwBEb2WMP+BWAegE8YAPQD8g4bwDkEGFACeABwCmAE1UAzGAEt5usFtUAnExpw4h1m7bv2HNy1i0BXMMCi7wMAO4ALAihVADdTFhQYcyhXEzBUNBg0HFBIWEViAKDQ8LJsIRExSSxFQmYKRGo6UqZkFkr2ThwRUv4CngkpFrkFFQ1tGGACMBlYBlU9I1NzdRgsIA)

utilRebind can safely be moved to the `return` statement because this class does not use `context.on` internally.

modules/osm/node.js
```diff
@@ -33,7 +37,7 @@ osmEntity.node = osmNode;
 
 osmNode.prototype = Object.create(osmEntity.prototype);
 
-Object.assign(osmNode.prototype, {
+const prototype = {
```
k-yle, line 40: this is required so we can use `typeof prototype` in the JSDoc comment above. No runtime impact.

modules/util/rebind.js
```diff
@@ -1,8 +1,16 @@
...
+/**
+ * @template T
+ * @template S
+ * @template {keyof S} Args
+ * @param {T} target
+ * @param {S} source
+ * @param {...Args} args
+ * @returns {T & Pick<S, Args>}
+ */
+export function utilRebind(target, source, ...args) {
```
k-yle, line 11: this had to be refactored to use [spread arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) instead of the [`arguments` keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments), because JSDoc doesn't support the `arguments` keyword.

there's a new unit test to confirm that [`this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) is still correctly bound.

k-yle (commented):
some comments to explain why this works

tordans:
Very nice!

tyrasd (approved):
Great! Thanks

PR #10619, opened 2024-12-20
store history in the OPFS instead of localStorage to bypass 5 MB size limit

### LINKED ISSUE #3239: Switch to async storage (i.e. indexdb instead of localstorage)

### LINKED ISSUE #11160: Migrate history storage from localStorage to IndexedDB and add async preferences API
### Description

Currently, iD uses `localStorage` for session data which has a hard 5MB limit, causing data loss and failed saves during extended mapping sessions or when working with large datasets. This storage constraint forces users to frequently restart their sessions and lose work progress, significantly impacting the mapping experience for power users and complex editing workflows.

This issue proposes migrating to `IndexedDB` which provides `50-100MB+` storage capacity (50x+ increase), asynchronous non-blocking operations, and atomic transactions. The implementation will include automatic migration of existing `localStorage` data, graceful fallbacks, and a new `asyncPrefs API` alongside the existing synchronous interface to maintain backward compatibility while enabling future server synchronization capabilities.

### Screenshots

_No response_

k-yle:
Closes #3239, Closes #11160

localStorage has a 5 MB size limit, where as the [Origin private file system](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) or [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) supports at least 50 MB if you have enough hard-drive space.

We still store a single boolean in localStorage because we need to synchronously check if there are unsaved changes.


tordans:
JFYI there is [some talk](https://github.com/facebook/Rapid/issues/1078) about migrating to IndexedDB on the Rapid side ([search results](https://github.com/search?q=repo%3Afacebook%2FRapid+localStorage&type=issues)).

k-yle:
Following a comment on slack, i realised i haven't posted any updates here...

IndexedDB turned out to be much more performant than the OPFS (anecdotally; I haven't done a proper analysis), because it means we can get rid of [this slow `JSON.stringify` call](https://github.com/openstreetmap/iD/pull/10619/files#diff-581667dc79034bf2c51fd303b3097f86cbdc53df347f03d92b719dba298f7cbdL507), since objects can be directly written to IndexedDB. 

That `JSON.stringify` is currently one of the main performance bottlenecks.

tyrasd:
Superseeded by #11141 (which is essentially the same as this PR with a few minor tweaks, and unit tests).

PR #10630, opened 2024-12-24
Fix mouse cursor on textarea scrollbars

### LINKED ISSUE #9507: Wrong cursor over scrollbar
### URL

_No response_

In the Tags window, place the cursor (mouse pointer) over the scrollbar that I have circled in the image.

The cursor remains an "I" (工). It should become an arrow.

mm

2.24.2

Chrome

draunger:
## DESCRIPTION
# APPROACH
I HAVE USED WEBKIT PROPERTY IN CSS TO CHANGE THE POINTER 
## ATTACHED ISSUE :
Closes #9507

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2664,7 +2664,7 @@ button.raw-tag-option svg.icon {
     white-space: pre;
 }
 .tag-text::-webkit-scrollbar-thumb:hover { 
-    cursor: pointer; /* Change mouse pointer to a hand */
+    cursor: auto;
 }
 
 .tag-text,
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2663,6 +2663,9 @@ button.raw-tag-option svg.icon {
     font-family: monospace;
     white-space: pre;
 }
+.tag-text::-webkit-scrollbar-thumb:hover { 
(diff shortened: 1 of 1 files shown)


css/80_app.css
```diff
@@ -2663,6 +2663,9 @@ button.raw-tag-option svg.icon {
     font-family: monospace;
     white-space: pre;
 }
+.tag-text::-webkit-scrollbar-thumb:hover { 
+    cursor: pointer; /* Change mouse pointer to a hand */
```
tyrasd: usually, scrollbars use the default  pointer, don't they? I would let the browser decide which pointer to use:

```suggestion
    cursor: auto;
```

tyrasd (changes requested):
Can you please apply the fix to all `textarea` elements, not just the raw tag editor?  It should best be defined next to where the style for `::-webkit-scrollbar` is specified in the css.

PS: I think your caps-lock key is stuck. :stuck_out_tongue_closed_eyes:

css/80_app.css
draunger: @tyrasd you are right :)

draunger:
@tyrasd i have change the cursor of all text area :)

tyrasd (approved):

PR #10632, opened 2024-12-28
Do not add ignore_issue if there are no other dynamic fixes, tries to fix #10614

### LINKED ISSUE #10614: "Add tunnel" quick fix not shown after already fixing one
### URL

_No response_

Create a waterway that crosses two roads.

When selecting the waterway, the validator correctly shows the quick fix menu the first time but after selecting "Add tunnel", the next issue only show an option to ignore the problem.

I understand that this happens because after the tunnel is created, the selection includes three waterways and the validator don't know what to do.
It should either:
- show the full quick fix menu
- automatically change selection to the next problematic segment
- don't show "ignore this issue" because it might mislead mappers into thinking there’s nothing else to do to fix it

And clicking the yellow box zooms in on the problematic way but still keeps the selection the same, so the quick fix menu is still incomplete.

2.30.4

Chrome, Firefox

Deeptanshu-sankhwar:
Tries to fix #10614

I have tried to solve the problem of the mapper getting mislead into thinking there is nothing else to fix by showing "ignore this issue" while adding a tunnel

I have added an extra check which prevents inserting the `ignore_issue` `validationIssueFix` if there are no other `dynamicFixes` in the context

### DIFF
--- a/modules/core/validation/models.js
+++ b/modules/core/validation/models.js
@@ -63,7 +63,7 @@ export function validationIssue(attrs) {
         var fixes = this.dynamicFixes ? this.dynamicFixes(context) : [];
         var issue = this;
 
-        if (issue.severity === 'warning') {
+        if (issue.severity === 'warning' && fixes.length > 0) {
             // allow ignoring any issue that's not an error
             fixes.push(new validationIssueFix({
                 title: t.append('issues.fix.ignore_issue.title'),


Deeptanshu-sankhwar:
https://github.com/user-attachments/assets/6bcf2e33-fad0-4069-bff4-39f910b7e3cb

Here is a demonstration video to show how it fixes the issue

tyrasd:
This does not really solve the issue. It would also break all other validation warnings that don't provide dynamic fixes.

PR #10633, opened 2024-12-28
fixed issue Merging two ways corrupts step_count #10492

### LINKED ISSUE #10492: Merging two ways corrupts step_count
### URL

_No response_

Follow-up to #8069 

Splitting a way in two results in i.e. step count of 4 => 3+1 / 2+2 depending on the length of the way itself.
However, merging two ways will result in 3+1 => Error (values are not the same), or 2+2 => 2 (merged, values are not summed)
Perhaps summing up the values, or alternatively (prompting to) delete the values (if situation has changed significantly), is better. 

See also https://github.com/MarcusWolschon/osmeditor4android/issues/1251#issuecomment-706021265

_No response_

2.30.4

Chrome

Darshit42:
this pr fixes #10492  ,changes are such that it adds up the values of tags

before fixing the issue tags such as step_count , parking:left, etc values were not adding up when lines were merged, this creates confusion to user on how merging works
merging two ways resulted in 3+4 => Error (values are not same) or 2+2 => 2 (values are not summed)

after this , when lines are merged it adds up the values of tags successfully 
merging two ways resulted in 3+4 => 7

https://github.com/user-attachments/assets/a8f074d4-56ac-4ab1-b8b4-d569d61694ed


Darshit42:
to anyone who's reading this , if i made any mistakes in raising this pr feel free to criticize it's my first time raising a pr in other repo besides mine 🙃

modules/osm/entity.js
```diff
@@ -144,20 +144,20 @@ osmEntity.prototype = {
...
+            } else if (!isNaN(t1) && !isNaN(t2)) {
+                changed = true;
+                merged[k2] = String(Number(t1) + Number(t2));
             } else if (t1 !== t2) {
                 changed = true;
-                merged[k] = utilUnicodeCharsTruncated(
-                    utilArrayUnion(t1.split(/;\s*/), t2.split(/;\s*/)).join(';'),
-                    255 // avoid exceeding character limit; see also context.maxCharsForTagValue()
-                );
+                merged[k2] = `${t1};${t2}`;
```
1ec5: The previous code intentionally eliminated duplicates from the list. Merging two features with identical values for a given key shouldn’t result in the value being repeated twice. Tests are failing because this code unconditionally concatenates the two values.

modules/osm/entity.js
```diff
@@ -144,20 +144,20 @@ osmEntity.prototype = {
...
+        for (var k2 in tags) {
+            var t1 = merged[k2];
+            var t2 = tags[k2];
             if (!t1) {
                 changed = true;
-                merged[k] = t2;
+                merged[k2] = t2;
+            } else if (!isNaN(t1) && !isNaN(t2)) {
+                changed = true;
+                merged[k2] = String(Number(t1) + Number(t2));
```
1ec5: While it may be appropriate to treat `step_count=*` as a summable number, this is not necessarily the case with every tag value that consists of digits. For example, if I merge two ways with `maxspeed=30` and `maxspeed=35`, iD should reject the merge because of conflicting tags rather than resulting in `maxspeed=65` (a speed limit of 65&nbsp;kilometers per hour).

1ec5 (commented):
Thanks for this pull request – the test failures appear to be pointing to unintended consequences of the proposed changes.

modules/osm/entity.js
Darshit42: ill look into these tags and make a list to exclude them from being adding up , this might fix the issue, if there's anything im missing please let me know @1ec5

modules/osm/entity.js
Darshit42: ill change it and fix the repeated value error

modules/osm/entity.js
1ec5: `step_count=*` is probably an outlier as a summable key. Some other keys like `hoops=*` (for the number of hoops on a basketball court) and `capacity=*` might be summable in some cases but not others, so I don’t think it would be safe to assume we can sum them.

PR #10639, opened 2024-12-30
allow duplicate values in semiCombo for some fields

### LINKED ISSUE #10635: semiCombo does not support duplicate values
### URL

https://osm.org/edit?way=26229179

Fixing this properly will require PRs in 3 different repositories, so I'm creating an issue to keep this organised.

From https://github.com/openstreetmap/id-tagging-schema/issues/1065#issuecomment-1837135803:

> BTW, there's also a bug when entering multiple `none` values to the `destination:symbol` key. If we add two of more `none` values separated with `;` iD only considers one. And if you grab and move one of those `none` values, iD eliminates all of them and leaves only one in the value part of the tag.

This bug/limitation also affects `*:lanes:*`, `seamark:*:colour`, as well as `destination:*`, where values like `…;none;none;…` or `red;white;red` are acceptable.

Untitled

_No response_

all

Microsoft Edge

k-yle:
Closes #10635 (part 2 of 3)

The main change was related to this delete button:
Untitled
It now has to delete by index, rather than by tag value. 

Everything else was a trivial change.

### DIFF
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -208,7 +208,9 @@ export function uiFieldCombo(field, context) {
             _comboData = _comboData.filter(filter);
         }
 
-        _comboData = objectDifference(_comboData, _multiData);
+        if (!field.allowDuplicates) {
+            _comboData = objectDifference(_comboData, _multiData);
+        }
         _combobox.data(_comboData);
 
         // hide the caret if there are no suggestions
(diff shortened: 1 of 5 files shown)


modules/ui/fields/combo.js
```diff
@@ -410,11 +417,12 @@ export function uiFieldCombo(field, context) {
         if (_isMulti) {
             t[d.key] = undefined;
         } else if (_isSemi) {
-            var arr = _multiData.map(function(md) {
-                return md.key === d.key ? null : md.key;
-            }).filter(Boolean);
+            let arr = _multiData.map(item => item.key);
+            arr.splice(d.index, 1);
```
tordans, line 430: Maybe add a comment here that this is the place where we actually delete the entry. I find the splice to be less obvious than the filtering that was used before in signaling that this is delete action.

tordans (approved):

modules/ui/fields/combo.js
k-yle: good idea, done

tyrasd (approved):
Thanks!

PR #10645, opened 2025-01-03
fix relation role dropdown not working

### LINKED ISSUE #10481: Relation member role comboboxes are not displaying available options when selected
### URL

https://www.openstreetmap.org/edit#map=20/42.3609187/-71.0557088

Steps to replicate the issue:
1. Find or create a relation of a type for which options are provided by the taginfo service (e.g., a restriction: no left turn)
2. Choose a feature that is part of the relation
3. Select the relation in the relations section of the feature editor
4. Click a member's role combobox

Depending on the relation, member role options are provided by the taginfo service with relevant choices when selecting member role combobox for a given relation. However, these options are not being displayed, leaving users with a combobox that functions as a text field.

A taginfo service request is triggered upon clicking a member role combobox (raw member editor), and a valid response is received, but the options are not displayed. The member role combobox for the same member, displayed in the relations section (raw membership editor) of the feature editor, shows available options when selected.

It appears the issue is caused when the mousedown event of a click on the member role input element bubbles to the member list item, which then triggers d3-drag handling, subsequently preventing the mouseup event from firing. This prevents the role options from being displayed as a result of the combobox mouseup event not firing. Stopping event propagation at the member's input wrapper, or in the combobox mousedown event, prevents the issue by preventing the drag events from triggering.

Stopping event propagation at the input wrapper or combobox prevents member dragging using the combobox, but still enables users to use the member header/label for reordering.

The issue is observed in Firefox version 130.0.1 and Chrome version 129.0.6668.60 on Windows 11.

Current state: relation member role options not displayed:
OpenStreetMap_RelationMemberRoleOptionsNotDisplayed

Expected behavior (relation member role options displayed):
OpenStreetMap_RelationMemberRoleOptionsExpected

2.30.2

Chrome, Firefox

k-yle:
Closes #10481 

Left-clicking on the input field should open up the dropdown, just like does in the _Membership_ (cf _Member_) section.

image

The reason for the inconsistency is that _Member_ section is reorderable using drag-and-drop, while the _Membership_ section is not, so `d3-drag` is not involved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Fix flickering of imagery metadata information in background panel ([#9754])
 * Immediately update raw tag key/value inputs when spaces have been trimmed ([#11206])
 * Fix duplicate values deleted when editing `destination:*` tags ([#10639], thanks [@k-yle])
+* Fix the relation role dropdown not opening on click ([#10645], thanks [@k-yle])
 #### :earth_asia: Localization
 #### :hourglass: Performance
 #### :rocket: Presets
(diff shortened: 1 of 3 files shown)


k-yle:
also fixed the css cursor issue

tyrasd (approved):

PR #10646, opened 2025-01-03
fix `destination_sign` relations not updated when splitting a way

### LINKED ISSUE #9188: Splitting a 'to' member of destination_sign relations adds an extra 'to' member
### URL

_No response_

Just split a way that is a 'to' member of any destination_sign relation.
destination_sign relations can have exactly one 'to' member.
Splitting a 'from' member also adds extra 'from', not sure if that is a problem too.

_No response_

Chrome

k-yle:
Closes #9188

[`destination_sign`](https://osm.wiki/Relation:destination_sign) can use the `intersection` role instead of `via`. 

There are still valid cases `destination_sign` relations won't be split properly, such as if there is no `from` role. But this helps with the most common cases

### DIFF
--- a/test/spec/actions/split.js
+++ b/test/spec/actions/split.js
@@ -1653,7 +1653,9 @@ describe('iD.actionSplit', function () {
         });
 
 
-        ['restriction', 'restriction:bus', 'manoeuvre'].forEach(function (type) {
+        ['restriction', 'restriction:bus', 'manoeuvre', 'destination_sign'].forEach(function (type) {
+            const viaRole = type === 'destination_sign' ? 'intersection' : 'via';
+
             describe('type = ' + type, function () {
                 var a = iD.osmNode({id: 'a', loc: [0, 0]});
                 var b = iD.osmNode({id: 'b', loc: [1, 0]});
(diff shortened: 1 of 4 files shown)


tordans (commented):
There are a few code comments that only reference the VIA now. Is there a way to update those without writing an essay on the topic? :)

test/spec/actions/split.js
```diff
@@ -1674,7 +1676,7 @@ describe('iD.actionSplit', function () {
                     iD.osmRelation({id: 'r', tags: {type: type}, members: [
                         {id: '-', role: 'from', type: 'way'},
                         {id: '~', role: 'to', type: 'way'},
-                        {id: 'd', role: 'via', type: 'node'}
+                        {id: 'd', role: via, type: 'node'}
```
tordans: When reading the changes, I did not immediately get this. Maybe a different name like `viaKey` would make it more explicit? (Code editors will also help with proper tool tips and highlighting, though)

k-yle:
> There are a few code comments that only reference the VIA now. Is there a way to update those without writing an essay on the topic? :)

good point, although I can't think of concise word that encompasses `via` and `intersection`. Something like `midpoint` is more verbose and possibly less obvious than `via`...

tyrasd (approved):

PR #10648, opened 2025-01-03
fix relation membership list using a non-deterministic order

k-yle:
When selecting a way, the list of relations uses an unpredictable order. 

This is annoying, but more importantly, it causes a rare bug: You can't merge ways if the relations were added in a different order. 

For example, merging these ways:..

<table>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

...will cause an unexpected error:

image

### DIFF
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -50,6 +50,8 @@ export function uiSectionRawMembershipEditor(context) {
     var _entityIDs = [];
     var _showBlank;
     var _maxMemberships = 1000;
+    /** @type {Set<string>} relations that were added after this panel was opened */
+    const recentlyAdded = new Set();
 
     function getSharedParentRelations() {
         var parents = [];
@@ -114,7 +116,37 @@ export function uiSectionRawMembershipEditor(context) {
             membership.role = roles.length === 1 ? roles[0] : roles;
         });
 
-        return memberships;
+        const existingRelations = memberships
+            .filter(membership => !recentlyAdded.has(membership.relation.id))
+            .map(membership => ({
+                ...membership,
+                // We only sort relations that were not added just now.
+                // Sorting uses the same label as shown in the UI.
+                // If the label is not unique, the relation ID ensures
+                // that the sort order is still stable.
+                _sortKey: [
+                    baseDisplayValue(membership.relation),
+                    membership.relation.id,
+                ].join('-'),
+            }))
+            .sort((a, b) => {
+                return a._sortKey.localeCompare(
+                    b._sortKey,
+                    localizer.localeCodes(),
+                    { numeric: true },
+                );
+            });
+
+
+        const newlyAddedRelations = memberships
+            .filter(membership => recentlyAdded.has(membership.relation.id));
+
+        return [
+            // the sorted relations come first
+            ...existingRelations,
+            // then the ones that were just added from this panel
+            ...newlyAddedRelations,
+        ];
     }
 
     function selectRelation(d3_event, d) {
@@ -184,6 +216,7 @@ export function uiSectionRawMembershipEditor(context) {
         }
 
         if (d.relation) {
+            recentlyAdded.add(d.relation.id);
             context.perform(
                 actionAddMembers(d.relation.id, _entityIDs, role),
                 t('operations.add_member.annotation', {
@@ -251,14 +284,6 @@ export function uiSectionRawMembershipEditor(context) {
 
         var graph = context.graph();
 
-        function baseDisplayValue(entity) {
-            var matched = presetManager.match(entity, graph);
-            var presetName = (matched && matched.name()) || t('inspector.relation');
-            var entityName = utilDisplayName(entity) || '';
-
-            return presetName + ' ' + entityName;
-        }
-
         function baseDisplayLabel(entity) {
             var matched = presetManager.match(entity, graph);
             var presetName = (matched && matched.name()) || t('inspector.relation');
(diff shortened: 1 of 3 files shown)


modules/ui/sections/raw_membership_editor.js
```diff
@@ -114,7 +114,14 @@ export function uiSectionRawMembershipEditor(context) {
...
 
-        return memberships;
+        return memberships
+            .map(membership => ({
+                ...membership,
+                // this property is added purely for sorting
+                _label: baseDisplayValue(membership.relation)
+            }))
+            // sort alphabetically based on the labels used in the UI
+            .sort((a, b) => a._label.localeCompare(b._label));
```
tordans:
```suggestion
                _sortKey: baseDisplayValue(membership.relation)
            }))
            // sort alphabetically based on the labels used in the UI
            .sort((a, b) => a._sortKey.localeCompare(b._sortKey));
```

tordans (approved):

modules/ui/sections/raw_membership_editor.js
```diff
@@ -114,7 +114,13 @@ export function uiSectionRawMembershipEditor(context) {
...
         });
 
-        return memberships;
+        return memberships
+            .map(membership => ({
+                ...membership,
+                _sortKey: baseDisplayValue(membership.relation)
+            }))
+            // sort alphabetically based on the labels used in the UI
+            .sort((a, b) => a._sortKey.localeCompare(b._sortKey));
```
1ec5: This still wouldn’t necessarily be a stable sort if the parent relations have the same label. Can we sort on the relation ID instead?

k-yle: To get the best of both worlds, could we sort by something like <code>{name}—{relationId}</code>? This means that relations would still be grouped by preset, and then sorted A-Z

<details>
<summary>sorting by relation ID means that the list is still chaotically ordered, from the user's perspective: (click to view screenshot)</summary>
image
</details>

1ec5: Yes, that would be reasonable. I guess ideally any memberships newly added in the current editing session would be listed after existing memberships, but that would be more difficult.

tordans: > any memberships newly added in the current editing session would be listed after existing memberships, but that would be more difficult

Or highlighted by some other means like color or text. But that could be a good contribution  issue  by itself (in which  we describe the likely best technical solution).

1ec5: I’m just concerned that the new membership tends to jump up somewhere in the middle of the list as soon as you add it. This differs from the behavior of the raw tag editor, so the user may end up setting a role on the wrong membership by accident. Can we temporarily prevent the membership list from resorting until the selection changes?

tordans: I see the UX issue. 

Would it be possible to have a visual order and an internal order (separate) and use the internal order for the merge use case that triggered this PR?
Or, maybe reorder right before doing the merge, so both ways use the same order when the merge happens?
Or, maybe only order when a way is selected, but not when a relation is added? (I assume the merge case would be resolved still because it requires a re-select(?) triggering a reorder?)

k-yle: to ensure the order is stable, I've changed `_sortKey` to `{label}-{relationId}`. 

To ensure that "recently added" relations appear at the bottom, we split out these members and don't sort them.

When the `selection` of the panel changes, we clear the "recently added" list.

modules/ui/sections/raw_membership_editor.js
```diff
@@ -114,7 +116,31 @@ export function uiSectionRawMembershipEditor(context) {
...
+                // We only sort relations that were not added just now.
+                // Sorting uses the same label as shown in the UI.
+                // If the label is not unique, the relation ID ensures
+                // that the sort order is still stable.
+                _sortKey: [
+                    baseDisplayValue(membership.relation),
+                    membership.relation.id,
+                ].join('-'),
+            }))
+            .sort((a, b) => a._sortKey.localeCompare(b._sortKey));
```
1ec5: Can we use `Intl.Collator`, setting options that include [numeric sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#numeric)? `String.localeCompare` does a lexicographical comparison, which is normally fine, but many of these labels will be numeric or contain numbers, particularly route relations, which don’t make as much sense when sorted alphabetically. It’s worse in OpenHistoricalMap’s fork, which appends a date range to every element’s label: OpenHistoricalMap/iD#184.

k-yle: done, using `String#localeCompare`'s [third argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#options), rather than directly invoking `Intl.Collator`, to avoiding writing a fallback [like we've done for other `Intl` APIs](https://github.com/openstreetmap/iD/blob/b2b397c7ce2325e9f770f586d7da811681723a21/modules/core/localizer.js#L229)

1ec5 (approved):

tyrasd (changes requested):
Please add a changelog entry. Otherwise it look good!

tyrasd (approved):

Dimitar5555:
This PR should resolve #9502 and #9788, correct?

tyrasd:
I'd think so. @k-yle can you double-check this please?

Issue #10661, opened 2025-01-07
When searching while hovering over a place search result, highlights become visually stuck


danieldegroot2:
### URL

https://www.openstreetmap.org/edit#map=19/41.961937/-88.720685

### How to reproduce the issue?

Minor visual bug.
When searching while hovering over a place search result, highlights become visually stuck

Go to left sidebar
Click on place search input
Move the cursor so it hovers over the empty area below place search input
Searching "ki", "Kish Hospital Drive" is highlighted as its the first search result which is hovered over.
Other "Kish Hospital Drive" is also highlighted while we keep typing "kish..."
Remove "ki..." and enter "dek..."
"DeKalb Avenue" is highlighted.
Other "DeKalb Avenue" is also highlighted while we continue typing "dekalb...".
Both "Kish Hospital Drive" results are also still highlighted.

### Screenshot(s) or anything else?

![image](https://github.com/user-attachments/assets/5d58cefe-48be-4da8-9fdf-019060283dee)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app, RapiD version at mapwith.ai/rapid

### What version numbers does this issue effect?

2.30.4, 2.31.0-dev, Rapid 2.5.0

### Which browsers are you seeing this problem on?

Chrome

draunger:
@danieldegroot2 i am interested in this issue...

danieldegroot2:
@tyrasd seems there is a regression when refreshing search after doing a coordinate search. (regardless of mouse position)

1. Use place search with i.e. `52.6412925, -7.2492597`. Returns single entry with coordinates. (as usual)
2. Select all (Ctrl+A) and remove search input (backspace).;

Or,

2. Click 'Search worldwide'. Returns second entry for '5, The Knoll'. Hover over this entry, it is highligted on the right. Click entry for '5, The Knoll'.

3. [any existing search input disappears.] When dragging map view, it glitches out / back to previous view. It is only possible to 'drag' the map view by using street level imagery forward/backward buttons, or slightly using the keyboard arrow keys.

Returns following error
```
Uncaught SyntaxError: Failed to execute 'querySelectorAll' on 'Element': '.-1' is not a valid selector.
    at SVGSVGElement.<anonymous> (selectorAll.js:7:17)
    at Ur.w_ [as selectAll] (selectAll.js:18:31)
    at St (util.js:145:10)
    at v (feature_list.js:342:13)
[...]

PR #10666, opened 2025-01-09
fixed-when searching while hovering over a place search result, highlights become visually stuck

draunger:
# description
## problem before the fix:
- when we search the place by placing the cursor on the empty area initially ,
-  then after writing the place, the search result comes and as the cursor is on the item 
- it highlight that item on the map , so after changing the place by placing the cursor on the same position as before 
- the item is remain highlighted or if we change the place without changing the position of the cursor the item 
- which is hovered before is highlighted along with the new result 
-  here you can see in this video attached below:-

https://www.loom.com/share/3ff0436cf2854bed8f390df3710ad5f1?sid=97a6949d-10f5-46e5-8144-60158e730c18

as you can see in this video when we write "dek" then kish hospitality drive also highlighted along with dekalb...

## approach:
- Basically i used a set data structure and push every id of highlighted item 
- when we change the search label then the previous highlighted items get unhighlighted 
- i used this in if-condition and iterate in set data structure so that all the previous highlighted item get unhighlighted 
- here you can see in this video atttached below :

https://www.loom.com/share/a0eff1338e69448b8c16d2d020039528?sid=411b93af-545b-48e7-9e71-a0d10a9180f6

# attached issue : 
https://github.com/openstreetmap/iD/issues/10661

### DIFF
--- a/modules/ui/feature_list.js
+++ b/modules/ui/feature_list.js
@@ -243,8 +243,8 @@ export function uiFeatureList(context) {
             return result;
         }
 
-        var highlight_after_mouseout = false;
-        var id_on_mouseover =-1;
+        var highlightAfterMouseout = false;
+        let idOnMouseover = new Set();
         function drawList() {
             var value = search.property('value');
             var results = features();
@@ -279,9 +279,10 @@ export function uiFeatureList(context) {
                   .attr('class', 'entity-name')
                   .call(t.append('geocoder.search'));
             }
-
-            if ((!value || !results.length) && highlight_after_mouseout  && id_on_mouseover !== -1) {
-                utilHighlightEntities([id_on_mouseover], false, context);
+            if (highlightAfterMouseout) {
+                idOnMouseover.forEach((value) => {
+                utilHighlightEntities([value], false, context); 
+                });
             }    
             list.selectAll('.no-results-item')
                 .style('display', (value.length && !results.length) ? 'block' : 'none');
@@ -339,15 +340,15 @@ export function uiFeatureList(context) {
 
         function mouseover(d3_event, d) {
             if (d.id === -1) return;
-            highlight_after_mouseout=true;
-            id_on_mouseover =d.id;
+            highlightAfterMouseout=true;
+            idOnMouseover.add(d.id);
             utilHighlightEntities([d.id], true, context);
         }
 
 
         function mouseout(d3_event, d) {
             if (d.id === -1) return;
-            highlight_after_mouseout=false;
+            highlightAfterMouseout=false;
             utilHighlightEntities([d.id], false, context);
         }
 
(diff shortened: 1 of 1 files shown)


tyrasd (dismissed):
Can you please describe your thoughts behind this bugfix? What were the causes of the bug to occur in the first place and how do your changes address these?

Your code should further match the style of the rest of the codebase: for example we use camel-case instead of underscores in variable names, and whitespace should be consistently useed around `=`.- Please clean this up.

Finally, please use a descriptive title for your pull request and don't shout in ALL-CAPS. :pray:

draunger:
@tyrasd i have describe my thoughts and also the describe the issue in detail , very sorry for ALL-CAPS  🙏🙏

tyrasd:
Ok, thanks for that.

The downside of your approach is that this "solution" it is merely addressing the symptom and not the cause of the underlying problem. Admittedly, this was not a trivial one to figure out, but here is what I noticed:

* There are event handlers set for both `mouseover` and `mouseout`, which should "in theory" already disable the highlighted state of the previously selected entity when the mouse enters a different element.
* For some reason, the `mouseout` handler is sometimes not called. There must be some underlying problem causing this event handler not to be triggered properly.
* One of the cases where the event handler is not called is when the DOM element is removed from the page (in `items.exit().remove()`). This makes sense: when there is no element, there are also no events. So, to fix this case, we need to manually call the `mouseout` handler when DOM elements are removed. Like this: `items.exit().each(d => mouseout(undefined, d)).remove();`
* This fixes some cases of the bug, but not all: when the mouse is on the first search result, the highlight can still get stuck. After a bit of debugging, I noticed that `items.enter()` had always 1 DOM element, even if the search result list had the same length as previously. This should not be the case, as d3 should never have to create new DOM elements when they already exist. So, something must be wrong with the data binding of the search results to the DOM.
* The culprit turned out to be a few lines further above: 
  ````js
            list.selectAll('.feature-list-item')
                .data([-1])
                .remove();
  ````
  This was added a long time ago with #2056.
* This indeed messes up the binding of the first DOM element of the `items` selection to the first search result, as it is always `remove`d whenever the search results list is updated.
* This explains why the `mouseout` event is not called for the first element of the search results list, when one types in the search field: As the element is always deleted and immediately recreated in the following `items.enter()` instruction, the mouse events don't have a chance to be properly triggered.
* I cannot see what useful things the `….data([-1]).remove()` instruction would be meant to doing here: it's not related to the functionality of the PR where it was added, and it generally does not seem to make much sense. :shrug:  My best guess is that it was accidentally included in the commit and overlooked in the PR review back then.
* Simply removing the culprit 3 lines of code seems to finally fully fix the original issue. Most importantly, this does not require adding any workaround and additional complexity to the code.

Issue #10653, opened 2025-01-05
Minimaps Go Black When Zoomed Out on Specific Backgrounds (OpenTopoMaps, Bing Maps Aerial, Mapbox Satellite)


Darshit42:
### How to reproduce the issue?

When using minimap functionality, the minimap turns completely black under the following conditions:

1. select a background such as OpenTopoMaps, Bing Maps Aerial, or Mapbox Satellite is selected
2. zoom out on the map

### Screenshot(s) or anything else?

https://github.com/user-attachments/assets/c30e4e2a-befa-43e6-b312-22cf0f7fa330

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.30.4

### Which browsers are you seeing this problem on?

Chrome

Darshit42:
i think issue is it fails to render the minimap after a certain limit , im working on this , if there's something else causing feel free to leave a comment

tyrasd:
My initial guess would be that the minimap may not properly respect the _minimum zoom level_ property of the respective background layers. @Darshit42 feel free to investigate this.

Darshit42:
i am investigating it, i will checkout it and if im successful i will make a pr

Darshit42:
@tyrasd 
![image](https://github.com/user-attachments/assets/56954bf4-2866-4df3-98b0-295fcd79477e)

this is background_source.js

i have made changes in this function so that the zoom size is always valid , is this correct? or do i need to make some other changes too?

this is the output before minimap was not able to render after 200km 

![image](https://github.com/user-attachments/assets/0ac10685-549e-4fd8-811b-ee1791df2925)


if everything's fine should i raise a pr?

Darshit42:
@tyrasd i tried running npm test and there are 4 errors which are all due to zooming out which causes minimap to go black , therefore when i fix the issue and remove the upper bound on the zoom extent these errors occur

![image](https://github.com/user-attachments/assets/4bba14b4-813b-4959-944a-121994eb9e14)

![image](https://github.com/user-attachments/assets/4f6f5bdd-e733-4536-989b-933651c6371b)


should i change test cases to fix these errors??

PR #10680, opened 2025-01-15
Fix: issue #10653 Minimap go black when zoomed out

Darshit42:
fixed the issue #10653 when we zoomed out minimap used to go black , now made a new variable clampedzoom to ensure minimap renderer zoom out is fixed so  that it dont overflow and respect the boundaries

before

image

after

image

### DIFF
--- a/modules/renderer/tile_layer.js
+++ b/modules/renderer/tile_layer.js
@@ -122,24 +122,25 @@ export function rendererTileLayer(context) {
         var requests = [];
         var showDebug = context.getDebug('tile') && !_source.overlay;
 
-        if (_source.validZoom(_zoom)) {
-            tiler.skipNullIsland(!!_source.overlay);
+        var clampedZoom = Math.max(_source.zoomExtent[0], Math.min(_zoom, _source.zoomExtent[1]));
 
-            tiler().forEach(function(d) {
-                addSource(d);
-                if (d.url === '') return;
-                if (typeof d.url !== 'string') return; // Workaround for #2295
-                requests.push(d);
-                if (_cache[d.url] === false && lookUp(d)) {
-                    requests.push(addSource(lookUp(d)));
-                }
-            });
+        tiler.skipNullIsland(!!_source.overlay);
 
-            requests = uniqueBy(requests, 'url').filter(function(r) {
-                // don't re-request tiles which have failed in the past
-                return _cache[r.url] !== false;
-            });
-        }
+        tiler().forEach(function(d) {
+            addSource(d);
+            if (d.url === '') return;
+            if (typeof d.url !== 'string') return; // Workaround for #2295
+            requests.push(d);
+            if (_cache[d.url] === false && lookUp(d)) {
+                requests.push(addSource(lookUp(d)));
+            }
+        });
+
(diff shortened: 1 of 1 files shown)


tyrasd (changes requested):
It doesn't really fix the issue, unfortunately.

Also, have you thought this through? Can you please explain how this change is supposed to fix the issue in the first place?

Darshit42:
i think its due to background map as it was occuring in mapbox satellite and after this it wasnt occuring , i thought of fixing the issue by setting up a maximum limit for zoom out , before if we zoomed out of certain limit map would just stop rendering , i set the clampedzoom to be equal to maximum zoomout capacity whenever it crosses the limit , so in a way its fixed on up limit of zoom rather than overflowing , here is a screen recording you can cross check, i will try digging more , i will check files but if you can give a hint or guide me it will be very helpful.





https://github.com/user-attachments/assets/e2c48a6b-e153-4889-89ad-6a8c5f5d4d77

tyrasd:
Sure, here are some hints:

* You're manipulating the value of the zoom level without adjusting the map projection itself. Some parts of the code are still using the original `projection` object which has the original zoom value: e.g. `_tileOrigin` is calculated based on `_projection.scale()`. Now the tile's x/y offsets are wrongly calculated in `imageTransform` because they partly use the clamped zoom value and partly the original one. You might got lucky in your test the cases, possibly because this error is small for zoom levels near 0.
* The `tiler` module is already clamping zoom values to the available zoom range (see `tiler.js` line 37), no need to implement this again.
* The main issue with the zoom clamping approach and the reason why we explicitly check for `validZoom` is the following: When using a clamped zoom level to "under-zoom" the map, one needs to draw a larger number of tiles to cover the same area. For example: If the layer is available only for zoom 12 to 18, and the minimap should be rendered for zoom 11, we need to request and draw 4 times as many tiles (because the zoom 12 tiles are only covering half the width and height of a zoom 11 tile that we would actually need). At zoom 11 this would not be a problem yet, but this number grows exponentially: at zoom 10, it would require 16 times as many tiles, at zoom 9 it would be 64 times as many, etc. – This means, we can only under-zoom a small amount of zoom levels until this becomes a serious problem.
* So what we want is the following: allow the minimap to display tiles at zoom level 0 for tiles layers which only have tiles up to zoom level 1. But we don't want to render the millions of tiles a tile source with a minimum zoom of 12 would require to be rendered as zoom 0.
* Here's one outline of an idea of a possible implementation: Introduce a property/flag in `tile_layer.js` that allows other modules to explicitly enable under-zooming by a certain specified number of zoom levels and pass this as an additional argument to `validZoom(_zoom, allowedUnderZoomAmount)`. In `map_in_map.js`: set this number to a value of 1 or 2.

Darshit42:
@tyrasd ive made the changes you told me in this branch of my fork https://github.com/Darshit42/iD/tree/fixed-minimap-issue #10694

i double checked in all the maps and found that apart from open topo map background there is no error in any other map , can you confirm and tell me if there's some other file i need to change , i ran the tests and all of them passed , if there's no issue can i raise the pr??

(also there is another case , we can turn validZoom in background_source.js to always return true , but same problem persists, in open topo map background it shows different tiles when rendering )

PR #10684, opened 2025-01-16
add indigenous languages not in CLDR

k-yle:
I didn't realise this was possible until I saw the CLDR override list today.

This PR adds 30 entries from [the australian wiki page](https://osm.wiki/Australian_Tagging_Guidelines/Australia%27s_First_Peoples). cc @andrewharvey, I see you edited this page recently


data/languages.json
```diff
@@ -122,6 +126,7 @@
   "ckb": {"nativeName": "کوردیی ناوەندی"},
   "clc": {},
   "co": {"nativeName": "corsu"},
+  "coa": {"nativeName": "Cocos Malay"},
```
1ec5: I think the intention is to use the native name rather than the English name in the `nativeName` property.

data/languages.json
```diff
@@ -37,6 +39,7 @@
   "asa": {"nativeName": "Kipare"},
   "ast": {"nativeName": "asturianu"},
   "atj": {},
+  "aus": {"nativeName": "Generic Australian Aboriginal"},
```
1ec5: We don’t have many intentional instances of language families in this list, but I suspect they all either omit the `nativeName` property or use a name that’s a compromise among the various languages within the family.

data/languages.json
k-yle: I might be missing something, but I think the current override system only lets you define `nativeName`, you can't define translated names. So misusing `nativeName` is the only easy solution unless we want to rework this build script. 

If we do change the build logic, we need to decide if all translated labels should be stored in this same JS object, or if the overrides should be synced to transifex[^1]

[^1]: Even if we sync these overrides to transifex, the default language (english) would still need to be defined in this file along with `nativeName`.

andrewharvey:
Fabulous! I did just add an entry on the wki as I was mapping https://www.openstreetmap.org/way/395478653. Would be great to have these appear as their actual language names rather than the ISO code in the iD presets.

I don't know enough about this specific file and how the nativeName field works to comment.

data/languages.json
andrewharvey: Is it still possible to translate it without the nativeName listed here? I don't know if there would be a compromise native name for `aus` from all the languages part of `aus`.

As for the best English, I would think "Australian languages" or "Australian Aboriginal languages".

tyrasd: > We don’t have many intentional instances of language families in this list

Do we have any? From what I can see, we at least have no ISO 639-2 codes that have the scope "Collective".

<hr>

Maybe it's best to omit codes for language families? I don't know anything about `aus` languages, but if there were a `name:gem` tagged somewhere, it would find it not very useful, as it could be anything from Icelandic, English, Dutch, Danish, Norwegian, Swedish, German or any other Germanic language. Btw, I actually found a couple of those in Romania ([example](https://www.openstreetmap.org/node/265006357)), where the tag was misused to apparently tag a local German dialect (it should have better been something like `[old_]name:bar`), but I digress.

data/languages.json
tyrasd: I would have said that the actual native name of the language would be preferred here, if known. 

We could in the future (not in this PR) enhance the build scripts to include one or a few "fallback names" in other languages (e.g. English or a common language of the region) for cases when we don't know the language's native name or to simply make it more accessible to an international audience.

I wouldn't go as far as starting to translate those on transifex, because it would essentially duplicate the work done by the CLDR project, where it would be better to [directly contribute there](https://cldr.unicode.org/#how-to-contribute).

tyrasd (commented):
If you like you can also add these languages to the list of `territoryLanguages` in `scripts/build_data.js` (see line 186): then they are included towards the top of the languages dropdown menu in the respective country.

data/languages.json
tyrasd: > Is it still possible to translate it without the nativeName listed here?

To clarify: Currently none of the language names are translated by "us". All language name translations come from the CLDR project.

CHANGELOG.md
```diff
@@ -71,6 +72,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 [#10624]: https://github.com/openstreetmap/iD/issues/10624
 [#10634]: https://github.com/openstreetmap/iD/issues/10634
 [#10650]: https://github.com/openstreetmap/iD/issues/10650
+[#10684]: https://github.com/openstreetmap/iD/issues/10684
```
tyrasd:
```suggestion
[#10684]: https://github.com/openstreetmap/iD/pull/10684
```

:wink:

data/languages.json
andrewharvey: `name:aus` currently has the most use of any of the other languages listed at https://wiki.openstreetmap.org/wiki/Australian_Tagging_Guidelines/Australia's_First_Peoples#Mapping_Indigenous_Names

It's used where the specific language is not known and so allows us to still tag it, until the specific language can be determined, and we should have some label for it beyond just `aus` which is misleading as a label.

1ec5: > Do we have any? From what I can see, we at least have no ISO 639-2 codes that have the scope "Collective".

Through CLDR, we have language codes that ISO 639-3 considers to be language families, such as Arabic, Chinese, and Kurdish. However, a single native name is pretty well established for those codes.

> It's used where the specific language is not known and so allows us to still tag it, until the specific language can be determined, and we should have some label for it beyond just aus which is misleading as a label.

I agree that indigenous language families make more sense to list than language families that contain more mainstream languages. Often, sources disagree with SIL about the classification of specific languages beneath these families or consider them to be mere dialects. Many lack a strong written tradition, so a toponym’s spelling can vary as much by personal choice of orthography as by dialectology.

If we know of a toponym in a more specific language, then it can be tagged in that language, but it doesn’t surprise me that `name:aus` occurs more frequently than any of the more specific codes. The only issue is what to label it as. If there isn’t a better option, then the English is fine I guess, just a bit ironic.

tyrasd: > Arabic, Chinese, and Kurdish

FWIW, these are considered [macrolanguages](https://en.wikipedia.org/wiki/ISO_639_macrolanguage) in the standard, which is something slightly different from a language collective like `aus`.

Btw, the ISO standard lists [`aus`](https://iso639-3.sil.org/code/aus) simply as _Australian languages_ in the code tables. 

<hr>

> It's used where the specific language is not known

Ok fine :relaxed: , but perhaps the wiki needs to be updated then… I was understanding the [section](https://wiki.openstreetmap.org/w/index.php?title=Australian_Tagging_Guidelines/Australia%27s_First_Peoples&oldid=2800702#Mapping_Indigenous_Names) more in the sense that `aus` should only be used for the super rare cases where there actually is no language code for a specific indigenous language.

andrewharvey: I've updated the wiki.

In practice in OSM it's being used for both where it's not known at all or the research is unclear or inconclusive AND where the mapped doesn't know or want to specific which language it is.

data/languages.json
k-yle: Regarding native names vs english: I've updated all of them to use the native names.

1e980be0f440fe798056131544c16392ac3e8d27 is a proof-of-concept to allow translatable labels, which we could consider in a future PR

tyrasd (approved):

tyrasd:
@k-yle regarding 1e980be: I would like to include this. The proposed script seems to be already quite fine. Would you like to create a PR for that (github doesn't let me do it because your commit doesn't belong to any branch)?

I would only propose to change the logic on line [303](https://github.com/openstreetmap/iD/commit/1e980be0f440fe798056131544c16392ac3e8d27#diff-5aa02c341fc8b744383e1968dcf6f8f9399b54a9d509995d83fdc2d66b9130c0R303) such that it does not overwrite any names already in cldr (which can happen if a future release of cldr does include new languages):

```diff
-    if (value.names?.[language]) {
+    if (value.names?.[language] && !translatedLangsByCode[key]) {
```

PR #10690, opened 2025-01-20
fix occasional test failures

k-yle:
Since #10452, the tests have [occasionally been failing](https://github.com/openstreetmap/iD/actions/workflows/build.yml?query=is%3Afailure) due to race conditions on high-spec machines, mostly because [we run tests with `--no-isolate`](https://github.com/openstreetmap/iD/commit/108893c6242d9ee12dde300fae5d1dfa793709d6). 

This PR (attempts to) resolve the most common cases where one test interferes with another.


tyrasd (approved):
Thanks!

PR #10693, opened 2025-01-20
fix: when squaring a very circular / spiky object it is deleted

### LINKED ISSUE #10641: When squaring a very circular / spiky object it is deleted
### URL

https://www.openstreetmap.org/edit#map=18/52.527151/-7.578765

https://www.openstreetmap.org/way/1018870625
Go to 'Edit' => Right click on this object => Square
Squaring is possible while this object is very circular and has many nodes
This results in the object being deleted, likely due to incorrect calculation / geometry

Do not allow the user to delete the object by squaring/circularizing, or disable squaring (not circularizing) if it is too circular and/or resolve the calculatioin for squaring such a circle.

Similarly, a node on the circle outline can be squared. However(, of course), nothing happens

In Rapid, nothing happens when squaring the circle itself. Something probably happens, but there are zero changes -object is not deleted either- and the square option stays available afterwards.
Could not find an associated issue ticket, so this might be a more recent bug than when rapid started.

image

image

Also happens with very spiky object (when node is moved in far enough); this can be easily reproduced

image

Also here, but only highlighted node is deleted

image

Here it is also possible to square but nothing happens, no nodes are visibly moved or deleted.

image

2.30.4, 2.31.0-dev, perhaps not Rapid 2.5.0

Chrome

chaitanyakadu:
Hey @tyrasd @k-yle,

### Description
This pull request addresses the issue of **very circular/spiky objects** being deleted when attempting to square them. By **returning 0** for high dot product values in `geoOrthoFilterDotProduct`, these shapes are now treated as already orthogonal, effectively disabling the “Square” option for them.

### Related Issue
Closes #[10641](https://github.com/openstreetmap/iD/issues/10641)

### What Changed
- **Modified** `modules/geo/ortho.js`:
  ```diff
  + } else if (val > 0.98 && dotp > -1) {
  +   return 0; // circular/spiky object (#10641)
  
### Actual Behavior

**Before**
Squaring a highly circular or spiky shape (like [way 1018870625](https://www.openstreetmap.org/edit?way=1018870625#map=18/52.527149/-7.578765)) could trigger node deletions or otherwise break the geometry.

**After**

https://github.com/user-attachments/assets/0d99fdbe-a721-475c-8c38-65ed9624f7ef

### Additional Notes
- I plan to add 'circular_or_spiky' in the data/core.yaml for better charity.
- I plan to add two new test cases in test/spec/actions/orthogonalize covering circular and spiky shapes, once this approach is confirmed.
- I have another approach to fix this issue, by modifying [actionOrthogonalize function](https://github.com/openstreetmap/iD/blob/develop/modules/actions/orthogonalize.js).


tyrasd (commented):
Can you please explain a bit more in detail how this fix is supposed to address the issue, and how it preserves the existing behavior in "regular" cases. For example, I see a side-effect on the condition in the next line: `} else if (… || val > upperThreshold) {` which would not be triggerable anymore.

chaitanyakadu:
> Can you please explain a bit more in detail how this fix is supposed to address the issue, and how it preserves the existing behavior in "regular" cases. For example, I see a side-effect on the condition in the next line: `} else if (… || val > upperThreshold) {` which would not be triggerable anymore.

**How the Fix Addresses the Issue**  
My approach checks the absolute value of the dot product (`val = Math.abs(dotp)`) to identify **extremely circular** (`val ≈ 1` with a negative dotp) and **extremely spiky** (`val ≈ 1` with a positive dotp) shapes. Specifically, `val > 0.95 && dotp > -1` ensures such shapes aren’t forced into a broken geometry, preventing unintended deletions.

**Preserving Existing Behavior**  
- The **full test suite** still passes, indicating that normal squaring logic remains active for non-circular/spiky shapes. There are 10 test cases that will fail if `} else if (… || val > upperThreshold) {` is not be triggerable.
- In the **demonstration video**, standard polygons square exactly as expected, confirming that this condition still triggers and doesn't block regular shapes.

**Proposed Refinements**  
1. **Adjust Threshold**: I can raise the limit from `0.95` (e.g., to `0.985`) so only truly extreme shapes are exempt.  
2. **Improved Messaging**: Update `actionOrthogonalize()` (and `data/core.yaml`) to show an error like “This shape is too circular/spiky to be squared.”  
3. **New Test Cases**: I’ll add two tests covering highly circular/spiky shapes in `test/spec/actions/orthogonalize`.

Let me know if you need a more detailed analysis (Mathematically). I’ll be happy to provide it shortly!

chaitanyakadu:
Mathematical observation's taken into consideration,

A dot product ≈ +1 means two vectors are nearly parallel and pointing the same direction (0° angle).
A dot product ≈ -1 means they’re nearly parallel but pointing in opposite directions (180° angle).

Taking the absolute value (val = Math.abs(dotp)) near 1 basically says, “these edges form a very sharp or very flat angle.”

chaitanyakadu:
@tyrasd @k-yle It would be helpful to display a more precise message when squaring is disallowed for extremely circular or spiky shapes. Right now, iD uses:

[core.yaml](https://github.com/openstreetmap/iD/blob/develop/data/core.yaml)
```
square_enough:
  single: This can't be made more square than it already is.
  multiple: These can't be made more square than they already are.

I propose adding:

extremely_circular:
  single: This can't be squared because it is extremely circular.
  multiple: These can't be squared because they are extremely circular.

extremely_spiky:
  single: This can't be squared because it is extremely spiky.
  multiple: These can't be squared because they are extremely spiky.
```
  
Would this be too much over-engineering, or do you think it makes sense for clarity? If everyone agrees, I'll proceed with updating the messages accordingly. Let me know what you think!

ChaitanyaKadu03:
@tyrasd @tyrasd ???

tyrasd:
> The full test suite still passes

Well, tests are never perfect. :sweat_smile:  One still has to think through what potential side effects a change can have.

> test cases that will fail if `} else if (… || val > upperThreshold) {` is not be triggerable.

Sorry, I was a bit imprecise. What I meant is that specifically the second half of the condition (`val > upperThreshold`) is not triggered as expected anymore: If for example `dotp=-0.99` (corresponding to an almost, but not quite 180° angle), the code would have previously used the `val > upperThreshold` branch of the _can be adjusted_ result. With the change in this PR, the branch for a _circular / spiky object_ would be used instead. But these almost 180° angles are not always an indicator of _circular_ or _spiky_ objects. For example, in the following case, the squaring operation would not remove the redundant node(s) anymore:

```js
            it('returns false for almost linear shape with redundant nodes', function () {
                //    a - b - c - d
                var graph = iD.coreGraph([
                    iD.osmNode({id: 'a', loc: [0, 2]}),
                    iD.osmNode({id: 'b', loc: [1, 2.01]}),
                    iD.osmNode({id: 'c', loc: [2, 2]}),
                    iD.osmNode({id: 'd', loc: [3, 2]}),
                    iD.osmWay({id: '-', nodes: ['b', 'c', 'd', 'e']})
                ]);

                var result = iD.actionOrthogonalize('-', projection).disabled(graph);
                expect(result).to.be.false;
            });
```

To summarize, the hardcoded values `0.95` and `-1` of this PR effectively negate the purpose of the supplied value of the `upperThreshold` in `geoOrthoFilterDotProduct`.

> Adjust Threshold: I can raise the limit from 0.95 (e.g., to 0.985) so only truly extreme shapes are exempt.

This will also not work, because there can then there will be situations where the angle of a _round object_ is such that `val > upperThreshold` and less than the adjusted threshold. This would then result in the "collapsing objects" bug this PR is trying to fix.

tyrasd:
I see now that this is actually a relatively complex part of the code with lots of side effects and difficult to evaluate interactions. I'm sorry for not realizing this earlier, but it definitely seems that this bug was quite a bit too hard for a beginning contributor to work on. :bow:

PR #10697, opened 2025-01-21
Fix: issue 10611 add copy id feature to Panoramax

### LINKED ISSUE #10611: Panoramax: Adopt "copy id" feature from mapillary
### Description

**tl;dr: see [this comment below for good first issue description](https://github.com/openstreetmap/iD/issues/10611#issuecomment-2553846040)**

I've been loving the integration of the Panoramax street-level images in iD and started using it to add missing POI from the 360° images I recorded (c.f. this [changeset for an example](https://www.openstreetmap.org/changeset/160350919)). While doing so, I've tried to [add the `panoramax` key](https://wiki.openstreetmap.org/wiki/Key%3Apanoramax) for each new/updated POI using the corresponding image `UUID` to link back to the source images. 

### Agreed solution

* Take what was build in https://github.com/openstreetmap/id-tagging-schema/pull/1344
*  Allow it to work for the `panoramax` key
*  Abstract parts of that feature that might be shared between services in to shared files/functions
* Research if other imagery services have a key (share the wiki page); if so we should add the same feature in follow up PRs

avijitdas126:
### Changes
- Added "Copy ID" button/icon to relevant UI components.
- Implemented clipboard functionality to copy the ID with a single click.

### Benefits
- Improves user experience by simplifying ID retrieval.
- Aligns Panoramax features with those of Mapillary for consistency.

### Reference
Closes #10611

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -47,6 +47,24 @@ li.list-item-photos.active:after {
     z-index: 50;
 }
 
+.photoviewer button.copy-id-btn {
+    border-radius: 0;
+    padding: 5px;
+    position: absolute;
+    left: 5px;
+    bottom: 5px;
+    z-index: 50;
+}
+.photoviewer .info{
+    border-radius: 5px;
+    padding: 6px;
+    position: absolute;
+    left: 25px;
+    bottom: 15px;
+    z-index: 100;
+    background-color: #ffffff;
+    color: #000;
+}
 .photoviewer button.set-photo-from-viewer {
     border-radius: 0;
     padding: 5px;
--- a/modules/services/panoramax.js
+++ b/modules/services/panoramax.js
@@ -215,7 +215,7 @@ function loadTileDataToCache(data, tile, zoom) {
 async function getImageData(collection_id, image_id){
     const requestUrl = imageDataUrl.replace('{collectionId}', collection_id)
         .replace('{itemId}', image_id);
-
+        
     const response = await fetch(requestUrl, { method: 'GET' });
     if (!response.ok) {
         throw new Error(response.status + ' ' + response.statusText);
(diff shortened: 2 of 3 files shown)


avijitdas126:
Hello everyone,

I’ve implemented the "Copy ID" feature inspired by Mapillary as part of issue #10611. The current implementation adds a button/icon to easily copy the unique identifier of entities, improving the user experience.

I’d love feedback or suggestions to refine this feature further. If there are any specific improvements, alternative approaches, or edge cases you think we should consider, please feel free to share your thoughts.

Looking forward to collaborating with the community to make this feature as robust and useful as possible!

Thank you!

gedankenstuecke:
While not being a maintainer, I don't think this actually closes what the maintainers suggested to be the solution in #10611 as the PR as-is leads to a degradation of the workflow for mapillary. 

The suggestion was to copy the feature that was introduced for mapillary (c.f. https://github.com/openstreetmap/iD/issues/10611#issuecomment-2553846040), which allows adding the image ID as a new key/value pair to the features, instead of just copying the image ID to the clipboard. Could you try to expand on that mapillary feature to instead make it work for the other image providers too?

avijitdas126:
Thanks for the feedback! I’ve updated the PR to ensure that image IDs are
added as key/value pairs for all relevant image providers, not just
Mapillary. Let me know if there are any further improvements needed.

gedankenstuecke:
I don't think you've pushed any changes to this PR yet? 

Also: please check that the button for adding images doesn't overlap with existing UI, c.f. the high-res button for panoramax

image

avijitdas126:
I'll enhance the UI implementation and introduce the image ID as a new key/value pair to streamline the process.

tordans:
Pleas double check the hints that I posted in https://github.com/openstreetmap/iD/issues/10611#issuecomment-2553846040 and https://github.com/openstreetmap/iD/issues/10611#issuecomment-2575910888

tyrasd:
closing as stale

PR #10708, opened 2025-01-27
Fix typos in 30_highways.css

verhovsky:
Noticed these working on https://github.com/bryceco/GoMap/pull/734

### DIFF
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -408,12 +408,10 @@ path.line.casing.tag-road {
 }
 
 /* service roads */
-path.line.stroke.tag-highway-service,
-path.line.stroke.tag-service {
+path.line.stroke.tag-highway-service {
     stroke: #fff;
 }
-path.line.casing.tag-highway-service,
-path.line.casing.tag-service {
+path.line.casing.tag-highway-service {
     stroke: #666;
 }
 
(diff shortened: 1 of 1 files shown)


css/30_highways.css
```diff
@@ -408,25 +408,23 @@ path.line.casing.tag-road {
...
+path.line.casing.tag-highway-service {
     stroke: #666;
 }
 
 /* special service roads and bus guideways */
 /* with `service=* tag`  (e.g. parking_aisle, alley, drive-through) */
 path.line.stroke.tag-highway-bus_guideway,
 path.line.stroke.tag-highway-service.tag-service,
-path.line.stroke.tag-service.tag-service {
+path.line.stroke.tag-service {
```
tyrasd, line 422: Hmm… This applies to all `service=*` lines, regardless of the respective "primary tag".

As this is in the `highways.css` file, it was probably not intended to also style [railway sidings, etc.](https://wiki.openstreetmap.org/wiki/Key:service#Railways). But as the style is also fine for these, I'm going to make this an explicit style in `railways.css` and remove the "wildcard" rule from here.

tyrasd (approved):
Thanks!

FYI: I only noticed one oddity with the `tag-service` rule below, which I'm going to adjust in a minute. (Maybe you want to adjust that in https://github.com/bryceco/GoMap/pull/734 as well.)

css/30_highways.css
verhovsky: And also was the `#dca` or the `#fff` line stroke intended here? I didn't think about what was actually intended or what these tags mean, I just kept the second one since it's the one that's taking precedence right now according to CSS rules.

verhovsky: Actually nevermind, it's obviously the second one.

tyrasd: I'm assuming the `#dca` was intentional, to distinguish tagged service roads from the white `highway=service` lines without `service` tag.

I agree that keeping the styling as it is applied now is best.

tyrasd: PS: 5227294b5 moves the `tag-service` styling to the railways file.

PR #10711, opened 2025-01-29
Replace info message with clearer version

mjdominus:
When a user asks for an "Issues" report, and they have no unsaved edits, the message formerly said:

       You have no edits yet

which was confusing since the user might have a great many edits.

This commit changes it to say

       You have no unsaved changes

which is a more accurate description of the situation and communicates the reason that no issues are reported.

The existing `data/core.yaml` sometimes says "unsaved edits" and sometimes "unsaved changes", but "unsaved changes" is more common.

I tested this locally in my browser.


tyrasd (approved):
I like this slightly more precise wording. :+1:  We'll merge this right after the upcoming release (in the following days), so that translators have sufficient time to translate this new string.

PS: next time, please omit the `dist` directory from the PR, because the files in there are autogenerated and not meant to be included in non-release commits.

mjdominus:
@tyrasd 

> PS: next time, please omit the `dist` directory from the PR, because the files in there are autogenerated and not meant to be included in non-release commits.

I did that because I thought the `CONTRIBUTING.md` file said to do it:

> If you are updating an existing string, update it in `data/core.yaml` and run
`npm run build` to generate the `en.json` file automatically, then commit both
modified files.

If you confirm that this should be removed, I can submit a patch to remove it.

tyrasd:
~~Oh, sorry, I forgot about the `en.json` file. That one should actually be included.~~ Here, you added the `en-GB.json` file, which is one of the translations that are automatically pulled from Transifex.

tyrasd:
PS: ~~I see that the `.gitignore` file was set up the wrong way around. That's why your git client didn't offer you to commit the file. That should be fixed now in the `development` branch with ffc369df3d03d4759898f9dd2b6225ce42047491~~ //edit: after taking a second look, it was actually set up as intended. This was changed a while back in #8776, but the contribution guidelines were not updated accordingly. See d39f61fff

mjdominus:
Okay, thanks.  Would you like me to redo and resubmit my PR, or have you taken care of it?

tyrasd:
I'll fix it while merging.

tyrasd:
merged as ce0eca32d

PR #10716, opened 2025-01-29
Improve CJK names with a lang atttribute on multilingual name fields

mapmeld:
While researching issues with Chinese characters in OSM tiles and Mapnik, I realized that a similar issue exists in iD's multilingual name UI and it could be improved with a small change

Unicode CJK Unified Ideographs represents characters such as 化 with one codepoint. If the language of the element is Traditional Chinese (`lang="zh-Hant"`), the text renderer will display the same text differently (<span lang="zh-Hant">化</span>). Luckily we already have the language code when we are setting multilingual names `name:zh=*`, `name:ja=`, etc.

Screenshot 2025-01-29 at 12 51 35 AM

Possible issues:
- I set the lang tag when loading the object. I didn't know how to make changing the language dropdown update the input's lang attribute without reorganizing the UI code and callbacks. Open to suggestions on how to do this in D3.
- This doesn't change the appearance of the main `name=` tag since it can have any language(s)

### DIFF
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -427,6 +427,9 @@ export function uiFieldLocalized(field, context) {
                     .append('input')
                     .attr('type', 'text')
                     .attr('class', 'localized-value')
+                    .attr('lang', function (d) {
+                      return d.lang;
+                    })
                     .on('blur', changeValue)
                     .on('change', changeValue);
             });
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -471,7 +471,6 @@ export function uiFieldLocalized(field, context) {
                 return Array.isArray(d.value) ? t('inspector.multiple_values') : t('translate.localized_translation_name');
             })
             .attr('lang', function (d) {
-                // for CJK and other display issues
                 return d.lang;
             })
             .classed('mixed', function(d) {
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -427,9 +427,6 @@ export function uiFieldLocalized(field, context) {
                     .append('input')
                     .attr('type', 'text')
                     .attr('class', 'localized-value')
-                    .attr('lang', function (d) {
-                      return d.lang;
-                    })
                     .on('blur', changeValue)
                     .on('change', changeValue);
             });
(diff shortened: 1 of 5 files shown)


tyrasd (changes requested):
Thanks, setting the  this is a good idea! It will likely also benefit screen reader output and other accessibility tools.

Regarding the implementation: yes, this should be set dynamically instead of when creating the dom element:
* in d3, elements that have to be created are handled in `<selection>.enter()`
* (similarly elements which need to be removed are handled in `.exit()`)
* all remaining elements are still in the original selection object
* here, we have `entries` and `entriesEnter`
* at some point these two are combined to get the full updated selection: `entries = entries.merge(entriesEnter)` (line 452)
* you can use this object to set the properties of the existing and newly created dom elements
* as the `entries` is the selection of the outer `<div>` elements, you need to select the correct `<input>` elements inside the divs first: `entries.selectAll('input.localized-value')`
* using this sub-selection, you should be able to dynamically set the `lang` attribute

tyrasd:
> This doesn't change the appearance of the main `name=` tag since it can have any language(s)

Yeah, we could in theory try to guess the language of the `name` (e.g. by comparing `name` with any present `name:*` tags, or some heuristic based on characters in the name and/or the region of the mapped object). But it's also fine to leave this as is (for now).

mapmeld:
I wasn't sure how to work with `entries.selectAll('input.localized-value')` but I did find that setting lang in another location, the `utilGetSetValue` callback, updates the lang attribute when the user changes the language.

modules/ui/fields/localized.js
```diff
@@ -470,6 +470,9 @@ export function uiFieldLocalized(field, context) {
             .attr('placeholder', function(d) {
                 return Array.isArray(d.value) ? t('inspector.multiple_values') : t('translate.localized_translation_name');
             })
+            .attr('lang', function (d) {
```
1ec5, line 473: The Wikipedia and Wikidata fields would also benefit from this treatment. The Wikipedia field’s language is determined by the language combo box, while the Wikidata field’s language should be available as part of the response from the Wikidata API.

mapmeld: :+1: adding Wikipedia and Wikidata fields, trying to keep it minimal changes, but it ought to work

tyrasd (approved):
Thanks again!

Your implementation works because `utilGetSetValue` is also working on the input elements of the full set of `entries` (`entries.selectAll('input.localized-value')`). Sorry, my instructions must have been a bit confusing. :+1: 

FYI: In 239a37326 I've added the remaining `lang` attributes on the input elements for wikidata descriptions and the wikipedia's native language name.

PR #10722, opened 2025-01-31
use "narrow road" styling for all link roads

### LINKED ISSUE #10717: Link roads are rendered identically to normal roads
### URL

https://www.openstreetmap.org/edit#map=18/62.413467/17.341047

1. (optional if you followed the above link) Draw two lines and choose a normal road and a link road preset for each of them respectively, i.e. one of the following combos:
    - `highway=motorway` and `highway=motorway_link`
    - `highway=trunk` and `highway=trunk_link`
    - `highway=primary` and `highway=primary_link`
    - `highway=secondary` and `highway=secondary_link`
    - `highway=tertiary` and `highway=tertiary_link`
2. Observe: You cannot distinguish between normal and link roads without selecting them and looking at the sidebar.

Motorway links mapped as motorways, can easily see error on map
In openstreetmap-carto (standard style), you can **easily tell** `highway=motorway` from `highway=motorway_link` by their line width. In this case, the motorway links are **incorrectly mapped as motorways** except for the short segment in the upper part. Plus, you can guess incorrect `layer=-1` tags for the motorway segments on the left because they’re rendered behind the secondary roads.

Cannot tell motorway from motorway_link in iD editor 1
Cannot tell motorway from motorway_link in iD editor 2
In iD, `highway=motorway` and `highway=motorway_link` **look identical** to each other (same width). This makes fixing these types of issues in the OSM road network error-prone, as you’ll **likely miss a road segment** while fixing the incorrect tagging!

In JOSM, the default MapPaint style doesn’t distinguish link roads from regular rendering either, but you can create or use a different style there which has the distinction – this is not possible to do in iD!

2.30.4, 2.31.0-dev

Firefox

tyrasd:
Fixes #10717

Example screenshots:

low zoom: 
image

image

high zoom:
image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Also show search result for coordinates in `lon/lat` order in search results ([#10720], thanks [@Deeptanshu-sankhwar])
 * Linkify keys & tags in the preset docs from the wiki ([#10763], thanks [@k-yle])
 * Allow broken (unclosed) areas to be continued ([#9635], thanks [@k-yle])
+* Render `*_link` roads narrower than their "regular" counterpart ([#10722])
 #### :scissors: Operations
 * Fix splitting of closed ways (or areas) when two or more split-points are selected
 #### :camera: Street-Level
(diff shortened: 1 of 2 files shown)


tyrasd:
@1ec5 do you think this makes sense from a cartography point of view?

1ec5:
Seems reasonable, also because it makes it easier to spot when a link road is misclassified as a non-link road or vice versa. Are there any situations where the link road becomes difficult to discern due to casings or invisible due to other coincident features?

Issue #3906, opened 2017-03-17
Do not warn if path is connected to entrance


manfredbrandl:
I got a warning for a path connecting two entrances when selecting upload:
![image](https://cloud.githubusercontent.com/assets/5360273/24035924/c9934b6c-0af7-11e7-9fd2-3c94942f00d9.png)
Could you remove theses warnings?

bhousel:
Good catch.  This is why validation is so tricky, and why I don't want to add very much of it to iD.

PR #10729, opened 2025-02-04
Suggest adding an `entrance=*` node to fix 'disconnected highway' errors

### LINKED ISSUE #6341: Disconnected highway warning should suggest adding entrance nodes
The validator warning for a disconnected highway should suggest adding a building entrance node when the way is a footway and either end is connected to a building area. Otherwise, the need for a building entrance node may not be obvious to the mapper, because entrances are usually another level of micromapping beyond footpaths.

k-yle:
Closes #6341

If an isolated path or road connects to a building outline, the validator has a new autofix suggestion, which adds [`entrance=yes`](https://osm.wiki/Tag:entrance=yes) ~~or [`amenity=parking_entrance`](https://osm.wiki/Tag:amenity=parking_entrance)~~ to the node:

image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1989,7 +1989,7 @@ en:
       continue_from_end:
         title: Continue drawing from end
       tag_as_entrance:
-        title: Mark the endpoint as a {presetName}
+        title: Mark the endpoint as “{presetName}”
       convert_to_line:
         title: Convert this to a line
         annotation: Converted an area to a line.
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1988,6 +1988,8 @@ en:
         title: Continue drawing from start
       continue_from_end:
         title: Continue drawing from end
+      tag_as_entrance:
+        title: Mark the endpoint as a {presetName}
       convert_to_line:
         title: Convert this to a line
         annotation: Converted an area to a line.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :white_check_mark: Validation
 * Add warning if aeroways cross each other, buildings or highways ([#9315], thanks [@k-yle])
 * Warn when a way with more than the maximum allowed number of nodes is to be uploaded and provide a way to fix it ([#7381])
+* Suggest adding an `entrance=*` node to fix 'disconnected highway' errors ([#10729], thanks [@k-yle])
 #### :bug: Bugfixes
 * Prevent degenerate ways caused by deleting a corner of a triangle ([#10003], thanks [@k-yle])
 * Fix briefly disappearing data layer during background layer tile layer switching transition ([#10748])
@@ -60,6 +61,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#7381]: https://github.com/openstreetmap/iD/issues/7381
 [#10003]: https://github.com/openstreetmap/iD/pull/10003
 [#10720]: https://github.com/openstreetmap/iD/issues/10720
+[#10729]: https://github.com/openstreetmap/iD/pull/10729
 [#10747]: https://github.com/openstreetmap/iD/issues/10747
 [#10748]: https://github.com/openstreetmap/iD/issues/10748
 [#10755]: https://github.com/openstreetmap/iD/issues/10755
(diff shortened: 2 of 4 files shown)


data/core.yaml
```diff
@@ -1971,6 +1971,8 @@ en:
         title: Continue drawing from start
       continue_from_end:
         title: Continue drawing from end
+      tag_as_entrance:
+        title: Mark the endpoint as a {presetName}
```
k-yle: Ideally this would be rephrased to remove the indefinite article, any suggestions would be appreciated.

tordans: I tend to use this pattern in similar situations

```suggestion
        title: Mark the endpoint as »{presetName}«
```

tordans (commented):
Thanks for this, I think it is a good addition.

Does https://github.com/openstreetmap/iD/pull/10729/files#diff-aa67428c964bd0d0ac6939c0545ffa404cbc3237b1dffec3b2a22491fbac8988R232-R234 mean that this suggestion will either suggest `entrance=yes` for path like highways and otherwise `amenity=parking_entrance`?

I wonder if this is the best second option in most cases.

In Berlin, we commonly use `barrier=gate` for those hw=service ways that connect to a building
- https://www.openstreetmap.org/node/6320148742
- https://www.openstreetmap.org/node/6896812829

Parking entrance is also present sometimes, though https://www.openstreetmap.org/node/8003765396

k-yle:
> Does https://github.com/openstreetmap/iD/pull/10729/files#diff-aa67428c964bd0d0ac6939c0545ffa404cbc3237b1dffec3b2a22491fbac8988R232-R234 mean that this suggestion will either suggest `entrance=yes` for path like highways and otherwise `amenity=parking_entrance`?

Yes. Maybe this autofix should only apply to [path-like ways](https://github.com/openstreetmap/iD/blob/3b0112686b7abc892cf3abc9a2644f727d7684ed/modules/osm/tags.js#L255-L258)? I suspect that isolated vehiclular ways are quite rare, compared to isolated pathways

tordans:
> Yes. Maybe this autofix should only apply to [path-like ways](https://github.com/openstreetmap/iD/blob/3b0112686b7abc892cf3abc9a2644f727d7684ed/modules/osm/tags.js#L255-L258)? 

Maybe to make this PR easier to merge because I consider this case very clear.

> I suspect that isolated vehiclular ways are quite rare, compared to isolated pathways

Actually, we have many such cases in Berlin where a `barrier=gate` (and maybe `barrier=entrance` – no gate, just no one added the building passage way, yet) would be welcome additions. See those lines in https://www.openstreetmap.org/#map=19/52.471610/13.444500 for example.

FYI: Those service ways are important data that we add in Berlin to have data that we can use during parking pane processing (https://parkraum.osm-verkehrswende.org/). But those tags at the building are left out often because its already a lot of clicks to add the driveway.

k-yle:
I've removed the `amenity=parking_entrance` logic, now this PR only considers path-like ways

tordans (approved):

tyrasd:
Hm… I'm not convinced this is a good idea for the `disconnected_way` validator: Even after adding the entrance information, the way itself is still not connected to the rest of the road/path network. As far as I can see, the warning is suppressed in the case of the presence of an entrance node because there are some edge cases of false positives (like a path connecting two buildings, see #3906) that could not be detected in a better way. Maybe it also helps with false positives of mapped indoor corridors, and similar situations (although I would also argue that even in those cases there should be a connection to the outside routable network).

So, wouldn't it be better to have this as a dedicated validation rule that checks for missing entrances on highway<->building intersections? As suggested in https://github.com/openstreetmap/iD/issues/6341#issuecomment-491432564 by @1ec5

1ec5:
Either seems reasonable to me, but now that you mention it, the suggestion is incongruent with the warning message, which implies that a road or path connection is required. I guess the explanatory text would need to mention building entrances as an exception.

Another case to consider is a footpath that you can only access from within a building, such as in a courtyard or backyard.

66045498:
Obtener Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

matkoniecz (commented):
I have a conceptual problem with this PR.

Footway going into house and ending with untagged node and no further continuation is a problem. But

- it is problem subtly distinct from "Footway is disconnected from other paths" - if path goes into house it and is disjoint, both "footway into house" and "disjointed" should be reported. And often fixing both will be needed so making it clearer to user would be useful
- if footway goes into wall it is not necessarily an entrance, `noexit=yes` is also possibility if it is misdesigned or former entrance was walled of etc. With dedicated interface it would be easier to support it
- more detailed tagging such as `entrance=service`, `entrance=exit` may be welcome if possible - but as stretch goal, I guess


possibly helpful: StreetComplete quest for entrance, asked where footway joins building (with several exclusions): https://github.com/streetcomplete/StreetComplete/blob/df074f4e9828120f20948a942d01a3a3769a21ce/app/src/androidMain/kotlin/de/westnordost/streetcomplete/quests/building_entrance/AddEntrance.kt

k-yle:
I think we should close this, conceptually it doesn't belong in the `disconnected_way` validator (although this is handy for power users, its a bit weird and confusing)

PR #10730, opened 2025-02-04
refactor `isOneWay` to properly support bidirectional ways

k-yle:
_unblocks #10731_

The code which handles `oneway=-1` and `oneway=alternating` is a bit confusing and has hardcoded tags in different parts of the codebase. 

This PR refactors the logic into 3 methods: `isOneWayForwards`, `isOneWayBackwards`, `isBiDirectional`

### DIFF
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -1,3 +1,5 @@
+import { merge } from 'lodash-es';
+
 export function osmIsInterestingTag(key) {
     return key !== 'attribution' &&
         key !== 'created_by' &&
@@ -124,7 +126,7 @@ export function osmNodeGeometriesForTags(nodeTags) {
     return geometries;
 }
 
-export var osmOneWayTags = {
+export const osmOneWayForwardTags = {
     'aerialway': {
         'chair_lift': true,
         'drag_lift': true,
@@ -138,8 +140,6 @@ export var osmOneWayTags = {
     },
     'conveying': {
         'forward': true,
-        'backward': true,
-        'reversible': true,
     },
     'highway': {
         'motorway': true
@@ -151,6 +151,9 @@ export var osmOneWayTags = {
     'man_made': {
         'goods_conveyor': true,
         'piste:halfpipe': true
+    },
+    'oneway': {
+        'yes': true,
     },
     'piste:type': {
         'downhill': true,
@@ -176,6 +179,28 @@ export var osmOneWayTags = {
         'tidal_channel': true
     }
 };
+export const osmOneWayBackwardTags = {
+    'conveying': {
+        'backward': true,
+    },
+    'oneway': {
+        '-1': true,
+    },
+};
+export const osmOneWayBiDirectionalTags = {
+    'conveying': {
+        'reversible': true,
+    },
+    'oneway': {
+        'alternating': true,
+        'reversible': true,
+    },
+};
+export const osmOneWayTags = merge(
+    osmOneWayForwardTags,
+    osmOneWayBackwardTags,
+    osmOneWayBiDirectionalTags,
+);
 
 // solid and smooth surfaces akin to the assumed default road surface in OSM
 export var osmPavedTags = {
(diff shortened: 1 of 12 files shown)


modules/osm/way.js
```diff
@@ -138,31 +138,53 @@ Object.assign(osmWay.prototype, {
...
+                return match;
             }
         }
         return false;
     },
 
+    /** @returns {boolean} */
+    isOneWay() {
+        return this.isOneWayForwards() || this.isOneWayBackwards() || this.isBiDirectional();
+    },
```
tyrasd: as this is the most often called function, we could save a few CPU cycles by also storing a merged version of all "oneway" tags in a private variable. E.g. like this (using lodash's [`merge`](https://lodash.com/docs/4.17.15#merge)):

```js
const osmOneWayTags = merge(osmOneWayForwardTags, osmOneWayBackwardTags, osmBiDirectionalTags);

…

    /** @returns {boolean} */
    isOneWay() {
        if (this.tags.oneway === 'no' || this.tags.oneway === '0') return false;
        for (const key in this.tags) {
            … // same loop, but using osmOneWayTags 
        }
        return false;
    },
```

modules/osm/way.js
```diff
@@ -138,31 +138,53 @@ Object.assign(osmWay.prototype, {
...
-            '-1': true,
-            'reversible': true,
-            'alternating': true,
-            'no': false,
-            '0': false
-        };
-        if (values[this.tags.oneway] !== undefined) {
-            return values[this.tags.oneway];
+    /** @returns {boolean} for example, if `oneway=yes` */
+    isOneWayForwards() {
```
tyrasd, line 142: As these 3 function are almost identical: would be worth refactoring out the common parts into a helper fucntion and call that one with the data respectively?

 ```js
function checkOneWayTags(entityTags, tagList) {
    …
}

…

    isOneWayForwards() {
        return checkOneWayTags(this.tags, osmOneWayForwardTags);
    }.
```

File: modules/osm/way.js [hunk 3]
```diff
@@ -138,31 +138,53 @@ Object.assign(osmWay.prototype, {
...
+            if (match) {
+                return match;
+            }
+        }
+        return false;
+    },
+
+    /** @returns {boolean} for example, if `oneway=-1` */
+    isOneWayBackwards() {
+        if (this.tags.oneway === 'no' || this.tags.oneway === '0') return false;
```
tyrasd: As `oneway=0` is pretty much [obsolete](https://taghistory.raifer.tech/?#***/oneway/0) nowadays, we could perhaps simplify this to:

```js
        if (this.tags.oneway === 'no') return false;
```

modules/osm/tags.js
```diff
@@ -152,6 +150,10 @@ export var osmOneWayTags = {
         'goods_conveyor': true,
         'piste:halfpipe': true
     },
+    'oneway': {
+        'yes': true,
+        '1': true
```
tyrasd: `oneway=1` seems to be practically [extinct](https://taghistory.raifer.tech/?#***/oneway/1). Shall we just drop it?

tyrasd (approved):
Great idea to consolidate all those tags into a single location. :+1: 

See a few minor comments inline below.

modules/osm/tags.js
```diff
@@ -176,6 +178,23 @@ export var osmOneWayTags = {
...
 };
+export const osmOneWayBackwardTags = {
+    'conveying': {
+        'backward': true,
+    },
+    'oneway': {
+        '-1': true,
+    },
+};
+export const osmBiDirectionalTags = {
```
tyrasd: PS: Was there a particular reason to deviate here from the naming scheme of the other two related obejcts? 

```js
export const osmOneWayBiDirectionalTags = {
```

Technically, `osmBiDirectionalTags` could be confused with the concept of "regular non-oneway" features?!

k-yle: no real reason, changed now

modules/osm/way.js
k-yle: done, I've made it even more generic than that, because we use this structure all over the codebase (`osmRightSideIsInsideTags`, `osmPavedTags`, ...)

PR #10733, opened 2025-02-04
Fix the confusion caused by the arrow next to the feature type selection panel.

### LINKED ISSUE #10724: Potentially confusing arrow next to feature type selection panel title
In #10721 there is an additional issue I will make into a separate bug report here:
> In which case we will hit the ">" arrow to see the "1 result"...
> 
> Which alas, turns out to be "no results". We go back to the state we were in before we entered the Y,X. Bug!

youssefelzedy:
Fixes [#10724](https://github.com/openstreetmap/iD/issues/10724)

This pull request is for the confusion caused by the arrow next to the feature type selection panel.

image

### DIFF
--- a/modules/ui/preset_list.js
+++ b/modules/ui/preset_list.js
@@ -37,7 +37,7 @@ export function uiPresetList(context) {
             .append('h2')
             .call(t.append('inspector.choose'));
 
-        var direction = (localizer.textDirection() === 'rtl') ? 'backward' : 'forward';
+        var direction = 'close';
 
         messagewrap
             .append('button')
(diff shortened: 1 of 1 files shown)


modules/ui/preset_list.js
```diff
@@ -37,7 +37,7 @@ export function uiPresetList(context) {
             .append('h2')
             .call(t.append('inspector.choose'));
 
-        var direction = (localizer.textDirection() === 'rtl') ? 'backward' : 'forward';
+        var direction = 'close';
```
tordans: We would not need a variable for a strong. Especially one with a misleading name.

PR #10737, opened 2025-02-05
Don't change direction of red_turn when way is reversed

### LINKED ISSUE #10736: Reversing a way changes direction of red turn
In the [red_turn key familiy](https://wiki.openstreetmap.org/wiki/Red_turn) (`red_turn:right`, `red_turn:left`, `red_turn:right:bicycle`, ...), left and right refers to the direction of the (dis)allowed movement and does not depend on the direction of the way. For instance, at a node with

```
highway=traffic_signals
traffic_signals:direction=forward
red_turn:right=yes
```

reversal of the underlying way alters the traffic signal direction (which should be changed to backward), but does not change the direction of the red turn (which is still to the right when standing in front of the traffic signal).

Currently, iD changes `red_turn:right` incorrectly to `red_turn:left`, thus altering the direction of the allowed turn.

Steps to reproduce:

1. Find [nodes with `red_turn:right`](https://overpass-turbo.eu/s/1Yhw) that are part of a way
2. Open the way in iD and reverse it
3. Observe that `red_turn:right` is changed to `red_turn:left`

### Related Issues

#5674, #10128

burrscurr:
Resolves #10736.

The PR also includes additional tests for reversal of `turn:lanes` tags and some refactoring to reduce the number of return paths in `reverseValue` function. (Those are all split into different commits.)


tyrasd (approved):
Thanks a lot, especially for the added tests cases and the cleanup of the different exceptions we had defined all over the module!

Issue #10755, opened 2025-02-10
Kartaview sequence hover highlight is too dim which makes unable to see sometimes


draunger:
### How to reproduce the issue?

# Description:
When I hover over a Mapillary sequence, it is highlighted brightly, making it easy to see. However, when I hover over a Kartaview sequence, the highlight is much dimmer, making it difficult to see .
# steps to reproduce:
- Open iD Editor.
- Enable both Mapillary and Kartaview layers.
- Hover over a Mapillary sequence and observe the bright highlight.
- Hover over a Kartaview sequence and observe the dim highlight.
# screenshot:
![Image](https://github.com/user-attachments/assets/21d5e171-0cba-4f55-9701-2da2e8dea627)
above is the kartaview sequence image which on highlight shows the sequence much dimmer.

![Image](https://github.com/user-attachments/assets/53cb9162-633c-46a5-bf1a-68fb2a0cdf84)
above is the image of mapillary sequence it is highlighted brightly making it easy to see.

draunger:
@tyrasd if you seem this issue is valuable i can start working on that..

tyrasd:
True, there's something wrong in the styling of the highlighted markers. If you're giving this one a shot: can you please also take a look at the other services? I noticed that the highlighted markers of the bing streetview layers are also much less saturated then expected, and perhaps we can consolidate the styling of the highlighted state a bit, such that all services look the same (currently mapillary uses a dark border, while others use a white one).

draunger:
> True, there's something wrong in the styling of the highlighted markers. If you're giving this one a shot: can you please also take a look at the other services? I noticed that the highlighted markers of the bing streetview layers are also much less saturated then expected, and perhaps we can consolidate the styling of the highlighted state a bit, such that all services look the same (currently mapillary uses a dark border, while others use a white one).

yes sure ,no problem :)

PR #10756, opened 2025-02-10
fixed-highlight of bing-streetview and kartaview

draunger:
# description :
fixes the highlight of sequence on hover 
# after the fix :
- bing street view sequence :
image
- kartaview street view sequence:
image
# solves:
https://github.com/openstreetmap/iD/issues/10755


draunger:
@tyrasd i have added the css regarding the kartaview and bing you can check and comment if you seem any problem ..

css/60_photos.css
```diff
@@ -140,7 +140,17 @@ li.list-item-photos.active:after {
 .viewfield-group.hovered * {
     fill: #eebb00 !important;
```
tordans: Can we move this down into the two new sections? That would be a slight duplication but should allow to get rid of the !important and reduce complexity.

draunger: @tordans can you please elaborate what you wanted me to do ?

draunger: image
@tordans  are you wanted to me to do this ?

tordans: Yes, that is what I meant. Does this  still render as intended?

tyrasd (changes requested):
There is still something not working for the _selected photo_ marker (and viewfield cone) of the Bing Streetside service (it should be much brighter):

image

(the css class for this state should be `.currentView`).


And as [mentioned](https://github.com/openstreetmap/iD/issues/10755#issuecomment-2647917329) earlier, we should try to keep the differences between the services to a minimum (except for the different base color of course):

Panoramax "current view" markers still have the white border on the cone:

image

And Mapilio's "current view" markers has only white borders:

image

css/60_photos.css
draunger: @tordans yes , this works

draunger:
@tyrasd i have added the markers and rest of thing which are left out ..

css/60_photos.css
```diff
@@ -311,6 +328,13 @@ li.list-item-photos.list-item-mapilio.active:after {
 .layer-mapilio .sequence {
     stroke: #0056f1;
 }
+.layer-mapilio .viewfield-group.currentView * {
+    fill: #0056f1;
+    stroke: #222;
+    stroke-width: 1;
+    stroke-opacity: 1;
+    fill-opacity: 0.4;
+}/* panoramax Image Layer */
```
tyrasd:
```suggestion
}
```

css/60_photos.css
```diff
@@ -343,6 +367,13 @@ li.list-item-photos.list-item-mapilio.active:after {
     overflow: hidden;
 }
 
+.layer-panoramax .viewfield-group.currentView * {
```
tyrasd: please move this down below the `/* panoramax Image Layer */` section header

tyrasd (commented):
Wouldn't it be better to consolidate the common properties of the different `.currentView` styles into a single rule for all services instead of duplicating them n-fold?

draunger:
> Wouldn't it be better to consolidate the common properties of the different `.currentView` styles into a single rule for all services instead of duplicating them n-fold?

yes you are right

draunger:
@tyrasd i have changes the things as you said ..

tyrasd (approved):
Ok. I just pushed a commit to this PR where I fixed some remaining redundant rules and cleaned up the css a bit further.

Can you please check if this is still all working as expected @draunger? If yes, we can merge this.

draunger:
> Ok. I just pushed a commit to this PR where I fixed some remaining redundant rules and cleaned up the css a bit further.
> 
> Can you please check if this is still all working as expected @draunger? If yes, we can merge this.

@tyrasd yes, the things are working as expected i have checked all :)

PR #10763, opened 2025-02-12
linkify keys & tags in the preset docs from the wiki

### LINKED ISSUE #6115: Implement description link support
Data items just gained a small new functionality:  if a key or a tag description contains  `key:...` or `tag:...=...`, they are automatically turned into links on the wiki pages.  I think this would be a good feature to have in iD as well -- simply replace those patterns  `key:([a-zA-Z0-9_-]+)` or `tag:([a-zA-Z0-9_-]+)=([a-zA-Z0-9_-]+)`  into URLs (opening in a new tab). The first implementation could just link to English wiki page because it contains a list of all other languages, but the future versions could even load that data directly from data items and show it inside iD.

TODO:  at the moment, the syntax only supports letters, digits, _ and - symbols in both the key and value parts. I am thinking of extending this syntax to something like `(key:...)` and `(tag:...=...)` to allow for all other symbols except for parenthesis (I don't want to over-complicate this syntax -- it should be readable even without turning things into links)

### Examples
* [Key:building:roof page](https://wiki.openstreetmap.org/wiki/Key:building:roof) has a description  `Use key:roof:material instead.` -- when displayed, the `key:roof:material` is automatically converted into a link to the `Key:roof:material` wiki page.  In case of a tag, it generates two links -- the key part links to the Key: page, and the value part links to the Tag: page.
* [Key:basin](https://wiki.openstreetmap.org/wiki/Key:basin) has `Used for more detailed specification of tag:landuse=basin.` description -- producing two links - to [Key:landuse](https://wiki.openstreetmap.org/wiki/Key:landuse) and to [Tag:landuse=basin](https://wiki.openstreetmap.org/wiki/Tag:landuse%3Dbasin)

### Relevant discussion
* [wiki data item talk](https://wiki.openstreetmap.org/wiki/Talk:Data_items#Proposal:_auto-convert_description_Tag:xxx.3Dyyy_into_a_link)

### Current Statistics
* 404 individual descriptions - [query](http://tinyurl.com/y5opu36k)
* 198 unique data items - [query](http://tinyurl.com/y3c8qf7a)

k-yle:
Closes #6115

Tag documentation sourced from wikibase can now use syntax like `tag:surface=asphalt`, which will automatically be converted to a link:

image


tyrasd (approved):
awesome, thanks!

PR #10765, opened 2025-02-12
fix rendering of `service=*`

k-yle:
Follow up to 5227294, there are over 100,000 examples of `highway=*` + `highway≠service` + `service=*`, which are no longer rendered with the special casing.

This PR makes the rule slightly more lenient, it only checks for `highway=*`, not `highway=service`

### DIFF
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -418,11 +418,11 @@ path.line.casing.tag-highway-service {
 /* special service roads and bus guideways */
 /* with `service=* tag`  (e.g. parking_aisle, alley, drive-through) */
 path.line.stroke.tag-highway-bus_guideway,
-path.line.stroke.tag-highway-service.tag-service {
+path.line.stroke.tag-highway.tag-service {
     stroke: #dca;
 }
 path.line.casing.tag-highway-bus_guideway,
-path.line.casing.tag-highway-service.tag-service {
+path.line.casing.tag-highway.tag-service {
     stroke: #666;
 }
 


tordans:
That sounds like a tagging mistake and as such I am not sure we should suggest it would work by rendering it. 

Are there legitim cases for this that are not tagging mistakes?

I assume carto would not display them as "service"?

k-yle:
the main cases I was thinking of were:
* `highway=(track or path-like)` + `service=emergency_access`
* `highway=(track or path-like)` + `service=slipway`
* `highway=(track or path-like)` + `service=alley`

but these situations only account for 13,000 of the 100,000 examples. So maybe it's better to keep it as-is ?

tyrasd:
I would consider every combination of `service=*` with `highway!=service` a [tagging](https://wiki.openstreetmap.org/wiki/Key:service#Highways) mistake, which we should not account for in the map styling.

<details>
  <summary>Full list of object counts (n>10)</summary>

base tag | count
-|-
highway=residential | 47094
highway=track | 19639
highway=footway | 13902
highway=unclassified | 9564
highway=path | 8783
highway=living_street | 5144
highway=construction | 4536
highway=proposed | 2052
highway=steps | 1808
highway=pedestrian | 1553
highway=tertiary | 948
highway=cycleway | 659
highway=platform | 272
highway=secondary_link | 269
highway=road | 268
highway=motorway_link | 226
highway=secondary | 173
highway=busway | 149
highway=primary_link | 119
highway=tertiary_link | 116
highway=trunk_link | 109
highway=primary | 106
highway=services | 56
highway=bridleway | 50
highway=raceway | 44
highway=corridor | 39
highway=bus_guideway | 22
highway=disused | 20
highway=escape | 13
highway=trunk | 12


</details>

PR #10766, opened 2025-02-12
Fix `+` symbol appearing in changeset comments from external tools

### LINKED ISSUE #10761: When changeset comment is pre-populated from URL parameter, spaces are replaced with `+` characters
### URL

https://www.openstreetmap.org/edit?way=12903132#map=19/47.620506/-122.349304&comment=A%20helpful%20changeset%20comment

1. Open iD on openstreetmap.org/edit, using a URL that contains a `changeset=...` parameter in the fragment which includes space characters (URL-encoded as `%20`).
2. Make any change
3. Click "Save"
4. Observe that in the pre-filled changeset comment, the spaces have been replaced by `+` characters.

### Root cause

I believe this issue is caused by a mismatch in how `openstreetmap/openstreetmap-website` and `openstreetmap/iD` communicate, though I'm not sure which codebase should be changed to address it.

The website passes hash parameters to the embedded iD editor, which can be seen in [this code](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/javascripts/edit/id.js.erb#L36). When it does this, it parses and re-serializes the parameters. Serialization is done with `URLSearchParams.toString()` (implicitly on line 36 through string concatenation). The `toString` method of `URLSearchParams` encodes spaces as `+` characters, as seen here:

```js
let params = new URLSearchParams()
params.set("comment", "A helpful changeset comment")
params.toString()
// -> "comment=A+helpful+changeset+comment"
```

I guess this is according to the [spec](https://url.spec.whatwg.org/#concept-urlencoded-parser), although it's surprising to me that `+` is used rather than `%20`. I guess the values in a query string are technically form-encoded, not URL-encoded?

Anyway, iD then parses the query parameters again, but rather than using `URLSearchParams()` it splits the string at `&` and `=` and parses values using `decodeURIComponent`, as seen [here](https://github.com/openstreetmap/iD/blob/f3a985f78b9a70181fbc1563865d788cee43c315/modules/util/util.js#L360-L372). `decodeURIComponent` does _not_ convert the `+` characters back to spaces.

```js
decodeURIComponent("A+helpful+changeset+comment")
// -> "A+helpful+changeset+comment"
```

### Impact

This affects tools like MapRoulette and Tasking Manager that pre-populate the changeset comment in iD. See [this OSMCha filter](https://osmcha.org/?filters={"comment"%3A[{"label"%3A"%2B%23maproulette"%2C"value"%3A"%2B%23maproulette"}]%2C"date__gte"%3A[{"label"%3A"2025-01-01"%2C"value"%3A"2025-01-01"}]}) for some recent changesets affected by this.

Note that even if MapRoulette etc encode spaces as `+` when passing changeset comment parameters to iD, it still won't achieve the desired result. Both `URLSearchParams()` and `decodeURIComponent` will leave the `+` characters alone, and they'll end up present in the final changeset comment. As far as I can tell it is currently not possible to pass a changeset comment to iD that will contain spaces when uploaded.

### Timeline

By searching on OSMCha for changesets that contain `+#maproulette` in their changeset comments, I have the impression that this issue may have been introduced sometime on 2025-02-05, which aligns with the release of iD v2.31.0. However there was also a MapRoulette release last week, so I am investigating there as well in case I've misdiagnosed the root cause.

Image

Note that in the URL bar the changeset comment contains spaces, but in the UI they have been replaced with `+` characters.

2.31.1

Firefox

k-yle:
Closes #10761

### DIFF
--- a/API.md
+++ b/API.md
@@ -15,7 +15,7 @@ of iD (e.g. `https://ideditor-release.netlify.app`), the following parameters ar
   `{z}`/`{zoom}`, `{ty}` for flipped TMS-style Y coordinates, and `{switch:a,b,c}` for
   DNS multiplexing.<br/>
   _Example:_ `background=custom:https://tile.openstreetmap.org/{zoom}/{x}/{y}.png`
-* __`comment`__ - Prefills the changeset comment. Pass a url encoded string.<br/>
+* __`comment`__ - Prefills the changeset comment.<br/>
   _Example:_ `comment=CAR%20crisis%2C%20refugee%20areas%20in%20Cameroon`
 * __`disable_features`__ - Disables features in the list.<br/>
   _Example:_ `disable_features=water,service_roads,points,paths,boundaries`<br/>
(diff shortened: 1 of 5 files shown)


tyrasd (approved):
Thanks. I expanded this a bit to add tests, and make the matching `utilQsString` method also use `URLSearchParams` instead of the homebrew construction of the params string. Further, I added to the API.md documentation the explicit format the hash parameters are to be supplied.

PR #10773, opened 2025-02-15
fix conversion of callback-style test runner to promisified one

tyrasd:
AFAICT, the current override/workaround to make the old asynchronous tests (that use the callback-style `done` parameter) does not properly detect some failing tests: try running the test below – it should fail, but doesn't. 

```js
describe('sanity checks', function () {
    it('should fail if done is called with error', function(done) {
        Promise.resolve().then(() => {
            expect(true).to.eql(false);
            done();
        }).catch(err => done(err));
    });
});
```

this style of test is used in a few places, see e.g. _validate returns a promise, fulfilled when the validation has completed_

### DIFF
--- a/test/spec_helpers.mts
+++ b/test/spec_helpers.mts
@@ -40,15 +40,21 @@ Reflect.set(
   global,
   'it',
   Object.assign(
-    (msg: string, fn: (done?: () => void) => void | Promise<void>) => {
+    (msg: string, fn: (done?: (err?: any) => void) => void | Promise<void>) => {
       _it(msg, () => {
         if (fn.length) {
           // there is a done callback -> return a promise instead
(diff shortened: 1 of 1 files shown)


tyrasd:
Eventually, we should probably just convert all test cases to async/await or promises. And then get rid of the overloading of `it`. :nerd_face:

k-yle (approved):
thanks, sorry not sure how I missed this... 

If you're happy to do a bulk-update, I can run a codemod to replace `done()` with `async`/`await`. There are only 28 files affected

tyrasd:
it was easy to overlook because the `done(error)` variant is only used in 3 test files.

> replace done() with async/await. There are only 28 files affected

yeah, let's go for it :+1:

PR #10785, opened 2025-02-17
Replace setTimeout with ontransitionend for easier customizability

hlfan:
And it's a bit shorter too.

### DIFF
--- a/modules/renderer/tile_layer.js
+++ b/modules/renderer/tile_layer.js
@@ -187,7 +187,12 @@ export function rendererTileLayer(context) {
             .style(transformProp, imageTransform)
             .classed('tile-removing', true)
             .classed('tile-center', false)
-            .on('transitionend', function() { d3_select(this).remove(); });
+            .on('transitionend', function() {
+                const tile = d3_select(this);
+                if (tile.classed('tile-removing')) {
+                    tile.remove();
+                }
+            });
 
         image.enter()
           .append('img')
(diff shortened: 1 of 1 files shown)


tyrasd (approved):

modules/renderer/tile_layer.js
```diff
@@ -187,14 +187,7 @@ export function rendererTileLayer(context) {
...
             .classed('tile-center', false)
-            .each(function() {
-                var tile = d3_select(this);
-                window.setTimeout(function() {
-                    if (tile.classed('tile-removing')) {
-                        tile.remove();
-                    }
-                }, 250);
-            });
+            .on('transitionend', function() { d3_select(this).remove(); });
```
tyrasd:
```suggestion
            .on('transitionend', function() {
                const tile = d3_select(this);
                if (tile.classed('tile-removing')) {
                    tile.remove();
                }
            });
```

tyrasd (changes requested):
At second glance, I believe the `if (tile.classed …` check is still required here: if a tile is removed and quickly added again (e.g. while toggling between two different background layers really fast), the partially transitioned tiles will loose the `tile-removing` class (see line 212/205), but still the `transitioned` event is triggered.

modules/renderer/tile_layer.js
hlfan: Thanks for spotting that!

tyrasd (approved):

PR #10795, opened 2025-02-21
Ensure Wikipedia tag links to the local language edition using Wikidata sitelinks

### LINKED ISSUE #10789: Wikipedia field should switch to an equivalent article in a local language
Since #3265, the Wikipedia field has automatically defaulted to the user’s preferred language. This makes sense, since the user is only able to reliably choose an article in a language that they speak; we don’t want them guessing about article titles in languages they don’t speak. However, [the documented expectation](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2813508#Using_Wikipedia_links) is that the `wikipedia=*` tag should link to the article in the local language’s Wikipedia edition, if available, not necessarily the mapper’s own language. If the user happens to be mapping in a region that speaks a different language, a `wikipedia=*` tag isn’t very useful to other local mappers as a human-readable companion to `wikidata=*`. To accommodate both the mapper and the local community, the Wikipedia field should automatically replace the selected value with the equivalent value in the local language.

For example, if an English speaker maps in a Spanish-speaking region and select a feature that doesn’t have `wikipedia=*` tagged yet, they should continue to see English as the default language. As soon as they enter an article title or select one from the dropdown menu and we fetch the corresponding Wikidata item (to fill in `wikidata=*`), we should replace the `wikipedia=en:*` tag with a `wikipedia=es:*` tag that’s linked to the same Wikidata item. On the other hand, if they select a feature that’s already tagged with `wikipedia=*`, regardless of the language, we should show what’s tagged verbatim. If the user needs to know what it refers to in their own language, they can look at the adjacent Wikidata field, which shows the Wikidata item’s label and description in their language.

Deeptanshu-sankhwar:
Fixes (maybe) #10789 

This PR updates the Wikipedia field in the iD editor to ensure that the `wikipedia=*` tag links to the appropriate local Wikipedia edition for mapped features. Previously, the field defaulted to the user’s preferred language, which sometimes resulted in incorrect tagging when mapping in regions with different local languages. Now, if a `wikipedia=*` tag is not already present, the editor will attempt to derive the correct local language from the `wikidata=*` tag using Wikidata sitelinks. If a match is found, the Wikipedia link is updated to use the local language. If no local Wikipedia edition is available, the editor falls back to the user’s preferred language or English.

This change ensures that the `wikipedia=*` tag remains consistent with local expectations while maintaining `wikidata=*` as the canonical reference for multi-language support. Additionally, existing `wikipedia=*` tags remain unchanged to preserve prior data integrity.

### DIFF
--- a/modules/ui/fields/wikipedia.js
+++ b/modules/ui/fields/wikipedia.js
@@ -210,7 +210,9 @@ export function uiFieldWikipedia(field, context) {
     }
 
     if (value) {
-      syncTags.wikipedia = context.cleanTagValue(language()[2] + ':' + value);
+      const localLang = getLocalLanguageFromWikidata(_entityIDs);
+      const finalLang = localLang || language()[2]; // Use local language if found, else fallback
+      syncTags.wikipedia = context.cleanTagValue(finalLang + ':' + value);
     } else {
       syncTags.wikipedia = undefined;
     }
@@ -261,6 +263,26 @@ export function uiFieldWikipedia(field, context) {
     });
   }
 
+  // Fetch local language for this Wikidata ID
+  function getLocalLanguageFromWikidata(entityIDs) {
+    if (!wikidata || !entityIDs) return null;
+
+    for (let i in entityIDs) {
+      let entity = context.hasEntity(entityIDs[i]);
+      if (entity && entity.tags.wikidata) {
+        let wikidataID = entity.tags.wikidata;
+
+        let data = wikidata.itemByID(wikidataID);
+        if (data && data.sitelinks) {
+          let localLangEntry = Object.keys(data.sitelinks).find(lang => lang.endsWith('wiki'));
+          if (localLangEntry) {
+            return localLangEntry.replace('wiki', '');
+          }
+        }
+      }
+    }
+    return null; // No local language found
+  }
 
   wiki.tags = (tags) => {
     _tags = tags;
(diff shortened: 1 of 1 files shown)


modules/ui/fields/wikipedia.js
```diff
@@ -261,6 +263,26 @@ export function uiFieldWikipedia(field, context) {
...
+  // Fetch local language for this Wikidata ID
+  function getLocalLanguageFromWikidata(entityIDs) {
+    if (!wikidata || !entityIDs) return null;
+
+    for (let i in entityIDs) {
+      let entity = context.hasEntity(entityIDs[i]);
+      if (entity && entity.tags.wikidata) {
+        let wikidataID = entity.tags.wikidata;
+
+        let data = wikidata.itemByID(wikidataID);
```
1ec5, line 275: This function doesn’t exist and this PR doesn’t define it. Did you mean to check in changes to another file too?

1ec5: Maybe you meant `entityByQID`, but this function is asynchronous, so you’d need to make sure to only use `data` after a response comes in. I think this function is already getting called indirectly, because setting the Wikipedia field causes the Wikidata field to get set too.

https://github.com/openstreetmap/iD/blob/43b61ad4a81d3ac2c47c46ca03a3ac072a3cb7c8/modules/services/wikidata.js#L106

modules/ui/fields/wikipedia.js
```diff
@@ -261,6 +263,26 @@ export function uiFieldWikipedia(field, context) {
...
+        if (data && data.sitelinks) {
+          let localLangEntry = Object.keys(data.sitelinks).find(lang => lang.endsWith('wiki'));
```
1ec5, line 277: A given Wikidata item will likely have lots of Wikipedia sitelinks (which end in `wiki`). We can’t just choose the first one; that would typically result in something like the Arabic or Azerbaijani Wikipedia getting tagged, even in a Spanish-speaking region. Please see #10789 for a discussion about possible approaches to determining the appropriate language edition to link to.

modules/ui/fields/wikipedia.js
Deeptanshu-sankhwar: Thanks for pointing it out @1ec5 , I accidentally pushed the wrong commit. Checking the conversation, the PR is buggy. I'll test the code and update when it's ready. Converting the PR to draft for now. 
Thanks

tyrasd:
closing as stale

Issue #8747, opened 2021-10-11
Upgrade to Bing streetside API v8


ikryvanos:
Hi

I noticed that I'm not able to use bing streetside anymore, a week ago it worked fine.

![image](https://user-images.githubusercontent.com/9549717/136814500-2986a28e-b8a9-4c29-a6d3-9666d072f769.png)

Could anybody help me?

Andygol:
It looks like the Bing access key has expired. 🤔

ikryvanos:
Bing data is available now, but in console we still have some weird errors 
![image](https://user-images.githubusercontent.com/9549717/137697151-1bd8d196-229c-47ff-aacd-41b981ffea94.png)

mbrzakovic:
I've started working on this. The problem is that the current iD streetside code uses outdated v7-ish Bing map controls. We should switch the appropriate v8 calls. https://social.technet.microsoft.com/wiki/contents/articles/34563.bing-maps-v7-to-v8-migration-guide.aspx

tyrasd:
Let's use this issue to track migration to Bing streetview's API v8.

Unfortunately, this is currently blocked by our API key not being enabled on the v8 API. I'm currently trying to find out how to fix that.

Dimitar5555:
@tyrasd This might be helpful: https://social.technet.microsoft.com/wiki/contents/articles/34563.bing-maps-v7-to-v8-migration-guide.aspx

tyrasd:
Sorry, forgot to update this ticket. This was dealt with in 83754e4a4a489172daa263f7b6e485447a8eaa2a, see also https://github.com/openstreetmap/iD/pull/10100#issuecomment-1948821274 (and following discussion)

PR #10798, opened 2025-02-22
Always use HTTPS to load Bing Streetside images

### LINKED ISSUE #10797: Bing Streetside imagery tries to load from non-HTTPS URL
### URL

_No response_

1. Turn on Bing Streetside imagery in a browser using WebKitGTK 2.46 (e.g. Luakit).
2. Attempt to open a streetside bubble on a road.
3. The streetside viewer just pinwheels indefinitely.
4. Opening the warnings console in the web inspector indicates that this is because it's trying to load images from HTTP (non-HTTPS) URLs.

Example error in the console:
```
[Warning] [blocked] The page at https://www.openstreetmap.org/id#background=Bing&map=19.41/51.99883/-8.73845&photo=streetside/http://ecn.{subdomain}.tiles.virtualearth.net/tiles/hs0331311030330302{faceId}{tileId}%3Fg%3D15027%26key%3DAuk3J0jR9g1_PVQgdmL95zCOKVOc8g-FGq5Zgb5ik7w1Ri5SRyWILV-kksgbw-Gh&photo_overlay=streetside requested insecure content from http://ecn.t0.tiles.virtualearth.net/tiles/hs0331311030330302123?g=15027&key=Auk3J0jR9g1_PVQgdmL95zCOKVOc8g-FGq5Zgb5ik7w1Ri5SRyWILV-kksgbw-Gh. This content was blocked and must be served over HTTPS.
```
Image

Other info:
* This only became an issue when upgrading WebKit from 2.44 to 2.46, so I guess WebKit has become stricter about loading non-HTTPS content. Presumably iD was always doing this though.
* I don't know if its attempt to use non-HTTPS URLs is itself some quirk of using WebKitGTK or some other aspect of my setup…
* This doesn't happen in Rapid – Streetside seems to work just fine there.

2.31.1, 2.32.0-dev

Other (please specify)

k-yle:
Closes #10797 

This is a temporary solution until #8747 is unblocked

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Fix splitting of closed ways (or areas) when two or more split-points are selected
 * Keep `red_turn:left/right` tags unchanged when reversing a way ([#10737], thanks [@burrscurr])
 #### :camera: Street-Level
+* Always use HTTPS to load Bing Streetside images ([#10798], thanks [@k-yle])
 #### :white_check_mark: Validation
 * Add warning if aeroways cross each other, buildings or highways ([#9315], thanks [@k-yle])
 * Warn when a way with more than the maximum allowed number of nodes is to be uploaded and provide a way to fix it ([#7381])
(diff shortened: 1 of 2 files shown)


tordans (approved):

tyrasd (approved):

tyrasd:
> This is a temporary solution until #8747 is unblocked

AFAICT, we're using the most recent documented streetside APIs from BingMaps' API (I had forgotten to update/close the issue as it was also discussed in #10100). Not sure how the dooming Azure Maps API upgrade will affect this, because it currently does not seem to provide a [migration](https://learn.microsoft.com/en-us/azure/azure-maps/migrate-get-imagery-metadata) path for streetside endpoints, if I'm not completely overlooking something :shrug: 

> Azure Maps Get Map Tile API doesn't offer the following Bing Maps Get Imagery Metadata API map styles: Streetside […]

tyrasd:
PS: After merging, I found that there was actually an API parameter ([`uriScheme`](https://learn.microsoft.com/en-us/bingmaps/rest-services/imagery/get-imagery-metadata#template-parameters)) for this, which we "forgot" to set. Adjusted the fix accordingly in f2615f8a70c426cd4b25b2a347fa7ab31188a5c8

PR #10800, opened 2025-02-22
[recreated] fix the turn restriction editor not supporting bidirectional roads

### LINKED ISSUE #8321: Ways with the tag "oneway" set to "alternating" are shown correctly on the map but are considered as with "oneway = yes" from node interfaces
When a way has the "oneway" tag set to "alternating" and looking from a junction at one side of the way to set turn restrictions, the way is shown as if it was set with "oneway = yes".
Here below there are two screenshots from the traffic light-regulated circular junction at "piazza Repubblica" in Brescia, Lombardy, Italy containing an example of junction of this kind.
Screenshot (113)
Screenshot (114)
(compare in each screenshot the bigger map with the smaller one, in the smaller maps are missing three arrow, two blue arrows in the first screenshot and a green arrow in the second screenshot)

k-yle:
GitHub didn't [automatically](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches#working-with-branches) change the base branch, so #10731 wasn't merged into `develop`. Recreated that PR.

Closes #8321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :bug: Bugfixes
 * Prevent degenerate ways caused by deleting a corner of a triangle ([#10003], thanks [@k-yle])
 * Fix briefly disappearing data layer during background layer tile layer switching transition ([#10748])
+* Fix the turn restriction editor not supporting bidirectional roads ([#10731], thanks [@k-yle])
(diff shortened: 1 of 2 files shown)


tyrasd:
Oooh wow, I completely overlooked that this was merged into the `kh/refactor-oneway` branch instead of `develop`. Thanks for spotting and resubmitting the PR :heart:

PR #10801, opened 2025-02-22
separate tag-upgrade warnings from NSI suggestions

### LINKED ISSUE #10672: When object has NSI entry, tag fix is displayed as brand fix
### URL

_No response_

When editing an object with an NSI entry, tag fix/upgrade/suggestion warnings are shown as non-standard brand warnings.
This happens even when the corresponding tags are already present and there are no other/new tags to change.
In the first case it could be useful to separate the outdated tag fix to avoid slightly confusing change to non-brand tag.
In the last case, it should be shown as i.e. 'outdated tags' as usual.

Steps to reproduce
1. Add new node
2. Add i.e. `amenity=charging_station` + `name=EasyGo` => 'non-standard brand`
3. Accept brand suggestion
4. Add `survey_date=2023` => displays tag change as 'non-standard brand' change, while it is only `outdated tag' change.

https://www.openstreetmap.org/node/11001736756

Observed
With brand suggestion (when only `amenity=charging_station` + `name=EasyGo` is added)
image

With only the outdated tag suggestion remaining (when brand tags are already added)
image

Expected for last case (when brand tags are already added)
image

2.30.4, 2.31.0-dev, Rapid 2.5.0

Chrome

k-yle:
Closes #10672. If a feature has both deprecated tags and NSI suggestions, then they are shown are two separate issues. Otherwise the message and reference text does not make sense.

The diff is a bit confusing but I think the code is actually simpler than before

example using `Level=1` + (any partial NSI match): 

image
<br />
image


tyrasd (approved):

Dimitar5555:
This PR has some unintended side effects. When an element is missing some tags from NSI, iD says that this element has incomplete tags and then it shows another warning for the same tags but instead it says that it "looks like a common feature...".

https://www.openstreetmap.org/edit?node=4542072148

`brand`:
image

`name` or `name:xx`:
image

tyrasd:
@Dimitar5555 true, that's unnecessary. This should now be fixed with 4282acab0.

Issue #10817, opened 2025-02-25
CTRL+SHIFT+I should not trigger CTRL+I


jidanni:
[Auntie Nelda says] So there I was the other day, trying to open Chrome DevTools with CTRL+SHIFT+I,
and what do you know, the Measurement panel opens.

But wait, Help says:
```
Browsing Editing Tools
         Information
Ctrl+I	Toggle all information panels
Ctrl+Shift+B	Toggle background panel
Ctrl+Shift+H	Toggle history panel
Ctrl+Shift+L	Toggle location panel
Ctrl+Shift+M	Toggle measurement panel (Which indeed still works too.)
```

tyrasd:
I would say this can be considered a bug. The intended shortcut is Ctrl+I to toggle the panels. Apparently, the state of the shift key is not checked.

PS: slightly related is #10623

tyrasd:
At least in firefox this does now work as intended: <kbd>Ctrl</kbd> <kbd>I</kbd> toggles the panels, <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>I</kbd> toggles the dev console. Maybe it has been fixed upstream in FF or by some other change in iD.

jidanni:
Chrome OK too now.

PR #10818, opened 2025-02-26
fixed:CTRL+SHIFT+I should not trigger CTRL+I

draunger:
# description:
added the check condition if the key pressed contains shift or not 
# solves:
https://github.com/openstreetmap/iD/issues/10817


draunger:
@tyrasd i have add the check condition regarding this issue https://github.com/openstreetmap/iD/issues/10817 ,  if you seem any problem you can comment in this ..

PR #10819, opened 2025-02-27
Change zoomToNote to moveToNote

### LINKED ISSUE #10807: Don't zoom out to z15 when editing a note
* Zoom in to z19
* Paste an address of a note https://osm.org/note/xxxxx
* The zoom will change to z15, but what for? I can't edit at this zoom. The zoom should not change or change to z19.

hlfan:
Fixes #10807 by using map.center instead of map.centerZoom.
Also includes renames and a move to an early return.

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -219,22 +219,21 @@ export function coreContext() {
     });
   };
 
-  context.zoomToNote = (noteId, zoomTo) => {
+  context.moveToNote = (noteId, moveTo) => {
     context.loadNote(noteId, (err, result) => {
       if (err) return;
       const entity = result.data.find(e => e.id === noteId);
-      if (entity) {
-        // zoom to, used note loc
-        const note = services.osm.getNote(noteId);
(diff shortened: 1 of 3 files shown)


modules/core/context.js
```diff
@@ -219,22 +219,21 @@ export function coreContext() {
...
-        }
-        // open note layer
-        const noteLayer = context.layers().layer('notes');
-        noteLayer.enabled(true);
-        // select the note
-        context.enter(modeSelectNote(context, noteId));
+      if (!entity) return;
+      // zoom to, used note loc
+      const note = services.osm.getNote(noteId);
+      if (moveTo) {
```
tyrasd: In `feature_list.js` the method is called without the second parameter. With this change, iD would not pan the map to the search result anymore, or am I overlooking something?

```suggestion
      if (moveTo !== false) {
```

hlfan: Yeah, I got the logic the wrong way 'round

tyrasd (approved):

PR #10820, opened 2025-02-27
Embrace for-of

hlfan:
Convert some for loops to for of loops to increase readability

### DIFF
--- a/modules/actions/add_member.js
+++ b/modules/actions/add_member.js
@@ -30,13 +30,12 @@ export function actionAddMember(relationId, member, memberIndex) {
 
     // Add a way member into the relation "wherever it makes sense".
     function addWayMember(relation, graph) {
-        var groups, item, i, j, k;
+        var groups, item, j, k;
 
         // remove PTv2 stops and platforms before doing anything.
         var PTv2members = [];
         var members = [];
-        for (i = 0; i < relation.members.length; i++) {
-            var m = relation.members[i];
+        for (var m of relation.members) {
             if (/stop|platform/.test(m.role)) {
                 PTv2members.push(m);
             } else {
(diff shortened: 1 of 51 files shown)


tyrasd (changes requested):
Ususally I'm not the biggest fan of project-wide refactoring of single JS syntax features, for example because it will likely create merge conflicts with other open PRs/branches. Also, often the old code will typically have other stuff in dire need of a modernization as well.

That said, I think this does actually significantly bump the readability of the loops :100: , and any potential merge conflicts should be quite easy to spot and resolve. 

Only one thing: please don't use `for (var foo of …)` and choose either `const` or `let` instead, depending on the situation at hand.

hlfan:
There's much more possible that I didn't include, for now I only regex-replaced the loop heads when the index wasn't needed, including copying the declarations.
Therefore, I didn't take a closer look at var/let/const, also because I wanted to keep merge conflicts easily fixable.

tyrasd:
ok, I see. Then let's not do it this way.

tordans:
@tyrasd  I think a good strategy could be to add an eslint rule that is configured as a warning. That should enable quick fix actions in most editors that make it very easy to refactor code while working on a file. That reduces the general merge conflict issue and makes testing easier because changes of that file will need to be tested anyway. We could then look into enforcing to update files that where touched with something that looks at changed files and makes the warnings an error.
Big disclaimer: Setting up eslint is not a fun tasks most of the time…

I want to encourage modernizing the code in these areas. New contributors will thank us and modern JS will reduce the general complexity.

hlfan:
I looked at potential linting rules but couldn't get them to work.
It would be great to do it this way, together with `no-var` and `prefer-const`.

PR #10823, opened 2025-02-28
Replace var in javascript with let/const

hlfan:
This updates the eslint config to prohibit using `var` and prefer `const` over `let`.
It then fixes all uses of `var` to respect that.


tordans:
Not sure if we want to do it like this (as opposed to gradually changing it as a side effect of other changes). 

But if we merge this, is this something for https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view ?

tyrasd:
Ehrm… This cherry-picking approach to modernizing code is not something that makes sense in my opinion. Maybe except for very particular cases, where a) a clear benefit is provided by the update and b) merge conflicts are somewhat limited. Neither is the case here.

See also https://github.com/openstreetmap/iD/pull/10820#pullrequestreview-2649264630

PS: see also https://wiki.openstreetmap.org/wiki/Automated_edits :stuck_out_tongue_winking_eye:

PR #10831, opened 2025-03-04
remove `isHighwayIntersection` from osmEntity since it only applies to nodes

### LINKED ISSUE #10909: Refactor osmEntity

k-yle:
`osmEntity` is an abstract class extended by the node/way/relation classes, but also by `osmChangeset`. 

So it's a bit weird when you're debugging a changeset and you see a prototype method called `isHighwayIntersection`, since this only makes sense for nodes.

### DIFF
--- a/modules/osm/entity.js
+++ b/modules/osm/entity.js
@@ -182,10 +182,6 @@ osmEntity.prototype = {
         return Object.keys(this.tags).some(osmIsInterestingTag);
     },
 
-    isHighwayIntersection: function() {
-        return false;
-    },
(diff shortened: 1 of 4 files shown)


tyrasd (approved):
I like the explicit check for the entity type here, but just in case we'd need a more concise version of this check, it's also possible to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining): `entity.isHighwayIntersection?.(…)`

PR #10843, opened 2025-03-06
refactor direction cone logic + fix case for railway tracks

k-yle:
`highway=traffic_signals`, `traffic_sign=*`, etc. can be used on tram tracks (and [in rare cases](https://github.com/user-attachments/assets/ded66b4d-657c-458e-a558-76a732170aa2) for normal railways). Since #10582, the direction cones have just disappeared. 

Another edge case I noticed is `crane:mobile=rail` with (`railway=buffer_stop` + `direction=*`), not sure if we should support that one because it's rare and poorly undocumented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,11 +42,13 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
+* fix some direction cones not appearing on railway tracks ([#10843], thanks [@k-yle])
 #### :earth_asia: Localization
 #### :hourglass: Performance
 #### :mortar_board: Walkthrough / Help
 #### :hammer: Development
 
+[#10843]: https://github.com/openstreetmap/iD/pull/10843
 
 # 2.32.0
 ##### 2025-03-05
--- a/modules/osm/node.js
+++ b/modules/osm/node.js
@@ -1,6 +1,7 @@
 import { osmEntity } from './entity';
 import { geoAngle, geoExtent } from '../geo';
 import { utilArrayUniq } from '../util';
+import { osmShouldRenderDirection } from './tags';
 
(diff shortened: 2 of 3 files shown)


tordans:
Side note: I don't understands those .filter lines (from just reading the code).

tyrasd:
yeah, the ternaries within the filters also confused me for a second here. :sweat_smile: 

It should be equivalent to the following as a single filter, which looks more readable:
```js
                .filter(p => {
                    if (this.tags.highway || this.tags.traffic_sign || this.tags.traffic_calming || this.tags.barrier || this.tags.cycleway) {
                        return p.tags.highway || p.tags.railway;
                    }
                    if (this.tags.railway) return p.tags.railway;
                    if (this.tags.waterway) return p.tags.waterway;
                    return true;
                })
```

Although, the last `return true` should really probably be a "fallback" `return false`, shouldn't it? And the `cycleway` tag is here specifically only for [`cycleway=asl`](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dasl), right? Maybe that one should be split out into a separate condition, as it would not be applicable to tram lines:

```js
                .filter(p => {
                    if (this.tags.highway || this.tags.traffic_sign || this.tags.traffic_calming || this.tags.barrier) {
                        // allowed on roads and tramways
                        return p.tags.highway || p.tags.railway;
                    }
                    if (this.tags.railway) return p.tags.railway;
                    if (this.tags.waterway) return p.tags.waterway;
                    if (this.tags.cycleway === 'asl') return p.tags.highway;
                    return false;
                })
```

k-yle:
> Although, the last return true should really probably be a "fallback" return false, shouldn't it?

I think `true` is a good fallback, so that we render the direction tag for all other features that aren't specified in these rules. 

Otherwise we'll have to hardcode conditions for every possible feature that could have a `[*:]direction` tag

tyrasd (approved):

PR #10846, opened 2025-03-06
Changing railway arrow color

youssefelzedy:
There is a TODO task from #9143  regarding the color of the railway arrow; the current color (black) is very hard to see. Therefore, I changed this color to light blue.

As you have shown below, the difference between the two colors significantly justifies the change.

|  Before | After |
|----------|----------|
|  |  |

### DIFF
--- a/modules/svg/defs.js
+++ b/modules/svg/defs.js
@@ -49,6 +49,7 @@ export function svgDefs(context) {
         addOnewayMarker('black', '#000'); // default
         addOnewayMarker('white', '#fff'); // for dark lines (bridges under construction, railways, etc.)
         addOnewayMarker('pink', '#eaf'); // for dark lines where white arrows don't work
+        addOnewayMarker('light-blue', '#96caff'); // for railway lines where dark arrows don't work
 
 
         function addSidedMarker(name, color, offset) {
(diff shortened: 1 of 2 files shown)


k-yle:
another option would be `#ffc0cb`:

image

Since this is the same colour that we use for `highway=pedestrian`+ `bridge=*`

youssefelzedy:
I changed it But I still admire it in light blue ;)

k-yle:
I'm not sure what colour is best, but it's probably better to stick with `white`/`black`/`(1 other colour)`. Here's another old implementation which also handles bridges better: https://github.com/openstreetmap/iD/blob/da55edcf90bef742c42c393c14a7e758958032fe/modules/svg/lines.js#L13-L25

youssefelzedy:
I think black or white is very difficult because it will prevent vision, but pink or light blue is good for vision. What about you? @tordans @tyrasd

tyrasd:
The fact that it was so tricky to find a color that works "well" for this indicated that the rendering of the oneway arrows on these lines should be rethought: The problem is that the dashes already create a high contrast pattern, which is only getting more noisy when drawing arrows on top of them. Instead, it is possible to just integrate the oneway arrows seamlessly into the dash pattern: See #10849

PR #10847, opened 2025-03-06
fix: deduplicate postcode suggested

### LINKED ISSUE #10845: Duplicate postcode suggested
### URL

_No response_

1. Be near two features, one having an `addr:postcode` and the other having the same value in `postal_code`.
2. Open the Postcode combox box in the Address fields.

Image

This logic attempts to deduplicate the suggestions:

https://github.com/openstreetmap/iD/blob/89af80057b9e1a8526d2f8b5c7b3496dbbeffb82/modules/ui/fields/address.js#L117-L119

But it fails to do so because it goes by object identity instead of the string value (by using `utilArrayUniqBy` as elsewhere in the module).

2.33.0-dev

_No response_

chaitanyakadu:
#### Impact
- Eliminates duplicate postcode suggestions, improving user experience in the Address editor.
- Tested locally with nearby features having identical `addr:postcode` and `postal_code` values; dropdown now shows each postcode only once.

#### Related Issue
Fixes [#10845](https://github.com/openstreetmap/iD/issues/10845)

@tyrasd @tordans @k-yle

### DIFF
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -114,9 +114,10 @@ export function uiFieldAddress(field, context) {
     }
 
     function getNearPostcodes() {
-        return [... new Set([]
+        const postcodes = []
             .concat(getNearValues('postcode'))
-            .concat(getNear(d => d.tags.postal_code, 'postcode', 200, 'postal_code')))];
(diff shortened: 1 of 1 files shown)


k-yle (approved):
thanks!

PR #10851, opened 2025-03-08
Change follow tool to follow lines with multiple same parents

### LINKED ISSUE #8750: Follow operation should prefer the focused way

### LINKED ISSUE #11489: Allow continuing drawing using follow function if two ways/areas have the same geometry

jperelli:
When mapping a river tidal flat, if the river already has a coast and I want to add a second layer of tidal flat on top of the river feature, I cant use the `F` key to follow the coast of the river because there are 2 shared areas with the same line. I want to be able to follow for a third feature that shares the same linestring.

Just added a condition for that to work.

This is the feature working. The following node must be the pointed with the red arrow because both nodes share the same multiple parents.
(This cannot be made in current iD because the nodes have multiple parents, the follow tool only follows lines with one parent.)

This is where it stops, because there are multiple different alternatives to take for the following node to be chosen. (list of node parents are different)

### DIFF
--- a/modules/behavior/draw_way.js
+++ b/modules/behavior/draw_way.js
@@ -457,7 +457,11 @@ export function behaviorDrawWay(context, wayID, mode, startGraph) {
 
             const featureType = getFeatureType(lastNodesParents);
 
-            if (lastNodesParents.length !== 1 || secondLastNodesParents.length === 0) {
+            const sameAllParents =
+                lastNodesParents.length === secondLastNodesParents.length &&
+                lastNodesParents.every(w => secondLastNodesParents.some(n => n.id === w.id));
+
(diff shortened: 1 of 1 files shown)


k-yle:
see also: #9340 and #8978 which are two alternative approaches for solving this problem

jperelli:
Thanks @k-yle I see the interest in changing the algorithm.

In this solution the behaviour is the same, the change is minimal.
Maybe it can be considered to be merged because it solves the issue with minimal impact in the current behaviour of the tool while the other PRs can be focused in the discussion of a more general different approach, then when a discussion is agreed, the other PRs can overwrite this code.

tyrasd:
Closing in favor of #9340, sorry for the delay. Thanks for your effort anyway! :bow:

PR #10856, opened 2025-03-10
Panoramax: Adopt "copy id" feature from mapillary

### LINKED ISSUE #10611: Panoramax: Adopt "copy id" feature from mapillary

MohamedAli00949:
Hi there,

This is my code to add the `copy id` button at the Panoramax image

### Changes
- "Copy ID" button component.

### Issue
Closes https://github.com/openstreetmap/iD/issues/10611


tordans:
From reading the code I don't see how this PR could resolve the requirements described in the issue, specifically https://github.com/openstreetmap/iD/issues/10611#issuecomment-2553846040 and https://github.com/openstreetmap/iD/issues/10611#issuecomment-2575910888

MohamedAli00949:
As I understood the issue, I found the requirements are to add the  `copy id` button to the photo viewer, such as below, and research imagery services such as Panoramax photo viewer. I have replied with the research results.
Screenshot 2025-03-10 111753

So, what is wrong with my implementation? and How to improve it.
 
Thanks in advanced

MohamedAli00949:
@tordans, Do you mean that I haven't used the `panoramax=*` field and want to use it at the copy button instead of passing it to the function or another thing?

MohamedAli00949:
@tordans, after reading the related modules to this feature, I have found another solution to solving this problem using hash (search hash params class, or using context). Is this suitable?

tyrasd (changes requested):
it should work like the corresponding button on mapillary photos:

image

MohamedAli00949:
Ok, But I will use the copy icon instead of the plus

MohamedAli00949:
Hi @tyrasd, 
I have changed the component, but the street-side photo viewer appears as the following.
image

I suggest the following style:
image
If my suggestion is good, I will commit these changes.

tyrasd:
please use the same icon as used in the mapillary viewer. otherwise this would be very confusing.

tordans:
@tyrasd I think the main thing we need to agree on first is, what solution does resolve the linked issue. I outlined my proposal in [the links above](https://github.com/openstreetmap/iD/pull/10856#issuecomment-2709891771). The core idea of that is, to use what we have for Mapillary and abstract it to reuse of other image providers. This is not what is being proposed here as far as I understand the code but instead a copy button that would require to manually add the key and paste the value. As  a result we would have two features doing something very similar – one better then the other.

My recommendation would be to clarify what we are looking for so this PR might be adjusted … or otherwise closed as it is not what we should merge IMO.

tyrasd:
> The core idea of that is, to use what we have for Mapillary and abstract it to reuse of other image providers. 

I agree, that's basically what I meant with _it should work like the corresponding button on mapillary photos_ above. :shrug: 

A copy-to-clipboard button is definitely not going to cut it here.

gedankenstuecke:
I've rephrased the original issue I had opened to make it clearer that the "agreed" solution is copying the mapillary-functionality

MohamedAli00949:
I am so sorry, for the misunderstanding.
Latterly, I added the functionality to set the photo button which was implemented before at mapillary.
But, there's to other issues,

The first is the plus icon, which provides a bad experience and I suggest replacing it with the following icon (such as @tordans has been suggested before at mapillary implementation).
image

The second is the position of the zoom icons in the panormax photo as in the following image and I suggest moving them down a bit.
image

If one of my suggestions or both isn't right tell me to change them.

modules/ui/photoviewer.js
```diff
@@ -146,15 +150,18 @@ export function uiPhotoviewer(context) {
...
 
                 const graph = context.graph();
                 const entities = context.selectedIDs()
                     .map(id => graph.entity(id));
 
-                if (entities.map(entity => entity.tags.mapillary)
-                    .every(value => value === activeImage?.id)) {
+                const entitiesWithTag = entities.map(entity => entity.tags[tagName] ? entity.tags[tagName] : undefined).filter(e => e);
+                const isPhotoUsed = entitiesWithTag
+                    .some(value => value === activeImage?.id);
```
tordans: This part checks if the "looked at" image id is already present in the tags value, right?
We should make this something like `value.includes(activeImage?.id)` to account for the edge case that multiple image ids are used separated by `;`, eg. "`<id1>;<id2>`". (We talked about that practice while reviewing https://github.com/openstreetmap/id-tagging-schema/pull/1344)

modules/ui/photoviewer.js
```diff
@@ -75,30 +75,34 @@ export function uiPhotoviewer(context) {
 
 
         function setPhotoFromViewerButton() {
-            if (services.mapillary.isViewerOpen()) {
-                if (context.mode().id !== 'select' || !(layerStatus('mapillary') && getServiceId() === 'mapillary')) {
+            const service = getServiceId();
+            const isActiveService = ['mapillary', 'panoramax'].includes(service) && services[service].isViewerOpen();
```
tordans: Lets extract this array into something like a `const PHOTO_SERVICES_WITH_TAG_BUTTON = []` to give it a bit more visibility that this is the place where we configure the services. Please check the code conventions on such const (naming in placement in file); I don't know how that is done elsewhere in the codebase.

We are also making the assumption here that the service name and tag key will be the same. That is true for those two but might be different for others. We could account for this now… but it feels like premature optimization so I would do it in a refactoring later when the need comes up.

modules/ui/photoviewer.js
```diff
@@ -75,30 +75,34 @@ export function uiPhotoviewer(context) {
...
+            const mapillaryOrPanoramax = (layerStatus(service) && ['mapillary', 'panoramax'].includes(service));
```
tordans: - Lets use a more generic name for the const
- Lets reuse the same array defined once above

Can you double check the conditions, I think that isActiveService already accounts for the second condition in this check, right?

modules/ui/photoviewer.js
```diff
@@ -75,30 +75,34 @@ export function uiPhotoviewer(context) {
...
+
+            if (isActiveService) {
+                if (context.mode().id !== 'select' || !(mapillaryOrPanoramax)) {
```
tordans:
```suggestion
                if (context.mode().id !== 'select' || !mapillaryOrPanoramax) {
```

modules/ui/photoviewer.js
```diff
@@ -75,30 +75,34 @@ export function uiPhotoviewer(context) {
...
                         });
                     }
-                    buttonShowHide();
+                    buttonShowHide(service);
                 }
 
-                function setMapillaryPhotoId() {
-                    const service = services.mapillary;
-                    const image = service.getActiveImage();
+                function setMapillaryOrPanoramaxPhotoId() {
```
tordans: Like above, lets look for a name that is more generic. 

What would be a good name to describe the general feature? "addPhotoServiceTag", "applyTagOfPhotoService",…

modules/ui/photoviewer.js
```diff
@@ -146,15 +150,18 @@ export function uiPhotoviewer(context) {
...
+            function buttonShowHide(tagName) {
+                const activeImage = services[tagName].getActiveImage();
 
                 const graph = context.graph();
                 const entities = context.selectedIDs()
                     .map(id => graph.entity(id));
 
-                if (entities.map(entity => entity.tags.mapillary)
-                    .every(value => value === activeImage?.id)) {
+                const entitiesWithTag = entities.map(entity => entity.tags[tagName] ? entity.tags[tagName] : undefined).filter(e => e);
```
tordans: That should be the same but shorter, right?

```suggestion
                const entitiesWithTag = entities.map(entity => entity.tags[tagName]).filter(Boolean);
```

modules/ui/photoviewer.js
tordans: Interesting to see the change from `every` to `some`. That was a bug in the previous implementation, right?

tordans (changes requested):
Thanks for the update!

tordans:
Given the research you did in https://github.com/openstreetmap/iD/issues/10611#issuecomment-2709845395 I only see on other service that has a dedicated wiki page about a dedicated tag which is https://wiki.openstreetmap.org/wiki/Key:kartaview

I don't think Kartaview is especially active / used in OSM and the tag has low usage with ~600 https://taginfo.openstreetmap.org/keys/kartaview but we could look into adding this as well. The Kartaview UI might require extra work though, in which case I would rather skip it for later due to the low usage.

tordans:
Re https://github.com/openstreetmap/iD/pull/10856#issuecomment-2719703334

**Choosing the icon**

Personally, I agree that the plus is a no go given the +/- for Zooming and that from the list we https://github.com/openstreetmap/iD/pull/10046#issuecomment-2039107528 researched last time the one you chose is the best choice. Martin might see it differently but I suggest to go with that for now.

**Position of button**

- Ideally we find one that works for all selected services
- Ideally it is on the right because the tags are left of the preview (I think it feels unnatural to have the button on the right of the image viewer… but that might just be me)
- Can we move the +/- via custom CSS for panoramax?
- Otherwise we have to move the button. It looks like the best option is the bottom left picking a place that works for both services (given the different footer heights of both services); Ideally we don't need to change the position per service… — If so, that should might be something to abstract into the config CONST i mentioned in https://github.com/openstreetmap/iD/pull/10856/files/3e705dd7f72a52810a73bbeadca32b3fc158bf12#diff-f6d31cddb0ba27a935e3cb9e3aa4d4ec8ee2b68f92e4dcc26ca42c48724d7e9c

modules/ui/photoviewer.js
MohamedAli00949: nice, thanks

modules/ui/photoviewer.js
MohamedAli00949: > Interesting to see the change from `every` to `some`. That was a bug in the previous implementation, right?

This is right, I have found this bug and I have rewritten this line to refactor the function.

MohamedAli00949: > This part checks if the "looked at" image id is already present in the tags value, right? We should make this something like `value.includes(activeImage?.id)` to account for the edge case that multiple image ids are used separated by `;`, eg. "`<id1>;<id2>`". (We talked about that practice while reviewing [openstreetmap/id-tagging-schema#1344](https://github.com/openstreetmap/id-tagging-schema/pull/1344))

Okay, this is more effective, I will apply it.

Thanks, a lot.

modules/ui/photoviewer.js
MohamedAli00949: Yeah, I think  `addPhotoServiceTag` is more expressional to its implementation.

MohamedAli00949:
> Given the research you did in [#10611 (comment)](https://github.com/openstreetmap/iD/issues/10611#issuecomment-2709845395) I only see on other service that has a dedicated wiki page about a dedicated tag which is https://wiki.openstreetmap.org/wiki/Key:kartaview
> 
> I don't think Kartaview is especially active / used in OSM and the tag has low usage with ~600 https://taginfo.openstreetmap.org/keys/kartaview but we could look into adding this as well. The Kartaview UI might require extra work though, in which case I would rather skip it for later due to the low usage.

Yeah, this is right. 
But I can't see this field used in schema fields.

MohamedAli00949:
The final result:

image
image

MohamedAli00949:
@tordans @tyrasd, I have made the required changes. please give them another look.

tyrasd (approved):
The picker icon is a good choice! I found some minor issues in the PR (see a929a0510b1b86bc05f87b4a646a96c0631e0327, 0632a26fcd26577a1c4172a7ff0a2f6f17196835) and changed a few variable/function names to be more consistent with the naming. But otherwise it works well, and would be easy to extend to more photo service providers, should we need to add more in the future. Thanks!

MohamedAli00949:
> The picker icon is a good choice! I found some minor issues in the PR (see [a929a05](https://github.com/openstreetmap/iD/commit/a929a0510b1b86bc05f87b4a646a96c0631e0327), [0632a26](https://github.com/openstreetmap/iD/commit/0632a26fcd26577a1c4172a7ff0a2f6f17196835)) and changed a few variable/function names to be more consistent with the naming. But otherwise it works well, and would be easy to extend to more photo service providers, should we need to add more in the future. Thanks!

Thanks, Martin, I have seen your changes and I will consider him at the next PR.

PR #10864, opened 2025-03-11
fix: hotkey to select tags ( Alt + 1,2,3, ... )

### LINKED ISSUE #10850: Hotkey to select tags ( Alt + 1,2,3, ... )
### Description

When creating an area, point, or line, often the user has to schlepp over and select the same first or second selection on the type list. This is a small annoyance, but does add friction when trying to put in many similar features at once. A great QoL feature would allow a user to select a recently selected tag without moving the mouse all the way over to the menu. Currently we do not have anything bound to Alt + (number), which would be a good way to choose the first, second, third, etc. choice on the list of recently chosen types. 
This feature would conform to the clean and simple design of iD, while adding an excellent QoL feature when adding many similar features at once. 

### Screenshots

_No response_

chaitanyakadu:
This PR adds keybindings for "Alt + 1..9" to the Raw Tag Editor (`uiSectionRawTagEditor`), allowing users to quickly focus the value input field of the corresponding tag row (1st to 9th) in the tag list. 

Fixes
[#10850](https://github.com/openstreetmap/iD/issues/10850) 

@tyrasd @tordans @k-yle can you kindly review and let me know if additional modifications are needed! Should I implement test cases for this feature?


chaitanyakadu:
@1ec5 @tyrasd @tordans @k-yle can you kindly review and let me know if additional modifications are needed!

inbesea:
Any reason this has not been merged yet? Would really love this feature to be added in.

PR #10867, opened 2025-03-12
Update Asian language data

SteveLz:
* Added Languages:
`ko-Hani`: [wiki](https://wiki.openstreetmap.org/wiki/Key:name:ko-Hani) [taginfo](https://taginfo.openstreetmap.org/keys/name%3Ako-Hani)
`vi-Hani`: [wiki](https://wiki.openstreetmap.org/wiki/Key:name:vi-Hani) [taginfo](https://taginfo.openstreetmap.org/keys/name%3Avi-Hani)
`prk`: [iso](https://iso639-3.sil.org/code/prk) [Wikipedia](https://en.wikipedia.org/wiki/Wa_language) [taginfo](https://taginfo.openstreetmap.org/keys/name%3Aprk) (official language of Wa State)
`ksw`: [iso](https://iso639-3.sil.org/code/ksw) [Wikipedia](https://en.wikipedia.org/wiki/S%27gaw_Karen_language) (official language of Kayin State / KNU)
`blk`: [iso](https://iso639-3.sil.org/code/blk) [Wikipedia](https://en.wikipedia.org/wiki/Pa%27O_language) (recognized minority language)
* territory_languages changes (include in build_data.js):
Add CJKV script variants commonly used in OSM but not provided by CLDR (`zh-Hans`300k+, `zh-Hant`500k+, `zh-Latn-pinyin`100k+ in CN; `ja-Hira`300k+, `ja-Latn`100k+ in JP; `ko-Latn`100k+, `ko-Hani`60k+ in Korea; `nan-Hant`, `name:nan-Latn-tailo`, `name:nan-Latn-pehoeji` 40k+ in TW)
Add the missing second and third most commonly used languages in Taiwan: Minnan and Hakka
Add S'gaw Karen (2M speakers), Chinese (1.6M speakers; official language of SR1, SR2(Wa State), SR4), Pa'O (860k speakers), Wa (500k speakers) in Myanmar


tyrasd (dismissed):
Hi. Thank you for the contribution. Please note the following comments:

Could you please provide a short summary about the added languages with links to sources like wikipedia, [ISO code tables](https://iso639-3.sil.org/), the osm wiki and/or taginfo as applicable.

I noted that you included `kar`, which is AFAICS not an individual language, but a group of languages ([Collective](https://iso639-3.sil.org/code/kar)), which we want to avoid in favor of the respective concrete languages. There are a few exceptions, for example if there is already significant use in OSM (but that does not seem to be the case [here](https://taginfo.openstreetmap.org/keys/name%3Akar#overview)).

Also, please don't modify the files `data/languages.json` and `data/territory_languages.json` manually, as these are automatically generated by combining the information in CLDR and the overrides in the build scripts. Instead, please only change the `scripts/language_names.js` (for `languages.json`) and `scripts/build_data` (for `territory_languages.json`). I also noticed that some of your changes in the scripts are slightly different than the changes in the data file. For example, some of the language names are slightly different and `zh-Bopo` is missing in the `scripts`. Can you please make sure that they are in sync?

Thanks!

tyrasd:
@SteveLz thanks for the added links and the updates to `language_names.js`. What's now still missing are the adjustments for the territory languages in `build_data.js` and a bit of description/sources for those changes.

matkoniecz:
have we ever got "and a bit of description/sources for those changes." ?

Are these in one of many commit descriptions?

matkoniecz:
@SteveLz have we got this sources somehow?

SteveLz:
> have we ever got "and a bit of description/sources for those changes." ?
> 
> Are these in one of many commit descriptions?

The new content added to `territory_languages.json` is generated by `build_data.js`, so the two are consistent.

matkoniecz:
> What's now still missing are the adjustments for the territory languages in build_data.js and a bit of description/sources for those changes.

from what I see "a bit of description/sources for those changes." remains missing and whether territory_languages.json is generated by build_data.js is not changing that

or is it present somewhere?

scripts/build_data.js
```diff
@@ -192,6 +192,17 @@ function generateTerritoryLanguages() {
     'ulk', 'wbp', 'wrh', 'wth', 'wyi', 'xdk', 'xni', 'xph', 'xrd', 'zku'
   ]; // https://github.com/openstreetmap/iD/pull/10684
   territoryLanguages.nz.push('rrm'); // https://github.com/openstreetmap/iD/pull/10684
+  territoryLanguages.cn.splice(0, 1, 'zh-Hans', 'zh', 'zh-Hant', 'zh-Latn-pinyin');
+  territoryLanguages.cn.splice(27, 0, 'vi-Hani');
```
1ec5, line 196: We don’t need to splice it into a specific index. We should just push it onto the end of the array. The array gets converted into an unordered object anyways.

matkoniecz:
These changes are patching CLDR. Are there issues submitted to CLDR requesting to fix it also there?

I would say that any patch overriding CLDR should have link to issue there.

tyrasd:
Most of those changes seem to be improvements that should be upstreamed to CLDR. Please open ticket(s) on https://unicode-org.atlassian.net/jira/software/c/projects/CLDR/list (see also https://github.com/unicode-org/cldr/blob/main/CONTRIBUTING.md and https://cldr.unicode.org/requesting_changes#requesting-updates-to-locale-data-through-a-ticket), and let the experts decide whether e.g. the added territory languages do make sense or not.

The overrides here should be limited to an absolute minimum (e.g. to specifics of OSM, like if OSM was using nonstandard language or country codes for whatever reason).

scripts/build_data.js
tyrasd: actually, the order does make a small difference here, as the more popular languages in a given country should be listed before the less popular ones (in the build output it is an array, e.g. https://github.com/openstreetmap/iD/pull/10867/changes#diff-8c14116848fdea998154bdf0a9918d2862ca2fde33d2151bddb6efc3ec22403dR50).

But the way it is implemented here is much too fragile, and would need to be adjusted every time we update CLDR and they have additions/changes to the upstream list.

PR #10868, opened 2025-03-12
added pgdown-pgup shortcut keys for kartaview

### LINKED ISSUE #10370: Streetlevel layers: keyboard shortcuts for previous/next photo
Follow up for #10360: We could be to let Page up / Page down keys cycle through the individual photos of a sequence (except when a vertex node of a way is selected – there these keys are already used to cycle through the nodes of the way).

draunger:
# description
added the pagedown and pageup keys for kartaview street level photos.
# approach:
used utilkeybinding function for adding the shortcut.
# solves:
https://github.com/openstreetmap/iD/issues/10370

### DIFF
--- a/modules/services/kartaview.js
+++ b/modules/services/kartaview.js
@@ -1,12 +1,12 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
 import { json as d3_json } from 'd3-fetch';
 import { zoom as d3_zoom, zoomIdentity as d3_zoomIdentity } from 'd3-zoom';
-
+import * as d3 from 'd3';
 import RBush from 'rbush';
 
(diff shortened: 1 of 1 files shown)


k-yle (commented):
<kbd>PgUp</kbd>/<kbd>PgDown</kbd> are used in other parts of the codebase, you probaly need to check if `context.mode().id !== 'select'`. Otherwise the shortcut will do two different things. 

Also, this will logically conflict with #9731, whichever is merged last needs to be updated.

draunger:
@k-yle as the PgUp/PgDown keys are only be used in the moving of vertex node of a way , so for this i already checked for this condition as you can see in this image..
image

Issue #10878, opened 2025-03-14
Coordinate search does not work with / symbol


coteyka1209:
### Description

It works if you type something like: 46.112785 72.921033 or 46.112785,72.921033
But doesn't work if you type something like 46.112785/72.921033 or 46.112785 / 72.921033
fix pls :D

### Screenshots

![Image](https://github.com/user-attachments/assets/c5b7eae1-fc9b-427b-9e46-327c47046bd9)

danieldegroot2:
This has recently been added in https://github.com/openstreetmap/iD/commit/607f9534656126ae02976c84c9bc8771fbb7bb4c 
However, it seems to drop the decimal delimiter 😅 

Once this is fixed, you can try it out at https://ideditor.netlify.app/ . It will be available on openstreetmap.org in a future release.

Sidenote: Also recently `y,x` handling was added in #10725 . This is already live.

tyrasd:
> However, it seems to drop the decimal delimiter

yeah, that's a bug if you're on a locale that does not use `.` as a decimal separator: it will still try to parse the number as in the localized format and potentially interpret the `.`as a thousands separator (see also https://github.com/openstreetmap/iD/pull/10805#user-content-fn-1-b0224a503ca4fa775cec245969b5f5b5).

tyrasd:
this should now be fixed with #10879

danieldegroot2:
@tyrasd There are still unresolved comments on your commit
- https://github.com/openstreetmap/iD/commit/607f9534656126ae02976c84c9bc8771fbb7bb4c#r153726927
- https://github.com/openstreetmap/iD/commit/607f9534656126ae02976c84c9bc8771fbb7bb4c#r153725841
- https://github.com/openstreetmap/iD/commit/607f9534656126ae02976c84c9bc8771fbb7bb4c#r153727132

PR #10879, opened 2025-03-14
fix coordinate search with / symbol

draunger:
# description
added the dms matcher condition regarding the '/' symbol
# after the fix:
image
# solves:
https://github.com/openstreetmap/iD/issues/10878


tyrasd (dismissed):
thanks, somehow I overlooked that the condition for this case was missing in #10805

tyrasd (approved):
thanks, somehow I overlooked that the condition for this case was missing in #10805

Issue #6028, opened 2019-03-09
Bug: changing tag in "All tags" and selecting other element changes other element


SelfishSeahorse:
When changing a tag in "All tags" and then selecting another OSM element, iD copies some tag from the first element to the second one:

![Untitled](https://user-images.githubusercontent.com/14875493/54074330-7b6ddf00-4291-11e9-8e5a-ed9309e28b56.gif)

In this example, i changed the restaurant's `cuisine=*` tag and then selected the road, which resulted in a change of the road's `name=*` tag.

This problem occurs both with Safari 12.0.3 and Firefox 65.0.1.

bhousel:
well that sounds bad

SelfishSeahorse:
PS: This bug must have been introduced only recently (v2.14?). I've never noticed this error before.

brianhatchl:
I was able to reproduce this, and did not require an "actual" tag change.  I was able to select a POI and just double-click it's `name` key for example.  Then select a different POI and the name value of the first POI is applied.

It seems like the issue arises when a raw tag editor `change` event is dispatched on `blur` AND the blur event is triggered by a new feature being selected.  At the point the `changeTags` event is handled in entity editor the selected entity id has already changed.

BjornRasmussen:
I can also reproduce this easily.

bhousel:
I've been trying to reproduce this in Firefox 65 with no luck. 

@SelfishSeahorse / @brianhatchl  - can you confirm that it still shows up in http://preview.ideditor.com/master/  ?

I do believe there is probably an actual issue with the `change` and `blur` events, but I'm having a hard time to see it.

SelfishSeahorse:
> I've been trying to reproduce this in Firefox 65 with no luck.
> 
> @SelfishSeahorse / @brianhatchl - can you confirm that it still shows up in http://preview.ideditor.com/master/ ?

Yes, the problem still exists, both with Safari 12.1 and Firefox 66.0.2 (macOS 10.13.6).

When changing a tag value, then pressing the tabulator key and selecting another element, the tag after/below the one that has been changed is copied to the other element (as in the screencast above).

brianhatchl:
I was not able to reproduce the bug (as I've described it) on Firefox 66, but I'm still able to reproduce it on Chrome 73.

I go to this area, http://preview.ideditor.com/master/#background=Bing&disable_features=boundaries&map=19.99/18.51844/-72.41289

[...]

Issue #5878, opened 2019-02-09
While editing one tag, click delete button on another tag (raw tag editor)


sun-geo:
### Expected behavior
delete line function is working while editing other field 

### Actual behavior
delete line function is not working while editing other field

### Screenshots illustrating the current situation
example: - a 2/1 lane-road gets physical separation and you want change lane attributes
from:
lane:backward=2
lanes:forward=1
to:
lanes=2
So, the idea is to delete from line "lanes:backward=2" only the "backward" string and hit instantly the delete icon of line "lane:forward=1" for removing this entire line.
See here:
![del-del_bottom_line](https://user-images.githubusercontent.com/47290585/52520925-b3bcd600-2c6f-11e9-85d5-e164b61917d5.gif)

btw, I just noticed, it seems working when the supposed line for deleting is **not** the bottom one, see here:
![del-del_not_bottom_line](https://user-images.githubusercontent.com/47290585/52520929-be776b00-2c6f-11e9-8ef2-c5819952e4fa.gif)

bhousel:
You want to be able to hit the delete button on one tag while editing another tag?.... 

Ok, I don't know whether that is possible but I'll see if we can do that.. The problem is that clicking the button will blur the combo field, and I'm not sure whether the tag edit will reliably happen.

Also what browser is this?  I forget if I've asked you before, but for issues like this it matters.

bhousel:
Confirmed in Chrome - I was able to fix this by batching the events so that a combo field change/blur won't conflict with the click on the delete button.

sun-geo:
thank you

Issue #5847, opened 2019-02-03
tabbing through additional tags broken


GallusMax:
This seems to be unearthed in the latest (2019/01 !) version. More than annoying, when you trust the chosen value being committed by a TAB, when its not. Sometimes, often. Sometime it works ok.
PLEASE fix!

_Originally posted by @GallusMax in https://github.com/openstreetmap/iD/issues/1117#issuecomment-460056286_

bhousel:
Can you provide some more details about what you're seeing? Also which browser you are using?

I think it has something to do with the tag editor, but it's not clear from your bug report.

GallusMax:
Sure. 
Given You're in the advanced editing of item tags, altering tag / value pairs one by one.
The expected behaviour of the TAB would hop from a tag field to its value field, where after some typing several presets appear - or typing ends up with a new freetext value. Another TAB should ACCEPT the typed / selected value and hop to the next (possibly: new and empty) tag. 
(I didnt notice any other behaviour, yet my surprise)
In the current version a TAB will leave the value field empty from time to time. Even after manually entering free text(!). And its not even clear if it will then hop into a new line / a new and empty tag field.

Bowser version:
I often use Firefox (latest), but did never find any difference to Chrome e.g.

Hope to be helpful for you, bhousel, thanks!

GallusMax:
How to reproduce:
Choose an item o the map (a street?)
Expand the "All Tags" Part on the left
Put the Cursor in the "Tag" field on the left
Type "hig" - completion  "hway" is offered
Type TAB - tag is set to "highway", as expected, Cursor blinks in Value field on the right
Type "ser" - completion "vice" id offered, which is what you want
Type TAB - Value field is EMPTY again (which is not what _I_ expected), Cursor goes haywire, or on the next empty Tag field, whe I would look after it.

Expected: the offered expansion is committed by the TAB, and the next Tag field is opened.

GallusMax:
Duh - seems of course a Firefox Issue. Tested again with Firefox "Quantum" 59.0.1. Thats where it hurts. Will report whe I found an older version w/o this issue. (Sorry, did not user Chrome for editing for a long time)

Issue #4991, opened 2018-04-17
New tag value is lost when tab is used


Bonbadil:
Hello,

When adding tags directly on new or pre-existing nodes and using TAB to add a new one, the new tag is added to the list but not its value:
![lost value of new tag](https://user-images.githubusercontent.com/833234/38862445-41145d8c-4235-11e8-88a8-ba270246483b.gif)
Also, the focus do not return to the first column.

I can reproduce the bug easily, but not systematically, on the same node and in the same changeset. 

I'm using Firefox Developer Edition 60.0b12 (64 bits) on Windows 10. I've several extensions, if you can't reproduce the bug.
The JS console say nothing at this moment.

jguthrie100:
It works with Firefox 59.0.2 on Linux

bhousel:
Thank you for the detailed report!   I will try this out on a few different browsers and see if I can reproduce the issue.

govvin:
I confirm this on Firefox 59.02.

PR #10889, opened 2025-03-19
simplify raw tag editor ui and logic

### LINKED ISSUE #10871: Mistakenly triggering keyboard shortcuts
### Description

What about a setting to disable keyboard shortcuts while editing the tags table, or in general? While navigating the tag table with the Tab key, sometimes the table loses focus (either I press Tab too much, or I'm faster than the UI can handle), and the letters I type, intended to fill the cells, behave as keyboard shortcuts instead.

### Screenshots

_No response_

tyrasd:
* always show empty line at bottom of the tag list to allow adding more tags (replaces `+` button and tab logic)
* disable delete and info buttons on empty tag
* replace deferring of tag change events and re-rendering with a simpler logic (to fix issues like #10871): only set input field values if the rendering event was triggered by an external event, or by a change of the respective field in the raw tag editor itself
* skip delete and tag info buttons while navigating using tab: makes it more quick to get to where one typically needs to, deleting a key using the keyboard only is possible by emptying the tag's key, and advanced users are not typically using the `i` button anyway

todo:
* thoroughly test if this does not introduce any old or new bugs (see also #6028, #4991, #5878, #5847)
* add changelog (after release v2.33 is out to avoid merge conflict)

### DIFF
--- a/modules/ui/sections/raw_tag_editor.js
+++ b/modules/ui/sections/raw_tag_editor.js
@@ -1,5 +1,6 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
 import { select as d3_select } from 'd3-selection';
+import { isEmpty } from 'lodash-es';
 
 import { services } from '../../services';
 import { svgIcon } from '../../svg/icon';
@@ -7,10 +8,9 @@ import { uiCombobox } from '../combobox';
 import { uiSection } from '../section';
 import { uiTagReference } from '../tag_reference';
 import { prefs } from '../../core/preferences';
-import { localizer, t } from '../../core/localizer';
+import { t } from '../../core/localizer';
 import { utilArrayDifference, utilArrayIdentical } from '../../util/array';
 import { utilGetSetValue, utilNoAuto, utilRebind, utilTagDiff } from '../../util';
-import { uiTooltip } from '..';
 import { allowUpperCaseTagValues } from '../../osm/tags';
 import { fileFetcher } from '../../core';
 
@@ -42,7 +42,6 @@ export function uiSectionRawTagEditor(id, context) {
     var _readOnlyTags = [];
     // the keys in the order we want them to display
     var _orderedKeys = [];
-    var _showBlank = false;
     var _pendingChange = null;
     var _state;
     var _presets;
@@ -76,11 +75,8 @@ export function uiSectionRawTagEditor(id, context) {
             return { index: i, key: key, value: _tags[key] };
         });
 
-        // append blank row last, if necessary
-        if (!rowData.length || _showBlank) {
-            _showBlank = false;
-            rowData.push({ index: rowData.length, key: '', value: '' });
-        }
+        // append blank row last
+        rowData.push({ index: rowData.length, key: '', value: '' });
 
 
         // View Options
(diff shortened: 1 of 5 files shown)


modules/ui/sections/raw_tag_editor.js
```diff
@@ -260,7 +231,10 @@ export function uiSectionRawTagEditor(id, context) {
                 }
 
                 row.select('.inner-wrap')      // propagate bound data
-                    .call(reference.button);
+                    .call(reference.button)
+                    .select('.tag-reference-button')
+                    .attr('tabindex', -1)
```
1ec5, line 236: > skip delete and tag info buttons while navigating using tab: makes it more quick to get to where one typically needs to, deleting a key using the keyboard only is possible by emptying the tag's key, and advanced users are not typically using the i button anyway

I’ll definitely appreciate not having to tab past the 🗑️ and ℹ️ buttons. But keyboard navigation is not just for advanced users; it’s also for people who need to use the keyboard more often for ergonomic reasons. Advanced users probably do use the ℹ️ button too, though maybe not as often. On the other hand, this would make the raw tag editor more consistent with the tab order in the fields.

GitHub has some controls (such as the code editor) that co-opt the Tab key but allow Tab to do the normal thing with a modifier key or after pressing Escape. Maybe we can come up with a key combination that always toggles the ℹ️ when either a field or a raw tag is focused?

modules/ui/sections/raw_tag_editor.js
```diff
@@ -160,32 +156,6 @@ export function uiSectionRawTagEditor(id, context) {
             .merge(list);
 
 
-        // Container for the Add button
```
1ec5, line 163: > replaces + button and tab logic

The blank row is an elegant approach, but will removing this button make the remaining ➕ button in the relations section less intuitive? I think there’s probably already some initial confusion over whether clicking it creates a new relation or adds to an existing relation. The analogue to this change would be to always show an empty relation membership row.

modules/ui/sections/raw_tag_editor.js
tyrasd: > key combination that always toggles the ℹ️ 

That's an awesome suggestion! As this shortcut would be typically pressed while the cursor is in an input box, it must be a shortcut with a modifier key. Something like `Ctrl` + `?` could work…

modules/ui/sections/raw_tag_editor.js
tyrasd: Good idea, definitely worth experimenting with.

modules/ui/sections/raw_tag_editor.js
1ec5: > Something like `Ctrl` + `?` could work…

Unfortunately that particular shortcut could cause problems in some keyboard layouts and operating systems. The equivalent to <kbd>Ctrl</kbd><kbd>?</kbd> on macOS would be ⇧⌘?, which in the U.S. English layout would be ⇧⌘/ just like the system-wide shortcut for searching menu items. I’m guessing the most obvious key combinations for help content have already been taken, but it’s OK if the key combination is obscure, as long as it exists.

1ec5: On macOS, there’s a system-wide [Full Keyboard Access](https://support.apple.com/guide/mac-help/change-keyboard-settings-for-accessibility-mchlae61a6de/mac) setting for whether the tab order only includes text input controls or whether it includes every user input control. Regardless of this setting, holding down Option causes the Tab key to include every user input control, at least in Safari. Ideally we’d make sure this functionality still works for those who rely more heavily on the keyboard for navigation.

Issue #10906, opened 2025-03-24
Items in Hong Kong/Macau are wrongly getting Chinese address hints


Vectorial1024:
### URL

N/A

### How to reproduce the issue?

1. Find any feature inside Hong Kong/Macau that does not have its address field filled
2. Using the editor, add an Address field for editing
3. See the confusing Address fields

### Screenshot(s) or anything else?

For example:

![Image](https://github.com/user-attachments/assets/fe9e26f8-8e2c-4f85-9bef-aed28dfaee21)

------

For both Hong Kong and Macau, most of these fields are unused. It was the discussion of territory-level country codes (https://github.com/openstreetmap/iD/issues/10899#issuecomment-2745909687) that I then noticed: both Hong Kong and Macau are currently using the Chinese address hints.

Still, similar to the idea/situation in #10899 for phone/fax hints, both Hong Kong and Macau are still using its own separate address system with no known plans of migration to the Chinese system.

As such, the editor needs to be updated.

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.32.0

### Which browsers are you seeing this problem on?

Other (please specify)

Issue #10899, opened 2025-03-22
Items in Hong Kong/Macau are wrongly getting Chinese phone/fax hint text


Vectorial1024:
### URL

N/A

### How to reproduce the issue?

1. Find any feature inside Hong Kong/Macau that does not have its `phone`/`'fax`/(or related fields) filled
2. Using the editor, add a e.g. Telephone field for editing
3. See "+86" in the hint text

### Screenshot(s) or anything else?

For example, the phone field has hint text "+86 [...]".

![Image](https://github.com/user-attachments/assets/97176807-24cd-4fef-9c73-512a016b4314)

------

Hong Kong (currently officially Hong Kong Special Administrative Region, People's Republic of China, or "HKSAR, PRC") and Macau (currently officially Macau Special Administrative Region, People's Republic of China, or "MSAR, PRC") used to be separate political entities as colonies of the UK and Portugal respectively.

Both Hong Kong and Macau got their own country code when the relevant telecomms standards were first drafted:
- Hong Kong: +852
- Macau: +853

This is distinct from the "mainland Chinese" code +86 currently being used by People's Republic of China ("PRC", also commonly known simply as "China").

As part of the deals about decolonization and Chinese takeover (Hong Kong 1997; Macau 1999), it was agreed these country codes would be kept in use, and as of now, there are no plans to migrate the phone numbers to use the +86 prefix. The +852 and +853 phone numbers are currently still in use, as have been pre-takeover. (Note: we should not consider the technically possible case of "what happens in 2047/2049".)

Yet, the iD editor is completely unaware of this situation, and is showing the Chinese code +86 instead.

Those who know about this situation will correctly fill in +852/+853, but I am concerned those unaware would just blindly trust the editor and fill in +86, and then we will have invalid phone numbers documented in OSM.

Please review the relevant logic regarding the hint texts and cater for the on-the-ground situation of +852/+853 in both Hong Kong and Macau.

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.32.0

### Which browsers are you seeing this problem on?

Other (please specify)

Kovoschiz:
I doubt if any attributes are handling HK and Macau separately from PRC. Most prominently `addr:*=` suggestions are mostly irrelevant. I can try to work on them if they are a significant problem. 
Peculiar `addr:postcode=` mentioned at first https://github.com/openstreetmap/iD/pull/2246#issuecomment-45256055

[...]

PR #10904, opened 2025-03-24
Enable territory-level phone hints

Vectorial1024:
As guided by https://github.com/openstreetmap/iD/issues/10899#issuecomment-2745909687 , this PR enables the editor to look at the territory-level country code, and therefore fixes #10899. (Note: address formatting will be in another issue/PR)

Hopefully there are no unexpected side effects.

With this PR, a random example feature in Hong Kong:

Screenshot 2025-03-24 104247

And also, a random example feature in Macau:

Screenshot 2025-03-24 104337

PR #10907, opened 2025-03-24
Describe Hong Kong/Macau address format

Vectorial1024:
While reading https://github.com/openstreetmap/iD/issues/10899#issuecomment-2745909687 and therefore preparing for #10906, it is discovered that the editor does not currently know about Hong Kong's address format. Also, the Macau address format is inaccurate.

As such, this PR simply adds in Hong Kong's address format and updates Macau's address format.

------

Since Hong Kong and Macau are essentially city-states, any address field at or above "city" is not needed.

Also, both [Hong Kong](https://en.wikipedia.org/wiki/Postal_codes_in_Hong_Kong) and Macau do not use post codes. Technically, China has assigned post codes for addresses in Hong Kong and Macau, but those are only used for CN->HK and CN->MO mails, and are clearly out of scope and need not be included.

The density of urban development in Hong Kong and Macau necessitates including the floor and the unit number in the address field.

Care has been taken so that Hong Kong's format follows the UK while Macau's format follows the Portuguese.


Vectorial1024:
The local Hong Kong / Macau mapping community needs to be consulted about this change. As such, it is now a draft.

Vectorial1024:
@k-yle This does NOT fix #10906. This is only a preparation to an upcoming fix.

k-yle (commented):
sorry, I misread the issue description. issue now unlinked. If you want to implement it in this PR, it should be fairly straightforward. This line:
https://github.com/openstreetmap/iD/blob/4a1f27b1f6d6a322bd4563ac41aa7f8beb8edd07/modules/ui/fields/address.js#L276
would need to be changed to something more like this:
```js
countryCode = [
  countryCoder.iso1A2Code(center, { level: 'territory' }),
  countryCoder.iso1A2Code(center),
];
```

and then everywhere in this file that references `_countryCode` would need to support an array instead of a string.

This ensures that territories with no definitions (like the [Cook Islands](https://osm.org/way/516477220)) can fallback to the country (`nz` in this case)

tyrasd:
> The local Hong Kong / Macau mapping community needs to be consulted

@Vectorial1024 have you already started this discussion?

Vectorial1024:
@tyrasd I just started it in these few hours, and will need time to get results. The basic idea is already in this PR, but the discussion is aimed to see whether additional fields are also needed.

For a Hong Kong example, it is customary to write out the "region" of an address (e.g. "Kowloon"), but due to #10908, it doesn't seem possible to use the editor to write in such "region".

And regarding these "regional" addresses I am not sure about how to document them anyway since it isn't usually done.

tyrasd:
Ok, sure, no hurry. But just to clarify: are you going to start this discussion? :blush:

Vectorial1024:
@tyrasd Sorry, you mean discussing the contents of this PR...?

If you meant what the actual address format should be, then I have started it (asking around in the OSM-linked Discord). Or did you mean I should ask elsewhere eg the OSM wiki page?

I'm new to this anyway, but it seems to me the prominent local mappers can be found on Discord.

tyrasd:
Oh ok, got it. I just wanted to make sure that this is actually in motion. And yes, I mean the "content" of the PR, i.e. the actual address tagging to use in Hong Kong / Macau.

I'd say the discord a good start and should be fine for now. :+1: Would you mind sharing the channel where this is discussed, so that others can follow along or chime in?

Eventually, it would be great if the results of the discussion are also documented on the wiki (e.g. [here](https://wiki.openstreetmap.org/wiki/Addresses#Country_specific_rules_and_sources)), but that's for a later stage.

Vectorial1024:
The Discord channel is here https://discord.gg/qWsNYjrt ; also pinging @Kovoschiz 

And it seems there is no prior consensus whatsoever; we've been basically avoiding this topic since it's not exactly fully necessary. It's gonna take a while.

Kovoschiz:
Yes, I have read here. Thought we should focus the tagging discussion in Discord, reserving Github for the technical details of how to implement this.

novolife:
https://www.hongkongpost.hk/en/about_us/tips/correct_address/index.html

Kovoschiz:
We have looked at it before. The HK Post example doesn't help, as it doesn't even cover the most common case of housing estates, and phases or towers in a complex. The "village, town, or district" line   fails, as villages are often written together with the larger (and intermediate) location they are associated with (ps most District Council "districts" aren't useful for addressing either). The HK Post suggestion system is very broken, poorly formatted, disorganized, messy, and chaotic.  

Yes, we have an idea of how addresses should be written here. What's being discussed is how they should fit in `addr:*=` in OSM.

matkoniecz:
> @[Vectorial1024](https://github.com/Vectorial1024) Vectorial1024 marked this pull request as draft [11 months ago](https://github.com/openstreetmap/iD/pull/10907#event-16951408614)

Are you planning to progress with this PR to point where it can be reviewed?

Vectorial1024:
Hi there! Thanks for the reminder. Too many things have happened / are happening behind the scenes, so eventually I really forgot about this.

Unfortunately, as noted by @Kovoschiz , the official guides are chaotic. While we digested the guide, we also noticed that the irl situation is also chaotic. We still don't have a clear idea of what this addressing format should look like.

To avoid blocking this further, I am thinking of submitting a "basic version" for now, and then we can submit another PR improving this after we have more information / knowledge of how these addressing works.

Vectorial1024:
I have a strong feeling the iD editor simply doesn't know what is "Hong Kong". Even if this PR is accepted, we may simply see no change of behavior.

We may also need to review another loosely-related PR: https://github.com/openstreetmap/iD/pull/10904

Kovoschiz:
The difficult conceptual and technical levels are determining when and how to use either or both `addr:suburb=` vs `addr:town` 
So this is fine by stopping at largest being `addr:street=` / `addr:place=` (to form the minimal necessary components) 
However, I would ask to add `addr:door=` for store "units", as `addr:unit=` has become numerously used for "parts" of complexes/"buildings" (in the meaning of the originally proposed Continental format) https://wiki.openstreetmap.org/wiki/Proposal:Addr_keys_(2011-04)#Examples

Vectorial1024:
Good catch; I forgot some HK urban (industrial) buildings do use `addr:unit`, which means we do need a separate `addr:door`.

This should be a minor change; almost there.

Edit: we can also take this chance to resolve the merge conflict.

Vectorial1024:
The door field has been added to the Hong Kong address hint. However, while tie iD editor can correctly place the value into the `addr:door` tag, it should be noted that the iD editor currently does not have translations for this door field. This may be a concern.

matkoniecz:
> I have a strong feeling the iD editor simply doesn't know what is "Hong Kong". Even if this PR is accepted, we may simply see no change of behavior.

Have you tested is it actually working?

(Honk Kong detection seems to work for #10904 at least but it may use a different mechanism)

> The Discord channel is here https://discord.gg/qWsNYjrt ;

how can you get onto that Discord? This link does not work

Kovoschiz:
@matkoniecz I don't know where that link came from. The documented one still works. https://wiki.openstreetmap.org/wiki/Discord https://discord.com/invite/mhvmg7S

Vectorial1024:
I probably made a new time-limited invite link to Discord (which is different from the one on the Wiki), but this is going off-topic.

matkoniecz:
Maybe I am testing something badly but it looks like this PR has no effect?

Have you tested it?

note

> I have a strong feeling the iD editor simply doesn't know what is "Hong Kong". Even if this PR is accepted, we may simply see no change of behavior.

screen-2026-02-27-12-45-10
screen-2026-02-27-12-45-18

Vectorial1024:
My bad; the related code issue should have been https://github.com/openstreetmap/iD/pull/10914 (one-four), not 10904 (zero-four) . By pure coincidence the phone detection issue and the address detection issue have highly similar issue numbers and I keep mixing them up when dealing with both issues.

But my overall understanding is that each module detects countries separately, resulting in two independent but thematically related PRs.

matkoniecz:
so both this PR and #10914 by itself will not result in any user visible changes

I will try to test both at the same time

matkoniecz:
screen-2026-03-06-12-14-00

when I test both together, merged (with also latest master branch merged in) I get some bogus error - I have nothing in pl.


how it looks for you?

Vectorial1024:
@matkoniecz known issue see https://github.com/openstreetmap/iD/pull/10907#issuecomment-3952185696

> [...] it should be noted that the iD editor currently does not have translations for this door field. This may be a concern.

Kovoschiz:
Forgot to ask: Can we change the field label text "Housename" etc together here, or it belongs to /openstreetmap/id-tagging-schema/ ? Transifex is for the language not locations, and the countries are for national dialects?

matkoniecz:
To what Housename would be changed?

Vectorial1024:
Whatever happens, I have no appetite to deal with the translations surrounding `addr:housename` in this PR. Just open a new issue/PR for that.

matkoniecz:
> @matkoniecz known issue see https://github.com/openstreetmap/iD/pull/10907#issuecomment-3952185696

> >    [...] it should be noted that the iD editor currently does not have translations for this door field. This may be a concern.

I am not 100% sure what is going on...

But It seems you define a new string to be translated so you should include base language version (US English) in a base string file. Translations will be collected as usual.

Vectorial1024:
With some code searches, it seems to me this repo doesn't contain any translation for `addr:*` at all, so whether I like it or not, we literally can't deal with translation issues discovered thus far. We have to deal with them somewhere else.

Vectorial1024:
Update: I found the translations; they are in `openstreetmap/id-tagging-schema`.

@Kovoschiz

Kovoschiz:
Oh, I have seen the `*!jp:` before. Forgot about it, and didn't look beyond the search result excerpt.

matkoniecz:
Should we wait until https://github.com/openstreetmap/id-tagging-schema/pull/2023 leaves draft state and addresses feedback?

Kovoschiz:
Err it's confusing why they reacted here, but not there. I will do it now.

Kovoschiz:
Wait, I don't understand. Isn't #10914 needed for me to preview it? But that wants here to be done first?

matkoniecz:
basically, both need to be approved at the same time

Vectorial1024:
Correct. Back then I validated my PR idea/code by modifying the CN address hint format in my local copy of the iD editor when the logic isn't there yet.

Vectorial1024:
Upon review, by attempting to "make it fancy", I have essentially made this PR (and the territory-level code PR) un-mergeable, which is blocking progress. This is entirely my mistake and misjudgement.

In the coming days I will go through the details again and make something mergeable.

Vectorial1024:
Superceded by https://github.com/openstreetmap/iD/pull/12704

Issue #9499, opened 2023-02-08
City/country name is not localised


Dimitar5555:
### How to reproduce the issue?

1. Change your locale Bulgarian (bg)
2. Go to 42.69093, 23.33091 and make some edit
3. Upload it
4. Notice the text on the left panel ("Благодарим ви за подобряването на картата около Sofia, Bulgaria.")

### Screenshot(s) or anything else?

The text should be using the features' names in the selected locale. The text in this case should be "Благодарим ви за подобряването на картата около София, България"
![image](https://user-images.githubusercontent.com/19364673/217494499-eeeb3a5c-3ec7-419d-aa6a-c8eb236fc585.png)

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.24.1

k-yle:
The english part (`Sofia, Bulgaria`) comes from [Nominatim](https://nominatim.openstreetmap.org), which uses your browser's locale, whereas the rest of iD uses [the settings for your OSM account](https://www.openstreetmap.org/preferences/edit). I have a similar problem, text from Nominatim is in te reo māori or spanish, but the rest of iD is in english. This is fixed in #9501

PR #10911, opened 2025-03-24
consider `name:*` tags with the full locale code when labelling features

### LINKED ISSUE #10219: Feature labels are based on generic name:zh tag when using Traditional or Simplified Chinese locale
### Description

Currently, neither `name:zh-Hant` nor `name:zh-Hans` are supported to be displayed on the map, but only the ambiguous `name:zh`, which is either traditional or simplified, is displayed.

### Screenshots

_No response_

### LINKED ISSUE #10882: Add support for PT-BR items names inside iD
### Description

Hi!

I'm starting to improve names in PT-BR (after discussing with community), because in some not so rare cases, the names differ in the PT and PT-BR variants. For example, regarding country names, we have:

`name:pt=Irão`
`name:pt-BR=Irã`

There are dozens of these cases, and I'm starting mapping these little by little. However, when I do add/fix the name, in iD the label, in the map, is still in PT, even though my whole setup is in PT-BR. Example:

Image

You can check [editing directly here](https://www.openstreetmap.org/edit#map=19/32.647415/54.564736).

I'm not sure if it's a quick fix, but I saw something related to data/languages and data/territory_languages.

If I add

"pt-BR": {"base": "pt", "nativeName": "Português Brasileiro"},

and update

"br": ["pt-BR", "pt", "en", "de", "it", "vec", "ja", "es", "kgp", "ko", "yrl", "gub", "xav"],

(not in those files, but in scripts/language_names.js and scripts/build_data)

Will it work?

Thanks!

### Screenshots

_No response_

### LINKED ISSUE #11315: Support zh-Hant and zh-Hans
### Description

Osm now has Chinese(Simplified) Chinese(Traditional) Chinese(Hong Kong) three Chinese language options. However, the iD editor always shows name:zh, ignoring name:zh-Hans and name:zh-Hant. There is a fact that outside of China, name:zh sometimes is traditional script, sometimes is simplified script(Just depends on the first mapper's habits) . So I think we should have a fallback that checks if there are name:zh-Hant or name:zh-Hans.
I have made a PR in https://github.com/openstreetmap/iD/pull/11082#discussion_r2139658080 . But the pr was not merged. Now the osm language system have updated, so I think maybe I should create a new PR to solve the problem.

### Screenshots

Image
Image
This is an example that I chose Chinese(Simplified), but iD editor displayed traditional Chinese script in name:zh.

### LINKED ISSUE #11731: [localizer] Consider the variant subtag
Follow up to https://github.com/openstreetmap/iD/pull/10911#discussion_r2651256463

If the user's locale includes a [variant subtag](https://taginfo.geofabrik.de/asia:indonesia:kalimantan/sources/languages/subtags?filter=variant), (such as [`-tarask`](http://r12a.github.io/app-subtags/index?lookup=-tarask) or [`-oxendict`](http://r12a.github.io/app-subtags/index?lookup=-oxendict)), iD should consider the variant when generating every possible combination of the [maxmimised](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/maximize) langauge tag: 

https://github.com/openstreetmap/iD/blob/b36341dfa347578fe0add2f6c61b6c35124f2ab5/modules/util/localizer.ts#L6-L9

For example: Given [`zh-Latn-pinyin`](http://r12a.github.io/app-subtags/index?lookup=zh-Latn-pinyin), [`utilExpandLocaleCode`](https://github.com/openstreetmap/iD/blob/b36341dfa347578fe0add2f6c61b6c35124f2ab5/modules/util/localizer.ts#L23) should return 8 values:
* `zh-CN`, `zh-Latn-CN`, `zh-Latn`, `zh` (unchanged)
* `zh-CN-pinyin`, `zh-Latn-CN-pinyin`, `zh-Latn-pinyin`, `zh-pinyin` (new)

---

This is blocked by [browser support for](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/variants#browser_compatibility) [`Intl.Locale#variants`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/variants). 

It would also have no impact, since none of iD's supported languages include a variant subtag. This might change if either:
- #11730 is implemented; or
- iD is localised into a new language which includes a variant subtag.

k-yle:
Closes #10219, Closes #10882, Closes #11315, Closes #11731, Closes #11082 (superseeds that PR)

Thanks to [`Intl.Locale#maximize`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/maximize), we can expand the user's locale code to add [likely subtags](https://www.unicode.org/reports/tr35/#Likely_Subtags), then try every combination of `language` [+ `script` [ +`region`]] until we find a matching `name:*` tag.

&nbsp;

For example, we can expand [`zh-CN`](https://r12a.github.io/app-subtags/index?lookup=zh-CN) to [`zh-Hans-CN`](https://r12a.github.io/app-subtags/index?lookup=zh-Hans-CN), and expand [`zh-TW`](https://r12a.github.io/app-subtags/index?lookup=zh-TW) to [`zh-Hant-TW`](https://r12a.github.io/app-subtags/index?lookup=zh-Hant-TW), without hardcoding these cases.

Then, we use the most appropriate `name:*` tag, first checking `name:zh-Hans-CN`, then `name:zh-CN`, then `name:zh-Hans`, then `name:zh`.

This also fixes the [`pt-BR`](https://r12a.github.io/app-subtags/index?lookup=pt-BR) case from #10882


tyrasd (approved):
It improves the current implementation significantly, so :+1: from my side. Feel free to merge as is, or to address the comments below.

1. while this works well for the case reported in #10882, it does not seem to fully address the original issue from #10219:
  * the [common](https://taginfo.openstreetmap.org/search?q=name%3Azh) keys for Chinese names include ones with script codes (e.g. `name:zh-Hans` or `name:zh-Hant`), but if, for example, a user has `navigator.languages = ['zh-Hans-CN']`, the labels would not be shown for tags like `name:zh-Hans`.
2. It feels slightly odd that this is tied to iD's UI translations.
  * Let's say a user has `navigator.language == 'lld'`. They will get iD in `en`, as that's the best available fallback for UI translations. Ideally, they should see labels in `lld`
  * Similarly, if `navigator.languages = ['de-AT', 'de']`. iD would be in `de`, but why would the user not get labels for `de-AT`
3. Also, what's your opinion to extend this even further:
  * when a user has multiple languages in their `navigator.languages`, we could check the full fallback list: e.g. when a user has `navigator.languages = ['lld', 'it', 'de', 'en']`, features could be labeled by the respective languages in that order
  * we could consider "looking ahead" for any region codes in the `name:` tag. e.g. if a user has `navigator.languages = ['pt']`, it might be ok to also display a label for a `name:pt-BR` tag, if there is no `name:pt`
  * same as the previous point, but for scripts: if `navigator.languages = ['zh-Hans-CN']`, (then `localizer.localeCodes() == ['zh-CN']`,) it would be ok to display labels in `name:zh-Hans`, maybe even in other scripts like `name:zh-Hant` or `name:zh-Latn-pinyin`

tyrasd:
PS: do you have a clue what could cause [this CI error](https://github.com/openstreetmap/iD/actions/runs/14036599855/job/39295988320?pr=10911#step:6:457)?

     Error: AssertionError: expected [ 'en' ] to strictly equal [ 'en-US', 'en' ]

k-yle:
> 1. while this works well for the case reported in [#10882](https://github.com/openstreetmap/iD/issues/10882), it does not seem to fully address the original issue from [#10219](https://github.com/openstreetmap/iD/issues/10219)

I was planning to send a third PR which does the special-casing for `zh`, but I thought I'd seperate it from this PR which is much more straightforward. So lets discuss this in #10911

> 2. It feels slightly odd that this is tied to iD's UI translations. [...]
> 3. Also, what's your opinion to extend this even further: [...]

Yeah, I also think this would be useful, but maybe in a separate PR? While this would be nice for me personally, it might be unexpected for others (#9499 in one such example)

tyrasd:
Let's prioritize #11082 for now, as that one is addressing the more often occurring situation in practice (i.e. a user has multiple locales as their preference and also solves the `zh-*` issue) and names like `pt-BR` are quite rare in OSM at the moment.

k-yle:
Since #11082 is stale, I've redone this PR to use [`Intl.Locale#maximize`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/maximize), which is an awesome new browser API.

This PR can now match `zh-CN` to `zh-Hans` without hardcoding any exceptions, so it solves both issues and superseeds #11082. See updated PR description

modules/util/localizer.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (region) codes.add(`${language}-${region}`);
+    if (script) codes.add(`${language}-${script}`);
+    codes.add(language);
+
+    return codes;
+}
+
+
+/**
+ * expands a locale code like `jp` to `ja-Jpan-JP`,
```
1ec5:
```suggestion
 * expands a locale code like `ja` to `ja-Jpan-JP`,
```

modules/util/localizer.ts
```diff
@@ -0,0 +1,45 @@
+import { utilArrayUniq } from './array';
+
+function collectLocale({ script, region, language }: Intl.Locale) {
```
1ec5: Some common tags require a variant code, such as the `tarask` in [`name:be-tarask=*`](https://taginfo.openstreetmap.org/keys/name:be-tarask). `maximize()` returns `be-Cyrl-BY-tarask`. There’s a `variant` property but only Firefox supports it.

This is an edge case that we can ticket out as tail work. iD doesn’t currently have any localizations that require a variant. The closest is our “ar-AA” locale, which should probably be just “ar”. (I think Transifex made up a country code for the Unitag variant.)

1ec5 (commented):
Nice find!

modules/util/localizer.ts
k-yle: I've opened #11731 to track this

1ec5 (approved):
This looks good to go from a code perspective. Before merging, please spot-check iD in any tricky localizations like Arabic, Chinese, or Serbian in places that might have secondary localized names in those languages.

k-yle:
i made one change ([5fe8f6e..601f5c8](https://github.com/openstreetmap/iD/compare/5fe8f6e..601f5c8)), so that base language is the last fallback, to fix case #3 below. 

seems like this works well:
1. 🆕 [`locale=zh-CN` uses `name:zh-Hans`](https://pr-10911--ideditor.netlify.app/#locale=zh-CN&id=n3343705059)
2. [`locale=zh-TW` prefers `name:zh` over `name:nan`](https://pr-10911--ideditor.netlify.app/#locale=zh-TW&id=r5478819)
3. 🆕 [`locale=zh-HK` prefers `name:zh-Hant` over `name:zh`](https://pr-10911--ideditor.netlify.app/#locale=zh-HK&id=n244085114).
4. 🆕 [`locale=pt-BR` prefers `name:pt-BR` over `name:pt`](https://pr-10911--ideditor.netlify.app/#locale=pt-BR&id=n424312026)
5. [`locale=ar-AA` with `name:ar-Latn`](https://pr-10911--ideditor.netlify.app/#locale=ar-AA&id=w579336200)
6. [`locale=sr` prefers `name:sr` over `name:sr-Latn`](https://pr-10911--ideditor.netlify.app/#locale=sr&id=n6188902930)
7. 🆕 [`locale=sr` prefers `name:sr-Cyrl` over `name:sr-Latn`](https://pr-10911--ideditor.netlify.app/#locale=sr&id=r9440629)

_🆕 means the behaviour was changed by this PR_

1ec5 (approved):

k-yle:
if there are no objections, i think we could merge this one?

1ec5:
I implemented osm-americana/diplomat#30 based on the approach taken here. It works well. Diplomat needs to handle #11731 and even more complex language tags, so I modified the implementation to calculate a powerset instead of hard-coding specific combinations of language tag components. But I think what we have here is more than adequate for the issues iD users are noticing.

daishu0000:
> I implemented [osm-americana/diplomat#30](https://github.com/osm-americana/diplomat/pull/30) based on the approach taken here. It works well. Diplomat needs to handle #11731 and even more complex language tags, so I modified the implementation to calculate a powerset instead of hard-coding specific combinations of language tag components. But I think what we have here is more than adequate for the issues iD users are noticing.

I agree, we should not hard-coding the language combinations. For example, there are "name:zh name:zh-Hans name:zh-Hant name:zh-Hans-CN name:zh-Hans-SG name:zh-Hant-TW name:zh-Hant-HK name:zh-Hant-MO" and endless name:zh tags. If we just hard-coding, we can't handle all situations.

daishu0000:
Actually, I think we should allow users to customize language fallback solution, in [My Advanced Preferences](https://www.openstreetmap.org/preferences/advanced), users can define more than one preferred languages, can we allow users to directly write osm language tag(e.g. name:zh name:zh-Latn-pinyin) there? This way users can customized the language variants order.

k-yle:
> Actually, I think we should allow users to customize language fallback solution, [...] 

perhaps that should be discussed in #11730 ?

matkoniecz:
> Closes https://github.com/openstreetmap/iD/issues/10219 (part 2)

so after this PR is merged #10219 should not be closed, right?

k-yle:
sorry I forgot to update that part of the PR description, fixed now. #10219 will be fully fixed by this PR

test/spec/util/localizer.ts
```diff
@@ -0,0 +1,21 @@
+
+describe(iD.utilExpandLocaleCode, () => {
+    // these tests depend on the JS runtime's bundled ICU data,
+    // so they should only cover common cases that don't differ
+    // between runtimes.
+    it.each`
+        input               | output
+        ${'en'}             | ${['en', 'en-Latn-US', 'en-US', 'en-Latn']}
```
matkoniecz: why `'en-Latn-US'` is preferred over `'en-US'` but `en` over `'en-Latn'`?

(not saying it does not make sense, just I do not understand yet what is going on)

test/spec/util/util.js
```diff
@@ -352,6 +352,22 @@ describe('iD.util', function() {
...
+        describe('localised names', () => {
+            it('sanity check', () => {
+                // these tests depend on the locale of the test environment.
+                // check what the locale is first, to avoid confusion.
+                expect(iD.localizer.expandedLocaleCodes()).toStrictEqual(['en-US', 'en-Latn-US', 'en-Latn', 'en']);
+            });
+            it('considers the original locale code first', () => {
+                expect(iD.utilDisplayName({ tags: { name: '5', 'name:en': '4', 'name:en-Latn': '3', 'name:en-Latn-US': '2', 'name:en-US': '1' } }, undefined, true)).toBe('1');
+            });
+            it('fallbacks to the maxmimised locale', () => {
```
matkoniecz:
```suggestion
            it('fallbacks to the maximized locale', () => {
```

test/spec/util/util.js
```diff
@@ -352,6 +352,22 @@ describe('iD.util', function() {
         it('uses only the housenumber for map labels', () => {
             expect(iD.utilDisplayName({ tags: { 'addr:housenumber': '31', 'addr:street': 'Princes Street' } }, undefined, true)).to.eql('31');
         });
+
+        describe('localised names', () => {
+            it('sanity check', () => {
+                // these tests depend on the locale of the test environment.
+                // check what the locale is first, to avoid confusion.
```
matkoniecz: is there some standard way to set test locale? or official docs?

if yes - maybe linking them here would make sense?

(this test is failing for me and initial googling failed to find answer how to fix it)

test/spec/util/localizer.ts
matkoniecz: and note it may affect anyone running tests after this is merged, including test attempts of something unrelated

1ec5: The fallback order I’m using in Diplomat and OSM Americana is `en-Latn-US`, `en-Latn`, `en-US`, `en`. (That’s what I meant by “powerset” in https://github.com/openstreetmap/iD/pull/10911#issuecomment-3941498387.) The middle ones are more or less tied, but I’d err on the side of ensuring readability for someone who doesn’t necessarily know a writing system other than their own.

k-yle: > why `'en-Latn-US'` is preferred over `'en-US'` but `en` over `'en-Latn'`?

My original approach was to make the exact match come first (the user requested `en` so they get `en`).

But I implemented this before seeing Minh's [more advanced version](https://github.com/osm-americana/diplomat/pull/30) (which i think is brilliant btw), so we could change it to match diplomat's ordering if that's preferred (?)

matkoniecz: Personally I think that @1ec5 ordering seems to make more sense, but I lack expertise or experience or anything similar in this area.

1ec5: Yeah, ordering from most specific to least specific is important because of languages like Chinese. Browsers generally indicate the maximized locale, but mappers rarely tag that much detail because it doesn’t make a difference to the place name.

test/spec/util/util.js
k-yle: > is there some standard way to set test locale? or official docs?

sadly no, see https://github.com/vitest-dev/vitest/issues/9539#issuecomment-3808207511

the only way is to use `LC_ALL=en-US npm test`, but this command is platform-specific (different syntax on windows)

matkoniecz: maybe if it is Linux specific, maybe mention it? And mention magic incantation for Windows if we find it?

k-yle: fixed now, so the unit tests work regardless of your computer's language

test/spec/util/localizer.ts
k-yle: sorry for the delay, i had to think about this a bit. I've copied the powerset approach from diplomat, with [2 changes](https://github.com/openstreetmap/iD/blob/d200f4826659c88c00ee7522344f43456217b89a/modules/util/localizer.ts#L34-L45) that i find useful:

1. if `x !== new Intl.Locale(x).baseName`, then include and prioritise `x`. This is required to support [`cmn`](https://r12a.github.io/app-subtags/index?lookup=cmn), which the browser transforms into the macrolanguage [`zh`](https://r12a.github.io/app-subtags/index?lookup=zh)[^1]
2. if the final array of locale codes does not include the original input, then include and prioritise the original. This is required to support [extensions](https://www.unicode.org/reports/tr35/#UnicodeSubdivisionIdentifier) like  `-u-sd-`, for example [`name:mi-u-sd-nzwko`](https://taginfo.osm.org/keys/name:mi-u-sd-nzwko)

unit tests added for both cases. I'm not certain if these are a good idea, and/or whether they're worth including in diplomat


[^1]: this also affects deprecated language codes like [`iw`](https://r12a.github.io/app-subtags/index?lookup=iw), but that's irrelevant. Ideally [`name:iw`](https://taginfo.osm.org/keys/name:iw) should not exist in OSM.

1ec5: Feel free to post issues about any improvements that could potentially apply to Diplomat. I’ve been unsure what to do about cmn/zh. Despite it being a more specific code, BCP&nbsp;47 and CLDR in practice favor zh-Hans-CN for backwards compatibility. As a result, some mappers have gone around replacing `name:cmn=*` with `name:zh-Hans=*`. I suspect this is motivated more by interoperability with the broader software ecosystem rather than a particular OSM data consumer. It probably deserves a forum discussion at some point.

Diplomat does have a later step to prefer the Unicode extensions, but a more elegant approach would be welcome.

matkoniecz: BTW, I found https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3_Retirements.tab via https://iso639-3.sil.org/code_tables/download_tables listing tripple letter deprecated language codes

are there other double letter deprecated language codes except iw, ji, in ?

k-yle: somehow i missed your question sorry. yes, the javascript API will transform 9 deprecated language codes:

- `bh` → `bho` (Bhojpuri)
- `in` → `id` (Indonesian)
- `iw` → `he` (Hebrew)
- `ji` → `yi` (Yiddish)
- `jw` → `jv` (Javanese)
- `mo` → `ro` (Romanian)
- `sh` → `sr` (Serbian)
- `tl` → `fil` (Filipino)
- `tw` → `ak` (Akan)

<details>
<summary>(determined using brute force)</summary>

```js
for (let i = 0; i < 26; i++) {
  for (let j = 0; j < 26; j++) {
    const old = String.fromCharCode(97+i) + String.fromCharCode(97+j);
    const nēw = new Intl.Locale(old).language;
    if (old !== nēw) {
      const label = new Intl.DisplayNames(navigator.languages, { type: 'language' }).of(nēw);
      console.log(`${old} → ${nēw} (${label})`);    
    }
  }
}
```

</details>

the only one i was suprised to see was [`tl`](https://taginfo.osm.org/keys/name:tl), which is quite common in OSM, and iD also uses [`tl` for localisations](https://github.com/openstreetmap/iD/blob/develop/dist/locales/tl.min.json). i've added a test case for one "deprecated" code, but i don't think we need to anything else.

tyrasd (approved):

PR #10914, opened 2025-03-24
Enable territory-level address hints

### LINKED ISSUE #10906: Items in Hong Kong/Macau are wrongly getting Chinese address hints

Vectorial1024:
As guided by https://github.com/openstreetmap/iD/issues/10899#issuecomment-2745909687 , this PR enables the editor to look at the territory-level country code, and therefore fixes #10906.

Hopefully there are no unexpected side effects.

The major beneficiaries of this PR are most probably the Hong Kong and Macau mapping communities. However, for those communities, this PR is **blocked by #10907**, so at this stage it will exist as a draft. But still, feature-wise, this PR is complete, just need to wait for a good timing.

I see no need to rush this through before #10907 is done, but if there are other territory-level countries that also benefit from this PR but already has a well-defined address format, feel free to convert this from a draft to an actual PR (I think it is only me that can do the conversion?).

### DIFF
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -273,6 +273,7 @@ export function uiFieldAddress(field, context) {
                 countryCode = t('intro.graph.countrycode');
             } else {
                 var center = extent.center();
+                // even if we do not have data for this territory/country, the editor will generate a default format for us.
                 countryCode = countryCoder.iso1A2Code(center, { level: 'territory' });
             }
             if (countryCode) {
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -273,7 +273,7 @@ export function uiFieldAddress(field, context) {
                 countryCode = t('intro.graph.countrycode');
             } else {
                 var center = extent.center();
-                countryCode = countryCoder.iso1A2Code(center);
+                countryCode = countryCoder.iso1A2Code(center, { level: 'territory' });
             }
             if (countryCode) {
                 _countryCode = countryCode.toLowerCase();


Vectorial1024:
In addition, there is a comment elsewhere that is suggesting changes to this PR: https://github.com/openstreetmap/iD/pull/10907#pullrequestreview-2709776125

So this will stay in draft for longer.

Vectorial1024:
I will work on #10904 first, to confirm how this idea should be done (both should be similar), and then do this one.

Vectorial1024:
I wrongly mentioned the Egyptian phone hint problem. Both this and the phone hint issue unfortunately have very similar tracker number.

Vectorial1024:
To clarify for the watchers, the main reason why this needs to wait for #10907 is that afaik there are no existing non-sovereign address hints which can be used to show to an uninvolved reviewer that the new logic works. Then, #10907 is useless without this PR.

Vectorial1024:
Superceded by https://github.com/openstreetmap/iD/pull/12704

PR #10916, opened 2025-03-24
Describe Egypt's phone format

### LINKED ISSUE #10915: Getting Egyptian address hints in Egypt is wrong
### URL

https://www.openstreetmap.org/edit#map=18/30.053622/31.243917

How do we reproduce the issue?
Find any feature inside Egypt that does not have its phone/'fax/(or related fields) filled
Using the editor, add a e.g. Telephone field for editing
See "+31" in the hint text

as the same problem in issue no. #10899  The country code of Egypt is not in the list on the phone format

For example, the phone field has hint text "+31".
Image

2.33.0-dev

Other (please specify), Chrome

youssefelzedy:
While discussing #10899 I realized that the editor also does not currently know about Egypt's phone number format, so I marked the issue #10915  for that and adding it.

As such, this PR is added in Egypt's phone number format.


data/phone_formats.json
```diff
@@ -18,6 +18,7 @@
   "dm": "+1-767-555-1234",
   "do": "+1-809-555-1234",
   "es": "+34 989 12 34 56",
+  "eg": "+20 1 234567890",
```
tyrasd: let's keep the list alphabetically sorted
```suggestion
  "eg": "+20 1 234567890",
  "es": "+34 989 12 34 56",
```

tyrasd:
[Wikipedia](https://ar.wikipedia.org/wiki/%D8%A3%D8%B1%D9%82%D8%A7%D9%85_%D8%A7%D9%84%D9%87%D8%A7%D8%AA%D9%81_%D9%81%D9%8A_%D9%85%D8%B5%D8%B1) seems to suggest that the separator should be a `-` and the digits should be divided into groups of 3 and 4 digits with `+20-1-234-5678`. Is that correct?

youssefelzedy:
Ok i will change it

tyrasd (approved):

PR #10917, opened 2025-03-24
create a random phone number if the local phone format is not specified in the ID

youssefelzedy:
While discussing #10916  @tyrasd suggested that he make a random phone number, as we did not find any number registered in our phone number format list

now I have created a new function that does the following:
1. **Generates a Random Country Code**  
   - Creates a country code between **`+1` and `+99`**.  
   - Example: `+45`, `+83`, `+9`  
2. **Determines a Random Phone Number Length**  
   - The total number (including the country code) is **10 or 11 digits**.  
   - Adjusts the phone number length based on the country code length.  
3. **Generates a Random Phone Number**  
   - Fills the number with random digits (`0-9`).  
4. **Formats the Phone Number with Spaces**  
   - Inserts a space **every 3 digits**, except for the last **4 digits** which remain grouped.  
   - Example format: `+45 123 456 7890`  
5. **Returns a Complete Formatted Phone Number**  
   - Example outputs:  
     - `+5 123 456 7890`  
     - `+78 987 654 321`  
     - `+99 543 210 9876`

### DIFF
--- a/modules/ui/fields/input.js
+++ b/modules/ui/fields/input.js
@@ -345,13 +345,34 @@ export function uiFieldText(field, context) {
         }
     }
 
+    function generateRandomPhoneNumber() {
+        // Generate a random country code between +1 and +999
+        const countryCode = `+${Math.floor(Math.random() * 99) + 1}`;
+
+        // Generate a random phone number with 10 or 11 digits
+        const phoneNumberLength = Math.floor(Math.random() * 2) + 10 - (countryCode.length - 1);
+        let phoneNumber = '';
+
+        for (let i = 0; i < phoneNumberLength; i++) {
+            phoneNumber += Math.floor(Math.random() * 10);
+        }
+
+        // Format the number with spaces every 3 or 4 digits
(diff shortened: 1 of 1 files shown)


tyrasd:
Why? :see_no_evil:

Sorry for the confusion, but in https://github.com/openstreetmap/iD/issues/10915#issuecomment-2748391361, I meant that the current default placeholder `+31…` _is_ just an arbitrary, unspecified, non-specific phone number. In that sense it is kind of _random_, but still perfectly fine as a placeholder if we don't have anything better. Making it actually random does not improve it at all.

If you want to improve it, open a PR on the tagging repository and replace the [`+31`](https://github.com/openstreetmap/id-tagging-schema/blob/v6.10.0/data/fields/phone.json#L9) placeholder with a country code that does not exist (maybe something like `+876` or something else that's fun).

youssefelzedy:
I understood wrongly 🤦 😂
In general, I will work on it, I will first make sure that +876 does not belong to any country number

PR #10921, opened 2025-03-25
Fix: Separate address formats for Hong Kong and Macau from Mainland China

Oscar-Qiu:
Edited address_format.json and address.js to enable custom display of Macau/HongKong address.

macau
hk


tyrasd:
Hi. I think this is already addressed in #10907 / #10914.

PR #10926, opened 2025-03-28
fix: merging two ways corrupts step_count

### LINKED ISSUE #10492: Merging two ways corrupts step_count

chaitanyakadu:
This PR addresses issue [#10492](https://github.com/openstreetmap/iD/issues/10492) by fixing the `step_count` and other bunch of tags corruption when merging ways and enhances tag merging logic.

#### Summary of Changes
- Updated `mergeTags` in `entity.js` to sum `step_count` and other numeric tags correctly during way merges.
- Added support for `sum`, `avg`, `max`, and `min` operations for tags like `capacity`, `maxspeed`, `maxheight`, etc.
- Modified `join.js` to exclude these tags from conflict checks.

#### Optimization
- Centralized tag definitions in a new `tagOperations.js` module to eliminate duplicate lists in `entity.js` and `join.js`, improving maintainability.
- Enabled merging of ways even when values differ for certain tags (e.g., `step_count`, `maxspeed`), applying appropriate operations instead of failing.
- Expanded the scope of #10492 beyond summing (e.g., `step_count`) to include averaging (`maxspeed`), maximizing (`maxheight`), and minimizing (`min_age`) operations for broader applicability.

#### Video
https://github.com/user-attachments/assets/346b3247-c868-48f0-be48-d14348229db7

#### Recommendation
@k-yle @1ec5 @tyrasd
Should we consider moving tagOperations.js file to another directory (./modules/osm/*) to align with iD’s config system?


tyrasd (changes requested):
I think this goes quite a bit too far. Here are some concerns with the current implementation:

* for most of the tags included in this PR it does not make sense to reduce values while merging:
  * `maxheight` / `maxweight` are limits: if anything, the minimum of two objects must be kept.
  * it does simply not make any sense to average `maxspeed` values
  * in all cases except for the `step_count` and `parking:*:capacity` there is a loss of information when reducing the tag values to a single value: e.g. when one were to combine two roads with `width=4` and `width=2`, the resulting `width=3` is not reflecting the reality at all. the previous result `width=4;2` at least indicates that the width is not uniform along the road
* the averaging would drop precision when the values were decimal numbers (e.g. `width=0.5` + `width=0.8` would result in `width=0`)
* the `parking:*:capacity` tags are currently not handled specifically when splitting a way. if the merging operation combines the values, the inverse must also be true for the splitting operation.
* `parking:both:capacity` is missing

I would recommend to simplify this PR:

* make a single list of _summable_ tags, which only includes `step_count` and `parking:*:capacity` for now.
* drop all logic for the `avg`/`max`/`min` cases
* move the list of _summable_ tags to `modules/osm/tags.js`
* extend the logic for splitting values in `modules/actions/split.js` also for the `parking:*:capacity` tag keys
* implement the logic for merging the tags in `modules/actions/join.js` instead of the `entity.js` module

chaitanyakadu:
The latest changes addresses bellow requests:
- [✅] make a single list of summable tags, which only includes step_count and parking:*:capacity for now.
- [✅] drop all logic for the avg/max/min cases
- [✅] move the list of summable tags to modules/osm/tags.js
- [✅] extend the logic for splitting values in modules/actions/split.js also for the parking:*:capacity tag keys
- [❌] implement the logic for merging the tags in modules/actions/join.js instead of the entity.js module

@tyrasd I made multiple attempts to implement the logic for merging the tags in modules/actions/join.js, which often lead to unnecessary side effects. The logic was implemented inside [var action](https://github.com/openstreetmap/iD/blob/develop/modules/actions/join.js#L28). 
Keeping the logic in entity.js seems more appropriate, as all mergeTags-related code is consolidated there. Moving it to join.js would split the logic across two files, increasing complexity for future contributors.

Another issue in current approach is that bellow function is cloned thrice. To maintain DRY code can you recommend a central file where this function could be exported.
```
function matchKeys(key) {
        return summableTags.some((tag) => {
            var regex = new RegExp('^' + tag, 'i' );
            return regex.test(key);
        });
}
```

tyrasd (approved):
>  implement the logic for merging the tags in modules/actions/join.js, which often lead to unnecessary side effects

I've pushed some additional changes to address the remaining issue :arrow_up: . The `osmEntity` module should really not have any logic that is specific to certain operations. It is now solved by an additional parameter of `.mergeTags` that specifies extra tags which to use "as is", in order to avoid side effects of the generic `valueA;valueB` merge logic.

PR #10935, opened 2025-04-04
`directionalCombo`: handle `[key]=left|right|both` schema (Support `sidewalk=right`)

### LINKED ISSUE #10839: Enhance the `directionalCombo` to support `sidewalk=both|left|right`
In https://github.com/openstreetmap/id-tagging-schema/pull/454#issuecomment-2702860236 we came to the conclusion to add support to display sidewalk data on highway values.

There are three tagging schema in use. We support two already, but need to add the third one (which is in use 2.248.360 times).

In https://github.com/openstreetmap/iD/pull/9587 the `directionalCombo` was improved to support two of the variations. We now want to add the third which only applies to the `sidewalk` tag.
* Support `sidewalk=both` => and display it as `sidewalk=yes` (which is `sidewalk:both=yes`)
* Support `sidewalk=left` => and display it as `sidewalk:left=yes + sidewalk:right=no`
* Support `sidewalk=right` => and display it as `sidewalk:left=no + sidewalk:right=yes`

tordans:
This resolves https://github.com/openstreetmap/iD/issues/10839

Following what was done in https://github.com/openstreetmap/iD/pull/9587/files

---

This is ready for review, see https://github.com/openstreetmap/iD/pull/10935#issuecomment-2973569136


tordans:
@k-yle @tyrasd can you give me some pointers as to why the specs don't work and if this is generally a sane way to do this?

gy-mate:
@k-yle @tyrasd Could you please help @tordans with the failing test specs? As @waldyrious previously [pointed out](https://github.com/openstreetmap/id-tagging-schema/pull/1507#issuecomment-2779865864), it would be great to have this feature!

Many thanks in advance! :)

k-yle (commented):
@gy-mate we discussed this on slack, there's two lines that need to be added [to `onChange`](https://github.com/openstreetmap/iD/blob/e07dcddfee4c26e4f8e59b967cb0ac96bc619b5a/modules/ui/fields/directional_combo.js#L96) which will fix the tests and the ui bug:

```diff
+ if (otherValue === 'left') otherValue = otherKey.endsWith(':right') ? 'yes' : 'no';
+ if (otherValue === 'right') otherValue = otherKey.endsWith(':left') ? 'yes' : 'no';
```

(commenting here since I can't start a thread on unchanged lines)

tordans:
@k-yle thanks a lot, that got me on the right track, finally. 
Here is the current code with green specs:


https://github.com/user-attachments/assets/0ca9d731-2ebf-4fdc-a40f-1a49b9d6383d

IMO this is finished, now, and ready for a "ready to merge" review.

gy-mate:
Awesome, thank you so much! :)

tordans:
On a separate note, the specs are not as nice as I wanted them to be. Especially https://github.com/openstreetmap/iD/pull/10935/files#diff-f10edc7d646966248a759f9414d26ed695bf57d21db38f18b76f2146d8528d63R231-R243
I can only test left and right separately and check that one value was updated correctly but have to ignore the other values (see "ignore" comment). Ideally we had a check below that that would show that after triggering the update for each side, that both values are now updated. I tried duplicating https://github.com/openstreetmap/iD/pull/10935/files#diff-f10edc7d646966248a759f9414d26ed695bf57d21db38f18b76f2146d8528d63R225-R229 again for that but that does not work.

I think it's fine now, but curious about how to actually do it…

tyrasd (approved):
While testing, I found that the PR did not handle the main use case for which we need it (https://github.com/openstreetmap/id-tagging-schema/pull/1507) very gracefully. For example, when the preset uses `sidewalk:both` as the _common_ key, the transformation of `sidwalk=left` would not result in the fields to show it like as if it was `sidewalk:left/right=yes/no`. There were also some corner cases that were not handled correctly, e.g. a present tag of the `otherKey` should never be overwritten by the fallback mechanism (e.g. `sidewalk=left + sidewalk:right=separate` should not leave the `sidewalk:right` tag untouched when changing the _left_ value of the field).

Lastly, I also found some tricky edge cases for when a multi-selection is containing several OSM objects that use different tagging "variants" (e.g. one tagged as `sidewalk:left=yes` and once `sidewalk=both`). Strictly speaking, that was already a problem in the current code (e.g. reproducible when there is one `cycleway:left=lane` and one `cycleway=track`: it would show as if the multiselection of both had a _Bike Track_ on the right side of both and only conflicting values for the left side), but as this very much related to the code of this PR anyway, I also addressed it was well.

See e2cf8e979499f7c70b4bbac1a30e326baf4722fc (and 66d3e23f) for the updated logic. //edit PS: it also generalized the code a little further, as to not assume that `:left`/`:right` are the two allowed direction options (there could be new fields in the future using `:forward:`/`backward` for example) nor that the `:<direction>` part is always at the end of the tag key (there's already the `parking:<direction>:orientation` field for which this is not the case).

Issue #2891, opened 2015-12-30
Multiple relations of same category (and name) are only shown once


IntendedCrash:
When trying to add an entity as a member to an already existing relation, the relation list shows relations of the same category (and with the same name) only once. Especially generic multipolygon relations (e. g. landuse=forest) without any name create such kinds of collisions. If there are multiple relations of that kind in the area, the users can see only one thus they are adding the object maybe to the wrong one just because it's the only one visible. Even if they would be aware of the situation, there is no possibility to select a relation not shown in the list.

Reason for this is the construction of the keys for the selection list. The function [RawMembershipEditor.relations()](https://github.com/openstreetmap/iD/blob/d8e627b9b13af1e1d03e8001673cdcb67955f892/js/id/ui/raw_membership_editor.js#L57) creates non-unique values, which are redirected via [d3.combobox.render()](https://github.com/openstreetmap/iD/blob/66e8f78aa77f96ff860aceb8b5c188d67b32d9c1/js/lib/d3.combobox.js#L209) as keys to [d3_selectionPrototype.data()](https://github.com/openstreetmap/iD/blob/f67d51492a81a325ed1a63ff3dcf2e4df7d7349c/js/lib/d3.v3.js#L1127) where they are required to be unique.

My proposal for a solution would be to implement a value check within [RawMembershipEditor.relations()](https://github.com/openstreetmap/iD/blob/d8e627b9b13af1e1d03e8001673cdcb67955f892/js/id/ui/raw_membership_editor.js#L67) before returning the result array. If there are multiple relations with the same value in result, add the respective IDs to the affected values. Thus keys for the selection list would be unique and also users can distinguish between multiple relations (because a list with "Forest", "Forest", "Forest" wouldn't be very useful when making multiple edits; would require to look for the correct relation every time again).

bhousel:
Thank you for the very descriptive bug report.
I fixed this in 4dd4ccb using your suggestion to append the id to any duplicate names.

PR #10942, opened 2025-04-08
disambiguate relations with same names in membership editor

tyrasd:
Shows the relations' ids as hover text in membership editor as a last resort in case multiple relations have otherwise the exact same label (and type).

Also shows the "regular" relation label in form of the hover text for non-ambiguous relations. This is useful to see the full name in case it is cut off with an ellipisis (…) for long-ish labels.

PS: This extends upon #2891


modules/ui/sections/raw_membership_editor.js
```diff
@@ -419,7 +415,8 @@ export function uiSectionRawMembershipEditor(context) {
             .attr('class', 'member-entity-name')
             .classed('has-colour', d => d.relation.tags.colour && isColourValid(d.relation.tags.colour))
             .style('border-color', d => d.relation.tags.colour)
-            .html(function(d) {
+            //.attr('title', d => JSON.stringify(d))
```
1ec5: This comment can be removed, right?

tyrasd: Right, I forgot to clean up this debug statement. Thanks for spotting it!

PR #10949, opened 2025-04-09
suggest housenumber/housename values from surrounding areas

### LINKED ISSUE #10946: The editor does not suggest addr:housename values
While determining the exact ability for the editor to auto-suggest the various address fields, I have found that the editor does NOT auto-suggest the `addr:housename` field.

For example, when looking at features inside a shopping mall in London, UK ("Surrey Quays Shopping Centre"), those features do not have values for the `addr:housename` field, and yet while the shopping mall itself has `addr:housename`, the features inside do not get that name suggested as a possible value of `addr:housename`.

See a feature inside the shopping mall: https://www.openstreetmap.org/edit?node=10745161070#map=19/51.495245/-0.047218

Image

The `addr:housename` field does not have a drop down, which means the editor cannot suggest any values to this field.

Is this an oversight or is this a design decision?

tyrasd:
If a feature has an address field and the feature is fully encompassed by another feature with address tags: also offer the values for `addr:housenumber` and `addr:housename` as dropdowns with autocomplete functionality. This can be useful for POIs like shops that are inside of a larger building and thus share the same address (incl. housenumber/housename).

Fixes #10946.


tordans:
A different mapping style that would fit this use case would be a address node that is located in the same building shape as the shop node that I am editing. 

In cities that is a common style when address data was imported and or the shapes don't match the address.

tyrasd:
Good point. That was luckily also relatively straight forward to implement using the existing logic.

Vectorial1024:
While we are at it, perhaps we may consider another case:

Buildings inside `landuse=*` should probably also get housename hints. Currently this is not the case in the iD editor.

Not sure about any international examples, but at least in Hong Kong, some property development may include several buildings (e.g. Blocks 1 through 8). Usually there will be a single `landuse=residential` enclosing the 8 buildings, and it is the `landuse=residential` that gets its own `addr:housename` (and also `addr:housenumber`). Each of the 8 buildings in this example does not have its own "street + housenumber" address. but will simply state they are inside the same property development, ala same `addr:housename` value as the property development `landuse=residential`.

Would this be possible to implement for the iD editor?

tyrasd:
@Vectorial1024 do you have an example or two where this is the case?

Vectorial1024:
Sure! @tyrasd I can provide a real-life example.

Consider The Chinese University of Hong Kong ("CUHK"), a prominent university from Hong Kong. You may have heard of CUHK in recent years through various means.

But still, some prominent features of CUHK are that it is quite literally built upon a hill, but retains the traditional university planning of having many individual buildings instead of "modern style" few but large buildings. In OSM terms, it is a relation with `type=multipolygon;landuse=education` ([link](https://www.openstreetmap.org/relation/7802779)) that encloses dozens of buildings.

It is easy to notice there are many named streets inside CUHK, but when addressing for the dozens of buildings inside CUHK, `addr:street` and `addr:housenumber` is not used. We just write "CUHK" as the `addr:housename` instead.

The `addr:housename` of the main campus is provided via another `landuse=education` feature ([link](https://www.openstreetmap.org/way/33277733)).

Example addresses:

- Faculty of Engineering: `[...] Ho Sin-Hang Engineering Building, The Chinese University of Hong Kong [...]` (https://www.erg.cuhk.edu.hk/erg/ContactUs); the building is [this](https://www.openstreetmap.org/way/135312912)
- Faculty of Science: `[...] Science Centre, CUHK [...]` (https://www.sci.cuhk.edu.hk/faculty/contact-us/faculty-office/); the building is [this](https://www.openstreetmap.org/way/134410446)
- Faculty of Arts: `[...] Fung King Hey Building, The Chinese University of Hong Kong [...]` (https://www.arts.cuhk.edu.hk/web/); the building is [this](https://www.openstreetmap.org/way/135265380)

etc. All these buildings are inside the main campus of CUHK, and so should take `addr:housename` as "The Chinese University of Hong Kong".

------

We can see buildings inside a `landuse=*` feature should still get `addr:housename` auto-suggestions.

tyrasd:
Ok, large campuses like universities, industrial or office complexes and maybe even gated communities are also in this category. But these are complicated: the situation is typically quite specific to the respective location and could be quite different for different university campuses, for example.

I would say that it is fine for mappers to continue to map these specific addresses individually/manually.

<hr>

>  should take addr:housename as "The Chinese University of Hong Kong".

I'm not so sure about this. When looking at the institute's website:

image

I think this should be tagged as (`addr:unit/door=606`) `addr:housename=Ho Sin-Hang Engineering Building` and probably `addr:place=The Chinese University of Hong Kong`. This seems also like the institutes are currently mapped on the campus: [example](https://www.openstreetmap.org/way/546320461)

PR #10967, opened 2025-04-15
Don't treat closed `waterway=dam` as area without `area=yes`

quincylvania:
`waterway=dam` can be mapped as either a line or an area, and consumers like renderers will want to treat the two cases differently. Generally if a tag can be applied to either geometry, and `area=yes` tag is required to differentiate them. However, iD currently has a hardcoded exception to this for dams, meaning the `area=yes` tag is missing for many dams-as-areas in OSM. This means that data consumers will need to manually add the same exception in order to handle these properly, which isn't great. Removing the exception will help encourage unambiguous mapping, making it easier to parse the data.

### DIFF
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -63,9 +63,6 @@ export var osmAreaKeysExceptions = {
         wash: true,
         ventilation_shaft: true
     },
-    waterway: {
-        dam: true
-    },
     amenity: {
         bicycle_parking: true
     }


tyrasd:
> Generally if a tag can be applied to either geometry, and `area=yes` tag is required to differentiate them.

This does not match the [documentation](
https://wiki.openstreetmap.org/w/index.php?title=Tag:waterway%3Ddam&oldid=2811398#How_to_map) of the intended usage of the tag `waterway=dam` as areas.

It would also break the rendering of ~80% of the existing dam areas in iD, that are not tagged with `area=yes`. For some additional context: it seems that the remaining 20% (i.e. closed `waterway=dam` ways with `area=yes`) were also not contributed "organically", as the growth stopped as soon as iD's handling was adjusted to match the _how to map_ section of the wiki in 04a07092d24a35dd6275b423efa3e733b6563d27.

PR #10999, opened 2025-04-26
:bug: Warn on same layer crossing regardless of bridge / tunnel

### LINKED ISSUE #7809: Don't mark crossing paths of different layers as a problem
Title says it all, I think. I added a path on top of an underground path, and it was marked as a problem and I was asked if I wanted to connect them. They obviously can not connect since one is underground and the other is not :) The underground path was labeled as layer=-1, and the top one as layer=1.

Link to the place: https://ideditor.netlify.app/#background=geovekst-nib&disable_features=boundaries,rail&id=w-6&map=19.28/59.91377/10.78802

Thanks!

### LINKED ISSUE #10631: A bridge or tunnel should not suffice to stop warnings about crossing ways
### URL

_No response_

1. Draw two crossing ways, one with `highway=residential` and `bridge=yes`, one with `waterway=river`.
2. Draw two crossing ways with different `layer` but neither `bridge` nor `tunnel`.

Actual:

Neither is caught by the validator.

Expected:

1 should be flagged: With the noted exceptions of indoor/corridor/building/etc, crossing ways (highway, waterway, railway) having the same `layer` is never correct: They should either intersect or have different layers.
2 should be flagged: If two highway/waterway/railway ways cross each other, it's a virtual certainty that one is a bridge or tunnel. Tagging just the layer is not enough.

Having a bridge at layer=0 cross a non-bridge at layer=0 may seem unambiguous, but it confuses rendering. Screenshots near https://www.openstreetmap.org/changeset/160586698 with Tracestrack Topo:

Screenshot 2024-12-25 15 19 02Screenshot 2024-12-25 15 18 57Screenshot 2024-12-25 15 18 53Screenshot 2024-12-25 15 18 49

The sidewalk has layer=1; the highway has layer=0; the sidewalk casing obscures the roadway.

2.31.0

_No response_

homersimpsons:
Fixes #10631


homersimpsons:
Shoutout to the project, it was really easy to setup a dev environment on GitHub Codespace and run iD and its test!

Here you can see the new check working:
image

- One way is a bridge, the other is not
- Both have the same defined layer
- => The warning is raised

test/spec/validations/crossing_ways.js
```diff
@@ -373,6 +373,11 @@ describe('iD.validations.crossing_ways', function () {
         verifySingleCrossingIssue(validate(), null);
     });
 
+    it('flags road bridge crossing road on the same layer', function() {
```
homersimpsons, line 394: I'm not sure about the test positions in this file, but here seems to be the most sensible place.

modules/validations/crossing_ways.js
```diff
@@ -89,36 +89,21 @@ export function validationCrossingWays(context) {
...
         if (featureType1 === 'waterway' && featureType2 === 'highway' && tags2.man_made === 'pier') return true;
         if (featureType2 === 'waterway' && featureType1 === 'highway' && tags1.man_made === 'pier') return true;
 
-        if (featureType1 === 'building' || featureType2 === 'building' ||
-            taggedAsIndoor(tags1) || taggedAsIndoor(tags2)) {
-            // for building crossings, different layers are enough
-            if (layer1 !== layer2) return true;
-        }
-        return false;
+        if (tags1.layer !== undefined && tags1.layer === tags2.layer) return false; // Warn if both have the same defined layer
```
homersimpsons, line 96: I think I can improve the comments, but I don't really have a great wording at this moment.

modules/validations/crossing_ways.js
```diff
@@ -89,36 +89,21 @@ export function validationCrossingWays(context) {
...
+
+        const isElement1Bridge = allowsBridge(featureType1) && hasTag(tags1, 'bridge');
+        const isElement2Bridge = allowsBridge(featureType2) && hasTag(tags2, 'bridge');
+        if (isElement1Bridge ^ isElement2Bridge) return true; // Either one is bridge, the other is not
+        
+        const isElement1Tunnel = allowsTunnel(featureType1) && hasTag(tags1, 'tunnel');
+        const isElement2Tunnel = allowsTunnel(featureType2) && hasTag(tags2, 'tunnel');
+        if (isElement1Tunnel ^ isElement2Tunnel ) return true; // Either one is tunnel, the other is not
```
tordans: Is the ^ correct here? Not ||?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_operators

homersimpsons: Yes, because in case we have 2 `tunnel` crossing we want to warn (in my opinion) if they have the same (or no layer).

I can expend the xor for it to be more readable if you prefer:
```javascript
(isTunnel1 && !isTunnel2) || (!isTunnel1 && isTunnel2) 
```

I will add a test case also specifically for this.

tordans: TIL. Never saw JS [XOR](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR) in the wild. Personally I think using && and || is more common and easier to understand even if a bit longer…

Dimitar5555: `isElement1Tunnel != isElement2Tunnel` seems clearer and easier to read.

modules/validations/crossing_ways.js
```diff
@@ -89,36 +89,21 @@ export function validationCrossingWays(context) {
...
         // don't flag crossing waterways and pier/highways
         if (featureType1 === 'waterway' && featureType2 === 'highway' && tags2.man_made === 'pier') return true;
         if (featureType2 === 'waterway' && featureType1 === 'highway' && tags1.man_made === 'pier') return true;
 
-        if (featureType1 === 'building' || featureType2 === 'building' ||
-            taggedAsIndoor(tags1) || taggedAsIndoor(tags2)) {
-            // for building crossings, different layers are enough
-            if (layer1 !== layer2) return true;
-        }
-        return false;
```
tordans: Is this `building` case really still covered by the code? Are there tests that make sure it is?

tordans (commented):
Just one thing caught my eye. Cannot tell too much else about this without diving deep because I don't know too much about the code and test coverage. This is something for Martin to review IMO.

modules/validations/crossing_ways.js
homersimpsons: > Is this building case really still covered by the code?

Yes, it falls in both
- the early exit case `if (tags1.layer !== undefined && tags1.layer === tags2.layer) return false;` 
- and the general case `return (tags1.layer || '0') !== (tags2.layer || '0');` 

as those handle anything and hence the `building`s.

> Are there tests that make sure it is?

I believe yes, for example by this test:
https://github.com/openstreetmap/iD/blob/4dc4e7dbdc87a847589c3584f19860bef9ce9aad/test/spec/validations/crossing_ways.js#L151-L155

homersimpsons:
Should I improve / change anything here?

> This is something for Martin to review IMO.

Looks like I cannot assign you @tyrasd

homersimpsons:
@tyrasd do you want me to do anything to help getting this merged? Like adding tests? Feel free to asks and possibly suggest some changes.

To sum up a bit the changes:
1. The main change is to warn if 2 elements have the same defined `layer` and intersect
2. I rewrote the `tunnel` / `bridge` part to be more concise and explicit (and to remove the case that is already handled by (1))
3. The `building` part is already included in the more general check (1) as discussed in https://github.com/openstreetmap/iD/pull/10999#discussion_r2063755697
4. I added 2 tests for the above changes, all the other tests are untouched

test/spec/validations/crossing_ways.js
```diff
@@ -383,6 +388,11 @@ describe('iD.validations.crossing_ways', function () {
         verifySingleCrossingIssue(validate(), null);
     });
 
+    it('flags road tunnel crossing road on the same layer', function() {
+        createWaysWithOneCrossingPoint({ highway: 'residential', tunnel: 'yes', layer: '-1' }, { highway: 'tertiary', layer: '-1' });
+        verifySingleCrossingIssue(validate(), {});
+    });
+
```
k-yle, line 413:
```suggestion

    it('flags a crossing where only one feature is indoor', () => {
        createWaysWithOneCrossingPoint({ highway: 'path' }, { highway: 'path', level: '1' });
        verifySingleCrossingIssue(validate(), {});
    });
```
i think one more test case would be good, to confirm that this behaviour is unchanged (1 indoor feature + 1 non-indoor feature)

modules/validations/crossing_ways.js
```diff
@@ -89,36 +89,21 @@ export function validationCrossingWays(context) {
...
+        if (tags1.layer !== undefined && tags1.layer === tags2.layer) return false; // Warn if both have the same defined layer
+
+        const isElement1Bridge = allowsBridge(featureType1) && hasTag(tags1, 'bridge');
+        const isElement2Bridge = allowsBridge(featureType2) && hasTag(tags2, 'bridge');
+        if (isElement1Bridge !== isElement2Bridge) return true; // Either one is bridge, the other is not
+
```
k-yle:
```suggestion

```
_([from eslint](https://eslint.org/docs/latest/rules/no-trailing-spaces))_

k-yle (approved):
the idea and implementation both look good to me, one suggestion below:

modules/validations/crossing_ways.js
homersimpsons: Thanks (usually my IDE fixes this automatically but looks like GitHub codespace does not no this by default). Fixed and force-pushed.

Note that I didn't delete the entire line as suggested but I just removed the extra spaces as I find that the code is way more readable with the "bridge block" and the "tunnel block". Ping me if you really want to delete the entire line.

test/spec/validations/crossing_ways.js
homersimpsons: thanks for the suggestion, added as-is

k-yle:
if there are no objections, i think this could be merged

k-yle:
I'm going to merge this one, there's been no other comments in the past 6 months, and I think this is an uncontroversial approach to solving the issue, which manages to keep all existing unit tests passing.

PR #11003, opened 2025-04-27
feat-deletion-in-changeset-comment dropdown

### LINKED ISSUE #10957: Add "delete" (×) button next to each changeset comment in dropdown
### Description

# descripton :

When entering a new changeset comment in the Upload to OpenStreetMap panel, the previously entered comments appear in a dropdown.
I’d like to have a "×" (cross/delete) button next to each comment in the dropdown list to allow removing individual saved comments.

# proposed behavior :

- Add a small × icon next to each dropdown comment.

- On clicking the × icon then: 
    - the selected comment is deleted from the saved history.
    - it no longer appears in future sessions.

### Screenshots

Image

As you can see in this image there is no option to remove a particular comment from the dropdown.

draunger:
# description :
- added the deletion of changeset comment 
# fixes :
https://github.com/openstreetmap/iD/issues/10957

### DIFF
--- a/modules/ui/changeset_editor.js
+++ b/modules/ui/changeset_editor.js
@@ -74,10 +74,12 @@ export function uiChangesetEditor(context) {
                 osm.userChangesets(function (err, changesets) {
                     if (err) return;
 
+                    var deleted = (JSON.parse(localStorage.getItem('deletedChangesetComments') || '[]')).map(s => s.trim().toLowerCase());
                     var comments = changesets.map(function(changeset) {
                         var comment = changeset.tags.comment;
                         return comment ? { title: comment, value: comment } : null;
-                    }).filter(Boolean);
+                    }).filter(Boolean)
+                      .filter(c => !deleted.includes(c.value.trim().toLowerCase()));
 
                     commentField
                         .call(commentCombo
--- a/modules/ui/combobox.js
+++ b/modules/ui/combobox.js
@@ -390,10 +390,37 @@ export function uiCombobox(context, klass) {
                 })
                 .attr('title', function(d) { return d.title; })
                 .each(function(d) {
+                    var option = d3_select(this);
                     if (d.display) {
-                        d.display(d3_select(this));
+                        d.display(option);
                     } else {
-                        d3_select(this).text(d.value);
+                        option.text(d.value);
+                    }
+                    // Only for comment combobox: add cross button, floated right, minimal style
+                    if (klass === 'comment') {
+                        option.append('button')
+                            .attr('class', 'combobox-delete')
+                            .attr('title', 'Delete this comment')
+                            .style('float', 'right')
+                            .style('font-size', '16px')
+                            .style('cursor', 'pointer')
+                            .text('×')
+                            .on('mousedown', function(event) {
+                                event.stopPropagation();
+                                event.preventDefault();
+                            })
+                            .on('click', function(event, d) {
+                                event.stopPropagation();
+                                event.preventDefault();
+                                let deleted = JSON.parse(localStorage.getItem('deletedChangesetComments') || '[]');
+                                let valueNorm = d.value.trim().toLowerCase();
+                                if (!deleted.map(s => s.trim().toLowerCase()).includes(valueNorm)) {
+                                    deleted.push(d.value);
+                                    localStorage.setItem('deletedChangesetComments', JSON.stringify(deleted));
+                                }
+                                d3_select(this.parentNode).remove();
+                            });
+
                     }
                 })
                 .on('mouseenter', _mouseEnterHandler)


modules/ui/changeset_editor.js
```diff
@@ -74,10 +74,12 @@ export function uiChangesetEditor(context) {
                 osm.userChangesets(function (err, changesets) {
                     if (err) return;
 
+                    var deleted = (JSON.parse(localStorage.getItem('deletedChangesetComments') || '[]')).map(s => s.trim().toLowerCase());
```
tordans, line 77: This logic `.map(s => s.trim().toLowerCase());` is duplicated in a lot of places. Do we really need it? We only ever compare the full string, to it should be OK to compare it as it. Otherwise we might want to extract it into a helper?

Another way would be to only store and compare hashes of those lines… but I don't think that is needed / has benefits.

modules/ui/changeset_editor.js
```diff
@@ -74,10 +74,12 @@ export function uiChangesetEditor(context) {
...
                     var comments = changesets.map(function(changeset) {
                         var comment = changeset.tags.comment;
                         return comment ? { title: comment, value: comment } : null;
-                    }).filter(Boolean);
+                    }).filter(Boolean)
+                      .filter(c => !deleted.includes(c.value.trim().toLowerCase()));
```
tordans, line 82: The `filter(Boolean)` should always be the last IMO.

modules/ui/combobox.js
```diff
@@ -390,10 +390,37 @@ export function uiCombobox(context, klass) {
...
                 .each(function(d) {
+                    var option = d3_select(this);
                     if (d.display) {
-                        d.display(d3_select(this));
+                        d.display(option);
                     } else {
-                        d3_select(this).text(d.value);
+                        option.text(d.value);
+                    }
+                    // Only for comment combobox: add cross button, floated right, minimal style
```
tordans, line 399: Can we somehow extract this into a separate file? I don't think it is ideal when this code block has to be read by every one who works on this component when it is really a separate component or feature.

modules/ui/combobox.js
```diff
@@ -390,10 +390,37 @@ export function uiCombobox(context, klass) {
...
+                                event.preventDefault();
+                            })
+                            .on('click', function(event, d) {
+                                event.stopPropagation();
+                                event.preventDefault();
+                                let deleted = JSON.parse(localStorage.getItem('deletedChangesetComments') || '[]');
+                                let valueNorm = d.value.trim().toLowerCase();
+                                if (!deleted.map(s => s.trim().toLowerCase()).includes(valueNorm)) {
+                                    deleted.push(d.value);
+                                    localStorage.setItem('deletedChangesetComments', JSON.stringify(deleted));
```
tordans, line 419: How will this get reset? Like, when I click it 100 times, all those will be in the stack forever, right? Should we maybe allow for a max of 100 items and remove the last items? Like a max array size…

tordans (commented):
A few thoughts I had when reading the code in Github.

modules/ui/changeset_editor.js
tyrasd: I think in this case it is ok like this: the `.map(…).filter(Boolean)` belongs together, as it is just a shortcut/trick to not have to write `.filter(funct).map(funct')` with two times almost the same function body.

modules/ui/changeset_editor.js
tyrasd: I would even go as far as to store the changeset id of the to be ignored comments instead of the actual literal comment string itself. That would be more memory efficient and be the most stable implementation.

modules/ui/combobox.js
tyrasd: It should be safe to remove any stored entries that do not match any changeset that is currently returned by the OSM API anymore.

modules/ui/combobox.js
```diff
@@ -390,10 +390,37 @@ export function uiCombobox(context, klass) {
...
+                        d.display(option);
                     } else {
-                        d3_select(this).text(d.value);
+                        option.text(d.value);
+                    }
+                    // Only for comment combobox: add cross button, floated right, minimal style
+                    if (klass === 'comment') {
+                        option.append('button')
+                            .attr('class', 'combobox-delete')
+                            .attr('title', 'Delete this comment')
```
tyrasd, line 403: User facing texts should never be hardcoded to English. See https://github.com/openstreetmap/iD/wiki/How-to-Use-iD%27s-Localizer-Module for how to add and use translatable strings.

modules/ui/combobox.js
tyrasd: yeah… at the very least we need to make this not hardcoded to a specific UI field: The logic should be reusable for other fields like the changeset `source` for example.

I could think of two ways to approach this:
* leave the rendering code here, but introduce a new configuration method to enable "removable" entries for a specific combobox (`uiCombobox(…).setRemovable(callback)`). The callback function is called whenever an `×` is clicked on.
* use the existing possibility to supply the [`display` property](https://github.com/openstreetmap/iD/blob/3a10bf808c1cf98085b6e3241e10ef5ab0ff76de/modules/ui/combobox.js#L10-L11) on the to be rendered comments , and append the `×` and respective logic there.

As the logic is currently only needed for the changeset editor, maybe the second approach is more efficient…

modules/ui/changeset_editor.js
tyrasd: what also needs to be taken into account in the store-changeset-id approach is that in case that there exists more than one changeset with the to be ignored comment, all respective changeset ids need to be added to the list. but this does come with an implicit benefit: when the user at some later point manually enters a previously deleted comment, it will be again visible in the suggestions (until deleted again). When storing the comment itself, a user would not be able to undo an accidental deletion of a specific comment.

tyrasd:
closing as stale

PR #11015, opened 2025-05-01
remove empty line from drag_note.js

McLaynV:
It confuses some tools if a `*.js` file starts with an empty line.


tyrasd (approved):
I also added the remaining couple of files that had the same "issue"

PR #11016, opened 2025-05-03
Translations core.yaml `ID`

tordans:
Resolves https://github.com/openstreetmap/iD/pull/10046#discussion_r2064037631


1ec5 (approved):
Thanks!

Issue #8477, opened 2021-04-29
Improve warning for unsquare buildings


aawiseman:
The not-square building warning seems somewhat inconsistent, I don't always see it when mapping. For example, I didn't square this one and didn't get a warning https://www.openstreetmap.org/changeset/103864837

Maybe in addition to the existing warning there can be another such as if there are four nodes in the building, check if the angles of the ways are between 75 and 105 degrees, or something like that. I would guess most buildings with just 4 nodes should be square, rectangle, etc.

cliffordsnow:
My experience is that many buildings new mappers draw have more than four nodes. I would include everything with 4 or more nodes for validation.

k-yle:
the inconsistent behaviour is caused by a bug on this line:

https://github.com/openstreetmap/iD/blob/2a5c68aa04cb4279e5af0d24f00bf7ae9b0635dd/modules/validations/unsquare_way.js#L58-L59

If you have never customised this degree input:

<img width="374" height="43" alt="Image" src="https://github.com/user-attachments/assets/0c4d23aa-ddce-4318-98e7-a39050528e95" />


...then `prefs()` returns `null`, and `isFinite(null)` is true because of javascript nonsense 😆. Therefore, the default threshold is 0° (off) instead of 5°

So: to activate this validator, you just need to interact with this input field once

Issue #6332, opened 2019-05-07
Add fix to tag real unsquare buildings as unsquare


quincylvania:
We want buildings to have nice square corners in OSM—unless they're not square in the real world. iD should offer a fix like "Tag as accurate" that will add a tag like `unsquare=yes` to indicate a building is known to be truly irregular. This will prevent iD and other tools from continuing to falsely flag the issue in the future.

![Screen Shot 2019-05-07 at 3 19 46 PM](https://user-images.githubusercontent.com/2046746/57326923-93b87980-70db-11e9-8af8-6fbb02a53fb3.png)

Other false validation issues are already tagged similarly in OSM, such as with [`noexit=yes`](https://wiki.openstreetmap.org/wiki/Key:noexit) and [`noname=yes`](https://wiki.openstreetmap.org/wiki/Key:noname).

Nakaner:
Although `noname=yes` is common, it is not that common that it can serve as an argument in favour of introducing `unsquare=yes`. In difference to `noexit=yes`, `unsquare=yes` and `noname=yes` only serve as a workaround for quality assurance tools. `noexit=yes` also conveys information for map users: There road ends here.

Some people prefer to tag as complete as possible and add `oneway=no`, `cycleway=no`, `lit=no` etc. to any way. However, such a practice is not base on a broad consensus and if you dig deep enough in the history of user blocks in OSM, you might find blocks set due to an excessive use of negative binary tags.

I think that iD does not need this tag and should only validate buildings if they have been added or modified in the current session. If doing so, they will be reported once which does not bother that much.

Adding such a tag is not a simple change as it might seem to be and I ask you to discuss it with the broader community on the Tagging mailing list.

pnorman:
I agree that this needs to be proposed to the broader community and discussed before encouraging people to add such a tag.

quincylvania:
I did this! I ended up using `nosquare` for the tag to align it with `noname` and `noexit`. Mappers can now fully resolve all unsquare way issues. If you manually square the feature later, then the tag is automatically removed.

![unsquare demo](https://user-images.githubusercontent.com/2046746/57468685-f6805100-7252-11e9-864e-045feb7b772d.gif)

@pnorman said:

> I agree that this needs to be proposed to the broader community and discussed before encouraging people to add such a tag.

[...]

PR #11020, opened 2025-05-05
create a new warning tier for low-severity suggestions

### LINKED ISSUE #6101: Add lower issue severity for inquisitive issues
iD's current validation system presets issues as problems: either Warnings or Errors. However, there are times such as #5835 where a feature is not particularly wrong but could use more detail.

We should add a third kind of issue, a Prompt, that just asks the user for more information if they know it. These would mostly be phrased as questions with one-click answers. They would not appear on the upload panel. Generally, they could be disabled by default.

Screen Shot 2019-03-25 at 2 52 25 PM

### LINKED ISSUE #10743: Adopt Rapid's issue level 'suggestion' for lower severity issues
### Context

Rapid recently added a third issue type "Suggestion" to the existing Warning and Errors.
- This is what it looks like https://github.com/facebook/Rapid/issues/1623
- Most of the code changes are https://github.com/facebook/Rapid/commit/d75885150760ecc9fe8c6bfc84e761f7d4cc4d70 and https://github.com/facebook/Rapid/commit/3e44d588fc05a68562df54f5b60950fbed1ca251

We have an existing issue on this in https://github.com/openstreetmap/iD/issues/6101 which links to quite a few other issues that would benefit from this feature.

The Rapid code changed quite a bit since it's fork but in general the existing implementation should make it a lot easier to implement this in iD.

### Goal

- Test out Rapid's implementation and discuss if this is what we want for iD.
   This is about the issue panel itself but also about the sidebar, status bar, changelog tags and other parts that the issues system touches.

- Adopt the changes from Rapid to iD. 
   This might require to change the implementation from what Rapid does in case this used Rapid specific code path that iD does not have.

---

FYI I added this as a suggestion for a GSoC Project (as a package with #10757): https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2025/Project_ideas#iD_editor

k-yle:
Closes #6101, Closes #10743

Created a new third tier for validation issues. `suggestion` (blue) is the lowest tier

image

I've changed the 'Unsquare Corners' validator to this new low-severity type, since it's currently off by default[^1] so this change won't have any noticable impact. We can discuss where to use this tier in a followup issue, this PR is just the technical implementation.

Comparison of the severity tiers:

|  | `suggestion` |`warning` | `error` |
|-|--------|--------|--------|
| Prevents the changes from being uploaded |  |  | ✅ | 
| Shown on the changeset page | | ✅ | ✅ |
| Added to the changeset tags as `warning:*` | | ✅ | ✅ |
| Shown in the 'issues' panel | | ✅ | ✅ |
| Shown when a feature is selected | ✅  | ✅ | ✅ |

[^1]: I can't tell whether it's intentionally off by default, it actually seems to be a bug (#8477 ?)

### DIFF
--- a/modules/core/validation/models.js
+++ b/modules/core/validation/models.js
@@ -4,7 +4,7 @@ import { t } from '../../core/localizer';
 export function validationIssue(attrs) {
     this.type = attrs.type;                // required - name of rule that created the issue (e.g. 'missing_tag')
     this.subtype = attrs.subtype;          // optional - category of the issue within the type (e.g. 'relation_type' under 'missing_tag')
-    this.severity = attrs.severity;        // required - 'warning' or 'error'
+    this.severity = attrs.severity;        // required - 'suggestion' or 'warning' or 'error'
     this.message = attrs.message;          // required - function returning localized string
     this.reference = attrs.reference;      // optional - function(selection) to render reference information
     this.entityIds = attrs.entityIds;      // optional - array of IDs of entities involved in the issue
@@ -63,7 +63,7 @@ export function validationIssue(attrs) {
         var fixes = this.dynamicFixes ? this.dynamicFixes(context) : [];
         var issue = this;
 
-        if (issue.severity === 'warning') {
+        if (issue.severity === 'warning' || issue.severity === 'suggestion') {
             // allow ignoring any issue that's not an error
             fixes.push(new validationIssueFix({
                 title: t.append('issues.fix.ignore_issue.title'),
@@ -87,6 +87,12 @@ export function validationIssue(attrs) {
     };
 
 }
+
+validationIssue.ICONS = {
+    suggestion: '#iD-icon-info',
+    warning: '#iD-icon-alert',
+    error: '#iD-icon-error'
+};
 
 
 export function validationIssueFix(attrs) {
--- a/modules/ui/commit_warnings.js
+++ b/modules/ui/commit_warnings.js
@@ -13,6 +13,9 @@ export function uiCommitWarnings(context) {
             .getIssuesBySeverity({ what: 'edited', where: 'all', includeDisabledRules: true });
 
         for (var severity in issuesBySeverity) {
+            // don't show suggestions on the changeset page
+            if (severity === 'suggestion') continue;
+
             var issues = issuesBySeverity[severity];
 
             if (severity !== 'error') {      // exclude 'fixme' and similar - #8603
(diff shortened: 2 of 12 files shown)


modules/ui/commit_warnings.js
```diff
@@ -34,7 +37,7 @@ export function uiCommitWarnings(context) {
 
             containerEnter
                 .append('h3')
-                .call(severity === 'warning' ? t.append('commit.warnings') : t.append('commit.errors'));
+                .call(t.append(`commit.${severity}s`));
```
tordans, line 40: One advantage of using the https://github.com/openstreetmap/iD/pull/11020/files#diff-677e75df5348a6b4af7e5becaec89f67be3bae43dcf3b2b6a9535bd5781dfa5eR91-R96 in for translations as well is those translations strings are easier to find with global search.

modules/ui/issues_info.js
```diff
@@ -29,7 +29,8 @@ export function uiIssuesInfo(context) {
         var liveIssues = context.validator().getIssues({
             what: prefs('validate-what') || 'edited',
             where: prefs('validate-where') || 'all'
-        });
+        }).filter(issue => issue.severity !== 'suggestion');
```
tordans, line 32: I wonder if we should make the table from https://github.com/openstreetmap/iD/pull/11020#issue-3039567699 an object in code and use that for filters like this one. This would move the documentation in code, which I quite like…

I am thinking about something like…

```
const showIssue = {
 preventChanges: { suggestion: false, warning: false, error: true },
 showOnChangesetPage: { suggestion: false, warning: true, error: true },
 countInChangesetTag: { suggestion: false, warning: true, error: true },
 etc: { suggestion: false, warning: false, error: true },
}
```

So this filter would become something like `.filter(issue => showIssue.showOnChangesetPage[issue.severity])`

modules/core/validation/models.js
```diff
@@ -88,6 +88,12 @@ export function validationIssue(attrs) {
 
 }
 
+validationIssue.ICONS = {
+    suggestion: 'info',
+    warning: 'alert',
+    error: 'error'
```
tordans: Those are always used  like this…
Is there an advantage to have this split up and recomposed below?
It makes it easier to search those strings when they are pre-composed in one place.

```suggestion
    suggestion: '#iD-icon-info',
    warning: '#iD-icon-alert',
    error: '#iD-icon-error'
```

tordans (approved):
Very happy to see this happening, thanks. I added a few inline suggestions.

1ec5:
> I can't tell whether it's intentionally off by default, it actually seems to be a bug

I think it might be related to the outcry in #6332.

tyrasd (approved):

modules/core/validation/models.js
k-yle: yeah that seems helpful, done

tyrasd (approved):

PR #11027, opened 2025-05-06
refresh numeric fields on blur

### LINKED ISSUE #10018: `width=0,1` saved as `width=1` in locales that use a period as the decimal separator, such as English and Chinese
### URL

_No response_

create node

tag `barrier=bollard`

add `width=0,1`

trigger confusion involving several people - see https://community.openstreetmap.org/t/what-width-0-can-mean-here/106666 https://www.openstreetmap.org/changeset/142886665 https://www.openstreetmap.org/note/3994430

Peek 2023-12-03 20-03

tried search for `comma` and `width` in open issues here and tagging schema

2.27.3

Firefox

tyrasd:
addresses #10018: after leaving the input field, the numeric value will be always updated (printed in the user's locale number format). This specifically re-formats any "misplaced" thousands separators, e.g. in the case where a user with English locale typed in `0,1` which results in the tag value to be set to `1`. :arrow_right:  this is now reflected in the UI fields appropriately, and the user can correct their mistake if they had intended to enter `0.1`. Similarly, if a user writes `12,3445` the thousands separator will be updated to the "correct" position (i.e. `123,445`), making the user aware that they might typed a digit too many.

Example behavior:

English locale:

https://github.com/user-attachments/assets/95ec4395-9f15-44fa-921e-96ffd68ea078

German locale (`,` used as decimal separator / `.` used as thousands separator[^1]):

https://github.com/user-attachments/assets/4c1625be-eda7-4d5e-a5ce-43ae9a68064f

[^1]:  note that `.` can also always be used for input in "raw number format", i.e. `<digits>.<digits>` (no thousands separator allowed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,12 +43,14 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
+* Refresh numeric input fields after leaving focus with the value that is stored in the tag ([#11027])
 #### :earth_asia: Localization
 #### :hourglass: Performance
 #### :mortar_board: Walkthrough / Help
 #### :hammer: Development
 
 [#10970]: https://github.com/openstreetmap/iD/pull/10970
+[#11027]: https://github.com/openstreetmap/iD/pull/11027
 
 
 # v2.34.0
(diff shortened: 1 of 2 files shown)


1ec5 (approved):
This looks fine to me at a glance. There are a lot of cases to account for now. If we end up needing to modify this logic further, we should break it out into something that can be unit tested.

PR #11029, opened 2025-05-07
try to make NSI suggestion message less misleading

### LINKED ISSUE #6517: Make issue warning clearer when offering to add `brand:wikidata` or `brand:wikipedia`
Whole point of `brand:wikidata` and `brand:wikipedia` is that name alone (or combined with other indicator like object type and location) is not sufficient to identify object as belonging to a specific brand.

Despite that iD is asking user to add `brand:wikidata` and `brand:wikipedia` and describes this action as "XYZ looks like a brand with nonstandard tags" and offers user to "Upgrade the tags".

There are multiple problems here

- iD is adding new tags, not "upgrading" ones that are present, so description is misleading
- such heuristic usually gives good results, but there are some false positives. iD must mention to user that he is responsible for verifying this and that suggested change may be wrong
- what is actually changed should be displayed in form readable to user
- Even after clicking "i" button this tags are still not linking to Wikipedia/Wikidata making verification harder

See for example http://preview.ideditor.com/master/#background=Bing&disable_features=boundaries&id=w488118123&map=20.29/38.74517/-75.15497

Selection_004

After #6443 it was improved a bit but is still far from OK version.

Note also that

> All features of the same brand should be tagged the same way.

displayed after clicking "i" is wrong - for example shops of the same brand may have different `opening_hours` or other tags.

### LINKED ISSUE #9528: It is not possible for a mapper to understand why an invalid NSI suggestion was made
### URL

https://www.openstreetmap.org/edit?editor=id&node=3196433293#map=20/53.95182/-1.08503

When iD makes an incorrect suggestion (as it often does, due to errors in the name suggestions index) it is not possible for a mapper to understand why the suggestion was made or otherwise verify it.

To reproduce:

Pick a cafe named after one of the most popular surnames in the country (Robinson).  
Edit with iD.
iD says that it "looks like a common feature with nonstandard tags"
Click "i" and this is displayed:

    Some features, for example retail chains or post offices, are expected to have certain tags in common.
    Suggested updates:
    - amenity=cafe
    + amenity=pub
    + brand=Robinsons
    + brand:wikidata=Q7353116
    + brewery=Robinsons
    + operator=Robinsons
    + operator:wikidata=Q7353116
    
In this case this is clearly utter rubbish, and that's yet another issue for the NSI, but what's an issue for here is that there is no way for me to find out _why_ iD is making this suggestion.

What It'd like to be able to do is to identify the culprit who added the invalid suggestion and help them to understand why the mechanical edit that they are trying to make is invalid.  Unfortunately there's simply no way for a user (most of whom won't even know that the NSI is even a thing) to do this

https://www.openstreetmap.org/node/3196433293/history is the object concerned here.

2.25.1

Other (please specify)

### LINKED ISSUE #9529: NSI suggestions are treated as valid when they are often not
### URL

https://www.openstreetmap.org/edit?editor=id&node=3196433293#map=20/53.95187/-1.08486

(related to https://github.com/openstreetmap/iD/issues/9528)

When iD makes an incorrect NSI suggestion, the language iD uses suggests that the NSI suggestion is likely correct (which is often untrue).

To reproduce:

Pick a cafe named after one of the most popular surnames in the country (Robinson).
Edit with iD.
iD says that it "looks like a common feature with nonstandard tags" and offers me the chance to "upgrade the tags"

This is wrong.  Sometimes NSI suggestions could be considered as "upgrades" (adding a wikidata field maybe, or pushing a particular tag schema).  This is not what is being suggested here - instead the suggestion is to change something mapped as one feature to another.

https://www.openstreetmap.org/node/3196433293/history

2.25.1

Other (please specify)

### LINKED ISSUE #9592: Altering the text when an NSI suggestion is displayed.

matkoniecz:
fixes #9529
fixes #6517
fixes #9592
fixes #9528

NSI may be wrong or incomplete which may result in wrong suggestions

some mappers are unaware that QA suggestions in iD are not always correct, sometimes resulting in "upgrading" thousands of objects without real benefit and damaging data

this is an attempt to make people more aware that brain must be engaged while doing QA

disclaimer: I am not a native speaker

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1909,7 +1909,7 @@ en:
       noncanonical_brand:
         message: "{feature} may be matching one of known brands"
         message_incomplete: "{feature} may be matching one of known brands"
-        reference: " Some features, for example retail chains or post offices, are expected to have certain tags in common. This set of suggested tags is based on comparing the name and other existing tags to a database of major chains. If this is merely a coincidence, select “Tag as not the same.”"
+        reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common. This set of suggested tags is based on comparing the name and other existing tags to a database of major chains. If this is merely a coincidence, select “Tag as not the same.”"
     point_as_area:
       message: '{feature} should be a point, not an area'
     point_as_line:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1907,9 +1907,9 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} may or may not benefit from changing this tags"
-        message_incomplete: "{feature} may or may not benefit from adding this tags"
-        reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
+        message: "{feature} may be matching one of known brands"
+        message_incomplete: "{feature} may be matching one of known brands"
+        reference: " Some features, for example retail chains or post offices, are expected to have certain tags in common. This set of suggested tags is based on comparing the name and other existing tags to a database of major chains. If this is merely a coincidence, select “Tag as not the same.”"
     point_as_area:
       message: '{feature} should be a point, not an area'
     point_as_line:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1907,7 +1907,7 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} may or may not benefit from chaning this tags"
+        message: "{feature} may or may not benefit from changing this tags"
         message_incomplete: "{feature} may or may not benefit from adding this tags"
         reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
     point_as_area:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1907,8 +1907,8 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} looks like a common feature with nonstandard tags"
-        message_incomplete: "{feature} looks like a common feature with incomplete tags"
+        message: "{feature} may or may not benefit from chaning this tags"
+        message_incomplete: "{feature} may or may not benefit from adding this tags"
         reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
     point_as_area:
       message: '{feature} should be a point, not an area'
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1907,8 +1907,8 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} may be matching one of known brands"
-        message_incomplete: "{feature} may be matching one of known brands"
+        message: "{feature} was automatically matched to one of templates"
+        message_incomplete: "{feature} was automatically matched to one of templates"
         reference: "This change is automatically guessed based on similarity to other branded features. If this guess is wrong, select “Tag as not the same.”"
     point_as_area:
       message: '{feature} should be a point, not an area'
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1909,7 +1909,7 @@ en:
       noncanonical_brand:
         message: "{feature} may be matching one of known brands"
         message_incomplete: "{feature} may be matching one of known brands"
-        reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common. This set of suggested tags is based on comparing the name and other existing tags to a database of major chains. If this is merely a coincidence, select “Tag as not the same.”"
+        reference: "This change is automatically guessed based on similarity to other branded features. If this guess is wrong, select “Tag as not the same.”"
     point_as_area:
       message: '{feature} should be a point, not an area'
     point_as_line:


matkoniecz:
PR triggered by https://community.openstreetmap.org/t/proposed-automated-edit-removal-of-crossing-markings-yes-tags-introduced-in-undiscussed-automated-edits/129614

tordans:
Can you share a screenshot of this UI element with updated wording (by changing it in the inspector)?

data/core.yaml
```diff
@@ -1907,8 +1907,8 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} looks like a common feature with nonstandard tags"
-        message_incomplete: "{feature} looks like a common feature with incomplete tags"
+        message: "{feature} may or may not benefit from changing this tags"
+        message_incomplete: "{feature} may or may not benefit from adding this tags"
```
1ec5: You have the right idea, but this is pretty confusing to see in a list of warnings. In principle, every warning can be right or wrong – that’s why it isn’t an error – so the title should simply describe what seems to be the problem. What the user really needs is a hint as to what might make the warning ignorable. For example:

> **{feature} almost matches the standard tags for a well-known {type} brand**
> Some features, for example retail chains or post offices, are expected to have certain tags in common. This set of suggested tags is based on comparing the name and other existing tags to a database of major chains. If this is merely a coincidence, select “Tag as not the same.”

Filling in `{type}` would require some code changes, but I suspect it would address most of the cases of changing the feature to a very wrong type, or at least reduce plausible deniability.

matkoniecz: > this is pretty confusing to see in a list of warnings

I think that user currently are too sure and being mislead, so them being confused would be improvement :)

> In principle, every warning can be right or wrong – that’s why it isn’t an error

many mappers are clearly unaware of this

> What the user really needs is a hint as to what might make the warning ignorable. For example:

but message that would make users less sure AND less confused would be even better

I will try to implement this one, but if anyone is interested in doing so: feel free to make own PR!

Or maybe I will stuff all of that text into the title?

matkoniecz:
> Can you share a screenshot of this UI element with updated wording (by changing it in the inspector)?

https://www.openstreetmap.org/way/248378684/history#map=19/51.553051/-0.286428

before:
screen-2025-05-07-06-59-52

after, @1ec5 suggestion applied:
screen-2025-05-07-07-01-43

note: I did no actual testing

data/core.yaml
1ec5: > I think that user currently are too sure and being mislead, so them being confused would be improvement :)

If you can special-case this title to only appear to overconfident users, then sure, but not every user is so confident. “May or may not” voids the message of any meaning. To an English speaker at least, it reads as legalese. If the title is misleading, then let’s make it more descriptive and direct, not less. To a mapper who doesn’t understand raw tags, what triggered the warning and why could it be a problem potentially?

> Or maybe I will stuff all of that text into the title?

Hedging is a good idea, but the place to do it is in the explanation. Rest assured, the user will see it by default when selecting the feature, but it won’t appear in places like the Issues sidebar panel or Save screen, where the user doesn’t have an opportunity to act on the warning anyways.

matkoniecz: https://github.com/openstreetmap/iD/pull/11029#issuecomment-2857055954 now has image how it is supposed to look like

k-yle:
Now that we have #11020, it might be worth considering if these "warnings" should be downgraded to "suggestions". I've also seen many people [waste their time](https://osm.org/changeset/160937101) adding tags like `operator:type=government` to every park, because the changeset page made it seem like a serious problem that they need to fix, even if they didn't edit that feature.

image

data/core.yaml
```diff
@@ -1907,9 +1907,9 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} looks like a common feature with nonstandard tags"
-        message_incomplete: "{feature} looks like a common feature with incomplete tags"
-        reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
+        message: "{feature} may be matching one of known brands"
+        message_incomplete: "{feature} may be matching one of known brands"
```
1ec5: These two strings exist to distinguish two cases:

* The suggestion changes or deletes a tag, so it needs more scrutiny
* The suggestion is purely additive, so it’s presumably less risky

If we no longer need this distinction, we can simplify the following code:

https://github.com/openstreetmap/iD/blob/5c721808c61e39d488a2a44ffb4a0f29b9b32c8d/modules/validations/outdated_tags.js#L159-L161

Then again, maybe we do still need to make some suggestions look riskier than others?

1ec5: > Now that we have #11020, it might be worth considering if these "warnings" should be downgraded to "suggestions". I've also seen many people [waste their time](https://osm.org/changeset/160937101) adding tags like `operator:type=government` to every park, because the changeset page made it seem like a serious problem that they need to fix, even if they didn't edit that feature.

Yes, I think the `message_incomplete` case might be a good candidate for this.

data/core.yaml
1ec5: This is the best I can do without modifying how these messages are invoked inside outdated_tags.js:

```yaml
        message: "{feature} has tags inconsistent with a known entity"
        message_incomplete: "{feature} lacks some expected tags for a known entity"
        reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common. The tags on this feature mostly resemble those of an entry in the Name Suggestion Index, except for some differences. If this is merely a coincidence, select “Tag as not the same.”"
```

I don’t feel great about using the word “entity”, but “brand” is too narrow, since NSI also covers operators, transit systems, and flags.

matkoniecz:
> Now that we have https://github.com/openstreetmap/iD/pull/11020, it might be worth considering if these "warnings" should be downgraded to "suggestions".

this also makes sense but I would consider it as a separate improvement (also, where you change that?)

oh right, #11020 is not even merged yet - maybe it can be part of #11020 ?

data/core.yaml
1ec5: On second thought, the `message_incomplete` case may be less critical, but the `message` case is the one we’re less confident about. Giving `message` a higher severity level than `message_incomplete` could send the wrong message about our confidence in the warning.

data/core.yaml
matkoniecz: "inconsistent with" and "expected tags" seem to me to be too encouraging and not clear that suggestion may be utterly wrong

k-yle:
> also, where you change that?

https://github.com/openstreetmap/iD/blob/8f9514c546cad692584eb50f905493f753a7f211/modules/validations/outdated_tags.js#L152

this line would become `suggestion` instead of `warning`

> maybe it can be part of #11020

i was hoping to keep the (possibly controversial) changes seperate from the technical implementation. but if there's no opposition i can add it to #11020

matkoniecz:
> i was hoping to keep the (possibly controversial) changes seperate from the technical implementation

probably a good idea to avoid quagmire blocking merge

data/core.yaml
matkoniecz: > If we no longer need this distinction, we can simplify the following code:

I would be fine with such change, I would be fine with distinct messages of people have a good idea.

data/core.yaml
1ec5: Yeah, I’m trying to think of a good term for something that looks suspiciously similar except for some differences that we the software are unsure about. Like an [uncanny valley](https://en.wikipedia.org/wiki/Uncanny_valley), but I don’t think we would get away with calling it that for this user base.

The point is not to hold up the “known entity” as a gold standard, but as something worth investigating, without prematurely casting judgment. If I were writing something other than a warning message, like a changeset comment or StreetComplete question, I would be asking something like:

> Does this {feature} belong to the {name} {type} chain?

So an equivalent set of warnings would be something like:

> {feature} is suspected to belong to a chain based on its tags
> {feature} has most but not all of a chain’s usual tags

data/core.yaml
tyrasd: What about using:

```suggestion
        message: "{feature} may be matching a known brand with nonstandard tags"
        message_incomplete: "{feature} may be matching a known brand with incomplete tags"
```

matkoniecz: "may be matching a known brand with nonstandard tags" sounds like brand has nonstandard tags

michaelblyons: Alternative formulations for `message`:

- `feature` may be matching a known brand, but with nonstandard tags
- `feature` may be using nonstandard tags on a known brand
- `feature` may be using nonstandard tags on a matching known brand

(No pressure; just an attempt to remediate Mat's last comment.)

k-yle:
#11020 is now merged. @matkoniecz do you want to use that new UI for this PR?

matkoniecz:
I would prefer to decouple text tweaking from level of severity tweaking. Also I would need to track down where it is defined.

matkoniecz:
on looking at it again:

`        message: "{feature} looks like a common feature with nonstandard tags"`
`        message: "{feature} may be matching one of known brands"`

this is effectively the same, this change makes no sense at all. To make sense at all some more serious change is needed, I will consider rolling back or changing it again.

matkoniecz:
note: strength of pushing NSI upgrades will be reduced with https://github.com/openstreetmap/iD/pull/11030/changes

matkoniecz:
one more attempt to both communicate what is going on, do not write a book that will not be read and to discourage blind "upgrades"

data/core.yaml
```diff
@@ -1921,9 +1907,9 @@ en:
         message: "{feature} has incomplete tags"
         reference: "Some features should have additional tags."
       noncanonical_brand:
-        message: "{feature} looks like a common feature with nonstandard tags"
-        message_incomplete: "{feature} looks like a common feature with incomplete tags"
-        reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."
+        message: "{feature} was automatically matched to one of templates"
+        message_incomplete: "{feature} was automatically matched to one of templates"
```
michaelblyons, line 1911: The English doesn't make sense. How about either:
...matched to a template
OR
...matched to one of our templates

data/core.yaml
```diff
@@ -1907,9 +1907,9 @@ en:
...
```
1ec5, line 1911: The passive voice might be complicating things in English. If this ends up being a blue suggestion box instead of a warning, I guess we’d want the message to be positive, not negative, but if we make it too non-urgent, the user will wonder why we’re even bothering them about it.

I think we should strive to reuse existing terminology instead of technical terms that the user has never encountered before.

What if:

```suggestion
        message: "{feature}’s tags almost match a common feature"
        message_incomplete: "{feature} matches a common feature except for some extra tags"
```

But to reiterate https://github.com/openstreetmap/iD/pull/11029#discussion_r2076715721, I really think a better message would mention the feature type that got matched, so any dissonance would be more obvious. That would be a matter of cloning `currEntity`, applying the suggested tags to it, and calling `utilDisplayLabel`.

https://github.com/openstreetmap/iD/blob/92667076b7cddea9fdafd93b45f13dba8cf1479e/modules/validations/outdated_tags.js#L157-L161

Or just getting the matching preset:

https://github.com/openstreetmap/iD/blob/92667076b7cddea9fdafd93b45f13dba8cf1479e/modules/util/utilDisplayLabel.js#L20-L22

matkoniecz:
trying to fish for suggestions in https://github.com/openstreetmap/iD/issues/9528#issuecomment-4049562907

matkoniecz:
It got stuck in annoying way, with feedback and variants scattered across multiple places.
So I am preparing summary of ideas and feedback so far.
Extra comments and ideas are welcome, especially if I mangled something in summary.
I plan to print it out and review on the next opportunity (when in bus/queue etc)

======================================================

NSI may be wrong or incomplete which may result in wrong suggestions

some mappers are unaware that QA suggestions in iD are not always correct, sometimes resulting in "upgrading" thousands of objects without real benefit and damaging data

this is an attempt to make people more aware that brain must be engaged while doing QA

message is supposed to be
- clear and as simple as possible
- be of a reasonable length
- make clear that it is suggested by an automated fallible process and message may be wrong
- discourage blind "upgrading"
- cannot mention brands as sadly NSI is not limited to brands

------------------------------


the current version that I am trying to change:

>message: "{feature} looks like a common feature with nonstandard tags"
>message_incomplete: "{feature} looks like a common feature with incomplete tags"
>reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common."

------------------------------

initial version ( https://github.com/openstreetmap/iD/pull/11029/changes/be6175c5aafcaa47921b6399ef05e4eaa88e05ac )
message: "{feature} may or may not benefit from changing this tags"

[1ec5: (...) this is pretty confusing to see in a list of warnings (...)](https://github.com/openstreetmap/iD/pull/11029/changes/be6175c5aafcaa47921b6399ef05e4eaa88e05ac#r2076715721)

------------------------------

1ec5 suggestion ( https://github.com/openstreetmap/iD/pull/11029/changes/be6175c5aafcaa47921b6399ef05e4eaa88e05ac#r2076715721 )

```
message: "{feature} has tags inconsistent with a known entity"
message_incomplete: "{feature} lacks some expected tags for a known entity"
reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common. The tags on this feature mostly resemble those of an entry in the Name Suggestion Index, except for some differences. If this is merely a coincidence, select “Tag as not the same.”"
```

matkoniecz: "inconsistent with" and "expected tags" seem to me to be too encouraging and not clear that suggestion may be utterly wrong

[SomeoneElseOSM](https://github.com/openstreetmap/iD/issues/9528#issuecomment-4049562907): 

I'm certainly not convinced by #11029 (comment) :)

In the example "Chinese Restaurant Young's looks like a common feature with nonstandard tags" is rubbish for a couple of reasons:

    it shouldn't be changing pubs into restaurants. There is crossover, but the mapper should decide.
    the words "common" and "nonstandard" are both loaded (as is "upgrade" in the other message)

The low-hanging fruit is "upgrade" - instead of "upgrade the tags" say "change the tags as suggested below".

Next, perhaps also change "a common feature with nonstandard tags" to "it might be a member of a chain that already exists in OSM". The challenge here is that that change has to be verifiable by the mapper - they have to be able to click something to see (in this example) what the chain "Young's" actually is, who owns it and where it is. None of that is currently possible within iD.

    re: https://github.com/openstreetmap/iD/pull/11030/changes was applied some time ago (not released yet)

    rere: "Accept the suggested updates" still suggests that the "new" version is better than the old version, which most of the time it won't be. I used "change" because that's neutral and doesn't imply a "best thing to do".

------------------------------

1ec5 alternative suggestion ( [Does this {feature} belong to the {name} {type} chain?](https://github.com/openstreetmap/iD/pull/11029/changes/be6175c5aafcaa47921b6399ef05e4eaa88e05ac#r2078065182) )

```
{feature} is suspected to belong to a chain based on its tags
{feature} has most but not all of a chain’s usual tags
```

-----------------------------

[my second version](https://github.com/openstreetmap/iD/pull/11029/changes/12d5e0436cf303dce5518b8c215bf7b152e6df0f)

```
message: "{feature} may be matching one of known brands"
message_incomplete: "{feature} may be matching one of known brands"
reference: "Some features, for example retail chains or post offices, are expected to have certain tags in common. This set of suggested tags is based on comparing the name and other existing tags to a database of major chains. If this is merely a coincidence, select “Tag as not the same.”"
```

my problem with this text: may be still too enthusiastic and too pushing

reference may be unreasonably long

-----------------------------

[tyrasd](https://github.com/openstreetmap/iD/pull/11029/changes/12d5e0436cf303dce5518b8c215bf7b152e6df0f#r2137276153)

```
message: "{feature} may be matching a known brand with nonstandard tags"
message_incomplete: "{feature} may be matching a known brand with incomplete tags"
```

my problem with this text: "may be matching a known brand with nonstandard tags" sounds like brand has nonstandard tags

-----------------------------

[michaelblyons](https://github.com/openstreetmap/iD/pull/11029/changes/12d5e0436cf303dce5518b8c215bf7b152e6df0f#r2303825685)

```
feature may be matching a known brand, but with nonstandard tags
feature may be using nonstandard tags on a known brand
feature may be using nonstandard tags on a matching known brand
```

"(No pressure; just an attempt to remediate Mat's last comment.)"

-----------------------------

[my next version](https://github.com/openstreetmap/iD/pull/11029/changes/371bb50718961ee524bed4cb63ec6c766f357e29)

intention was to make more clear that it is outcome of an automated process

```
message: "{feature} was automatically matched to one of templates"
message_incomplete: "{feature} was automatically matched to one of templates"
```

michaelblyons:
The English doesn't make sense. How about either:
...matched to a template
OR
...matched to one of our templates


1ec5:
The passive voice might be complicating things in English. If this ends up being a blue suggestion box instead of a warning, I guess we’d want the message to be positive, not negative, but if we make it too non-urgent, the user will wonder why we’re even bothering them about it.

I think we should strive to reuse existing terminology instead of technical terms that the user has never encountered before.

What if:

```
message: "{feature}’s tags almost match a common feature"
message_incomplete: "{feature} matches a common feature except for some extra tags"
```

But to reiterate #11029 (comment), I really think a better message would mention the feature type that got matched, so any dissonance would be more obvious. That would be a matter of cloning currEntity, applying the suggested tags to it, and calling utilDisplayLabel.

PR #11030, opened 2025-05-07
do not suggest that QA suggestion is always improvement

matkoniecz:
in many cases users thought that all suggestions are improvement

this change attempts to lessen this false impression

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2066,8 +2066,8 @@ en:
       tag_this_as_lower:
         title: Tag this as lower
       upgrade_tags:
-        title: Upgrade the tags
-        annotation: Upgraded old tags.
+        title: Change the tags
+        annotation: Changed old tags.
       use_different_layers:
         title: Use different layers
       use_different_layers_or_levels:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2067,7 +2067,7 @@ en:
         title: Tag this as lower
       upgrade_tags:
         title: Change the tags
-        annotation: Changed old tags.
+        annotation: Accepted the suggested updates
       use_different_layers:
         title: Use different layers
       use_different_layers_or_levels:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2066,7 +2066,7 @@ en:
       tag_this_as_lower:
         title: Tag this as lower
       upgrade_tags:
-        title: Change the tags
+        title: Accept the suggested updates
         annotation: Accepted the suggested updates
       use_different_layers:
         title: Use different layers


matkoniecz:
PR triggered by https://community.openstreetmap.org/t/proposed-automated-edit-removal-of-crossing-markings-yes-tags-introduced-in-undiscussed-automated-edits/129614 and https://github.com/openstreetmap/iD/pull/11029

data/core.yaml
```diff
@@ -2066,8 +2066,8 @@ en:
       tag_this_as_lower:
         title: Tag this as lower
       upgrade_tags:
-        title: Upgrade the tags
-        annotation: Upgraded old tags.
+        title: Change the tags
+        annotation: Changed old tags.
```
1ec5: There’s already another action with a similar title:

https://github.com/openstreetmap/iD/blob/7d6dda002b53efcf3574f49ea9a37eff7800c6d1/data/core.yaml#L86-L87

Instead of “upgrade the tags”, how about “accept the suggested tags”, since that’s what the diff below calls them anyways?

matkoniecz: > There’s already another action with a similar title:

do you know here it is used?

> accept the suggested tags

I would refer to go further away from implying it is a good change, but this is also improvement compared to the current state

1ec5: `change_tags` is when you manually change a tag. So if you twiddle one of the fields, the Undo button’s tooltip will say “Undo: Changed tags.”

The other potential source of confusion is that we’d wind up with two options for changing the tags. The second option indicates the effect of the change, but the first does not:

Change the tags. Tag as not the same “Volvo”.

“Complete the tags” would match the warning’s current title, while “Accept the suggested updates” would match the line “Suggested updates” below. But if you find that too optimistic, then we’d need to come up with a replacement for that string too. Personally, I don’t find it too leading, as long as the problem statement in #11029 maintains transparency with the user.

SomeoneElseOSM:
In my experience iD's suggestions often make things worse (often when a previous tagging was not quite right for a real-life feature, but iD's "suggestions" are far worse).

"changed" is certainly better than "upgraded" - my personal suggestion for wording would likely be much ruder and unlikely to be accepted here :)

data/core.yaml
tyrasd: > Accept the suggested updates

Yeah, that sounds like an eloquent solution for this label. :+1:

data/core.yaml
```diff
@@ -2066,8 +2066,8 @@ en:
       tag_this_as_lower:
         title: Tag this as lower
       upgrade_tags:
-        title: Upgrade the tags
-        annotation: Upgraded old tags.
+        title: Change the tags
```
matkoniecz:
```suggestion
        title: Accept the suggested updates
```

data/core.yaml
matkoniecz:
```suggestion
        annotation: Accepted the suggested updates
```

matkoniecz: changed to this one then

note: I think it is still not making clear that this automated changes may be damaging, but better than what we have now

note: I am not native speaker, I cannot judge is it Engrish or English

tyrasd (approved):
Looks good, but let's merge it after the upcoming release. So that translators have time to localize it.

matkoniecz:
> In my experience iD's suggestions often make things worse (often when a previous tagging was not quite right for a real-life feature, but iD's "suggestions" are far worse).

Are they coming from NSI suggestions or from https://github.com/openstreetmap/id-tagging-schema/blob/main/data/deprecated.json or from somewhere else?

(not sure are there issues opened already about these problems)

PR #11052, opened 2025-05-15
Route relation labeling (again)

### LINKED ISSUE #8759: Omit disambiguator from relation type in member/membership editor

tyrasd:
A slightly different approach for #8759 (and a bit beyond) to fix the issue with overly long NSI preset names for PT route relations. Supersedes  #10583[^1].

* Uses the regular non-NSI base preset for the labeling in the relation membership (and relation members) editors
* Shows the icon of the NSI preset next to the base preset name
* Shows the route's `ref` tag in a (colored[^2]) "shield" 

Some before/after examples:

 - 

<hr>

 - 

<hr>

 - 

<hr>

 - 

<hr>

[^1]: Different to the approach in #10583, this is not relying on implicit assumptions about NSI's naming scheme and does not interfere with (translated) names of regular presets.
[^2]: If applicable, see also #9424


modules/ui/feature_list.js
```diff
@@ -332,12 +333,26 @@ export function uiFeatureList(context) {
                 .attr('class', 'entity-type')
                 .text(function(d) { return d.type; });
 
+            label.each(function(d) {
+                if (d.entity?.type !== 'relation') return;
+                const hasColor = d.entity.tags.colour && isColourValid(d.entity.tags.colour);
+                const hasRef = d.entity.tags.ref;
+                if (hasColor || hasRef) {
+                    const color = isColourValid(d.entity.tags.colour) ? d.entity.tags.colour : '#555';
```
1ec5: Many recreational routes are associated with multiple colors, which could be tagged as a semicolon-delimited list in `colour=*` (regardless of any `osmc:symbol=*`).

modules/ui/feature_list.js
```diff
@@ -332,12 +333,26 @@ export function uiFeatureList(context) {
                 .attr('class', 'entity-type')
                 .text(function(d) { return d.type; });
 
+            label.each(function(d) {
+                if (d.entity?.type !== 'relation') return;
+                const hasColor = d.entity.tags.colour && isColourValid(d.entity.tags.colour);
```
1ec5: Recently on the forum we were [discussing alternative keys](https://community.openstreetmap.org/t/define-check-and-automated-correct-string-values-for-the-tag-colour-for-buildings/129467/10) that often associate routes with colors. `colour=*` is usually appropriate when a route is associated with a color. However, `ref:colour=*` is used in some places, particularly Spain, where they don’t attempt to uniquely color-code the routes in a network, but the color is still very important for recognizability. If iD uses the color as a background color for something that looks like a route marker, then this would be a good opportunity to use this key. Unfortunately, id-tagging-schema doesn’t have a `ref:colour=*` field yet. It’s a touchy subject, because people who don’t understand the Spanish system or others like it tend to rush to judgment about how they’re doing it wrong.

modules/ui/feature_list.js
tyrasd: Yeah, I was pondering about `osmc:symbol` for hiking routes (but only for a future iteration on top of this feature). Multi-value `colour` tags seem to exist ([example](https://www.openstreetmap.org/relation/7840712)), but are not very frequent outside of the Netherlands:

[!](https://overpass-turbo.eu/s/24ld)

css/80_app.css
```diff
@@ -2894,21 +2881,35 @@ img.tag-reference-wiki-image {
...
+.section-raw-membership-editor .member-row img.member-entity-icon,
+.section-raw-member-editor .member-row img.member-entity-icon {
+    display: inline-block;
+    width: auto;
+    height: 18px;
+    margin-left: 5px;
+    margin-bottom: -4.5px;
+}
+.section-raw-membership-editor .member-row span.member-entity-ref-color,
+.section-raw-member-editor .member-row span.member-entity-ref-color,
```
matkoniecz, line 2925: section-raw-member-editor?

where it has effects?

modules/osm/tags.js
```diff
@@ -339,6 +341,22 @@ export function isColourValid(value) {
     return true;
 }
 
+export function getRelationColor(tags, fallback) {
+    let color, textColor;
+    if (tags['ref:colour'])       color = tags['ref:colour'];
+    else if (tags.colour)         color = tags.colour;
+    const isValid = isColorValid(color);
+    if (!isValid)                 color = fallback;
+    if (tags['ref:colour_tx'])    textColor = tags['ref:colour_tx'];
```
matkoniecz, line 367: note that relation with `colour=red` `ref:colour_tx=red` will result in unreadable text

is it OSM data bug?

I see no doc at https://taginfo.openstreetmap.org/keys/ref%3Acolour_tx#wiki

I see mention at https://wiki.openstreetmap.org/wiki/Key:ref:colour which suggests it would be bug, so I guess that using it as is is fine?

css/80_app.css
1ec5: The member editor is what appears at the bottom of the inspector when a relation is selected. The membership editor is what appears at the bottom of the inspector when a relation’s member is selected.

modules/ui/feature_list.js
matkoniecz: > It’s a touchy subject, because people who don’t understand the Spanish system or others like it tend to rush to judgment about how they’re doing it wrong.

I have not rushed to yell "you are doing it wrong way" but I definitely had such thought. Is there write-up somewhere which explains why `ref:colour` rather than `colour` makes sense?

(if it makes sense it would make sense to support it, including iD presets)

css/80_app.css
matkoniecz: so `raw-membership-editor` is for membership editor?

There is no special raw membership editor?

modules/ui/feature_list.js
1ec5: [The `ref:colour=*` documentation](https://wiki.openstreetmap.org/wiki/Key:ref:colour#Disambiguation) alludes to the situation. It’s primarily important in Spain, where colors on route shields apparently have political overtones. I’ve had someone explain the considerations there to me before, but it goes over my head. From my perspective as an OSM Americana developer, some of the distinctions seem like they could be represented by more specific `network=*` values instead, but other route colors seem to have no rhyme or reason to them, which would justify the presentational keys. Might be worth asking on the forum for a fuller explanation.

matkoniecz: I am mostly confused why not use regular `colour` - I ma not asking why it gets tagged but why use different key than normal?

css/80_app.css
1ec5: Not sure what you’re asking, but the member and membership editors appear in the same location in the UI, in mutually exclusive situations. It does make sense to style them both consistently, as `ref=*` and `network=*` can both appear on relations and relation members, depending on the feature type.

matkoniecz: > but the member and membership editors appear in the same location in the UI, in mutually exclusive situations

it does not seem mutually exclusive?

what about relations that are having members (so member editor is appearing) and are also part of other relations (route master one for example) so membership editor is appearing.

In fact membership editor will appears always, I think to enable adding relation into another relation as member.

(I am not nitpicking for sake of nitpicking, I am trying to confirm that I understood what you wanted to express)

1ec5: > what about relations that are having members (so member editor is appearing) and are also part of other relations (route master one for example) so membership editor is appearing.

Yes, you would see both in that case.

modules/osm/tags.js
matkoniecz: what about

```
ref=REF
colour=#00ffff
ref:colour=orange
ref:colour_bg=red
ref:colour_tx=orange
route=bicycle
type=route
```

should it appear orange on orange?

See also https://www.openstreetmap.org/relation/15627027 and https://www.openstreetmap.org/note/5180615

modules/ui/feature_list.js
1ec5: [This highway route](https://www.openstreetmap.org/relation/14047150) is `colour=red` because it’s known as the Red Route; one would identify it by color alone. [This subway route](https://www.openstreetmap.org/relation/2851728) is tagged (more or less) `colour=green` because it’s known as the Green Line and the branding is green through and through.

By contrast, [AP-1](https://www.openstreetmap.org/relation/404438) has (more or less) `ref:colour=blue` because the _route markers_ are blue. That’s the only aspect of the route that’s blue. Normally we don’t tag this kind of detail, because all the routes in a given network share the same route marker style, but Spanish routes are more idiosyncratic.

matkoniecz:
I asked on https://t.me/OSMes/175014 for feedback

Not sure is it necessary to supporti `ref:colour_bg` somehow.

Robot8A:
It's something the Spanish community came with for road shields, but I see how it could be used beyond it too.

For example in Zaragoza, Spain, each public transport line has a specific visual representation:
https://detalier.com/portfolio/red-de-autobuses-de-zaragoza/ (this is the website of the designer studio who did the visual representation, as the official website of the bus company is inaccesible from a non-Spanish IP, but it can be seen there too)

Day lines are shown with a given background colour, with white text on stops. Night lines instead are dark blue background and a given text colour.

The question is whether that is interesting to show in iD or not.

And also probably how if we have so much complexity with osmc:symbol, why couldn't we have it as well for ref colour.

tyrasd:
> Not sure is it necessary to supporti `ref:colour_bg` somehow.

I would say not, because [`ref:colour_bg`](https://wiki.openstreetmap.org/wiki/Key:ref:colour) seems to be intended for the _background colour of the road sign_, which we don't care about here.

modules/osm/tags.js
tyrasd: > is it OSM data bug?

yes, I would say so

hlfan:
> fix the issue with overly long NSI preset names for PT route relations

⚖️

matkoniecz:
it is kind of fitting for trains to have train-shaped icons...

well, I guess that solution would be to have limit on how long icon is allowed to be? Have you created already an issue for that?

1ec5:
Long wordmark icons are precisely why name-suggestion-index has historically preferred Twitter and Facebook icons (which have to be square) over Wikimedia Commons logos. Does ÖBB NightJet have a [small logo or icon <small>(P8972)</small>](https://www.wikidata.org/wiki/Property:P8972) that could be added to [this Wikidata item](https://www.wikidata.org/wiki/Q28127371)? I believe NSI already prefers small logos over full logos.

tordans:
Site note: we are putting too much stuff onto this one line...

Robot8A:
> Long wordmark icons are precisely why name-suggestion-index has historically preferred Twitter and Facebook icons (which have to be square) over Wikimedia Commons logos. Does ÖBB NightJet have a [small logo or icon (P8972)](https://www.wikidata.org/wiki/Property:P8972) that could be added to [this Wikidata item](https://www.wikidata.org/wiki/Q28127371)? I believe NSI already prefers small logos over full logos.

Probably a version with just the letters NJ, which are used when abbreviating the train service (see https://iechub.rfi.it/ArriviPartenze/ArrivalsDepartures/Monitor?Arrivals=False&Search=tarvisio&PlaceId=3443 during night time as an example)

Edit: I wrote my message without realizing @hlfan actually already solved this https://www.wikidata.org/w/index.php?title=Q28127371&diff=prev&oldid=2469124147

PR #11082, opened 2025-06-02
Improve multilingual name fallback logic based on user locale

### LINKED ISSUE #10219: Feature labels are based on generic name:zh tag when using Traditional or Simplified Chinese locale

### LINKED ISSUE #11315: Support zh-Hant and zh-Hans

daishu0000:
## Introduction

This PR enhances the multilingual support for feature names in the map by allowing multiple fallback languages to be used when displaying `name:*` tags. This improves localization and ensures that users see names in their preferred or most appropriate language.

## Main Changes

- **Multiple language fallbacks for names:**  
Instead of checking only a single name:<lang> tag, the logic now iterates through a prioritized list of language codes based on the user's locale preferences. If no localized tag is found in any of the preferred languages, it falls back first to name:en, and if that is also not available, finally falls back to the general name tag.

- **Special handling for Chinese variants:**  
  Added remapping logic to better support Chinese language variants by mapping locale codes to appropriate language tags.

### DIFF
--- a/modules/core/localizer.js
+++ b/modules/core/localizer.js
@@ -6,6 +6,19 @@ import { utilStringQs } from '../util';
 import { utilArrayUniq } from '../util/array';
 import { presetsCdnUrl } from '../../config/id.js';
 
+const rematchCodes = {
+    // below is not needed, because it can be solved by split('-')[0].toLowerCase()
+    // 'ar-AA': 'ar',
+    // 'pt-BR': 'pt',
+    // 'pt': 'pt-PT',
+
+
+    'zh-CN': ['zh-Hans','zh'],
+    'zh-TW': ['zh-Hant','zh'],
+    'zh-HK': ['zh-Hant','zh'],
+    'ko-Hani': ['ko-Hani'],
+};
+
 let _mainLocalizer = coreLocalizer(); // singleton
 let _t = _mainLocalizer.t;
 
@@ -48,6 +61,7 @@ export function coreLocalizer() {
     // `_localeCodes` must contain `_localeCode` first, optionally followed by fallbacks
     let _localeCodes = ['en-US', 'en'];
     let _languageCode = 'en';
+    let _languageCodes= [];
     let _textDirection = 'ltr';
     let _usesMetric = false;
     let _languageNames = {};
@@ -57,6 +71,7 @@ export function coreLocalizer() {
     localizer.localeCode = () => _localeCode;
     localizer.localeCodes = () => _localeCodes;
     localizer.languageCode = () => _languageCode;
+    localizer.languageCodes = () => _languageCodes;
     localizer.textDirection = () => _textDirection;
     localizer.usesMetric = () => _usesMetric;
     localizer.languageNames = () => _languageNames;
@@ -169,6 +184,20 @@ export function coreLocalizer() {
 
     function updateForCurrentLocale() {
         if (!_localeCode) return;
+
+        for (let localeCode of _localeCodes) {
+            if (localeCode in rematchCodes) {
+                let languageCodeList = rematchCodes[localeCode];
+                for (let code of languageCodeList) {
+                    _languageCodes.push(code);
+                }
+            } else {
+                localeCode = localeCode.split('-')[0].toLowerCase();
+                _languageCodes.push(localeCode);
+            }
+        }
+
+        _languageCodes.push("en")
 
         _languageCode = _localeCode.split('-')[0];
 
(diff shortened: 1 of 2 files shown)


daishu0000:
Hi @tyrasd, Could you please check my PR? I think it's an important PR for the users that want to see names of different languages in ID editor. And it also can solve the problem of showing traditional and simplifiled Chinese. Thank you very much!

tordans:
@daishu0000 did you follow the conversation at ~https://github.com/openstreetmap/iD/pull/11082~  https://github.com/openstreetmap/iD/pull/10911 ?
It looks like a solution is within reach.

daishu0000:
> @daishu0000 did you follow the conversation at #11082 ? It looks like a solution is within reach.

but this PR is #11082 :joy:

tordans:
> > @daishu0000 did you follow the conversation at #11082 ? It looks like a solution is within reach.
> 
> but this PR is #11082 😂

Sorry, https://github.com/openstreetmap/iD/pull/10911 :-)

modules/core/localizer.js
```diff
@@ -6,6 +6,13 @@ import { utilStringQs } from '../util';
 import { utilArrayUniq } from '../util/array';
 import { presetsCdnUrl } from '../../config/id.js';
 
+const rematchCodes = {
+    'zh-CN': ['zh-Hans','zh'],
+    'zh-TW': ['zh-Hant','zh'],
+    'zh-HK': ['zh-Hant','zh'],
+    'ko-Hani': ['ko-Hani'],
```
tyrasd, line 13: What's the purpose of rematching `ko-Hani` to `ko-Hani` again?

modules/core/localizer.js
```diff
@@ -170,6 +179,20 @@ export function coreLocalizer() {
     function updateForCurrentLocale() {
         if (!_localeCode) return;
 
+        for (let localeCode of _localeCodes) {
+            if (localeCode in rematchCodes) {
+                let languageCodeList = rematchCodes[localeCode];
```
tyrasd, line 184: this should be a `const` variable

```suggestion
                const languageCodeList = rematchCodes[localeCode];
```

similarly, also the two `for (const … of …)` loops

tyrasd (commented):
This look good in general. See one question below and a minor code style improvement suggestion.

modules/core/localizer.js
daishu0000: if not do that, it will get a result of ko, because we only use the text before "-" by default
```
      } else {
          localeCode = localeCode.split('-')[0].toLowerCase();
          _languageCodes.push(localeCode);
      }
```

tyrasd: Maybe I'm overlooking something, but in this case, `_languageCodes` will still include the original `ko-Hani`, as the source code you mentioned just adds additional fallback locales to the existing list of `_languageCodes`. So my question is more precisely: Is there a reason we don't want to include `ko` as a fallback code if a user has `ko-Hani` specified?

Is it because in this case we cannot assume that the user is able to read Hangul script if they only specify `ko-Hani` as their user locale? If that's the reasoning behind it, I think we should generalize it to any locale that is a _language-script_ pair. E.g. `sr-Latn` is another prominent example.

matkoniecz: ping @daishu0000

modules/core/localizer.js
matkoniecz: ping @daishu0000

1ec5:
This approach would be obviated by #10911.

modules/core/localizer.js
daishu0000: > ping @daishu0000

I think #10911 #11731 Maybe the better solution than me

Issue #11079, opened 2025-05-30
Bidi tags


NeatNit:
### URL

https://master.apis.dev.openstreetmap.org/edit?node=4359512471

### How to reproduce the issue?

A live example: https://www.openstreetmap.org/edit?node=3645671154

Some tags have bidirectional text in values. The direction of the rendered text is determined by the first strong character in it: https://www.unicode.org/reports/tr9/#The_Paragraph_Level (P2)

In HTML this is achieved by adding `dir="auto"` to the element containing the text. This is why it's rendered correctly here: https://master.apis.dev.openstreetmap.org/node/4359512471

![Image](https://github.com/user-attachments/assets/717acf08-2f30-4a69-bc27-222e3e488fd7)

But in iD they are all rendered in LTR - note especially the value of "he-en" and compare with above:

![Image](https://github.com/user-attachments/assets/bbaae7a1-f3f5-4147-a36a-9b7917ecc456)

If you change your OSM preferences to Hebrew or Arabic, they are all rendered RTL instead, which is also wrong:

![Image](https://github.com/user-attachments/assets/034d27be-4f6e-456f-aeff-81728900ccaa)

By adding `dir="auto"` to all input elements, the desired result is achieved (but may need some tweaking in CSS):

![Image](https://github.com/user-attachments/assets/c6bab322-97e1-42d1-a1c8-00a52c4bc86d)

This was achieved by inspecting the inputs and then running this in console: `for (let x of document.getElementsByTagName("input")) {x.setAttribute("dir", "auto");}` (just as a quick hack)

1ec5:
The fields in the Fields section should definitely respect the predominant writing direction via `dir="auto"`. I’m less sure about the raw tag editor in the Tags section, since the point is to treat every value as code. Could initial or final punctuation cause additional confusion, especially if the user switches to text view, where every line is in the form `k=v` or `k=*`?

#9037 tracks more or less this same issue, but it seems to be focused on the raw tag editor, probably an oversight.

NeatNit:
> I’m less sure about the raw tag editor in the Tags section, since the point is to treat every value as code.

I'm not really sure what you mean here. Not every value is a code, and those that are will be in English or numeric anyway so would be rendered LTR by `dir="auto"`. I think I'm missing something.

> Could initial or final punctuation cause additional confusion, especially if the user switches to text view, where every line is in the form `k=v` or `k=*`?

[...]

PR #11091, opened 2025-06-03
Add "dir=auto" to fields with potentially bidirectional text

tyrasd:
Addresses #11079 by setting `dir=auto` to all text input fields that potentially contain bi-directional text.

Unfortunately, this comes with a significant drawback: as the text _direction_ of an input field also affects the direction of the input field's _placeholder_ text,  `dir=auto` means that placeholders will always be rendered left to right. This is because the "value" of the input field itself is empty in the situation where the placeholder is to be shown, and an empty string apparently falls back to left to right. :shrug: 

Apparently, this is the intended behavior of the [placeholders](http://www.w3.org/html/wg/drafts/html/master/forms.html#the-placeholder-attribute) of input fields, and it is not possible to set the placeholder's text direction independently from the input field itself, but there's a documented workaround: 

> In situations where the control's content has one directionality but the placeholder needs to have a different directionality, Unicode's bidirectional-algorithm formatting characters can be used in the attribute value

todo: 
* [ ] fix placeholders' orientation
* [ ] also include address field input elements

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -243,6 +243,12 @@ textarea.mixed::placeholder {
     font-style: italic;
 }
 
+.ideditor[dir='rtl'] input:placeholder-shown[dir="auto"] {
+    /* hard-set rtl for placeholder texts in rtl locales, see
+       https://github.com/openstreetmap/iD/pull/11091#issuecomment-2936953197 */
+    direction: rtl;
+}
+
 /* keytraps need to be invisible yet not be display:none or visibility:hidden */
 .keytrap {
     width: 0;
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -494,6 +494,7 @@ export function uiFieldCombo(field, context) {
         _input = _input.enter()
             .append('input')
             .attr('type', 'text')
+            .attr('dir', 'auto')
             .attr('id', field.domId)
             .call(utilNoAuto)
             .call(initCombo, _container)
--- a/modules/ui/fields/input.js
+++ b/modules/ui/fields/input.js
@@ -91,6 +91,7 @@ export function uiFieldText(field, context) {
         input = input.enter()
             .append('input')
             .attr('type', field.type === 'identifier' ? 'text' : field.type)
+            .attr('dir', 'auto')
             .attr('id', field.domId)
             .classed(field.type, true)
             .call(utilNoAuto)
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -172,6 +172,7 @@ export function uiFieldLocalized(field, context) {
         input = input.enter()
             .append('input')
             .attr('type', 'text')
+            .attr('dir', 'auto')
             .attr('id', field.domId)
             .attr('class', 'localized-main')
             .call(utilNoAuto)
@@ -426,6 +427,7 @@ export function uiFieldLocalized(field, context) {
                 wrap
                     .append('input')
                     .attr('type', 'text')
+                    .attr('dir', 'auto')
                     .attr('class', 'localized-value')
                     .on('blur', changeValue)
                     .on('change', changeValue);
--- a/modules/ui/fields/textarea.js
+++ b/modules/ui/fields/textarea.js
@@ -33,6 +33,7 @@ export function uiFieldTextarea(field, context) {
 
         input = input.enter()
             .append('textarea')
+            .attr('dir', 'auto')
             .attr('id', field.domId)
             .call(utilNoAuto)
             .on('input', change(true))
--- a/modules/ui/fields/wikidata.js
+++ b/modules/ui/fields/wikidata.js
@@ -60,6 +60,7 @@ export function uiFieldWikidata(field, context) {
         searchRowEnter
             .append('input')
             .attr('type', 'text')
+            .attr('dir', 'auto')
             .attr('id', field.domId)
             .style('flex', '1')
             .call(utilNoAuto)
@@ -113,6 +114,7 @@ export function uiFieldWikidata(field, context) {
         enter
             .append('input')
             .attr('type', 'text')
+            .attr('dir', 'auto')
             .call(utilNoAuto)
             .classed('disabled', 'true')
             .attr('readonly', 'true');
--- a/modules/ui/fields/wikipedia.js
+++ b/modules/ui/fields/wikipedia.js
@@ -113,6 +113,7 @@ export function uiFieldWikipedia(field, context) {
     _titleInput = _titleInput.enter()
       .append('input')
       .attr('type', 'text')
+      .attr('dir', 'auto')
       .attr('class', 'wiki-title')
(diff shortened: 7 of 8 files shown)


NeatNit:
Interesting trouble with the placeholder text, I didn't think about this. I don't like the workaround suggested in that link, it wouldn't fix text alignment. I think dynamically setting the `dir` attribute to `ltr`/`rtl` while the input is empty is the better option. My JS is rusty but I'll play around with it in jsfiddle and post here when I have something.

NeatNit:
https://jsfiddle.net/nz0cgush/

This gives ideal behavior, in my opinion.

```html
<input type="text" id="thing" placeholder="שלום!" dir="auto" oninput="recalc_dir(this)" onload=recalc_dir(this)>
```

```js
// I hereby place this script/demo/fiddle in the public domain
// so don't worry about any copyright nonsense

function recalc_dir(o) {
  if (o.value !== "") {
    o.dir = "auto";
  } else {
    o.dir = resolve_text_direction(o.placeholder);
  }
}

function resolve_text_direction(text) {
  // run relevant part of UBA to determine the "paragraph direction" of the given text
  // https://www.unicode.org/reports/tr9/#The_Paragraph_Level
  // Look for existing implementations - might be part of bidi-js https://www.npmjs.com/package/bidi-js
  // annoying that JS/browsers don't expose this as a standard library function, as they need to implement it anyway for dir="auto".
  
  // hard-coded for demo:
  return "rtl";
}
```

Essentially: when there's any value in the input, it sets `dir="auto"` and lets the browser do its thing. When the value is empty, it sets `dir="rtl"` or `dir="ltr"` based on the placeholder text. A full implementation for determining the text direction of the placeholder text is needed. I'm not sure yet if the npm package I linked to actually has the necessary function, but even if not it'd be fairly easy to extract that information from the functions it does expose.

tyrasd:
> I think dynamically setting the dir attribute to ltr/rtl while the input is empty is the better option.

Yeah, something like that would also work. I'm slightly tending towards the recommended solution with the RLE characters, as that does not require to add additional event handlers, which can be tricky to debug. Also, for simplicity I would just fix the direction for the placeholder texts according to the locale's script is (or `ltr` if we know that a placeholder comes from an untranslated source like taginfo). Yes, placeholders from missing translations might have quirky punctuation, but that's only an additional indication that someone should translate the respective string(s).

> It wouldn't fix text alignment

The alignment can be fixed independently from the text "rtl/ltr" problem with css. In fact, I'm also tending towards fixing all "tabular" text to a single alignment: When the UI is in a `rtl` locale (e.g. Hebrew), all text is right-aligned, including raw OSM tags or names in English (and still using the correct respective text direction, of course). And vice versa for `ltr` locales (e.g. English). This helps with readability by keeping consecutive lines aligned horizontally and generally reduces visual noise. I think that makes the most sense from a typographic point of view.

I tried to find examples of how similar situations are handled in reality with mixed directional text (e.g. on road signs, advertisements, maps, etc.), but it seems like the problem is in practice almost always worked around by either justifying or centering all text, by adjusting font sizes to get an almost justified result, or by using completely independent columns for the different scripts. None of those methods work for dynamic and "tabular" data like in iD. Do you maybe have additional examples of how this is approached (well) in practice @NeatNit ?

NeatNit:
> I'm slightly tending towards the recommended solution with the RLE characters, as that does not require to add additional event handlers, which can be tricky to debug.

As you are far more experienced than me in general web development and in iD specifically, I'll just accept this as fact, but I stand by my earlier opinion. That said, alignment should be the only visual difference between the two solutions.

> Also, for simplicity I would just fix the direction for the placeholder texts according to the locale's script is (or `ltr` if we know that a placeholder comes from an untranslated source like taginfo).

That should work.

> Yes, placeholders from missing translations might have quirky punctuation, but that's only an additional indication that someone should translate the respective string(s).

With the embedding solution (the Unicode formatting characters), if you use LTR embedding for untranslated strings and RTL embedding for translated strings, I think punctuation would be rendered correctly in all cases.

> The alignment can be fixed independently from the text "rtl/ltr" problem with css.

You're giving me some silly ideas. I might share them soon. ;)

> In fact, I'm also tending towards fixing all "tabular" text to a single alignment: When the UI is in a `rtl` locale (e.g. Hebrew), all text is right-aligned, including raw OSM tags or names in English (and still using the correct respective text direction, of course). And vice versa for `ltr` locales (e.g. English). This helps with readability by keeping consecutive lines aligned horizontally and generally reduces visual noise. I think that makes the most sense from a typographic point of view.
>
> I tried to find examples of how similar situations are handled in reality with mixed directional text (e.g. on road signs, advertisements, maps, etc.), but it seems like the problem is in practice almost always worked around by either justifying or centering all text, by adjusting font sizes to get an almost justified result, or by using completely independent columns for the different scripts. None of those methods work for dynamic and "tabular" data like in iD. Do you maybe have additional examples of how this is approached (well) in practice @NeatNit ?

This is a tough question with no right answer, but when the column width is narrow enough, I personally prefer each cell to be aligned independently. For technical people like me it's also a good way to tell whether the string is resolved as LTR or RTL, which isn't always obvious from the rendered text alone. For example:

ABC אבג

אבג ABC

Without alignment, you wouldn't know there's a difference between these two by just looking at them.

I will look for examples, but I don't have any on hand. I'd guess in 90% of cases people don't make a conscious decision and just use whatever default behavior the software they're using gave them. I have definitely seen both options in the wild - that is, where an entire column is either left- or right-aligned, or where each cell is aligned independently based on its content.

Worth noting: Excel and LibreOffice Calc by default align each cell independently. I also feel like this is the more commonly used option elsewhere, but atm I don't have the data or examples to back it up.

NeatNit:
> > The alignment can be fixed independently from the text "rtl/ltr" problem with css.
> 
> You're giving me some silly ideas. I might share them soon. ;)

https://jsfiddle.net/bn7jfxth/

```html
<input type="text" placeholder="&#x202B;עברית!&#x202E;" dir="auto">
<br>
<input type="text" placeholder="English!" dir="auto">
```
```css
input:placeholder-shown[dir="auto"][placeholder^="\00202B"] {
  direction: rtl;
}
```

How's that for a compromise?

Edit: come to think of it, the embedding characters aren't needed. Whichever code would have added the embedding characters should instead add the class `rtl-placeholder`:

```html
<input type="text" placeholder="עברית!" dir="auto" class="rtl-placeholder">
<br>
<input type="text" placeholder="English!" dir="auto">
```
```css
input:placeholder-shown.rtl-placeholder[dir="auto"] {
  direction: rtl;
}
```

The code could use whatever heuristic to determine this.

tyrasd:
> `input:placeholder-shown.rtl-placeholder[dir="auto"] { direction: rtl; }`

Oh, for some reason, I thought it's impossible to override the DOM attribute `dir=auto` with a specific direction in CSS. But it seems to work fine. So yeah, that does seem to be the most elegant solution. :+1: Thanks! :bow:

PR #11093, opened 2025-06-03
Enable dependabot for GitHub Actions dependencies

Harvester57:
This PR adds the following:

- Dependabot configuration for GitHub Actions
- Explicit version pinning for GitHub Actions, to improve Depdendabot alerting

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,9 @@ updates:
       - "chore-dependabot"
     ignore:
       - dependency-name: "sinon"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "chore-dependabot"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,9 +21,9 @@ jobs:
         node-version: ['20', '22']
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v4.2.2
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm clean-install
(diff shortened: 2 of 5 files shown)


Harvester57:
I also played with CodeQL for TypeScript/Javascript, and nodejsscan, as you can see here: https://github.com/openstreetmap/iD/compare/develop...Harvester57:iD:develop, but the results (cf. https://github.com/Harvester57/iD/security/code-scanning) are not very convincing, so I left that part out

tordans:
Will this create single PRs per dependency or one Issue where all changes are collected (this dashboard like issue).
Personally, I would prefer the dashboard because getting all those irrelevant notification mails is annoying :-).

---

Something different: Do you know if https://github.com/openstreetmap/id-tagging-schema/pull/1255 makes sense and if we want it here as well?

tyrasd (approved):
Thanks!

tyrasd:
> Will this create single PRs per dependency

I assume so, but as the github actions only really use a small handfull of dependencies, I don't see a huge influx of new PRs.

If you mean this on a general basis: yes, I see that the dependabot PRs can become a bit spammy at times. We could try something else, if you like. But I'm currently failing to find the configuration option to make it do something else than a PR per dependency upgrade. Am I overlooking something?

> Something different: Do you know if https://github.com/openstreetmap/id-tagging-schema/pull/1255 makes sense and if we want it here as well?

Interestingly, in the iD repo, dependabot does already do what you've outlined in the mentioned issue: see e.g. https://github.com/openstreetmap/iD/pull/11085/files
At first glance, I'm not sure what's different in the tagging repo that it behaves differently there… :thinking: But let's investigate this in the other issue, shall we?

Harvester57:
You can group the dependencies PRs into various groups, such as dev and prod environments, or only security-related updates, etc.

The documentation  is here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates

Harvester57:
Here is how it looks when grouped: https://github.com/Harvester57/iD/pull/2

I'll try to separate production and dev dependencies in two different groups, then you can have a different scheduling strategy depending on the type (daily for prod, weekly for dev, for example)

PR #11099, opened 2025-06-04
Use up-to-date NSI when staging

ginnyTheCat:
This updates NSI when staging (like it's already done for ELI) as otherwise the version from `2024-10-14` is used which is quite old.
https://github.com/openstreetmap/iD/blob/c277bd0b765af53789cbf50e16c10ca8e9a5d8f8/package-lock.json#L5848

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -32,9 +32,9 @@ jobs:
       - run: npm run dist
         working-directory: './id-tagging-schema'
       - run: mkdir dist/id-tagging-schema && mv id-tagging-schema/dist dist/id-tagging-schema/dist
-      # build iD using freshest version of presets and ELI
+      # build iD using freshest version of presets, ELI and NSI
       - run: npm clean-install
-      - run: npm install editor-layer-index
+      - run: npm install editor-layer-index name-suggestion-index
       - run: mkdir dist/data
       - run: npm run imagery
       - run: npm run all


tyrasd:
Does it matter which particular version is included in iD? I'm asking because iD only directly uses the js parts of the npm package and I would assume that if something had changed in the js code parts of NSI, there would be a new minor version for that (e.g. `6.1.20250605`). The actual NSI tagging data itself is loaded via CND requests using a dynamic version like this:
`https://cdn.jsdelivr.net/npm/name-suggestion-index@6.0/dist/…` – thus the preview (like the production instances) of iD should always get the most recent cached version of the data set from the CND server.

Bumping the dependency in the package-lock file would not change this behavior, unless we also change how iD is loading the nsi data. For example, instead of loading `@6.0` we could also load it as `…/name-suggestion-index@^6.0.20250601/…` which would ensure that users would always request and get at least the indicated patch version.[^1]

[^1]: this is similar how it is done for the id-tagging-schema, but the tagging schema performs even an additional step to make sure that the loaded versions of the different files of the tagging-schema data are always loaded for the same version. this is AFAIK less of an issue with how NSI is structured, but if we were to change this, it would make sense to also adapt this more robust mechanism for the NSI data.

ginnyTheCat:
Ah, thank you so much for your response. I'm sorry for not researching the code more before opening this. It already being loaded from a CDN is a way better solution than this even is.

PR #11103, opened 2025-06-05
Add signatures audit to CI build pipeline

Harvester57:
Hi !

This PR introduce the following changes, related to security/quality of the build pipeline:

- Add formal names for each build step
- Audit production packages for known vulnerabilities, after checkout and before performing the clean installation
- Audit signatures of installed packages after installation, to detect potential software supply chain attacks
- Add linting step before building
- Move the FORCE_COLOR environment variable as a global variable, to avoid repetition for every build steps

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,8 +30,6 @@ jobs:
       uses: actions/setup-node@v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
-    - name: NPM - Audit for vulnerabilities in production packages
-      run: npm audit --audit-level=info --omit=dev
     - name: NPM - Perform clean install
       run: npm clean-install
     - name: NPM - Audit packages signatures
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,9 +25,9 @@ jobs:
 
     steps:
     - name: Checkout the code
-      uses: actions/checkout@v4.2.2
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v4.4.0
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
     - name: NPM - Perform clean install
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,8 @@ jobs:
       run: npm clean-install
     - name: NPM - Audit packages signatures
       run: npm audit signatures
+    - name: NPM - Lint codebase
+      run: npm run lint
     - name: NPM - Build all
       run: npm run all
     - name: NPM - Run tests
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,8 +34,6 @@ jobs:
       run: npm clean-install
     - name: NPM - Audit packages signatures
       run: npm audit signatures
-    - name: NPM - Lint codebase
-      run: npm run lint
     - name: NPM - Build all
       run: npm run all
     - name: NPM - Run tests
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,8 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - name: NPM - Perform clean install
       run: npm clean-install
+    - name: NPM - Audit packages signatures
+      run: npm audit signatures
     - name: NPM - Build all
       run: npm run all
     - name: NPM - Run tests
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,17 +24,12 @@ jobs:
         node-version: ['20', '22']
 
     steps:
-    - name: Checkout the code
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
-    - name: NPM - Perform clean install
-      run: npm clean-install
-    - name: NPM - Audit packages signatures
-      run: npm audit signatures
-    - name: NPM - Build all
-      run: npm run all
-    - name: NPM - Run tests
-      run: npm run test
+    - run: npm clean-install
+    - run: npm audit signatures
+    - run: npm run all
+    - run: npm run test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,11 +24,15 @@ jobs:
         node-version: ['20', '22']
 
     steps:
-    - uses: actions/checkout@v4.2.2
+    - name: Checkout the code
+      uses: actions/checkout@v4.2.2
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm clean-install
-    - run: npm run all
-    - run: npm run test
+    - name: NPM - Perform clean install
+      run: npm clean-install
+    - name: NPM - Build all
+      run: npm run all
+    - name: NPM - Run tests
+      run: npm run test
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,9 @@ on:
 permissions:
   contents: read
 
+env:
+  FORCE_COLOR: 2
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -27,11 +30,5 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm clean-install
-      env:
-        FORCE_COLOR: 2
     - run: npm run all
-      env:
-        FORCE_COLOR: 2
     - run: npm run test
-      env:
-        FORCE_COLOR: 2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,8 @@ jobs:
       uses: actions/setup-node@v4.4.0
       with:
         node-version: ${{ matrix.node-version }}
+    - name: NPM - Audit for vulnerabilities in production packages
+      run: npm audit --audit-level=info --omit=dev
     - name: NPM - Perform clean install
       run: npm clean-install
     - name: NPM - Audit packages signatures
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -11,8 +11,8 @@ jobs:
     name: Check for spelling errors
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4.2.2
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -21,17 +24,21 @@ jobs:
...
-      env:
-        FORCE_COLOR: 2
+    - name: NPM - Audit for vulnerabilities in production packages
+      run: npm audit --audit-level=info --omit=dev
+    - name: NPM - Perform clean install
+      run: npm clean-install
+    - name: NPM - Audit packages signatures
+      run: npm audit signatures
+    - name: NPM - Lint codebase
+      run: npm run lint
```
tyrasd: `npm run test` already runs the linter, so this would be running it twice which is unncessary.

.github/workflows/build.yml
```diff
@@ -21,17 +24,21 @@ jobs:
...
-      env:
-        FORCE_COLOR: 2
-    - run: npm run all
-      env:
-        FORCE_COLOR: 2
-    - run: npm run test
-      env:
-        FORCE_COLOR: 2
+    - name: NPM - Audit for vulnerabilities in production packages
+      run: npm audit --audit-level=info --omit=dev
```
tyrasd: What's the benefit of adding this to the test workflow? We've already set up dependabot to alert us about vulnerabilities directly though github (i.e. `https://github.com/<org>/<repo>/security/dependabot`), so this wouldn't add anything new, would it?

.github/workflows/build.yml
```diff
@@ -21,17 +24,21 @@ jobs:
...
+    - name: NPM - Perform clean install
+      run: npm clean-install
+    - name: NPM - Audit packages signatures
+      run: npm audit signatures
```
tyrasd: Maybe I am misunderstanding what `audit signatures` does, but under which condition can a freshly installed instance fail this step?

tyrasd (dismissed):
Honestly, I'm not the biggest fan of repeating a command like `npm run dist` just slightly differently worded as the step's name: `NPM - Build dist`. That's just needlessly adding lines to the script without any benefit. :shrug: 

On the concrete suggestions to the changed workflow logic: I've got a few questions, see them inline below.

.github/workflows/build.yml
Harvester57: You're entirely right, will remove this step !

.github/workflows/build.yml
Harvester57: That's a fair question. While you will already have a fresh install after the previous step, a clean install only verifies the integrity of the downloaded packages against your local package-lock.json file, an audit of the signatures checks that the packages downloaded was signed by the expected maintainer of the package, even in the event of a takeover of the account publishing on the NPM repository, thus providing an additional defense against a supply chain attack on one of the packages you pull. Is that the mechanism you had in mind for signature verification ?

.github/workflows/build.yml
Harvester57: Agreed, I'll remove this step :)

Harvester57:
> Honestly, I'm not the biggest fan of repeating a command like `npm run dist` just slightly differently worded as the step's name: `NPM - Build dist`. That's just needlessly adding lines to the script without any benefit. 🤷

Fair enough, I'll remove the redundant names for the steps

.github/workflows/build.yml
```diff
@@ -21,17 +24,12 @@ jobs:
         node-version: ['20', '22']
 
     steps:
-    - uses: actions/checkout@v4.2.2
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
k-yle, line 27: what's the benefit of pinning to a commit hash for **first-party** packages published by GitHub themselves?

Harvester57: I'm not sure there is such a distinction between first-party and third-party actions, in the sense that the checkout action, and all other GitHub official actions, are following the same development model as any other action: they have a public GH repository (https://github.com/actions/checkout), that is open to external contributors, and therefore the threat model is the same for this action as any other one.

They are not immune to a repository takeover that would push or rewrite an existing tag, and I would argue that given the ubiquity of this particular checkout action, it is even more important to pin them.

k-yle: I see, I was wondering becuase I regularly hear the opposite advice: never pin to a commit hash, because the commit hash could come from a fork of the repository (which is [a well known GitHub quirk](https://trufflesecurity.com/blog/anyone-can-access-deleted-and-private-repo-data-github)
<sup>[_better source needed_]</sup>)

For example, a malicious contributor could change this line to the following:
```suggestion
    - uses: actions/checkout@5a96951a20872f15b767339256a01fe5c4d1478f # v4.123.123
```
And for a code reviewer, it's a lot of work to manually check if the commit hash matches the version in the code comment. So the code reviers might not bother to check, since "actions/checkout" looks legitimate.

In that example above, commit [5a9695](https://github.com/actions/checkout/commit/5a9695) comes from a deleted fork, so it could contain any malicious code.

...and GitHub Actions will happily pull the malicious commit from the fork and run it, ignoring the explicit repository name of "actions/checkout":
image

---

anyway.... this is all very hypothetical, so don't read this as a code review, I'm just curious about the rationale behind this change

tordans: But `actions/checkout` is actually maintained by Github, right? At least [the COC](https://github.com/actions/checkout?tab=coc-ov-file) suggests so. TBH I have a higher trust that Github has their merges under control … or resolves them quickly than any other dependency we have in iD. Which is why I would vote to keep this simple and follow usage recommendations they provide https://github.com/actions/checkout?tab=readme-ov-file#usage

Harvester57: @k-yle: manual updates of Actions SHA should be prohibited and only PRs opened by Dependabot should be reviewed and merged, especially since you already have Dependabot configured for this repository.

Even GitHub's own Secure Reference manual recommend to pin the SHA of third party Actions: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

@tordans: it is indeed maintained by GitHub, but the question is, why should you treat Actions developed by GitHub any differently than any other Action ? From your perspective, they are part of your supply chain, and you use them to deploy staging instances and build your code in your workflows.

I would argue that, given their market size, and give the amount of repos using these basic Actions (checkout, setup, ...), it makes them a way more interesting target than any other third party Action. You might expect (and this could be the case in practice) that they have additional security measures and processes in place for these repositories, but their employees and code owners are still vulnerable to phishing, or any other type of attack that you could read about in other supply chain attacks in the past few years.

Regarding the argument of speed of resolution, you might be right, but if I look at the iD repository statistics of your workflows (https://github.com/openstreetmap/iD/actions/metrics/usage?dateRangeType=DATE_RANGE_TYPE_CURRENT_WEEK), the build job ran on average 10 times per day for the last week, and the deploy to staging instance 3.5 times per day, which in my opinion leaves plenty of time to an attacker that compromised a tagged Action to run its code.

Harvester57: I'd like to point out that you already have a very good best practice in place (https://github.com/openstreetmap/iD/pull/9273) by reducing the GITHUB_TOKEN default permissions, very positive !

k-yle: i find this topic quite interesting, because this approach basically moves the onus to the downstream maintainer to manually enforce your rule that _"manual updates of Actions SHA should be prohibited"_

and the likelihood of someone forgetting this rule is significantly higher than the likelihood of github's own repository getting compromised.

which is why i'm quite suprised that [github's advice](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) is the opposite of [defensive design](https://en.wikipedia.org/wiki/Defensive_design). There's also a mistake on that docs page which affects the viability of this solution. The docs proclaim:

> Dependabot only creates alerts for vulnerable actions that use semantic versioning and will not create alerts for actions pinned to SHA values. [[source]](https://docs.github.com/en/actions/reference/security/secure-use#monitoring-the-actions-in-your-workflows)

But this contradicts https://github.com/dependabot/dependabot-core/pull/5951 😆 suggesting that (at least part of) those docs are 3 years out-of-date

Harvester57: I'm failing a bit to see how it's opposing the defensive design paradigm, could you expand on this ? To be noted that pinning hashes is also recommended by the OpenSSF in their Scorecard approach, with the same rationale, and the same logic to let the dependencies be updated by automated means: https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

tyrasd (approved):
Thanks for the PR and the extra background information @Harvester57. I did also learn quite a bit in this thread. At first I was of the impression that it might be overkill to pin dependencies from the very same company that provides us also with the rest of our build infrastructure (i.e. dependabot and the run environment itself, which _theoretically_ could also become compromised).

But I agree that there is some merit to apply the same guideline that we have for other dependencies (i.e. pin versions to their commit hash; for npm dependencies via the package-lock file) also for all other dependencies, regardless of who the maintainer is.

Yes, we might get some extra dependabot PRs from time to time, but that shouldn't be an issue in practice. And any PRs from external contributors that update the hashes (in the gh workflows or package-lock) should automatically ring the bells to be extra cautious when reviewing a PR.

.github/workflows/build.yml
tyrasd: PS: I've now applied the version/hash pinning also to the other two workflows (codespell and transifex) that we use in this repo.

k-yle: sorry, I realised I never replied to this discussion 😅 For completeness, I meant that pinning to a commit hash requires maintainers to do extra work[^1], with no safeguards to ensure that the hashes are actually checked.

Generally speaking, requiring people to remember and follow an SOP isn't great defensive design, and in other industries this would be one the [weakest types of defenses](https://en.wikipedia.org/wiki/Hierarchy_of_hazard_controls).

---

So it's easy for a malicious user to fool a maintainer into accepting code that looks legitimate, such as this line:
```yaml
- uses: actions/checkout@85b8272e24166014adcf9ff6ef3701031878ec80 # v5.0.0
```
It includes the words _actions/checkout_, so many people would assume that this code must come from the [actions/checkout](https://github.com/actions/checkout) repo. but it doesn't!

if we open that commit (https://github.com/actions/checkout/commit/85b8272e24166014adcf9ff6ef3701031878ec80) the author and content match [the real v5 commit](https://github.com/actions/checkout/commit/08c6903cd8c0fde910a37f88322edcfb5dd907a8), the purported author is a github employee, and the URL contains _actions/checkout_. So it looks very legitimate.   
There is [a warning banner](https://github.com/user-attachments/assets/a7e6bbd2-60d9-4846-ac2d-4c9bcb760482), but it only explains the _problem_, not the _risk_. And unfortunately this banner is [understandably](https://en.wikipedia.org/wiki/Alarm_fatigue) oft [ignored](https://www.stepsecurity.io/blog/the-github-warning-everyone-ignores-this-commit-does-not-belong-to-any-branch) because of [several](https://keturn.net/blog/posts/2021/01/whats-lost-in-the-squash/#:~:text=with%20no%20indication) known [false-positives](https://github.com/orgs/community/discussions/19021).

In reality, [`85b8272`](https://github.com/actions/checkout/commit/85b8272e24166014adcf9ff6ef3701031878ec80) is not from the `actions/checkout` repo. It's from a deleted fork, and contains [malicious code](https://github.com/actions/checkout/blob/85b8272e24166014adcf9ff6ef3701031878ec80/dist/index.js#L1) added several commit ago, with fake commit history [replayed](https://git-scm.com/docs/git-replay) ontop of the malicious commit.

---

So anyway, this approach works for us, but it's shockingly easy to fool people when you use commit hashes


[^1]: (manually validating every commit hash added by external contributors)

PR #11104, opened 2025-06-05
Update development dependencies

Harvester57:
Bumps the npm-dev-dependencies group with 2 updates in the / directory: [editor-layer-index](https://github.com/osmlab/editor-layer-index) and [name-suggestion-index](https://github.com/osmlab/name-suggestion-index).

Updates `editor-layer-index` from `3e15818` to `87b6ab6`
- [Commits](https://github.com/osmlab/editor-layer-index/compare/3e15818cc1da3945a3e3747a9203942dde88a90a...87b6ab66b2a4dbf2407e33120e789b0526133ed1)

Updates `name-suggestion-index` from 6.0.20241014 to 6.0.20250601
- [Changelog](https://github.com/osmlab/name-suggestion-index/blob/main/CHANGELOG.md)
- [Commits](https://github.com/osmlab/name-suggestion-index/compare/v6.0.20241014...v6.0.20250601)

---
updated-dependencies:
- dependency-name: editor-layer-index dependency-version: 87b6ab66b2a4dbf2407e33120e789b0526133ed1 dependency-type: direct:development dependency-group: npm-dev-dependencies
- dependency-name: name-suggestion-index dependency-version: 6.0.20250601 dependency-type: direct:development update-type: version-update:semver-patch dependency-group: npm-dev-dependencies ...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3718,8 +3718,10 @@
     },
     "node_modules/editor-layer-index": {
       "version": "0.0.0",
-      "resolved": "git+ssh://git@github.com/osmlab/editor-layer-index.git#3e15818cc1da3945a3e3747a9203942dde88a90a",
-      "dev": true
+      "resolved": "git+ssh://git@github.com/osmlab/editor-layer-index.git#87b6ab66b2a4dbf2407e33120e789b0526133ed1",
+      "integrity": "sha512-UtETrMxFgZHgOWuWcxHWIbaYaF3ZlhrPYFSaOw5SwPazgCZSm0sBqFViXD7LKCKGv+Rpuw0VQCxhY7BnldBwwA==",
+      "dev": true,
+      "license": "BSD"
     },
     "node_modules/electron-to-chromium": {
       "version": "1.5.161",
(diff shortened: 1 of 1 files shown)


Harvester57:
Ah, this is a slight duplicate of this PR, I believe: https://github.com/openstreetmap/iD/pull/11099

tyrasd:
Yeah, the data of the name-suggestion-index is only used indirectly through a CDN, that's why I don't think there's a benefit in bumping that dependency regularly (except of course if there were a proper new minor release with code changes). See also https://github.com/openstreetmap/iD/pull/9576#issuecomment-2944114433

tyrasd:
`editor-layer-index` is similar, but also a bit different (as it is not available in a versioned form through a CDN). It is updated manually right before every release (see https://github.com/openstreetmap/iD/blob/develop/RELEASING.md).

bhousel:
> `editor-layer-index` is similar, but also a bit different (as it is not available in a versioned form through a CDN). It is updated manually right before every release (see https://github.com/openstreetmap/iD/blob/develop/RELEASING.md).

I would be remiss to not mention that Quincy and I did build a working version of ELI that does have all the geojson bugs fixed and published to a CDN with versioning - and that we had switched iD over to using it before Mikel forced me out of the iD maintainer position and made Quincy revert the change.  

Anyway you should really switch to it!  We we even happy to keep it license compatible with ELI, contrary to whatever misinformation you might have heard!

Is it time to reopen this conversation?
https://github.com/openstreetmap/iD/issues/7425

PR #11124, opened 2025-06-12
issues:11120 -automated id extraction from a pasted url in tags ( only for contacts)

### LINKED ISSUE #11120: Handle url->id mapping, id->url click linking for key:contact=*

Eshan276:
as the issue  #11120 says , "It would be idea, if pasting the URL transparently returned just the id (username, etc) from the URL. And if rendering it reversed this process when generating the link so it could be edited."

the linkification part was done and PR is also there [(comment)](https://github.com/openstreetmap/iD/issues/11120#issuecomment-2964956960), 

in this PR i have pushed the code , to auto extract the ids from url in tags(contacts) .

### DIFF
--- a/modules/ui/sections/raw_tag_editor.js
+++ b/modules/ui/sections/raw_tag_editor.js
@@ -194,7 +194,29 @@ export function uiSectionRawTagEditor(id, context) {
             .call(utilNoAuto)
             .on('focus', interacted)
             .on('blur', valueChange)
-            .on('change', valueChange);
+            .on('change', valueChange)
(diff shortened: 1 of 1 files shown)


k-yle:
I'm going to close this, because of https://github.com/openstreetmap/iD/issues/11120#issuecomment-2996612693 

The other (uncontroversial) parts of that issue have since been implemented in https://github.com/openstreetmap/openstreetmap-website/pull/6197 and https://github.com/openstreetmap/iD/pull/9988

PR #11141, opened 2025-06-18
Integrate indexeddb for history and prefs sync with server

### LINKED ISSUE #3239: Switch to async storage (i.e. indexdb instead of localstorage)

### LINKED ISSUE #11160: Migrate history storage from localStorage to IndexedDB and add async preferences API

Deeptanshu-sankhwar:
### Changes Made

• **Added IndexedDB storage** using `idb-keyval` library to replace localStorage for history data
• **Created `asyncPrefs` API** for asynchronous preference operations alongside existing `prefs`
• **Implemented `migrateHistoryData()`** to automatically migrate existing localStorage history to IndexedDB
• **Updated history module** to save/restore from IndexedDB instead of localStorage
• **Added test cases** for migration function and asyncPrefs functionality

### Dependencies
• Added `idb-keyval` for IndexedDB operations
• Added `fake-indexeddb` for testing

This enables better data persistence while maintaining backward compatibility.

migration_flowchart


modules/core/preferences.js
```diff
@@ -1,5 +1,8 @@
 // https://github.com/openstreetmap/iD/issues/772
 // http://mathiasbynens.be/notes/localstorage-pattern#comment-9
```
tordans: This commend might be outdated.

Deeptanshu-sankhwar: Thanks for you pointing it out @tordans , shall I remove the comment from `preferences.js` then?

Deeptanshu-sankhwar:
Closes #11160

tyrasd:
I found one more comment (in [`context.js#L294`](
https://github.com/Deeptanshu-sankhwar/iD/blob/73b2981beabb749072fb8e143f2ef1fc46569dad/modules/core/context.js#L294)) which is now outdated, and could need an update:

```js
  // Debounce save, since it's a synchronous localStorage write,
  // and history changes can happen frequently (e.g. when dragging).
  context.debouncedSave = _debounce(context.save, 350);
```

I assume there is still merit to keep the debouncing logic on the history-saving function (`debouncedSave`), even though it's less of an issue after switching from localstorage to indexeddb. But perhaps we could reduce the debouncing time to something that is triggering slightly more frequently (e.g. like 100ms). Do you think that makes sense @Deeptanshu-sankhwar ?

modules/core/history.js
```diff
@@ -355,6 +355,7 @@ export function coreHistory(context) {
                 _tree = coreTree(_stack[0].graph);
                 _checkpoints = {};
             }
+            _pausedGraph = null;
```
tyrasd, line 354: This variable looks to be unused? What is/was its purpose?

modules/core/history.js
```diff
@@ -652,20 +653,25 @@ export function coreHistory(context) {
             if (lock.locked() &&
                 // don't overwrite existing, unresolved changes
                 !_hasUnresolvedRestorableChanges) {
-                const success = prefs(getKey('saved_history'), history.toJSON() || null);
 
-                if (!success) dispatch.call('storage_error');
+                const historyData = history.toJSON();
+                if (historyData) {
+                    asyncPrefs.set(getKey('saved_history'), historyData);
+                    prefs(getKey('has_saved_history'), true);
```
tyrasd: I'd say it would be more stable if this was only set after the async set is complete:

`asyncPrefs.set(…).then(() => prefs(…, true))`

modules/core/history.js
```diff
@@ -652,20 +653,25 @@ export function coreHistory(context) {
...
-        // delete the history version saved in localStorage
+        // delete the history version saved in IndexedDB
         clearSaved: function() {
             context.debouncedSave.cancel();
             if (lock.locked()) {
                 _hasUnresolvedRestorableChanges = false;
-                prefs(getKey('saved_history'), null);
+
+                asyncPrefs.set(getKey('saved_history'), undefined);
+                prefs(getKey('has_saved_history'), null);
```
tyrasd: see above

modules/core/preferences.js
```diff
@@ -41,10 +45,57 @@ function corePreferences(k, v) {
...
 export { corePreferences as prefs };
+
+export const asyncPrefs = {
+  /** @param {string} key */
+  get(key) {
+    if (corePreferences(key)) {
+      const parsed = JSON.parse(corePreferences(key));
+      corePreferences(key, null);
+      return parsed;
+    }
```
tyrasd: when/how is this code branch used?

modules/core/preferences.js
```diff
@@ -41,10 +45,57 @@ function corePreferences(k, v) {
...
+
+    return get(key);
+  },
+  set,
+};
+
+export async function migrateHistoryData() {
+  const historyKeyPattern = /^iD_.*_saved_history$/;
```
tyrasd: This looks unnecessarily complicated. the storage key for the history data is already known (`getKey('saved_history')`), so there's no need to do this with a regular expression. Or am I overlooking something?

test/spec/core/history.js
```diff
@@ -571,4 +567,54 @@ describe('iD.coreHistory', function () {
             expect(history.difference().deleted().length).to.eql(1);
         });
     });
+
+    describe('#replaceLocalStorageWithIndexedDB', function() {
+        beforeEach(function() {
+            var FDBFactory = require('fake-indexeddb/lib/FDBFactory');
```
tyrasd: please use `import …` syntax to load dependencies also in test files

test/spec_helpers.mts
```diff
@@ -6,6 +6,12 @@ import 'happen';
 import fetchMock from 'fetch-mock';
 import envs from '../config/envs.mjs';
 
+// Setup global IndexedDB mock for all tests
```
tyrasd: what's the reason to go with this setup over fake-indexeddb's auto setup (`import "fake-indexeddb/auto";`)?

tyrasd (changes requested):
Please rebase the PR on the current develop branch to solve the conflict in `package-lock.json`.

Also see some inline remarks below.

modules/core/history.js
Deeptanshu-sankhwar: Hello @tyrasd , `_pausedGraph` variable is used to temporarily store a graph state when change event dispatching is pause, I've set `_pausedGraph = null` in the `reset` function to ensure that any paused state is cleared when the history is reset

modules/core/history.js
Deeptanshu-sankhwar: I have chained setting the `has_saved_history`, inside the async call now

modules/core/history.js
Deeptanshu-sankhwar: made the same changes here

modules/core/preferences.js
Deeptanshu-sankhwar: this change was relating if in case the migration logic fails, but this appears to be very redundant instead, I have removed the redundant code from here, since `migrateHistoryData` function would take care of moving key, values from localstorage to indexeddb

modules/core/preferences.js
Deeptanshu-sankhwar: Thank you for pointing, I have use `saved_history` directly to get the history keys, instead of regex

test/spec_helpers.mts
Deeptanshu-sankhwar: I wasn't aware of `fake-indexeddb/auto`, I have used it now directly instead of cherry picking modules from `fake-indexeddb`, it is very helpful to automatically setup mock implementation of IndexedDB

tyrasd (approved):

PR #11158, opened 2025-06-27
Render climbing routes

### LINKED ISSUE #11133: Add rendering for climbing route features
### Description

Right now, climbing features don't have any styling
> Image

[Location](https://ideditor.netlify.app/#locale=en&map=17.54/68.16505/16.59361&disable_features=boundaries&background=geovekst-nib)

The issues connected to https://github.com/openstreetmap/id-tagging-schema/issues/31 work on addding presets for those features.

I think we should look into adding some styling for those ways as well.

### Screenshots

Looking at existing styles (Left to right: Ladder, Steps, Slipway(Drivable), Chair Lift) I think something that looks similar to Ladder/Steps is good, maybe with a red color like the chair lift (something that highlights the "danger" :)).

Image

harahu:
Custom rendering for climbing routes. Work in progress, but I thought I'd share my branch as a start.

Closes #11133


modules/renderer/features.js
```diff
@@ -210,6 +210,10 @@ export function rendererFeatures(context) {
         return tags['piste:type'];
     });
 
+    defineRule('climbing', function isClimbing(tags){
```
tyrasd: I'm hondering: Is a new feature category really warranted for something as rare as the climbing features?

If we decide to add one, we also need to add the respective strings to `core.yaml` (`en -> feature -> climbing -> description / tooltip`, currently around line 950 in the file), so that these texts can be properly displayed and translated.

harahu: You shouldn't read too much intention into the changes I've made in this PR. I'm quite unfamiliar with the iD code base, so my approach to getting rendering of climbing routes to work was just to follow the pattern established by pistes. In other words, this PR is currently just me doing cargo culting.

I want to achieve two things with this PR:
1. get iD to render climbing routes in some pretty way,
2. get iD to understand that climbing routes are routable, such that paths that are only accessible by way of climbing routes aren't identified as routing islands in the editor.

All code that doesn't contribute to those two goals can be stripped out of the PR.

harahu: I dropped the addition of this rule now.

tyrasd (approved):
Looks great, thanks!

PR #11182, opened 2025-07-02
support RegExs for `prerequisiteTag`s

k-yle:
Split out of #9511. That PR had logic to support RegExs, which was added 2 years ago, but the more recent discussion hasn't considered this part. So I'm moving this into a seperate PR. 

The ~~only~~ fields that need this are [`seamark:*:colour_pattern`](https://taginfo.osm.org/keys/seamark:topmark:colour_pattern), which should only be shown if [`seamark:*:colour`](https://taginfo.osm.org/keys/seamark:topmark:colour) contains a `;`

_edit: ...and perhaps [`parking:SIDE:orientation=*`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/parking/side/orientation.json), see https://github.com/openstreetmap/iD/pull/11182#issuecomment-3036050854_

### DIFF
--- a/modules/ui/field.js
+++ b/modules/ui/field.js
@@ -354,6 +354,9 @@ export function uiField(context, presetField, entityIDs, options) {
                         return prerequisiteTag.valueNot !== value;
                     }
                     if (prerequisiteTag.value) {
+                        if (prerequisiteTag.value.startsWith('/') && prerequisiteTag.value.endsWith('/')) {
+                            return new RegExp(prerequisiteTag.value.slice(1, -1)).test(value);
+                        }
                         return prerequisiteTag.value === value;
                     }
                 } else if (prerequisiteTag.keyNot) {


tordans:
This would also need some docs changes, right?

k-yle:
Yes, but only if we still want to do this. I'll leave this PR here for further discussion. There are no other use-cases that I can think of

tordans:
@k-yle would this help in this case… https://github.com/ideditor/schema-builder/issues/137#issuecomment-2087984901 ?

```
const regex = /^(?!no$)(?!separate$).+/i;

console.log(regex.test("yes"));      // true
console.log(regex.test("maybe"));    // true
console.log(regex.test("no"));       // false
console.log(regex.test("separate")); // false
console.log(regex.test(""));         // false
console.log(regex.test(null));         // true <-- is this the issue here?
```

k-yle:
yes, it could solve this case: https://regexr.com/8fs7p

**EDIT:** actually, we can already solve this using #9511:
```json
{
  "key": "parking:left:orientation"
  "prerequisiteTag": { "key": "parking:left", "valuesNot": ["", "no", "separate"] }
}

**Comment by matkoniecz** [comment]
> The only fields that need this are [seamark:*:colour_pattern](https://taginfo.osm.org/keys/seamark:topmark:colour_pattern), which should only be shown if [seamark:*:colour](https://taginfo.osm.org/keys/seamark:topmark:colour) contains a ;

I am bit confused. Is it a theoretical use? Is it possible to test this PR?

I tried to test it on https://www.openstreetmap.org/way/336736331/history#map=19/53.981440/14.580307 which gets no `seamark:light:colour` (and it seems that https://wiki.openstreetmap.org/wiki/Key%3Aseamark%3Alight%3Acolour recommends to not add one)

Then I tried on https://www.openstreetmap.org/node/11233267310 and it just shows as generic seamark.

Then I grepped for colour_pattern in iD tagging schema repo and found no matches.

**Comment by matkoniecz** [comment]
@k-yle is there an actual use for this change? Or is intended to support not yet implemented preset?

If yes maybe it would make to implement it in synch so this change is testable?

**Comment by k-yle** [comment]
- it would help with [`parking:SIDE:orientation=*`](https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/parking/side/orientation.json), which is an existing preset. see [#11182 (comment)](https://github.com/openstreetmap/iD/pull/11182#issuecomment-3036050854) for further discussion 
- it would also with [`seamark:*:colour_pattern`](https://taginfo.osm.org/keys/seamark:topmark:colour_pattern), which should only be shown if [`seamark:*:colour`](https://taginfo.osm.org/keys/seamark:topmark:colour) contains a `;`. Used [>28k times](https://taginfo.openstreetmap.org/search?q=colour_pattern), but not in the presets yet.


in any case, i think this should wait until after schema builder v7, or be closed as https://github.com/openstreetmap/iD/labels/wontfix-low-impact ~wontfix-niche

**Comment by tordans** [comment]
>parking:SIDE:orientation=*,

This might be unrelated… But I have a unresolved "todo" from a community chat month ago where I talked about making the preset rendering of directional combo in iD smarter by adding a guard that a sub fields can only be active when the main field is present (and not "no"-case). I think this targets the same idea as changes here. We would move logic from schema to the implementation… but it will likely be easier to handle the SIDE part.

**Comment by k-yle** [comment]
yeah, that works too, but a generic solution would only check values like `no` and `none`, whereas `parking:SIDE:orientation=*` also needs to check `separate`. 

so it might be hard to solve that paticular case with a generic solution

File: modules/ui/field.js [hunk 1]
```diff
@@ -339,21 +347,18 @@ export function uiField(context, presetField, entityIDs, options) {
...
-                        return !prerequisiteTag.valuesNot.includes(value);
-                    }
                     if (prerequisiteTag.valueNot) {
                         return prerequisiteTag.valueNot !== value;
                     }
-                    if (prerequisiteTag.values) {
-                        return prerequisiteTag.values.includes(value);
-                    }
                     if (prerequisiteTag.value) {
+                        if (prerequisiteTag.value.startsWith('/') && prerequisiteTag.value.endsWith('/')) {
```
matkoniecz, line 357: I would say that it should be first allowed by https://github.com/ideditor/schema-builder#prerequisitetag

currently there is no indicator that r expression is valid there

k-yle:
closing for now, since you could argue that this would be a breaking change, and therefore would need to wait for schema-builder v8

lets start by updating the docs in schema builder first, then updating iD if the schema-builder change if accepted

PR #11184, opened 2025-07-02
Fix amenity=shelter incorrectly classified as landuse

### LINKED ISSUE #11153: Imprecise "landuse" feature type
### URL

https://www.openstreetmap.org/edit#map=20/52.0669047/5.1164352

When a closed way gets the tag amenity=shelter (without building tag) and you filter away in the map data on landuse features, the shelter disappears. I wouldn't expect this behaviour, since an amenity isn't a landuse. I would expect it would be filtered under "other features". Could this be changed? Much appreciated!

Image
Image

_No response_

2.34.1

Microsoft Edge

youssefelzedy:
# Fix amenity=shelter incorrectly classified as landuse
Fixes #11153

When filtering map data, closed ways with `amenity=shelter` (without a building tag) were incorrectly being classified as "landuse" features and would disappear when landuse features were filtered out. Users expected amenities like shelters to appear under "other features" instead.

The `isLanduse` function in `modules/renderer/features.js` was using a overly broad check `!!tags.amenity` that treated **any** amenity tag as landuse, regardless of whether the amenity actually represents land use.

1. **Added `osmLanduseAmenityTagValues` whitelist** in `modules/osm/tags.js` containing only amenity values that represent actual land use areas
2. **Updated landuse rule** to use specific amenity checking instead of broad `!!tags.amenity`

#### `modules/osm/tags.js`
- Added `osmLanduseAmenityTagValues` object with 12 specific amenity types that represent actual land use:
  - Area-based amenities: `parking`, `motorcycle_parking`, `bicycle_parking`
  - Educational facilities: `school`, `college`, `university` 
  - Large institutional areas: `hospital`, `prison`, `grave_yard`, `place_of_worship`
  - Public spaces: `marketplace`

#### `modules/renderer/features.js`
- Updated landuse rule to use: `(tags.amenity && osmLanduseAmenityTagValues[tags.amenity])`
- Replaced broad check: `!!tags.amenity`

- `amenity=shelter` now correctly appears under "other features"
- Point amenities (bench, ATM, restaurant, etc.) no longer misclassified as landuse
- Area amenities that are actual land use (parking, universities) remain correctly classified
- Maintains existing behavior for legitimate landuse amenities
- Follows established code patterns in the project

The current whitelist is based on the [OSM Land Use wiki table](https://wiki.openstreetmap.org/wiki/Land_use#Table_of_some_existing_land_use_tags). Additional amenity types could be added if they represent actual land use areas, following the same pattern.

This approach provides a maintainable solution that centralizes OSM tag semantics in `modules/osm/tags.js` while fixing the classification issue.


tyrasd (approved):
looks good, thanks!

modules/renderer/features.js
```diff
@@ -2,7 +2,8 @@ import { dispatch as d3_dispatch } from 'd3-dispatch';
 
 import { prefs } from '../core/preferences';
 import { osmEntity, osmIsInterestingTag, osmLifecyclePrefixes } from '../osm';
-import { osmLanduseAmenityTagValues } from '../osm/tags.js';
+import { osmLanduseAmenityTagValues, osmLifecyclePrefixes } from '../osm/tags.js';
+import { osmEntity } from '../osm';
```
youssefelzedy, line 4: `osmEntity` is duplicated in lines 4 and 6

tyrasd: yes, that was my mistake. I've ultimately addressed that in the final merge commit, see https://github.com/openstreetmap/iD/commit/a6ac48cd512195a9ad27b0f327be163182b68505#diff-1b592a9dc03ac1954ba91b21ebec500e0e533a51108ccb941d6c8e343267a515

But thanks for noticing it anyway! :bow:

Issue #9589, opened 2023-04-18
Don't let users attach e.g., paths to Address Interpolation


jidanni:
### Description

Don't let, or at least warn, when users attempt to attach real life objects to things like
[address interpolation](https://wiki.openstreetmap.org/w/index.php?title=Key:addr:*&uselang=en#Tags_for_interpolation_ways). I mean it is even more abstract than a political boundary.

### Screenshots

![20230418T154532](https://user-images.githubusercontent.com/1230959/232709069-cc96d6cf-ae51-44d7-9b5c-f813ab6ba436.jpg)
(User attempts to attach path to https://www.openstreetmap.org/way/264949636 . Good thing I stopped him just in time :-))

tyrasd:
closing in favor of the more generic issue #9488

Issue #6631, opened 2019-07-07
Warn when landuse is connected to highways


SilentSpike:
Probably one of the most common issues I fix while mapping which can be an easy mistake to make for newer mappers (and highly tedious to correct).

As far as I'm aware there is no valid reason for landuse to be connected to the highway network with the exception of a highway endpoint (e.g. service road which leads to a carpark).

1ec5:
I’ve spilled a lot of ink over the years arguing that connecting landuse to highways can be a _good_ thing in some cases, depending on [regional norms](https://wiki.openstreetmap.org/wiki/Ohio/Map_features#Land_use). But clearly it isn’t welcome in many parts of the world, so a warning would be useful where mappers want the ability to avoid connections to landuse/landcover.

Still, the warning would need to make exceptions for some scenarios. Given the parking lot example above, I assume the proposal would affect more than just `landuse` areas. Here are some edge cases off the top of my head:

* A park has a named entrance that you drive into. An `entrance` node connects a `highway=service` way with an `amenity=park` area, but it is not the endpoint of the `highway=service` way.
* A greenway trail has a named trailhead. An `highway=trailhead` node connects a `highway=footway` way with an `amenity=park` area.
* A gated residential subdivision is surrounded by a white picket fence, and a street leads past a gate into the subdivision. The closed `barrier=fence` way is connected to the entire `landuse=residential` area representing the subdivision. A `barrier=gate` node connects a `highway=residential` way to the `barrier=fence` way and thus the `landuse=residential` area.
   * Same, but for a driveway leading into a fenced-off wastewater treatment plant, power substation, or [aerodrome](https://www.openstreetmap.org/#map=17/39.03891/-84.65006).
* A road fords a river that is mapped as an area. The `highway=unclassified` road is connected to the riverbank way.
* A boat ramp is connected to a lake that’s mapped as an area. A `leisure=slipway` node connects a `highway=service` way with a `natural=water` area.
* A pedestrian plaza rings the outside of an urban park, with greenspace in the interior. A `highway=pedestrian` multipolygon’s outer way is entirely connected to a `leisure=park` area.
* A sidewalk hugs the edge of a parking lot, and the entire curb is flush with the parking lot, so a wheelchair user can pass from one to the other at any point. A `highway=footway` `footway=sidewalk` way is completely connected to an `amenity=parking` area.
   * Same, but for a pedestrian plaza.
[...]

PR #11200, opened 2025-07-10
add validator about power lines crossing other lines

### LINKED ISSUE #8446: [validator] don't allow or warn if user connect power=line and highway=*
https://www.openstreetmap.org/changeset/102238506
https://www.openstreetmap.org/node/8589623953
don't allow (=if the user clic for the next node of a highway=* on top of a power=line, don't add a shared node but only a node for the highway=*) and/or warn if user connect power=line and highway=*

### LINKED ISSUE #9488: Power lines connected to roads, landuses etc.
### Description

Currently it is very easy to create an extra node on a power line and connect it to anything (like a landuse, road, waterway). In almost all cases, such connections are not desirable.

It would be very nice if such connections are not allowed by default. Alternatively, there can be a warning about it with an option to disconnect and delete the node, or to add some power related tag.

### Screenshots

_No response_

nlehuby:
This PR add a new check in the validators about power lines connected to non power feature.

Capture d’écran du 2025-07-10 15-56-28

Fix #8446 & #9488 

I don't have much experience with javascript development, so please let me know what can be improved.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1828,12 +1828,14 @@ en:
         reference: "Crossing bridges should use different layers."
       bridge-bridge_connectable:
         reference: "Crossing bridges should be connected or use different layers."
+      building-power_connectable:
+        reference: "Power line that enters into a building should be connected using a power terminal."
       indoor-indoor:
         reference: "Crossing indoor features should use different levels."
       indoor-indoor_connectable:
         reference: "Crossing indoor features should be connected or use different levels."
       power-other:
-        reference: "Power lines should not be connected to other features."
+        reference: "Only connect power lines to features that are physically connected in the real world."
     disconnected_way:
       title: Disconnected Ways
       tip: "Find unroutable roads, paths, and ferry routes"
@@ -2000,6 +2002,8 @@ en:
         title: Connect using a crossing
       connect_using_ford:
         title: Connect using a ford
+      connect_using_power_terminal:
+        title: Connect using a power terminal
       continue_from_start:
         title: Continue drawing from start
       continue_from_end:
@@ -2009,8 +2013,8 @@ en:
         annotation: Converted an area to a line.
       delete_feature:
         title: Delete this feature
-      delete_connecting_node:
-        title: Delete the connecting node
+      disconnect_feature:
+        title: Disconnect the feature
       extract_point:
         title: Extract this point
       ignore_issue:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1829,7 +1829,7 @@ en:
       bridge-bridge_connectable:
         reference: "Crossing bridges should be connected or use different layers."
       building-power_connectable:
-        reference: "Power line that enters into a building should be connected using a power terminal."
+        reference: "Use a power terminal to explicitly indicate that the power line enters the building at this location."
       indoor-indoor:
         reference: "Crossing indoor features should use different levels."
       indoor-indoor_connectable:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1786,6 +1786,13 @@ en:
       detached:
         message: "{feature} is too close to {feature2}"
         reference: "Separate points should not share a location."
+    connected_ways:
+      title: Connected Ways
+      message: "{feature} is connected to {feature2}"
+      building-power_connectable:
+        reference: "Use a power terminal to explicitly indicate that the power line enters the building at this location."
+      power-other:
+        reference: "Only connect power lines to features that are physically connected in the real world."
     crossing_ways:
       title: Crossing Ways
       message: "{feature} crosses {feature2}"
@@ -1828,14 +1835,10 @@ en:
         reference: "Crossing bridges should use different layers."
       bridge-bridge_connectable:
         reference: "Crossing bridges should be connected or use different layers."
-      building-power_connectable:
-        reference: "Use a power terminal to explicitly indicate that the power line enters the building at this location."
       indoor-indoor:
         reference: "Crossing indoor features should use different levels."
       indoor-indoor_connectable:
         reference: "Crossing indoor features should be connected or use different levels."
-      power-other:
-        reference: "Only connect power lines to features that are physically connected in the real world."
     disconnected_way:
       title: Disconnected Ways
       tip: "Find unroutable roads, paths, and ferry routes"
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2007,6 +2007,7 @@ en:
         title: Connect using a ford
       connect_using_power_terminal:
         title: Connect using a power terminal
+        annotation: Connected power line to building using a power terminal.
       continue_from_start:
         title: Continue drawing from start
       continue_from_end:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1832,6 +1832,8 @@ en:
         reference: "Crossing indoor features should use different levels."
       indoor-indoor_connectable:
         reference: "Crossing indoor features should be connected or use different levels."
+      power-other:
+        reference: "Power lines should not be connected to other features."
     disconnected_way:
       title: Disconnected Ways
       tip: "Find unroutable roads, paths, and ferry routes"
@@ -2007,6 +2009,8 @@ en:
         annotation: Converted an area to a line.
       delete_feature:
         title: Delete this feature
+      delete_connecting_node:
+        title: Delete the connecting node
       extract_point:
         title: Extract this point
       ignore_issue:
--- a/modules/validations/connected_ways.js
+++ b/modules/validations/connected_ways.js
@@ -384,38 +384,20 @@ export function validationConnectedWays(context) {
             onClick: function(context) {
                 var loc = this.issue.loc;
                 var edges = this.issue.data.edges;
+                var connectingNode = this.issue.data.crossingNode;
 
-                context.perform(
-                    function actionConnectCrossingWays(graph) {
-                        // create the new node for the points
-                        var node = osmNode({ loc: loc, tags: connectionTags });
-                        graph = graph.replace(node);
-
-                        var nodesToMerge = [node.id];
-                        var mergeThresholdInMeters = 0.75;
-
-                        edges.forEach(function(edge) {
-                            var edgeNodes = [graph.entity(edge[0]), graph.entity(edge[1])];
-                            var nearby = geoSphericalClosestNode(edgeNodes, loc);
-                            // if there is already a suitable node nearby, use that
-                            // use the node if node has no interesting tags or if it is a crossing node #8326
-                            if ((!nearby.node.hasInterestingTags() || nearby.node.isCrossing()) && nearby.distance < mergeThresholdInMeters) {
-                                nodesToMerge.push(nearby.node.id);
-                            // else add the new node to the way
-                            } else {
-                                graph = actionAddMidpoint({loc: loc, edge: edge}, node)(graph);
-                            }
-                        });
-
-                        if (nodesToMerge.length > 1) {
-                            // if we're using nearby nodes, merge them with the new node
-                            graph = actionMergeNodes(nodesToMerge, loc)(graph);
-                        }
-
-                        return graph;
-                    },
-                    t('issues.fix.connect_crossing_features.annotation')
-                );
+                if (connectingNode && connectionTags) {
+                    var node = context.graph().entity(connectingNode);
+                    if (node) {
+                        var updatedTags = Object.assign({}, node.tags, connectionTags);
+                        context.perform(
+                            function updateNodeTags(graph) {
+                                return graph.replace(node.update({ tags: updatedTags }));
+                            },
+                            t('issues.fix.connect_using_power_terminal.title')
+                        );
+                    }
+                }
             }
         });
         fix._connectionTags = connectionTags;
--- a/modules/validations/connected_ways.js
+++ b/modules/validations/connected_ways.js
@@ -1,0 +1,428 @@
+import { isEqual } from 'lodash-es';
+
+import { actionAddMidpoint } from '../actions/add_midpoint';
+import { actionMergeNodes } from '../actions/merge_nodes';
+import { operationDisconnect } from '../operations/disconnect';
+import { geoExtent,  geoLineIntersection, geoSphericalClosestNode,  } from '../geo';
+import { osmNode } from '../osm/node';
+import { osmFlowingWaterwayTagValues, osmRailwayTrackTagValues, osmPowerTagValues, osmRoutableAerowayTags, osmRoutableHighwayTagValues } from '../osm/tags';
+import { t } from '../core/localizer';
+import { utilDisplayLabel } from '../util/utilDisplayLabel';
+import { validationIssue, validationIssueFix } from '../core/validation';
+
+
+export function validationConnectedWays(context) {
+    var type = 'connected_ways';
+
+    // returns the way or its parent relation, whichever has a useful feature type
+    function getFeatureWithFeatureTypeTagsForWay(way, graph) {
+        if (getFeatureType(way, graph) === null) {
+            // if the way doesn't match a feature type, check its parent relations
+            var parentRels = graph.parentRelations(way);
+            for (var i = 0; i < parentRels.length; i++) {
+                var rel = parentRels[i];
+                if (getFeatureType(rel, graph) !== null) {
+                    return rel;
+                }
+            }
+        }
+        return way;
+    }
+
+
+    function hasTag(tags, key) {
+        return tags[key] !== undefined && tags[key] !== 'no';
+    }
+
+    // discard
+    var ignoredBuildings = {
+        demolished: true, dismantled: true, proposed: true, razed: true
+    };
+
+
+    function getFeatureType(entity, graph) {
+
+        var geometry = entity.geometry(graph);
+        if (geometry !== 'line' && geometry !== 'area') return null;
+
+        var tags = entity.tags;
+
+        if (tags.aeroway in osmRoutableAerowayTags) return 'aeroway';
+
+        if (hasTag(tags, 'building') && !ignoredBuildings[tags.building]) return 'building';
+        if (hasTag(tags, 'highway') && osmRoutableHighwayTagValues[tags.highway]) return 'highway';
+        if (hasTag(tags, 'landuse')) return 'landuse';
+
+        // don't check railway or waterway areas
+        if (geometry !== 'line') return null;
+
+        if (hasTag(tags, 'railway') && osmRailwayTrackTagValues[tags.railway]) return 'railway';
+        if (hasTag(tags, 'waterway') && osmFlowingWaterwayTagValues[tags.waterway]) return 'waterway';
+
+        if (hasTag(tags, 'power') && osmPowerTagValues[tags.power]) return 'power';
+        //if (hasTag(tags, 'disused:power') && osmPowerTagValues[tags.power]) return 'power';
+        //if (hasTag(tags, 'construction:power') && osmPowerTagValues[tags["construction:power"]]) return 'power';
+
+        return null;
+    }
+
+    function isLegitCrossing(tags1, featureType1, tags2, featureType2) {
+        if (featureType1 === 'power' && featureType2 === 'power') return true;
+        return false;
+    }
+
+    /**
+     * @returns {object | null} the tags for the connecting node, or null if the entities should not be joined
+     */
+    function tagsForConnectionNodeIfAllowed(entity1, entity2, graph) {
+        var featureType1 = getFeatureType(entity1, graph);
+        var featureType2 = getFeatureType(entity2, graph);
+
+        /**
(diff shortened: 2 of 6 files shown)


flacombe:
I completely second this contribution. We highly need this. Thank you @nlehuby!

data/core.yaml
```diff
@@ -1832,6 +1832,8 @@ en:
         reference: "Crossing indoor features should use different levels."
       indoor-indoor_connectable:
         reference: "Crossing indoor features should be connected or use different levels."
+      power-other:
+        reference: "Power lines should not be connected to other features."
```
1ec5: How about something like this, to avoid leaving the impression that power lines can’t connect to other power infrastructure like `power=tower`?

```suggestion
        reference: "Only connect power lines to features that are physically connected in the real world."
```

modules/validations/crossing_ways.js
```diff
@@ -74,6 +75,8 @@ export function validationCrossingWays(context) {
         if (hasTag(tags, 'railway') && osmRailwayTrackTagValues[tags.railway]) return 'railway';
         if (hasTag(tags, 'waterway') && osmFlowingWaterwayTagValues[tags.waterway]) return 'waterway';
 
+        if (hasTag(tags, 'power') && tags.power === 'line') return 'power';
```
1ec5: Like `power=line`, `power=minor_line` and `power=cable` shouldn’t be connected to roadways and land use areas either. However, if we add `power=minor_line`, we’ll need to add an exception for [connections to buildings](https://www.openstreetmap.org/way/1415262985) and possibly other features.

modules/validations/crossing_ways.js
```diff
@@ -485,6 +502,18 @@ export function validationCrossingWays(context) {
                         icon: 'iD-icon-layers',
                         title: t.append('issues.fix.use_different_levels.title')
                     }));
+                } else if (isCrossingPower) {
+                    fixes.push(new validationIssueFix({
+                        icon: 'iD-operation-delete',
+                        title: t.append('issues.fix.delete_connecting_node.title'),
+                        entityIds: [this.data.crossingNode],
+                        onClick: function(context) {
+                            var operation = operationDelete(context, [this.issue.data.crossingNode]);
```
1ec5: This option should disconnect the power line, not delete the connecting node. The connecting node could be an intersection between two roadways or a node that defines the shape of a landuse area. Deleting the node could damage the other connected features’ geometry or topology.

1ec5 (commented):
Thanks for taking the time to implement this feature, and for your patience.

modules/validations/crossing_ways.js
flacombe: We should consider adding [power=terminal](https://wiki.openstreetmap.org/wiki/Tag:power%3Dterminal) on the connecting node with the building.
Other validators like Osmose will warn for an unfinished power line [here](https://www.openstreetmap.org/node/13005124893)

By te way that's really cool to map [service supplies](https://wiki.openstreetmap.org/wiki/Proposal:Service_supplies_proposal)

1ec5: Good idea, I didn’t know about that tag. Looks like openstreetmap/id-tagging-schema#653 is tracking adding a preset for `power=terminal`. Let’s wait until that happens before extending this validation rule to `power=minor_line`. Then we’ll be able to offer a second option to fix the warning by adding `power=terminal` to the connecting node. That would get us a little closer to being able to warn about [islands in the power network](https://community.openstreetmap.org/t/how-to-open-an-id-pr-with-a-suggested-warning-for-disconnected-power-lines/128056) as well.

flacombe:
When seeking for _non-power_ features a power line could be connected to, we should take care of life cycle prefixes.
A power=line connecting to disused:power=line shouldn't raise warning.

That requires to consider *:power=* and power=* as power features (disused:power=line shouldn't connect to roads or landuse as well)

modules/validations/crossing_ways.js
nlehuby: Indeed, you are right.

I've switched to operationDisconnect (instead of operationDelete).

It's a bit annoying because I do think in most case, we want to keep the node in the other feature but we do want to delete the node on the power line... but I didn't manage to do that.

modules/validations/crossing_ways.js
nlehuby: I've also created the PR to add the power=terminal preset : https://github.com/openstreetmap/id-tagging-schema/pull/1637

data/core.yaml
```diff
@@ -1828,10 +1828,14 @@ en:
         reference: "Crossing bridges should use different layers."
       bridge-bridge_connectable:
         reference: "Crossing bridges should be connected or use different layers."
+      building-power_connectable:
+        reference: "Power line that enters into a building should be connected using a power terminal."
```
1ec5: Nit: active voice:

```suggestion
        reference: "Use a power terminal to explicitly indicate that the power line enters the building at this location."
```

modules/validations/crossing_ways.js
1ec5: Yes, that is unfortunate. Maybe we could ticket out a followup improvement that warns about an extraneous vertex along a power line, especially if it causes a kink in the line. It could offer to straighten the line (by removing the node) or add a tower or pole in case it’s missing. This warning might need some extra care, because sometimes the line can branch without a pole, as I see frequently when mapping residential service drops.

modules/validations/crossing_ways.js
1ec5, line 1: By the way, the warning’s title says the power line “crosses” the building, whereas “connects to” would be more accurate. It isn’t a big deal, but maybe the code already allows us to customize the language straightforwardly.

1ec5 (approved):
This looks good now, but we should wait until openstreetmap/id-tagging-schema#1637 lands so that the fix suggestions use supported tags.

modules/validations/crossing_ways.js
flacombe: For the last point about mid span branches, here are two facts:
* It doesn't make a kink the line, the branch occurs in straight line between towers
* Mappers must use power=connection + line_management=branch on the node, so there is always `power` on every nodes of lines.

1ec5: > It doesn't make a kink the line, the branch occurs in straight line between towers

What I mean is that power lines should be straight between towers, but some are very crooked, including most [untouched TIGER-imported power lines](https://www.openstreetmap.org/way/16074038) but also potentially some lines where users have inadvertently added an errant node along the way. We could have a separate validation rule about such issues.

> Mappers must use power=connection + line_management=branch on the node, so there is always `power` on every nodes of lines.

This isn’t guaranteed to be the case before the user clicks on the option to delete the node. We’d need to add another option for adding a `power=connection` `line_management=branch` node to prevent the user from damaging the existing features (quite possibly deleting one of the ways altogether). Moreover, id-tagging-schema has no preset for `power=connection` yet, so we can’t expect the user to know about that prerequisite.

tyrasd (commented):
Hi, sorry for the long wait in getting this reviewed. I noticed some things while trying out the branch and have a few general questions/remarks:

#### :bug: bug when terminal node is already is tagged

When the connection to a building already has tags (e.g. a simple `note`), the fix adds a second vertex at the same coordinates, see below:

https://github.com/user-attachments/assets/9a93ded7-379d-495e-90cc-52e0e1f5fd78

In such cases this specific fix should either be not offered. Or, if it can be done safely, the fix should merge the new tags onto the existing node

#### compatibility with (proposed) preset

https://github.com/openstreetmap/id-tagging-schema/pull/1637 seeks to add `power=terminal` as a preset, which  (currently) has `line_attachment=anchor` as `addTags`. The fix in this PR does not add `line_attachment`, so the user would get a second validation warning about the incompleteness of the node's tags right after applying the fix. This could be addressed by adding the `line_attachment` tag here as well, or by removing `line_attachment` from the preset's `addTags`. (This is probably best to be discussed in the id-tagging-schema PR.)

#### interaction with `layer` tag on building

When a building is tagged with e.g. `layer=1`, the validation warning is not displayed. I assume this is not intentional, but rather a consequence of adding the logic in to `crossing_ways` which already handles `layer` tags for the detection of other data quality issues, and this resulted in an unwanted side effect?

In fact, it might be cleaner to extract this validation into a separate module instead of making the `crossing_ways` validation even more complex than it already is. Any reusable logic could be refactored out into a `util` library or simply be shared between the two modules directly.

#### `power=terminal` on non-end nodes

Is it reasonable to assume `power=terminal` can also be used on nodes that are not a start or end point of a way? The validation fix will happily offer to add the tag also to middle vertices. I have only ever seen it used on the extreme points of a way, never as a middle vertex, but maybe that's a false extrapolation of mine.

#### order of offered fixes

For example, when a power line connects to a building, but continues straight on, it was probably most likely the case that the mapper accidentally snapped the building to a power line (or vice versa). In this case, the first offered fix to add a power terminal is probably not the best to use by an inexperienced user to fix the issue.

Maybe the order of the offered fixes could be optimized to always show the most appropriate fix first. See #11329 for a general discussion around this topic.

#### other tags that can allow power line – building connection

From looking at the [wiki](https://wiki.openstreetmap.org/wiki/Tag%3Apower%3Dterminal), I had the impression that `power=insulator` could technically also be used to correctly tag a connection of a building with a power line. Especially in the case where the way itself does not end at that point, but it e.g. is rather a part of the power line's support structures, that might be even the only correct way to tag such situations.

Are there maybe even more tags that would allow connecting a power line to a non-power feature?

#### other features power lines should not connect to

The validation will only warns when power lines are connected to roads or buildings. But a similarly invalid case would be when e.g. a power line is connected to other structures (e.g. `man_made=bridge`, `natural=water`, `landuse=meadow`, etc.). Would it not make sense to extend this validator to also take those into account? Actually, as power features apparently should only interact with other map features at very specific features with well-documented tags, would it not make sense to warn on all occasions where the connection is between a power line and a non-power feature? Or would that lead in too many false positives?

#### "disconnect" fix results in two nodes at the exact same coordinate

When using the fix to disconnect the features, the result is two nodes with exact same coordinates. That is not ideal for multiple reasons (e.g. it is easy to accidentally snap the nodes together again, it will be flagged by many QA tools, data consumers that only look at the coordinates of the features might still end up with a false connection at that point, etc.). For the normal disconnect operation this is not a problem, as the mapper is already interacting with the particular intersection node and will likely follow up the disconnect operation by a move operation to reorganize the nodes appropriately. But that is not the case when the disconnect is the result of a validation fix.

Maybe an approach like ome ideas to approach this better:

* If power line is almost straight at the to disconnect point: instead offer to remove the node on the power line (the node should remain only on the highway or building).
* Otherwise this means that the power line has a kink at that location, which is strange, as power lines typically can only bend at a power pole or tower (or insulator/portal/terminal). In that case, the best fix might be to offer to create a new power pole or tower next to the intersecting feature. Maybe that's a little overkill, so instead the following could be an alternative:
* Move the disconnected node of the power line a little bit to the side of the road or building.

#### extendability to non-power features

Some other OSM features can also benefit from a validation that checks for suspicious connections, e.g. path and address interpolation (#9589), highway and landuse (#6631), highway and administrative boundary, etc.

It would be awesome if the implementation in this PR could be flexible enough to support for easy extension to other feature combinations like the ones mentioned in the future.

nlehuby:
Thanks for the extensive review.

I started reworking the code to create a dedicated module for validating incorrectly connected ways.

I fixed the following remarks along the way 
- interaction with layer tag on building
- compatibility with the preset
- order of the fixes
- other valid tags for building-power connection
- other features power lines can not connect to (I've only added landuse for now, still to be expanded)
- better wording in the warning message ("is connected to" instead of "crosses")

Now comes the difficult part:
- see if we can reuse the node instead of creating another one in the same position
- see if we can avoid proposing a terminal if it's in the middle of the power line

Knowing that I haven't managed to have the fix to delete the node on the power line while keeping it on the other feature :disappointed:

nlehuby:
I've updated the fixes:
- only propose to add power=terminal if last node of the power feature way
- add power=terminal to the connecting node instead of creating a new one at same position
- the disconnect fix now disconnects the features + removes the node from the power feature.
I chose to always delete the node, without trying to calculate whether it significantly altered the geometry of the way, because I feel that most of the time it is a connection error and not an omission of a power tower or other power support. But there is always the possibility of doing a more detailed check, as proposed by tyrasd if it is a little too drastic.

Apart from this suggestion, I believe I have addressed all the comments made in the various reviews :)

nlehuby:
Hello,

Would it be possible to have another code review on this proposal? I would very much like to see this in iD.

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+        if (hasTag(tags, 'waterway') && osmFlowingWaterwayTagValues[tags.waterway]) return 'waterway';
+
+        if (hasTag(tags, 'power') && osmPowerTagValues[tags.power]) return 'power';
+        //if (hasTag(tags, 'disused:power') && osmPowerTagValues[tags.power]) return 'power';
+        //if (hasTag(tags, 'construction:power') && osmPowerTagValues[tags["construction:power"]]) return 'power';
+
+        return null;
+    }
+
+    function isLegitCrossing(tags1, featureType1, tags2, featureType2) {
```
hlfan: You define tags here, but don't use them. And `isLegitConnection` might be a better name here.

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+            // if the way doesn't match a feature type, check its parent relations
+            var parentRels = graph.parentRelations(way);
+            for (var i = 0; i < parentRels.length; i++) {
+                var rel = parentRels[i];
+                if (getFeatureType(rel, graph) !== null) {
+                    return rel;
+                }
+            }
+        }
+        return way;
```
hlfan:
```suggestion
        if (getFeatureType(way, graph) !== null) return way;
        // if the way doesn't match a feature type, check its parent relations
        for (const rel of graph.parentRelations(way)) {
            if (getFeatureType(rel, graph) !== null) return rel;
        }
```

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+                ]
+            ]);
+
+            // Optimize by only checking overlapping segments, not every segment
+            // of overlapping ways
+            segmentInfos = tree.waySegments(extent, graph);
+
+            for (j = 0; j < segmentInfos.length; j++) {
+                segment2Info = segmentInfos[j];
+
```
hlfan:
```suggestion
            for (const segment2Info of tree.waySegments(extent, graph)) {
```

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+                                edge: [n1.id, n2.id]
+                            },
+                            {
+                                way: way2,
+                                featureType: way2FeatureType,
+                                edge: [nA.id, nB.id]
+                            }
+                        ],
+                        crossPoint: point,
+                        crossIndex: way1NodesId.indexOf(intersectingNode),
```
hlfan: Since this is only called once per issue, I don't think it's worth it to precompute `way1NodesId`.
```suggestion
                        crossIndex: way1Nodes.findIndex(n => n.id === intersectingNode),
```

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+        } else if (entity.type === 'relation') {
+            return entity.members.reduce(function(array, member) {
+                if (member.type === 'way' &&
+                    // only look at geometry ways
+                    (!member.role || member.role === 'outer' || member.role === 'inner')) {
+                    var entity = graph.hasEntity(member.id);
+                    // don't add duplicates
+                    if (entity && array.indexOf(entity) === -1) {
+                        array.push(entity);
+                    }
```
hlfan: Using a `Set` might be the better option here.

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+        // declare these here to reduce garbage collection
+        var wayIndex, connectionsIndex, connections;
+        for (wayIndex in ways) {
+            connections = findConnectionOnWay(ways[wayIndex], graph, tree);
+            for (connectionsIndex in connections) {
+                issues.push(createIssue(connections[connectionsIndex], graph));
+            }
+        }
+        return issues;
+    };
```
hlfan:
```suggestion
    let validation = function (entity, graph) {
        const tree = context.history().tree();
        const issues = [];
        for (const way of waysToCheck(entity, graph)) {
            for (const connection of findConnectionOnWay(way, graph, tree)) {
                issues.push(createIssue(connection, graph));
            }
        }
        return issues;
    };
```

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+
+        var connectionTags = tagsForConnectionNodeIfAllowed(entities[0], entities[1], graph);
+
+        var featureType1 = wrongConnection.wayInfos[0].featureType;
+        var featureType2 = wrongConnection.wayInfos[1].featureType;
+
+
+        var isCrossingPower = hasTag(entities[0].tags, 'power') || hasTag(entities[1].tags, 'power');
+
+        var subtype = [featureType1, featureType2].sort().join('-');
```
hlfan:
```suggestion
        const edges = wrongConnection.wayInfos.map(w => w.edge);
        const featureTypes = wrongConnection.wayInfos.map(w => w.featureType);

        const connectionTags = tagsForConnectionNodeIfAllowed(entities[0], entities[1], graph);

        const isCrossingPower = entities.some(e => hasTag(e.tags, 'power'));

        const subtype = featureTypes.sort().join('-');
```

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+
+                fixes.push(new validationIssueFix({
+                    icon: 'iD-operation-delete',
+                    title: t.append('issues.fix.disconnect_feature.title'),
+                    entityIds: [this.data.connectingNode],
+                    onClick: function(context) {
+                        var graph = context.graph();
+                        var entity1 = graph.entity(this.issue.entityIds[0]);
+                        var entity2 = graph.entity(this.issue.entityIds[1]);
+                        var powerEntity = hasTag(entity1.tags, 'power') ? entity1 : hasTag(entity2.tags, 'power') ? entity2 : null;
```
hlfan:
```suggestion
                        const powerEntity = this.issue.entityIds.map(id => graph.entity(id)).find(e => hasTag(e.tags, 'power'));
```

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+                        fixes.push(makeConnectWaysFix(lessLikelyConnectionTags));
+                    }
+                }
+
+
+                // repositioning the features is always an option
+                fixes.push(new validationIssueFix({
+                    icon: 'iD-operation-move',
+                    title: t.append('issues.fix.reposition_features.title')
+                }));
```
hlfan: Is repositioning the features always an option here? I don't think so.

modules/validations/connected_ways.js
```diff
@@ -0,0 +1,428 @@
...
+                    var node = context.graph().entity(connectingNode);
+                    if (node) {
+                        var updatedTags = Object.assign({}, node.tags, connectionTags);
+                        context.perform(
+                            function updateNodeTags(graph) {
+                                return graph.replace(node.update({ tags: updatedTags }));
+                            },
+                            t('issues.fix.connect_using_power_terminal.title')
+                        );
+                    }
```
hlfan:
```suggestion
                    if (!node) return;
                    const tags = {...node.tags, ...connectionTags};
                    context.perform(
                        graph => graph.replace(node.update({ tags } )),
                        t('issues.fix.' + fixTitleID + '.annotation')
                    );
```

hlfan (commented):
In general, newly written code should use `let` and `const` instead of `var`.
Using more syntactical sugar also helps to make the logic clearer.

modules/validations/connected_ways.js
nlehuby: :thinking: is it ? I will need to convert it to an array in the end right ?

modules/validations/connected_ways.js
nlehuby: This creates unsafe reference to the intersectingNode variable

nlehuby:
Thanks a lot for the review @hlfan , I've updated all the var and try to improve the code as suggested.

modules/validations/connected_ways.js
hlfan: Yes, but you can do that after the loop (`reduce`) and take advantage of the faster implementation. With this, for many large relations, the execution time scales linearly instead of quadratically.

modules/validations/connected_ways.js
hlfan: Not if you declare `intersectingNode` lazily in the loop when only the intersections remain:
```js
                const intersectingNode = [nAId, nBId].find(id => id === n1.id || id === n2.id);

```

nlehuby: ok, it makes sense, thanks !

PR #11216, opened 2025-07-16
Add warning when road type is replaced with `highway=construction` without `construction=...` tag

### LINKED ISSUE #11278: iD makes it very easy to remove the road type information when setting `highway=construction`
## Summary

It is currently very easy with iD to remove the road type information by changing a `highway=xyz` way to `highway=construction` without adding `construction=xyz`. The most common cause of this seems to be that a mapper selects the "road under construction" preset, but it is also possible to achive the same result by editing the tags directly. This issue occurs frequently (see numbers section), and is likely in almost all cases not the mapper's intention.

When `construction=xyz` tag is not set, the road type information is lost from the tags. While humans are usually pretty good at identifying the lost road type information from context or the edit history, this is difficult for data users. Rendering applications like OSM-Carto display such roads as generic roads under construction. StreetComplete resets the road type to `highway=road` when a user determines that construction has finished, unless a `construction=xyz` tag was present.

## How common is this?

I analysed roads that were changed to `highway=construction` that did have a road type before. I distinguish 3 cases:

 - ✅ A: the road type was preserved in the change (`construction=xyz` was set in  the same changeset)
 - ✅ B: `construction=xyz` is present, but `xyz` is a different road type than before (probably an intentional change by the mapper)
 - ⚠️ C: no `construction=...` tag was set in the changeset, road type is lost

Of the third category, where the road type was lost, I considered four sub-cases depending on whether the road type was later restored. In all of those cases, a mapper did retstore the road type in a later changeset (or will need to do so in the future):

1. Road type restored when construction ended (via `highway=xyz`)
2. Road type restored while still in construction (via `construction=xyz`)
3. Road type unknown, road still in construction (`highway=construction` without `construction` key)
4. The road type was reset, after construction has ended (reset to `highway=road`, usually by StreetComplete)

From looking at changesets, Case 1 often are highly active mappers, which edit tags directly and simply restore the road type when construction has finished. Often the initial change and the fix change are done by the same mapper. In the second case, usually some other mapper recognizes the lost road type (e.g. due to the unusual Carto renderig) and fixes the road type. Case 3 usually means that no one has cared enough to restore the road type. This appears to occur often in areas with low mapper density, where issues like this one aren't catched as fast. In the fourth case, most of the time, some mapper using StreetComplete determined construction to be finished, which means that in absence of road type information `highway=road` is used as a fallback.

Note that all cases in C are problematic. In each case, the road type is initially lost, requiring an additional changeset by a mapper. This fix happens pretty often, either combined with other changes by the original mapper (good-ish) or by someone else (bad, requires extensive changeset comments explaining that road type should be kept, often weeks/months/years later). However, a substantial proportion of the changes where the road type is lost initially are not fixed for a long time.

### The numbers

All numbers are for roads set to highway=construction in 2024 in Germany.

All iD edits:

| Category                                 |   Count               |
|------------------------------------------|-----------------------|
| all highway=construction changes         |  2431 (100.0% of all) |
|  -> road type preserved unchanged        |  1320 ( 54.3% of all) |
|  -> road type preserved changed          |   120 (  4.9% of all) |
|  -> road type lost                       |   991 ( 40.8% of all) |
|  ->   -> restored at construction end      |   536 ( 22.0% of all) |
|  ->   -> restored while in construction    |   250 ( 10.3% of all) |
|  ->   -> still without road type           |   163 (  6.7% of all) |
|  ->   -> reset to highway=road             |    42 (  1.7% of all) |

All non-iD edits:

| Category                                 |   Count               |
|------------------------------------------|-----------------------|
| all highway=construction changes         |  3413 (100.0% of all) |
|  -> road type preserved unchanged        |  2514 ( 73.7% of all) |
|  -> road type preserved changed          |   329 (  9.6% of all) |
|  -> road type lost                       |   570 ( 16.7% of all) |
|  ->   -> restored at construction end      |   177 (  5.2% of all) |
|  ->   -> restored while in construction    |   259 (  7.6% of all) |
|  ->   -> still without road type           |    96 (  2.8% of all) |
|  ->   -> reset to highway=road             |    38 (  1.1% of all) |

iD edits of new mappers (`changesets_count` < 100):

| Category                                 |   Count               |
|------------------------------------------|-----------------------|
| all highway=construction changes         |   495 (100.0% of all) |
|  -> road type preserved unchanged        |    86 ( 17.4% of all) |
|  -> road type preserved changed          |    17 (  3.4% of all) |
|  -> road type lost                       |   392 ( 79.2% of all) |
|  ->   -> restored at construction end      |   170 ( 34.3% of all) |
|  ->   -> restored while in construction    |   156 ( 31.5% of all) |
|  ->   -> still without road type           |    61 ( 12.3% of all) |
|  ->   -> reset to highway=road             |     5 (  1.0% of all) |

### Conclusions

The numbers are not all included here, but I also checked some other regions (Spain, Poland, California, India), which have mostly similar results:

 - Of all edits where road type information is lost, a **majority is done with iD** (usually about 60%-70%)
 - As a percentage of their edits, **iD mappers are much more likely** than other mappers to  **remove the road type** (~40-ish% of edits, compared to ~15ish% for other editors -- India is an exception here)
 - Compared to more experienced iD mappers, **less experienced mappers are much more likely to remove the road type** in their edits, often in 80%-90% of their `highway=construction` edits!

### Keep fixing manually (status quo)

- wastes contributor time
- often "too late" (e.g. months/years after road type info was lost)

### Fix suggestions / validation in editors

(see #11216 for an suggestion for how to implement this)

 - should be easy to implement
 - teaches users that road type should be kept
 - also applies to direct tag edits
 - cannot address non-iD changes (but this is out of scope for this repo here)

### In iD Tagging scheme?

Actually, applying the construction preset should generally preserve the road type. Can this be expressed in the iD presets and somehow fixed there? (Pro: would help other editors that use the presets)

 - only applies to changes via presets (not direct tag changes)

### Create a fixup-bot that restores the information automatically from tag history

 - fix up edits directly after they have happened
 - does not prevent the "bad" edits
 - does not teach users that the road type should be retained

burrscurr:
## Summary

Add a warning when a `highway=<road_type>` is replaced with `highway=construction` without a `construction=<road_type>` tag. The warning comes with a fix that allows to restore the previous road type in the `construction` key. For now, the warning is only created when it can be fixed automatically, that is, when the unedited way had a `highway!=road` tag.

Possible follow-ups:
- Investigate whether the `construction=...` tag can be set when the "road under construction" preset is applied
- add a warning (perhaps a suggestion?) to set `construction=...` key for newly added ways

### Context

Context: This can happen when a user selects the "road under construction" preset or manually replaces the `highway` tag with `construction`. If the previous road type is not restored in the `construction=...` tag, the information which road type is being constructed is being lost in the tagging. From looking at affected changesets, it seems that typically, iD is used and it appears that the removal of the road type information was unintentional. Under these circumstances, iD should issue a warning. Data consumers shouldn't be expected to interpret tags based on the edit history.

To mitigate the consequences of removed road type information, mappers and data consumers would need to analyze the way history. For data consumers, such interpolations are often not feasible. StreetComplete for instance, will set `highway=road` in its "is this construction finished" quest if there is no `construction=...` key available (which is a reasonable default to assume based on exiting tags). 

There are many examples of `highway=road`s that were created as a result of a manual change `

| road type information lost (usually with iD) | Reset to `highway=road` (usually with SC) | Notes |
|--------------------------|-------------------------|-------------|
| https://www.openstreetmap.org/way/314491796/history/12 | https://www.openstreetmap.org/way/314491796/history/14 | `highway=secondary` before |
| https://www.openstreetmap.org/way/915874283/history/3 | https://www.openstreetmap.org/way/915874283/history/5 | `highway=footway` before |
| https://www.openstreetmap.org/way/375991058/history/13 | https://www.openstreetmap.org/way/375991058/history/14 | |

Overpass Link: https://overpass-turbo.eu/s/27Xa (`surface` filter helps to select roads that have a more rich history and are thus more likely to be true positives)

## Test Plan

Added some unit tests.

Until i figure out how to do this in tests, i manually verified that:

 - changing an existing `highway=xyz` to `highway=construction` triggers the warning and that the fix adds `construction=xyz`
 - no warning on existing `highway=construction` + `construction=xyz`
 - no warning for existing `highway=construction` without `construction`

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2042,6 +2042,8 @@ en:
         title: Remove the name
       reposition_features:
         title: Reposition the features
+      restore_road_type:
+        title: Restore previous road type
       reverse_feature:
         title: Reverse this feature
       select_preset:
--- a/modules/validations/missing_tag.js
+++ b/modules/validations/missing_tag.js
@@ -62,26 +62,60 @@ export function validationMissingTag(context) {
         }
 
         // flag an unknown road even if it's a member of a relation
-        if (!subtype && isUnknownRoad(entity)) {
-            subtype = 'highway_classification';
+        if (!subtype) {
+            if (isUnknownRoad(entity)) {
+                subtype = 'highway_classification';
+            }
         }
 
         if (!subtype) return [];
 
-        var messageID = subtype === 'highway_classification' ? 'unknown_road' : 'missing_tag.' + subtype;
-        var referenceID = subtype === 'highway_classification' ? 'unknown_road' : 'missing_tag';
-
         // can always delete if the user created it in the first place..
-        var canDelete = (entity.version === undefined || entity.v !== undefined);
-        var severity = (canDelete && subtype !== 'highway_classification') ? 'error' : 'warning';
+        var userCreatedEntity = (entity.version === undefined || entity.v !== undefined);
+
+        // If tags are missing, display a warning with a dynamic fix offering to select a
+        // (new) preset.
+        let messageID = `issues.missing_tag.${subtype}.message`;
+        let referenceID = 'issues.missing_tag.reference';
+        let mainFixLabel = 'issues.fix.select_preset.title';
+        let mainFixAction = function(context) {
+            context.ui().sidebar.showPresetList();
+        };
+        let mainFixIcon = 'iD-icon-search';
+        let severity = 'warning';
+        // Whether the warning should contain a fix that will (attempt) to delete
+        // the entity with missing tags.
+        let offerDeletionFix = true;
+
+        switch (subtype) {
+            // An entity lacks tags to make it relevant, for example:
+            // - no tags at all (and not a way in a relation or a node in a way)
+            // - a node without tags that is not part of a way)
+            // - a relation without a type=... tag
+            case 'any':
+            case 'descriptive':
+            case 'relation_type':
+                if (userCreatedEntity) {
+                    severity = 'error';
+                }
+                break;
+            // A way tagged as highway=road (a more precise highway value should be
+            // used).
+            case 'highway_classification':
+                messageID = 'issues.unknown_road.message';
+                referenceID = 'issues.unknown_road.reference';
+                mainFixLabel = 'issues.fix.select_road_type.title';
+                break;
+        }
 
         return [new validationIssue({
             type: type,
             subtype: subtype,
             severity: severity,
+            // depends on messageID, t (localizer)
             message: function(context) {
                 var entity = context.hasEntity(this.entityIds[0]);
-                return entity ? t.append('issues.' + messageID + '.message', {
+                return entity ? t.append(messageID, {
                     feature: utilDisplayLabel(entity, context.graph())
                 }) : '';
             },
@@ -91,40 +125,36 @@ export function validationMissingTag(context) {
 
                 var fixes = [];
 
-                var selectFixType = subtype === 'highway_classification' ? 'select_road_type' : 'select_preset';
-
                 fixes.push(new validationIssueFix({
-                    icon: 'iD-icon-search',
-                    title: t.append('issues.fix.' + selectFixType + '.title'),
-                    onClick: function(context) {
-                        context.ui().sidebar.showPresetList();
-                    }
+                    icon: mainFixIcon,
+                    title: t.append(mainFixLabel),
+                    onClick: mainFixAction
                 }));
 
-                var deleteOnClick;
+                if (offerDeletionFix) {
+                    var deleteOnClick;
+                    var id = this.entityIds[0];
+                    var operation = operationDelete(context, [id]);
+                    var disabledReasonID = operation.disabled();
+                    if (!disabledReasonID) {
+                        deleteOnClick = function(context) {
+                            var id = this.issue.entityIds[0];
+                            var operation = operationDelete(context, [id]);
+                            if (!operation.disabled()) {
+                                operation();
+                            }
+                        };
+                    }
 
-                var id = this.entityIds[0];
-                var operation = operationDelete(context, [id]);
-                var disabledReasonID = operation.disabled();
-                if (!disabledReasonID) {
-                    deleteOnClick = function(context) {
-                        var id = this.issue.entityIds[0];
-                        var operation = operationDelete(context, [id]);
-                        if (!operation.disabled()) {
-                            operation();
-                        }
-                    };
+                    fixes.push(
+                        new validationIssueFix({
+                            icon: 'iD-operation-delete',
+                            title: t.append('issues.fix.delete_feature.title'),
+                            disabledReason: disabledReasonID ? t('operations.delete.' + disabledReasonID + '.single') : undefined,
+                            onClick: deleteOnClick
+                        })
+                    );
                 }
-
-                fixes.push(
-                    new validationIssueFix({
-                        icon: 'iD-operation-delete',
-                        title: t.append('issues.fix.delete_feature.title'),
-                        disabledReason: disabledReasonID ? t('operations.delete.' + disabledReasonID + '.single') : undefined,
-                        onClick: deleteOnClick
-                    })
-                );
-
                 return fixes;
             }
         })];
@@ -135,7 +165,7 @@ export function validationMissingTag(context) {
                 .enter()
                 .append('div')
                 .attr('class', 'issue-reference')
-                .call(t.append('issues.' + referenceID + '.reference'));
+                .call(t.append(referenceID));
         }
     };
 
--- a/modules/validations/missing_tag.js
+++ b/modules/validations/missing_tag.js
@@ -1,3 +1,4 @@
+import { actionChangeTags } from '../actions/change_tags';
 import { operationDelete } from '../operations/delete';
 import { osmIsInterestingTag } from '../osm/tags';
 import { t } from '../core/localizer';
@@ -34,6 +35,19 @@ export function validationMissingTag(context) {
         return entity.type === 'way' && entity.tags.highway === 'road';
     }
 
+    // There was a highway=<road_type> tag that was changed to highway=construction without
+    // adding a construction=<road_type> tag.
+    function hadRoadTypeChangedToConstruction(context, entity) {
+        let origGraph = context.history().base();
+        if (!origGraph.hasEntity(entity.id)) {
+            return false;
+        }
+        const origTags = origGraph.entity(entity.id).tags;
+        const wasKnownType = 'highway' in origTags && origTags.highway !== 'road' && origTags.highway !== 'construction';
+        const isUnknownType = entity.tags.highway === 'construction' && !('construction' in entity.tags);
+        return entity.type === 'way' && wasKnownType && isUnknownType;
+    }
+
     function isUntypedRelation(entity) {
         return entity.type === 'relation' && !entity.tags.type;
     }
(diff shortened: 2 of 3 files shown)


modules/validations/missing_tag.js
```diff
@@ -62,26 +62,60 @@ export function validationMissingTag(context) {
...
+                }
+                break;
+            // A way tagged as highway=road (a more precise highway value should be
+            // used).
+            case 'highway_classification':
+                messageID = 'issues.unknown_road.message';
+                referenceID = 'issues.unknown_road.reference';
+                mainFixLabel = 'issues.fix.select_road_type.title';
+                break;
+        }
```
burrscurr, line 145: General idea of this restructuring: list defaults for missing_tags validations, and select specific overrides for the different warning subtypes.

data/core.yaml
```diff
@@ -2042,6 +2042,8 @@ en:
         title: Remove the name
       reposition_features:
         title: Reposition the features
+      restore_road_type:
+        title: Restore previous road type
```
burrscurr, line 2046: Is it possible to add translations within this PR? Or will this need to be done separately after a merge?

modules/validations/missing_tag.js
```diff
@@ -34,6 +35,19 @@ export function validationMissingTag(context) {
         return entity.type === 'way' && entity.tags.highway === 'road';
     }
 
+    // There was a highway=<road_type> tag that was changed to highway=construction without
+    // adding a construction=<road_type> tag.
+    function hadRoadTypeChangedToConstruction(context, entity) {
+        let origGraph = context.history().base();
+        if (!origGraph.hasEntity(entity.id)) {
+            return false;
+        }
```
burrscurr, line 44: I needed to add this to make the tests pass, but I don't really understand why. Does this catch the case where the user creates a new way that did not previously exist?

modules/validations/missing_tag.js
```diff
@@ -62,26 +76,82 @@ export function validationMissingTag(context) {
         }
 
         // flag an unknown road even if it's a member of a relation
-        if (!subtype && isUnknownRoad(entity)) {
-            subtype = 'highway_classification';
+        if (!subtype) {
+            if (isUnknownRoad(entity)) {
+                subtype = 'highway_classification';
+            } else if (hadRoadTypeChangedToConstruction(context, entity)) {
+                subtype = 'highway_classification_construction';
```
burrscurr, line 83: Perhaps there is a more descriptive and concise name for what this warning is?

modules/validations/missing_tag.js
```diff
@@ -62,26 +76,82 @@ export function validationMissingTag(context) {
...
+            case 'highway_classification_construction':
+                messageID = 'issues.unknown_road.message';
+                referenceID = 'issues.unknown_road.reference';
+                mainFixLabel = 'issues.fix.restore_road_type.title';
+                mainFixIcon = 'iD-icon-undo';
+                offerDeletionFix = false;
+                mainFixAction = function(context) {
+                    // We checked in hadRoadTypeChangedToConstruction() that a highway tag
+                    // did exist before the edit and can thus just collect it here.
+                    const origRoadType = context.history().base().entity(entity.id).tags.highway;
```
burrscurr, line 136: This is quite a long call, most of it was already checked in the `hasRoadTypeChangedToConstruction` function -- how to catch errors here?

modules/validations/missing_tag.js
```diff
@@ -62,26 +76,82 @@ export function validationMissingTag(context) {
...
+                    let newTags = Object.assign({}, entity.tags);   // shallow copy
+                    newTags.construction = origRoadType;
+                    context.perform(
+                        actionChangeTags(entity.id, newTags),
+                        t('operations.change_tags.annotation')
+                    );
```
burrscurr, line 142: I copied this over and adjusted it from a different validation. Hope this fits.

test/spec/validations/missing_tag.js
```diff
@@ -122,4 +122,18 @@ describe('iD.validations.missing_tag', function () {
...
+    });
+
+    // Arguably, this should be a warning in the future, but for now, just updated ways are flagged.
+    it('ignores newly created highway=construction without construction key', function() {
+        createWay({ highway: 'construction' });
+        var issues = validate();
+        expect(issues).to.have.lengthOf(0);
+    });
+
+    // TODO: Add test that updates an existing road and creates a warning
```
burrscurr, line 138: I would like to add tests where existing data is modified to trigger the warning. The setup in this file seems to just assume the created way was a user edit. How can I set up data as if it was pre-existing OSM data?

burrscurr (commented):
I think this should reasonably be ready for a initial review. I'll need to add some more tests, which I currently don't know how to do. (See comment.)

(Note: I'm a JS newbie so please excuse the probably pretty non-idiomatic style)

tyrasd:
#10844 will add a UI to change features between different lifecycle statuses, including from `highway=XYZ` to `highway=construction + construction=XYZ` (and also vice versa). Do you think there is still merit to include this validation once the lifecycle feature is implemented in iD? It would still trigger when people use the raw tag editor, but I would have thought that these users are not the main target of this PR.

burrscurr:
#10844 looks very helpful! If it provides a better way of changing `highway=xyz` to "under construction", that would be great! I suspect that currently, in most cases, the road type information is lost due to selecting the "road under construction" preset. If #10844 makes this more unlikely or even replaces this option, that would be great.

I also suspect that there are some mappers who set `highway=construction` as a raw tag directly. How common is this? Difficult to tell. If I get around to doing this, I'll try to quantify how often the road type information is lost. This could help to measure whether #10844 is (sufficiently) effective at preventing this. Should I open an issue to document the results and the general problem this PR tries to solve – to keep this PR clean?

> Do you think there is still merit to include this validation once the lifecycle feature is implemented in iD?

I think a "road type information lost" warning has a very good signal:noise ratio and is easy to implement. It is difficult to think of cases where removing the road type is accurate and reflects the user's intentions – perhaps if construction is ongoing, but it is unforeseeable which type of road will be constructed? So, from my perspective, it appears like a worthwhile addition nonetheless.

burrscurr:
For what it's worth, the issue of lost road type information and the fallback to a generic value after finished construction has just come up in a discussion in the German forum: https://community.openstreetmap.org/t/streetscomplete-setzt-highway-road/132873. (This does not really help to answer the question whether #10844 can solve this issue, but it is evidence that this is a real-world issue that is affecting mappers and general data quality. The linked example in the post is of a [pretty new iD mapper that changed](https://www.openstreetmap.org/changeset/167165927) a `highway=footway` to `highway=construction`.)

tyrasd:
> I'll try to quantify how often the road type information is lost. […] Should I open an issue to document the results and the general problem this PR tries to solve – to keep this PR clean?

Sound great, feel free to post it here, or open a new issue. Whatever you think fits best.

burrscurr:
Numbers available in #11278!

TLDR: the road type is frequently lost when setting highway=construction in iD, on average in about 40% of edits; newer mappers remove the road type in ~80-90% of cases.

(exact numbers depend on region, but are fairly consistent)

data/core.yaml
matkoniecz: translations are handled separately

modules/validations/missing_tag.js
matkoniecz: `highway_construction_type_missing` ?

PR #11222, opened 2025-07-17
Fix tile edge blending

hlfan:
Since I came across Leaflet's tiling discussion and 9e110cfda41787aeda5547d6b37e5fa56330587c again:

Currently, tiles later in the DOM spill over earlier ones with the extruded edges due to the increased opacity.
Actually, you'd want to tile the layer first and only apply the filter on everything together.
That way, the edges mix with a more correct proportion (still not perfect, but much better).

The impact is pretty difficult to spot; I've had to pick a [layer with very high local contrast](https://github.com/osmlab/editor-layer-index/blob/gh-pages/sources/europe/at/ViennaMehrzweckkarte-situation.geojson) and do A-B testing constantly to occasionally see an affected pixel.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4382,8 +4382,8 @@ img.tile {
     transition: opacity 1ms linear;
 }
 
-.layer-background img.tile,
-.map-in-map-background img.tile {
+.layer-background,
+.map-in-map-background {
     filter: url(#alpha-slope5);
 }
 


hlfan:
Here is an example at https://ideditor.netlify.app/#map=20.20/48.19974/16.36746&disable_features=address_points,points,traffic_roads,service_roads,paths,buildings,building_parts,indoor,landuse,boundaries,water,rail,pistes,aerialways,power,past_future,others&background=none&overlays=wien.gv.at-gp-situation:

| tile-wise filter | layer spanning filter |
|--------|--------|
| image | image |

tyrasd (approved):
Thanks, this makes sense. The difference really _is_ subtle for most layers (e.g. the example in the original issue #10589 is indistinguishable), but let's do this properly! 😊

hlfan:
> let's do this properly

Sooo, about that, this filter location gets overwritten by the _Display Options_ filters.

hlfan:
@tyrasd Sorry about this...

tyrasd:
oh, true, I also forgot about that. It's now fixed by applying the alpha filter to the inner `.layer` div instead.

PR #11239, opened 2025-07-22
sync user preferences from the openstreetmap API

### LINKED ISSUE #11305: Preferences in iD Are Not Synchronized Across Devices
### Description

Currently, user preferences in the iD editor are stored only in the browser’s local storage. This limits them to a single device and browser profile. If a user switches devices, uses a different browser, or clears their cache, all their settings such as background imagery, overlay selections, panel positions, and validation rules are lost and must be manually reconfigured.

This results in an inconsistent editing experience, especially for contributors who map from multiple devices or participate in mapping events using shared equipment. To address this, preferences should be stored on the OpenStreetMap server and associated with the user’s account. When the user logs in, iD should retrieve these preferences from the `/api/0.6/user/preferences.json` endpoint and apply them automatically.

Solving this will require adding features in iD to read preferences from the API when authentication completes, apply them immediately in the editor, and push updated preferences back to the API whenever they are changed locally. There will also need to be a mechanism for handling offline edits by queuing preference updates and syncing them once a connection is restored.

### Screenshots

_No response_

Deeptanshu-sankhwar:
## Technical Changes

- **modules/core/context.js**
  - On initialization, sets the OSM connection for the preferences module.
  - Listens for the `authDone.preferences` event on the OSM connection to trigger a sync of user preferences from the server after authentication.
  - If already authenticated on load, immediately syncs preferences from the server.

- **modules/core/preferences.js**
  - Adds `setOsmConnection(connection)` to allow the preferences module to use the current OSM connection.
  - Implements `syncWithServer()` to:
    - Fetch all user preferences from the OSM API (`/api/0.6/user/preferences.json`) and replaces them into local storage.
    - Push any pending local changes to the server.
  - Updates the core preferences setter to:
    - Immediately push changes to the server via `PUT` or `DELETE` requests.
    - Queue changes if the OSM connection is unavailable, and retry them on the next sync.
  - Notifies registered listeners when preferences are updated from the server.

- **test/spec/validations/**
  - Updates tests to mock the OSM connection and verify the new preference sync logic.
  
  Closes #11305


matkoniecz:
> @[Deeptanshu-sankhwar](https://github.com/Deeptanshu-sankhwar) Deeptanshu-sankhwar marked this pull request as draft [6 months ago](https://github.com/openstreetmap/iD/pull/11239#event-19133039615)

do you plan to finish this PR?

EDIT: @tyrasd wanted to look at it

PR #11240, opened 2025-07-22
run tests on node v22-24, remove v20

k-yle:
node v20 [is in maintenance mode](https://nodejs.org/en/about/previous-releases) now

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


tyrasd (approved):
Thanks. Initially, I was wondering if it might be better to keep supporting node's LTS versions as long as they are under maintenance. But to be frank, I believe only few people would be actually benefit from that, as systems that ship with node 20 seem to be rare (e.g. ubuntu's LTS version 24.04 only has node 18).

PR #11253, opened 2025-07-26
make the imagery denylist RegExs case-insensitive

k-yle:
Currently you can just bypass the denylist by making one character uppercase...


tyrasd (approved):
good catch, thanks!

PR #11257, opened 2025-07-28
Prevent direction rendering on outbound one-ways from all-way stop signs

### LINKED ISSUE #9879: all way stop icon vs. one way streets
### URL

https://www.openstreetmap.org/edit?node=235431065#map=20/42.15600/-87.77108

Shouldn't there only be three dark cones show for this "all way" stop,
being that one of the ways is one-way outbound only?
Just like there are three white bars on the pavement in the photo, not four.

20230903T072811

_No response_

Chrome

zandonella:
Fixes #9879

Description:
All-way stop signs were incorrectly showing direction cones on outbound one-way roads, which implied incoming traffic where none would be coming from.

Changes:
Updated the `.filter()` logic in the `osmNode.directions()` method to skip outbound one-way roads when the current node has the `highway=stop` tag.

Notes:
This is scoped specifically to stop signs to match the original issue, but this logic could be extended to other node types through a list of node types to check or by removing the stop sign check entirely (which would make every node have this logic). 

I implemented this within `.filter()` instead of `osmShouldRenderDirection()` because I was checking the first node in the one-way to determine if it was outbound or not, and accessing that value within `osmShouldRenderDirection()` would require refactoring the function. If there is another way to check if it is outbound, or if refactoring the `osmShouldRenderDirection()` would be more desirable, then I'd be happy to try another approach!

Original issue:
image

Fix:
image

### DIFF
--- a/modules/osm/node.js
+++ b/modules/osm/node.js
@@ -120,7 +120,13 @@ const prototype = {
 
             var nodeIds = {};
             resolver.parentWays(this)
-                .filter(way => osmShouldRenderDirection(this.tags, way.tags))
+                .filter(way => {
+                    // check for outbound one-way ways from stop signs
+                    const isOutbound = way.tags.oneway === 'yes' && way.nodes[0] === this.id;
+                    const isStopSign = this.tags.highway === 'stop';
+                    if (isOutbound && isStopSign) return false;
+                    return osmShouldRenderDirection(this.tags, way.tags);
+                })
                 .forEach(function(parent) {
                     var nodes = parent.nodes;
                     for (i = 0; i < nodes.length; i++) {
--- a/modules/osm/node.js
+++ b/modules/osm/node.js
@@ -1,7 +1,7 @@
 import { osmEntity } from './entity';
 import { geoAngle, geoExtent } from '../geo';
 import { utilArrayUniq } from '../util';
-import { osmShouldRenderDirection } from './tags';
+import { osmShouldRenderDirection, hasPartialOneway } from './tags';
 
 export const cardinal = {
     north: 0,               n: 0,
@@ -120,17 +120,27 @@ const prototype = {
 
             var nodeIds = {};
             resolver.parentWays(this)
-                .filter(way => {
-                    // check for outbound one-way ways from stop signs
-                    const isOutbound = way.tags.oneway === 'yes' && way.nodes[0] === this.id;
-                    const isStopSign = this.tags.highway === 'stop';
-                    if (isOutbound && isStopSign) return false;
-                    return osmShouldRenderDirection(this.tags, way.tags);
-                })
+                .filter(way => osmShouldRenderDirection(this.tags, way.tags))
                 .forEach(function(parent) {
                     var nodes = parent.nodes;
                     for (i = 0; i < nodes.length; i++) {
                         if (nodes[i] === this.id) {  // match current entity
+                            const isStopOrSignal = (this.tags.highway === 'stop' && this.tags.stop === 'all') || this.tags.highway === 'traffic_signals';
+                            // handle outbound one-way ways
+                            if (isStopOrSignal && parent.isOneWay() && !hasPartialOneway(parent)) {
+                                // forward one-way, inbound is from previous node
+                                if (parent.isOneWayForwards() && i > 0) {
+                                    nodeIds[nodes[i - 1]] = true;
+                                }
+                                // backward one-way, inbound is from next node
+                                if (parent.isOneWayBackwards() && i < nodes.length - 1) {
+                                    nodeIds[nodes[i + 1]] = true;
+                                }
+                                // ignore outbound
+                                continue;
+                            }
+
+                            // default rendering logic for all nodes
                             if (lookForward && i > 0) {
                                 nodeIds[nodes[i - 1]] = true;  // look back to prev node
                             }
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -347,6 +347,29 @@ export function osmShouldRenderDirection(vertexTags, wayTags) {
     return true;
 }
 
+/**
+ * Checks if a way has partial oneway tags that invalidate strict one-way behavior.
+ * @param {Object} way The way to check.
+ * @returns {boolean} True if the way has partial oneway tags, false otherwise.
+ */
+
+export function hasPartialOneway(way) {
+    // iterate over all tags in the way
+    for (const key in way.tags) {
+        if (key.startsWith('oneway:')) {
+            const value = way.tags[key].toLowerCase();
+
+            // Functionally two way traffic
+            if (value === 'no') return true;
+            // Reverse one-way
+            if (value === '-1' && way.isOneWayForwards()) return true;
+            // Backward one-way
+            if (value === 'yes' && way.isOneWayBackwards()) return true;
+        }
+    }
+    return false;
+}
+
 export var osmSummableTags = new Set([
     'step_count',
     'parking:both:capacity',


tordans:
From a mapping point of view: this signs is very likely not in the middle of the intersection. It would be better to place the sign nodes away from the center where the stop line is. This will prevent this error and make it clear for data consumers what to do here. Direction signs on vertex habe been reported before and the solution mentioned before was to map them differently.

tyrasd:
> It would be better to place the sign nodes away from the center where the stop line is.

Well, the [wiki recommends](https://wiki.openstreetmap.org/w/index.php?title=Tag:highway%3Dstop&oldid=2738567#All-way_stop) to use the tag on the intersection node for all-way stops, specifically: 

> All-way stops should be shown using `highway=stop` at the intersection node.

:shrug:

modules/osm/node.js
```diff
@@ -120,7 +120,13 @@ const prototype = {
 
             var nodeIds = {};
             resolver.parentWays(this)
-                .filter(way => osmShouldRenderDirection(this.tags, way.tags))
+                .filter(way => {
+                    // check for outbound one-way ways from stop signs
+                    const isOutbound = way.tags.oneway === 'yes' && way.nodes[0] === this.id;
```
tyrasd: There is also the (rare-ish) case of [`oneway=-1`](https://wiki.openstreetmap.org/wiki/Tag:oneway=-1), which would be outbound when `way.nodes[way.nodes.length - 1] === this.id`.

tyrasd: And, it could be the case that a oneway way passes through the intersection without having its start or end node on the intersection. For example:

```text
     a
     |
b----c====d
     |
     e
```

Here we have 3 ways: b-c, c-d, a-c-e. If a-c-e is a oneway, the direction cone should only be drawn in one direction (on the segment a-c in the `oneway=yes` case, or on c-e when `oneway=-1`).

Maybe it makes sense to add the check into the `forEach` method, which already evaluates the direction of the way also when the vertex is not at the start/end of the way?

tyrasd (changes requested):
> this logic could be extended to other node types

Out of the top of my head, I can think of one other common tag that would benefit from this fix: `highway=traffic_signals` can also be mapped on the intersection vertex (example: [node/41338269](https://www.openstreetmap.org/node/41338269)), which would only affect inbound roads. Can you please add that tag in addition to `highway=stop` please?

> I implemented this within `.filter()` instead of `osmShouldRenderDirection()`

I think that's fine, as `osmShouldRenderDirection` only determines whether the way-vertex combination should get the directions drawn at all, or if the line should be skipped.

Maybe it would make sense to integrate the logic into the `.forEach`, though, which already evaluates the direction of the vertex respective to the way. See also the related inline comment below.

modules/osm/node.js
k-yle: there are [other tags that imply `oneway`](https://github.com/openstreetmap/id/blob/develop/modules/osm/tags.js#L153), so it'd be better to use [some of the existing helper functions](https://github.com/openstreetmap/iD/blob/992c712c9f2b96227a38fadabed15d37fcbb0cf7/modules/osm/way.js#L145-L170). For example:
```suggestion
                    const isOutbound = way.isOneWayForwards() && way.nodes[0] === this.id;
```

modules/osm/node.js
```diff
@@ -120,7 +120,13 @@ const prototype = {
...
+                    const isStopSign = this.tags.highway === 'stop';
```
k-yle:
```suggestion
                    const isStopSign = this.tags.highway === 'stop' && this.tags.stop === 'all';
```
I think we need to check for `stop=all`, because otherwise this codes breaks the logic for nodes that are explicitly tagged as `direction=backward`.

This is also a good reminder of https://github.com/openstreetmap/id-tagging-schema/pull/1539 which should probably be extended to other countries that don't have `stop=all`

modules/osm/node.js
k-yle: in some cases, it is actually logical for a stop sign to face the outbound direction on a one-way road.


For example, `oneway=yes` + `oneway:bus=no` ([example](https://osm.org/way/704118142)) and [other variants of `oneway:*`](https://taginfo.openstreetmap.org/search?q=oneway%3A). These roads are two-way for some vehicles, so the stop sign could conceivably face the _wrong_ (outbound) direction.

likewise for [`oneway:*=-1`](https://osm.wiki/Tag:oneway:bicycle=-1), which is oneway in the opposite direction for some vehicles.

tyrasd: Good point. So, to consider these cases, we'd need add an additional check (after `way.isOneWayForwards/Backwards` to look if there exists a tag that might invalidate the oneway-ness of the road (i.e. `oneway:*=no`).

Luckily, the [`cycleway=opposite*`](https://wiki.openstreetmap.org/wiki/Tag:cycleway%3Dopposite) tags have been deprecated some time ago, so one less exception to consider.

zandonella: I've implemented and fixed all of the other requested changes. Thanks for the helpful suggestions!

For this part, is there already a helper function or preferred pattern for detecting `oneway:*=no` tags? I’m a bit stuck on the best way to handle that.

I’m thinking the most reliable behavior would be if any of these tags are present, skip the outbound filtering and let the previous rendering apply since the road is functionally two-way for some types of traffic and should show all direction cones.

k-yle: @zandonella did you push you changes?

>  is there already a helper function or preferred pattern for detecting `oneway:*=no` tags?

No, there's no such helper yet, but you could put your code in a re-usable helper function if you think it might be useful elsewhere in the future. 

> if any of these tags are present, skip the outbound filtering

sounds good to me. But you would need to check if (`oneway:*=no`) or (`oneway:*=-1` and [isOneWayForwards](https://github.com/openstreetmap/iD/blob/992c712c9f2b96227a38fadabed15d37fcbb0cf7/modules/osm/way.js#L145)) or (`oneway:*=yes` and [isOneWayBackwards](https://github.com/openstreetmap/iD/blob/992c712c9f2b96227a38fadabed15d37fcbb0cf7/modules/osm/way.js#L152))

zandonella:
I added a helper function to `/modules/osm/tags.js` to check for any additional oneway tags. I took a look at the tag info provided, and I think this should cover all cases, but if this is missing something, it should be super easy to add in. I was also able to fix all of the other requested changes and all of the provided example nodes appear to be rendering properly.

Zaneo:
What if there is `oneway:bicycle=no` will it still not render?

modules/osm/node.js
```diff
@@ -79,146 +70,97 @@ const prototype = {
...
+            if (!lookForward && !lookBackward) return;
+
+            var nodeIds = {};
+            resolver.parentWays(this)
+                .filter(way => osmShouldRenderDirection(this.tags, way.tags))
+                .forEach(function(parent) {
+                    var nodes = parent.nodes;
+                    for (i = 0; i < nodes.length; i++) {
+                        if (nodes[i] === this.id) {  // match current entity
+                            const isStopOrSignal = (this.tags.highway === 'stop' && this.tags.stop === 'all') || this.tags.highway === 'traffic_signals';
```
matkoniecz, line 128: can this, or a bit more code can be cleanly packaged into a separate function?

NOTE: I am not a maintainer, maybe wait for actual iD maintainer, note that @k-yle reviewed it and made no such request (except "No, there's no such helper yet, but you could put your code in a re-usable helper function if you think it might be useful elsewhere in the future."), so maybe this should be ignored

matkoniecz:
> What if there is oneway:bicycle=no will it still not render?

From reading `hasPartialOneway` it seems it should work properly but I have not tested it

PR #11261, opened 2025-07-29
right-align input fields with expected numeric content

### LINKED ISSUE #9677: Right-align numeric fields
#8769 changed the behavior of numeric fields to accept numbers in the user’s language instead of POSIX English. This is a significant change to users of some languages, who may have formed muscle memory around entering raw tag values in the non-raw field over many years. I believe even these users would be accustomed to entering numbers in their language in other Web applications, due to the prevalence of `<input type="number">`. However, we need to clue them into the fact that the behavior has suddenly changed.

Since we didn’t first prime the user by applying a monospaced font to numeric fields for a release or two, as suggested in https://github.com/openstreetmap/iD/pull/8769#issuecomment-1453926011, how about right-aligning numeric fields? This is a common practice in UIs on many platforms, not to mention on calculators. It would make a clearer distinction between real numeric fields and those that happen to take numbers most of the time, a point that even confuses yours truly: https://github.com/openstreetmap/iD/issues/3615#issuecomment-1564729750. That said, it would have little effect in right-to-left languages where the fields are already right-aligned.

tyrasd:
Resolves #9677.

Affected are fields of the type:

* `number`
* `roadHeight`
* `roadSpeed`
* (implicitly also the `layer` subfield of the `structure` field)
* `date`

Not included are `addr:housenumber`, `addr:postcode` which can be numeric, but are often of a mixed nature.

Example (just for fun in a locale with a localized decimal separator):

image
<br>
<br>

I'm not quite 100% convinced that this is really a benefit overall:
* :heavy_plus_sign: It brings the numeric value closer to the unit/measurement or the `+ / -` buttons
* :heavy_minus_sign: It's a little harder to see if a field is empty when quickly scanning from top to bottom
* the biggest benefit of right-aligning numbers is when numbers are represented in tabular form (e.g. on a spreadsheet), where it affords one to quickly compare numeric values by their length and their most significant digit. This is not really the case for numeric fields in iD.

What do you think? //cc @1ec5


tordans:
I don’t think this is an improvement. It puts the data where the UI is very dense making it harder to see. And makes the data harder to scan.

PR #11265, opened 2025-07-30
try to fix a few flaky unit tests

k-yle:
Same as #10690. Because [we enabled `--no-isolate`](https://github.com/openstreetmap/iD/commit/108893c6242d9ee12dde300fae5d1dfa793709d6), some unit tests which mutate global state can occasionally fail. 

see code review comments for more details

### DIFF
--- a/modules/core/history.js
+++ b/modules/core/history.js
@@ -148,7 +148,8 @@ export function coreHistory(context) {
 
             if (transitionable) {
                 var origArguments = arguments;
-                d3_select(document)
+                return new Promise(resolve => {
+                  d3_select(document)
                     .transition('history.perform')
                     .duration(duration)
                     .ease(d3_easeLinear)
@@ -158,11 +159,12 @@ export function coreHistory(context) {
                         };
                     })
                     .on('start', function() {
-                        _perform([action0], 0);
+                        resolve(_perform([action0], 0));
                     })
                     .on('end interrupt', function() {
-                        _overwrite(origArguments, 1);
+                        resolve(_overwrite(origArguments, 1));
                     });
+                });
 
             } else {
                 return _perform(arguments);
--- a/test/spec/services/osm_wikibase.js
+++ b/test/spec/services/osm_wikibase.js
@@ -277,12 +277,12 @@ describe('iD.serviceOsmWikibase', function () {
         }),
         status: 200,
         headers: { 'Content-Type': 'application/json' }
-    });
+      });
 
       wikibase.getEntity({ key: 'amenity', value: 'parking', langCodes: ['fr'] }, callback);
 
       await setTimeout(50);
-      expect(parseQueryString(fetchMock.calls()[0][0])).to.eql(
+      expect(parseQueryString(fetchMock.calls(/action=wbgetentities/)[0][0])).to.eql(
         {
           action: 'wbgetentities',
           sites: 'wiki',
--- a/test/spec/ui/flash.js
+++ b/test/spec/ui/flash.js
@@ -18,20 +18,14 @@ describe('iD.uiFlash', function () {
             .remove();
     });
 
-    it('flash is shown', function() {
-        iD.uiFlash(context).duration(200)();
+    it('flash is shown and goes away', async () => {
+        iD.uiFlash(context).duration(0)();
         var flashWrap = d3.selectAll('.flash-wrap');
         var footerWrap = d3.selectAll('.main-footer-wrap');
         expect(flashWrap.classed('footer-show')).to.be.ok;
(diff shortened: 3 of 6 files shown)


modules/core/history.js
```diff
@@ -158,11 +159,12 @@ export function coreHistory(context) {
                         };
                     })
                     .on('start', function() {
-                        _perform([action0], 0);
+                        resolve(_perform([action0], 0));
```
k-yle, line 162: instead of waiting x milliseconds, which is error-prone, returning a promise makes this test more deterministic

test/spec/services/osm_wikibase.js
```diff
@@ -277,12 +277,12 @@ describe('iD.serviceOsmWikibase', function () {
...
         status: 200,
         headers: { 'Content-Type': 'application/json' }
-    });
+      });
 
       wikibase.getEntity({ key: 'amenity', value: 'parking', langCodes: ['fr'] }, callback);
 
       await setTimeout(50);
-      expect(parseQueryString(fetchMock.calls()[0][0])).to.eql(
+      expect(parseQueryString(fetchMock.calls(/action=wbgetentities/)[0][0])).to.eql(
```
k-yle, line 285: there might be other services calling API requests, so explicitly specify which mock we're interested in

test/spec/ui/flash.js
```diff
@@ -1,9 +1,16 @@
...
 
+vi.mock('d3-timer');
+
 describe('iD.uiFlash', function () {
     var context;
 
     beforeEach(function() {
+        // With this mock, a timeout of 0 will be immediately resolved, any
+        // other number will never resolve, to make the tests more deterministic.
+        vi.mocked(d3_timeout).mockImplementation((cb, timeout) => timeout === 0 && cb());
```
k-yle: see code comments. Currently these tests have `duration: 200` and `setTimeout: 225`, this change avoids the need for precise timing

test/spec/validations/osm_api_limits.js
```diff
@@ -2,7 +2,10 @@ describe('iD.validations.osm_api_limits', function () {
     var context;
 
     beforeEach(function() {
-        iD.services.osm = { maxWayNodes: function() { return 10; } };
+        iD.services.osm = {
+            maxWayNodes: () => 10,
+            on: () => undefined,
```
k-yle, line 7: this incomplete mock was causing the error:

> `osmService.on` is not a function

in the short time period after the test completes, but before it's cleaned up. only happens on slow machines.

test/spec/ui/flash.js
tyrasd: To be honest, this one confuses me a little… :thinking: 

* For the `flash is shown` test: with a duration of 0 the flash will be hidden with no delay, but the test still works because the transition only finishes on the next tick. So far so good.
* The `flash goes away` test uses 200ms duration and checks the result after 225ms. But if the mocked version of d3's internal timeout never resolves (as `duration > 0`), how does the test still pass?
* I tried to replace the mock implementation with something stupid like `throw new Error()`, but everything continues to pass. Also, when I checked how often the mocked function was called, it reports 0. What am I overlooking? :see_no_evil: 

PS: Would it perhaps be even easier to just combine the two test cases into one that checks the "visible" state once right at the beginning and once again for the "hidden" state after the 225 ms timeout? Then there would never be the possibility of a race condition between the two tests that operate on the same DOM elements.

```js
    it('flash is shown and goes away', async function() {
        iD.uiFlash(context).duration(100)();
        const flashWrap = d3.selectAll('.flash-wrap');
        const footerWrap = d3.selectAll('.main-footer-wrap');
        expect(flashWrap.classed('footer-show')).to.be.ok;
        expect(footerWrap.classed('footer-hide')).to.be.ok;
        await setTimeout(200);
        d3.timerFlush();
        expect(flashWrap.classed('footer-hide')).to.be.ok;
        expect(footerWrap.classed('footer-show')).to.be.ok;
    });
```

k-yle: yeah, this one is pretty weird. I went with your suggestion, and:
* changed the `.duration(100)` to `0`, because it doesn't matter, we just need ≥1 process tick
* also called `d3.timerFlush()` _before_ `setTimeout`. I don't see any purpose in calling it after `setTimeout`

Seems to pass consistently now[^1]


[^1]: using my (very unscientific) test method of running `npm run test:spec` 30 times with random values of [`--max-old-space-size`](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-mib) while doing CPU-intensive stuff in the background 😆

tyrasd (approved):

PR #11268, opened 2025-07-31
move enable/disable buttons to the top of the list

### LINKED ISSUE #8878: Move "Deactivate/Activate all" to top of list

ved17-git:
Moved the "Activate all" and "Deactivate all" options from the bottom to the top of the object selection list to improve visibility and accessibility.
This ensures users can easily discover and use these actions without needing to scroll through a long list.

Fixes: #8878
image

### DIFF
--- a/modules/ui/sections/map_features.js
+++ b/modules/ui/sections/map_features.js
@@ -20,31 +20,33 @@ export function uiSectionMapFeatures(context) {
             .append('div')
             .attr('class', 'layer-feature-list-container');
 
-                var footer = containerEnter
-            .append('div')
-            .attr('class', 'feature-list-links section-footer');
 
-        footer
-            .append('a')
-            .attr('class', 'feature-list-link')
-            .attr('role', 'button')
-            .attr('href', '#')
-            .call(t.append('issues.disable_all'))
-            .on('click', function(d3_event) {
-                d3_event.preventDefault();
-                context.features().disableAll();
-            });
+var footer = containerEnter
+    .append('div')
+    .attr('class', 'feature-list-links section-footer')
+    .style('margin-top', '-25px');
 
-        footer
-            .append('a')
-            .attr('class', 'feature-list-link')
-            .attr('role', 'button')
-            .attr('href', '#')
-            .call(t.append('issues.enable_all'))
-            .on('click', function(d3_event) {
-                d3_event.preventDefault();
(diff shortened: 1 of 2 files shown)


tordans:
Styling: the padding looks off. I think the space between headline and action links should be quite a bit smaller.

ved17-git:
> Styling: the padding looks off. I think the space between headline and action links should be quite a bit smaller.

Fixed the padding — the spacing between the headline and action links is now reduced.

tyrasd (changes requested):
Hi. Thanks for giving this another try. I think having those buttons on the start of the section does slightly improve the usability. However there are still some things I noticed, that still need to be addressed:

* Somehow the indentation of the moved code was messed up in this PR. Please restore the proper indentation
* Please move the CSS to the appropriate css style file (`80_app.css`), and maybe introduce a new css class for this kind of "table header" (replacing the `.section-footer` class, which does not really fit anymore if the buttons are on the top)
* It's probably best to completely remove the 30px height offset (instead of just reducing the margin to 5px)
* There is another list that uses the same disable/enable all buttons: in the validation/issues sidebar. This also needs to be changed to keep the UI consistent.
* `package-lock.json` should not be committed if no dependencies are changed

One last thing while we're at it: This is just an idea, and you don't have to address it, but I noticed that those enable/disable buttons are actually the only (?) ones in iD that use an "unstyled" link instead of a proper button UI element, which seems odd. I assume this choice was made on purpose back in the day (e.g. to de-emphasise the importance of these buttons), but if we change the buttons to the top that could introduce potential confusion. It would be possible to visually group the buttons to the list by introducing a border (and changing the links to proper `<button>` elements with hover styles, etc.) like in the following mockup:

matkoniecz:
@ved17-git do you plan to continue working on that PR?

matkoniecz:
Ping @ved17-git

tyrasd:
closing as stale

PR #11269, opened 2025-07-31
Add user-defined keyboard shortcuts for presets (#11170)

### LINKED ISSUE #5873: Quick feature creation from favorite presets

### LINKED ISSUE #11170: Quickly select a preset
This issues is part of https://github.com/openstreetmap/iD/issues/11168

@kaligrafy, @tyrasd and @tordans meet online to discuss how and advanced user feature to quickly select a preset could be integrated into iD.

---

**Goals**
* A flow that allows trained users to very quickly add features to iD, eg. hundreds of driveways
* At the same time, keep iD an editor that is easy to understand for new mapper

---

**Context**
The V3-Prototype (link in https://github.com/openstreetmap/iD/issues/11168) explored a different UX for adding features. Instead of Geometry > Preset the main UX is Preset > Geometry. Part of this rework was adding favorites which are accessible via the top icon bar. Each favorite in kaligrafy/iD has a shortcut.
> Image

**The current flow is…**
1. Press the shortcut eg. `9` => The UI is in drawing mode for this preset with the selected geometry type (line)
2. Draw on the map => Done

**Key questions:**

When we want to port this to iD, key questions are…
- What are the consequences of adding many shortcuts to iD?
   - Good to know: The default shortcuts are defined in the code. However they can be overwritten for locales via the translation system. The shortcut code has a collision detection (or can have). Changing languages after a custom shortcut is defined, can cause issues.
- How do we handle the different UX when it comes to adding features (Geometry > Preset)

---

# Possible solutions to add this to iD:

## (a) Reintroduce V3

Evaluation: This would be a lot of work technically but also likely introduce many new interaction issues with other features. So nothing we want to pursue here…

## (b) Favorites + Any Shortcuts (for a Geometry type)

- We could add a UI to add a shortcut to a favorite. Part of defining the shortcut would be to also specify the geometry type (line, point, area).
- We could introduce favorites to iD. Coupling shortcuts and favorites could make it easer to understand which presets actually have a shortcut.

**UI** and **Flow**: _See (c)_, but with any shortcut (modifier key + any key) and collision handling.

_Disadvantage:_
- Compared to (c) is is more complex to define the shortcut
- Compared to (c) it opens up more issues when changing translations (which might introduce new collisions) 

## (c) Favorites + Number-Shortcuts (for a Geometry type)

This is the same as (b) but _only allowing numbers as presets shortcuts_. 

_Advantages:_
- No collision with existing or translated shortcuts (`1`,`2`,`3` are reserved)
- Easier to explain/understand

_Disadvantages:_
- Only having `4-0` as numbers is too limiting. We should also have `44`, `555` etc. Which might be hard to implement…

**Idea to use this that requires a Geometry first**

**UI:** Create this kind of a shortcut:
> Define Shortcut (Number): ```[____]```
> Hint: Provide a number like `5` or `55` to select this preset. Use the shortcut after you drew a geometry.

**Flow:** We have a shortcut `33` for bike stand. (In this case the shortcut is for a preset, no additional Geometry is defined except by the preset)
1. press `1` on the map to draw a point geometry 
   => point tool 
   => click on map to create a point
2. press `33` while point is selected 
    => selects bike stand preset
   (TBD: We have to handle the fact that the focus is in the search bar at this point, see (d))
3. done

**Idea to use this that auto-applies a geometry**

**UI:** Create this kind of a shortcut:
> Define Shortcut (Number): ```[____]```  (**Point**|Line|~~Area~~)
> Hint: Provide a number like `5` or `55` to select this preset. A shortcut requires a default geometry type.

- The Point|Line|Area is a toggle button (likely as icons)
- One is always selected
- Those that are not allowed by the preset definition are grayed out/unselectable

**Flow:** We have a shortcut `33` for bike stand and defined point geometries as the default (when we created the shortcut).
1. press `33` on the map
    => point tool (for bike stands)
    => click on the map to create the bike stand point
3. done

And if I want to draw the bike stand as an area…
- We could add the "select geometry" step as in the variant before (`3` (area) then `33` (bike stand))
- We could add "transform to … (point,line,area)" actions which would allow to quickly change the geometry type (via a toggle point>line>area>point…)
- We could add modifier keys that apply to all favorites
   - `33` is the default geometry type (whatever we picked)
   - shift+33 forces points (if allowed)
   - alt+33 forces areas (if allowed)
   - shift+alt+33 forces lines (if allowed)
   - … always showing an error message if the geometry type is not allowed, falling back to the default

## (d) Custom search terms for the preset search

- We could introduce custom search terms for presets. Given the existing UX, those would work very similar to shortcuts.

_Advantages:_
- Similar workflow to (c)
- No shortcuts at all, so no collisions and related complexity
- Reuses an existing interaction pattern

_Disadvantages:_
- Unconventional, so not what users will expect at first; needs a bit more explaining
- Only works in the Idea-Variation that requires a geometry to be preset…

**UI:** Create this kind of a "shortcut":
> Quicksearch: ```[____]``` (any String)
> Hint: This quicksearch term will allow you to find the preset quickly when using the preset search. These terms are personal to you and stored in your OSM user profile.

**Flow:** We have a quicksearch `bs` for bike stand.
1. press `1` on the map to draw a point geometry 
   => point tool 
   => click on map to create a point
   => the focus is now in the feature selection search bar (existing feature)
2. press `bb` which add the text to the search bar
    => our bike stand is returned as the first result
3. press `enter` to select the first result (existing feature)
    => selects bike stand preset
4. done

**Idea to use this that auto-applies a geometry**

This flow does not work for this idea.

---

### We need a way to apply and remove the shortcuts

As a user, I need to apply and remove shortcusts.

Possible solutions are…
- https://github.com/openstreetmap/iD/issues/11066 could open up space for this; this idea is more about object actions, than preset actions, though…
- `todo`

### We need a way to list all favorite and mange them:

As a user, I need to understand which favorites have custom shortcuts. 

Possible solutions are…
- Add a start Icon to the preset, that makes it stand out in the list
- Add a virtual preset-category for "Favorites" to quickly list them all
- Move the favorites in some UI element next to the geometry selection, similar to what V3 did
   > Image

kaligrafy:
Implements a new feature allowing users to assign custom keyboard shortcuts (numbers 8-999) to any preset, with shortcuts persisted in browser storage.

Features:
- Inline shortcut editor in preset inspector with edit/save/cancel/delete buttons
- Smart multi-digit detection with timeout handling (e.g., "23" for custom presets)
- Preserves built-in shortcuts (1-3) with override capability for longer shortcuts
- Visual feedback via flash notifications when shortcuts are activated
- Input validation with user-friendly error messages
- Proper cleanup to prevent infinite buffer growth
- Full tooltip support for all UI elements

UI Components:
- New shortcut editor integrated into feature type section
- Trash can icon for delete, checkmark for save, X for cancel
- Error messages displayed below input with clear range requirements
- Edit button hides during edit mode for clean interface

Technical Implementation:
- modules/core/preset_shortcuts.js: Storage and management layer
- modules/behavior/preset_shortcuts.js: Keyboard event handling
- modules/ui/sections/feature_type.js: Inline editor UI
- localStorage persistence for user sessions
- Proper D3 behavior pattern with utilRebind

The feature allows rapid preset selection during mapping workflows while maintaining compatibility with existing keyboard shortcuts.

Fixes #11170
Partially fixes https://github.com/openstreetmap/iD/issues/5873

Made with the help of Claude Sonnet 4 (Thinking) from Anthropic.

### DIFF
--- a/docs/preset_shortcuts.md
+++ b/docs/preset_shortcuts.md
@@ -1,0 +1,164 @@
+# Preset Shortcuts Feature
+
+This document describes the preset shortcuts feature that allows users to create custom keyboard shortcuts for frequently used presets.
+
+## Overview
+
+The preset shortcuts feature enables users to assign number keys (8-999) to their favorite presets, allowing for rapid feature creation. When a preset has an assigned shortcut, users can simply press the number key(s) to either:
+
+1. Enter drawing mode with that preset (if no features are selected)
+2. Apply the preset to currently selected features
+
+## User Interface
+
+### Setting Shortcuts
+
+1. **Select a preset**: When a preset is selected in the feature type section, an inline shortcut editor appears below the preset name
+2. **View current shortcut**: The editor shows "Shortcut: none" or "Shortcut: [number]" if one is already set
+3. **Edit shortcut**: Click the edit button (pencil icon) to open the inline editor
+4. **Enter a number**: Type a number between 8 and 999 in the small input field
+5. **Save or Cancel**: Click the checkmark to save, X to cancel, or trash icon to remove existing shortcut
+
+### Using Shortcuts
+
+Once a shortcut is set, users can:
+
+- **Press the shortcut number**: Type the assigned number to activate the preset
+- **Multi-digit shortcuts**: For shortcuts like "22" or "555", type the digits in sequence within 800ms
+- **Single-digit shortcuts**: For shortcuts like "8" or "9", press once
+
+### Removing Shortcuts
+
+To remove an existing shortcut:
+- Click the edit button to open the inline editor
+- Click the trash icon to immediately remove the shortcut
+- Alternatively, clear the input field and save to remove the shortcut
+
+## Technical Implementation
+
+### Architecture
+
+The feature consists of three main components:
+
+1. **Core Shortcuts Manager** (`modules/core/preset_shortcuts.js`)
+   - Manages shortcut storage in localStorage
+   - Handles validation and conflict resolution
+   - Provides APIs for getting/setting shortcuts
+
+2. **Keyboard Behavior** (`modules/behavior/preset_shortcuts.js`)
+   - Captures number key presses
+   - Implements multi-digit detection with timeout
+   - Activates presets when shortcuts are pressed
+
+3. **UI Integration** (`modules/ui/sections/feature_type.js`)
+   - Adds inline shortcut editor to preset display
+   - Handles inline editing with save/cancel/remove actions
+   - Updates display based on shortcut status
+
+### Data Storage
+
+Shortcuts are stored in localStorage under the key `preset_shortcuts` as a JSON object:
+
+```json
+{
+  "8": "amenity/restaurant",
+  "22": "highway/footway", 
+  "555": "amenity/parking"
+}
+```
+
+### Keyboard Handling
+
+The keyboard behavior uses a buffering system with timeout:
+
+1. **Number pressed**: Added to buffer, timeout started (800ms)
+2. **Additional numbers**: Reset timeout, continue buffering
+3. **Timeout reached**: Process buffer as complete shortcut
+4. **Non-number key**: Clear buffer
+
+This allows for both single-digit (8, 9) and multi-digit (22, 555) shortcuts.
+
+### Conflict Resolution
+
+- **Range conflicts**: Only numbers 8-999 are allowed (1-7 reserved for drawing modes)
+- **Duplicate shortcuts**: When assigning a shortcut already in use, the previous assignment is removed
+- **Validation**: Real-time feedback prevents invalid inputs
+
+## Integration Points
+
+### Existing Systems
+
+The feature integrates with several existing iD systems:
+
+- **Preset Manager**: Uses existing preset lookup and validation
+- **Drawing Modes**: Leverages existing point/line/area drawing modes
+- **Action System**: Uses existing `actionChangePreset` for applying presets
+- **Localization**: All user-facing text supports translation
+- **Storage**: Uses existing preferences system for persistence
+
+### UI Components
+
+- **Modal System**: Uses existing modal framework
+- **Tooltips**: Uses existing tooltip system
+- **Icons**: Uses existing SVG icon system
+- **CSS**: Follows existing styling patterns
+
+## Configuration
+
+### Wait Duration
+
+The timeout for multi-digit shortcuts can be configured:
+
+```javascript
+ui.presetShortcuts.waitDuration(1000); // Set to 1 second
+```
+
+### Shortcut Range
+
+The valid shortcut range is hardcoded to 8-999 to avoid conflicts with:
+- Numbers 1-3: Reserved for drawing modes (point, line, area)
+- Numbers 4-7: Reserved for potential future drawing modes
+
+## Error Handling
+
+### Validation Errors
+
+- **Out of range**: Numbers below 8 or above 999 are rejected
+- **Invalid format**: Non-numeric input is rejected
+- **Conflicts**: Users are warned when shortcuts conflict
+
+### Runtime Errors
+
+- **Storage failures**: Gracefully handled if localStorage is unavailable
+- **Preset not found**: Silently ignored if preset ID is invalid
+- **Invalid geometry**: Preset application skipped if geometry doesn't match
+
+## Accessibility
+
+- **Keyboard navigation**: Full dialog navigation with Tab/Enter/Escape
+- **Screen readers**: Proper labeling and ARIA attributes
+- **Visual feedback**: Clear validation messages and states
+- **Focus management**: Automatic focus to input field
+
+## Performance Considerations
+
+- **Lazy loading**: Shortcuts loaded only when first accessed
+- **Event efficiency**: Keyboard handler uses capture phase for performance
+- **Memory usage**: Minimal overhead with efficient data structures
+- **Storage optimization**: Compact JSON storage format
+
+## Browser Compatibility
+
+- **localStorage**: Required for shortcut persistence
+- **Keyboard events**: Standard event handling, works in all modern browsers
+- **CSS**: Uses standard properties with vendor prefixes where needed
+
+## Future Enhancements
+
+Potential improvements for future versions:
+
+1. **Import/Export**: Allow users to backup and share shortcut configurations
+2. **Visual indicators**: Show shortcut numbers on preset icons
+3. **Search integration**: Include shortcuts in preset search results
+4. **Conflict warnings**: Proactive warnings when setting conflicting shortcuts
+5. **Custom ranges**: Allow administrators to configure valid shortcut ranges
--- a/modules/core/preset_shortcuts.js
+++ b/modules/core/preset_shortcuts.js
@@ -1,0 +1,140 @@
+import { prefs } from './preferences';
+import { dispatch as d3_dispatch } from 'd3-dispatch';
+import { utilRebind } from '../util';
+
+/**
+ * Core Preset Shortcuts Manager
+ *
+ * Manages user-defined keyboard shortcuts for presets. This module provides
+ * functionality to store, retrieve, and manage preset shortcuts that allow
+ * users to quickly activate presets using number keys 8-999.
+ *
+ * Features:
+ * - Store shortcuts in localStorage for persistence across sessions
+ * - Support numeric shortcuts from 8-999 (1-7 reserved for drawing modes)
+ * - Validate shortcuts to ensure they're within the allowed range
+ * - Handle conflicts when multiple presets try to use the same shortcut
+ * - Dispatch events when shortcuts are added, removed, or changed
+ *
+ * Usage:
+ *   const shortcuts = corePresetShortcuts();
+ *   shortcuts.setShortcut('amenity/restaurant', '42');
+ *   const presetId = shortcuts.getPreset('42'); // 'amenity/restaurant'
+ *
+ * Events:
+ *   'shortcutAdded' - fired when a shortcut is assigned to a preset
+ *   'shortcutRemoved' - fired when a shortcut is removed from a preset
+ *   'shortcutChanged' - fired when a shortcut is modified
+ */
+export function corePresetShortcuts() {
+    const dispatch = d3_dispatch('shortcutAdded', 'shortcutRemoved', 'shortcutChanged');
+
+    let _shortcuts = {};
+    let _loaded = false;
+
+    // Load shortcuts from localStorage
+    function loadShortcuts() {
+        if (_loaded) return;
+
+        try {
+            const stored = prefs('preset_shortcuts');
+            if (stored) {
+                _shortcuts = JSON.parse(stored);
+            }
+        } catch (error) {
+            console.error('Error loading shortcuts:', error);
+            _shortcuts = {};
+        }
+        _loaded = true;
+    }
+
+    // Save shortcuts to localStorage
+    function saveShortcuts() {
+        try {
+            prefs('preset_shortcuts', JSON.stringify(_shortcuts));
+        } catch (error) {
+            console.error('Error saving shortcuts:', error);
+        }
+    }
+
+    const presetShortcuts = {
+        // Get shortcut number for a preset ID
+        getShortcut: function(presetId) {
+            loadShortcuts();
+            return Object.keys(_shortcuts).find(shortcut => _shortcuts[shortcut] === presetId);
+        },
+
+        // Get preset ID for a shortcut number
+        getPreset: function(shortcut) {
+            loadShortcuts();
+            return _shortcuts[shortcut];
+        },
+
+        // Set a shortcut for a preset
+        setShortcut: function(presetId, shortcut) {
+            loadShortcuts();
+
+            // Validate shortcut format (8-999)
+            const num = parseInt(shortcut, 10);
+            if (isNaN(num) || num < 8 || num > 999) {
+                throw new Error('Shortcut must be a number between 8 and 999');
+            }
+
+            // Remove any existing shortcut for this preset
+            const existingShortcut = Object.keys(_shortcuts).find(s => _shortcuts[s] === presetId);
+            if (existingShortcut) {
+                delete _shortcuts[existingShortcut];
+            }
+
+            // Remove any preset using this shortcut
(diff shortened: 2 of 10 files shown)


kaligrafy:
Add the shortcut icon left of the info icon in the preset panel on the left. Also, when clicking this shortcut icon, open the preferences panel or a modal window. The preferences panel should also have the shortcuts section where we could edit shortcuts for any preset there. (Still subject to discussion though)

tordans:
Very short notes from the meetup
- form in a modal; away from the feature - trigger modal via icon next to info-i
   - modal content
     > Define a shortcut for Preset <NAME>
     > To use the shortcut, eg. 33:
     > * On the map, press "33"
- selected existing way => short cut 33 => does not apply preset (but 77 would)
- fresh line => short cut 33 => noes not apply preset (but 77 would)
- short cut 33 (or 77) => draw works great
- ideally: when shortcut is clicked, show preset in the left panel (not just bottom)
- list current shortcuts in preference section, action to edit, action to delete

---

Also: Thanks a lot for this great feature!

kaligrafy:
Issue with shortcuts starting with 1-3 digits like 33 is fixed.
Also, now the shortcut preset is shown on the left
It also forbids change an selected object to a preset with a different geometry (for instance you cannot change a line object to an area preset)

The shortcut editor now opens in a modal window and the shortcut button appear left of the reference/info button.

The preferences pane now show shortcuts and makes it possible to edit them.

kaligrafy:
Some screenshots:

image
Clicking on the left button will open modal window below:

Edit in modal:

image

After some shortcuts have been set, you can edit them in the preferences pane:

image

When no shortcut is defined:

image

kaligrafy:
@tyrasd hi! Is it possible to get a timeline on how and when to implement this feature? Thanks!

kaligrafy:
little bump :-)

tordans:
During the [last community](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Notes) chat Martin agreed that if Kyle, Pierre-Leo and I find the time to polish and review this feature we can do so and merge it. Keeping him in the loop during relevant changes. This should be based on what Martin, Pierre-Leo and Tobias discussed during the two Meetings we had for this (and other) features.

---

Last weekend I looked into the status of this PR with a focus on the UX side of things. The last plan for UX we discussed was a bit different from what we have here ATM. But at the same time, after working with it, I think it still need refinement (more below).

I prototyped the changes locally but the code is not finished yet to push it. I hope to find time to do so soon.

Here are is an overview about the changes and status quo:

From a end user point of view, we explain the feature as a way to mark presets as favorite. Part of being a favorite is, that a preset gets a shortcut.

The UI to set/remove the preset is moved in the preset-section of the sidebar. The blue star is an active favorite, the gray star can be activated. Once active, a Favorite gets a shortcut. Reminder: The shortcuts can only be numbers. Right now, the shortcuts are auto-numbered (more below). The shortcut is visible below the star to make discovering the number easy. The star also has a tooltip like the Node/Way/Area (missing in the screenshot).

* Bildschirmfoto 2026-01-19 um 09 06 14
* Bildschirmfoto 2026-01-19 um 09 06 03

Each shortcut is listed in the preferences panel of the user. The two lists are in sync. The list also shows the geometries that this preset will work on. This is important when users want to understand why a specific shortcut does (not) show in the presets-favorites list where only those are listed that work with the selected object's geometry. The list also shows the shortcut and allows to delete them.

The main feature of the list is to sort the shortcuts by drag and drop. The idea in the last meeting was, that we wanted to avoid a UI to set the shortcut by hand by just auto-assigning numbers. The sorting was supposed to make it easy to reassign the numbers.

This is also the area that I think we need to re-work. Right now, deleting a shortcut from place 2 means, that all following numbers get reassigned, which is very confusing and means I cannot just remember that "5 is always crossing".
My next iteration would be to keep the list but allow for a UI to change the shortcut. The Drag-Dop-UI would then be removed.

* Bildschirmfoto 2026-01-19 um 09 05 57

Lastly, we show the favorites in the UI to select a preset. This is either for a new geometry, or when changing the preset for an existing object. It uses the same group-feature we already have. The idea was to have the Favorites on top of the list and always open. But I am not sure this is what we will end up with, given the likely duplication with the "last used" list of presets.

* Bildschirmfoto 2026-01-19 um 09 06 48
* Bildschirmfoto 2026-01-19 um 09 06 42

* Bildschirmfoto 2026-01-19 um 09 06 25
* Bildschirmfoto 2026-01-19 um 09 06 33
* This also shows how the 3 favorites get filtered to only show the one that works with the selected geometry "line".

---

Once I had a change to rework this UI more and get the code presentable I will report back. Ideally @kaligrafy  and @k-yle  can then review the current state (code + UX). Looking forward to get this into a mergable state!

kaligrafy:
@tordans Thanks for the proposition. Looks clean! Can you provide the code for it in a separate branch? I would like to work on this and progress further. Thanks!

tordans:
@kaligrafy sorry about that, here it the rebased code: https://github.com/tordans/iD/tree/pr/11269-favorites-shortcuts – I cannot quite remember which part I was stuck on and why I did not push it back then. I suggest you rebase this PR and then cherry pick my commit to then modify it…

matkoniecz (commented):
Maybe I am on extreme case here, but for me "with shortcuts persisted in browser storage." would not be too useful and I would some want to import/export them (to oms.org user preferences? to/from text file)

Maybe I would expect them to be shared between different machines where I use iD, but people using different OSes may actively NOT want this

comment as I am far from sure should this comment about  browser storage not being enough is making a good point or have I got on extreme side

modules/core/preset_shortcuts.js
```diff
@@ -0,0 +1,140 @@
...
+
+        // Set a shortcut for a preset
+        setShortcut: function(presetId, shortcut) {
+            loadShortcuts();
+
+            // Validate shortcut format (8-999)
+            const num = parseInt(shortcut, 10);
+            if (isNaN(num) || num < 8 || num > 999) {
+                throw new Error('Shortcut must be a number between 8 and 999');
+            }
```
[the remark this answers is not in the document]
matkoniecz: can you even somehow get one such input there?

modules/ui/shortcut_editor.js
```diff
@@ -0,0 +1,200 @@
...
+                showError(t('preset_shortcut.error_empty'));
+                return;
+            }
+
+            // Validate range
+            const num = parseInt(value, 10);
+            if (isNaN(num) || num < 8 || num > 999) {
+                showError(t('preset_shortcut.error_invalid_range'));
+                return;
+            }
```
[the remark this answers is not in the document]
matkoniecz: see https://github.com/openstreetmap/iD/pull/11269#discussion_r3995657412

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+## Integration Points
+
+### Existing Systems
+
+The feature integrates with several existing iD systems:
+
+- **Preset Manager**: Uses existing preset lookup and validation
+- **Drawing Modes**: Leverages existing point/line/area drawing modes
+- **Action System**: Uses existing `actionChangePreset` for applying presets
+- **Localization**: All user-facing text supports translation
```
matkoniecz, line 96: why we describe such things in this documentation here?

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+
+This allows for both single-digit (8, 9) and multi-digit (22, 555) shortcuts.
+
+### Conflict Resolution
+
+- **Range conflicts**: Only numbers 8-999 are allowed (1-7 reserved for drawing modes)
+- **Duplicate shortcuts**: When assigning a shortcut already in use, the previous assignment is removed
+- **Validation**: Real-time feedback prevents invalid inputs
+
+## Integration Points
```
matkoniecz, line 87: is it leftover instructions for LLM that generated existing code? Is it useful as committed documentation?

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+3. **Timeout reached**: Process buffer as complete shortcut
+4. **Non-number key**: Clear buffer
+
+This allows for both single-digit (8, 9) and multi-digit (22, 555) shortcuts.
+
+### Conflict Resolution
+
+- **Range conflicts**: Only numbers 8-999 are allowed (1-7 reserved for drawing modes)
+- **Duplicate shortcuts**: When assigning a shortcut already in use, the previous assignment is removed
+- **Validation**: Real-time feedback prevents invalid inputs
```
matkoniecz, line 85: what you mean by that?

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Integration Points
+
+### Existing Systems
+
+The feature integrates with several existing iD systems:
+
+- **Preset Manager**: Uses existing preset lookup and validation
+- **Drawing Modes**: Leverages existing point/line/area drawing modes
```
matkoniecz, line 94: not sure is it making to add it to the persistent documentation, but once merged "Leverages existing" become tautologically clear, as anything merged will be exististing

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+- **Action System**: Uses existing `actionChangePreset` for applying presets
+- **Localization**: All user-facing text supports translation
+- **Storage**: Uses existing preferences system for persistence
+
+### UI Components
+
+- **Modal System**: Uses existing modal framework
+- **Tooltips**: Uses existing tooltip system
```
matkoniecz, line 102: no need to document things like that, OBVIOUSLY it is not implementing own tooltip system

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+
+The valid shortcut range is hardcoded to 8-999 to avoid conflicts with:
+- Numbers 1-3: Reserved for drawing modes (point, line, area)
+- Numbers 4-7: Reserved for potential future drawing modes
+
+## Error Handling
+
+### Validation Errors
+
+- **Out of range**: Numbers below 8 or above 999 are rejected
```
matkoniecz, line 126: this repeats the same info again...

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+- **Visual feedback**: Clear validation messages and states
+- **Focus management**: Automatic focus to input field
+
+## Performance Considerations
+
+- **Lazy loading**: Shortcuts loaded only when first accessed
+- **Event efficiency**: Keyboard handler uses capture phase for performance
+- **Memory usage**: Minimal overhead with efficient data structures
+- **Storage optimization**: Compact JSON storage format
+
```
matkoniecz, line 149: is it LLM generated? Because it really needs human review to drop irrelevant parts not worth documenting and removal of puffery

(and this makes me less enthusiastic about reviewing this feature in general if it is LLM-generated in a similar way elsewhere)

docs/preset_shortcuts.md
matkoniecz: the same for other similar claims

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+## User Interface
+
+### Setting Shortcuts
+
+1. **Select a preset**: When a preset is selected in the feature type section, an inline shortcut editor appears below the preset name
+2. **View current shortcut**: The editor shows "Shortcut: none" or "Shortcut: [number]" if one is already set
+3. **Edit shortcut**: Click the edit button (pencil icon) to open the inline editor
+4. **Enter a number**: Type a number between 8 and 999 in the small input field
+5. **Save or Cancel**: Click the checkmark to save, X to cancel, or trash icon to remove existing shortcut
```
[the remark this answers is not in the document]
kaligrafy: yes we need to choose better icons and discuss this.

docs/preset_shortcuts.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Existing Systems
+
+The feature integrates with several existing iD systems:
+
+- **Preset Manager**: Uses existing preset lookup and validation
+- **Drawing Modes**: Leverages existing point/line/area drawing modes
+- **Action System**: Uses existing `actionChangePreset` for applying presets
+- **Localization**: All user-facing text supports translation
+- **Storage**: Uses existing preferences system for persistence
```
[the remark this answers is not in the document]
kaligrafy: It says it supports translation, the translations are in transidex, not here.

docs/preset_shortcuts.md
kaligrafy: just that the regex will refuse invalid shortcut characters, but this needs to be fixed (see above)

docs/preset_shortcuts.md
kaligrafy: yes this can be removed

docs/preset_shortcuts.md
kaligrafy: can be removed

docs/preset_shortcuts.md
kaligrafy: yes the documentation is, because I make too many typos when I write documentation, so prefer to let the LLM write it correctly once, but I agree there is too much stuff here

kaligrafy:
> Maybe I am on extreme case here, but for me "with shortcuts persisted in browser storage." would not be too useful and I would some want to import/export them (to oms.org user preferences? to/from text file)
> 
> Maybe I would expect them to be shared between different machines where I use iD, but people using different OSes may actively NOT want this
> 
> comment as I am far from sure should this comment about browser storage not being enough is making a good point or have I got on extreme side

I have a copy of this feature in my fork that saves presets and preferences into a json file that can be then re-uploaded in another instance. It works well, but may be for a separate PR, since it will save/restore all the prefs, not only the shortcuts.

Issue #8327, opened 2021-01-27
Display custom map data from WMS endpoints


1ec5:
#7510 added support for WMS tokens like `bbox`, `height`, and `proj` in custom backgrounds. This extends the custom background feature to be compatible with ArcGIS MapServers and ImageServers. However, for custom map data URL templates, iD only supports TMS and quadkey tokens and output in Mapbox Vector Tile format. It should also support WMS tokens and ArcGIS server JSON to make the feature compatible with more existing data sources.

The idea is merely to load ArcGIS FeatureServer layers in the same way that custom map data can already be loaded into iD: as a reference layer rather than a way to directly add data to OSM. I think further streamlining the workflow would be out of scope for this issue.

## Rationale

The code for resolving URL templates should be shared between the custom background and custom map data modules to improve code coverage and consistency.

Basic WMS support can be seen as a baby step compared to the full ArcGIS FeatureServer integration proposed in #4164, but I think there is value in just getting WMS support to parity with the existing support for TMS and local files. The U.S. mapping community has discovered many appropriately licensed sources of aerial imagery from local government agencies; these sources have been a key ingredient in promoting local community involvement. But many of these agencies also provide non-imagery layers that we’ve yet to tap into, even in a bespoke, manual mapping workflow.

For example, an Ohio state agency provides a [FeatureServer](https://geo.oit.ohio.gov/arcgis/rest/services/LBRS/Butler_LBRS/FeatureServer/) containing [public domain](http://gis3.oit.ohio.gov/LBRS/_downloads/docs/White%20Paper-LBRS_2011.pdf#page=4) address points for most of the state. Even though I’m not planning an automated import of this address data in the near future, I would like to be able to quickly look up the address of an individual building as I manually map it, since it would be nice to have this one address in the meantime, and my manual edit will make the building less likely to be included in an initial phase of a future building+address import. The same dataset comes with a [MapServer](https://geo.oit.ohio.gov/arcgis/rest/services/LBRS/Butler_LBRS/MapServer/), but it doesn’t label the address points and doesn’t have dynamic layers enabled, so I’m forced to fire up JOSM to load the FeatureServer. By comparison, a similar MapServer in Silicon Valley has saved us countless hours of manual hunting in our [POI import](https://wiki.openstreetmap.org/wiki/Santa_Clara_County,_California/Social_distancing_protocol_import) there, but only because it was correctly styled with labels in the first place.

[...]

PR #11274, opened 2025-08-03
Support bbox token for custom map data

honnip:
this is useful for services that export the features in geojson, I'm not sure if it is a common case

### DIFF
--- a/modules/services/vector_tile.js
+++ b/modules/services/vector_tile.js
@@ -95,6 +95,8 @@ function vtToGeoJSON(data, tile, mergeCache) {
 function loadTile(source, tile) {
     if (source.loaded[tile.id] || source.inflight[tile.id]) return;
 
+    var bbox = tile.extent.bbox();
+
     var url = source.template
         .replace('{x}', tile.xyz[0])
         .replace('{y}', tile.xyz[1])
@@ -104,7 +106,8 @@ function loadTile(source, tile) {
         .replace(/\{switch:([^}]+)\}/, function(s, r) {
             var subdomains = r.split(',');
             return subdomains[(tile.xyz[0] + tile.xyz[1]) % subdomains.length];
-        });
+        })
+        .replace('{bbox}', `${bbox.minX},${bbox.minY},${bbox.maxX},${bbox.maxY}`);
 
 
     var controller = new AbortController();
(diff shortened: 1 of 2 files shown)


modules/services/vector_tile.js
```diff
@@ -104,7 +106,8 @@ function loadTile(source, tile) {
         .replace(/\{switch:([^}]+)\}/, function(s, r) {
             var subdomains = r.split(',');
             return subdomains[(tile.xyz[0] + tile.xyz[1]) % subdomains.length];
-        });
+        })
+        .replace('{bbox}', `${bbox.minX},${bbox.minY},${bbox.maxX},${bbox.maxY}`);
```
k-yle, line 110: is the intention of this PR to support WMS urls? (#8327)

* If yes, shouldn't we use [the same projection logic](https://github.com/openstreetmap/iD/blob/91ee41bcae182e60e8a7211b38da783fcf9015b1/modules/renderer/background_source.js#L141-L159) as the background renderer? 
* If no, is there any documentation for the type of URL that you're trying to support? Otherwise there is a risk that implementing support for `{{bbox}}` using this format will make it more complicated to support #8327 in the future

honnip: I thought it was the same bbox as the one used by bus.osm.kr. It's just the coordinates of the corners of the bounding box (not projected?). But yes, it seems different from the WMS token..

tyrasd: > bus.osm.kr

At first glance, this API does not even return geojson, but instead some custom json payload with embedded lat/lon values. That wouldn't be something that the vector tiles module could consume.

matkoniecz:
> this is useful for services that export the features in geojson, I'm not sure if it is a common case


can you give example of one, so this PR can be tested?

(and to be sure that it solves real problem, not theoretical one)

matkoniecz:
I propose to close it as not testable and solving a theoretical problem

matkoniecz:
@honnip 

> this is useful for services that export the features in geojson, I'm not sure if it is a common case


can you give example of one, so this PR can be tested?

k-yle:
yeah i agree, i'm going to close this given the dicussion in this thread: https://github.com/openstreetmap/iD/pull/11274#discussion_r2250934051

PR #11281, opened 2025-08-04
Retain some ref:* tags on buildings during node extraction

### LINKED ISSUE #11266: Extract POI also extracts ref:GB:uprn from building way
### URL

https://www.openstreetmap.org/way/190715109

1. find a building where the POI info is also on the building area
2. right click and press Extract
3. observe that ref:GB:uprn has been moved to the POI instead of staying on the building area

wiki: https://wiki.openstreetmap.org/wiki/Key:ref:GB:uprn
there are 3,280,125 uses of ref:gb:uprn

2.35.3

Firefox

youssefelzedy:
# Fix: Retain all ref:* tags on building areas during extraction

**Fixes #11266**

## Problem
When extracting a node from a building area, country-specific reference tags like `ref:GB:uprn`, `ref:linz:building_id`, etc. were being transferred to the extracted point instead of remaining with the building polygon. This happened because the code only retained a hardcoded list of reference tags.

As mentioned in the issue, @k-yle  suggestion to add specific tags to `buildingKeysToRetain` doesn't work well.

## Solution
Replace the hardcoded approach with a regex pattern to retain all reference tags (any tag starting with `ref:`) on building areas during extraction. The fix adds `key.match(/^ref:.{1,}/)` to the existing building tag retention logic.

## What this fixes
-  Building reference tags remain on the building area after extraction
-  Works for all country-specific reference formats
-  The extracted point gets other appropriate tags but not building reference IDs
-  Behavior is consistent across all international reference tag formats

---

This approach is more robust and future-proof compared to maintaining a hardcoded list of country-

### DIFF
--- a/modules/actions/extract.js
+++ b/modules/actions/extract.js
@@ -86,7 +86,8 @@ export function actionExtract(entityID, projection) {
                 // don't transfer building-related tags
                 if (buildingKeysToRetain.indexOf(key) !== -1 ||
                     key.match(/^building:.{1,}/) ||
-                    key.match(/^roof:.{1,}/)) continue;
+                    key.match(/^roof:.{1,}/) ||
+                    key.match(/^ref:.{1,}/)) continue;
             }
             // leave `indoor` tag on the area
             if (isIndoorArea && key === 'indoor') {
--- a/modules/actions/extract.js
+++ b/modules/actions/extract.js
@@ -48,7 +48,7 @@ export function actionExtract(entityID, projection) {
 
         var keysToCopyAndRetain = ['source', 'wheelchair'];
         var keysToRetain = ['area'];
-        var buildingKeysToRetain = ['architect', 'building', 'height', 'layer', 'nycdoitt:bin'];
+        var buildingKeysToRetain = ['architect', 'building', 'height', 'layer', 'nycdoitt:bin', 'ref:GB:uprn', 'ref:linz:building_id'];
 
         var extractedLoc = d3_geoPath(projection).centroid(entity.asGeoJSON(graph));
         extractedLoc = extractedLoc && projection.invert(extractedLoc);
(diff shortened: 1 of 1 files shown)


tordans:
Are there tests that handle this case?

youssefelzedy:
> Are there tests that handle this case?

@tordans, What do you mean by test handle this case?
If you meant there is any test case for handling `ref:*` tags, no, I recently just added this regex pattern

On wiki: https://taginfo.openstreetmap.org/search?q=ref%3A
There are a lot of `ref` tags here, so we shouldn't add all of them manually

modules/actions/extract.js
```diff
@@ -86,7 +86,8 @@ export function actionExtract(entityID, projection) {
                 // don't transfer building-related tags
                 if (buildingKeysToRetain.indexOf(key) !== -1 ||
                     key.match(/^building:.{1,}/) ||
-                    key.match(/^roof:.{1,}/)) continue;
+                    key.match(/^roof:.{1,}/) ||
+                    key.match(/^ref:.{1,}/)) continue;
```
k-yle: sorry for not explaining this clearly, the goal of #11266 is just to add the 1 (or maybe 2) mentioned keys into the [`buildingKeysToRetain` list](https://github.com/openstreetmap/iD/blob/91ee41bcae182e60e8a7211b38da783fcf9015b1/modules/actions/extract.js#L51). 

We can't make a change that affects every [`ref:*`](https://osm.wiki/Key:ref#Extensions) key, since these keys have varying purposes:
- Some like [`ref:GB:uprn`](https://osm.wiki/Key:ref:GB:uprn) should stay with the building
- Others like [`ref:vatin`](https://osm.wiki/Key:ref:vatin) should stay with the extracted POI


So, we should only make a change to the tags that have been requested by someone

k-yle (changes requested):

modules/actions/extract.js
youssefelzedy: Oh, okay, no problem. I was also confused 😆. You mean we just need to add `ref:linz:building_id` and `ref:GB:uprn` now?

k-yle: yep I think just those two tags for now. if you want to think about a more sophisticated solution, i'd suggest discussing that in a new issue first.

k-yle (approved):
thanks!

PR #11282, opened 2025-08-04
Fix/clean_tags_spaces

### LINKED ISSUE #11276: Don't remove spaces after semicolons in fields that are not of the comma-separated type
Example: https://overpass-api.de/achavi/?changeset=169600340

[note] is a field of "any text" type. iD shouldn't remove spaces there. I think removing spaces should be restricted to comma-separated type fields like [cuisine] or [beauty].

Waqibsk:
Fixes #11276 

Removes spaces after semicolons only for specific comma-separated tags (e.g., `cuisine`, `beauty`, `payment:*`). 
And Preserves spacing for all other tags

### DIFF
--- a/modules/util/clean_tags.js
+++ b/modules/util/clean_tags.js
@@ -12,9 +12,13 @@ export function utilCleanTags(tags) {
 
 
     function cleanValue(k, v) {
-        function keepSpaces(k) {
-            return /_hours|_times|:conditional$/.test(k);
+
+        // Compact list: no space after semicolon
+        function isCommaSeparated(k) {
+             return /^(cuisine|beauty|payment:cash|payment:credit_cards|leading|trailing|both)$/.test(k);
         }
+
+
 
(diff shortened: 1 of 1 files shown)


Waqibsk:
Hello  :)

I wanted to know if my current approach looks valid:

I'm compacting semicolon-separated values only for specific known keys like cuisine, beauty, etc., using a regex inside **isCommaSeparated(k).**

Does this seem like the right direction for handling tag cleaning?

Also, is there an existing list of tags that are known to use comma-separated formats? I'd like to make sure we're handling all such cases consistently — currently i guess the build is failing due to mismatched spacing.

Would appreciate any feedback on this !

tyrasd:
>  if my current approach looks valid

not really, this just inverts the logic: Now he have a short and largely incomplete list of tags. There would be lots of tags missing for which the tag cleanup does make sense.

PR #11290, opened 2025-08-07
Add shortcut to toggle overlay layer visibility

### LINKED ISSUE #10972: Add shortcut to toggle visibility of all overlay layers
### Description

Similarly to the `Alt + W` shortcut that hides the data layer, it would be nice to have a shortcut to turn on/off the visibility of the overlay layers.
I suggest `Alt + Q` but be careful with `Alt` shortcuts on Firefox because the `Alt` key also reveal the top menu bar in some situations. Might be better to move these 2 shortcuts to `Shift`.
This was briefly discussed here for the Strava iD extension https://github.com/cmoffroad/id-strava-heatmap-extension/issues/12#issuecomment-2813112235 .

zandonella:
Added a shortcut to toggle the visibility of all overlay layers to `Shift + O`. This follows the `Shift + <key>` pattern set by newer shortcuts (e.g., `Shift + P`) and, as far as I'm aware, does not conflict with any browser shortcuts. I chose O as the key as it is adjacent to the existing `Shift + P` shortcut on QWERTY keyboards, and is an intuitive letter for "Overlays" to be on.

I used essentially the same logic as `toggleStreetSide` for `toggleAllOverlays` with some modifications due to the differences in layer structure. Because the list of selectable overlays has a variable number of overlays, I also had to modify `updateLayerSelections` to stop them from unchecking on rerender.

Closes #10972 

Keybind in keyboard shortcuts:
image

Visibility on (default):
image

Visibility toggled:
image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2413,6 +2413,7 @@ en:
         osm_data: "Toggle OpenStreetMap data"
         minimap: "Toggle minimap"
         highlight_edits: "Highlight unsaved edits"
+        overlays: "Toggle all overlays"
       selecting:
         title: "Selecting features"
         select_one: "Select a single feature"
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -80,6 +80,11 @@
             "text": "shortcuts.browsing.enable.streetlevel"
           },
           {
+            "modifiers": ["⇧"],
+            "shortcuts": ["O"],
+            "text": "shortcuts.browsing.display_options.overlays"
+          },
+          {
             "shortcuts": ["background.minimap.key"],
             "text": "shortcuts.browsing.display_options.minimap"
           },
--- a/modules/ui/sections/overlay_list.js
+++ b/modules/ui/sections/overlay_list.js
@@ -10,6 +10,9 @@ import { uiSection } from '../section';
 
 export function uiSectionOverlayList(context) {
 
+    let _overlaysHidden = false;
+    let _savedOverlays = [];
+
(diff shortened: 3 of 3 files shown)


tyrasd:
FWIW, the [iD Strava Heatmap](https://addons.mozilla.org/en-US/firefox/addon/id-strava-heatmap/) plugin adds `Shift+Q` as a keyboard shortcut for toggling the overlay layers. That has the small benefit of being usable with the same hand on the keyboard as one would use to toggle the OSM data layer (`Alt+W`). On the other hand, `Shift+O` can be easier to remember (O = **o**verlays). :thinking:

Nekzuris:
What's the point of a 2 hands shortcut?

And every mapper who need this shortcut are already familiar with `Shift + Q`.

k-yle (changes requested):
thanks for the PR. given https://github.com/openstreetmap/iD/issues/10972#issuecomment-5260977497, I'd say we should use <kbd>⇧ Shift</kbd> <kbd>Q</kbd> ?

matkoniecz:
@zandonella ping?

matkoniecz (changes requested):
> Visibility toggled:

can you recover it without pressing magical shortcut?

we will end with the same problem as https://github.com/openstreetmap/iD/issues/12805 - people clicking it by accident and not being aware how to undo it

iff we really need it then overlay panel needs button or other way to recover from it

------------------------------------------------------

(and if we start saving config state it gets worse, as user on mobile cannot click such shortcut at all)

Issue #8214, opened 2020-11-21
ID Dark Mode


CharliePlett:
Hey, could we please have a dark mode?

quincylvania:
Hi and welcome, @CharliePlett.

Dark mode definitely sounds cool. I mention it as a missing vision feature [in our accessibility documentation](https://github.com/openstreetmap/iD/blob/develop/ACCESSIBILITY.md#-vision-accessibility). The toughest part will likely be maintaining two UI styles that are both look and work great.

mxdanger:
@quincylvania Hi, I just wanted to add to this conversation and say that before we can begin working on a dark mode, all the CSS color values in ID should be replaced with CSS variables. Compatibility will not be a problem as every browser [supports CSS variables](https://caniuse.com/css-variables) except for IE, which ID has discontinued support for.

This makes standardization easier for the existing light theme and implementing dark mode would be as easy as setting up CSS variables that overwrites the light theme. This would allow for easily maintaining both themes side by side.

All the theme switching would be handled by the browser based on what the user has already selected as their theme preference for their OS. This does not require any custom JS theme switcher and is the recommended way for accessibility and respecting the user's choice. Both code snippets use light mode as the default fallback if the user has no preferred theme.

Example with media query `prefers-color-scheme` for changing the CSS variables. Related link: [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)
```css
:root {
    --color-text-primary: #000;
}
@media (prefers-color-scheme: dark) {
    :root {
        --color-text-primary: #fff;
    }
}
```

The meta `color-scheme` tag is also particularly useful as it changes the user agent stylesheet, and for ID it would be primarily used for dark themed scrollbars and plain form controls such as checkboxes or radio buttons. Related links: [web.dev](https://web.dev/color-scheme/), [CSSWG]( https://drafts.csswg.org/css-color-adjust/)
```html
<meta name="color-scheme" content="light dark">
```

quincylvania:
We haven't abandoned support for IE 11, though it's true we're not actively testing it nor expecting all features to work perfectly. I think we can still use CSS variables but have PostCSS convert them to legacy CSS at build time.

For dark mode, lots of sites have a Light/Dark/Match System selection control that I think works well. We could add one for iD in the Preferences pane.

1ec5:
> We haven't abandoned support for IE 11

By way of an update, we mostly if not completely dropped IE11 support in #8811.

[...]

PR #11292, opened 2025-08-07
Separate UI styling from JS

hlfan:
Make the accent separate from the general button style in preparation for #8214.

### DIFF
--- a/modules/ui/entity_editor.js
+++ b/modules/ui/entity_editor.js
@@ -1,4 +1,5 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
+import { select as d3_select } from 'd3-selection';
 import deepEqual from 'fast-deep-equal';
 
 import { presetManager } from '../presets';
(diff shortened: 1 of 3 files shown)


modules/ui/entity_editor.js
```diff
@@ -142,10 +142,10 @@ export function uiEntityEditor(context) {
...
                 // flash the button to indicate the preset changed
                 context.container().selectAll('.entity-editor button.preset-reset .label')
-                    .style('background-color', '#fff')
-                    .transition()
-                    .duration(750)
-                    .style('background-color', null);
+                    .classed('flash-bg', true)
+                    .on('animationend', function() {
+                        this.classList.remove('flash-bg');
+                    });
```
tyrasd: maybe using d3 wrapper functions could make this slightly more readable (as it mirrors the `.classed('flash-bg', true)` on the previous line)?

```suggestion
                    .on('animationend', function() {
                        d3_select(this).classed('flash-bg', false));
                    });
```

hlfan: done.

tyrasd (approved):

PR #11300, opened 2025-08-11
 fix: allow 0 as input value in layer tag

### LINKED ISSUE #11086: Don't make the Layer selector suddenly disappear when passing 0
"I'll just oh,

Image

change this from layer=-1 to oh, layer=1", I said to myself with confidence, as I placed my mouse upon the "⏶" button and proceeded to click.

You'll never guess what happened next.

Waqibsk:
Fixes #11086 

screenshots of the changes made:
<div>
Screenshot 2025-08-11 212839
->
Screenshot 2025-08-11 212702

</div>

### DIFF
--- a/modules/ui/fields/radio.js
+++ b/modules/ui/fields/radio.js
@@ -149,8 +149,7 @@ export function uiFieldRadio(field, context) {
         // Layer
         if (layer && showLayer) {
             if (!layerField) {
-                layerField = uiField(context, layer, _entityIDs, { wrap: false })
-                    .on('change', changeLayer);
+                layerField = uiField(context, layer, _entityIDs, { wrap: false });
             }
             layerField.tags(tags);
             field.keys = utilArrayUnion(field.keys, ['layer']);
@@ -222,13 +221,6 @@ export function uiFieldRadio(field, context) {
         dispatch.call('change', this, t, onInput);
     }
 
-
-    function changeLayer(t, onInput) {
-        if (t.layer === '0') {
-            t.layer = undefined;
-        }
-        dispatch.call('change', this, t, onInput);
-    }
 
 
     function changeRadio() {


tyrasd (approved):

modules/ui/fields/radio.js
```diff
@@ -223,13 +222,6 @@ export function uiFieldRadio(field, context) {
     }
 
 
-    function changeLayer(t, onInput) {
-        if (t.layer === '0') {
-            t.layer = undefined;
```
k-yle, line 228: The wiki [says](https://osm.wiki/Key:layer):

> 0 should not be used explicitly

does this sentance need to be updated? `layer=0` [is used 40,000 times](https://taginfo.openstreetmap.org/tags/layer=0), but usage continues to decrease. This PR might change that trend. 

just double checking so that we don't end up with [another situation like this](https://osm.wiki/ID/Controversial_Decisions)

tyrasd: I think the wiki is correct in that the default 0 value should not be specified explicitly in most cases. But there are some exception cases where it can be helpful to still add it for clarity: I see it sometimes used with indoor corridors and building passages (example [1](https://www.openstreetmap.org/way/627391825), [2](https://www.openstreetmap.org/way/159986443)) to make it clear that they are in fact on ground level.

The UI silently removing any 0 values made it impossible to map this information in these cases, and the behaviour was generally confusing when the field suddenly got emptied upon reaching or entering the zero value.

> usage continues to decrease. This PR might change that trend.

Do you think so? I wouldn't have expected this change to have a huge impact: there are probably not many mappers that relied on nudging the layer tag to zero using the +/- buttons in order to clear it. One can still manually remove the value by just deleting the content of the input field, or, what's probably most applicable in most use cases, by using the delete icon on the whole structure field.

PR #11304, opened 2025-08-13
fix layer buttons broken

k-yle:
#11300 removed the `onChange` event listener, so the buttons don't work anymore


tyrasd (approved):
Thanks, sorry for missing this bug in #11300

PR #11308, opened 2025-08-14
Add dark mode

### LINKED ISSUE #8214: ID Dark Mode

hlfan:
Fixes #8214.
This builds on the original work from @AntonKhorev to make it able to sync with the OSM preference. See https://github.com/openstreetmap/openstreetmap-website/pull/6323 for that.

Making the theme respond anytime results in variables like `--bg-filter` and `--bg-blend-mode` to not need to watch for media query changes in JS.
Other styling changes away from JS have already been made in #11292.

### DIFF
--- a/css/10_colors.css
+++ b/css/10_colors.css
@@ -1,0 +1,57 @@
+.ideditor {
+    --bg-color: #fff; /* ~Bootstrap .bg-body */
+    --bg-color-2: #f6f6f6; /* ~Bootstrap .bg-body-tertiary */
+    --bg-color-3: #ececec; /* ~Bootstrap .bg-body-secondary */
+    --active-bg-color: #f1f1f1; /* in the middle between --bg-color-2 and --bg-color-3 */
+    --border-color: #ccc; /* ~Bootstrap .border-secondary-subtle */
+    --transparent-border-color: #cccccc94; /* --border-color with alpha value to match rgba(0, 0, 0, .1) over --bg-color-2 */
+    --border-color-2: #ddd; /* mix 66% of --border-color with white */
+    --outset-color: #fcfcfc; /* brighter than --border-color with white */
+    --shadow-color: #bbb; /* mix 91.5% of --border-color with black */
+    --text-color: #333;
+    --muted-text-color: #aaa;
+    --passive-bg-color: #cccccc; /* = --border-color */
+    --passive-text-color: #888;
+    --widget-color: #222;
+    --widget-bg-color: #eee;
+    --nested-bg-color: #eff2f7;
+    --nested-highlight-color: #e3e8ef;
+    --nested-border-color: #ccd5e3;
+    --selected-bg-color: #e8ebff;
+    --warning-bg-color: #ffc; /* mix 60% of --warning-label-color with white */
+    --warning-label-color: #ffa;
+    --warning-highlight-color: #ff8; /* ~saturate(1.5) of --warning-label-color */
+    --warning-border-color: #fb2;
+    --warning-text-color: #b15500;
+    --warning-highlight-text-color: #7f3d00;
+    --warning-icon-color: #f90;
+}
+
+@media (prefers-color-scheme: dark) {
+    .ideditor {
+        --bg-color: #212529;
+        --bg-color-2: #2b3035;
+        --bg-color-3: #343a40;
+        --active-bg-color: #30353b;
+        --passive-bg-color: #41464b;
+        --border-color: #41464b;
+        --transparent-border-color: #41464b94;
+        --border-color-2: #2b2e32;
+        --outset-color: #54595d;
+        --shadow-color: #3b4045;
+        --text-color: #dee2e6;
+        --muted-text-color: #6e7276;
+        --widget-color: #e9ecef;
+        --widget-bg-color: #31373c;
+        --nested-bg-color: #2d2e33;
+        --nested-highlight-color: #33363b;
+        --nested-border-color: #3e434e;
+        --selected-bg-color: #3b3c49;
+        --warning-bg-color: #332701;
+        --warning-label-color: #554102;
+        --warning-highlight-color: #5f4100;
+        --warning-border-color: #7e3b00;
+        --warning-text-color: #e98c38;
+        --warning-highlight-text-color: #f0ac71;
+    }
+}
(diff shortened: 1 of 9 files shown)


css/10_colors.css
```diff
@@ -0,0 +1,177 @@
...
+    --error-bg-color: #300;
+    --error-label-color: #500;
+    --error-highlight-color: #600;
+    --error-border-color: #800;
+    --error-text-color: #e33;
+    --error-highlight-text-color: #e66;
+}
+
+@media (prefers-color-scheme: dark) {
+    .ideditor:not([data-theme="light"]) {
```
k-yle: if we're going to use [data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/How_to/Use_data_attributes), should they be documented in [./API.md](https://github.com/openstreetmap/id/blob/develop/API.md)? That file is currently a collection of every way that consumers can customise iD.

(for context) I see that the OSM website [already adds 2 `data-*` attributes to the iD container](https://github.com/hlfan/openstreetmap-website/blob/4dff9d60cec071b1eb6d58b7a6d19c5d887f305f/app/views/site/id.html.erb#L17-L18), but it seems like iD doesn't actually use these (?)

hlfan: These element datasets are used for the core context initialization, so setting it from there would be aligned better with the other options.

But since this PR is already quite large, I'd like to keep this CSS-only and have the JS api integration done separately.
I could see `.theme-light` or `.theme-dark` be set with
```js
iD.coreContext().theme(container.dataset.theme).init()
```
However, I don't feel comfortable with making changes there yet.

k-yle: Until now, the `data-` attributes were only used by the OSM-website to pass data from [one of their files](https://github.com/hlfan/openstreetmap-website/blob/4dff9d60cec071b1eb6d58b7a6d19c5d887f305f/app/views/site/id.html.erb#L18) to [another internal file](https://github.com/hlfan/openstreetmap-website/blob/4dff9d60cec071b1eb6d58b7a6d19c5d887f305f/app/assets/javascripts/id.js#L26). iD has no code related to these `data-` attributes. So this PR is introducing root `data-` attributes to the iD codebase for the first time.

---

but if your intention is to introduce some sort of `.theme()` method to `coreContext` which updates the `data-theme`, then there's no issue, because no one would directly interact with the `data-` attributes, so we don't need to document it.

hlfan: Do you think CSS classes would be better suited for this than data attributes?

k-yle: if `coreContext` can be used to set the theme, then imo it doesn't matter what method is used internally. maybe @tyrasd has further thoughts

css/80_app.css
```diff
@@ -2315,6 +2317,7 @@ input.date-selector {
 
 .form-field-input-restrictions .restriction-container {
     position: relative;
+    color: #333;
```
k-yle: a few lines below here is `.restriction-help`, which has the same bg/fg colours in light and dark mode, making it noticably brighter than everything else

just checking if that's intentional?

image

k-yle (commented):
I used this PR for a few edits and i think it looks really nice :) just my personal opinion, i don't know much about colour contrast guidelines in dark mode

one question about the design below:

css/80_app.css
hlfan: I noticed it too on adopting the original work, but I left it for now since it didn't seem as blinding. However, I can take another look at it.

css/10_colors.css
hlfan:
```diff
diff --git a/modules/core/context.js b/modules/core/context.js
index 3219ff60e..d3be29955 100644
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -419,12 +419,32 @@ export function coreContext() {
   };
 
 
+  /* Theme */
+  const availableThemes = ['dark', 'light'];
+  let _theme;
+  if (availableThemes.includes(context.initialHashParams.theme)) {
+    _theme = context.initialHashParams.theme;
+  }
+  function applyTheme() {
+    availableThemes.forEach(theme =>
+      context.container().classed(`theme-${theme}`, theme === _theme)
+    );
+  };
+  context.theme = function(val) {
+    if (!arguments.length) return _theme;
+    if (availableThemes.includes(val)) _theme = val;
+    else _theme = undefined;
+    applyTheme();
+    return context;
+  };
+
   /* Container */
   let _container = d3_select(null);
   context.container = function(val) {
     if (!arguments.length) return _container;
     _container = val;
     _container.classed('ideditor', true);
+    applyTheme();
     return context;
   };
   context.containerNode = function(val) {
```

That's what I've bodged together for the context side so far. Not including this in the PR so it can be squash-merged separately.

k-yle: cool, since it'll be controllable through coreContext I think we can resolve this thread, whether we use classes or `data-` attributes internally won't matter

hlfan:
I've been using https://pr-11308--ideditor.netlify.app/dist/iD.css for regular mapping too for a few weeks and couldn't find other light mode remains.

That said, I also don't know much about colour contrast guidelines in dark mode. But just from the hwb() way I converted the colors, at least the ones I added _should_ yield a higher contrast.

tyrasd (approved):
I found a few niche additional UI elements that still needed special styling for dark mode (fixed already; see added commits to the branch):



image



And I took the opportunity to slightly clean up the styling in a couple of places (e.g. removing some unused styles, reducing hardcoded colors in `80_app.css`, etc., see the remaining added commits). I hope that's fine.

hlfan:
> I hope that's fine.

Totally!
The `.field-help-nav-item:hover` background should probably be adjusted as well.

Also [Anton Khorev](https://github.com/AntonKhorev) should be credited too.

And as for the `coreContext` integration, please take a look at https://github.com/openstreetmap/iD/pull/11308#discussion_r2289893294.

Nekzuris:
Thank you for this!

But I think it lacks contrast, the background is too light.
[This](https://userstyles.world/style/15596/openstreetmap-dark-theme) is imo better.

tyrasd:
> the background is too light

My understanding is that this was based on the colors of the dark theme of the osm website, which is also not using a pitch-black background. Feel free to propose tweaks to the theme, though.

hlfan:
Pitch black on an HDR monitor looks worse, and while I do think the backgrounds could be a touch darker, having the colors lining up with the OSM website was a thing I left as is from the draft of @AntonKhorev.

hlfan:
Another file that should be updated is ACCESSIBILITY.md, as it mentions dark mode at its end.

PR #11313, opened 2025-08-15
Setup prettier and eslint to autoformat code

tordans:
See Slack chat for more. This is a example what this could look like.

1. Setup Prettier, ESLint, Editorconfig
2. Run formatting
3. Introduce `.git-blame-ignore-revs` for step 2

Also…
- Update VS Code setting recommendations
- Add a prettier plugin to auto format imports, so one does not have to think about this and it is the same across editors (this is why TS is added as a dependency)
- Update the editorconfig to simplify the setup
- The prettier setup uses [the new CLI](https://prettier.io/blog/2025/06/23/3.6.0) which is faster

Prettier also autoformats markdown files. As with everything related to Prettier and ESLint formatting, one has to give up a bit of control which will help to think more about the code than the formatting…

---

One consequence is, that existing PRs will have to run the formatting to align their code and then rebase. Did not test this workflow, yet.

Known TODOS

- [ ] Review CLI failure – but lets first talk about if this is something that will continue
- [ ] Test how existing PRs have to behave


tyrasd:
I'm not a fan of this, for similar reasons as outlined in https://github.com/openstreetmap/iD/pull/10820#pullrequestreview-2649264630: there is very little to gain from such a refactoring, it potentially breaks existing PRs and branches, and does not actually improve the quality of the code.

Maybe one exception:

I would be OK with doing a one-off reformatting of the couple of js files we have that use the non-standard 2-space indentation (as a `.git-blame-ignore-revs` commit). That would actually improve the codebase as it makes it consistently use the same indentation everywhere.

The changes to `.editorconfig` look fine, though. :+1:

PR #11316, opened 2025-08-16
Validation: non standard dashes in dash-sensitive tag values

### LINKED ISSUE #9735: convert non-standard dashes to standard dashes in the Opening Hours field
### URL

https://www.openstreetmap.org/changeset/138191928

Upload opening hours using a non-standard dash

the standard dash seems to be - ([Hyphen-Minus](https://unicodeplus.com/U+002D))
other similar dashes:
`‐` HYPHEN 
`‑` NON-BREAKING HYPHEN
`‒` FIGURE DASH
`–` EN DASH
`﹘`SMALL EM DASH
`۔‎`  ARABIC FULL STOP
`⁃` HYPHEN BULLET
`˗` MODIFIER LETTER MINUS SIGN
`−` MINUS SIGN
`➖` HEAVY MINUS SIGN
`Ⲻ` COPTIC CAPITAL LETTER DIALECT-P NI
(similar characters found from https://www.unicode.org/Public/security/15.0.0/confusables.txt, there may be more)

this leads to parsing fails in some applications, eg. also see https://github.com/organicmaps/organicmaps/issues/3888. it should also be handled at the application level, but it's good to have standard data as well.

_No response_

_No response_

_No response_

Waqibsk:
Fixes: #9735

This PR adds a new validation module that detects non-standard dashes in OSM tag values where from–to syntax (ranges, schedules, conditionals) is commonly used.

<li>Detects invalid dash characters (e.g., EN DASH, EM DASH, MINUS SIGN, etc.) in dash-sensitive keys such as opening_hours, service_times, collection_times, and *:conditional.
<li>
Provides an auto-fix option allowing users to replace all non-standard dashes with the standard ASCII hyphen-minus (-).


tordans:
Do we typically have software tests for those kind of validators?

modules/validations/invalid_dashes.js
```diff
@@ -0,0 +1,150 @@
...
+function isDashSensitiveKey(key) {
+  return DASH_SENSITIVE_KEYS.has(key) || key.endsWith(':conditional');
+}
+
+    // Regex for all non-standard dashes and similar characters
+//  ~        → Tilde, sometimes mistaken for a dash
+//  \u2010   → Hyphen (‐)
+//  \u2011   → Non-breaking hyphen
+//  \u2012   → Figure dash (‒)
+//  \u2013   → En dash (–)
```
1ec5: Also the em dash (U+2014).

modules/validations/invalid_dashes.js
```diff
@@ -0,0 +1,150 @@
...
+        var issues = [];
+        function showReferenceDash(selection) {
+            selection.selectAll('.issue-reference')
+                .data([0])
+                .enter()
+                .append('div')
+                .attr('class', 'issue-reference')
+                .call(t.append('issues.invalid_dashes.reference'));
+        }
+    const DASH_SENSITIVE_KEYS = new Set([
```
1ec5: Lists like this go in modules/osm/tags.js. Once openstreetmap/id-tagging-schema#1537, we can remove this list in favor of checking for a field `type` of `schedule`.

modules/validations/invalid_dashes.js
```diff
@@ -0,0 +1,150 @@
...
+//  \u02D7   → Modifier letter minus sign (˗)
+//  \u2212   → Mathematical minus sign (−)
+//  \u2796   → Heavy minus sign (➖)
+//  \u2CBA   → Coptic capital letter sampi (Ⲻ),
+
+var invalidDashRegex = /[~\u2010\u2011\u2012\u2013\uFE58\u06D4\u2043\u02D7\u2212\u2796\u2CBA]/;
+
+
+        function containsInvalidDash(text) {
+            return invalidDashRegex.test(text);
```
1ec5: Fancy dashes are valid within [comments](https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:comment), so don’t flag them. Let’s test the case where the value contains fancy dashes both inside and outside a comment – the automatic fix should replace the fancy dash outside the comment but not the one inside the comment.

/ref https://github.com/openstreetmap/iD/issues/9735#issuecomment-1623895198

Waqibsk:
Hey, I’ve added the tests for this validator in `test/spec/validations/invalid_dashes.js`. Let me know if there’s anything else I should update or improve :)

modules/validations/invalid_dashes.js
```diff
@@ -0,0 +1,150 @@
...
+                    deleteOnClick = function(context) {
+
+                        var id = this.issue.entityIds[0];
+                        var operation = operationDelete(context, [id]);
+                        if (!operation.disabled()) {
+                            operation();
+                        }
+                    };
+                }
+fixes.push(
```
1ec5: There’s some confusing indentation starting here.

modules/validations/invalid_dashes.js
1ec5: Similarly, if this validator rule looks at `*:conditional=*`, then it should only flag fancy dashes inside of conditions (after the `@`, before the next `;`).

tordans: @1ec5 for the second commend, do you have an example string that you have in mind?

1ec5: The conditional restriction syntax can technically apply to any key, including keys that accept freeform text, such as [`source:maxspeed:conditional=*`](https://taginfo.openstreetmap.org/keys/source%3Amaxspeed%3Aconditional) and [`note:access:conditional=*`](https://taginfo.openstreetmap.org/keys/note%3Aaccess%3Aconditional).

That said, OSM currently only has elements with [invalid en and em dashes in conditional tags](https://overpass-turbo.eu/s/2a0H). For example, [this way](https://www.openstreetmap.org/way/1387600593/history/1) is tagged:

```
access:conditional=no @ (2025 May 17–2025 Dec 31)
```

tordans: Fascinating, OSM always finds new ways to make a data consumers live harder :-). Instead of an ISO 8601 date format `2025-05-17` [the Wiki actually documents](https://wiki.openstreetmap.org/wiki/Conditional_restrictions) a natural-language variant `2025 May 17`.
> `motor_vehicle:conditional=no @ (2018 May 22-2018 Oct 07)`
> `motorcycle:conditional=yes @ Jun 01-Oct 01`

I would argue though, that all those should use a hyphen-minus as well. Everything else just makes parsing those strings way to complex.
Same for `bicycle:conditional=yes @ (20:00-09:00)` and similar.

Meaning, I thing we should apply the validation and replacement on the whole value.

1ec5: As I said, the en and em dashes currently in the database are invalid. But there’s no reason why a source or note can’t have an en or em dash in it. So we’d either need to exclude each of those possible keys or, more robustly, limit the check to just the condition part of a conditional tag. A regular expression like `@[^(;]+|@ *\([^)]+)` should be able to find conditions. (It casually looks for both `value @ condition; …` and `value @ (condition); …`.) Then we’ll need an additional step to exclude any comment.

modules/validations/invalid_dashes.js
```diff
@@ -40,20 +35,38 @@ function isDashSensitiveKey(key) {
 //  \u2212   → Mathematical minus sign (−)
 //  \u2796   → Heavy minus sign (➖)
 //  \u2CBA   → Coptic capital letter sampi (Ⲻ),
+//  \u2014   → EM DASH (—)
+
+var invalidDashRegex = /[~\u2010\u2011\u2012\u2013\uFE58\u06D4\u2043\u02D7\u2212\u2796\u2CBA\u2014]/;
+        function replaceInvalidDashesOutsideComments(text) {
```
1ec5: Something strange is happening with the indentation in your editor. Can you use your code editor’s reformat/reindent functionality to tidy up your additions to the codebase, or look into why it isn’t doing that automatically for you? Thank you.

modules/validations/invalid_dashes.js
tordans: Now I understand. Had to re-read the "comment" concept for `opening_hours` (which is quite "interesting" TBH).
I am personally not sure if we actually need to cover this edge case. One can always use the ignore option for the validation warning and changing the dashes would IMO not wrangle the comment in any meaningful way. But of course, the cleanest way would be to go the regex path…

1ec5: Personally, I think we should tread a little more carefully on validator warnings that touch tags that iD doesn’t even support – such as `*:conditional=*` – because the user won’t have a meaningful opportunity to review their changes. If this overcomplicates the task, we can always omit support for `*:conditional=*` for now and leave it as tail work for another PR.

modules/validations/invalid_dashes.js
```diff
@@ -40,20 +35,38 @@ function isDashSensitiveKey(key) {
...
+            let match;
+
+            while ((match = regex.exec(text)) !== null) {
+                if (match[0].startsWith('"')) {
+                    result += match[0];
+                } else {
+                    // replace invalid dashes outside quotes
+                    result += match[0].replace(invalidDashRegex, '-');
+                }
+            }
```
1ec5: As far as I can tell, this isn’t replacing the dash in `foo–foo` (the case without any comments) or `"foo–bar" foo–foo`.

Please add some tests to verify various edge cases that could arise with this code.

modules/validations/invalid_dashes.js
1ec5: We could also leave comments as tail work if necessary, as long as the community doesn’t get the impression that the comment syntax is restricted beyond what’s in the opening hours syntax, just because iD is replacing certain characters. For instance, we don’t expect the Arabic full stop in the machine-readable parts of opening hours syntax, but in an Urdu-speaking country, it could appear in a comment (to denote [an abbreviation](https://en.wikipedia.org/wiki/Full_stop#Shahmukhi)).

I realize this is getting into hypotheticals, but again we should be careful because of iD’s very limited support for the opening hours syntax outside of validation. The idea here was to chip off just the aspects of validation that can safely occur before we have proper user-friendly editing support for opening hours syntax.

Anyhow, I think we’re already pretty close with this PR, so let’s just make sure the comment-related code is working as intended.

Waqibsk:
Does ID supports nested comments(eg:  `" foo"bar"foo "` ) in tags ?
in that case .... the current approach of replacing dashes will break

tordans:
> Does ID supports nested comments(eg: `" foo"bar"foo "` ) in tags ? in that case .... the current approach of replacing dashes will break

I don’t understand the question. Can you share the full tag example? 

The linked opening hour wiki page has a section on the „comment“ part of the tag value and has an „allowed chars“ reference that defines `"` as disallowed.

Waqibsk:
Sorry for the confusion ,

I was asking whether a user could include double quotes "" inside a comment in a tag value. For example:
`opening_hours:  Mo-Fr 08:00-12:00 open "service for "VIP" members"`

Would that be allowed??

tordans:
> I was asking whether a user could include double quotes "" inside a comment in a tag value. For example: `opening_hours: Mo-Fr 08:00-12:00 open "service for "VIP" members"`
> 
> Would that be allowed??

It can happen, because we only have conventions an no real validation for values anywhere in OSM.
But in this case, the Wiki is quite explicit (which it is not often…)
> <comment_character>	Any character except `"`.	

Source: https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:comment

So the app should not crash or perform actions that "break" the data, but other than that we don't have to handle this case IMO.

modules/osm/tags.js
```diff
@@ -373,3 +373,9 @@ export var osmSummableTags = new Set([
     'parking:left:capacity',
     'parking:left:capacity'
 ]);
+export  const DASH_SENSITIVE_KEYS = new Set([
```
tyrasd: we typically use camel casing for exported constants:

```suggestion
export  const dashSensitiveKeys = new Set([
```

modules/validations/invalid_dashes.js
```diff
@@ -0,0 +1,181 @@
...
+                .data([0])
+                .enter()
+                .append('div')
+                .attr('class', 'issue-reference')
+                .call(t.append('issues.invalid_dashes.reference'));
+        }
+
+        function isDashSensitiveKey(key) {
+            return DASH_SENSITIVE_KEYS.has(key) || key.endsWith(':conditional');
+        }
```
tyrasd: It would be cleaner to move this method to `osm/tags.js` and export it instead of the tag set.

tyrasd (changes requested):
I started looking through the code of this PR, and noticed some minor issues (e.g. `dist` files should not be committed, and the two inline comments below). But then I looked through the affected [OSM data](https://overpass-turbo.eu/s/2a8J) and noticed that the original issue is a little more complicated: most cases where non-ascii dashes are used are from situations where the tags are not matching the opening hours syntax at all (e.g. because they were copy-pasted from a website, or when entering manually not really knowing the exact specifications).

In those cases, simply replacing the dashes does not fix the problem, and arguably makes the data quality problem even slightly worse. Here are some typical examples: [1](https://www.openstreetmap.org/node/12578121900), [2](https://www.openstreetmap.org/node/13063912526), [3](https://www.openstreetmap.org/node/12949138717), [4](https://www.openstreetmap.org/node/13056292242), [5](https://www.openstreetmap.org/node/12638679035), [6](https://www.openstreetmap.org/node/12463528583), [7](https://www.openstreetmap.org/way/490289844), [8](https://www.openstreetmap.org/node/12958943652), [9](https://www.openstreetmap.org/node/2749599320), [10](https://www.openstreetmap.org/way/1414973629).

That said, there exist also a decent amount of cases where the only mistake in the tag is the use of the wrong dash (examples: [11](https://www.openstreetmap.org/node/12776380806), [12](https://www.openstreetmap.org/node/12953794227), [13](https://www.openstreetmap.org/node/12983649896)). 

Therefore, I think a better way to approach this would be the following:

1. include [opening_hours.js](https://www.npmjs.com/package/opening_hours)
2. parse the tag using the library (note that for some tags, the `mode` has to be set to `2`)
3. if the library cannot parse the string, there will be an exception raised. in this case it might be better to just give up
4. if the tag can be parsed, but has warnings that the library can fix (i.e. `hours.getWarnings().length > 0 && hours.prettifyValue() !== <tagvalue>`), offer the "prettified" value as a fix
5. (optional) it seems like opening_hours.js does not check for all of the mentioned dash-alternatives in its parser. For example, it knows about tilde and en/em-dashes, but apparently not the more fancy ones. It would be great if those could be added [upstream](https://github.com/opening-hours/opening_hours.js) to the library as a PR.

This approach would have the additional benefit of being a first step towards #974, as that also requires parsing the tag values.

modules/validations/invalid_dashes.js
k-yle: > [^](https://github.com/openstreetmap/iD/pull/11316#discussion_r2280735877)  ...  (after the `@`, before the next `;`)

not just `;`, `AND` is also a valid operator in the conditional syntax right?  

[example here](https://kyle.kiwi/osm-conditional-restrictions/#yes+@+(09:00-17:00+AND+weight+%3C+3.5)) (but don't trust [this library](http://github.com/k-yle/osm-conditional-restrictions), the wiki is [pretty ambiguous](https://osm.wiki/Conditional_restrictions#Tagging:_full_scheme) about the permitted syntax)

1ec5:
> This approach would have the additional benefit of being a first step towards https://github.com/openstreetmap/iD/issues/974, as that also requires parsing the tag values.

If we lead with a validator rule before a GUI, my concern would be that we could easily trap the user with a warning about an invalid `opening_hours=*` value that we won’t help them fix. But if opening_hours.js’s “prettified” value would be a reliable replacement in common cases, then that would be a compelling argument for adding a warning based on that validator.

As I mentioned [on the forum](https://community.openstreetmap.org/t/automated-edit-proposal-convert-non-standard-dashes-to-standard-dashes/110206/15), a syntax issue requiring esoteric tagging knowledge should really be the focus of a MapRoulette challenge, with any in-editor validator being merely a backup. That said, OpenHistoricalMap’s fork of iD took the same approach you describe for the similarly complex Extended Date/Time Format: OpenHistoricalMap/iD#187 OpenHistoricalMap/iD#188.

Waqibsk:
>include [opening_hours.js](https://www.npmjs.com/package/opening_hours)

Should I implement this approach?
It works for `opening_hours`, `service_times`, and `collection_times`, but in that case we won’t be able to handle `*:conditional` tags

tyrasd:
@Waqibsk 

> Should I implement this approach?

That's up to you. As outlined above, the current simple search-replace approach is too limited to really be useful. I _think_ the `opening_hours.js` solution is not too much more complex to implement if you follow the steps 1 to 4/5 above. But I can understand if you say that's outside of what you like to do.

> but in that case we won’t be able to handle *:conditional tags

It would be fine to omit the `*:conditional` tags for now. Although, supporting them as well would "only" require a little extra pre-processing to fish out the conditions from the tags. For example like the following: `for (const rule of tag.split(';')) { const condition = rule.match(/@(.*)/)[1]); }` should work in principle. Then one could only feed in the _condition_ parts of the tag into `opening_hours.js` and do the rest the same as for the other tags.

Waqibsk:
>there will be an exception raised. in this case it might be better to just give up

 should we also create a warning for this (e.g. “invalid format”) so the user gets feedback rather than silence?

tyrasd:
I'd say it is better to ignore such cases for now (as explained by @1ec5 [above](https://github.com/openstreetmap/iD/pull/11316#issuecomment-3201271215)): the warning would only be confusing for most iD users as they would not have an easily actionable way to fix the error. There are external QA tools (e.g. osmose / maproulette) that can handle the more complicated issues.

1ec5:
Yeah, we can treat a more comprehensive warning as tail work and track it in a separate issue. That would be an opportunity for others to come up with, say, natural language parsers etc. to guess the intended meaning and offer suggestions.

matkoniecz:
> most cases where non-ascii dashes are used are from situations where the tags are not matching the opening hours syntax at all (e.g. because they were copy-pasted from a website, or when entering manually not really knowing the exact specifications).

> In those cases, simply replacing the dashes does not fix the problem, and arguably makes the data quality problem even slightly worse. Here are some typical examples: [1](https://www.openstreetmap.org/node/12578121900), [2](https://www.openstreetmap.org/node/13063912526), [3](https://www.openstreetmap.org/node/12949138717), [4](https://www.openstreetmap.org/node/13056292242), [5](https://www.openstreetmap.org/node/12638679035), [6](https://www.openstreetmap.org/node/12463528583), [7](https://www.openstreetmap.org/way/490289844), [8](https://www.openstreetmap.org/node/12958943652), [9](https://www.openstreetmap.org/node/2749599320), [10](https://www.openstreetmap.org/way/1414973629).

would it make sense to accept this code for some subset of tags, that would exclude opening_hours-like fields?

matkoniecz:
@Waqibsk have you seen this comment? What you think about restricting this PR a not and avoid oh-related complexity for now?

Waqibsk:
Will it really help , if we apply this validation to other tags (which are not `opening_hours`-like ) ? I think the main issue was regarding `opening_hours` like tags only

1ec5:
#9735 requested this feature for opening hours keys specifically. Did we address the issue with comments inside opening hours? If so, I think we can scope the PR to just opening hours and similar keys. Conditional tagging would be nice to have, but it’s more difficult and possibly less well defined.

test/spec/validations/invalid_time_range.js
```diff
@@ -0,0 +1,105 @@
...
+        expect(issue.entityIds[0]).to.eql('w-1');
+    });
+
+    it('ignores correct syntax', function () {
+        createWay({ opening_hours: ' Mo-Fr 10:00-18:00 ' });
+        var issues = validate();
+        expect(issues).to.have.lengthOf(0);
+    });
+
+    it('flags multiple  syntax errors in time range keys', function () {
```
matkoniecz, line 59:
```suggestion
    it('flags multiple syntax errors in time range keys', function () {
```

double space nitpick

test/spec/validations/invalid_time_range.js
```diff
@@ -0,0 +1,105 @@
...
+        var replaceFix = issues[0]
+            .dynamicFixes(context)
+            .find(
+                (f) => f.title.stringId === 'issues.fix.fix_time_range.title'
+            );
+        replaceFix.onClick.call({ issue: issues[0] }, context);
+        var entity = context.hasEntity('w-1');
+        expect(entity.tags.opening_hours).to.eql('Mo-Fr 10:00-18:00');
+        expect(entity.tags.service_times).to.eql('10:00-20:00');
+    });
```
matkoniecz, line 104: can you add test for `Mo-Fr 10:00-18:00 "comment including – (some special dash) which is valid syntax"` confirming it will not raise complaints?

see https://openingh.openstreetmap.de/evaluation_tool/?EXP=Mo-Fr+10%3A00-18%3A00+%22comment+including+%E2%80%93+%28some+special+dash%29+which+is+valid+syntax%22&lat=48.7769&lon=9.1844&mode=0

test/spec/validations/invalid_time_range.js
matkoniecz: @Waqibsk

test/spec/validations/invalid_time_range.js
Waqibsk: really sorry for the delay, i have added the tests

data/core.yaml
```diff
@@ -1634,6 +1635,13 @@ en:
         message: '{feature} has an invalid email address'
         message_multi: '{feature} has multiple invalid email addresses'
         reference: 'Email addresses must look like "user@example.com".'
+    invalid_time_range:
+      title: Invalid time range syntax 
+      tip: Fix the invalid time range syntax 
+      message: '{feature} contains invalid time range  '
```
matkoniecz, line 1641: is it possible to phrase it in more human-readable form more explaining what got broken?

Is it intended to cover far more than dashes in future? Maye mention non-standard dashes here specifically?

PR #11319, opened 2025-08-17
Add Korean phone number format

aaei924:
I added South and North Korean phone number format.
Region codes are for Seoul(02) and Pyongyang(02).


data/phone_formats.json
```diff
@@ -36,6 +36,8 @@
   "jm": "+1-876-555-1234",
   "jp": "+81-3-1234-5678",
   "kn": "+1-869-555-1234",
+  "kp": "+850-2-123-4567",
+  "kr": "+82 2-123-4567",
```
tyrasd: for South Korea: if spaces are preferred (ITU-T/DIN formatting), then it should be the following, right?

```suggestion
  "kr": "+82 2 123 4567",
```

local example: https://www.openstreetmap.org/node/8398939870

aaei924: Yeah, it can also be written even without hyphens. However, when Korean people write phone number, especially in domestic format, using hyphen is much more preferred in local convention.

tyrasd: I'm just asking because the OSM wiki (https://wiki.openstreetmap.org/wiki/Key:phone) recommends to use either the `phone=+<country code> <area code> <local number>` or the format with all dashes `+<country code>-<area code>-<local number>`. I would like to avoid showing a format that is not recommended by the OSM wiki in the placeholder text in the field in iD. :blush: 

Looking through [taginfo](https://taginfo.geofabrik.de/asia:south-korea/keys/phone#values), it seems like in South Korea the version without hyphens seems to be the most used one. Let's go with that one.

tyrasd (approved):

data/phone_formats.json
aaei924: The reason mappers have used phone number without hyphen is that the description in Wiki is saying to use formats as '+CC XXX XXX XXX' (without hyphen). People can easily think it is compelled that they should input values without hyphen.
This is not a big problem because there's no difference in actual dialling between with or without hyphen. However, according to my whole life in Korea, it's difficult for me to agree that writing phone number without hyphen is widely used.

tyrasd: Hmm… I looked around a bit on some websites of Korean businesses: it turns out only very few include the international `+82` prefix. But the few examples I found that do, use the format with all hyphens; examples: [1](https://www.ewha.ac.kr/ewhaen/index.do), [2](https://www.trazy.com/spot/40/beer-oclock-nightlife-bar), [3](https://www.muhann.com/eng/main/main.html#location).

In that case, it would probably be better to suggest to use the format as `+82-2-123-4567`. Would that make sense?

aaei924: Yes, that one makes sense.

PR #11323, opened 2025-08-19
process `terms` and `aliases` as arrays

k-yle:
Blocked by https://github.com/ideditor/schema-builder/pull/227, this would need to be merged at the same time that we upgrade to id-tagging-schema v7


matkoniecz:
maybe it would make sense to make milestone for that id-tagging-schema v7 PRs?

tyrasd:
this is now included directly in #12358

PR #11324, opened 2025-08-19
chore: update eslint config to support TS files

k-yle:
we have a small number of typescript files, but the current eslint config doesn't validate these files. This PR also sets up the CI to run typechecking on those files. 


tyrasd (approved):
Thanks!

PR #11327, opened 2025-08-20
directly handle "emergency" access values as area-key exceptions

tyrasd:
…instead of relying on the presence of dummy presets in the tagging schema.

This would unblock https://github.com/openstreetmap/id-tagging-schema/pull/1680, see also that issue for background.


k-yle (approved):
I think there's nothing blocking this PR, right?

PR #11332, opened 2025-08-21
configure preview deployments for Pull Requests

k-yle:
Every PR now gets deployed to netlify, similar to how it works in [openstreetmap/id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema).

In this repo, we use the [GitHub Deployments API](https://docs.github.com/en/rest/deployments/deployments) instead of posting a PR comment, just to reduce noise and notifications. The link to the deployed version is still shown in the PR page:

Untitled

Demo: [PR #1](https://github.com/k-yle/id-deploy-test/pull/1) and [PR #2 (from a fork)](https://github.com/k-yle/id-deploy-test/pull/2). It won't work for this PR itself, it'll only work once this is merged

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,15 +21,39 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['22', '24']
+        node-version:
+          - { v: 22 }
+          - { v: 24, isLatest: yes }
 
     steps:
     - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-    - name: Use Node.js ${{ matrix.node-version }}
+    - name: Use Node.js ${{ matrix.node-version.v }}
       uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: ${{ matrix.node-version.v }}
     - run: npm clean-install
     - run: npm audit signatures
     - run: npm run all
     - run: npm run test
+
+    - name: '[PR Preview] Upload dist folder'
+      if: ${{github.event_name == 'pull_request' && matrix.node-version.isLatest}}
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      with:
+        name: dist
+        path: |
+          dist
+
+    - name: '[PR Preview] Store PR metadata'
+      if: ${{github.event_name == 'pull_request' && matrix.node-version.isLatest}}
+      run: |
+        touch ./pr_metadata.ini
+        echo "pr=${{github.event.number}}" >> ./pr_metadata.ini
+        echo "ref=${{github.sha}}" >> ./pr_metadata.ini
+
+    - name: '[PR Preview] Upload PR metadata'
+      if: ${{github.event_name == 'pull_request' && matrix.node-version.isLatest}}
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      with:
+        name: pr_metadata
+        path: ./pr_metadata.ini
--- a/scripts/pr_preview.js
+++ b/scripts/pr_preview.js
@@ -1,0 +1,52 @@
+// @ts-check
+
+/** @param {import('@actions/github-script').AsyncFunctionArguments} _ */
+export function afterDeploy({ core, github, context }) {
+    /**
+     * @param {string} pr
+     * @param {string} ref
+     */
+    return async (pr, ref) => {
+        core.info(`Creating deployment event in PR #${pr} for '${ref}'…`);
+        const deployment = await github.rest.repos.createDeployment({
+            auto_merge: false,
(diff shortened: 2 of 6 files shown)


scripts/pr_preview.js
```diff
@@ -0,0 +1,50 @@
...
+            environment: `pr-preview-${pr}`,
+            transient_environment: true,
+            description: `Preview deployment for Pull Request #${pr}`,
+            required_contexts: [],
+        });
+
+        core.info('Marking deployment as complete…');
+        await github.rest.repos.createDeploymentStatus({
+            state: 'success',
+            environment_url: `https://pr-${pr}--ideditor-preview.netlify.app`,
```
k-yle:
```suggestion
            environment_url: `https://pr-${pr}--ideditor.netlify.app`,
```
this needs to be changed before merging. The current URL is just for the demo

eslint.config.js
```diff
@@ -30,7 +30,6 @@ export default [
        'indent': ['off', 4],
        'keyword-spacing': 'error',
        'linebreak-style': ['error', 'unix'],
-       'no-await-in-loop': 'error',
```
hlfan: I'd rather have `await Promise.all(artifacts.filter().map())` than the removal of this rule.

.github/workflows/build.yml
```diff
@@ -21,15 +21,39 @@ jobs:
...
-        node-version: ${{ matrix.node-version }}
+        node-version: ${{ matrix.node-version.v }}
     - run: npm clean-install
     - run: npm audit signatures
     - run: npm run all
     - run: npm run test
+
+    - name: '[PR Preview] Upload dist folder'
+      if: ${{github.event_name == 'pull_request' && matrix.node-version.isLatest}}
+      uses: actions/upload-artifact@v4
```
tyrasd: we should now stick to always pin the exact sha hash for the exact version we want to use, see #11103

```suggestion
      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```

(there is one more occurrence of this in this file and a couple more `uses` in `deploy-pr-preview.yml` )

scripts/pr_preview.js
tyrasd: Is there anything else that needs to be configured for this to work on the `ideditor.netlify.app` project? IIRC not, but let me know if there's something to change.

tyrasd (commented):
Thanks, this looks great with the gh deployments integration! 

Other than the small question/suggestion below (and @hlfan's suggestion to keep the no-await-in-loop linter rule), it's looking really good. Feel free to swap over to the final deployment URL when you think it's ready to be merged. :relaxed:

scripts/pr_preview.js
k-yle: URL changed. 

I don't think there's any extra configuration required, because I just created a netlify account yesterday, and the demo worked immediately with no settings changed in the netlify portal.

scripts/pr_preview.js
```diff
@@ -0,0 +1,50 @@
...
+        await Promise.all(artifacts.data.artifacts.map(async (artifact) => {
+            if (artifact.name === 'dist' || artifact.name === 'pr_metadata') {
+                core.info('Cleaning up artifact…');
+                await github.rest.actions.deleteArtifact({
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    artifact_id: artifact.id
+                });
+            }
+        }));
```
hlfan: Always logging deletions might be better than having to count the clean-up lines:
```suggestion
        const deletions = artifacts.data.artifacts
            .filter(artifact => artifact.name === 'dist' || artifact.name === 'pr_metadata')
            .map(artifact =>
                github.rest.actions.deleteArtifact({
                    owner: context.repo.owner,
                    repo: context.repo.repo,
                    artifact_id: artifact.id
                }));
        core.info(`Cleaning up ${deletions.length} artifacts…`);
        await Promise.all(deletions);
```

k-yle: sure, done

tyrasd (approved):

k-yle:
oh, it [failed](https://github.com/openstreetmap/iD/actions/runs/17159017019/job/48683204240) because the `NETLIFY_*` env variables aren't accessible, @tyrasd are these variables limited to the `develop` branch at the moment? I don't have access to the corresponding settings page to check

tyrasd:
yeah, I just realized that I have set up the secrets as github "environments", which needs to be specified for the respective workflows, see 2e71cbe54406cf7bd2301374ad9d591ad2a35c65. It works now (example PR: https://github.com/openstreetmap/iD/pull/11334). :bow:

tyrasd:
FYI: I just configured the API key for iD on the OSM sandbox / test instance https://api06.dev.openstreetmap.org/ to allow the use on `https://pr-…--ideditor.netlify.app/` domains. Unfortunately, the OSM API does not allow to use wildcards in the landing URL for oauth2 applications, so I have manually whitelisted a whole bunch on domains (currently for pr ids 11388..11999), allowing to test also uploads from the PR preview instances of iD (after switching to the dev API connection). Feel free to ping me when additional subdomains are needed.

PR #11333, opened 2025-08-21
support discarding tags, not just keys

k-yle:
corresponding PR for https://github.com/ideditor/schema-builder/pull/231

### DIFF
--- a/modules/actions/discard_tags.js
+++ b/modules/actions/discard_tags.js
@@ -13,7 +13,8 @@ export function actionDiscardTags(difference, discardTags) {
 
       for (let i = 0; i < keys.length; i++) {
         const k = keys[i];
-        if (discardTags[k] || !entity.tags[k]) {
+        const v = entity.tags[k];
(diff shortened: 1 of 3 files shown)


modules/actions/discard_tags.js
```diff
@@ -13,7 +13,8 @@ export function actionDiscardTags(difference, discardTags) {
 
       for (let i = 0; i < keys.length; i++) {
         const k = keys[i];
-        if (discardTags[k] || !entity.tags[k]) {
+        const v = entity.tags[k];
+        if (discardTags[k] === true || (typeof discardTags[k] === 'object' && discardTags[k][v]) || !entity.tags[k]) {
```
tyrasd, line 17: Btw: At first, I was wondering what the `|| !entity.tags[k]` is doing here… Then I realized that this is responsible for actually deleting a tag when one fully clears the value of a tag, e.g. in the raw tag editor (where `entity.tags[k] === ''`). :sweat_smile:

tyrasd (approved):

PR #11334, opened 2025-08-21
enhance "incompatible source" warnings for changesets

### LINKED ISSUE #8400: changeset tag warnings:incompatible_source
when a use put a comment or a source on the objet with google maps, a changet tag warnings:incompatible_source is added, very easy to catch via osmcha for ex.
https://github.com/openstreetmap/iD/pull/3064
https://www.openstreetmap.org/changeset/100360273

but it fail to detect the issue in the changeset tag source
https://www.openstreetmap.org/changeset/100734135

could this be improved ?

tyrasd:
* not only warn on keyword "google", but:
* use same list as is used by the "incompatible source" validator for map features
* also warn when an incompatible source is mentioned in the "sources" field / changeset tag
* include number of invalid sources in uploaded changeset comment or source in `warnings:incompatible_source` changeset tag, fixes #8400
* show multiple warnings if multiple incompatible sources are present in a tag

[example changeset](https://api06.dev.openstreetmap.org/changeset/435213)

<hr>

as this changes and introduces strings to translate, this should wait for the v2.37, so open todo:

* [x] add to changelog for v2.37-dev


modules/ui/commit.js
```diff
@@ -185,11 +184,24 @@ export function uiCommit(context) {
...
             .getIssuesBySeverity({ what: 'edited', where: 'all', includeIgnored: true, includeDisabledRules: true })
             .warning
             .filter(function(issue) { return issue.type !== 'help_request'; });    // exclude 'fixme' and similar - #8603
 
+        // also include "incompatible_source" warnings caused by the changeset tags
+        function checkIncompatibleSources(str) {
+            return incompatibleRules
+                .filter(rule =>
+                    rule.regex.test(str) &&
+                    !rule.exceptRegex?.test(str)
```
hlfan: This section is also more or less a duplicate of `incompatibleRules.find` in `checkIncompatibleSource`.
Exporting `getIncompatibleSource(str): {id, match}` instead would be more DRY.

PR #11341, opened 2025-08-22
add missing changelog entries

k-yle:
&shy;


tyrasd (approved):
Thanks for catching those! :heart:

PR #11310, opened 2025-08-14
Port Maproulette integration from Rapid to iD

tordans:
Closes https://github.com/openstreetmap/iD/issues/10758

This is a take at porting the Maproulette integration from Rapid to iD.
It tried the LLM route for this using Cursor with ChatGPT-5 and a lot of nudging and prompting.

**[You can test this code in this manual preview](https://id-mapillary-integration.netlify.app/#map=19.21/52.47707/13.42995&background=Bing&disable_features=points,boundaries&locale=en&maproulette=true)**
Here is [the same location in Rapid for comparison](https://rapideditor.org/edit#map=19.21/52.47708/13.43021&maproulette=true&background=Bing&disable_features=points,boundaries&locale=en).

(Note to self: [Netlify project](https://app.netlify.com/projects/id-mapillary-integration/overview))

**Open todos**

- [ ] Improve the icon; it's way to hard to see right now and does not look nice.
- [ ] Test if the features Rapid has actually work
    - [ ] Going through the features PR by PR in Rapid should be a good start https://github.com/facebook/Rapid/pulls?q=is%3Apr+is%3Aclosed+maproulette
- [ ] Review the code – I did only a very high level code review myself and only looked at the result in the browser up until now
- [ ] Rework data flow to use https://github.com/maproulette/maproulette-backend/issues/1037

**Features that I tested**
Right sidebar
- activate MR via Checkbox
- preserve active state in URL
- activates on fresh (re)load of editor via URL
- filter by MR campaign ID (same as rapid, input need to loose focus to submit)
- MR campaign ID is preserved as state in URL
- activates on fresh (re)load of editor with selected campaign via URL

API
- same number of requests as rapid in the network tab

Left sidebar
- left side panel shows the same content, including Mustache tags
- "Open in MR" opens MR task

Map
- MR tasks show up and update when changing the settings in the right sidebar

**Known issues**

- [ ] Translations don't show on the preview even with `&locale=en` but they do show up locally

**Open to test**

- [ ] Mustache form replacement
- [ ] MR not logged in error / flow
- [ ] Actually submitting the form
- [ ] Submitting the form + "next" checkbox

**Missing feature**

- [ ] Rapid has this nice feature which highlights the OSM element that is referenced in the MR task (when hovering the "Kontext way/123" in the instructions) in some situations. Copilot suggested that this will not work in iD because it does not have the necessary event handlers to do this. I did not research this more, yet…
   - image

PR #11346, opened 2025-08-24
Feat: Click to switch features that are crossing ways

### LINKED ISSUE #11328: Intersection warning: consider adding button to select the other feature
### Situation

The iD editor currently detects intersections before submitting. However, sometimes it is not possible to easily select the other feature to fix tagging, which would then resolve the intersection warning.

### Suggestion

Proposing to add a button to the box to quickly select the other feature during the intersection resolution.

The final result could be like:

```
[ ] Tag this as higher
[ ] Tag this as lower
[ ] Select the other feature
[ ] Reposition the features
[ ] Ignore this issue
```

### Screenshots

Image

Captured at version 2.35.3

Waqibsk:
Fixes #11328

This PR , turns the labels of the crossing  features  into a link which when clicked on would select that  feature.
This makes it easier for users to navigate between crossing features.

Screenshot:
Screenshot 2025-08-24 232411

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -408,7 +408,6 @@ export function validationCrossingWays(context) {
             crossingTypeID += '_connectable';
         }
 
-        // Differentiate based on the loc rounded to 4 digits, since two ways can cross multiple times.
         var uniqueID = crossing.crossPoint[0].toFixed(4) + ',' + crossing.crossPoint[1].toFixed(4);
 
         return new validationIssue({
@@ -419,10 +418,52 @@ export function validationCrossingWays(context) {
                 var graph = context.graph();
                 var entity1 = graph.hasEntity(this.entityIds[0]),
                     entity2 = graph.hasEntity(this.entityIds[1]);
-                return (entity1 && entity2) ? t.append('issues.crossing_ways.message', {
-                    feature: utilDisplayLabel(entity1, graph, featureType1 === 'building'),
-                    feature2: utilDisplayLabel(entity2, graph, featureType2 === 'building')
-                }) : '';
+               return entity1 && entity2
+                   ? function (selection) {
+                         selection.html(
+                             t('issues.crossing_ways.message', {
+                                 feature:
+                                     '<span class="feature-link feature1"></span>',
+                                 feature2:
+                                     '<span class="feature-link feature2"></span>',
+                             })
+                         );
+
+                         selection
+                             .select('.feature1')
+                             .text(
+                                 utilDisplayLabel(
+                                     entity1,
+                                     graph,
+                                     featureType1 === 'building'
+                                 )
+                             )
+                             .style('cursor', 'pointer')
+                             .style('text-decoration', 'underline')
+                             .on('click', () =>
+                                 context.enter(
+                                     modeSelect(context, [entity1.id])
+                                 )
+                             );
+
+                         selection
+                             .select('.feature2')
+                             .text(
+                                 utilDisplayLabel(
+                                     entity2,
+                                     graph,
+                                     featureType2 === 'building'
+                                 )
+                             )
+                             .style('cursor', 'pointer')
+                             .style('text-decoration', 'underline')
+                             .on('click', () =>
+                                 context.enter(
+                                     modeSelect(context, [entity2.id])
+                                 )
+                             );
+                     }
+                   : '';
             },
             reference: showReference,
             entityIds: entities.map(function(entity) {
@@ -511,6 +552,7 @@ export function validationCrossingWays(context) {
                 .append('div')
                 .attr('class', 'issue-reference')
                 .call(t.append('issues.crossing_ways.' + crossingTypeID + '.reference'));
+
         }
     }
 
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -409,7 +409,6 @@ export function validationCrossingWays(context) {
         }
 
         // Differentiate based on the loc rounded to 4 digits, since two ways can cross multiple times.
-
         var uniqueID = crossing.crossPoint[0].toFixed(4) + ',' + crossing.crossPoint[1].toFixed(4);
 
         return new validationIssue({
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -408,6 +408,8 @@ export function validationCrossingWays(context) {
             crossingTypeID += '_connectable';
         }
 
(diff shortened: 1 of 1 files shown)


tordans:
Functionally I think this is a good idea. Rapid has a „highlight on hover“ for their MapRoulette integration which I find even more useful because it does not change the current selection. I looked into this for #11310 but iD might be missing something to use this easily. 

Visually I think we should have a look if there is any existing „Text as action“ style that we can reuse like the blue color or something else. 

Semantically the span should be a button tag IMO.

tyrasd:
> highlight on hover

do you mean similar to the blue outline that is currently already shown for the secondary affected feature of the validation?

image

(The code responsible for this is `utilHighlightEntities` in `util/util.js`.)

Or did you mean something different? :thinking: 

> because it does not change the current selection

Note that just identifying the features mentioned in the validation message is not mainly what this PR tries to solve: The main use case is to be able to quickly go to the other feature in case fixing the issue is more easily be done by modifying the other feature.

tyrasd:
> Semantically the span should be a button tag IMO.

Something like this?



Hmm, I don't think that looks great. I'd rather go with a simple link (`<a>`), to visually mark the text as individually clickable (i.e. blue text, correct cursor; avoiding the use of inline css). This is a common UX pattern used in iD and there are even already a couple of places in iD where `<a>` links lead one directly to select a specific map feature (e.g. in Osmose QA issues).

tordans:
> > highlight on hover
> 
> do you mean similar to the blue outline that is currently already shown for the secondary affected feature of the validation?
> …
> (The code responsible for this is `utilHighlightEntities` in `util/util.js`.)

Yeah, that's it, thanks. Now that I think about it again, the different to Rapid was something else … but not relevant here.

> > because it does not change the current selection
> 
> Note that just identifying the features mentioned in the validation message is not mainly what this PR tries to solve: The main use case is to be able to quickly go to the other feature in case fixing the issue is more easily be done by modifying the other feature.

👍 

> > Semantically the span should be a button tag IMO.
> 
> Something like this?

No, I was only referring to the semantics, not the visual representation, mainly from an accessibility and general clean-html point of view where actions elements should be either `a` tags or `button` tags (with `a` tags reserved for something that has a `href` and `button` tags for something that has a `onclick` or form submit), but not `span` or `div`. Visually both can be styles as the other, which is based on a completely different set of guidelines.

So +1 for the link style look of the second screenshot.

tyrasd (changes requested):
Hi @Waqibsk . This is a good start. The following would be to be adjusted still:

* feature1 is already the selected feature, thus it does not need to be clickable
* as mentioned above: it would be better if the clickable part of feature2 is a `<a>` element, instead of a custom styled `<span>`
* currently, this only adds the link for the _crossing ways_ validation. I believe other some of the other validation modules also have warnings involving more then one feature. Could you please add the same link also to those validators?

I noticed that the implementation is quite a bit more convoluted than expected (and using `selection.html` which should be avoided in general). I see why you chose to approach the implementation that way, as the localizer module did not have a functionality to accept a callback function to specify additional html markup for the replacement strings. I did now extend the functionality to add this missing functionality in #11347. Please take a look and rebase your PR on that branch, as it should allow you to greatly reduce the overhead in this PR. I think you should end up with something like the following:

```js
return entity1 && entity2
    ? t.append('issues.crossing_ways.message', {
        feature: utilDisplayLabel(entity1, graph, featureType1 === 'building'),
        feature2: selection => selection
            .append('a')
            .classed('feature-link', true)
            .text(utilDisplayLabel(entity2, graph, featureType2 === 'building'))
            .click(…)
    }) : '';
```

Waqibsk:
>feature1 is already the selected feature, thus it does not need to be clickable

I think currently, the way iD handles `crossing ways` validation, the order of the features doesn’t change after switching. For example, if a `railway track (feature1)` is crossing a `river (feature2)`, and the `railway`is selected first, then after clicking the `river` link, the `railway `still remains in the “first” position . Because of this, the user cannot switch back to the `railway` unless both features are made clickable.
That said, if we prefer not to make the currently selected feature clickable, a conditional could be added to handle this

tyrasd:
>  the order of the features doesn’t change after switching.

True. But still: only the not already selected feature should be clickable, regardless of if it's the first or second feature. Should none of the selected features exactly match the current selection (e.g. when multiple features are selected), then both links can be clickable.

Waqibsk:
Screenshots:
<div>
Screenshot 2025-08-28 233854
Screenshot 2025-08-28 233845
</div>

modules/validations/crossing_ways.js
```diff
@@ -531,6 +587,7 @@ export function validationCrossingWays(context) {
                 .append('div')
                 .attr('class', 'issue-reference')
                 .call(t.append('issues.crossing_ways.' + crossingTypeID + '.reference'));
+
```
matkoniecz: do we need that new empty line?

modules/validations/crossing_ways.js
```diff
@@ -446,10 +447,65 @@ export function validationCrossingWays(context) {
                 var graph = context.graph();
                 var entity1 = graph.hasEntity(this.entityIds[0]),
                     entity2 = graph.hasEntity(this.entityIds[1]);
-                return (entity1 && entity2) ? t.append('issues.crossing_ways.message', {
-                    feature: utilDisplayLabel(entity1, graph, featureType1 === 'building'),
-                    feature2: utilDisplayLabel(entity2, graph, featureType2 === 'building')
-                }) : '';
+     return entity1 && entity2
```
matkoniecz: it seems that something weird went with indentation here

matkoniecz (commented):
no proper code review yet (but I spotted some very minor things - given how long PR waited I can also fix them myself), but I tested and it works beautifully - thanks!

k-yle:
this is blocked by #11347, and will conflict with #9888

so i think #11347 needs to be merged first[^1]

[^1]:  [#9888](https://github.com/openstreetmap/iD/pull/9888) could also be merged first, or last. doesn't really matter

matkoniecz:
> this is blocked by https://github.com/openstreetmap/iD/pull/11347

interesting, seemed to work entirely fine in my testing

k-yle:
yeah it works fine, but iirc, the only purpose of #11347 was to make the implementation in this PR cleaner

matkoniecz:
> yeah it works fine, but iirc, the only purpose of https://github.com/openstreetmap/iD/pull/11347 was to make the implementation in this PR cleaner

and it can be now done here!

matkoniecz:
@Waqibsk https://github.com/openstreetmap/iD/pull/11347 was merged, can you update this PR to use code improvements enabled by it?

PR #11347, opened 2025-08-25
Allow string replacements to use callback functions

tyrasd:
This allows:

1. to implement replacement parts that can inject additional html markup[^1]
  ```js
  selection.call(t.append('stringId', {
      key1: 'replacementString',
      key2: sel => sel.append('a')
          .attr('href', 'https://example.com')
          .text(…);
  })
  ```
2. to have the correct locales on a replacement part if the respective substring is not yet translated in the locale. In the following example code, if `operations._unavailable` is translated, but `operations.….title` isn't, this will now produce the correct `lang` attribute for each individual part of the string: `<span lang="lang1">…translation for operations._unavailable…</span><span lang="en">…fallback translation for operations.….title…</span>`:
  ```js
  t.append('operations._unavailable', {
      operation: t.append(`operations.${_operation.id}.title`))
  });
  ```
  

[^1]: this was previously only possible via a hacky solution using d3's `.html` method (e.g. `t.html(stringId, { key: { html: '<a href="">…</a>' } })`) which we want to avoid, as it makes html injection attacks potentially possible, if the input strings are not very carefully escaped.

Open todos:
* [x] (after PR is merged) update [wiki](https://github.com/openstreetmap/iD/wiki/How-to-Use-iD%27s-Localizer-Module)
* [x] update existing code to make use of this functionality
  * [x] replace most remaining uses of 't.html'
  * [x] make use of nested `t.append`
* [x] add to changelog


modules/ui/sections/map_style_options.js
```diff
@@ -65,7 +67,9 @@ export function uiSectionMapStyleOptions(context) {
 
         label
             .append('span')
-            .html(function(d) { return t.html(name + '.' + d + '.description'); });
+            .ech(function(d) {
```
hlfan, line 70: This is causing errors in osm prod @tyrasd:
```suggestion
            .each(function(d) {
```

1ec5: Fixed in 0241c7653d4a93692f59b7aa1021242e63a3de72 and revert posted in openstreetmap/openstreetmap-website#6917.

Issue #9606, opened 2023-04-30
Custom Map Data Settings: Supports KML, how about KMZ?


jidanni:
Here we see KML is supported,
![20230430T165302](https://user-images.githubusercontent.com/1230959/235344545-f5c9fed0-15f4-4ee8-80f5-9fead352b669.jpg)
Well if KMZ were also supported, users wouldn't need to futz with converting it themselves.

tyrasd:
Unfortunately, KMZ is not supported by the library we use to convert geodata to geojson. See https://github.com/mapbox/togeojson/issues/119

tordans:
> See [mapbox/togeojson#119](https://github.com/mapbox/togeojson/issues/119)

Side note: it might be worth switching to https://togeojson.docs.placemark.io/#how-does-this-differ-from-mapboxtogeojson - that does not support this format either but is better maintained.

tyrasd:
Sorry, misclicked the reopen button.

We do use tmcw's version of the library (see https://github.com/openstreetmap/iD/blob/v2.35.3/package.json#L52), but for the same reason as in the linked issue also that one does not support KMZ. 😊

jidanni:
To convert a KMZ file to GeoJSON format, I simply use
```
ogr2ogr -f GeoJSON output.geojson input.kmz #just like:
ogr2ogr -f GeoJSON output.geojson input.kml
```

Network links aren't any different in KML vs. KMZ.
But I am not asking you to start supporting network links.
I'm just asking you to support the 100 times smaller KMZ format.

tyrasd:
The issue is that KMZ is using the zip file format under the hood, which is quite complex, as the data could be compressed in a variety of different compression algorithms.[^1] Note that as KMZ files are essentially zip files under the hood, you don't even need ogr2ogr to convert it first: Just change the file extension to `.zip` and unpack it using your Operating System or file compression tool of choice. The contents of the zip file are the KML file you can feed into iD.

[^1]: The most common one, _deflate_, would be somewhat reasonably possible to implement using a [DecompressionStream](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream/DecompressionStream), but the alternative compression methods would be unfeasible. That would result in some KMZ files working, and some KMZ files not working. There I think it is better to ask the user convert the files to a supported format first.

PR #11353, opened 2025-08-27
compress changesets with gzip before upload

### LINKED ISSUE #6231: Compressed changeset upload
The changeset upload endpoint can handle gzip compressed payloads when providing an additional `'Content-Encoding': 'gzip'` http header in the following request:

```
            // Upload the changeset..
            var options = {
                method: 'POST',
                path: '/api/0.6/changeset/' + changesetID + '/upload',
                options: { header: { 'Content-Type': 'text/xml' } },
                content: JXON.stringify(changeset.osmChangeJXON(changes))
            };
```

Currently, both on production and dev instance support this option. See https://github.com/openstreetmap/operations/issues/193 for a bit of motivation why this could be useful.

Not at all time critical -> v3 topic

k-yle:
Closes #6231

I didn't realise [`CompressionStream`](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream) is now supported in all major browsers, until I saw Martin's [comment](https://github.com/openstreetmap/iD/issues/9606#issuecomment-3224083850) yesterday. 

This makes it possible to implement #6231 without adding [pako](http://npm.im/pako) as a dependency. For old browsers, we fallback to a normal upload. 

One downside is that the Chrome DevTools only show the compressed payload, making it slightly less convenient to debug:

image

tested with the dev server ([cs437286](https://master.apis.dev.openstreetmap.org/changeset/437286) + a few others)


tyrasd:
> One downside is that the Chrome DevTools only show the compressed payload, making it slightly less convenient to debug:

Same for firefox's dev tools. :see_no_evil: But that seems to be more like a bug/deficiency in the respective dev tool, as they should be able to see the `Content-Encoding` header and decompress the payload accordingly. I found https://bugzilla.mozilla.org/show_bug.cgi?id=1832779 on Firefox' bug tracker for this.

tyrasd (approved):

Issue #11215, opened 2025-07-16
Erroneous NSI warning


Limfjorden:
### URL

https://www.openstreetmap.org/edit?way=265236425#map=19/54.832294/11.163538

### How to reproduce the issue?

1. Create an object with the following tags (or go to the URL provided above):

building=retail
name=OK Plus
shop=convenience
brand:wikidata=Q130240297
brand=OK Plus

2. Observe iD wanting to change most of the tags including wikidata to follow the NSI for Circle K.

### Screenshot(s) or anything else?

<img width="387" height="570" alt="Image" src="https://github.com/user-attachments/assets/5dda38fc-ec9b-4b72-a3e7-a5e3154271d2" />

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app

### What version numbers does this issue effect?

2.35.0

### Which browsers are you seeing this problem on?

Firefox

tyrasd:
One solution for this particular case would be to add "OK Plus" to the name-suggestion-index (https://github.com/osmlab/name-suggestion-index).

But I would agree that the algorithm is a bit too aggressive in matching the NSI data to existing OSM data in this case, as here the brand has a ~~completely different name~~[^1], and is even fully qualified with its own wikidata entry.

[^1]: this is actually not the case: the NSI preset does have `OK` as a [`matchName`](https://github.com/osmlab/name-suggestion-index/blob/98ed1067e38238fddedfa39694985d2f5eca3c3c/data/brands/shop/convenience.json#L1326-L1332)

bhousel:
WE can fix this on the NSI side.
https://github.com/osmlab/name-suggestion-index/issues/11345

PR #11372, opened 2025-09-01
Fixes unexpected NSI matches

mudit06mah:
### Issue Fixed:
#11215 

### Issue:
**Before:**
NSI was suggesting tag upgradation for "OK Plus" with "Circle K"
image

**After:**
No such error is suggested.

### PR Explanation:
When gathering all the names, for example in "OK Plus"
A tuple was generated for "OK Plus" as well as "OK".
When comparing "ok" with "circlek", we obtain a match (I believe it happens because "ok" is a very short string)
I fixed it by directly comparing the complete search string so that it doesn't accidentally match with strings it isn't completely similar to.

### Feedback:
I could not understand why we're using `gatherNames()` and `gatherTuples()`.

### DIFF
--- a/modules/services/nsi.js
+++ b/modules/services/nsi.js
@@ -540,7 +540,23 @@ function _upgradeTags(tags, loc) {
     if (hits[0].match !== 'primary' && hits[0].match !== 'alternate') break;  // a generic match, stop looking
 
     const searchName = tuple.n.toLowerCase().replace(/[^a-z0-9]/g, '');
-    if (searchName.length < 3) continue;  // skip too-short names
+    for (let i = hits.length - 1; i >= 0; i--) {
+      const matchItem = _nsi.ids.get(hits[i].itemID);
+      if (matchItem) {
+        const matchName = (matchItem.tags.name || '').toLowerCase().replace(/[^a-z0-9]/g, '');
+        if (searchName && matchName) {
+          if (searchName.length < 5) {
+            //directly compare short names:
+            if (searchName !== matchName) {
+              hits.splice(i, 1);
+            }
+          }
+        }
+      }
+    }
+    
+    if (!hits.length) continue;  // if all matches were filtered out, try next tuple
+    
 
     // A match may contain multiple results, the first one is likely the best one for this location
     // e.g. `['pfk-a54c14', 'kfc-1ff19c', 'kfc-658eea']`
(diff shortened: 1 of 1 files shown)


bhousel:
> Still here's a fix for this Issue.

How is this PR supposed to fix the issue?

mudit06mah:
> > Still here's a fix for this Issue.
> 
> How is this PR supposed to fix the issue?

The tag upgradation was being suggested because "OK" was a really small string and was somehow matching with "circlek", so I skipped all strings with size less than 3 
, thus removing unwanted suggestions.

If you think a better method is to match the string before skipping it, then I will add that

mudit06mah:
I just realised why this is not a good fix.
I apologise for the inconvenience and I will fix it first thing tomorrow

mudit06mah:
I have fixed my PR by comparing small strings instead of checking, (so for example, name like "kfc" will not be skipped).
Hence this issue has been fixed without causing any more problems.

bhousel:
Why not just add the "OK Plus" brand to NSI?

These are the "OK" branded convenience stores we are tracking already:
https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=ok
That page lists the different name tags that will match, and which locations around the world that it can be expected to be found.

mudit06mah:
> Why not just add the "OK Plus" brand to NSI?
> 
> These are the "OK" branded convenience stores we are tracking already: https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=ok That page lists the different name tags that will match, and which locations around the world that it can be expected to be found.

I think the issue here is that the current algorithim is forcefully matching brands like "OK Plus" and "Circle K" which are completely different names, and hence I fixed that.
I didn't know that the brand "OK" is already being tracked. But even if it is, I fixed the code to not match different names like "OK Plus" and "Circle K".

I will add "OK Plus" to the NSI list if it is still needed.

UKChris-osm:
It seems [Circle K](https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=ok#circlek-04f01c) includes a `matchName` for **ok**, but I'm not sure why - perhaps we could also remove this `matchName` to stop the suggestion?

bhousel:
> It seems [Circle K](https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=ok#circlek-04f01c) includes a `matchName` for **ok**, but I'm not sure why - perhaps we could also remove this `matchName` to stop the suggestion?

Yeah I agree, it seems like we should 
1. Remove the "ok" and "ok-mart" matchNames for the worldwide Circle K.  
2. Review existing items for "OK" branded stores in the parts of the world where they have it (China, Hong Kong, Macau, also  a Chilean and a South African brands that seems unrelated).
3. Add a new item for "OK Plus" scoped to wherever that is used.

I opened https://github.com/osmlab/name-suggestion-index/issues/11345

PR #11379, opened 2025-09-03
fix `railway:signal:position` rendering errors

hlfan:
These weren't caught by the review of #10303


tyrasd (approved):
Good catch, thanks!

PR #11385, opened 2025-09-04
Fix payment wikidata

### LINKED ISSUE #11164: Incorrect data item by payment: key prefix in iD
`payment:` key prefix in ID has discouraged data item.

Image

Image

Correct data item:

Image

srujanaelicherla:
This PR introduces a temporary fix for the payment field in the iD editor. When a key starts with payment:, it now correctly uses the Wikidata ID Q550 to ensure proper linking.

Closes #11164

Changes Included:

Updated getEntity function in osm_wikibase.js to force payment: keys to use Q550.
No other functional changes to the iD editor.

Notes for Reviewers:

Verify that payment: keys return the correct Wikidata entity.
Confirm no unintended side effects on other key lookups.

### DIFF
--- a/modules/osm/node.js
+++ b/modules/osm/node.js
@@ -106,7 +106,7 @@ const prototype = {
             if (SIDES.has(sideTag?.toLowerCase()) && !hideSideTag) {
                 rawValues.push({
                     type: 'side',
-                    value: this.tags.side.toLowerCase(),
+                    value: sideTag?.toLowerCase(),
                 });
             }
 


README.md
```diff
@@ -52,3 +52,6 @@ iD also bundles portions of the following open source software.
 ## Thank you
 
 Initial development of iD was made possible by a [grant of the Knight Foundation](https://blog.mapbox.com/large-investment-in-openstreetmap-from-knight-foundation-cf7aa00534db).
+
+Practice line
+Srujana
```
k-yle: I don't think a 'temporary solution' is a good idea, since it's quite easy to fix this properly.

The easiest solution would be in the [id-tagging-schema repository](https://github.com/openstreetmap/id-tagging-schema). We could add `"reference": { "key": "payment:*" }` to this field (see [documentation here](https://github.com/ideditor/schema-builder#reference))

---

Alternatively, we could fix the code itself, which [already](https://github.com/openstreetmap/iD/blob/135397dc9762fb185c78ef3c6f1bfdd9bdf23934/modules/ui/field.js#L215-L217) special cases [`multiCombo` fields](https://github.com/ideditor/schema-builder#:~:text=multiCombo) fields (but not correctly). 

Currently we call the API with `?titles=payment`, but we want to use `?titles=payment:*`  instead. This could be fixed by changing [this line](https://github.com/openstreetmap/iD/blob/135397dc9762fb185c78ef3c6f1bfdd9bdf23934/modules/ui/field.js#L216):

```diff
  if (d.type === 'multiCombo') {
-   referenceKey = referenceKey.replace(/:$/, '');
+   referenceKey = referenceKey.replace(/:$/, ':*');
  }
```

However, there are [4 other places](https://grep.app/search?f.repo=openstreetmap%2FiD&q=uiTagReference%28) that call `uiTagReference(`, so we need to check those too. And we need to check how this affects the [17 other multiCombo fields](https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema+%22type%22%3A+%22multiCombo%22&type=code).

k-yle (changes requested):

README.md
srujanaelicherla: Thanks @k-yle for the detailed guidance 
I understand that a proper fix is better than a temporary patch.
I’ll explore the two approaches you mentioned — starting with adding "reference": { "key": "payment:*" } in id-tagging-schema, since that seems the cleanest.
If that works, I can update this PR accordingly.

srujanaelicherla:
Thanks @k-yle for the clear guidance earlier. 
I’ve now opened a PR in id-tagging-schema to add the proper reference mapping for payment:* → [openstreetmap/id-tagging-schema#1701](https://github.com/openstreetmap/id-tagging-schema/pull/1701?utm_source=chatgpt.com)

This should resolve the issue in a clean way, without the temporary patch in osm_wikibase.js.
Once that schema change is merged and released, iD should automatically pick it up.

Happy to adjust or close this PR depending on how the maintainers want to proceed.

Issue #5228, opened 2018-08-16
Conflict resolution popup, potential endless loop in conflict resolution


mmd-osm:
Steps to reproduce:

1. JOSM: Create way with 2 nodes, tags= highway=path
2. JOSM: Upload change
3. iD: Load area, including way
4. JOSM: Add a third node in the middle of the way
5. JOSM: Upload change
6. iD: add a second tag to the way: surface=grass
7. iD: Upload changes
8. iD: Conflict detection kicks in, resolves conflict, and uploads again
9. iD gets stuck on the following screen:

![bildschirmfoto von 2018-08-16 09-36-26](https://user-images.githubusercontent.com/5842757/44194962-f164fa00-a137-11e8-9323-f41807e741fe.png)

There's also an exception on the javascript console:

![bildschirmfoto von 2018-08-16 09-43-40](https://user-images.githubusercontent.com/5842757/44195289-e3fc3f80-a138-11e8-891b-feae68c055e1.png)

Tested on http://preview.ideditor.com/master and dev instance. The change in question has been applied to the database: https://master.apis.dev.openstreetmap.org/way/4305175922 (note the 3 nodes, and the 2 tags).

Someone reported on the forum, that this counting up may also happen in an endless loop (!) going on for hours and hours. This is probably the case when the conflict resolution didn't work out at all, and the algorithm keeps trying and trying.

bhousel:
Thanks for reporting @mmd-osm  - the exception on the console explains exactly what the issue is 👍

mmd-osm:
Awesome. Any idea where this endless loop thing might come from? It would just show the popup, count from 0 to NNN, then start over again from zero and so on.

bhousel:
> Awesome. Any idea where this endless loop thing might come from? It would just show the popup, count from 0 to NNN, then start over again from zero and so on.

Hmm yeah in this case it wasn't looping, iD resolved all the conflicts and sent the merged changeset just fine.  This bug happened because the uploading code failed to call the final callback and let the saving code know that it was done.

There's also #5114 - I think that's what the person on the forum might have encountered.

Issue #8610, opened 2021-07-31
I get info that some things are not translated in Polish. I do not have Polish language translation selected.


matkoniecz:
https://www.openstreetmap.org/preferences is set to `en`

I get "missing pl translation..." in many places

![screen](https://user-images.githubusercontent.com/899988/127741369-0358111a-826f-4d31-9466-fb7584d42045.png)

Firefox 90.0 on  Lubuntu 20.04

In browser settings English is ranked over Polish

![screen01](https://user-images.githubusercontent.com/899988/127741396-442871e1-7f7b-4a50-802b-1c6e01e3f5ff.png)

mbrzakovic:
Hi @matkoniecz, I can't repro the issue. 
Does the issue persist after hard-refresh/refresh of the page?
Can you please provide endpoint url that you are targeting?
Additionally, if you could go to Inspect, and paste down the console log, that would be helpful.

matkoniecz:
It went away on its own (for now at least).

I am going to blame cosmic rays and reopen if I manage to reproduce it.

Thanks for checking, no idea for now what triggered it.

weech:
I have had this or a very similar error occur occasionally with RapiD for a while now, but I didn't use that enough for it to matter to me. With this new iD release, it occurs 10-20% of the time. I am also using Firefox on Linux, and accessing iD from the main OSM website. At first I though it might be an extension, but I disabled any extension that could be blocking the script, and it still occurs. I think it's a CORS issue. Here's the top few unique lines of the console from the most recent time it happened.

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://cdn.jsdelivr.net/npm/@openstreetmap/id-tagging-schema@3/dist/translations/en.min.json. (Reason: CORS request did not succeed).

TypeError: NetworkError when attempting to fetch resource. id-a7caaa0f3f6288026cbae41d1ab0115f47bca016cdcd62a2d26e56bb35e42903.js:6:22283
    h https://www.openstreetmap.org/assets/id-a7caaa0f3f6288026cbae41d1ab0115f47bca016cdcd62a2d26e56bb35e42903.js:6
```
I also have no idea how to reproduce it, but if I figure out how, I'll let you know. The issue normally doesn't persist after a refresh, but it did at least once.

mbrzakovic:
@weech thanks for the info. App not being able to get translation file could result in the screen shot that @matkoniecz attached. I will keep my eyes open on this.

[...]

PR #11388, opened 2025-09-05
fix crash when uploading changesets with more than 10,000 features

k-yle:
Now that iD stores the history in [`IndexedDB`](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), it's possible to create enourmous changesets, which might reach the limit of 10,000 features.  Until now, this was almost impossible.

For example, you can easily reach the limit by copy-pasting a backyard water tank 400 times[^1]. If you try to upload >10,000 items, all your edits are lost when the API errors.

There are two solutions:
1. prevent uploading changesets with over 10,000 features
2. refactor the uploader so that it splits changesets into chunks of 10,000. This will be a huge change because it needs to consider merge conflicts, while also chunking [in a valid order](https://github.com/osmlab/osm-api-js/blob/2f3758927c4cc7441dd324b8728f90f6be2305df/src/api/changesets/chunkOsmChange.ts#L11-L24). Since this logic already has [longstanding bugs](https://github.com/openstreetmap/iD/issues/5228) and has no unit tests, I don't think it's worth all the effort for a rare situation.

So this PR implements option No.1:

image

test plan:
* in the devtools' network tab, right-click `/api/capabilities`, click <kbd>Override content</kbd>, and change `10000` to something smaller like `10`.

[^1]: 10000 ÷ [25 nodes per water tank] = 400


data/core.yaml
```diff
@@ -1981,6 +1981,8 @@ en:
       max_way_nodes:
         message: Way has too many vertices
         reference: "The maximum number of vertices of a way is {maxWayNodes}. Split the feature into smaller parts or otherwise reduce the number of nodes of the affected way."
+      max_changeset_elements:
+        reference: "You've edited {changesetElements} elements in this changeset, but the maxmium number allowed is {maxChangesetElements}. Therefore, your edit is too large to upload. You need to export your changes as an osmChange file, and use a different editor to upload it."
```
hlfan:
```suggestion
        reference: "You've edited {changesetElements} elements in this changeset, but the maximum number allowed is {maxChangesetElements}. Therefore, your edit is too large to upload. You need to export your changes as an osmChange file and use a different editor to upload it."
```

k-yle: oops, thanks

tyrasd:
My first thought here was: How does this improve the mapping session of a particular user? The only benefit of catching this in the upload dialog over blindly uploading the changeset anyway is that in this case the error message does not come from the API, but from iD itself (and is therefore slightly more verbose and potentially translated to the user's language).

The user is however equally lost in how to proceed if they don't know about JOSM, or where to ask for help.

Would it not be better if the editor would disallow further edits after the limit was reached (and automatically undo the offending action that brought it over the limit)?

PstasDev:
Then allow it to be enabled for superusers. People might like the modern and great UX, that is not present in JOSM.

k-yle:
> The only benefit of catching this in the upload dialog over blindly uploading the changeset anyway is that in this case the error message does not come from the API, but from iD itself 

actually at the moment iD crashes, so this PR at least prevents your work from getting lost. 


> Would it not be better if the editor would disallow further edits after the limit was reached (and automatically undo the offending action that brought it over the limit)?

if we want this level of complexity, then it might be better to just refactor the changeset uploading process to fix conflict resolution and support chunked uploads. then there's no error message required. 


I was hoping this could be a quick solution to prevent hours of work from being lost, until we have time for 1 of these long-term solutions

matkoniecz:
Maybe I am testing it the wrong way(TM) but I got endless loop of 409 errors when I tried to upload overly large CS in unmodified iD - not a crash.

screen-2026-02-27-13-02-22

matkoniecz:
And these are `The changeset 541617 was closed at 2026-02-27 12:01:00 UTC` which just makes me confused what is going on here.

matkoniecz:
With this PR upload is outright blocked. Interface is a bit confusing (grayed out upload, message appears only on hover) but as stop-gap against user losing their data I guess it is an improvement?

Though maybe that message can replace upload button message? Or it be changed? Or would it be too much work for translators?

BTW, I am hit again by #8610

screen-2026-02-27-13-09-56

PR #11389, opened 2025-09-05
add type-annotations to the geo/* files

k-yle:
just to make developing a bit more pleasant. I've used the typescript syntax since it's more concise, but JSDoc comments would have the same effect. JSDoc just involes 3-9 extra lines of code per function.

excluding `geoExtent` for now, since it's less straightforward


modules/geo/vector.ts
```diff
@@ -1,5 +1,8 @@
+export type Vec2 = [x: number, y: number];
+export type Vec3 = [x: number, y: number, z: number];
```
tyrasd: `Vec3` is not used anywhere as far as I can see, is it? I also can't think of a situation where we'd need it[^1] in the near future, so I'd probably omit it for now.

[^1]: except for the zoom/lat/lon `map` parameter, but that's not really a vector, and the order would be different again.

k-yle: sure, deleted

PR #11392, opened 2025-09-08
fix 'Toggle history panel' keyboard shortcut

### LINKED ISSUE #10479: 'Toggle history panel' keyboard shortcut overrides browser shortcuts on macOS
### URL

https://www.openstreetmap.org/edit

1. Open the URL
2. Click somewhere on the map
3. Press `⌘⇧H`
4. The iD _History panel_ opens instead of...
    - the home page of the browser (Safari `18.0`, Google Chrome `129.0.6668.71`, Microsoft Edge `129.0.2792.65`)
    - the History sidebar of the browser (Firefox `130.0.1`)

https://github.com/user-attachments/assets/ea1a6073-dbb2-4722-92b2-2902e67b4b5f

The issue is present on macOS `15.0` using the browsers above. I haven't tested Opera and Vivaldi.

I think this is the relevant line in the code: https://github.com/openstreetmap/iD/blob/0ce8136279055311e2355c98b4dfab5a5f96a3a4/modules/ui/sections/data_layers.js#L404

Maybe it could be solved by requiring an additional press on the ⌥ key:
```diff
        var historyPanelLabelEnter = panelsListEnter
            .append('li')
            .attr('class', 'history-panel-toggle-item')
            .append('label')
            .call(uiTooltip()
                .title(() => t.append('map_data.history_panel.tooltip'))
-                .keys([uiCmd('⌘⇧' + t('info_panels.history.key'))])
+                .keys([uiCmd('⌘⌥⇧' + t('info_panels.history.key'))])
                .placement('top')
            );
```

### To do after the issue is closed as completed

- [ ] Update the Wiki at https://wiki.openstreetmap.org/w/index.php?title=ID/Shortcuts&veaction=edit&section=8

2.30.2

Chrome, Firefox, Microsoft Edge, Safari

gy-mate:
Fix #10479.

### DIFF
--- a/modules/ui/sections/data_layers.js
+++ b/modules/ui/sections/data_layers.js
@@ -401,7 +401,7 @@ export function uiSectionDataLayers(context) {
             .append('label')
             .call(uiTooltip()
                 .title(() => t.append('map_data.history_panel.tooltip'))
-                .keys([uiCmd('⌘⇧' + t('info_panels.history.key'))])
+                .keys([uiCmd('⌘⌥' + t('info_panels.history.key'))])
                 .placement('top')
             );
 


tyrasd:
:thinking: 

This makes the history panel break the pattern of <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>key</kbd> that is currently shared between all info panels, which is not ideal. And chaning the shortcut would likely break the habits of many users who memorized the shortcut to toggle the panels.

I'm having a hard time understanding how #10479 even is a real issue in practice. For example, under what circumstance does one need to open the browser's home page while mapping? Also, as an easy workaround it is always possible to exit the focus from of the iD context by pressing <kbd>Ctrl</kbd> <kbd>L</kbd> (to move the cursor into the browser's address bar): then one can press any keyboard shortcut and only the browser's built-in functionality is triggered.

gy-mate:
_See my answer here: https://github.com/openstreetmap/iD/issues/10479#issuecomment-3315276673._

PR #11412, opened 2025-09-14
fix: Warnings for source:esri/Google_Open_Buildings #11410

### LINKED ISSUE #11410: Warnings for source:esri/Google_Open_Buildings
Image

The Google_Open_Buildings dataset is not proprietary and is released under an ODbL compatible license
This dataset is available by default on the rapid editor (which has automatically marked the source)
An exception should be made for this string

Ankitgkp:
#11410

Add exception for Google_Open_Buildings dataset in incompatible source validation

Update regex pattern to include both Google_Africa_Buildings and Google_Open_Buildings
Add test case for Google_Open_Buildings dataset
Google_Open_Buildings is ODbL compatible and should not be flagged as incompatible

Fixes issue where Google_Open_Buildings dataset was incorrectly flagged as
having an incompatible source, despite being released under an ODbL compatible
license and available by default in Rapid editor.


k-yle (approved):
looks good, thanks!

PR #11416, opened 2025-09-15
Fix data loss when changing language in localized name field multiselection

### LINKED ISSUE #11337: data loss when changing language in localized name field of a multiselection
To reproduce:

Ankitgkp:
Closes #11337

Fixes data loss when changing the language of localized name fields in multiselection. Previously, when multiple features had different values for the same localized tag and the user changed the language, only a single value was preserved. Now all values from the multiselection are properly transferred to the new language tag, preventing data loss.

### DIFF
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -31,12 +31,12 @@ export function uiFieldLocalized(field, context) {
     // the language() function will not work immediately.
     fileFetcher.get('languages')
         .then(loadLanguagesArray)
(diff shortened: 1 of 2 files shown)


tyrasd:
Can you please exclude whitespace-only changes to the code? Btw, we typically write anonymous function headers as `function(…) { …` without a space before the parenthesis.

Ankitgkp:
@tyrasd  I’ve raised another PR (#11420) that includes the fix.

PR #11420, opened 2025-09-16
Fix: preserve values when switching language in localized name multiselection

### LINKED ISSUE #11337: data loss when changing language in localized name field of a multiselection

Ankitgkp:
#11337

Resolved data loss issue where changing the language in a localized name field for multiple selected features removed existing tags without transferring values. Now values are correctly reassigned to the new language key.

### DIFF
--- a/modules/ui/fields/localized.js
+++ b/modules/ui/fields/localized.js
@@ -293,10 +293,14 @@ export function uiFieldLocalized(field, context) {
 
         var newKey = lang && context.cleanTagKey(key(lang));
 
-        var value = utilGetSetValue(d3_select(this.parentNode).selectAll('.localized-value'));
+        // Use existing value from the data, preserving multiselection values
+        var existingValue = d.value;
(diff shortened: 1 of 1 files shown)


tyrasd (changes requested):
This does not seem to fix the issue for me:

https://github.com/user-attachments/assets/637a8a93-f9ea-4736-8534-083cc1d3f6bc

I think we need to remember what the tag value was before the key change: `d.value` holds only an array of all values of the multiselection, and does not have the information which feature had which of the multiple names. See line 291 where the previous value is currently cleared: `tags[key(d.lang)] = undefined;` <- that's the value we want to store in the new tag (`tags[newKey] = …`). Also, I think we don't need to use `utilGetSetValue` on the  value input field, because when the language is changed, the correct value is already stored in the `tags` object, so the code can probably be simplified to a single variable holding the `value` to set.

Ankitgkp:
@tyrasd #11427 Please Review this PR

Issue #6789, opened 2019-08-22
Wikidata name locking blocks adding names in other languages


matkoniecz:
https://www.openstreetmap.org/edit?editor=id#map=21/41.86650/12.50386

![Selection_014](https://user-images.githubusercontent.com/899988/63497988-47941800-c4c5-11e9-98b9-97204a2b0028.png)

eliotb:
It is possible to add another language by going to the "All tags" and clicking "+"  
Entering "name:" will give a list of suggestions.
Not as smooth as using the + on the name entry you show.

matkoniecz:
I edited issue title to make it more clear

tohaklim:
I thought this was by design? Or are you referring to the use case where a wikidata-locked entry lacks multilingual names and novice users should be allowed to add them?

ferdinand0101:
This is probably the case
In my opinion the + in the name field should not be locked but the already existing name should stay locked

quincylvania:
@bhousel Do you have thoughts on how this should work? I think protecting existing names but allowing new ones would be okay.

bhousel:
> @bhousel Do you have thoughts on how this should work? I think protecting existing names but allowing new ones would be okay.

Yes I agree - It's probably ok to allow people to change the multilingual names..

lukf:
Why don't we offer an "Upgrade" to automatically add name tags for every language in the wikidata item? The name-locking implies that names are automatically synchronised with the wikidata item.

SilentSpike:
>Why don't we offer an "Upgrade" to automatically add name tags for every language in the wikidata item?

One problem with this is that wikidata labels aren't actually the same thing as the name of what the item represents, they're just identifying labels. The actual names of things should be stored in properties.

Another is that a single user is unlikely to be able to confirm that all languages are named accurately.

matkoniecz:
Note that Wikidata items would require review due to quality issues.

And anyway Wikidata items are often burdened by incompatible licences (for example Wikidata items include data, for example names, added from Google maps - sometimes indirectly, via Wikipedia).

Adamant36:
> Note that Wikidata items would require review due to quality issues.
I can attest to that. Names on Wikidata aren't stable at all as there's some edit warring that goes on with them. So I don't think it would be good to go with them directly.

real68er:
[...]

PR #11423, opened 2025-09-17
"Stop" autofilling localized names from wikipedia article titles

tyrasd:
This functionality was added many years ago in 6636d9ace4a5a859bf7c283148d92d36e6f2ec3f, but was almost rendered ineffective at some later point after which features with a `wikidata` tag don't allow changes to the name tag anymore. One can still get this behavior by manually removing the `wikidata` tag after linking a wikipedia article:

<details>

<summary>example of current behavior</summary>

https://github.com/user-attachments/assets/d873e446-b9ba-482b-887f-3279b76cdf8d

</details>

There are some arguments against having this functionality in iD even if the editing of names were relaxed (see #6789):

* ~~it is questionable whether automatically importing names from wikipedia is compatible with the odbl~~
* it is preferred that names are be mapped from local knowledge – missing names from external sources could always be merged in automatically in a data consumer's post-processing step
* wikipedia page titles do not always correspond to the language of the wiki: sometimes a wikipedia article's title is just the name of the place/feature in a foreign language (e.g. https://nl.wikipedia.org/wiki/Heidelberger_H%C3%BCtte) or includes extra disambiguation text (e.g. https://nl.wikipedia.org/wiki/Heidelberg_(Duitsland) ), etc.
* the implementation is also a little buggy: when the wikipedia tag is changed from one value to another, the fetched translations are not updated, resulting in the old page's names being autofilled


1ec5:
We’ve had many discussions over the years about autofilling tags from Wikidata and every time it raises a storm of controversy, so I’m surprised we had this functionality all along! You have a good case for removing this functionality as it’s currently implemented. However, just for the record:

> it is questionable whether automatically importing names from wikipedia is compatible with the odbl license

No, the issues with ODbL compatibility only go in the other direction. Wikipedia’s API gets the interwiki link data from Wikidata. Wikidata has a CC0 public domain dedication. OSM can accept public domain data (assuming it really is PD), but a CC0 project may or may not be able to accept data from an ODbL project. There’s a similar dynamic between OSM and OHM, which is also largely CC0.

This doesn’t detract from your other points, however. Just because it’s compatible with the license doesn’t necessarily mean it’s desirable.

PR #11427, opened 2025-09-18
Fix: preserve tag value when changing language in localized name field during multiselection

### LINKED ISSUE #11337: data loss when changing language in localized name field of a multiselection

Ankitgkp:
#11337
Preserves the previous tag value when changing the language in a localized name field during multiselection, preventing data loss.


tordans:
> image

Ankitgkp:
@tordans 
Screenshot 2025-09-18 at 5 05 56 PM
Screenshot 2025-09-18 at 5 06 15 PM

Issue #11436, opened 2025-09-23
Warn the user if a `url` field has a malformed URL


1ec5:
`url` fields such as `website=*` should validate input so that the user knows if they’re entering data that isn’t usable. For example, a schemaless URL like `website=example.com` causes the OSM website to incorrectly link to a relative URL: openstreetmap/openstreetmap-website#6407. Although this is incorrect behavior, it’s a sign that some user-facing data consumers might experience similar problems.

#6494 added a validation warning for malformed URLs in `url` fields such as `website=*`, but 9ccace9022e6c6cb37ea8638fce05cbf26972d14 later commented it out. In the discussion in #6831, we found it redundant to the red border that Firefox applied to malformed input automatically. Unfortunately, no other browser applied that border automatically, and Firefox has removed it as of version 143. (That discussion got derailed by an infeasible idea to automatically upgrade URLs from HTTP to HTTPS.)

We could uncomment the URL validation code, in which case I’d recommend replacing the custom regular expression with code that simply checks if a URL can be constructed:

```js
try {
  entity.tags.website.split(';').map(v => new URL(v));
} catch (e) {
  issues.push(…);
}
```

Alternatively, we could restore the behavior that Firefox used to have, but in every browser, by adding some CSS for the [`:invalid`](https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) pseudoclass.

/ref #8827

1ec5:
If we want to solve #6831, we could do so by detecting whether prepending a protocol would suppress the exception, then offering two suggested fixes to choose from, HTTP or HTTPS. It would be up to the user to determine which one is appropriate.

Issue #8827, opened 2021-11-27
Website link in editor links to openstreetmap.org when http/https prefix missing


andrewpmk:
### How to reproduce the issue?

Website link redirects to openstreetmap.org when http:// or https:// is missing.

To reproduce:
- Enter a website link in the website field of editor with the http:// or https:// prefix missing
- Click the "visit website" link to the right of the website (the arrow pointing up and to the right)
- For example, I was editing a funeral home and entered website www.turnerporter.ca
- Click on link, redirects me to https://www.openstreetmap.org/www.turnerporter.ca
- Should either redirect to https://turnerporter.ca/, or display a warning/error that the http:// or https:// prefix is missing and suggest that you add the prefix

### Screenshot(s) or anything else?

![image](https://user-images.githubusercontent.com/10249827/143688251-6c13d970-957f-4d8e-8c64-fff151b64c9c.png)
![image](https://user-images.githubusercontent.com/10249827/143688259-5f2caa85-416f-4390-b9b1-e5a41a4c7bd4.png)

### Which iD Editor versions do you see the issue on?

Released version at openstreetmap.org/edit

### Which browsers are you seeing this problem on?

Firefox

1ec5:
I think the field was set to `type="url"` at one point, so that the browser would highlight a malformed URL, as seen in https://github.com/openstreetmap/iD/issues/5658#issuecomment-450235505. These days, it’s just a `type="text"`. I wonder if that changed as part of factoring out id-tagging-schema as a separate repository (#8175 etc.).

tyrasd:
Btw: there used to be a validator for this, but it was removed in #6831

PR #11438, opened 2025-09-23
Restore URL validation for website and url fields

Ankitgkp:
#11436 
Improves URL validation in invalid_format.js by replacing regex-based validation with URL constructor validation as recommended in #8827. Validates multiple URL fields (website, url, contact:website, etc.) and handles semicolon-separated values. Prevents malformed URLs like "example.com" that cause incorrect relative linking behavior.

### DIFF
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -50,7 +50,10 @@ export function validationFormatting() {
         Object.keys(entity.tags).forEach(function(tag) {
             if (tag === 'image') return; // skip image
             if (!/(website|url)/i.test(tag)) return; // only process website/url tags
-            var value = entity.tags[tag].trim();
+            var raw = entity.tags[tag];
+            if (!raw) return; // skip empty or undefined
+            var value = raw.trim();
+            if (!value) return; // skip empty string after trim
             if (value.includes(';')) {
                 // If semicolon present, validate each part
                 var parts = value.split(';').map(function(s) { return s.trim(); });
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -46,67 +46,65 @@ export function validationFormatting() {
                 .call(t.append('issues.invalid_format.website.reference'));
         }
 
-        // URL field validation - check multiple possible URL tags (excluding image which allows File: format)
-        const urlTags = ['website', 'url', 'website:mobile', 'contact:website', 'contact:url', 'source:website', 'source:url'];
-
-        urlTags.forEach(function(tag) {
-            if (entity.tags[tag]) {
-                var value = entity.tags[tag].trim();
-                if (value.includes(';')) {
-                    // If semicolon present, validate each part
-                    var parts = value.split(';').map(function(s) { return s.trim(); });
-                    var invalidParts = parts.filter(function(x) { return !isValidURL(x); });
-                    if (invalidParts.length) {
-                        // Always warn if any split parts are invalid
-                        issues.push(new validationIssue({
-                            type: type,
-                            subtype: 'website',
-                            severity: 'warning',
-                            message: function(context) {
-                                var entity = context.hasEntity(this.entityIds[0]);
-                                return entity ? t.append('issues.invalid_format.website.message' + this.data,
-                                    { feature: utilDisplayLabel(entity, context.graph()), site: invalidParts.join(', ') }) : '';
-                            },
-                            reference: showReferenceWebsite,
-                            entityIds: [entity.id],
-                            hash: tag + '=' + invalidParts.join(),
-                            data: (invalidParts.length > 1) ? '_multi' : ''
-                        }));
-                    } else if (!isValidURL(value)) {
-                        // All split parts valid, but whole value still invalid
-                        issues.push(new validationIssue({
-                            type: type,
-                            subtype: 'website',
-                            severity: 'warning',
-                            message: function(context) {
-                                var entity = context.hasEntity(this.entityIds[0]);
-                                return entity ? t.append('issues.invalid_format.website.message',
-                                    { feature: utilDisplayLabel(entity, context.graph()), site: value }) : '';
-                            },
-                            reference: showReferenceWebsite,
-                            entityIds: [entity.id],
-                            hash: tag + '=' + value,
-                            data: ''
-                        }));
-                    }
-                } else {
-                    // No semicolon, validate whole value
-                    if (!isValidURL(value)) {
-                        issues.push(new validationIssue({
-                            type: type,
-                            subtype: 'website',
-                            severity: 'warning',
-                            message: function(context) {
-                                var entity = context.hasEntity(this.entityIds[0]);
-                                return entity ? t.append('issues.invalid_format.website.message',
-                                    { feature: utilDisplayLabel(entity, context.graph()), site: value }) : '';
-                            },
-                            reference: showReferenceWebsite,
-                            entityIds: [entity.id],
-                            hash: tag + '=' + value,
-                            data: ''
-                        }));
-                    }
+        // Refactored: Iterate all tags, skip 'image', process tags matching /(website|url)/
+        Object.keys(entity.tags).forEach(function(tag) {
+            if (tag === 'image') return; // skip image
+            if (!/(website|url)/i.test(tag)) return; // only process website/url tags
+            var value = entity.tags[tag].trim();
+            if (value.includes(';')) {
+                // If semicolon present, validate each part
+                var parts = value.split(';').map(function(s) { return s.trim(); });
+                var invalidParts = parts.filter(function(x) { return !isValidURL(x); });
+                if (invalidParts.length) {
+                    // Always warn if any split parts are invalid
+                    issues.push(new validationIssue({
+                        type: type,
+                        subtype: 'website',
+                        severity: 'warning',
+                        message: function(context) {
+                            var entity = context.hasEntity(this.entityIds[0]);
+                            return entity ? t.append('issues.invalid_format.website.message' + this.data,
+                                { feature: utilDisplayLabel(entity, context.graph()), site: invalidParts.join(', ') }) : '';
+                        },
+                        reference: showReferenceWebsite,
+                        entityIds: [entity.id],
+                        hash: tag + '=' + invalidParts.join(),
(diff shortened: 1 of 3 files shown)


modules/validations/invalid_format.js
```diff
@@ -40,31 +43,35 @@ export function validationFormatting() {
...
 
-        if (entity.tags.website) {
-            // Multiple websites are possible
-            // If ever we support ES6, arrow functions make this nicer
-            var websites = entity.tags.website
-                .split(';')
-                .map(function(s) { return s.trim(); })
-                .filter(function(x) { return !isSchemePresent(x); });
+        // URL field validation - check multiple possible URL tags
+        const urlTags = ['website', 'url', 'website:mobile', 'contact:website', 'contact:url', 'image', 'source:website', 'source:url'];
```
tordans: For `image` the „File:…“ format is not uncommon https://taginfo.openstreetmap.org/keys/image#values
I wonder if we should validate this here or if we should add a exception for this format.

Ankitgkp:
@tordans I've fixed the issue by excluding the image tag from URL validation to allow "File:..." values.

test/spec/validations/invalid_format.js
```diff
@@ -0,0 +1,119 @@
...
+            expect(issues).to.have.lengthOf(2);
+            issues.forEach(function(issue) {
+                expect(issue.type).to.eql('invalid_format');
+                expect(issue.subtype).to.eql('website');
+            });
+        });
+
+        it('should handle multiple URLs separated by semicolons', function() {
+            var entity = createPointWithTags({
+                website: 'https://example.com;invalid-url;http://valid.org'
```
1ec5, line 55: The [`website=*`](https://wiki.openstreetmap.org/wiki/Key:website) documentation recommends percent-encoding the semicolon `;` just in case, but technically it is a valid URL character per RFC&nbsp;1738. So `https://example.com;invalid-url;http://valid.org` is a valid URL, at least in Firefox.

Ideally mappers would percent-encode the semicolon, because otherwise it could be [OSM’s standard value delimiter](https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator). However, that’s a bit much to task of laypeople, especially since some browsers like Firefox no longer percent-escape on their behalf when copying from the location bar. Semicolons aren’t super common for the `website=*` key but they do sometimes occur in some of the other URL keys, about [700 times](https://qlever.cs.uni-freiburg.de/osm-planet/AgZe7E) in all.

If after splitting on a semicolon, the first URL is valid and the subsequent ones are not, we could try validating the entire tag value instead before deciding that it’s invalid.

Ankitgkp:
@tordans I've updated the URL validation to first use strict parsing, but also accept URLs with semicolons and other browser-accepted formats as a fallback. This should prevent valid URLs from being incorrectly flagged. Please review, thanks!

test/spec/validations/invalid_format.js
tordans: @1ec5 does 915afe31ab83c02f92ddf533a776127763fd1e58 match what you hand in mind? To me it looks the wrong way around. It should be: First check separate and only check the whole as a last effort. Right?

Ankitgkp:
@tordans Updated URL validation logic: we now always warn if any semicolon-separated part is invalid, and only fall back to validating the full value if all parts are valid.

modules/validations/invalid_format.js
```diff
@@ -40,31 +46,70 @@ export function validationFormatting() {
...
 
-        if (entity.tags.website) {
-            // Multiple websites are possible
-            // If ever we support ES6, arrow functions make this nicer
-            var websites = entity.tags.website
-                .split(';')
-                .map(function(s) { return s.trim(); })
-                .filter(function(x) { return !isSchemePresent(x); });
+        // URL field validation - check multiple possible URL tags (excluding image which allows File: format)
+        const urlTags = ['website', 'url', 'website:mobile', 'contact:website', 'contact:url', 'source:website', 'source:url'];
```
hlfan: I'd rather not have the parts repeated.
`uriTag = /\b(website|url)\b/` and a `for` loop block with `if (!uriTag.test(tag)) continue;` could clean it up.

test/spec/validations/invalid_format.js
1ec5: Agreed, otherwise, the validator wouldn’t flag the example I gave earlier, because the whole string would _technically_ be a valid URL, even though the mapper likely didn’t intend it to be a single URL.

modules/validations/invalid_format.js
Ankitgkp: Replaced the hard-coded urlTags array with a regex check over all entity.tags, skipping image since File: values are allowed. The semicolon handling and validation logic remain unchanged.

modules/validations/invalid_format.js
```diff
@@ -40,31 +46,71 @@ export function validationFormatting() {
...
-                    },
-                    reference: showReferenceWebsite,
-                    entityIds: [entity.id],
-                    hash: websites.join(),
-                    data: (websites.length > 1) ? '_multi' : ''
-                }));
+        // Refactored: Iterate all tags, skip 'image', process tags matching /(website|url)/
+        Object.keys(entity.tags).forEach(function(tag) {
+            if (tag === 'image') return; // skip image
+            if (!/(website|url)/i.test(tag)) return; // only process website/url tags
```
hlfan: This check also discards `image` tags, so the check above isn't necessary.
On the other side, it doesn't discard keys like `traffic:hourly`.

Ankitgkp: Updated the URL validation loop to follow your feedback
Removed the redundant image check (regex already excludes it)
Tightened the regex to`/\b(website|url)\b/i` so tags like `traffic:hourly`are no longer incorrectly matched
Continued to iterate dynamically over `entity.tags` instead of using a hard-coded list

modules/validations/invalid_format.js
```diff
@@ -40,31 +46,68 @@ export function validationFormatting() {
...
-            var websites = entity.tags.website
-                .split(';')
-                .map(function(s) { return s.trim(); })
-                .filter(function(x) { return !isSchemePresent(x); });
+        Object.keys(entity.tags).forEach(function(tag) {
+            if (!/\b(website|url)\b/i.test(tag)) return;
+            var raw = entity.tags[tag];
+            if (!raw) return;
+            var value = raw.trim();
+            if (!value) return;
```
hlfan:
```suggestion
        Object.entries(entity.tags).forEach(function([key, tag]) {
            if (!/\b(website|url)\b/i.test(key)) return;


            var value = tag?.trim?.();
            if (!value) return;
```

modules/validations/invalid_format.js
```diff
@@ -40,31 +46,68 @@ export function validationFormatting() {
...
+                            var entity = context.hasEntity(this.entityIds[0]);
+                            return entity ? t.append('issues.invalid_format.website.message',
+                                { feature: utilDisplayLabel(entity, context.graph()), site: value }) : '';
+                        },
+                        reference: showReferenceWebsite,
+                        entityIds: [entity.id],
+                        hash: tag + '=' + value,
+                        data: ''
+                    }));
+                }
```
hlfan, line 102: Most of the issue data could be built independently of the validity:
```js
const websiteValidationIssueBase = {
	type: type,
	subtype: 'website',
	severity: 'warning',
	message: function(context) {
		var entity = context.hasEntity(this.entityIds[0]);
		return entity ? t.append('issues.invalid_format.website.message' + this.data,
			{ feature: utilDisplayLabel(entity, context.graph()), site: this.value }) : '';
	},
	reference: showReferenceWebsite,
	entityIds: [entity.id],
	value,
	hash: tag + '=' + value,
	data: ''
};
```
```suggestion
                if (invalidParts.length) {
                    issues.push(new validationIssue({
                        ...websiteValidationIssueBase,





                        value: invalidParts.join(', '),



                        hash: tag + '=' + invalidParts.join(),
                        data: (invalidParts.length > 1) ? '_multi' : ''
                    }));
                } else if (!isValidURL(value)) {
                    // All split parts valid, but whole value still invalid
                    issues.push(new validationIssue({
                        ...websiteValidationIssueBase











                    }));
                }
            } else {
                // No semicolon, validate whole value
                if (!isValidURL(value)) {
                    issues.push(new validationIssue({
                        ...websiteValidationIssueBase











                    }));
                }
```

hlfan: I should've probably added that I meant the multiple newlines only as a diff aid, IMO these can all be collapsed.

modules/validations/invalid_format.js
```diff
@@ -40,31 +46,62 @@ export function validationFormatting() {
...
-                }));
+        Object.entries(entity.tags).forEach(function([key, tag]) {
+            if (!/\b(website|url)\b/i.test(key)) return;
+            if (!tag) return;
+            var value = tag.trim();
+            if (!value) return;
+            if (value.includes(';')) {
+
+                var parts = value.split(';').map(function(s) { return s.trim(); });
+                var invalidParts = parts.filter(function(x) { return !isValidURL(x); });
```
hlfan, line 70:
```suggestion
                var parts = value.split(';').map(s => s.trim());
                var invalidParts = parts.filter(x => !isValidURL(x));
```

modules/validations/invalid_format.js
```diff
@@ -40,31 +46,62 @@ export function validationFormatting() {
...
+                        hash: key + '=' + value,
+                        data: ''
+                    }));
+                }
+            } else {
+                // No semicolon, validate whole value
+                if (!isValidURL(value)) {
+                    issues.push(new validationIssue({
+                        ...websiteValidationIssueBase,
+
```
hlfan, line 96: As for https://github.com/openstreetmap/iD/issues/11436#issuecomment-3323878608:
```suggestion
                        key: key,
                        fixes: [{ protocol: 'https', icon: 'temaki-lock' }, { protocol: 'http', icon: 'iD-icon-info-filled' }]
                            .filter(f => isValidURL(...))
                            .map(f => new validationIssueFix({
                                icon: f.icon,
                                title: t('issues.fix.add_' + f.protocol + '.title'),
                                onClick: function(context) {...}
                            })),
```

hlfan (commented):
Apart from my other comments, the messages in `/data/core.yaml` are still missing.

Issue #11437, opened 2025-09-23
Walkthrough gets stuck after "delete point" step


SergeiPatiakin:
### How to reproduce the issue?

- Register as a new user
- Start the walkthrough for iD. Follow the instructions without skipping any steps
- In the Points section, follow the instructions for the "delete point" step

### Expected Behavior
- After successfully deleting a point, iD should provide instructions for the next step

### Observed Behavior
- After successfully deleting a point, iD provides no instructions

### Screenshot(s) or anything else?

<img width="1506" height="740" alt="Image" src="https://github.com/user-attachments/assets/15a18d95-40d0-4721-8824-3ee1b51fe20a" />

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.36.0

### Which browsers are you seeing this problem on?

Chrome

tyrasd:
fixed by #11442

clorofolle:
I am having the same exact bug still. I've tried both Firefox and Chrome. 
Should I open a new issue? The issue is completely the same as the one described here. 
The only difference in my case is that the language is set to "Italian".
The editor is still version 2.36.0, but I saw the bugfix was deployed?
Thank you.

<img width="2986" height="1500" alt="Image" src="https://github.com/user-attachments/assets/0c4b7f79-c065-40f8-a589-ad19e186048b" />

tyrasd:
The bugfix was not yet rolled out to the osm.org website. This will happen with v2.37 in the coming days. Sorry for the delay. :bow:

PR #11442, opened 2025-09-25
Adds optional chaining to hasEntity call (Fix Issue #11437)

bgo-bell:
I also experienced Issue #11437 when onboarding to openstreetmap recently and therefore thought it would be a fitting first issue to fix!

Upon investigation of the problem, I believe the reason it's not progressing to the next "undo" step in the walkthrough is because there is an error when trying to hide the labels near the mouse; by the time it is trying to hide the labels, the point has already been deleted, thus the call `graph.hasEntity(id)` returns `undefined`.

This then causes the error as seen below in the screenshot of the console. 

There are 3 different ways I thought of approaching this problem: 

1. **Change the logical ordering of events,** such that this labeling event happens before a deleted point event so that the error isn't thrown in the first place (i.e. this label hide can successfully happen). This seemed like both the most involved and most intrusive change. As a new contributor, I definitely don't have the knowledge of this project to make a change this big and therefore decided against it. 
2. **Change the `filterLabels` function** such that it conditionally filters based on if the entity is still available. This seemed like a good solution and I am potentially interested in moving forward with this one, but why I eventually decided against implementing it is similar to the reason cited above- as a new contributor I'm afraid of the other affects this could have throughout the project that I am not yet aware of at this point :) 
3. **Add some (slight) error handling.** This is the approach that I ended up opting for, through adding optional chaining to the `hasEntity` call within the `filterLabels` function. This effectively eliminates the TypeError as it safeguards against an undefined value being returned by `hasEntity`. _That being said,_ I am slightly concerned about the possibility of this not fully protecting against a TypeError, as I was having a hard time finding the definition of the `hasEntity` function, and am not sure if it could return something like a boolean, in which case this TypeError will persist. 

Anyways this is lots of words for a one character change LOL. Excited for a review and (hopefully) my first contribution <3 

Screenshot 2025-09-24 at 10 09 50 PM

### DIFF
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -719,7 +719,7 @@ export function svgLabels(projection, context) {
                     // in select mode: hide labels of currently selected line(s)
                     // to still allow accessing midpoints
                     // https://github.com/openstreetmap/iD/issues/11220
-                    context.mode().selectedIDs().includes(id) && graph.hasEntity(id).geometry(graph) === 'line');
+                    context.mode().selectedIDs().includes(id) && graph.hasEntity(id)?.geometry(graph) === 'line');
             hideIds.push.apply(hideIds, nearMouse);
             hideIds = utilArrayUniq(hideIds);
         }


tyrasd (approved):
Good catch, thanks for the fix! :bow:

PR #11484, opened 2025-10-15
Added info that there are more presets than visible on the screen

### LINKED ISSUE #11465: Add info that there are more presets than visible on the screen
Add a node. iD will suggest a few presets - at the top, the last ones used, and the first time you turn on the editor, there will only be about 10 presets.
When I started mapping on OSM, I thought that was all I could choose! Therefore, I suggest adding information at the very end, e.g.
`<text>`: "These are not all the possibilities. To see more, enter the name of the store or place you are looking for in the search box."

There are more features for lines and landuse to choose, but for nodes there are only a few. I think this information should be in all of them (nodes, ways, areas).

Image

akshat99812:
Previously, when adding a node in the editor, users only saw a limited set of presets (top suggestions and initial ~10 presets). This caused confusion for new mappers, who might think these are the only options available.

This PR adds an informational note at the start of the preset list
Screenshot 2025-10-16 at 2 36 05 AM

Closes: #11465

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1086,6 +1086,12 @@ a.hide-toggle {
     background: var(--bg-color-3);
 }
 
+.preset-search-info {
+    padding: 10px;
+    font-size: 12px;
+    color: #666;
+}
+
 .preset-icon-container {
     position: relative;
     width: 60px;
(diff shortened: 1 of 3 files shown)


tyrasd:
See https://github.com/openstreetmap/iD/issues/11465#issuecomment-3453100048

maro-21:
I would be in favor of having the text at the bottom after the user scrolls to the bottom of the list and sees there are no more options.

PR #11499, opened 2025-10-22
Restore URL validation

### LINKED ISSUE #6831: Offer to fix missing http:// in website tag
A website tag without http:// is highlighted as an issue - it would be good if iD offered to "upgrade" or fix this by adding the protocol.

Also, you'll notice I had to scroll down to actually see the website field. Perhaps iD could move fields which are an issue higher up in the list, or even into the issues section just below the yellow box?
website

### LINKED ISSUE #11436: Warn the user if a `url` field has a malformed URL

hlfan:
Continuing #11438 to close #11436 and resolve #6831.

### DIFF
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -1,6 +1,7 @@
 import { t } from '../core/localizer';
 import { utilDisplayLabel } from '../util/utilDisplayLabel';
-import { validationIssue } from '../core/validation';
+import { validationIssue, validationIssueFix } from '../core/validation';
+import { actionChangeTags } from '../actions/change_tags';
 
 export function validationFormatting() {
     var type = 'invalid_format';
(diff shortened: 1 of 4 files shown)


modules/validations/invalid_format.js
```diff
@@ -26,11 +27,28 @@ export function validationFormatting() {
...
 
-        /* see https://github.com/openstreetmap/iD/issues/6831#issuecomment-537121379
-        function isSchemePresent(url) {
-            var valid_scheme = /^https?:\/\//i;
-            return (!url || valid_scheme.test(url));
+        function isValidURL(url, strict = false) {
+            try {
+                // First try strict WHATWG parsing
+                const link = new URL(url);
+                return link.protocol === 'http:' || link.protocol === 'https:';
```
1ec5: I shouldn’t be surprised, but there are actually [a handful of (S)FTP website values](https://overpass-turbo.eu/s/2eAl) globally, mostly in New Zealand, and they actually make some sense. They’re pretty rare, but are they valid? Some browsers like Firefox have dropped native support for FTP in recent years, making it a less user-friendly option for `website=*`. (I could not find any Telnet or Gopher URLs, for what it’s worth.)

1ec5 (commented):
Can you summarize what’s changed since #11438, if anything?

hlfan:
The biggest addition are the dynamic fixes, but I also flattened the issue creation logic a bit and added messages in core.yaml.

See [this diff](https://github.com/Ankitgkp/iD/compare/663e4919d1453b52f1c936b6dc922522a5ba5e1a..hlfan:iD:url-format).

modules/validations/invalid_format.js
hlfan: That check was only intended to catch wikimedia_commons file prefixes, but the automatic insertion of slashes is the better option here.

1ec5 (approved):
This addresses the feedback I had in #11438.

PR #11511, opened 2025-10-24
fix layer tag overriden when adding a bridge or tunnel

k-yle:
if a line already has `highway=*` + `layer=-3`, and you click on the `tunnel` option in the UI, then it will replace `layer=-3` with `layer=-1` which is unexpected. 

However, if the `layer` is >0, then it does make sense to change it to -1.

The opposite applies for bridges. 

So, instead of blindly overriding the `layer` tag with `1` or `-1`, we now have a more nuanced solution which respects existing tag values (if they make sense).

see unit tests for more examples.

### DIFF
--- a/modules/ui/fields/radio.js
+++ b/modules/ui/fields/radio.js
@@ -20,6 +20,7 @@ export function uiFieldRadio(field, context) {
     var radioData = (field.options || strings.options || field.keys).slice();  // shallow copy
     var typeField;
     var layerField;
+    let _tags = {};
     var _oldType = {};
     var _entityIDs = [];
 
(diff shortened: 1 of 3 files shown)


modules/ui/fields/radio.js
```diff
@@ -263,6 +270,7 @@ export function uiFieldRadio(field, context) {
 
 
     radio.tags = function(tags) {
+        _tags = tags;
```
tordans, line 273: Does this do anything?

k-yle: yes, previously this component did not store the `tags` in it's local state (`_tags`). but now this component needs access to all the other tags, so we have to store the value. 

the name clash is it a bit weird, in this codebase [usually](https://grep.app/search?f.repo=openstreetmap%2FiD&q=+%3D+function%28val%29+%7B) the setter function calls the argument `val`

tyrasd (approved):

Issue #7508, opened 2020-04-11
Replace Transifex


comradekingu:
Following the logic of OSM and iD being a place free of tracking not by happen-chance,
it would be great if the translation platform followed suit. To me that is one of the central points of what is and should be the greater OSM ecosystem.

Transifex is not libre software, so it is a bad match all together.

As a translator I find Transifex infuriating, and I can't even log in to contribute to the OSM resource, as doing so would grant Transifex a license to use all my strings for their closed translation memory, which is an updated policy requirement they instated the day EU-GDPR went into effect.

Transifex also doesn't let anyone browse strings freely, so the potential for malice is greater. I'd link to the egregious terms and conditions if the whole site wasn't down right now. Requiring contributors to be over 18 is another detriment to its utility as a platform for OSM.

My recommendation is to use [Weblate](https://weblate.org/), either in as a self-hosted instance, or Hosted Weblate. https://weblate.org/hosting/
https://hosted.weblate.org/legal/ is a far cry from the likes of Transifex and Crowdin, but any platform that can compare itself to Weblate is good with me.

Edit: Full disclosure, I have contributed language fixes to Weblate, and other libre software translation platforms. I play a small role in OSM-NO. I have used all platforms extensively.

1ec5:
Thanks for starting this conversation. At the time this project chose Transifex, that platform was more open (open source too), and the obvious alternatives had significant downsides: #909. For example, Translatewiki.net required access to push changes to the repository automatically. (It looks like [that’s an option with Weblate](https://docs.weblate.org/en/latest/faq.html#what-are-the-options-for-pushing-changes-back-upstream) but there are alternative workflows.)

For what it’s worth, iD currently uses Transifex for free by virtue of being an open source project. Functionally speaking, the Transifex features that matter most to this project would include:

* Translation memory (since some terms are repeated many times across presets and across the community index)
* Maintainer-provided context, support for a decent translation format (YAML is OK but not the only possibility)
* Access to a community of translators who are familiar with translating OSM software (which we currently get by being part of the OpenStreetMap organization at Transifex)

(In my opinion, built-in plural/grammar support would be nice too.)

While we do have a copy of all the current translations, simply hooking the repository up to a different service would lose history and translation memory, so hopefully there’s a way to export that from Transifex. More importantly, the existing translators would need to migrate with us, so there would need to be ample communication.

[...]

PR #11512, opened 2025-10-24
Update CONTRIBUTING.md with further details about Transifex signup

andrewharvey:
After a contributor was unable to create an account to help with translations due to Transifex only allowing "business email's", we discovered there is a specific signup flow that must be followed to allow contributors to create accounts without a "business email".


andrewharvey:
another reason to replace transifex.... #7508

tyrasd (approved):

PR #11513, opened 2025-10-24
Change indent_size to tab in .editorconfig

malwikt:
Neovim shows this warning:
'editorconfig: invalid value for option indent_size: tab_width. indent_size must be a number'

Per specification it should either be a number or a 'tab'.

https://spec.editorconfig.org/#supported-pairs
```Set to a whole number defining the number of columns used for each indentation level and the width of soft tabs (when supported). If this equals tab, the indent_size shall be set to the tab size, which should be tab_width (if specified); else, the tab size set by the editor. The values are case-insensitive.```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,4 +16,4 @@ indent_size = 4
 [*.{js,mjs}]
 end_of_line = lf
 insert_final_newline = true
-indent_size = tab_width
+indent_size = tab


tordans:
According to https://github.com/openstreetmap/iD/pull/11313#issuecomment-3200198252 we can cleanup the file a bit more even, to keep it simpler: https://github.com/openstreetmap/iD/pull/11313/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5b

Wanted to extract this into a separate PR. Maybe you have time for a follow up?

PR #11526, opened 2025-10-29
Panel button overlap

### LINKED ISSUE #11516: Panel close button UI overlap
### URL

https://www.openstreetmap.org/edit?editor=id#map=16/21.38084/79.74344

## Description

When two info panels are opened in the sidebar, the last panel’s close button or edge content occasionally gets hidden behind the side menu, making it difficult for users to close or interact with it.

This happens because the last `.panel-container` element doesn’t have enough right-side spacing to remain fully visible when panels are stacked horizontally or when the layout compresses.

---

## Steps to Reproduce

1. Open one or more info panels in iD Editor.
2. Resize the window or view on a narrower browser width.
3. Observe that the last panel’s edge is hidden behind the menu area — the close button becomes difficult to click.

---

## Screenshots

**Before fix:** Before Fix 1 Before Fix 2

**After UI changes:** After Fix 1 After Fix 2

Already attached above

2.37.0-dev

Chrome

ved17-git:
This PR fixes a layout issue where the last opened info panel in the sidebar would partially hide behind the side menu.
Added right-side margin to .info-panels Updated CSS to ensure each panel remains fully visible.

Fixes #11516

image


css/80_app.css
```diff
@@ -4615,6 +4615,8 @@ img.tile-debug {
     -ms-user-select: element;
     pointer-events: none;
     overflow: hidden;
+    margin: 0 40px 2px 0
+
```
tordans, line 4619:
```suggestion
    margin: 0 40px 2px 0;
```

tyrasd:
This does not work well if the panel and menu don't actually overlap, and does also not work if iD is in right-to-left mode. See https://github.com/openstreetmap/iD/issues/11516#issuecomment-3469113828

Issue #11532, opened 2025-10-30
after uploading a changeset, both the changeset details and presets list are rendered overlapping in the panel


andrewharvey:
### How to reproduce the issue?

after uploading a changeset, both the changeset details and presets list are rendered overlapping in the panel

### Screenshot(s) or anything else?

<img width="679" height="1000" alt="Image" src="https://github.com/user-attachments/assets/87c6a03d-adf7-46da-9f84-088195953ae0" />

Vectorial1024:
Indeed, I also noticed this as #11530 but I wrongly thought it would be some sort of CDN laziness. I am here to report that unfortunately I also see this bug while using `2.37.2`:

<img width="399" height="463" alt="Image" src="https://github.com/user-attachments/assets/3ee8becb-9462-488a-a7b9-79afdbd34b46" />

radiotrefoil:
In addition;

- the Upload to OpenStreetMap sidebar is unable to be closed, and the previous editing session reappears when re-entering edit mode until the page is refreshed.
- The Sources dropdown menu has disappeared

novolife:
Yes, if you click Save and then cancel, the left panel will be covered by the upload panel and cannot be selected. After uploading, click on a feature to continue editing, the upload panel will still cover the feature panel.

novolife:
@tyrasd This could be a T0 bug😁. Please check it asap.

novolife:
I tested the [development mirror of develop branch](https://ideditor.netlify.app/) and this issue cannot be reproduced, but in the [stable mirror of release branch](https://ideditor-release.netlify.app/). Not sure whether the issue is with the OSM website or if [a commit](https://github.com/openstreetmap/iD/commits/develop/) fixed it.

maxwellward:
EDIT: as also mentioned by novolife, I cannot replicate this issue on the dev deployment, `2.38.0-dev`, so this may already have been fixed.

I can confirm I'm also experiencing this issue on Zen 1.14.9b (which is Firefox 141.0 under the hood) on MacOS Tahoe 26.0.1. iD editor version 2.37.2.

I can reproduce it 100% of the time with the following steps:

- Click the initial upload button but do not finish the changeset upload
- Close the upload panel
- Observe the upload panel displays over the details panel

I have heard of a number of others along with someone on Mastodon who are also experiencing this.

tyrasd:
[...]

PR #11533, opened 2025-10-31
fix: guard against undefined options in source filtering

hlfan:
Adds another guard to prevent #11532.

### DIFF
--- a/modules/ui/changeset_editor.js
+++ b/modules/ui/changeset_editor.js
@@ -89,7 +89,7 @@ export function uiChangesetEditor(context) {
                     // based on the values used in recent changesets.
                     const recentSources = changesets
                         .flatMap((changeset) => changeset.tags.source?.split(';'))
-                        .filter(value => !sourceField.options.includes(value))
+                        .filter(value => !sourceField.options?.includes(value))
                         .filter(Boolean)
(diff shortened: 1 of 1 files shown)


k-yle (commented):
this does fix the bug, but the proper solution would be to revert [this manual change](https://github.com/openstreetmap/id-tagging-schema/commit/94749b52c955c3c07a02441242fb891037a8c023#diff-0590621fdfad5ca2b4d752006e299978f572e544aac4176f072a27fb76c1db00R801) to the generated `dist` file. 

`source` has `"stringsCrossReference": "{source}"` , which is circular reference, hence why `field.options` is undefined.

Since that edit is not in the main branch, there's probably no point in sending a PR. We'll need @tyrasd to update that file and make a new release

tyrasd:
yeah, I had messed something up when creating the v6.13.3 `dist` release: I had an uncommitted test version of the `source.json` field in my directory which I though I had deleted, but wasn't. Sorry for that.

@hlfan I think there is still merit in merging this safeguard check. It could even come into play when people are running their own instance iD if nonstandard presets. Can you restore your branch, please?

hlfan:
Nice to see the tests pass too now @tyrasd.

tyrasd (approved):

PR #11540, opened 2025-11-02
Reduce duplication in utilKeybinding map initialization

hlfan:
Simplify utilKeybinding.keys and utilKeybinding.keyCodes by generating them programmatically instead of defining large static objects. This removes repeated values and makes the mappings easier to maintain without changing behavior.


modules/util/keybinding.js
```diff
@@ -222,204 +222,158 @@ utilKeybinding.modifierProperties = {
 utilKeybinding.plusKeys = ['plus', 'ffplus', '=', 'ffequals', '≠', '±'];
 utilKeybinding.minusKeys = ['_', '-', 'ffminus', 'dash', '–', '—'];
 
-utilKeybinding.keys = {
+utilKeybinding.keys = (() => {
+    const k = {};
     // Backspace key, on Mac: ⌫ (Backspace)
-    '⌫': 'Backspace', backspace: 'Backspace',
+    k['⌫'] = k.backspace = 'Backspace';
```
matkoniecz, line 228: I am confused why new variant is better here.

Why not define `utilKeybinding.keys = {` and later add num and f entries programatically?

hlfan: I'd like to fit more content on the same screen which is more difficult with this variant:
```suggestion
    utilKeybinding.keys['⌫'] = utilKeybinding.keys.backspace = 'Backspace';
```

matkoniecz: what is wrong with defining 

```
utilKeybinding.keys = {
    // Backspace key, on Mac: ⌫ (Backspace)
    '⌫': 'Backspace', backspace: 'Backspace',
    // Tab Key, on Mac: ⇥ (Tab), on Windows ⇥⇥
    '⇥': 'Tab', '⇆': 'Tab', tab: 'Tab',
    // Return key, ↩
    '↩': 'Enter', '↵': 'Enter', '⏎': 'Enter', 'return': 'Enter', enter: 'Enter', '⌅': 'Enter',
    // Pause/Break key
```

and then adding in loop num and f entries?

In general I am unsure is it worth having loops etc with just 12 entries.

This PR shortens code, but not sure is it really simplified.

(I have no strong opinion, I simply looked at diff - without real knowledge about this part of code)

hlfan: > what is wrong with defining utilKeybinding.keys and then adding in loop num and f entries?

that the keys' keys are not sorted - when I made f434ff4d41ef673fad7ed9f17c3797bda4e9e6b1 I didn't think of adding an exclusion first to demonstrate the issue.

k-yle (commented):
the propose style in this PR is [discouraged by eslint](https://eslint.org/docs/latest/rules/no-multi-assign) and would make it harder to add strict JSDoc annotations to this file, since we can't use the [`keyof` operator](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html) anymore ([test case](https://www.typescriptlang.org/play/?filetype=js#code/PTAEAEBcGcFoGMAWBTeBrAUB+B7AdtJKAOY44AmoAvKAN4aiOgDkg1MTMBcLAQgIbrQADv2TMANAyYAjfmiEiuzPgOHxREgL4BuDMABUeiJACeg5OWQAzOmmTGc10hQ2gA4mUp7gWbPkKgZShpabQxAgG02ZgBdagDecgA6GRUROKVZeTVmHX1DKFNzKxs7B3jyFz5Pbx8sABtkIidyABVkQgBhXmhkOIAKPKNCundnUC9QAYMhszo8ZAA3ZAAnFwm+gAYASh3csAA9AH56xvK2zu7emin8k1naKrWwG5ne2nml1fHn7d3dA8OQA))

tyrasd:
True, there's little benefit of this kind of refactoring. Let's keep it simple.

PR #11553, opened 2025-11-06
Update contribution guideline label links

hlfan:
And move away from the http://labl.es/svg icons


hlfan:
Huh, the label links don't work in the preview...

hlfan:
preview

PR #11560, opened 2025-11-07
Update help text about Street Level Photos providers

danieldegroot2:


k-yle (approved):
lgtm. and i think transifex will invalidate this string automatically

tyrasd (approved):
Thanks for the suggestion. I changed it a bit, so that it does not need to be the complete list of providers. This is because otherwise we'd need to keep the help text in sync every time providers are removed or added. I also noticed somthing odd about Microsoft's streetside service: the previous link to bing streetside is dead (redirecting to an azure maps page that does not mention street level imagery at all), the new bing maps site does not seem to offer street level imagery anymore, and the azure maps API (which is replacing the Bing Maps API) does also not have a replacement for the streetside service. It might be the case that microsoft silently deprecated the service, thus it might have to be removed soon-ish anyway. :shrug: Lastly, I put panoramax as the new first mentioned service, as it is the most "relevant" for the OSM mapping context.

PR #11569, opened 2025-11-10
Pr 11567 fix

jiyansh007:
fix(build): confirm fake-indexeddb@6.2.5 passes on Node 24

Locally verified Dependabot PR #11567.
- Ran full build + tests under Node 22 and 24 (both ✅)
- No code changes required, just confirming compatibility.
- Adds empty commit to retrigger CI for Node 24 environment.

Closes #11567.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,7 +70,7 @@
         "esbuild": "^0.25.12",
         "esbuild-visualizer": "^0.7.0",
         "eslint": "^9.39.1",
-        "fake-indexeddb": "^6.2.4",
+        "fake-indexeddb": "^6.2.5",
         "fetch-mock": "^11.1.1",
         "gaze": "^1.1.3",
         "happen": "^0.3.2",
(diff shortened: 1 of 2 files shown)


jiyansh007:
✅ Tested locally on Node 22 and Node 24 — all tests pass.
Opening pr no: #11569

jiyansh007:
@dependabot  rebase

k-yle (commented):
the CI errors that you're seeing are actually intermittent race conditions, not an issue with the PR itself. 

* For [job54955660322](https://github.com/openstreetmap/iD/actions/runs/19226752923/job/54955660322#step:7:409) from 08e4cecb: the issue is caused by us [running deferred work inside `requestIdleCallback`](https://github.com/openstreetmap/iD/blob/develop/modules/core/validator.js#L789), which is called after the test is unmounted.
* For [job54955682430](https://github.com/openstreetmap/iD/actions/runs/19226760186/job/54955682430#step:7:409) from 08e4cecb: we're [running code inside `setTimeout(..., 75)`](https://github.com/openstreetmap/iD/blob/697accfbd32c80ca480ffb28efca68574ab28f84/modules/ui/combobox.js#L137), but the test runner is so fast that the test finishes before the callback is called.

I don't know any easy solutions, but if you want to tackle this problem, that would be a good place to start

tyrasd:
let's continue to address the issues with the tests in a separate issue: #11579

PR #11571, opened 2025-11-11
Dynamically shift panels only when overlapping menu (fixes #11516)

### LINKED ISSUE #11516: Panel close button UI overlap

maddysib:
## Description
Fixes the issue where panel close buttons get hidden behind the side menu on narrow screens by dynamically detecting actual overlap and only adding margin when necessary.

## What's Changed
- **Dynamic Overlap Detection**: Uses `getBoundingClientRect()` to measure real-time panel/toolbar positions
- **Smart Margin Application**: Adds `margin-right` (or `margin-left` in RTL) **only** when panels actually overlap the menu
- **RTL Support**: Works correctly for right-to-left languages (`#rtl=true`)
- **Performance**: Uses CSS transitions for smooth margin changes during resize

## Technical Details
- Added `adjustPanelOverlap()` function in `modules/ui/info.js` that:
  - Measures toolbar and last panel positions
  - Compares boundaries to detect overlap
  - Toggles `panel-overlaps-menu` class accordingly
- Added CSS in `css/80_app.css` for conditional margin
- Listens to window resize events and panel toggle actions

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4658,6 +4658,22 @@ img.tile-debug {
     background: none;
     color: #ddd;
 }
+/* Add to the END of css/80_app.css */
+
+/* Dynamic margin for overlapping panels */
+.panel-container.panel-overlaps-menu {
+  margin-right: 12px; /* Adjust based on your testing */
+}
+
+[dir="rtl"] .panel-container.panel-overlaps-menu {
+  margin-left: 12px;
+  margin-right: 0; /* Reset LTR margin */
+}
+
+/* Smooth transition */
+.panel-container {
+  transition: margin 0.2s ease;
+}
 .ideditor[dir='rtl'] .panel-title button.close {
     float: left;
 }
--- a/modules/ui/info.js
+++ b/modules/ui/info.js
@@ -21,7 +21,33 @@ export function uiInfo(context) {
             active[k] = false;
         }
     });
+function adjustPanelOverlap() {
+        var panels = document.querySelectorAll('.panel-container');
+        if (panels.length === 0) return;
 
+        var toolbar = document.querySelector('.top-toolbar');
+        if (!toolbar) return;
+
+        var isRTL = document.documentElement.getAttribute('dir') === 'rtl';
+        var toolbarRect = toolbar.getBoundingClientRect();
+        var lastPanel = panels[panels.length - 1];
+        
+        if (!lastPanel) return;
+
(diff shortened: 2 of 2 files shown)


modules/ui/info.js
```diff
@@ -77,13 +103,15 @@ export function uiInfo(context) {
 
 
             // redraw the panels
+
```
matkoniecz, line 106: these empty lines are not needed, right?

modules/ui/info.js
```diff
@@ -140,6 +170,6 @@ export function uiInfo(context) {
                 });
         });
     }
-
+window.addEventListener('resize', adjustPanelOverlap);
```
matkoniecz, line 173: another whitespace-only review: this should be more intended, right?

css/80_app.css
```diff
@@ -4658,6 +4658,22 @@ img.tile-debug {
     background: none;
     color: #ddd;
 }
+/* Add to the END of css/80_app.css */
```
matkoniecz, line 4661: it seems you copy-pasted instructions from LLM or from whatever else gave this code to you

css/80_app.css
```diff
@@ -4658,6 +4658,22 @@ img.tile-debug {
...
+
+/* Dynamic margin for overlapping panels */
+.panel-container.panel-overlaps-menu {
+  margin-right: 12px; /* Adjust based on your testing */
```
matkoniecz, line 4665: `  margin-right: 12px; /* Adjust based on your testing */`

it looks like this is not a serious PR

tyrasd:
closing as stale

Issue #11445, opened 2025-09-27
wrong last modified of a way and a relation when changing a node


Mahabarata:
### URL

https://www.openstreetmap.org/edit?node=13114459674#map=19/-3.276268/11.444414

### How to reproduce the issue?

1. follow the link above
2. delete the wrong name and save your change
3. update the page : on the left panel, the node is indicated with a last modification by you a few second ago = the information is true
4. select the way that uses this node : the way is indicated with a last modification by you a few second ago. It is false : the way is exactly the same as before : same tags, same number of nodes, same location for all nodes,
5. select the relation of that way : the relation is indicated with a last modification by you a few second ago. It is false : the relation is exactly the same as before your saved it : same tags, same number of members, same number of nodes, and all members and nodes are at the same locations.

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.36.0

### Which browsers are you seeing this problem on?

Firefox

tyrasd:
Yes, in this particular case, the indicated time is not what one would expect. But that is mostly a limitation rooted in the data that iD has available from the OSM API: from a normal `map` call result it is not possible to distinguish the case of a pure tag edit from one where the mapper has actually (indirectly) changed the geometry of the way. Also, please note that the indicated time is not meant to be an authoritative value of the last modified[^1] time of the selected feature[^1]. The main goal is to let the  mappers know when the feature was last touched[^2] by another mapper, in order to understand whether the feature has been sitting around for a decade or if it was recently inspected by another mapper. For that purpose it is more useful to also consider child-edits instead of omitting them. See also https://github.com/openstreetmap/iD/pull/7629#discussion_r2242279753.

[^1]: definition pending
[^2]: Thinking of it: it might be worth to rename the text from "last modified …" to "last touched …" to indicate this intent more clearly 🤔 

PS: Please don't use the main OSM database for [test](https://www.openstreetmap.org/node/13114459674/history/2) data. There is a dedicated sandbox version of the OSM stack available at https://api06.dev.openstreetmap.org/ for such uses. 😊

Mahabarata:
I agree with the second footnote. Because you closed this issue, could you change that text a soon as possible, so as not to forget this improvement ?

k-yle:
I'm going to reopen this one so that we have 1 place to discuss future optimisations. Currently there are various comments spread across 5 different https://github.com/openstreetmap/iD/labels/wontfix issues.

[...]

PR #11576, opened 2025-11-13
Fixed last modified

### LINKED ISSUE #11575: In edit mode, selected way last modified shows updates from nodes not way itself
### URL

https://www.openstreetmap.org/edit?way=90774946#map=20/53.7405477/-2.0169201

* The way is shown in the bottom-left of the page as "[Last modified 13 months ago by ubegxj](https://www.openstreetmap.org/way/84198108/history)" 
* However if you click through to the object's history the actual details are given as "Edited over 5 years ago by [Fekneejit](https://www.openstreetmap.org/user/Fekneejit)"
* This is because the _geometry_ of the way's nodes were updated more recently by the user mentioned first, but not the tags on the way itself. The presentation of the "last modified" value is not limiting itself to the way alone, which was the expected behaviour.

_No response_

2.37.3

Chrome

ChehakTrehan:
Fixed incorrect 'Last modified' info where way timestamp included node updates.
Closes #11575
image
image

### DIFF
--- a/modules/ui/view_on_osm.js
+++ b/modules/ui/view_on_osm.js
@@ -20,10 +20,8 @@ export function uiViewOnOSM(context) {
         const link = selection.selectAll('.view-on-osm')
             .data(data, d => d.id);
 
-        // exit
         link.exit().remove();
 
-        // enter
         const linkEnter = link.enter()
             .append('a')
             .attr('class', 'view-on-osm')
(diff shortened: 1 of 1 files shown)


ChehakTrehan:
@tordans Please review this PR.
Thanks
Chehak

ChehakTrehan:
@tordans I have fixed the build failure.
Please review
Thanks
Chehak

k-yle (commented):
thanks for the PR, but I think we should discuss this first in #11445. There's some more background info in https://github.com/openstreetmap/iD/pull/7629#discussion_r2242279753

PR #11581, opened 2025-11-15
fix: address race conditions in validator (isDataLoaded undefined)

### LINKED ISSUE #11579: address race conditions in tests
[there] are actually intermittent race conditions [in the test suite]

* For [job54955660322](https://github.com/openstreetmap/iD/actions/runs/19226752923/job/54955660322#step:7:409) from 08e4cecb: the issue is caused by us [running deferred work inside `requestIdleCallback`](https://github.com/openstreetmap/iD/blob/develop/modules/core/validator.js#L789), which is called after the test is unmounted.
* For [job54955682430](https://github.com/openstreetmap/iD/actions/runs/19226760186/job/54955682430#step:7:409) from 08e4cecb: we're [running code inside `setTimeout(..., 75)`](https://github.com/openstreetmap/iD/blob/697accfbd32c80ca480ffb28efca68574ab28f84/modules/ui/combobox.js#L137), but the test runner is so fast that the test finishes before the callback is called.

I don't know any easy solutions, but if you want to tackle this problem, that would be a good place to start

_Originally posted by @k-yle in https://github.com/openstreetmap/iD/pull/11569#pullrequestreview-3463865913_

jiyansh007:
This PR fixes the intermittent race-condition test failures mentioned in #11579.

During unmount, deferred work inside `requestIdleCallback` and timed validation
continued running, and `context.connection()` returned a mock missing `isDataLoaded()`.
This caused:

  TypeError: osm.isDataLoaded is not a function

Fixes included:
- Added a safety guard before calling `osm.isDataLoaded`
- Updated test mocks to include `isDataLoaded()`
- Ensured async timers were flushed before teardown

After this change:
- All tests pass consistently
- No unhandled errors remain

### DIFF
--- a/modules/validations/missing_tag.js
+++ b/modules/validations/missing_tag.js
@@ -43,7 +43,7 @@ export function validationMissingTag(context) {
         var subtype;
 
         var osm = context.connection();
-        var isUnloadedNode = entity.type === 'node' && osm && !osm.isDataLoaded(entity.loc);
+        var isUnloadedNode = entity.type === 'node' && osm && typeof osm.isDataLoaded === 'function' && !osm.isDataLoaded(entity.loc);
 
         // we can't know if the node is a vertex if the tile is undownloaded
         if (!isUnloadedNode &&
--- a/package-lock.json
+++ b/package-lock.json
@@ -70,7 +70,7 @@
         "esbuild": "^0.25.12",
         "esbuild-visualizer": "^0.7.0",
         "eslint": "^9.39.1",
-        "fake-indexeddb": "^6.2.4",
+        "fake-indexeddb": "^6.2.5",
         "fetch-mock": "^11.1.1",
         "gaze": "^1.1.3",
         "happen": "^0.3.2",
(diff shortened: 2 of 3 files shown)


modules/validations/missing_tag.js
```diff
@@ -43,7 +43,7 @@ export function validationMissingTag(context) {
         var subtype;
 
         var osm = context.connection();
-        var isUnloadedNode = entity.type === 'node' && osm && !osm.isDataLoaded(entity.loc);
+        var isUnloadedNode = entity.type === 'node' && osm && typeof osm.isDataLoaded === 'function' && !osm.isDataLoaded(entity.loc);
```
k-yle, line 46: I wonder if it would be better to call `context.validator().reset()` in an `afterEach` block. That function calls `cancelIdleCallback`, thereby addressing the root cause. This reset command could possibly be added to [the spec_helpers file](https://github.com/openstreetmap/iD/blob/develop/test/spec_helpers.ts), so that it applies to every test. 

Likewise for combo: we could call `uiCombobox.off()`, which is a global singleton that cancels any pending `setTimeout`s.

---

i tried this locally, but since it's an intermittent issue, i also don't know if it truely solves the problem 😅

k-yle:
oh it looks like this was already fixed in c9318270c4c53595a00d6d08b86dd7ce58775d02

Issue #7964, opened 2020-09-09
Sort relation dropdown by most recently used


quincylvania:
_Broken out from https://github.com/openstreetmap/iD/issues/5997#issuecomment-689086673._

Relations in the "Add relation" dropdown selector are currently sorted by creation date. We could improve this by putting recently used relations at the top, even if they weren't created recently.

quincylvania:
Also, adding many features to one relation will be a lot easier once we do #7753 . But this is still worth doing.

PR #11588, opened 2025-11-17
keep recently edited relations in membership dropdown for entire session

### LINKED ISSUE #3622: Keep relations in relation membership dropdown for entire session
The relation membership dropdown only displays relations within the current viewport. As soon as you pan away from the relation’s existing members, the relation disappears from the dropdown. By contrast, in Potlatch (1 and 2), the list of available relations keeps growing as you pan around. I think iD should adopt Potlatch’s behavior, keeping all loaded relations in the dropdown until leaving iD.

Very often, I need to tag a place POI as a boundary relation’s `label` or `admin_centre`. In iD, this entails zooming out so that some part of the boundary is also visible – generally only possible for villages, because nothing is editable beyond a certain zoom level. Alternatively, I can drag the POI to the boundary and tag the relation membership with both features in view, but then it’s difficult to put the POI in exactly the right place again. With this issue fixed, I could pan to the boundary then pan back to the POI and add the relation membership without relocating the POI. There would be less need for #3487.

k-yle:
Closes #3622

The relation dropdown list now remembers relations that you've already added during this edit session. Even if those relations are no longer visible on screen.

Showing every downloaded relation as originally proposed in #3622 becomes pretty chaotic, I think this is a reasonable compromise which is still quite helpful

### DIFF
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -51,6 +51,8 @@ export function uiSectionRawMembershipEditor(context) {
     var _maxMemberships = 1000;
     /** @type {Set<string>} relations that were added after this panel was opened */
     const recentlyAdded = new Set();
+    /** @type {Set<string>} all relations that had members added during this edit session */
+    const previouslySelected = new Set();
 
     function getSharedParentRelations() {
         var parents = [];
@@ -216,6 +218,7 @@ export function uiSectionRawMembershipEditor(context) {
 
         if (d.relation) {
             recentlyAdded.add(d.relation.id);
+            previouslySelected.add(d.relation.id);
             context.perform(
                 actionAddMembers(d.relation.id, _entityIDs, role),
                 t('operations.add_member.annotation', {
@@ -314,8 +317,22 @@ export function uiSectionRawMembershipEditor(context) {
             });
         } else {
 
+            // any relation that has previously been added will remain
+            // in the dropdown for the rest of the edit session.
+            for (const id of previouslySelected) {
+                const relation = context.hasEntity(id);
+                if (relation && relation.id !== entityID) {
+                    result.push({
+                        relation,
+                        value: baseDisplayValue(relation),
+                        display: baseDisplayLabel(relation)
+                    });
+                }
+            }
+
             context.history().intersects(context.map().extent()).forEach(function(entity) {
                 if (entity.type !== 'relation' || entity.id === entityID) return;
+                if (previouslySelected.has(entity.id)) return; // already added above
 
                 var value = baseDisplayValue(entity);
                 if (q && (value + ' ' + entity.id).toLowerCase().indexOf(q.toLowerCase()) === -1) return;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased (2.38.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* Keep recently edited relations in relation membership dropdown for entire session ([#11588], thanks [@k-yle])
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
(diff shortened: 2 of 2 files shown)


modules/ui/sections/raw_membership_editor.js
```diff
@@ -216,6 +218,7 @@ export function uiSectionRawMembershipEditor(context) {
 
         if (d.relation) {
             recentlyAdded.add(d.relation.id);
+            allAdded.add(d.relation.id);
```
1ec5: This would end up being a negative counter decrementing from zero.

1ec5 (dismissed):

modules/ui/sections/raw_membership_editor.js
k-yle: oh i think the choice of variable names is a bit confusing, _adding_ doesn't mean creating a new relation, but rather _adding_ a member to a relation. 

for example:
* select [`w7634775`](https://pr-11588--ideditor.netlify.app/#disable_features=boundaries&map=20.57/-36.82279/174.80451&background=LINZ_NZ_Aerial_Imagery&id=w7634775)
* click <kbd>+</kbd> in the sidebar to add `w7634775` as a member of any relation
* pan somewhere completely different, and the relation you selected earlier will still be visible

1ec5: Ah, I see, so “selected” is probably a less confusing name internally. I hope users can figure out that that’s why these relations are lingering in the menu.

k-yle: If we think this will be confusing, we could exclude some relations that are always "local", for example [`type=restriction`](https://osm.wiki/Tag:type=restriction).

but as always when we hardcode random tags, it could cause even more confusion because no one can understand the logic

modules/ui/sections/raw_membership_editor.js
```diff
@@ -314,8 +317,22 @@ export function uiSectionRawMembershipEditor(context) {
             });
         } else {
 
+            // any relation that has previously been added will remain
+            // in the dropdown for the rest of the edit session.
```
1ec5, line 321: Would this be more intuitive if we sort some or all of the relations by recency instead of relation ID, as proposed in #7964?

k-yle: yeah I think that would help, but a much more significant improvement would be https://github.com/openstreetmap/id-tagging-schema/pull/1538 - this would filter out all the illogical options. 

e.g. it would stop you from adding a road directly to a superroute, or adding a tree to a boundary relation, etc. etc.

modules/ui/sections/raw_membership_editor.js
```diff
@@ -51,6 +51,8 @@ export function uiSectionRawMembershipEditor(context) {
     var _maxMemberships = 1000;
     /** @type {Set<string>} relations that were added after this panel was opened */
     const recentlyAdded = new Set();
+    /** @type {Set<string>} all relations that had members added during this edit session */
```
matkoniecz, line 54: why only added? maybe also keep ones where members were removed?

matkoniecz: If only added is intentional and desirable - then can you edit PR title?

matkoniecz:
From discussion: idea sounds nice, but there are some potential problems.

What about turn restrictions? If someone adds bunch of no-left-turn with excellent iD turn restrictor editor "helpfully" listing them forever is not going to be useful.

When user splits road with 50 bus relations they do not want to be haunted by this horde of Flixbus routes, probably.

Maybe limit it to some subset of relations?

k-yle:
yeah, I've changed my mind about this idea, after using this PR for several months.

- It's great for `type=route` and `type=boundary` relations (which are inherently large).
- but it's annoying for 'local' relations like `site=traffic_signals` or `type=restriction`, since their bbox is always small.

So I'm happy to close this and revisit it after https://github.com/openstreetmap/id-tagging-schema/pull/1538, when the list of relations will be significantly smaller[^1]

[^1]:  because implausible/invalid options won't be shown anymore

matkoniecz:
Which in turn is blocked by https://github.com/ideditor/schema-builder/issues/134

I guess that if anyone is interested in pushing them forward then commenting there on proposed code may be useful.

matkoniecz:
Now https://github.com/openstreetmap/id-tagging-schema/pull/1538 is the next thing to review/finish to get this one unblocked

PR #11589, opened 2025-11-17
add JSDoc annotations to context.js

k-yle:
Same purpose/rationale as #10618: autocomplete makes it much easier to work in this repository, all it takes is a few lines of comments.

The last few lines are a bit of a hack, but there's no other concise solution[^1]

[^1]: except for migrating this huge file to typescript, but that's pretty infeasible


tyrasd (approved):

tyrasd:
> except for migrating this huge file to typescript, but that's pretty infeasible 

hopefully we can do this one day. :innocent:

PR #11590, opened 2025-11-17
Fix confusing 'Point should be a point' warning

k-yle:
Currently, if you draw a line or area with [`traffic_calming=chicane`](https://osm.wiki/Tag:traffic_calming=chicane), you get this amusing warning:

image

This is because the preset's geometry [only allows `vertex`](https://github.com/openstreetmap/id-tagging-schema/blob/c3903741573b83a3a2b30936398f68a7694b331d/data/presets/traffic_calming/chicane.json#L7-L9), but the validator converts `vertex` to `point` for the translations.

---

Unrelated to #10523 despite looking similar.


matkoniecz (commented):
looks good to me, tested to confirm it works

Issue #10995, opened 2025-04-25
Street level imagery: First forward/backward interaction after switching sequences does not work


danieldegroot2:
### URL

Editing at https://ideditor.netlify.app/#disable_features=boundaries&map=19.73/52.66778/-7.25113&background=EsriWorldImageryClarity&photo_overlay=panoramax

### How to reproduce the issue?

On at least Panoramax, but I believe also other providers, the first time after switching sequences(/providers) I am unable to click forward/backward. Then the second time it works.
When I click forward/backward it always seems to already be done loading.
There are no errors in the browser console.

Sidenote: Could optionally disable buttons while it is still loading but when it takes long / fails to load an image it is actually useful to be able to go to the next/previous image (temporarily) because it fixes the loading issue.

(note, if spamming the back and forth buttons inside the same sequence it sometimes does not work but this is just buffer, which is okay.)

tyrasd:
This seems to work now. Maybe some of the other recent bugfixes fixed this problem as well. Can you please confirm whether this is fixed on https://preview.ideditor.com/ ?

danieldegroot2:
Not currently able to reproduce on either live or preview.

danieldegroot2:
Able to get it again. Button stays greyed out as seen below. Seems zoom/resolution independent.
It either thinks I am dragging (easily true) or it's because the cone has just finished recalculating in the background.
(it only seems to happen where cone has to recalculate but I could be wrong)

just go back and forth and switch between the back/forth button and the two sequences here at different speeds/pattern until you get it.

https://www.openstreetmap.org/edit#map=20/52.6384704/-7.2355404

https://ideditor.netlify.app/#disable_features=boundaries&map=19.00/52.63865/-7.23524&background=EsriWorldImageryClarity&photo_overlay=panoramax&photo=panoramax/3e3322c9-0791-4858-8efe-f8e4c5dbb8fe

![Image](https://github.com/user-attachments/assets/21116f2a-b871-464b-998d-09fad2b6a44f)

Got it two times in a row at https://api.panoramax.xyz/#pic=c209cf96-6ef7-44cd-8679-c044833eec7a&focus=pic

![Image](https://github.com/user-attachments/assets/b75fdc8f-aaf1-488b-8758-bb3a314fe921)

or if I keep using back and forth arrow to go to this image

https://www.openstreetmap.org/edit#map=20/52.6387568/-7.2350603

[...]

PR #11603, opened 2025-11-25
Fixed streetside viewfield

### LINKED ISSUE #11735: fix visibility of the direction icon for Bing streetside
### URL

https://www.openstreetmap.org/edit?editor=id#map=20/45.1828160/5.7292308

1. follow the link above
2. in the map data tab, select Bing streetside
3. on the map, click on one of the 360 view circle : a new small window appears with a Bing streetside view
4. click again on the same 360 view circle : a direction icon appears, which turns if you move the imagery of the streetside to the left or the right.

It should be great to have the direction icon at the first click on a 360 view circle.

note : I will not open another issue, but, in the small window of the street view, if you click on "View on Bing Maps", you will see that it doesn't work anymore. It seems impossible to fix it, the only alternative I found is to use this https://www.bing.com/api/maps/sdk/mapcontrol/isdk/setviewtostreetside, useless I think because less good than the small window of the street view in iD, especially if you enlarge that window.

_No response_

2.37.3

Firefox

JoshM2:
Currently when you have the Bing Streetside photo overlay turned on, the viewfield (yellow cone that shows you where on the map you are looking) doesn't show up unless the image viewer is already open. So the first time you click on one of the circles viewfield doesn't appear.

To fix this issue I moved .datum() to the end of the showViewer() function.

### DIFF
--- a/modules/services/streetside.js
+++ b/modules/services/streetside.js
@@ -667,7 +667,6 @@ export default {
       this._pendingSelected = null;
     }
 
-
     return this;
   },
 
--- a/modules/services/streetside.js
+++ b/modules/services/streetside.js
@@ -700,6 +700,8 @@ export default {
 
     let d = this.cachedImage(key);
 
+    let viewer = context.container().select('.photoviewer');
+    if (!viewer.empty()) viewer.datum(d);
     // Save the selected bubble until viewer is initialized
     this._pendingSelected = d;
 
--- a/modules/services/streetside.js
+++ b/modules/services/streetside.js
@@ -662,6 +662,12 @@ export default {
         .classed('hide', false);
     }
 
+    if (this._pendingSelected) {
+      context.container().select('.photoviewer').datum(this._pendingSelected);
+      this._pendingSelected = null;
+    }
+
+
     return this;
   },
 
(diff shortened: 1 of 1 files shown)


JoshM2:
I was looking through the issues to see if this was in there already and I found #10995 which is related to the problem I was having (step 2 in this comment mentions my problem https://github.com/openstreetmap/iD/issues/10995#issuecomment-2884379921).
The original code I submitted partially fixed this, the first time you clicked the button to move forward the image would change but the marker wouldn't, and on future clicks the marker would lag one circle behind the image you were actually viewing. I put back .datum() where it was originally, so now it is being called at the end showViewer() as well as the original place it was. It works now and both problems are fixed.

matkoniecz:
I was wondering what is so special about Bing that it requires such ugly code and compared it with Mapillary, which is without it.

Then discovered that Mapillary has exactly the same bug.

Not sure should it be done in the same PR, probably it would be preferable.

I also opened #11979 for bug/missing feature that probably should not be crammed into this PR.

-------------------------

I am not great fan of how this code looks like, but I also have no better ideas how to do this.

k-yle (changes requested):
i remember running into this issue with #11666, where i fixed it by [swapping the order of `showViewer()` and `selectImage()`](https://github.com/openstreetmap/iD/pull/11666/changes#diff-d4ce4c283bab22b22158738d4ec92b4f2e7fd6ee77c97616d03dfd83fcbcdcecL277).

I would suggest that as a simpler alternative. [This is the equivilant line](https://github.com/openstreetmap/iD/blob/958c98f452b00cd0541c5d4bea808cf9241aef6d/modules/svg/streetside.js#L104-L106) for Bing. copilot seems to agree

tyrasd:
closing as stale

PR #11619, opened 2025-11-28
Update lasso selection description in core.yaml

### LINKED ISSUE #11726: Merging deletes instead of merging

jidanni:
Sure, if the user could find
> multiselect_lasso: "Another way to select multiple features is to hold down the `{shift}` key, then press and hold down the {leftclick} left mouse button and drag the mouse to draw a selection lasso. All of the points inside the lasso area will be selected.

via a mouseover while looking at the keyboards shortcut list, this change wouldn't be needed.

Else users will bang their head: why can't I select those houses?

(No, a subsequent CTRL+up arrow won't select the way that the nodes are from either. Only works when one node is selected.)

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2420,7 +2420,7 @@ en:
         title: "Selecting features"
         select_one: "Select a single feature"
         select_multi: "Select multiple features"
-        lasso: "Draw a selection lasso around features"
+        lasso: "Lasso select (points only)"
         search: "Find features matching search text"
       with_selected:
         title: "With features selected"
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2420,7 +2420,7 @@ en:
         title: "Selecting features"
         select_one: "Select a single feature"
         select_multi: "Select multiple features"
-        lasso: "Draw a selection lasso to select points with area"
+        lasso: "Draw a selection lasso to select points in an area"
         search: "Find features matching search text"
       with_selected:
         title: "With features selected"
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2420,7 +2420,7 @@ en:
         title: "Selecting features"
         select_one: "Select a single feature"
         select_multi: "Select multiple features"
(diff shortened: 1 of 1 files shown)


data/core.yaml
```diff
@@ -2420,7 +2420,7 @@ en:
         title: "Selecting features"
         select_one: "Select a single feature"
         select_multi: "Select multiple features"
-        lasso: "Draw a selection lasso around features"
+        lasso: "Lasso select (points only)"
```
matkoniecz: maybe lighter modification would fit

```suggestion
        lasso: "Draw a selection lasso to select points with area"
```

? Is it still entirely correct?

The change as is seems to have quite distinct style from other entries

matkoniecz:
@jidanni ping? (note: I unblocked user so that ping will arrive)

jidanni:
Thank you. I finally got the notification.
> "Draw a selection lasso to select points with area"

Better would be
> "Draw a selection lasso to select points with**in** area"

Best would be
> "Draw a selection lasso to select (only) points within area"

The point is to hammer home that the user should only expect points will be selected.

Hmmm, then
> "Draw a selection lasso to select (only!) points within area"

Also "select" appears twice in the same sentence. Not so good.

matkoniecz:
Can you update PR once final text version will be found?

jidanni:
Last time I updated a PR, I crashed the entire GitHub.com . I dare not try again.

matkoniecz:
Than I propose to close it as abandoned PR and treat your future PRs as abandoned.

Also, I have no idea what you mean by "I crashed the entire GitHub.com". It is some misunderstanding.

jidanni:
I found a button to accept your changes and clicked it. I hope you are able to move forward with it now.

PR #11636, opened 2025-12-03
fix non-string properties not visible in the raw tag editor

k-yle:
If you load a gpx/geojson/mvt file via the Custom Map Data pane, only string properties work. Numbers/booleans/arrays/objects don't work.

image

### DIFF
--- a/modules/util/object.js
+++ b/modules/util/object.js
@@ -29,3 +29,28 @@ export function utilCheckTagDictionary(tags, tagDictionary) {
     }
     return undefined;
 }
+
+/**
+ * converts every value in an object to a string, if
+ * it's not already a string.
+ * @param {Record<string, unknown>} object
+ */
+export function stringifyProperties(object) {
+    /** @type {Tags} */
+    const tags = {};
+    for (const key in object) {
+        switch (typeof object[key]) {
+            case 'undefined':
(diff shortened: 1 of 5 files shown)


modules/util/object.js
```diff
@@ -29,3 +29,25 @@ export function utilCheckTagDictionary(tags, tagDictionary) {
...
+    const tags = {};
+    for (const key in object) {
+        switch (typeof object[key]) {
+            case 'undefined':
+                break; // skip property
+            case 'string':
+                tags[key] = object[key];
+                break;
+            default:
+                tags[key] = JSON.stringify(object[key]);
```
1ec5: Any idea if any of the supported formats can contain a `BigInt` value that would cause this call to throw a `TypeError`?

k-yle: good point. currently, none of the 3 parsing libraries[^1] support bigints, but if https://github.com/mapbox/pbf/pull/115 is accepted, that will change. added support now.

[^1]: [@tmcw/togeojson](https://npm.im/@tmcw/togeojson), [@mapbox/vector-tile](https://npm.im/@mapbox/vector-tile),  and [pbf](https://npm.im/pbf)

1ec5 (approved):
Thanks!

PR #11637, opened 2025-12-03
Styling improvements for the custom map data panel

### LINKED ISSUE #9628: missing margins for tags viewer of custom map data
### Description

in data block Custom Map Data not good show data information,
whey are eclipses. But key value have more room.
I think best solution for this is is table view.

### Screenshots

image

k-yle:
Closes #9628

just some padding and removal of the weird `<hr />`

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>image
 <td>image
</table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased (2.38.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* Minor usability improvements for custom map data ([#11635], [#11637], thanks [@k-yle])
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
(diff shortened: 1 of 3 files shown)


tordans:
Line breaks on the values and keys (break all) would be great too.

k-yle:
> Line breaks on the values and keys (break all) would be great too.

I tried this, but unfortunately it's non-straightforward, because the text is rendered in an `<input />` element, and replacing it with a `<textarea />` or `<div />` introduces a lot of complexity. Might be easier to keep this panel consitent with the normal raw tags editor?

tordans:
(For the future:)

> Might be easier to keep this panel consitent with the normal raw tags editor?

Don't see a reason to keep it the same code. I always thought of it as kind of a hack to re-use it given that we have to disable all but the table rendering in terms of features. There a few ideas in the rapid issues on how to build this out more and that would be great for quick mapping tasks that one does not want to create a MR challenge for.

k-yle:
i guess i'll abandon this because of d143a2d6021375b63f26f9394c0a942e8cb4ce20

PR #11643, opened 2025-12-05
Fix: Add missing English translation keys for inspector and modes

garvit-bhattt:
This PR adds missing English translation keys that were referenced in the iD UI but not defined in `core.yaml`.

Added keys include:
- inspector.zoom_to_key
- modes.add_note.key
- modes.inspect.title

These keys previously caused missing translation warnings in the developer console and inconsistent UI labeling. This update improves localization completeness and brings all referenced translation keys into `core.yaml`.

Tested by running:
npm run all
npm start

No missing translation warnings appear now.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -50,7 +50,7 @@ en:
       title: Note
       label: Add Note
       description: "Spotted an issue? Let other mappers know."
-      key: N
+      key: "Add note"
     add_preset:
       title: "Add {feature}"
     browse:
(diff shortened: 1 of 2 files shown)


data/core.yaml
```diff
@@ -703,6 +703,7 @@ en:
     locating: "Locating, please wait..."
     location_unavailable: Your location is unavailable.
   inspector:
+    zoom_to_key: "Zoom to"
```
k-yle, line 706: where did you see `zoom_to_key` being used? 

It should be `zoom_to.key`, which is already defined a few lines below

data/core.yaml
```diff
@@ -50,7 +50,7 @@ en:
       title: Note
       label: Add Note
       description: "Spotted an issue? Let other mappers know."
-      key: N
+      key: "Add note"
```
k-yle, line 53: this change seems to be unrelated to the PR description?

PR #11644, opened 2025-12-05
remove pointless dependency

k-yle:
When constructing a CDN URL for NSI assets, NSI's version number is parsed with "[vparse](https://npm.im/vparse)", and then the semver-patch part is removed.

As far as I can tell, this extra parsing is not required, because [the jsdeliver CDN](https://cdn.jsdelivr.net/npm/name-suggestion-index@~7.0/dist/json/nsi.min.json) supports any semver syntax. None of the other assets do this extra parsing.

cc @bhousel, since this you added this because of https://github.com/openstreetmap/iD/pull/8305#discussion_r663055241

### DIFF
--- a/modules/services/nsi.js
+++ b/modules/services/nsi.js
@@ -1,6 +1,4 @@
 import { Matcher } from 'name-suggestion-index';
-import parseVersion from 'vparse';
-
 import { fileFetcher, locationManager } from '../core';
 import { presetManager } from '../presets';
 
@@ -48,9 +46,7 @@ const notBranches = /(coop|express|wireless|factory|outlet)/i;
 //
 function setNsiSources() {
   const nsiVersion = packageJSON.dependencies['name-suggestion-index'] || packageJSON.devDependencies['name-suggestion-index'];
-  const v = parseVersion(nsiVersion);
-  const vMinor = `${v.major}.${v.minor}`;
-  const cdn = nsiCdnUrl.replace('{version}', vMinor);
+  const cdn = nsiCdnUrl.replace('{version}', nsiVersion);
   const sources = {
     'nsi_data': cdn + 'dist/json/nsi.min.json',
     'nsi_dissolved': cdn + 'dist/wikidata/dissolved.min.json',
--- a/package-lock.json
+++ b/package-lock.json
@@ -95,8 +95,7 @@
         "sinon-chai": "^4.0.1",
         "svg-sprite": "2.0.4",
         "typescript": "^5.9.3",
-        "typescript-eslint": "^8.56.0",
-        "vparse": "~1.1.0"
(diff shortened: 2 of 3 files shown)


bhousel:
Yeah this is fine, good suggestion @k-yle  👍 

I remember adding this in 8db1c1f mostly because Milos asked for it here https://github.com/openstreetmap/iD/pull/8305#discussion_r662213728

I remember being weakly against the idea of having code that automatically references the dependencies in `package.json`, but I didn't really have a strong opinion and didn't want to hold up the PR, so I added it.

My opinion is still that urls like these should be updated manually (by a human) and with intention.

package-lock.json
```diff
@@ -2485,7 +2481,6 @@
       "integrity": "sha512-GNWcUTRBgIRJD5zj+Tq0fKOJ5XZajIiBroOF0yvj2bSU1WvNdYS/dn9UxwsujGW4JX06dnHyjV2y9rRaybH0iQ==",
       "devOptional": true,
       "license": "MIT",
-      "peer": true,
```
matkoniecz: why this `peer` lines are disappearing?

matkoniecz: in the best style of JS ecosystem I found only https://github.com/npm/documentation/issues/1773

looking at https://github.com/npm/cli/pull/8645 and linked PRs it looks like removals are safe? But not entirely getting what is going on.

modules/services/nsi.js
```diff
@@ -48,9 +46,7 @@ const notBranches = /(coop|express|wireless|factory|outlet)/i;
 //
 function setNsiSources() {
   const nsiVersion = packageJSON.dependencies['name-suggestion-index'] || packageJSON.devDependencies['name-suggestion-index'];
-  const v = parseVersion(nsiVersion);
-  const vMinor = `${v.major}.${v.minor}`;
-  const cdn = nsiCdnUrl.replace('{version}', vMinor);
+  const cdn = nsiCdnUrl.replace('{version}', nsiVersion);
```
matkoniecz, line 49: I am trying to confirm that my test plan makes sense: it is executed at runtime, when user starts editing, right?

package-lock.json
k-yle: yeah, i need to update my npm version now that https://github.com/npm/cli/pull/8645 has finally been fixed. I've also removed the diff from this PR

modules/services/nsi.js
k-yle: yes, correct. We can check the browser's devtools to see if the correct version of NSI is loaded:

Untitled

The released version of iD currently uses `@7.0`, this PR will change it to `@~7.0` to exactly match the value in package.json

matkoniecz (commented):
NSI continues to work in my test

PR #11645, opened 2025-12-05
never apply the reverse operation to `railway:turnout_side`

k-yle:
[railway:turnout_side](https://osm.wiki/Key:railway:turnout_side) is relative to the [_facing_ direction](https://en.wikipedia.org/wiki/Facing_and_trailing) of the [switch](https://osm.wiki/Tag:railway=switch), not relative to the [OSM direction](https://osm.wiki/Forward_%26_backward,_left_%26_right).

Therefore, it should not be inverted when flipping the node, or reversing the parent way.


matkoniecz:
I hoped that wiki documentation will leave me less confused but after reading it I am still not sure

I posted https://wiki.openstreetmap.org/wiki/Talk:Tag:railway%3Dswitch#Turnout

claysmalley:
Excuse the sloppy drawing, but this is the idea. A standard switch has a straight setting and a diverging setting. The diverging track may be oriented to the left or the right of the straight track (this is`railway:turnout_side`). The direction of the underlying OSM way has no bearing on which side the turnout is on.

tyrasd:
Thanks. FWIW, there was also a recent PR #11944 to render this tag, so mistakes should be a little more obvious now.

PR #11653, opened 2025-12-08
fix failing CI

### LINKED ISSUE #1848: Staging deployment fails
Right now, the deployments are broken 
https://github.com/openstreetmap/id-tagging-schema/actions/workflows/staging.yml
Error: https://github.com/openstreetmap/id-tagging-schema/actions/runs/19984990119/job/57317572528#step:13:427

k-yle:
Closes https://github.com/openstreetmap/id-tagging-schema/issues/1848

A new langauge was added to transifex recently: [`be@tarask`](https://en.wikipedia.org/wiki/Tara%C5%A1kievica). Due to the `@` symbol, this code isn't a valid [BCP 47 language tag](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag), so [`Intl.Locale`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale) is throwing an error.

### DIFF
--- a/scripts/update_locales.js
+++ b/scripts/update_locales.js
@@ -51,6 +51,13 @@ dataShortcuts.forEach(tab => {
 
 let coverageByLocaleCode = {};
 
+/**
+ * transifex uses underscores and `@` for locale codes. iD only uses
+ * hyphens, so that all locale codes are valid BCP 47 language tags.
+ * @param {string} code
+ */
+const normaliseLocaleCode = (code) => code.replace(/[@_]/g, '-');
+
 // There's a race condition here, but it's highly unlikely that the info will
 // return after the resources. There's an error check just in case.
 asyncMap(resourceIds, getResourceInfo, gotResourceInfo);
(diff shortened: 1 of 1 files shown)


matkoniecz:
note that some codes are rematched in https://github.com/openstreetmap/iD/blob/develop/scripts/language_names.js#L5

no idea whether putting tarask there would suffice

k-yle:
> no idea whether putting tarask there would suffice

yeah, that would also solve the problem, but then the next time a new [language with an `@` symbol](https://explore.transifex.com/languages#:~:text=@) is added, it'll break again.

So i feel like making the code more robust would be a better solution

tyrasd (approved):
Found (and fixed) a small bug: at line 125 we need to use the "raw" language code, because `allStrings` still is indexed by whatever transifex returned). Otherwise the content of the `be-tarask` output is always empty.

In general: I found this language a bit hard to test, as none of my usual browsers let me set `be-tarask` as my preferred language.[^1] But it did work as expected in the openstreetmap-website integration when selecting `be-tarask` from the language selector there.

[^1]: in Firefox, I was able to manually override the `intl.accept_languages` setting via `about:config`

PR #11656, opened 2025-12-08
11655: show warning if invalid ref:REI input tag is found

### LINKED ISSUE #2055: Add warning for ref:REI=* values
### Description

`ref:REI` is a tag used in Italy for hiking relations, it has 13358 uses at the moment. As [wiki](https://wiki.openstreetmap.org/wiki/IT:Key:ref:REI), the value has to be of 8 characters, but sometimes people adds wrong values, with less or more characters. Would be possible for iD to warn the users about this problem?

At the moment there's a user doing gardening from time to time, but it would be cool if we could prevent this problem at the source, and use gardening time for other contributions. Thanks!

### Screenshots

_No response_

ankushKun:
Implements openstreetmap/id-tagging-schema#2055 

Added REI tag validation logic to modules/validations/invalid_format.js
Added en localization for the same to data/core.yaml

When an invalid value is entered to a ref:REI tag, a warning is displayed to the user

Valid formats: A string of length 8 characters `00000000` or in the format `00-00-0000`

image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1880,6 +1880,9 @@ en:
         message: '{feature} has an invalid URL'
         message_multi: '{feature} has multiple invalid URLs'
         reference: 'URLs must start with "http://" or "https://".'
+      ref_rei:
+        message: '{feature} has an invalid ref:REI value'
+        reference: 'ref:REI values must be 8 characters long.'
     line_as_area:
       message: '{feature} should be a line, not an area'
     line_as_point:
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -27,6 +27,15 @@ export function validationFormatting() {
                 .call(t.append('issues.invalid_format.email.reference'));
         }
 
+        function showReferenceRefRei(selection) {
+            selection.selectAll('.issue-reference')
+                .data([0])
+                .enter()
+                .append('div')
+                .attr('class', 'issue-reference')
+                .call(t.append('issues.invalid_format.ref_rei.reference'));
+        }
+
         function isValidURL(url, strict = false) {
             try {
                 // First try strict WHATWG parsing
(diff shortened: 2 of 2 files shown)


tordans:
@ankushKun please review the codebase on how alerts like this are handled. IMO you will find that we do not add them on a tag by tag basis manually deep in the code but instead search for abstractions like id-tagging-schema and similar.

ankushKun:
@tordans should I be adding the ref:REI tag to https://github.com/openstreetmap/id-tagging-schema first? since I cannot find it anywhere in the current codebases

matkoniecz:
@tordans Can tagging schema support such validations? Do we have any such formats specified there?

ivanbranco:
`Valid formats: A string of length 8 characters 00000000 or in the format 00-00-0000`

The REI reference doesn't include dashes, so the only valid format should be an 8 character string. More precise validations could be added (e.g. the first three characters are always letters), but even a simple length check would already be useful.

k-yle (commented):
rather than hardcoding random tags, it would be better to use [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema)'s existing [`pattern` field](https://github.com/ideditor/schema-builder#pattern), if at all.

This requires the key [`ref:REI`](https://osm.wiki/Key:ref:REI) to be defined in the tagging schema first. I think this is a good thing, because:
1. it makes us go through the usual process of determing whether [`ref:REI`](https://osm.wiki/Key:ref:REI) deserves a field in the UI
2. iD's validation issues should be fixable. A warning about a tag that's hidden in the UI is likely to be confusing for new users.

It's also worth considering if this warning is actually a good idea. What are mappers supposed to do if the value is invalid? Perhaps this would be better suited for [osmose](https://github.com/osmose-qa/osmose-backend)?

matkoniecz:
> It's also worth considering if this warning is actually a good idea. What are mappers supposed to do if the value is invalid? Perhaps this would be better suited for [osmose](https://github.com/osmose-qa/osmose-backend)?

or maybe create (limited) number of notes where format mismatches expected one? (only in places without nearby notes)

this way it is more likely to be looked by expert mappers actively looking for problems to fix

Marc-marc-marc:
> It's also worth considering if this warning is actually a good idea. What are mappers supposed to do if the value is invalid? 

delete the incorrect ref or fill the correct one

> Perhaps this would be better suited for [osmose](https://github.com/osmose-qa/osmose-backend)?

I think that a warning during mapping is very better than uploading a invalid data, wating a QA to detect it and/or open a not-for-qa-reporting (bad pratice ?) and wait that someone (maybe not the initial mapper) find why the 1st mapper fill an invalid data and where the correct one is

tyrasd:
see https://github.com/openstreetmap/iD/pull/11656#pullrequestreview-3626412076

PR #11657, opened 2025-12-09
Remove icon color override

hlfan:
Without the override, the icon automatically inherits the correct fill color.
And the text color also works in dark mode, see this before/after:
image

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1447,7 +1447,6 @@ button.preset-reset .label.flash-bg {
 
 .label-text .label-textannotation svg.icon {
     margin: 0 8px;
-    color: #333;
     opacity: 0.5;
     width: 14px;
     height: 14px;


tyrasd (approved):

matkoniecz (commented):
I tested that it fixes wikidata lock icon

is it expected to have also other effects?

hlfan:
I didn't find any other affected elements that would be affected, only this:

https://github.com/openstreetmap/iD/blob/0c8499db0dc2e45e4b64d24ef03ee75ab9ad93ee/modules/ui/field.js#L259-L261

tyrasd:
Not sure why I did not merge the PR after approving it. Sorry for the delay. :bow:

matkoniecz:
Also, thanks to @hlfan for this fix!

PR #11661, opened 2025-12-10
use JSON instead of XML for (almost) all OSM APIs

k-yle:
We can remove almost all the XML parsing code by using the OSM API's json endpoints. Deleted about 250 lines of code.

We already use the new JSON endpoints for most things. This PR migrates the last few endspoints:
- GET `/api/capabilities.json`
- GET `/api/0.6/notes.json`
- GET `/api/0.6/notes/{id}.json`
- POST `/api/0.6/notes/{id}/{action}.json`
- GET `/api/0.6/changesets.json`

This is faster, simpler, and makes the code more concise.

The last remaining endpoint that doesn't support JSON is _uploading changesets_, which is tracked by https://github.com/openstreetmap/openstreetmap-website/pull/5973

<details>
<summary>affected parts of the UI:</summary>

- in the changeset upload panel: the `comment` dropdown
- view notes + view note comments
- create note
- close note
- reopen note
- comment on note
- search for note by ID
- (imageryBlocklist)
- (max nodes per way)
- API online/offline status

</details>


matkoniecz (commented):
I retested following:
```
in the changeset upload panel: the comment dropdown
view notes + view note comments
create note
close note
reopen note
comment on note
search for note by ID
```

How following may be tested?

```
(imageryBlocklist)
(max nodes per way)
API online/offline status
```
?

I have not found any other new affected places.

matkoniecz:
BTW, the diff is beautiful. So much of code is now gone!

modules/services/osm.js
```diff
@@ -20,7 +20,7 @@ var dispatch = d3_dispatch('apiStatusChange', 'authLoading', 'authDone', 'change
 var urlroot = osmApiConnections[0].url;
 var apiUrlroot = osmApiConnections[0].apiUrl || urlroot;
 var redirectPath = window.location.origin + window.location.pathname;
-var oauth = osmAuth({
+var oauth = new osmAuth({
```
tyrasd: Was there a reason to change this to use the explicit constructor syntax? I guess it does work either way, but the [upstream example](https://github.com/osmlab/osm-auth?tab=readme-ov-file#example) is using this without `new`, so maybe sticking to that would be fine? 🤔

tyrasd:
> How following may be tested?
> 
> ```
> (imageryBlocklist)
> (max nodes per way)
> API online/offline status
> ```

I tested the more "rare" replies for the capabilities call (e.g. API online/offiline) on a local test instance of the OSM API, and it works as expected. :+1:

modules/services/osm.js
k-yle: oh you're right, the library [declares that `osmAuth` is an ES6 class](https://github.com/osmlab/osm-auth/blob/f2bc3ec6a41603a48a97c69c09d92f78e1702ed7/src/osm-auth.d.ts#L4), but acually it's just [a function](https://github.com/osmlab/osm-auth/blob/f2bc3ec6a41603a48a97c69c09d92f78e1702ed7/src/osm-auth.mjs#L21). 

I guess we should fix it upstream them, so I'll revert this change

---

_edit:_ opened https://github.com/osmlab/osm-auth/pull/147

PR #11666, opened 2025-12-12
add wikimedia commons as an photo service

### LINKED ISSUE #10640: Implement Wikimedia Commons as an image provider
### Description

I would like to have an implementation of wikimedia commons images with coordinates, there should be an overlay for that.

Basically the same overlay as for Mapillary, but for Wikimedia Commons (so you can use that for mapping). Ideally only show images under CC0 license (or else using the images is not allowed (i think?)).

### Screenshots

_No response_

k-yle:
Closes #10640

images from [Wikimedia Commons](https://commons.wikimedia.org) are now available alongside streetlevel imagery like Mapillary:

image

This uses Mediawiki's [Geosearch](https://www.mediawiki.org/wiki/API:Geosearch)+[Imageinfo](https://www.mediawiki.org/wiki/API:Imageinfo) API ([example](https://commons.wikimedia.org/w/api.php?format=json&formatversion=2&origin=*&action=query&generator=geosearch&ggsnamespace=6&ggslimit=500&ggsbbox=-36.835668247244385%7C174.76638793945315%7C-36.83676736421915%7C174.76776123046878&prop=coordinates%7Cimageinfo&iiprop=timestamp%7Cuser%7Curl%7Cextmetadata%7Cmetadata%7Csize&iiextmetadatafilter=LicenseShortName&iiurlwidth=300)), which was discussed in https://github.com/openstreetmap/iD/issues/10640#issuecomment-3137220974

The PR is a draft because it's blocked by several other PRs, and there are a lot of issues to consider:
- [ ] the API only provides basic metadata and EXIF data, but not [structured data](https://commons.wikimedia.org/wiki/Commons:Structured_data). 
  - [ ] therefore, the _date_ is the uploaded date, not the capture date. which could be years apart.
  - [ ] therefore, the _username_ is the uploader, not the author. which is often different (e.g. [flickr2commons](https://commons.wikimedia.org/wiki/Commons:Flickr2Commons)) 
  - [ ] no support for heading/bearing yet, although this could be obtained from the EXIF (`GPSImgDirection` & co.) if we decide not to use [structured data](https://commons.wikimedia.org/wiki/Commons:Structured_data).
- [ ] there's no filter to exclude incompatible licenses yet. I'm guessing CC-0 is the only compatible license?
  - [ ] The license information comes from [`LicenseShortName`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data). There is a also a boolean called [`AttributionRequired`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data), maybe we could filter based on `AttributionRequired=false` instead of hardcoding compatible license names ?
- [ ] are we adhering to [wikimedia's API terms](https://foundation.wikimedia.org/wiki/Policy:Wikimedia_Foundation_API_Usage_Guidelines)? I think so, but maybe we should tweak the tiling logic to further reduce the number of API requests
  - [ ] _share a kind ping to the [Wikimedia Foundation "SRE" Team](https://www.mediawiki.org/wiki/Wikimedia_Site_Reliability_Engineering), to bring this feature to their attention - as it could have an impact on API traffic_ - see [#11666 (comment 3661238897)](https://github.com/openstreetmap/iD/pull/11666#issuecomment-3661238897)

### DIFF
--- a/modules/services/commons.ts
+++ b/modules/services/commons.ts
@@ -1,0 +1,357 @@
+import { dispatch as d3_dispatch } from 'd3-dispatch';
+import RBush, { type BBox } from 'rbush';
+import type { Tile } from '../util/tiler';
+import { utilTiler, utilQsString, utilStringQs } from '../util';
+import { planePhotoFrame } from './plane_photo';
+import { services } from '.';
+import type { Projection } from '../geo/raw_mercator';
+import { searchLimited, type WithBbox } from '../util/partition';
+import type { Vec2 } from '../geo/vector';
+import { getRelativeDate } from '../util/date';
+import { uiTooltip } from '../ui';
+import { localeDateString } from '../renderer/background_source';
+
+export interface CommonsPhoto {
+    pageid: number;
+    ns: number;
+    title: string;
+    index: number;
+    coordinates: {
+        lat: number;
+        lon: number;
+        primary: '';
+        globe: 'earth';
+    }[];
+    imagerepository: 'local';
+    imageinfo: {
+        /** ISO Date */
+        timestamp: string;
+        user: string;
+        size: number;
+        width: number;
+        height: number;
+        thumburl: string;
+        thumbwidth: number;
+        thumbheight: number;
+        responsiveUrls: { [size: number]: string; };
+        url: string;
+        descriptionurl: string;
+        descriptionshorturl: string;
+        metadata: { name: string; value: unknown }[];
+        extmetadata: {
+            LicenseShortName: {
+                value: 'Public domain';
+                source: 'commons-desc-page';
+                hidden: '';
+            };
+        };
+    }[];
+}
+
+interface ApiResponse {
+    error?: {
+        code: string;
+        info: string;
+        docref: string;
+    };
+    batchcomplete?: boolean;
+    continue?: {
+        iicontinue: string;
+        cocontinue: string;
+        continue: string;
+    };
+    query?: {
+        pages: {
+            [pageId: string]: CommonsPhoto;
+        };
+    };
+}
+
+export function getCommonsPhotoLoc(photo: CommonsPhoto): Vec2 {
+    return [photo.coordinates[0].lon, photo.coordinates[0].lat];
+}
+export function getCommonsPhotoDate(photo: CommonsPhoto) {
+    return new Date(photo.imageinfo[0].timestamp);
+}
+
+
+const minZoom = 10; // TODO: reconsider
+const maxZoom = 20;
+
+interface Cache {
+    images: {
+        rtree: RBush<WithBbox<CommonsPhoto>>;
+        byPageId: { [pageId: string]: CommonsPhoto };
+    };
+    requests: {
+        loaded: { [tileId: string]: true };
+        inflight: { [tileId: string]: AbortController };
+    };
+};
+
+function createCache(): Cache {
+    return {
+        images: { rtree: new RBush(), byPageId: {} },
+        requests: { loaded: {}, inflight: {} }
+    };
+}
+let _cache = createCache();
+
+
+
+export default new class CommonsService {
+    dispatch = d3_dispatch('loadedImages', 'viewerChanged');
+    // @ts-expect-error -- hack
+    on = (...args) => this.dispatch.on(...args);
+
+    #photoViewer: ReturnType<typeof planePhotoFrame> | undefined;
+
+    activeImage: CommonsPhoto | undefined;
+
+    #createRequetUrl(bbox: BBox) {
+        // docs: https://www.mediawiki.org/wiki/Extension:GeoData#API
+        const qs = new URLSearchParams({
+            format: 'json',
+            formatversion: '2',
+            origin: '*', // for CORS
+            action: 'query',
+            generator: 'geosearch',
+            ggsnamespace: '6', // the `File:` namespace
+            ggslimit: '500',
+            ggsbbox: `${bbox.maxY}|${bbox.minX}|${bbox.minY}|${bbox.maxX}`,
+            prop: 'coordinates|imageinfo',
+            iiprop: 'extmetadata|metadata|size|timestamp|url|user',
+            iiextmetadatafilter: 'LicenseShortName',
+            iiurlwidth: '300',
+        });
+        return `https://commons.wikimedia.org/w/api.php?${qs}`;
+    }
+
+    loadTiles(projection: Projection) {
+        const tiler = utilTiler()
+            .zoomExtent([minZoom, maxZoom])
+            .skipNullIsland(true);
+
+        const tiles = tiler.getTiles(projection);
+        if (!tiles) return;
+        for (const tile of tiles) {
+            this.#loadTile(tile);
+        }
+    }
+
+    cachedImage(pageId: string) {
+        return _cache.images.byPageId[pageId];
+    }
+
+    async #loadTile(tile: Tile) {
+        const cache = _cache.requests;
+        const tileId = tile.id;
+
+        if (cache.loaded[tileId] || cache.inflight[tileId]) return;
+        const controller = new AbortController();
+        cache.inflight[tileId] = controller;
+
+        // @ts-expect-error -- blocked by other PR
+        const url = this.#createRequetUrl(tile.extent.bbox());
+
+        try {
+            const response = await fetch(url, { signal: controller.signal });
+            if (!response.ok) {
+                throw new Error(response.status + ' ' + response.statusText);
+            }
+            const data: ApiResponse = await response.json();
+
+            if (data.error) {
+                throw new Error(`${data.error.code}: ${data.error.info}`);
+            }
+
+            const newPhotos = Object
+                .values(data.query?.pages || {})
+                .filter(photo => photo.coordinates && photo.imageinfo); // TODO: temp hack, why are some missing coords?
+
+            // load the API response into the cache
+            for (const photo of newPhotos) {
+                _cache.images.byPageId[photo.pageid] = photo;
+            }
+            _cache.images.rtree.load(Object.values(newPhotos).map(photo => {
+                const [x, y] = getCommonsPhotoLoc(photo);
+                return { minX: x, maxX: x, minY: y, maxY: y, data: photo };
+            }));
+
+            this.dispatch.call('loadedImages');
+        } catch (ex) {
+            console.error(ex); // eslint-disable-line no-console
+        }
+        cache.loaded[tileId] = true;
+        delete cache.inflight[tileId];
+    }
+
+    reset() {
+        Object.values(_cache.requests.inflight).forEach(request => request.abort());
+        _cache = createCache();
+    };
+
+    /** get visible images from the cache */
+    getImages(projection: Projection): CommonsPhoto[] {
+        return searchLimited(5, projection, _cache.images.rtree).map(item => item.data);
+    }
+
+    getCachedImage(pageId: string) {
+        return _cache.images.byPageId[pageId];
+    }
+
+    /** update the currently highlighted bubble */
+    setStyles(context: iD.Context, hovered: CommonsPhoto | undefined) {
+        context.container()
+            .selectAll<HTMLElement, CommonsPhoto>('.layer-commons .viewfield-group')
+            .classed('highlighted', (d) => d.pageid === hovered?.pageid)
+            .classed('hovered', (d) => d.pageid === hovered?.pageid)
+            .classed('currentView', (d) => d.pageid === this.activeImage?.pageid);
+
+        context.container().selectAll('.layer-commons .viewfield-group .viewfield')
+            .attr('d', 'M 8,13 m -10,0 a 10,10 0 1,0 20,0 a 10,10 0 1,0 -20,0');
+    }
+
+    /** Updates the URL to save the current shown image */
+    #updateUrlImage(pageid: number | undefined) {
+        const hash = utilStringQs(window.location.hash);
+        if (pageid) {
+            hash.photo = `commons/${pageid}`;
+        } else {
+            delete hash.photo;
+        }
+        window.history.replaceState(null, '', `#${utilQsString(hash, true)}`);
+    }
+
+    /** Loads the selected image in the frame */
+    selectImage(context: iD.Context, id: number | string) {
+        const d = _cache.images.byPageId[id];
+
+        this.activeImage = d;
+        this.#updateUrlImage(d.pageid);
+
+        const viewerLink = d.imageinfo[0].url;
+
+        let viewer = context.container().select('.photoviewer');
+        if (!viewer.empty()) viewer.datum(d);
+
+        this.setStyles(context, undefined); // TODO: why?
+
+        let wrap = context.container()
+            .select('.photoviewer .commons-wrapper');
+
+
+        this.#photoViewer!
+            .showPhotoFrame(wrap)
+            .selectPhoto({ image_path: viewerLink });
+
+        let attribution = wrap.selectAll('.photo-attribution').text('');
+
+        attribution
+            .append('div')
+            .attr('class', 'attribution-row');
+
+        attribution
+            .append('a')
+            .attr('class', 'report-photo')
+            .attr('href', d.imageinfo[0].descriptionurl)
+            .attr('target', '_blank')
+            .attr('rel', 'noopener')
+            .text(d.imageinfo[0].extmetadata.LicenseShortName.value);
+
+        attribution
+            .append('span')
+            .text('|');
+
+        attribution
+            .append('span')
+            .text(getRelativeDate(new Date(d.imageinfo[0].timestamp)))
+            .call(uiTooltip()
+                .title(() => localeDateString(d.imageinfo[0].timestamp))
+                .placement('top'));
+
+        attribution
+            .append('span')
+            .text('|');
+
+        attribution
+            .append('a')
+            .attr('class', 'image-link')
+            .attr('target', '_blank')
+            .attr('rel', 'noopener')
+            .attr('href', `https://commons.wikimedia.org/wiki/User:${d.imageinfo[0].user}`)
+            .text(`User:${d.imageinfo[0].user}`);
+
+        return this;
+    }
+
+    showViewer(context: iD.Context) {
+        const wrap = context.container().select('.photoviewer');
+        const isHidden = wrap.selectAll('.photo-wrapper.commons-wrapper.hide').size();
+
+        if (isHidden) {
+            // hide the photo viewers from any other service
+            for (const service of Object.values(services)) {
+                if (service === this) continue;
+                if ('hideViewer' in service) {
+                    service.hideViewer(context);
+                }
+            }
+            wrap
+                .classed('hide', false)
+                .selectAll('.photo-wrapper.commons-wrapper')
+                .classed('hide', false);
+        }
+
+        return this;
+    }
+
+
+    #isLoaded = false;
+    // eslint-disable-next-line require-await
+    async ensureViewerLoaded(context: iD.Context) {
+
+        let imgWrap = context.container()
+            .select('#ideditor-viewer-commons-simple > img');
+
+        if (!imgWrap.empty()) imgWrap.remove();
+
+        if (this.#isLoaded) return;
+        this.#isLoaded = true;
+
+        let wrap = context.container()
+            .select('.photoviewer')
+            .selectAll('.commons-wrapper')
+            .data([0]);
+
+        let wrapEnter = wrap.enter()
+            .append('div')
+            .attr('class', 'photo-wrapper commons-wrapper')
+            .classed('hide', true)
+            .on('dblclick.zoom', null);
+
+        wrapEnter
+            .append('div')
+            .attr('class', 'photo-attribution fillD');
+
+        this.#photoViewer = planePhotoFrame(context, wrapEnter);
+        this.#photoViewer.event.on('viewerChanged', () => this.dispatch.call('viewerChanged'));
+    }
+
+    /** Hides the current viewer if shown, resets the active image and sequence */
+    hideViewer(context: iD.Context) {
+        let viewer = context.container().select('.photoviewer');
+        if (!viewer.empty()) viewer.datum(null);
+        this.#updateUrlImage(undefined);
+        viewer
+            .classed('hide', true)
+            .selectAll('.photo-wrapper')
+            .classed('hide', true);
+        context.container().selectAll('.viewfield-group, .sequence, .icon-sign')
+            .classed('currentView', false);
+
+        this.activeImage = undefined;
+
+        this.setStyles(context, undefined);
+    }
+};
--- a/modules/svg/local_photos.js
+++ b/modules/svg/local_photos.js
@@ -87,7 +87,7 @@ export function svgLocalPhotos(projection, context, dispatch) {
             .on('click.forward', () => stepPhotos(1))
             .text('▶');
 
-        return planePhotoFrame(context, viewerEnter)
+        return Promise.resolve(planePhotoFrame(context, viewerEnter))
             .then(planePhotoFrame => _photoFrame = planePhotoFrame);
     }
 
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -350,6 +350,20 @@ li.list-item-photos.list-item-mapilio.active:after {
     overflow: hidden;
 }
 
+/* commons Image Layer */
+li.list-item-photos.list-item-commons.active:after {
+    background-color: #069;
+}
+.layer-commons {
+    pointer-events: none;
+}
+.layer-commons .viewfield-group * {
+    fill: #069;
+    stroke: #ffffff;
+    stroke-opacity: .6;
+    fill-opacity: .6;
+}
+
 /* panoramax Image Layer */
 li.list-item-photos.list-item-panoramax.active:after {
     background-color: #ff6f00;
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1381,6 +1381,9 @@ en:
     tooltip: "Street-level photos from the Norwegian Public Roads Administration"
     publisher: "Norwegian Public Roads Administration"
     view_on: "View it on Vegbilder"
+  commons:
+    title: Wikimedia Commons
+    tooltip: Georeferenced photos with a compatible license from Wikimedia Commons
   mapillary_images:
     tooltip: "Street-level photos from Mapillary"
   mapillary_map_features:
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -382,6 +382,7 @@ export function coreContext() {
 
 
   /* Photos */
+  /** @type {ReturnType<typeof rendererPhotos>} */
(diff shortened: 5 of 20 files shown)


modules/services/commons.ts
```diff
@@ -0,0 +1,357 @@
...
+            format: 'json',
+            formatversion: '2',
+            origin: '*', // for CORS
+            action: 'query',
+            generator: 'geosearch',
+            ggsnamespace: '6', // File: namespace
+            ggslimit: '500',
+            ggsbbox: `${bbox.maxY}|${bbox.minX}|${bbox.minY}|${bbox.maxX}`,
+            prop: 'coordinates|imageinfo',
+            iiprop: 'timestamp|user|url|extmetadata|metadata|size',
```
valerio-bozzolan: Better to have these values listed in alphabetical order, for MediaWiki API netiquette (I don't remember where it's written). So:

```
extmetadata|metadata|size|timestamp|url|user
```

modules/svg/local_photos.js
```diff
@@ -87,7 +87,7 @@ export function svgLocalPhotos(projection, context, dispatch) {
             .on('click.forward', () => stepPhotos(1))
             .text('▶');
 
-        return planePhotoFrame(context, viewerEnter)
+        return Promise.resolve(planePhotoFrame(context, viewerEnter))
```
valerio-bozzolan, line 90: I see that the old line 56 is also auto-resolving, but I would expect to resolve externally. Example call:

```
ensureViewerLoaded().then(...)
```

What do you think about?

(This is not a blocking comment)

valerio-bozzolan:
> there's no filter to exclude incompatible licenses yet. I'm guessing CC-0 is the only compatible license

Uh?

My two cents:

- :white_check_mark: one thing is the image authorization (image license allowing to embed that image for all purposes)
    - :white_check_mark: All Wikimedia Commons pictures are allowed to be embedded in this web application for all purposes, as long as there is a link to Wikimedia Commons to see the image credits
    - :white_check_mark: Some Wikimedia Commons pictures require to cite the author(s) in a specific way, but a link to Wikimedia Commons is sufficient from the web app; the credits are necessary only if we share the image itself, or if you share modified versions of the image, so, not something about the OSM database itself
    - :white_check_mark: This pull request is NOT allowing people to create modified versions of the image... or put the image in the OSM database...
    - :white_check_mark: We are **not** allowing people to draw maps over images... we are not Bing Maps...
    - :white_check_mark: So, the web app does not create "derivative work". Really. I just look at a picture and I get the missing local knowledge. If that was a concern, it's a no-problem.
- :white_check_mark: one thing is the related edit (OpenStreetMap license)
    - :white_check_mark: Users can freely map as usual, whatever they looked before as source of information.
    - :white_check_mark: There is NO competition between Wikimedia Commons and OpenStreetMap (unlike Google StreetView - I guess you are thinking about that)

So I would say: what you are doing, is perfectly fine, and any photographer on Wikimedia Commons would thank you so much.

If an image should not be on Commons, it must be removed from Commons. But that's another issue.

valerio-bozzolan (commented):
https://pr-11666--ideditor.netlify.app/#disable_features=boundaries&map=20.15/-33.79739/151.28750&background=MAPNIK&photo_overlay=commons&photo=commons/54837769

:orange_circle: I tried your above demo ↑ and I noticed it does not load the image during the first page load.

I can reproduce by:

- open link in an anonymous tab

I can reproduce in Firefox 140ESR, or Chromium, from Debian GNU/Linux stable (trixie).

Workaround:

- click another image

k-yle:
> > there's no filter to exclude incompatible licenses yet. I'm guessing CC-0 is the only compatible license
>
> My two cents:
> [...]

unfortunately it might be more complicated than that 🫤  @1ec5 wrote a good summary [in this forum thread](https://c.osm.org/t/123973/4). That whole discussion is quite relevant. 

I think it would be safe if we filter based on [`AttributionRequired=false`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data), but perhaps it's worth asking the LWG for some advice, if no one is certain?

modules/services/commons.ts
k-yle: sure, done

modules/svg/local_photos.js
k-yle: oh, that's because I changed `planePhotoFrame`. it's no longer an async function, and this was the least intrusive fix, to avoid changing indentation.

i would consider this PR to be very much a draft. this part might get reverted layer, once the conflicting PRs are merged

valerio-bozzolan:
> unfortunately it might be more complicated than that 🫤 @1ec5 wrote a good summary [in this forum thread](https://c.osm.org/t/123973/4). That whole discussion is quite relevant.

Thanks for the link! Comments:

- about 1st blog post - https://community.openstreetmap.org/t/linking-cc-by-sa-4-0-images-via-wikimedia-commons-or-other-urls/123973/4:
   - it describes an interesting problem about the `image=` tag
   - this proposed code change does not endorse such `image=` tag, so, no-problem
   - examining the worst use case: if somebody decides to use this code proposal to introduce big mistakes, like, adding a `image=` tag, hot-linking JPG images from Wikimedia Commons, instead of just setting the Wikimedia Commons file page... that could be a problem, but:
       - People should just avoid to hot-link stuff. If the contributor hot-link a file instead of linking the credits page of Wikimedia Commons, that's a contribution error, in general, in my opinion. This code proposal does not encourage hot-linking.
           - If we think we are endorsing hot-linking, we can clarify this even more with a small phrase under every image (I mean, every image, not just Wikimedia Commons), but this has nothing to do with image licenses, or with this specific change.
       - **Also public domain files should not be hot-linked**. For multiple reasons, but mainly because of "moral rights" exist. So, when possible, everybody should try to set a minimum credit to multimedia files, even on public domain files. Example: the Monna Lisa is under the public domain, but we should still add a small credit to say that the author is Leonardo Da Vinci and not OsmAnd. This is a moral right, not part of copyright, so, it always applies. The solution in Wikimedia Commons is: not hot-linking the JPG, but linking the Wikimedia Commons page, as credit. So everyone reaches both the file, as well as the most-updated credits.
       - So this is not a problem created by this code feature.
- about 2nd blog post by Minh Nguyễn - https://community.openstreetmap.org/t/linking-cc-by-sa-4-0-images-via-wikimedia-commons-or-other-urls/123973/2
    - totally agree
- about the 4th blog post by Minh Nguyễn - https://community.openstreetmap.org/t/linking-cc-by-sa-4-0-images-via-wikimedia-commons-or-other-urls/123973/4
    - I do **not** agree.
    - I read worrying statements like «*If the photo shows the building’s wall because the hot air balloon hasn’t taken off yet, that’s still copying*»
    - My sincere feedback is that similar statements are not just wrong, but dramatic and destructive. Example: if I've taken a photo of a bridge, and if a mapper maps such bridge, I cannot accuse that mapper of copyright violation. I maybe need to invest millions in very good lawyers to convince a court that this extracting facts from images is a copyright violation. The reason is: **facts are not covered by copyright**. Accusing a mapper of "copying" facts from an image is just a boomerang accuse. Try to accuse somebody in such way, and a [justice of the peace](https://en.wikipedia.org/wiki/Justice_of_the_peace) would consider the accuser as a [copyright troll](https://en.wikipedia.org/wiki/Copyright_troll), or worst. I hope this reassures that user of the forum.
    - If what I'm saying is not enough: I can recommend to find some free time to book a [Creative Commons certification](https://creativecommons.org/about/cc-certificate-program/), since these are FAQs answered in this kind of professional courses
        - P.S. and incidentally I have such certification - "Creative Commons - Creative Commons Certificate for GLAM - 2021-12-21.png" - "Valerio Bozzolan" - that is why I recommend it - P.S. I don't earn any fee from recommending these things.

(P.S. will continue there)

valerio-bozzolan:
P.S. maybe a good idea to add a checkbox like this in the description, as last point:

- [ ] share a kind ping to the Wikimedia Foundation "SRE" Team, to bring this feature to their attention - as it could have an impact on API traffic - https://www.mediawiki.org/wiki/Wikimedia_Site_Reliability_Engineering

1ec5:
> My sincere feedback is that similar statements are not just wrong, but dramatic and destructive. Example: if I've taken a photo of a bridge, and if a mapper maps such bridge, I cannot accuse that mapper of copyright violation. I maybe need to invest millions in very good lawyers to convince a court that this extracting facts from images is a copyright violation. The reason is: facts are not covered by copyright. Accusing a mapper of "copying" facts from an image is just a boomerang accuse. Try to accuse somebody in such way, and a [justice of the peace](https://en.wikipedia.org/wiki/Justice_of_the_peace) would consider the accuser as a [copyright troll](https://en.wikipedia.org/wiki/Copyright_troll), or worst. I hope this reassures that user of the forum.

I was trying to explain the reasoning behind OSM’s “[whiter than white](https://wiki.openstreetmap.org/wiki/Whiter_than_white)” policy on contributions, which is a policy that the project is within its rights to choose. I was not making any legal accusations [from my armchair](https://community.openstreetmap.org/t/dont-copy-from-other-maps-rule/134810/9). Besides, there is a distinction between the act of copying and the act of violating a copyright. I was trying not to conflate the two concepts, but I’m afraid some participants on the forum have jumped to conclusions based on selective quotes of what I wrote. I have gone on the record [many](https://community.openstreetmap.org/t/aligning-osm-hiking-sections-with-wikivoyage/129850/20), [many](https://community.openstreetmap.org/t/poi-quality-and-usage-any-corporates-contributing/107488/10) [times](https://community.openstreetmap.org/t/mapping-temporary-road-closure-based-on-government-announcement-copyright-and-data-reliability-issues/136704/6) reminding folks that facts cannot be copyrighted. On that basis, I have successfully pushed OpenHistoricalMap to make more liberal use of sources than OSM. You would not find someone more eager for a pragmatic way forward for a feature such as this.

I’ll take that as feedback to be more concise next time. 😅

valerio-bozzolan:
(Imagine me on an armchair too lol)

Thanks. At this point I do not understand anymore _who_ is the remaining stakeholder sharing legal concerns against this feature, and - after our solid answers (yours included) about extracting facts from photos - _what_ are these remaining legal concerns, possibly sharing a plausible legal problematic scenario.

P.S. about "whiter than white" - great mind thinks alike! :smile: I mean that OSM (created the 1st July 2004) thinks like Wikimedia Commons (created the 7 Sept 2004) - https://commons.wikimedia.org/wiki/Commons:Project_scope/Precautionary_principle and vice-versa.

I just mean, again that:

- if a picture should not be placed in front of a map
- if a picture should not be used to extract facts and contribute to a map
- if its description page should not be linked from a map

Then, that specific file seems to have bigger issues and risks a deletion from Wikimedia Commons.

So, waiting for more unanswered plausible legal concerns, I would say again: green light to the code proposer and thanks again for your time! and thanks to any extra heroic code reviewers.

gy-mate:
>>> there's no filter to exclude incompatible licenses yet. I'm guessing CC-0 is the only compatible license
>> 
>> So I would say: what you are doing, is perfectly fine, and any photographer on Wikimedia Commons would thank you so much.
> 
> I think it would be safe if we filter based on [`AttributionRequired=false`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data), but perhaps it's worth asking the LWG for some advice, if no one is certain?

CC0 **_is_** the only ODbL-compatible CC license, unfortunately. See:

https://wiki.openstreetmap.org/wiki/Creative_Commons
https://blog.openstreetmap.org/2017/03/17/use-of-cc-by-data/

gy-mate:
> * The license information comes from [`LicenseShortName`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data). There is a also a boolean called [`AttributionRequired`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data), maybe we could filter based on `AttributionRequired=false` instead of hardcoding compatible license names ?

@k-yle In this case, I think `"Public domain"` and `"CC0"` `LicenseShortName`s could be hardcoded.

valerio-bozzolan:
The mentioned article mentions problems comparing datasets. It does not mention problems about "getting facts from a photo".

Yet again, facts appearing in an image are not subject to copyright. The potential threat you are describing is not existing and does not block this feature.

If somebody wants to conduct their own original research about gathering facts from photos, you're very welcome in doing this, but as a parallel research, so as to do not block the work under this feature, which is totally reasonable, as it allows mappers to create their original edits, thanks to their eyes, and their analisys of facts.

gy-mate:
> The mentioned article mentions problems comparing datasets. It does not mention problems about "getting facts from a photo".
> 
> Yet again, facts appearing in an image are not subject to copyright. The potential threat you are describing is not existing

I agree that facts (e.g. a tree standing at an intersection) cannot be copyrighted, but I think it _does_ matter how you become _aware_ of that fact. If you only know about that tree because you saw it on a CC-BY 4.0 photo or even Google Street View (maybe during armchair mapping), you cannot upload that tree to OSM. See: [https://wiki.openstreetmap.org/wiki/Legal_FAQ#Can_I_trace_data_from_Google_Maps/Nokia_Maps/...?](https://wiki.openstreetmap.org/wiki/Legal_FAQ#Can_I_trace_data_from_Google_Maps/Nokia_Maps/...?)

Therefore, I believe that showing non-ODbL-compatible images as a photo overlay would be harmful to OSM. But—based on the previous comments—there is a solution for this issue.

valerio-bozzolan:
Wikimedia Commons is a project by a like-minded organization. Not a map dataset competitor. Wikimedia Commons by itself is not blocked by that unrelated FAQ, which is about taking from competitors.

We already talked about competitors in the forum thread.

If the legal team wants to clarify this even more, highly appreciated in parallel, so to do not block this feature for no realistic reasons.

EDITED: just to clarify: indeed DO NOT COPY from Google Maps or Google StreetView! Indeed this is a problem! I totally agree. But fortunately unrelated.

gy-mate:
> Wikimedia Commons is a project by a like-minded organization. Not a map dataset competitor. Wikimedia Commons by itself is not blocked by that unrelated FAQ, which is about taking from competitors.

While the FAQ does mention competitor companies, it also says:

_Only sources with compatible licenses - such as US Government information released into the **public domain** - may be used as bases for adding OSM data._

So if a CC-BY photo is uploaded to Wikimedia Commons, it's very much affected by this and cannot be used for editing OSM. Furthermore, it doesn't really matter if Wikimedia Commons is a competitor or not as they do not hold copyrights over uploaded images. _Authors_ hold the copyrights and their license decision should be honored.

> If the legal team wants to clarify this even more, highly appreciated in parallel, so to do not block this feature for no realistic reasons.

This has already been clarified at https://osmfoundation.org/wiki/Licence/Licence_Compatibility#Specific_incompatible_licences — it's a valid concern.

k-yle:
_(assuming we have to restrict the licenses)_

> @k-yle In this case, I think "Public domain" and "CC0" LicenseShortNames could be hardcoded.

Ideally we could use [`AttributionRequired=false`](https://commons.wikimedia.org/wiki/Commons:Machine-readable_data) if this is acceptable, instead of hardcoding values. 

because there are other possible values like `No restrictions`[^1]  which are functionally equivalent , and it seems like there is no exhaustive list[^2], this is just a free-form text field.

[^2]: These parameters are dervied from `<span class="licensetpl_attr_*" style="display: none">false</span>` inside the license templates, and I don't know how to query for something like this


[^1]: Added by [Template:Flickr-no_known_copyright_restrictions](https://commons.wikimedia.org/wiki/Template:Flickr-no_known_copyright_restrictions), see [this example photo](https://commons.wikimedia.org/wiki/File:SS_KUT_in_Sydney_Harbour_(8156187163).jpg)

gy-mate:
Oh, I see, that makes sense!

valerio-bozzolan:
> So if a CC-BY photo is uploaded to Wikimedia Commons, it's very much affected by this and cannot be used for editing OSM

Please stay focused on what you already said: you are saying that Creative Commons licenses must be respected. So, you are saying that copyright must be respected. But, again, Creative Commons licenses **only work in the field of copyright**. Here, we are talking about learning facts from an image, which is not an aspect covered by copyright.

So, thanks, the whole concern is now noted, but not blocking this feature, which is about being able to look at a picture, while mapping. This is not about drawing over an image (derived work / modified form). This is not about copy-pasting a database (derived work). This is not about creating new images (derived work / modified form). This is not about extracting data from a competitor. Etc.

https://www.wipo.int/en/web/copyright § "FAQ What can be protected using copyright?"

https://creativecommons.org/licenses/by/4.0/legalcode.en § Scope § License Conditions

Thanks for sharing a potential concern which could be potentially concerning. This is not. Green light for this merge request. Wikimedia Commons authors will just only say "thank you" for making these materials more useful.

tordans:
@valerio-bozzolan to make sure I understand you correctly: You are saying that we don't need any special permission to map based on images that are licensed CC-BY nor do we need to do anything special like referencing the fact that we use the data anywhere (like we do with some aerial imagery sources). Because by mapping based on the imagery we don't do anything with the image that would be relevant for the given license.

I think we should look for a place in the OSM ecosystem where such a position was documented before. It make the decision easier if we have something from the OSM perspective. I started looking into this:

a. I was hoping the Wiki on the Mapillary-License would have something like this https://wiki.openstreetmap.org/wiki/Mapillary#License but unfortunately it does not. Given that all imagery from Mapillary is CC-BY-NC (old) and CC-BY-SA (current), the situation is the same, right?

b. The OSMF / Licensing Working Group has something about extracting and tracing data from CC-BY imagery from the perspective of aerial imagery: https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers
The fact that there is this waiver can be interpreted two ways: Either it is really needed to make what we do confirm with CC-BY. Or it is just to have something explicit that shows that there is really nothing special going on but everyone feels better to write it down…

valerio-bozzolan:
a. Thanks for mentioning Mapillary again. Mapillary **is** a potential competitor, so, even if copyright **may** be fine for Mapillary (since - in general - learning facts from an image has nothing to do with copyright) there are very tangible extra legal concerns in terms of unfair competition. I think we are welcome to discuss about Mapillary elsewhere, since it has very little to do with this specific feature.

b. [the mentioned waiver](https://osmfoundation.org/wiki/Licence/Waiver_and_Permission_Templates/Template_text_for_aerial_imagery_waivers) is designed to explicitly allow to draw over images, or trace over images, or create vector versions of images, map over images, etc. - these things may be considered "derived work" indeed. Drawing over an image or similar "derived work" can be related to copyright indeed; but fortunately, this feature is **not** about putting images as background to draw over them or trace. So, that waiver is completely unnecessary, since, even if produced, it would allow something totally different, not used by this feature.

tyrasd:
Hey all. Yes, the licensing questions are pretty intricate and need to be evaluated carefully.[^1] I will get in touch with the [LWG](https://osmfoundation.org/wiki/Licensing_Working_Group) for their view on this, and will report back here when I hear back from them.

[^1]: fwiw, initially I would agree with @valerio-bozzolan's interpretation that the situation should indeed different when mapping something from facts deduced from a (e.g. cc-by licensed) street level photo compared to when tracing over a similarly licensed orthophoto. And both are still something slightly different from when working with cc-by licensed vector data.

For the PR itself, here are some initial remarks of things that I noticed while playing with it:
* as the photos on commons typically do have meaningful titles, it would be nice to also show that alongside the user name / date.
* it currently always loads the full "original" resolution of the photos, which can lead to very slow load times for high resolution photos. Perhaps it would make sense to use a checkbox similarly to panoramax to only load the full resolution when requested, and otherwise use a scaled down version of the photo in the viewer.
* a few images on commons are in file formats that browsers typically cannot display natively (e.g. a tif in this [example](
https://pr-11666--ideditor.netlify.app/#disable_features=boundaries&map=16.19/49.42286/8.73111&background=Bing&photo_overlay=commons&photo=commons/157533155); maybe also raw camera formats?). Here it would make sense to either restrict to formats we know a browser can typically display (i.e. jpeg, png, webp, etc.), or use upstream file format conversion?

Danysan1:
> * it currently always loads the full "original" resolution of the photos, which can lead to very slow load times for high resolution photos. Perhaps it would make sense to use a checkbox similarly to panoramax to only load the full resolution when requested, and otherwise use a scaled down version of the photo in the viewer.
> * a few images on commons are in file formats that browsers typically cannot display natively (e.g. a tif in this [example](https://pr-11666--ideditor.netlify.app/#disable_features=boundaries&map=16.19/49.42286/8.73111&background=Bing&photo_overlay=commons&photo=commons/157533155); maybe also raw camera formats?). Here it would make sense to either restrict to formats we know a browser can typically display (i.e. jpeg, png, webp, etc.), or use upstream file format conversion?

Fixing the first point should automatically fix also the second point, as Commons thumbnails are served in jpg/png. Since the code in the PR is using the geosearch Commons API it's possible to simply use the `thumburl` response parameter there. Otherwise it's also possible to use the url `https://commons.wikimedia.org/wiki/Special:FilePath/${urlEncodedImageName}?width=${width}` (for example, using the image example linked above, https://commons.wikimedia.org/wiki/File:Landesamt_f%C3%BCr_Geoinformation_und_Landentwicklung_BaW%C3%BC_%E2%80%93_dop20rgbi_32_480_5474_1_bw_2023.tif has the thumbnail https://commons.wikimedia.org/wiki/Special:FilePath/File:Landesamt_f%C3%BCr_Geoinformation_und_Landentwicklung_BaW%C3%BC_%E2%80%93_dop20rgbi_32_480_5474_1_bw_2023.tif?width=300px ), but I think the `thumburl` option is more reliable.

k-yle:
@tyrasd did you hear back from the LWG?

danieldegroot2:
minor remarks; (I am aware not all are actionable)
- images are always shown as panorama(360); small images are usually shown as flat images.
-- there is no flat/360 filter.
-- full 360 images are not continuous; commons has a link to a separate 360 viewer.
- certain images are shown inconsistently or not at all at low zoom. (no example); such behavior has been also present in some other web-based viewers.
- images at same/similar location cant be navigated to (i.e. from other images) or are not shown (no example).
(not sure how to test/know for sure which ones should be/not shown.)
- images do not scale/resize properly, or not at all ([example](https://pr-11666--ideditor.netlify.app/#disable_features=boundaries&map=19.92/52.65404/-7.24955&background=Bing&photo_dates=2016-09-25_&photo_overlay=commons&photo=commons/172322559)); only adds empty viewer area(black).
- date is always shown in months, exact date only shown on hover (inconsistent with other providers); >12 months is useless/incomprehensive number for some users.
- username is always shown with `User:` prefix, which seems redundant for display purposes (though some providers show `@`)
(note, some users do not have a user page, or import has only external page as mentioned above.)
- link is shown as first / left most license text (inconsistent with other providers); usually shown as last / right most text.

valerio-bozzolan:
> username is always shown with User: prefix

Always always?

Edited:

If we are showing the `|author =`: then seeing `User:` is a feature (?). The credits should be shown as-is, and if the picture has `[[User:Foo]]` in the credits, we should show that as-is. If the credit is a more complex text like `[[User:Foo]], [[User:Bar|BarMan]] and contributors` we should show that as is (rendered as the wikitext API gives) so, `User:Foo, BarMan and contributors`. The credit line can be quite a complex wikitext thing.

But, if we are showing the structured data "[creator](https://www.wikidata.org/wiki/Special:EntityPage/P170)" - than we may want to show the "[author name string](https://www.wikidata.org/wiki/Special:EntityPage/P2093)" which may have no `User` prefix

tordans:
In January I got in contact with people from Wikimedia Germany to ask them for their view on this topic. They cannot do a full legal evaluation on this topic but they send me the message below to share here. I think this adds a valuable perspective to the question. 

 > Hello, I heard there may be some confusion regarding how we can use information from freely licensed images (such as photos on Wikimedia Commons) to improve the map (such as adding information about shops, etc.). Because I am an experienced copyright lawyer, I thought I could help clear up some questions that have been raised recently.
>
> The short version: We are entirely free to take any kind of information from both copyrighted and freely licensed images, be it information contained in the images themselves as well as information contained in the metadata (be it in the image description or in the actual metadata), totally without being required to worry about the copyright and/or the free license in the image in any way. We do not need to attribute the photographers, reproduce the license – there are no such restrictions whatsoever.
>
> Why is that? Because copyright law, and with that any free license, only applies to the image itself, and not to the information contained therein or in the metadata. You may have heard the phrase 'information cannot be copyrighted' before, and this is what this is about. Since copyright law does not protect information contained in images and in their metadata, free license can likewise not 'generate' any such protection.
>
> If we want to go even deeper, you may have heard of database works or the EU sui generis database right. Well, the information contained in an image or in its metadata in itself does not constitute a database under copyright law. Image collections, take the entirety of images on Wikimedia Commons, might be protected databases under certain conditions as a whole, but the information contained in individual images or in their metadata does not constitute protected elements thereof except in specific cases of 'curated' image metadata that we do not need to worry about here (e.g. catalogue metadata contained in a set of images depicting a museum collection).
>
> I hope that you find this helpful, and that this clears up any perceived restrictions to using information from both copyrighted and freely licensed images to improve the map. I am happy to answer any follow-up questions, and may update this post in case someone points out anything that I should add to make this easier to understand for everyone. This is really something that mappers should not be worrying about, so I want to make this as clear as possible.

gy-mate:
Thank you, that's great news!

Should https://wiki.openstreetmap.org/wiki/Street-level_imagery_services#With_photosphere_support be corrected then?

| Name              | May be used for OSM mapping |
|-------------------|-----------------------------|
| Wikimedia Commons | ~~depends on image license~~ **yes**    |

stkdiretto:
The statement that using images from Commons depends on the license was likely never true from a licensing point of view, so it should be corrected, IMO. I only found the classifications of the other street level photography providers through this PR – I find it highly doubtful whether there is any legal basis to prevent users of those services to use factual information found within the photos to use in OSM. I don't even see anything towards that in the linked T&C from e.g. StreetView.

(edit: This is my personal account, I was involved in the exchange from WMDE's POV)

gy-mate:
> I will get in touch with the [LWG](https://osmfoundation.org/wiki/Licensing_Working_Group) for their view on this, and will report back here when I hear back from them.

@tyrasd Have you heard back from them since?

gy-mate:
@grischard @mackerski @kathleenlu09 @sbma44 If you could post the LWG's standpoint as a comment here, that would be much appreciated! Thank you! :)

1ec5:
With my OpenHistoricalMap hat on, I’ll add that integration with Wikimedia Commons would be quite desirable, since Commons is already one of the most, er, common sources of contributions to the project via our fork of iD. We’re even more excited about tweaking this feature to also integrate with [Yesterdays](https://openstreetmap.us/our-work/charter-projects/yesterdays/).

One frequent source of hesitation among prospective contributors is the prospect of securing a clean copyright status for every source. We’re constantly having to reassure mappers that facts aren’t subject to copyright protection, so they can feel free to crib dates from newspaper articles, names from books, etc. without worrying at all about licensing.

That said, we do have to caution against copying details from copyrighted maps, because the line between facts and depictions of facts gets blurry fast. Fortunately, I think this is more of a consideration when using [Wikimaps Warper](https://warper.wmflabs.org/) to generate base layers than when using Wikimedia Commons to show an image in a window.

PR #11675, opened 2025-12-13
Fixed #9406: By treating aerialway=station as valid start/end point which suppresses the false warning

### LINKED ISSUE #9406: Footways in ID are flagged as disconnected when between ski lifts and ski piste
### URL

https://www.openstreetmap.org/way/713694683

Open ID, select the 'isolated' footway connecting gondola cable car to cable way/lift or cable way/lift to ski piste and the alert pops up of "Foot Path is disconnected from other roads and paths"

https://www.openstreetmap.org/way/713694683

Think the resolve would be to consider the cable car ways and maybe ski piste as (virtual) footway, or suppress the false alerts in these situations in ID Editor. See no alerts in OSM Inspector on this.

Note that to have seen similar on golf courses where paths are mapped from green side to next tee. No getting out but to hit the ignore button. Applies to all web browser, all ID versions. Not good.

_No response_

All

Vivaldi

Razen04:
## Description
Currently, iD warns for paths which are not connected to any other route any is contained within a building like gondola stations which is a false positive. This has been explained in detail in Issue #9406.

## Fix
The fix to this is simple, we just add an extra check in the `isConnectedVertex` function to check if a vertex in the path has `aerialway=station` and return `true` for that, which means the path is valid to have start/end in the gandola station even if they are not connected to any other roads or highway(which they aren't supposed to).

Fixes #9406 

## Before:

Screenshot 2025-12-13 at 3 53 10 PM

## After the if statement:

Screenshot 2025-12-13 at 3 53 47 PM

## References
This is similar to issue #3906 - which showed problems with foothpaths connecting two buildings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Use Röntgen icon set directly from upstream npm package ([#11784], thanks [@tordans])
 * Replace deprecated `document.createEvent`/`initEvent` with modern [Event] constructor ([#11870], thanks [@JaiswalShivang])
 * Fix crash in country combo field when entering unrecognized ISO country codes ([#11904], thanks [@JaiswalShivang])
-* Fixed showing "Disconnected Way" warnings for paths ending at aerialway stations and stopping locations. ([#9406], thanks [@Razen04])
+* Fixed showing "Disconnected Way" warnings for ways ending at aerialway stations and stopping locations. ([#9406], thanks [@Razen04])
 
 
 [#8464]: https://github.com/openstreetmap/iD/issues/8464
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Use Röntgen icon set directly from upstream npm package ([#11784], thanks [@tordans])
 * Replace deprecated `document.createEvent`/`initEvent` with modern [Event] constructor ([#11870], thanks [@JaiswalShivang])
 * Fix crash in country combo field when entering unrecognized ISO country codes ([#11904], thanks [@JaiswalShivang])
+* Fixed showing "Disconnected Way" warnings for paths ending at aerialway stations and stopping locations. ([#9406], thanks [@Razen04])
 
 
 [#8464]: https://github.com/openstreetmap/iD/issues/8464
@@ -93,6 +94,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
 [#11870]: https://github.com/openstreetmap/iD/issues/11870
 [#11904]: https://github.com/openstreetmap/iD/issues/11904
+[#9406]: https://github.com/openstreetmap/iD/issues/9406
 [#id-tagging-schema/pull/1507]: https://github.com/openstreetmap/id-tagging-schema/pull/1507
 [@ilias52730]: https://github.com/ilias52730
 [@Razen04]: https://github.com/Razen04
--- a/modules/validations/disconnected_way.js
+++ b/modules/validations/disconnected_way.js
@@ -151,6 +151,9 @@ export function validationDisconnectedWay() {
                 vertex.tags.entrance !== 'no') return true;
             if (vertex.tags.amenity === 'parking_entrance') return true;
 
+            // treat nodes inside gandola stations and ski lifts as connected
+            if (vertex.tags.aerialway === 'station') return true;
+
             return false;
         }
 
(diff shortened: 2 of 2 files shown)


modules/validations/mutually_exclusive_tags.js
```diff
@@ -11,13 +11,33 @@ export function validationMutuallyExclusiveTags(/* context */) {
...
+          return typeof rule === 'string' ? rule : `${rule.key}=${rule.value}`;
+        }
 
         let pairsFounds = tagKeyPairs.filter((pair) => {
-            return (pair[0] in entity.tags && pair[1] in entity.tags);
+            return matchesRule(pair[0], entity.tags) && matchesRule(pair[1], entity.tags);
         }).filter((pair) => {
             // noname=no is double-negation, thus positive and not conflicting. We'll ignore those
-            return !((pair[0].match(/^(addr:)?no[a-z]/) && entity.tags[pair[0]] === 'no') ||
-                     (pair[1].match(/^(addr:)?no[a-z]/) && entity.tags[pair[1]] === 'no'));
```
k-yle, line 20: these changes seem to be unrelated to the PR description. Are they meant to be a different git branch?

Razen04: Yeah I mistakenly worked on different issue on the same branch, my bad. I will fix it asap.

modules/validations/disconnected_way.js
k-yle, line 1: we have [another preset, called _Aerialway Stopping Location_](https://github.com/openstreetmap/id-tagging-schema/blob/5c85ba0b277f21739a58629aa29e2cea802dc655/data/presets/public_transport/stop_position_aerialway.json), which seems to be used in much the same way, and also seems to be a routing sink. 

Should this logic apply to that tag as well?

Razen04: I think `public_transport=stop_position && aerialway=*` should be treated the same way as `aerialway=station`.

In both cases, the mapper is explicitly marking a boundary where pedestrian movement ends and aerial transport begins.

This is different from public_transport=stop_position for buses, trams, or trains, where pedestrian connectivity is expected to continue.

So should I add that case too in the code?

matkoniecz: I think so, but I would be also fine with doing it as a next PR (note: not a maintainer)

Razen04: Let's see why k-yle says. The fix is just one line more i suppose.

matkoniecz: @k-yle do you have by any chance time to look into that?

k-yle: it should be a trivial change, i would fix it in this PR imo, so that we have a centralised place to this discuss this topic

Razen04: I'll do it.

Razen04:
I messed it up in between, lol. Anyways I think I made. the correct change, do look into it once.

matkoniecz:
Yeah, this 579 changed files looked a bit suspect :)

modules/validations/disconnected_way.js
matkoniecz: appears to be done, from what I see

modules/validations/disconnected_way.js
```diff
@@ -151,6 +151,11 @@ export function validationDisconnectedWay() {
                 vertex.tags.entrance !== 'no') return true;
             if (vertex.tags.amenity === 'parking_entrance') return true;
 
+            // treat nodes inside gondola stations and ski lifts as connected
+            if (vertex.tags.aerialway) {
+                if (vertex.tags.aerialway === 'station' || vertex.tags.public_transport === 'stop_position') return true;
```
matkoniecz, line 156: I guess it would also count say `public_transport=stop_position aerialway=no` ?

And  `public_transport=stop_position aerialway=whatever_gibberish` ?

Should requiring specifically `aerialway=station` or `public_transport=stop_position aerialway=yes` be needed?

CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 * Use Röntgen icon set directly from upstream npm package ([#11784], thanks [@tordans])
 * Replace deprecated `document.createEvent`/`initEvent` with modern [Event] constructor ([#11870], thanks [@JaiswalShivang])
 * Fix crash in country combo field when entering unrecognized ISO country codes ([#11904], thanks [@JaiswalShivang])
+* Fixed showing "Disconnected Way" warnings for paths ending at aerialway stations and stopping locations. ([#9406], thanks [@Razen04])
```
matkoniecz:
```suggestion
* Fixed showing "Disconnected Way" warnings for ways ending at aerialway stations and stopping locations. ([#9406], thanks [@Razen04])
```

Are we actually checking anywhere is it path? From what I see from code it will also treat `highway=motorway` ending on `aerialway=station` as connected

tyrasd:
I fixed this slightly differently in cd255c82323cefacf91c158503d0d768ff237cf9: the `aerialway` ways themselves can be considered to be part of the routable network in such cases. There's no need to do the hacky _treat vertex as routing source_ for this particular issue.

PR #11682, opened 2025-12-14
Make panel close button visually distinct (#11672)

### LINKED ISSUE #11672: Right side panel closer button looks too much like buttons below it
Image

Here the side panel closer "X" is stacked directly above the "+" etc., and with the same perfect spacing. It takes the [stick out from the side](https://github.com/alltheplaces/alltheplaces.xyz/issues/72#issuecomment-3646338273) approach, but alas, ended up right in "the checkmate path of the rook."

Sure, in the photo we can see it has a slightly different background.

But when not blown up, well, it looks all just part of the same dark theme.

So perhaps it could be jogged around or droop out off on the side, or [go for the green](https://github.com/alltheplaces/alltheplaces.xyz/issues/72#issuecomment-3646487346)... something, anything, to make it look different from the other buttons. Thanks.

Mayankaggarwal8055:
This PR increases the panel close button width from 40px to 50px, adjusts its background color to better distinguish it from adjacent tool buttons, and adds an explicit hover style to preserve clear interaction feedback.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4144,8 +4144,13 @@ li.issue-fix-item button:not(.actionable) .fix-icon {
 }
 
 .pane-heading button {
-    width: 40px;
+    width: 50px;
     border-radius: 0;
+    background-color:  black;
+}
+
+.pane-heading button:hover {
+    background-color: #121212; /* slightly lighter */
 }
 
 .pane-content {


tyrasd:
This didn't fix the issue in regular (non-dark mode) light theme, where the black background looks terrible. also, see #11672: this is a non-issue in practice, so not necessarily worth addressing. If you still really like working on it, you could try to implement it like described in the last paragraph of https://github.com/openstreetmap/iD/issues/11672#issuecomment-3675490804.

Mayankaggarwal8055:
Thanks for the feedback — that makes sense.
I hadn’t tested the change in the light theme, and I agree the black background doesn’t work there.

I understand why this was closed, given that it’s a minor/non-issue in practice. If it’s still useful, I’m happy to rework this following the approach described in the last paragraph of #11672, with theme-aware styling so light mode isn’t negatively affected.

Let me know whether you’d prefer that, or if it’s best left as-is.

PR #11683, opened 2025-12-14
(#11676) Prioritize parent relations in “Choose a parent relation” dropdown

### LINKED ISSUE #11676: Relation dropdown menu should prioritize parent relations of connected features
The Relations section’s “Choose a parent relation” dropdown menu shows all the relations in the current viewport, sorting them by relation ID. Most users don’t have much insight about relation IDs. Instead, we should try to sort the relations by something more directly relevant to the task at hand, such as prioritizing the parent relations of connected features.

#7964 proposes to sort the most recently used relations to the top. This is helpful when building a relation by adding its members individually, but it doesn’t help when making a spot fix, such as closing a gap in a broken relation. Besides, when I’m in the middle of building out a public transportation route, I’m usually juggling two or three different relations at a time (one for each direction, plus a route master relation), which might be named similarly. These relations would swap places frequently in an MRU list, making the task quite error-prone.

The most common relation types expect all or most of the relation’s members to be connected to each other. Of [the relation types that have presets](https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/type/) in id-tagging-schema, only `enforcement`, `public_transport`, and `site` mostly include disconnected members.

#11588 would add any other relation the user has previously modified in the same session. This would make the list more heterogeneous, so a more intelligent sorting order would be even more helpful.

Mayankaggarwal8055:
When working with relations, I noticed that the “Choose a parent relation” dropdown was sorted only alphabetically, which made it easy to miss the most relevant relation during quick edits—especially when a feature already belonged to multiple relations.

In this PR, I’ve adjusted the sorting so that relations the selected feature already belongs to appear at the top of the dropdown. All other relations continue to be sorted alphabetically, just as before.

This small change makes it quicker to find the right relation and helps reduce the chance of accidentally adding a feature to the wrong one during everyday mapping workflows.

### DIFF
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -115,52 +115,7 @@ export function uiSectionRawMembershipEditor(context) {
             membership.role = roles.length === 1 ? roles[0] : roles;
         });
 
-        const parentRelationIDs = new Set(
-            getSharedParentRelations().map(r => r.id)
-        );
-
-        const existingRelations = memberships
-            .filter(membership => !recentlyAdded.has(membership.relation.id))
-            .map(membership => ({
-                ...membership,
-                // We only sort relations that were not added just now.
-                // Sorting uses the same label as shown in the UI.
-                // If the label is not unique, the relation ID ensures
-                // that the sort order is still stable.
-                _sortKey: [
-                    baseDisplayValue(membership.relation),
-                    membership.relation.id,
-                ].join('-'),
-            }))
-            .sort((a, b) => {
-
-                const aIsParent = parentRelationIDs.has(a.relation.id);
-                const bIsParent = parentRelationIDs.has(b.relation.id);
-
-
-                // 1️⃣ Parent relations first
-                if (aIsParent && !bIsParent) return -1;
-                if (!aIsParent && bIsParent) return 1;
-
-
-                // 2️⃣ Same category → older alphabetical sort
-                return a._sortKey.localeCompare(
-                    b._sortKey,
-                    localizer.localeCodes(),
-                    { numeric: true },
-                );
-            });
-
-
-        const newlyAddedRelations = memberships
-            .filter(membership => recentlyAdded.has(membership.relation.id));
-
-        return [
-            // the sorted relations come first
-            ...existingRelations,
-            // then the ones that were just added from this panel
-            ...newlyAddedRelations,
-        ];
+       return memberships;
     }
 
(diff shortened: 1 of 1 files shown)


1ec5 (changes requested):
I think this is incorrectly touching the code for listing the relations that the selected feature is already a member of. #11676 is about the dropdown menu for adding the feature to another relation. The items are sorted around here, and no, it was not alphabetical:

https://github.com/openstreetmap/iD/blob/27638e5da88c39f0b0aa59236d55af873025f4f2/modules/ui/sections/raw_membership_editor.js#L330-L332

Before requesting a review, please try out your changes in a local instance of iD to make sure it does what was requested. Thank you.

Mayankaggarwal8055:
I’ve updated the PR with the latest changes.

The existing membership list behavior is unchanged, and the parent-first logic
is now applied only to the “Add to relation” dropdown while preserving the
original creation order. Tested locally.


before dropdown changes: 

Screenshot (2)


after dropdown chnages:

Screenshot (1)

modules/ui/sections/raw_membership_editor.js
```diff
@@ -327,9 +297,26 @@ export function uiSectionRawMembershipEditor(context) {
...
+            );
+
+            result.sort((a, b) => {
+
+                const aIsParent = !!(a.relation && parentRelationIDs.has(a.relation.id));
+                const bIsParent = !!(b.relation && parentRelationIDs.has(b.relation.id));
+
+                // 1) parent relations first
+                if (aIsParent && !bIsParent) return -1;
+                if (!aIsParent && bIsParent) return 1;
```
k-yle, line 315: this logic appears to be checking the parents of _this_ relation.

#11676 is proposing to sort based on the parents of _connected features_. This will require:
 - finding all connected features
    - for nodes, this refers to the node's parent ways
    - for ways, this refers to any way that shares a node with this way
 - find the parent relations of those connected features

This will likely be a challenging issue to implement. If you'd like to start with something easier, check out the issues that are tagged as https://github.com/openstreetmap/iD/labels/new%20contributor%20opportunity

Mayankaggarwal8055:
Thanks for the detailed clarification — I understand now that #11676 requires
prioritizing relations based on the parents of connected features (nodes/ways),
which is a much broader change than what I initially implemented.

Given the scope and complexity involved, I think it’s best for me to step back
from this PR for now and close it, rather than risk an incomplete solution.

Appreciate the guidance — it was helpful in understanding the underlying model.

Issue #11679, opened 2025-12-13
Capitalize the initial letter of genus, species, and taxon tag values


Marc-marc-marc:
### Description

Scientific taxon names follow nomenclature rules: the initial letter of taxon names (e.g. genus) must be capitalized. The web editor should :

- enforce (or validator warning) capitalization the initial letter
and/or
- filter or convert taginfo values to display them with capitalization of the initial letter
and/or
- autofix user input

for ex, user input genus=quercus
the correct value is genus=Quercus

tyrasd:
As far as I can tell, the non-capitalization of those tags doesn't seem to be a problem in practice: Almost all entered values do already follow the correct capitalization. I think this is mostly because the UI already guides the user well towards the expected format, for example by  the fields' placeholder text, autocomplete, etc.

> autofix user input

I don't think that's the best approach: It would just hide situations (from other mappers and QA tools) where a user accidentally enters something really invalid, e.g. `species=unknown`, `genus=birch`, etc.

> filter or convert taginfo values to display them with capitalization of the initial letter

This is in principle what iD already does, isn't it? That said, I did actually notice that the <kbd>Tab</kbd>-autocomplete does not properly capitalize the initial letter as expected. That's a bug, as the autocomplete should do this when a field does specify that it is case sensitive.

> validator warning

This might be useful, so that people can learn that the expected format is with a capital letter. On the other hand, given that this might not be a very common tagging mistake, it could also be sufficient if this particular tagging mistake is handled by external QA tools like Osmose.

Mayankaggarwal8055:
Hi @tyrasd ,  
I’ve opened PR #11712 to fix the tab-autocomplete behavior for case-sensitive fields, preserving the original suggestion casing.

Please let me know if this looks good or if any adjustments are needed.

matkoniecz:
there is a lot of fiddly thing here, for example capitalizing `genus=sugar mapple` is just hiding errors

cases like `genus=quercus` can be just mass-fixed as a clear and blatant mistakes (and in fact I added some to https://codeberg.org/matkoniecz/OpenStreetMap_cleanup_scripts/commit/5a36951b90ecf32d7a2c6872a4b6695e5cd72259 and started edit)


> That said, I did actually notice that the Tab-autocomplete does not properly capitalize the initial letter as expected. That's a bug, as the autocomplete should do this when a field does specify that it is case sensitive.

this is a bug and I will create a separate issue for it

matkoniecz:
[...]

PR #11685, opened 2025-12-15
Capitalized the initial letter of genus, species, and taxon tag values 

ChehakTrehan:
Fixes #[1697](https://github.com/openstreetmap/iD/issues/11679)

Automatically capitalize the first letter of scientific taxon tag values (genus, family, subfamily, tribe) when saving, ensuring correct nomenclature.


tyrasd:
as mentioned in https://github.com/openstreetmap/iD/issues/11679#issuecomment-3675198831, this is not the best approach to address the original issue.

PR #11691, opened 2025-12-15
Added a mutually exclusive key and value and updated the functions to work with string and object

### LINKED ISSUE #9449: Consider `crossing=unmarked` + `crossing_ref=zebra` as a tagging error
### Description

For pedestrian crossings (`highway=crossing` nodes), `crossing_ref=zebra` implies the presence of markings. Therefore the combination of `crossing=unmarked` + `crossing_ref=zebra` cannot be valid and should be considered an error.

An iD user in my area keeps incorrectly tagging crossings like this for some reason. It would be nice if this was caught by the validator to prevent these errors from creeping in.

### Screenshots

_No response_

Razen04:
Fixes #9449 

## Changes made:

1. Added a key and value pair in the `osmMutuallyExclusiveTagPairs` for `crossing=unmarked` and `crossing_ref=zebra`.
2. Updated the way to get pairsFound to handle both string and object.

## Before:
Screenshot 2025-12-15 at 7 11 04 PM

> There is no warning for `crossing_ref=zebra` and `crossing=unmarked`

## After:
Screenshot 2025-12-15 at 7 11 37 PM

> Now there will be a warning for those specific key=value pairs.

### DIFF
--- a/modules/validations/mutually_exclusive_tags.js
+++ b/modules/validations/mutually_exclusive_tags.js
@@ -38,7 +38,7 @@ export function validationMutuallyExclusiveTags(/* context */) {
         }).filter((pair) => {
             // noname=no is double-negation, thus positive and not conflicting. We'll ignore those
             return !(isIgnoredNoTag(pair[0], entity.tags) ||
-                     isIgnoredNoTag(pair[0], entity.tags));
+                     isIgnoredNoTag(pair[1], entity.tags));
         });
 
         // Additional:
--- a/modules/validations/mutually_exclusive_tags.js
+++ b/modules/validations/mutually_exclusive_tags.js
@@ -12,12 +12,33 @@ export function validationMutuallyExclusiveTags(/* context */) {
 
     const validation = function checkMutuallyExclusiveTags(entity /*, graph */) {
 
+        function matchesRule(rule, tags){
+          if (typeof rule === 'string') {
+            return rule in tags;
+          }
+          return tags[rule.key] === rule.value;
+        }
+
+        function isIgnoredNoTag(rule, tags){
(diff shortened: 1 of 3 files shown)


matkoniecz:
Can you try adding a test covering at least part of new functionality to tests?

matkoniecz:
Maybe `crossing_ref` and `crossing:markings=no` should also conflict while we are at this?

modules/validations/mutually_exclusive_tags.js
```diff
@@ -12,12 +12,33 @@ export function validationMutuallyExclusiveTags(/* context */) {
...
+
         let pairsFounds = tagKeyPairs.filter((pair) => {
-            return (pair[0] in entity.tags && pair[1] in entity.tags);
+            return matchesRule(pair[0], entity.tags) && matchesRule(pair[1], entity.tags);
         }).filter((pair) => {
             // noname=no is double-negation, thus positive and not conflicting. We'll ignore those
-            return !((pair[0].match(/^(addr:)?no[a-z]/) && entity.tags[pair[0]] === 'no') ||
-                     (pair[1].match(/^(addr:)?no[a-z]/) && entity.tags[pair[1]] === 'no'));
+            return !(isIgnoredNoTag(pair[0], entity.tags) ||
+                     isIgnoredNoTag(pair[0], entity.tags));
```
k-yle, line 41:
```suggestion
                     isIgnoredNoTag(pair[1], entity.tags));
```
this looks like a logic error

Razen04: Yeah my bad, I most probably forgot to change it after copying.

Razen04:
Fixed the loical bug and added test cases and also added the pair `crossing` and `crossing:markings=no` as the mutually_exclusive_pair.

PR #11692, opened 2025-12-15
adjusted housenumber and floor widths for tw

### LINKED ISSUE #11622: Floor field too wide, housenumber too narrow
Are you sure `Floor` needs such a wide field, at the expense of the housenumber?

Image

Let's check how many digits it would take to cut off:
```
addr:housenumber=7之12號
addr:housenumber=123456
```
Well in the USA house numbers get that large, and even have N,W,S,E in them between the digits too sometimes.

Since when does floor need that much room?
>Hey mom, I'm on the 123456th Floor.

Well, OK, "[mezzanine](https://en.wikipedia.org/wiki/Mezzanine)" floor. But, well, we can be like the elevator button: "M".

bhavyaKhatri2703:
image

adjusted housenumber, floor , unit in a ratio of 0.4 : 0.4 : 0.2

fixes #11622

### DIFF
--- a/data/address_formats.json
+++ b/data/address_formats.json
@@ -124,7 +124,12 @@
       ["postcode", "city", "district"],
       ["street+place"],
       ["housenumber", "floor", "unit"]
-    ]
+    ],
+    "widths": {
+        "housenumber": 0.4,
+        "floor": 0.4,
+        "unit": 0.2
+    }
   },
   {
     "countryCodes": ["jp"],


tyrasd:
b06496780f555b23afed0b8996afc1f2b2ba9041 has a more pragmatic approach to fix this: `addr:floor` contains similarly short values like `addr:housenumber` and ?`addr:unit`, thus it should always be rendered equally narrow like those other subfields. There's no need to add a manual exception for Taiwan for this field. Instead, this commit added a proper default width for `addr:floor`.

Issue #9115, opened 2022-05-19
Feature type golf = lateral_water_hazard and golf = water_hazard  removes water=* (pond, lake, reservoir, etc)


brierjon:
### URL

https://www.openstreetmap.org/way/699711932/history

### How to reproduce the issue?

Feature Types Lateral Water Hazard and Water Hazard remove water=* tags. Instead of tagging via Feature Type these should be  added fields to natural = water as to not remove water = lake, water=pond, water=reservoir etc as being a water hazard for golf is an attribute of water instead of primary feature of a pond or natural lake even if located within or next to a golf course.

Users may not know to edit tags directly when using ID and to prevent removal of applicable tags either selecting Lateral Water Hazard or Water Hazard or when adding reservoir, lake, pond etc designations for the water. At the very least selecting water hazard should not remove water = * tags if not made an attribute of water instead of feature type. By changing the relationship of how water hazards are tagged with water areas it would avoid accidental or unnoticed removal of water tags.

### Screenshot(s) or anything else?

- https://www.openstreetmap.org/way/30105408/history#map=18/39.23700/-84.25005
- https://www.openstreetmap.org/relation/1850518/history

### Which iD Editor versions do you see the issue on?

Released version at openstreetmap.org/edit

### Which browsers are you seeing this problem on?

Firefox

PR #11696, opened 2025-12-16
when changing presets, delete keys from the old preset which are not in the new preset

### LINKED ISSUE #2069: Feature type golf = lateral_water_hazard and golf = water_hazard  removes water=* (pond, lake, reservoir, etc)
### URL

https://www.openstreetmap.org/way/699711932/history

Feature Types Lateral Water Hazard and Water Hazard remove water=* tags. Instead of tagging via Feature Type these should be  added fields to natural = water as to not remove water = lake, water=pond, water=reservoir etc as being a water hazard for golf is an attribute of water instead of primary feature of a pond or natural lake even if located within or next to a golf course.

Users may not know to edit tags directly when using ID and to prevent removal of applicable tags either selecting Lateral Water Hazard or Water Hazard or when adding reservoir, lake, pond etc designations for the water. At the very least selecting water hazard should not remove water = * tags if not made an attribute of water instead of feature type. By changing the relationship of how water hazards are tagged with water areas it would avoid accidental or unnoticed removal of water tags.

- https://www.openstreetmap.org/way/30105408/history#map=18/39.23700/-84.25005
- https://www.openstreetmap.org/relation/1850518/history

Firefox

### LINKED ISSUE #11175: retagging stile leaves stile= tag behind
### URL

_No response_

1. select vertex within footway
2. select stile from feature list (`barrier=stile` gets applied)
3. select its type in iD interface (`stile=` gets applied)
4. select gate from feature list (`barrier=gate` gets applied, `stile=` remains behind - it should have been removed)

http://overpass-turbo.eu/s/1c3g finds about 300 other similar issues

number is quite low but it is result of ongoing cleanup, I also opened many notes like https://www.openstreetmap.org/note/4832435

problem is more general, see say "retagging amenity=restaurant with cuisine=obscure_value leaves cuisine #11037"

_No response_

_No response_

k-yle:
Previously discussed at https://github.com/openstreetmap/id-tagging-schema/pull/1622#issuecomment-3036657726
 - GoMap has [already implemented](https://github.com/bryceco/GoMap/commit/bf9236b759075cc58a8d61d38a5a07b8c61f16ce) the logic proposed here.
 - StreetComplete and CoMaps have their own [manually curated logic](https://github.com/mnalis/StreetComplete-taginfo-categorize).

---

Currently, if you change `building=yes` + `building:levels=5` to `landuse=grass`, then the grass will still have `building:levels`, which is a bit weird.

There are endless examples like this. This topic has been discussed many times in different PRs. [`removeTags`](https://github.com/ideditor/schema-builder#removetags) can solve this, but it's not clear when we should use `removeTags`. 

---

New logic:
* universal keys like `name`, `check_data` are always kept
* keys that do not exist in the tagging schema are always kept, like `ref:FR:some_random_tag`
* keys that are not used by the old preset are always kept
* keys that exists in the old preset, but not in the new preset, are deleted.

---

* fixes [#11175](https://github.com/openstreetmap/iD/issues/11175)
* fixes [id-tagging-schema#1622](https://github.com/openstreetmap/id-tagging-schema/pull/1622)
* fixes [id-tagging-schema#1621](https://github.com/openstreetmap/id-tagging-schema/pull/1621)
* fixes [id-tagging-schema#1624](https://github.com/openstreetmap/id-tagging-schema/pull/1624)
* fixes [id-tagging-schema#1686 (comment 2537104048)](https://github.com/openstreetmap/id-tagging-schema/pull/1686#discussion_r2537104048)
* fixes [#9115](https://github.com/openstreetmap/id-tagging-schema/issues/2069)
* discussed at [id-tagging-schema#1800](https://github.com/openstreetmap/id-tagging-schema/issues/1800)
* discussed at [#11175 (comment 3019700503)](https://github.com/openstreetmap/iD/issues/11175#issuecomment-3019700503)
* does not fix&shy; [id-tagging-schema#1283](https://github.com/openstreetmap/id-tagging-schema/issues/1283#issuecomment-2226890322) - this is a complex example where we still need [`removeTags`](https://github.com/ideditor/schema-builder#removetags)


tordans:
> universal keys like name, check_data are always kept

Is this part in the code already?

k-yle:
> Is this part in the code already?

yeah, right now iD doesn't delete any _fields_ when changing presets. 

So the only change is to the 4th bullet point.

tordans:
If we release this, we should look into adding a recommendation in the docs (probably schema builder) to hint at other apps that this is the common handling of this case. (Note to self)

tyrasd (approved):
This sounds like the most sensible thing do do.

I noticed that none of the tests seems to cover the case for a tag which does not exist in either preset… Shall we maybe add a test for that case while we're at it? //edit: added this in 6f92a296c

modules/actions/change_preset.js
```diff
@@ -18,9 +20,22 @@ export function actionChangePreset(entityID, oldPreset, newPreset, skipFieldDefa
...
                 newPreset.fields(loc).concat(newPreset.moreFields(loc))
                     .filter(f => f.matchGeometry(geometry))
-                    .map(f => f.key).filter(Boolean)
+                    .flatMap(f => f.allKeys())
+                    .filter(Boolean)
                     .forEach(key => preserveKeys.push(key));
             }
+
+            if (oldPreset) {
+                /** 'field-keys' are keys used by fields (different to the keys used by preset itself) */
```
tyrasd: does `/**` do anything special here? It looks like a regular inline comment, but the "jsdoc" syntax used makes this stand out a bit…

k-yle: yeah, it's a bit pointless for this local variable. fixed

matkoniecz:
> fixes https://github.com/openstreetmap/id-tagging-schema/issues/2069

how it does it?

https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/presets/golf/lateral_water_hazard.json#L4 does not have `water=` tags, so they will be still deleted - right?

When migrating from say https://github.com/openstreetmap/id-tagging-schema/blob/93113eccdddcedb502d951dc206c7add951b1d68/data/presets/natural/water/pond.json#L15

k-yle:
> > fixes https://github.com/openstreetmap/id-tagging-schema/issues/2069
> 
> how it does it?

If we add a field for `water=*` to the _Lateral Water Hazard_ preset, then the tag will be preserved when you switch between the presets _Lake_ and _Lateral Water Hazard_.

So it requires a change in this repo and iD-tagging-schema to fully fix that issue

matkoniecz:
I expected that https://github.com/openstreetmap/id-tagging-schema/issues/951 would be also fixed, but it still requires fiddling with paths.

matkoniecz:
> StreetComplete and CoMaps have their own [manually curated logic](https://github.com/mnalis/StreetComplete-taginfo-categorize).StreetComplete and CoMaps have their own [manually curated logic](https://github.com/mnalis/StreetComplete-taginfo-categorize).

I wonder whether it also should be partially used to purge things like ```"fhrs:.*", "old_fhrs:.*", "fvst:.*", "ncat", "nat_ref", "gnis:.*", "winkelnummer",
    "type:FR:FINESS", "type:FR:APE", "kvl_hro:amenity", "ref:DK:cvr(:.*)?", "certifications?",
    "transiscope", "opendata:type", "local_ref", "official_ref", "ref:thw", "ref:nav",
    "ref:GR:kepId", "ref:bufa", "ref:walmart", "ref:rspo", "ref:cue", "ref:regon",
    "ref:BY:trade_register", "ref:vb", "ref:comfort", "ref:bademilj.", "ref:HU:vatin",
    "ref:HU:company", "ref:ico", "ref:scoutingnl", "ref:anlegg", "ref:FR:sport_equip",
    "ref:FR:RNA", "ref:FR:CRTA", "ref:charity",```

PR #11697, opened 2025-12-16
Fix changeSet keys like hashtag being a suggestion for objects

### LINKED ISSUE #11625: iD suggests hashtags as a key on objects

iD will suggest _hashtags_ as a key on OSM objects even though this is a tag key that should be limited to changesets.

Image

2.37.3

Other (please specify) browser independent.

bhavyaKhatri2703:
added a new regex expression for excluding key names similar to _extraExcludedKeys. 
Not sure what else can be added to this expression.

Fixes #11625


modules/services/taginfo.js
```diff
@@ -15,6 +15,8 @@ var _popularKeys = {};
 // these will be returned by keys(), but taginfo will not be queried for values() requests
 var _extraExcludedKeys = /^(postal_code|via|((int_|loc_|nat_|official_|old_|ref_|reg_|short_|full_|sorting_|alt_|artist_|long_|bridge:|tunnel:)?name(:left|:right)?(:[a-z]+)?))$/;
 
+var _extraExcludedKeyNames = /^(hashtag|hashtags)$/;
```
hlfan:
```suggestion
var _extraExcludedKeyNames = /^(hashtags?|created_by)$/;
```

modules/services/taginfo.js
```diff
@@ -245,7 +247,7 @@ export default {
                 callback(err);
             } else {
                 var f = filterKeys(params.filter);
-                var result = d.data.filter(f).sort(sortKeys).map(valKey);
+                var result = d.data.filter(f).filter(function(d) { return !_extraExcludedKeyNames.test(d.key); }).sort(sortKeys).map(valKey);
```
hlfan:
```suggestion
                var result = d.data.filter(f).filter(d => !_extraExcludedKeyNames.test(d.key)).sort(sortKeys).map(valKey);
```

hlfan (commented):
This looks good to me now.

bhavyaKhatri2703:
am i allowed to tag someone for a review?

k-yle (approved):

Issue #10727, opened 2025-02-02
Include addr:housename to the address squema. Several Countries do not have mandatory addr:housenumber, but several buildings do have a name


jjiglesiasg:
### Description

In several countries (including #Bolivia), buildings has a name, but not addr:housenumber are displayed outside or inside those buildings , also those buildings contains bussiness inside of all kind, and peopple tends to orient themselves by the addr:streetname plus the name of the building where the bussiness and apartments are located, instead of numbers that maybe or not displayed.

Proposed Solution
Add addr:housename as desirable data to be collected and filled when inquiring for any kind of POIs, either including or nor addr:housenumber

Deeptanshu-sankhwar:
Hi @jjiglesiasg, thanks for sharing the issue. I have added `addr:housename` to the address formats for buildings in Bolivia. But I am not sure which all countries would align to the same problem, so I can update their address schemas as well. I would appreciate any insight into fetching list of such countries.

Meanwhile, this is how the preview comes to be for any building in Bolivia, in the PR above.

![Image](https://github.com/user-attachments/assets/859ed888-dc52-452a-bd1b-4f7335f3121e)

tyrasd:
@jjiglesiasg I only found a small number of objects in OSM in Bolivia which have `addr:housename` but no housenumber: https://overpass-turbo.eu/s/1Y9J

I assume this is mostly because addresses are not yet well mapped in the country, but it would still be good to see some examples of how such "housename" addresses of buildings look in reality, and ideally some hints towards that these house names are used as addresses. Thanks!

jjiglesiasg:
![Image](https://github.com/user-attachments/assets/759a43d3-f1cf-448f-a679-402c7e22cee8)
![Image](https://github.com/user-attachments/assets/730748c0-3dae-4e7a-82f9-4595a86531d9)
![Image](https://github.com/user-attachments/assets/29da0091-d6c1-4ef2-ad35-54bec43e21d9)
![Image](https://github.com/user-attachments/assets/cae53a92-f6d6-4974-8a72-beb13da6e869)
![Image](https://github.com/user-attachments/assets/2741eb25-96d5-4e1d-8bef-8e705dfc718a)
![Image](https://github.com/user-attachments/assets/748609d4-e896-46d7-a6d0-11553157d977)
![Image](https://github.com/user-attachments/assets/1fc95edd-61a7-4fc0-9245-c4379f0d12b2)
[...]

PR #11698, opened 2025-12-16
Added Morocco to address_formats.json

ilias52730:
In Morocco, most buildings have house names. A developer will call their building "Résidence Soraya" for exemple (see image). Similar issue was fixed for Bolivia (#10727).

image

Common practice currently is to add these names to the `name=*` tag. This happens because there are no clear guidelines, but also because the `addr:housename` field is missing from the default preset. Mappers see a prominent "Name" field and logically fill it in, even though `addr:housename` is the correct tag for the address data.

I made some extensive research about address formats in Morocco. I looked through documents from the country's postal service and the Universal Postal Union. I was surprised there's actually a very precise addressing scheme, detailed on [this document](https://codepostal.ma/annuaire.pdf) and [this one](https://www.upu.int/UPU/media/upu/PostalEntitiesFiles/addressingUnit/marEn.pdf):

```
Unit
House name
House number and street
Extra info (Lieu-dit/Place) if necessary and/or if no street
Postcode and City
```

An example of an apartment's address in an apartment building would be:

```
Appartement 4
Résidence Soraya
15 Rue Al Fourate
20000 Casablanca
```
So I came up with this preset for Morocco:
```
    "format": [
      ["housename"],
      ["housenumber", "street+place"],
      ["postcode", "city"]
    ]
```
- For mapping in Morocco (and in most countries), a `unit` field would be clutter for the majority of use cases (building polygons or ground-floor POIs). If a specific office in a building needs to be mapped, the user can manually add `addr:unit` and `addr:floor`.

- `street+place` is consistent with other presets. In many parts of Morocco, houses have numbers but no street names. This allows users to input the neighborhood/hamlet/village name (Lieu-dit) in the street/place field, which the editor can then change to `addr:place`.

I plan on documenting this on Morocco's wiki, change all past buildings I mapped from `name` to `addr:housename`, and encourage others to do so in the country.

### DIFF
--- a/data/address_formats.json
+++ b/data/address_formats.json
@@ -71,6 +71,14 @@
       ["postcode", "city"]
     ]
   },
+{
+    "countryCodes": ["ma"],
+    "format": [
+      ["housename"],
+      ["housenumber", "street+place"],
+      ["postcode", "city"]
+    ]
+  }
   {
     "countryCodes": ["mo"],
     "format": [
--- a/data/address_formats.json
+++ b/data/address_formats.json
@@ -71,14 +71,14 @@
       ["postcode", "city"]
     ]
   },
-{
+  {
     "countryCodes": ["ma"],
     "format": [
       ["housename"],
       ["housenumber", "street+place"],
       ["postcode", "city"]
     ]
-  }
+  },
   {
     "countryCodes": ["mo"],
     "format": [


tyrasd (approved):
Notably, `addr:housename` is only used only about 100 times in Morocco in OSM today ([source](https://taginfo.geofabrik.de/africa:morocco/search?q=addr%3A)).

But the documents you linked seem to be quite clear that the house name can be a part of the address in Morocco, so :+1: from my side.

ilias52730:
> Notably, `addr:housename` is only used only about 100 times in Morocco in OSM today ([source](https://taginfo.geofabrik.de/africa:morocco/search?q=addr%3A)).

Time for change! I’ll make sure to document it and communicate it to my fellow contributors. Thanks a lot.

ilias52730:
Any idea when the next release is scheduled, more or less?

tyrasd:
Early January most likely.

PR #11699, opened 2025-12-17
Now multiCombo Languages will show human readable languages instead of ISO codes

### LINKED ISSUE #11652: Languages field should be human-readable
The [Languages](https://github.com/openstreetmap/id-tagging-schema/blob/2475120b86a706bf2cbc5edefc3c914b1587e0e3/data/fields/language_multi.json) multicombo field for `language:*=*` displays raw ISO&nbsp;639 codes, but the user may not know the language code for the languages associated with the POI they’re mapping, especially if it’s a language they don’t speak. The field should display the human-readable name of each option.

Languages: vi

Similar to #7123, we have an opportunity to use [`Intl.DisplayNames.prototype.of()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/of) to come up with all these names without adding any additional burden on translators. Alternatively, we already pull in CLDR language names with [many manual overrides and extensions](https://github.com/openstreetmap/iD/blob/77a9c71f9d191ef3ce8ee0e625ace2dd88da37ad/scripts/language_names.js#L28) for the Multilingual Names field and could use that instead.

Razen04:
Fixes #11652 

## Changes made:
- `displayValue` function now looks at the field.key and if it is `language:`, it will use `localizer.languageName` on `tval`.
- `renderValue` function will also look at the field.key and it does the same like `displayName`
- The `comboData` array will now have `value=v` so that choosing 'Hindi' will not make tag as `language:Hindi=yes` and instead it will be `language:hi=yes`.

Before can be seen on the issue, here is the after image:

Screenshot 2025-12-17 at 8 52 55 PM

Btw I have a doubt, when we click on the dropdown we see some language only like Hindi is not present but when I type hi, I can see Hindi as an option and can set it, is it intentional?

### DIFF
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -115,11 +115,12 @@ export function uiFieldCombo(field, context) {
     function displayValue(tval) {
         tval = tval || '';
         // Issue #11652
-        if (field.key === 'language:'){
+        // Ignore language: key and when tval is not others
+        if (field.key === 'language:' && tval !== 'others'){
           let langName = localizer.languageName(tval);
           if (langName) {
             return langName;
-          };
+          }
         }
 
         var stringsField = field.resolveReference('stringsCrossReference');
@@ -142,7 +143,8 @@ export function uiFieldCombo(field, context) {
         tval = tval || '';
 
         // Issue #11652
-        if (field.key === 'language:'){
+        // Ignore language: key and when tval is not others
+        if (field.key === 'language:' && tval !== 'others'){
           let langName = localizer.languageName(tval);
           if (langName) return selection => selection.text(langName);
         }
@@ -190,6 +192,37 @@ export function uiFieldCombo(field, context) {
     }
 
     function getOptions(allOptions) {
+        // Get dropdown list for language: key via localizer instead of taginfo
+        if (field.key === 'language:') {
+          let codes = Object.keys(localizer.languages());
+
+          let options = codes.map((c) => {
+            let name = localizer.languageName(c);
+            return {
+              key: c,
+              value: name,
+              title: name,
+              display: selection => selection.text(name)
+            };
+          }).filter(Boolean);
+
+          options.sort((a, b) => {
+            return a.value.localeCompare(b.value);
+          });
+
+          // inserting others because it does not come via _dataLanguages
+          options.push({
+            key: 'others',
+            value: 'others',
+            title: 'others',
+            display: selection => selection.text('others'),
+            klass: 'raw-option'
+          });
+
+
+          return options;
+        }
+
         var stringsField = field.resolveReference('stringsCrossReference');
         if (!(field.options || stringsField.options)) return [];
 
@@ -241,6 +274,9 @@ export function uiFieldCombo(field, context) {
         var queryFilter = d => d.value.toLowerCase().includes(q.toLowerCase()) || d.key.toLowerCase().includes(q.toLowerCase());
         if (hasStaticValues()) {
             setStaticValues(callback, queryFilter);
+
+            // If it is language field, we don't need to request for values, we get it from getOptions
+            if (field.key === 'language:') return;
         }
 
         var stringsField = field.resolveReference('stringsCrossReference');
@@ -302,17 +338,9 @@ export function uiFieldCombo(field, context) {
                 const labelId = getLabelId(stringsField, v);
                 var isLocalizable = stringsField.hasTextForStringId(labelId);
                 var label = stringsField.t(labelId, { default: v });
-                // If this is a language field, force the label to be the human readable name
-                if (['language:'].includes(field.key)) {
-                    let langName = localizer.languageName(v);
-                    if (langName) {
-                        label = langName;
-                        isLocalizable = true;
-                    }
-                }
                 return {
                     key: v,
-                    value: v,
+                    value: label,
                     title: stringsField.t(`options.${v}.description`, { default:
                         isLocalizable ? label : (d.title !== label ? d.title : '') }),
                     display: addComboboxIcons(stringsField.t.append(labelId, { default: label }), v),
@@ -737,6 +765,8 @@ export function uiFieldCombo(field, context) {
                 .classed('raw-value', function(d) {
                     var k = d.key;
                     if (_isMulti) k = k.replace(field.key, '');
+                    // Ignore the raw-value class for key language:
+                    if (field.key === 'language:' && localizer.languageName(k) && k !== 'others') return false;
                     return !stringsField.hasTextForStringId('options.' + k);
                 })
                 .classed('draggable', allowDragAndDrop)
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -4,7 +4,7 @@ import { drag as d3_drag } from 'd3-drag';
 import * as countryCoder from '@rapideditor/country-coder';
 
 import { fileFetcher } from '../../core/file_fetcher';
-import { t } from '../../core/localizer';
+import { localizer, t } from '../../core/localizer';
 import { services } from '../../services';
 import { uiCombobox } from '../combobox';
 import { svgIcon } from '../../svg/icon';
@@ -114,6 +114,13 @@ export function uiFieldCombo(field, context) {
     // (for multiCombo, tval should be the key suffix, not the entire key)
     function displayValue(tval) {
         tval = tval || '';
+        // Issue #11652
+        if (field.key === 'language:'){
+          let langName = localizer.languageName(tval);
+          if (langName) {
+            return langName;
+          };
+        }
 
         var stringsField = field.resolveReference('stringsCrossReference');
         const labelId = getLabelId(stringsField, tval);
@@ -133,6 +140,12 @@ export function uiFieldCombo(field, context) {
     // (for multiCombo, tval should be the key suffix, not the entire key)
     function renderValue(tval) {
         tval = tval || '';
+
+        // Issue #11652
+        if (field.key === 'language:'){
+          let langName = localizer.languageName(tval);
+          if (langName) return selection => selection.text(langName);
+        }
 
         var stringsField = field.resolveReference('stringsCrossReference');
         const labelId = getLabelId(stringsField, tval);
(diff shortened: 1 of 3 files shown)


modules/ui/fields/combo.js
```diff
@@ -114,6 +114,13 @@ export function uiFieldCombo(field, context) {
     // (for multiCombo, tval should be the key suffix, not the entire key)
     function displayValue(tval) {
         tval = tval || '';
+        // Issue #11652
+        if (field.key === 'language:'){
+          let langName = localizer.languageName(tval);
+          if (langName) {
+            return langName;
+          };
```
1ec5: Nit: stray semicolon.

modules/ui/fields/combo.js
```diff
@@ -114,6 +114,13 @@ export function uiFieldCombo(field, context) {
     // (for multiCombo, tval should be the key suffix, not the entire key)
     function displayValue(tval) {
         tval = tval || '';
+        // Issue #11652
+        if (field.key === 'language:'){
+          let langName = localizer.languageName(tval);
```
1ec5, line 120: [`language:others=*`](https://taginfo.openstreetmap.org/keys/language%3Aothers) is a very common key that doesn’t correspond to an ISO language code. It’s a special subkey that indicates whether the set of `language:*=*` tags is complete or not. It currently appears as “others”, unlocalized but styled as if it’s localized:

others in blue.

I think we should special-case this key to fall through to the original code so that id-tagging-schema can provide a localized string for it: openstreetmap/id-tagging-schema#1860.

modules/ui/fields/combo.js
1ec5, line 1: > Btw I have a doubt, when we click on the dropdown we see some language only like Hindi is not present but when I type hi, I can see Hindi as an option and can set it, is it intentional?

By default, a multicombo field’s dropdown only shows the most common subkeys, [as reported by taginfo](https://taginfo.openstreetmap.org/search?q=language%3A). This is appropriate for most of the multicombo fields, but the set of valid language codes is determined by an ISO standard rather than the OSM mapping community.

Let’s add a special case to `getOptions()` that lists all the languages we know about. This would enable us to sort them alphabetically by the interface language, so that the user can find a language more easily. Since you’re using localizer to come up with these language names, you could create a getter for its `_dataLanguages` object and have `getOptions()` return its keys.

modules/ui/fields/combo.js
```diff
@@ -289,12 +302,20 @@ export function uiFieldCombo(field, context) {
...
+                    let langName = localizer.languageName(v);
+                    if (langName) {
+                        label = langName;
+                        isLocalizable = true;
+                    }
+                }
                 return {
                     key: v,
-                    value: label,
+                    value: v,
```
1ec5: Due to this change, you can search for the language code “hi” and get Hindi, but you can’t search for “hindi” and get Hindi. This also affects other combo and multicombo boxes. Some users will know the raw code but others (especially non-English speakers) will only know the localized display name. We want the user to be able to search for either. I’m pretty sure the underlying control’s implementation already knows how to search by the `key`, so `value` should be `label`.

modules/ui/fields/combo.js
1ec5, line 1: Not sure why, but the chips are showing up in monospaced font instead of variable-width font, like you see in the Payment field:

Tiếng Tây Ban Nha in monospaced font.
Visa in variable-width font.

1ec5 (commented):
This is a good start, thank you!

Razen04:
Sure I will work on the changes and let you know if I face any problem in between. These are the things I have understand I need to work on:

Things to do:
- [x] Ignore the language:others basically ignore the tval=others and not let it pass through the languageName().
- [x] Display all the languages we know about instead of getting the info via taginfo and also sort it alphabetically on the basis of it's display value.
- [x] Fix searching.
- [x] Fix monospace chip styling.  

Do let me know if there is something I missed.

modules/ui/fields/combo.js
Razen04: Since others is not localized and is a raw-option, should we show that in monospaced font or it should too be variable-width font?

1ec5: Anything that’s displayed raw should continue to be monospaced in black, while anything that’s localized should be variable-width in blue. There’s something controlling the font and color based on whether the control thinks it’s a raw value or not. It might have something to do with that change of `value` from `label` to `v`, but I haven’t experimented with it.

Razen04:
I was able to do these changes, do let me know if there is something wrong.

modules/ui/fields/combo.js
Razen04: Yeah the `raw-option` class was responsible for it

modules/ui/fields/combo.js
```diff
@@ -190,6 +192,37 @@ export function uiFieldCombo(field, context) {
...
+
+          options.sort((a, b) => {
+            return a.value.localeCompare(b.value);
+          });
+
+          // inserting others because it does not come via _dataLanguages
+          options.push({
+            key: 'others',
+            value: 'others',
+            title: 'others',
```
1ec5: We should allow id-tagging-schema to provide its own localized string for the title of this option. Can we make the initialization of this option more consistent with the generic initialization down below?

modules/ui/fields/combo.js
```diff
@@ -737,6 +765,8 @@ export function uiFieldCombo(field, context) {
                 .classed('raw-value', function(d) {
                     var k = d.key;
                     if (_isMulti) k = k.replace(field.key, '');
+                    // Ignore the raw-value class for key language:
+                    if (field.key === 'language:' && localizer.languageName(k) && k !== 'others') return false;
```
1ec5: This is fine for now, but we’ll probably want to find a more data-driven approach in the future, when we enhance similar fields like Payment Methods, Currency, and Country (#7123).

modules/ui/fields/combo.js
```diff
@@ -190,6 +192,37 @@ export function uiFieldCombo(field, context) {
...
+              key: c,
+              value: name,
+              title: name,
+              display: selection => selection.text(name)
+            };
+          }).filter(Boolean);
+
+          options.sort((a, b) => {
+            return a.value.localeCompare(b.value);
+          });
```
1ec5, line 217: `String.prototype.localeCompare` uses the browser’s language preference, but there are a few other ways for the user to customize iD’s interface language, such as setting `?locale=` in the URL. Pass in the locale code to ensure consistency. (Given the number of items to sort, best to store the locale outside of the `sort()` call.)

https://github.com/openstreetmap/iD/blob/aaaa972fd31834315f64b97bc8ceac60c3bdeaa7/modules/ui/sections/preset_fields.js#L88-L90

modules/ui/fields/combo.js
```diff
@@ -177,6 +192,37 @@ export function uiFieldCombo(field, context) {
     }
 
     function getOptions(allOptions) {
+        // Get dropdown list for language: key via localizer instead of taginfo
+        if (field.key === 'language:') {
+          let codes = Object.keys(localizer.languages());
+
+          let options = codes.map((c) => {
+            let name = localizer.languageName(c);
```
1ec5, line 201: There’s a bunch of obscure languages for which the menu now shows a language code because we don’t know the language name. These should be styled as raw values. We might want to omit them from the dropdown menu so that more common languages are easier to access.

modules/ui/fields/combo.js
Razen04: Yes, we will have to come up with something like specialKey boolean for payment, currency, language, and country or any others, that will be a better long-term solution for these type of keys.

Razen04:
Ok I will maintain consistency and apply the changes accordingly.

modules/ui/fields/combo.js
Razen04: Basically we need others to get localized in different languages, I think this can be done if we use stringsField to fill the 'others' option and push in the `options` array.

modules/ui/fields/combo.js
Razen04: I omitted them from the dropdown, is that okay?

Btw some names at the bottom of the dropdown are not in English, are those supposed to be like that or am I doing something wrong there?

Razen04:
I have made the changes, do let me know if somethings wrong.

modules/ui/fields/combo.js
```diff
@@ -716,6 +773,8 @@ export function uiFieldCombo(field, context) {
                 .classed('raw-value', function(d) {
                     var k = d.key;
                     if (_isMulti) k = k.replace(field.key, '');
+                    // Ignore the raw-value class for key language:
+                    if (field.key === 'language:' && localizer.languageName(k) && k !== 'others') return false;
```
1ec5: I entered a valid but rare ISO&nbsp;639 language code that iD doesn’t know about. It should appear in monospaced font, similar to `others`, but instead it appears in blue as if it’s localized. Would it be feasible to style unrecognized values as before?

cst in blue variable-width font.

modules/ui/fields/combo.js
1ec5: These names are coming from [languages.json](https://github.com/openstreetmap/iD/blob/1a86bc5f1765b3a784218bcaf28345663b56af07/data/languages.json). This is where we add languages that CLDR doesn’t know about, but that we want to show in the Multilingual Names field anyways. 

I think it’s fine. Ideally we’d make all these additional languages’ names translatable, but that’s a big effort for relatively obscure languages.

1ec5 (commented):
There’s one more styling issue. Once that’s fixed, it’ll be good enough to land.

modules/ui/fields/combo.js
Razen04: yeah i will do that quickly

Razen04: Screenshot 2025-12-22 at 1 31 19 PM

Ok?

1ec5 (approved):
Looks great, thank you for persisting through multiple iterations! In case you’re interested, #7123 would be a good opportunity for you to expand upon what you’ve built here. I haven’t opened an issue about the Currency field yet, but it’s the same idea there.

Razen04:
@1ec5 Great! I will look into the issue you mentioned, btw can you also look at some other PR's I have made if possible like: #11667 , #11691 if you get time. Thanks for the guidance.

PR #11700, opened 2025-12-17
Fix language initialization for Firefox when quality is present on the requested locales

dobladov:
Firefox returns locales like "en-US;q=0.5" in Accept-Language header which are not valid [BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag), the only valid values for `Intl.Locale` constructor.

This causes it to throw an exception [at this point](https://github.com/openstreetmap/iD/blob/490c86fa3985deee78cc541c7c1e3c3afd56b108/modules/core/localizer.js#L155) skipping the execution of [updateForCurrentLocale](https://github.com/openstreetmap/iD/blob/490c86fa3985deee78cc541c7c1e3c3afd56b108/modules/core/localizer.js#L134)

Screenshot 2025-12-17 at 19 18 13

This solution, strips quality values before passing locales to `Intl.Locale` constructor and fallbacks to the legacy logic in case of an exception.

Before
Screenshot 2025-12-17 at 18 54 31

After
Screenshot 2025-12-17 at 18 54 41

The test have been extended to cover cases of locales with quality [localizer.js](https://github.com/openstreetmap/iD/pull/11700/changes#diff-193e1de822f0df63bd5288d4c9debc093499286899eadf214bfee818c6f81740)

Related issue: https://community.openstreetmap.org/t/edit-map-with-id-dont-work/101362

### DIFF
--- a/modules/core/localizer.js
+++ b/modules/core/localizer.js
@@ -148,19 +148,21 @@ export function coreLocalizer() {
         /** @type {string[]} */
         let toUse = [];
         for (const locale of requestedLocales) {
-            if (supportedLocales[locale]) toUse.push(locale);
+            const safeLocale = locale.split(';')[0].trim(); // strip any quality values like ";q=0.5"
 
-            if ('Intl' in window && 'Locale' in window.Intl) {
+            try {
+              if ('Intl' in window && 'Locale' in window.Intl) {
                 // Full locale ('es-ES'), add fallback to the base ('es')
-                const localeObj = new Intl.Locale(locale);
+                const localeObj = new Intl.Locale(safeLocale);
                 const withoutScript = `${localeObj.language}-${localeObj.region}`;
                 const base = localeObj.language;
 
                 if (supportedLocales[withoutScript]) toUse.push(withoutScript);
                 if (supportedLocales[base]) toUse.push(base);
-            } else if (locale.includes('-')) {
-                // legacy logic: if Intl.Locale is not available
-                let langPart = locale.split('-')[0];
+              }
+            } catch {
+                // legacy logic: if Intl.Locale is not available or fails
+                let langPart = safeLocale.split('-')[0];
                 if (supportedLocales[langPart]) toUse.push(langPart);
             }
         }
--- a/test/spec/core/localizer.js
+++ b/test/spec/core/localizer.js
@@ -110,6 +110,12 @@ describe('iD.coreLocalizer', function() {
             [['zh-Hans-CN'], ['zh-CN', 'zh', 'en']],
             [['zh-Hans'], ['zh', 'en']],
             [['fr-Latn'], ['fr', 'en']],
+            // Firefox-style locales with quality values (see #11700)
+            [['en;q=0.5'], ['en']],
+            [['en-AU;q=0.9'], ['en-AU', 'en']],
+            [['zh-CN;q=0.8'], ['zh-CN', 'zh', 'en']],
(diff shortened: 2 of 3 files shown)


modules/core/localizer.js
```diff
@@ -148,19 +148,21 @@ export function coreLocalizer() {
         /** @type {string[]} */
         let toUse = [];
         for (const locale of requestedLocales) {
-            if (supportedLocales[locale]) toUse.push(locale);
+            const safeLocale = locale.split(';')[0].trim(); // strip any quality values like ";q=0.5"
```
1ec5, line 151: Is there some way to fold this into where we process the preference, so other code doesn’t behave inconsistently?

https://github.com/openstreetmap/iD/blob/13b96eb391068c9532b5eca18f67244a1315a8cc/modules/core/localizer.js#L72-L73

test/spec/core/localizer.js
```diff
@@ -110,6 +110,12 @@ describe('iD.coreLocalizer', function() {
             [['zh-Hans-CN'], ['zh-CN', 'zh', 'en']],
             [['zh-Hans'], ['zh', 'en']],
             [['fr-Latn'], ['fr', 'en']],
+            // Firefox-style locales with quality values (see #11700)
+            [['en;q=0.5'], ['en']],
+            [['en-AU;q=0.9'], ['en-AU', 'en']],
+            [['zh-CN;q=0.8'], ['zh-CN', 'zh', 'en']],
+            [['fr;q=0.7', 'en;q=0.3'], ['fr', 'en']],
+            [['zh-Hans-CN;q=1.0'], ['zh-CN', 'zh', 'en']],
```
1ec5, line 118: We’re using [`navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages), not accessing the raw `Accept-Languages` header:

https://github.com/openstreetmap/iD/blob/490c86fa3985deee78cc541c7c1e3c3afd56b108/modules/util/detect.js#L90-L91

Are you sure Firefox includes quality values in `navigator.languages`? I suppose it’s always possible to set that manually in about:config, but how do you do that in the UI? I currently have multiple languages set in my preferences:

vi, en-US, en, es

This results in:

* [`intl.accept_languages`](https://kb.mozillazine.org/About:config_entries#Intl.) setting in about:config: `vi,en-us,en,es`
* `Accept-Language` header: `vi,en-US;q=0.8,en;q=0.5,es;q=0.3`
* `navigator.languages`: `["vi", "en-US", "en", "es"]`

If this is a case that the user can only encounter if they manually set `intl.accept_languages` to an invalid format, then I’m not sure it’s a use case we should accommodate in iD.

dobladov: This is a good point, I don't remember modifying it in`about:config`, but I can see that it is set.

Screenshot 2025-12-17 at 20 51 14

Screenshot 2025-12-17 at 20 53 12

Could be that I modified it years ago due to this issue: https://unix.stackexchange.com/questions/127803/firefox-29-default-accept-language-bug

If that is the case, we can close the PR, I don't think we should sanitize it in this case.

1ec5: Fascinating! Yes, between the [Mozilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1005640) and [Red Hat](https://bugzilla.redhat.com/show_bug.cgi?id=1514536) tickets, it appears to be a long-gone issue. Hopefully not too many other users are similarly affected. If we find a similar edge case in the future, we can revive your contribution. Regardless, thank you for taking the time to explore a fix here. ❤️

Issue #7498, opened 2020-04-02
Route relation member sorting


tomasokol:
Hi, 
as Wiki says, a Bus route should have relation members in this order: 
1) bus stops (in the direction the bus is going)
2) roads (in the direction of the bus)

So iD should keep bus stops at the top of the list. That should be relatively easy, i suppose. 
Options are: 
1) still just one “window” for all members;
2) splitting the existing “window” into two, one containing stops, the other containing roads.

It would be really nice to have the stops ordered automatically by the iD. If it could analyze the route and sort the bus stops that sit along. And again, two options:
1) supporting all types of mapping (the simple one, with just public_transport=platform beside the road;
2) supporting only routes that are mapped by public_transport=platform along with public_transport=stop_position.

Also, there is a name for route=bus “Bus Route.” But there is none for “route=trolleybus,” as it is just a “Route” in iD.

quincylvania:
> Also, there is a name for route=bus “Bus Route.” But there is none for “route=trolleybus,” as it is just a “Route” in iD.

Okay, I added a Trolleybus Route preset.

> It would be really nice to have the stops ordered automatically by the iD.

iD could definitely be smarter about placing new members when adding them to route relations. But I'm not sure if it should be reordering existing members or not.

tomasokol:
> But I'm not sure if it should be reordering existing members or not.

Validator could suggest, that lines are not in order (there is a gap / members are not in the right order). If it should be automatically sorted, I don't know.

Though it could be automatically putting all the stops and stopping positions above lines/roads, in the list. I don't see any situation when somebody would not want that to happen or wanted to keep it sorted differently.

HarelM:
I have a request to support bike route relations (and even super relations).
I did add support for showing those in my sure, but the editors are mostly using iD for editing, and they have a hard time sorting relations of bike routes (which are similar to bus routes in that aspect).
It would be great if there was a button to allow sorting ways in relation, much like JOSM.
I'm having a hard time convincing editors to move from iD to JSOM (I also prefer iD over JSOM so I can relate), so having this functionality in iD would be a great a addition to these editors.
Here's an example of super relation bike route:
https://www.openstreetmap.org/relation/19598237
As you can see, it has some subroutes and they have a lot of ways in then. 
The ways are sometimes not sorted out, mainly because there's not an easy way to do it as part of iD.

If this is something I can contribute, I'll be happy to do so, but I wouldn't know where to start, so I'll need some guidance.
[...]

Issue #5821, opened 2019-01-28
Optional UI sound effects


quincylvania:
Many apps use subtle sound effects to reinforce their interfaces and give audial feedback. Messaging apps bloop upon sending, video and sound editing apps click when snapping clips end-to-end, and various apps thunk upon errors. Games widely use sound effects for all sorts of input like button presses and focus changes.

There are various places in iD where I think minor, tasteful sound effects could create a richer editing experience and help the vision-impaired.

- Click upon snapping to a node or way when moving or drawing vertices
- Snip upon splitting a way
- Error thunk upon trying to create a self-intersecting way
- Selection sound upon picking a preset
- Cheerful completion sound upon successful changeset upload
- More…

Of course users would be able to mute sounds at any time.

mmd-osm:
Yeah, back in the days, Potlatch 0.10c had a nifty beep sound when saving a changeset. That was like 11 years ago. 

And it's still around ~~today~~: https://github.com/openstreetmap/openstreetmap-website/blob/7bd1cda/public/potlatch/beep.mp3

https://github.com/openstreetmap/potlatch/blame/40b208026d0fe3261856c0fb5cba4266dbd1147c/potlatch.as#L40-L42

antimundo:
I think UI sounds can give a lot of feedback to the user and clarify what is happening, but if something like this was implemented, I would do add it with the purpose of been useful and provide feedback while trying not be very intrusive, I don't like the approach of adding sounds that makes the editor sound like a video game.

Also if implemented, these sound effects should be optional.

PR #11701, opened 2025-12-17
Improved rendering/handling of relation members 

tyrasd:
Several improvements related to how one can interact with relation members:

(For the screenshots below: on the left is new UI, right is the previous one.)

- more space for member label by putting buttons next to role input and add a "grab handle pattern" to indicate drag&drop functionality

 

- add a button to download all members of a relation also from the relation members list (previously this was only available from the relation _membership_ list of the relation's members)

- if members are connecting ways, the white space between the consecutive members is collapsed and the grab icon is replaced with arrows indicating the connection:

 

- this can make gaps in a route more easily visible
 

 
- The icons indicate the connectivity in the respective direction for members with `forward/backward` roles in more complex relations

- allow merging additional members onto a selected relation (works on a multiselection of one relation and 1+ other features: first select the relation, then shift-select the members to add)

- add operation to remove members from a relation (see above, but only works on a multiselection that contains only the relations and features that are already its members)

- slightly more descriptive tooltip texts for the "merge" operation depending on what is going to be merged.

## todo:

* [ ] unit tests for the connectivity-checking code


svg/iD-sprite/icons/icon-grab-connects-both.svg
1ec5, line 1: It took me a bunch of squinting at your examples before realizing that the pattern consists of down arrows on one side and up arrows on the other side. I thought this was just a grippy pattern, similar to icon-grab.svg.

modules/ui/sections/raw_member_editor.js
```diff
@@ -148,29 +271,65 @@ export function uiSectionRawMemberEditor(context) {
...
+                        if (d.connections.prev && d.connections.next || d.connections.loops) {
+                            d3_select(this).call(svgIcon('#iD-icon-grab-connects-both'));
+                        } else if (d.connections.prev) {
+                            d3_select(this).call(svgIcon('#iD-icon-grab-connects-prev'));
+                        } else if (d.connections.next) {
+                            d3_select(this).call(svgIcon('#iD-icon-grab-connects-next'));
+                        } else {
+                            d3_select(this).call(svgIcon('#iD-icon-grab'));
+                        }
+                    });
```
1ec5, line 321: Showing the connections is a good idea, but I really don’t find the backwards/forwards distinction intuitive at all. It isn’t clear enough to me that the grippies are _supposed_ to connect, probably because the labels get in the way.

In case the grippy patterns don’t work out, here are some alternative ideas:

* Keep the gap between each member. Insert the arrows in the gap. Increase the gap and insert some obvious indicator wherever there are discontinuities.
* Keep the arrow indicators where they are, but remove the handles. Make it more like JOSM’s “model railroad” schematic, with lines running continuously up and down the section, but ideally without the gaps that JOSM has.
* Combine all the handles for the connected members into a single handle. Dragging that handle moves all the members together, sort of like in a jigsaw puzzle.
* Give each row a jigsaw puzzle piece shape. The pieces fit together when they’re connected on the map. They don’t fit if there’s a gap or the roles conflict.

modules/actions/merge_members.js
```diff
@@ -0,0 +1,53 @@
...
+
+        return graph;
+    };
+
+
+    action.disabled = function(graph) {
+        const relationCount = entityIDs.filter(entityID =>
+            graph.entity(entityID).type === 'relation')
+            .length;
+        if (relationCount !== 1) return 'not_eligible';
```
1ec5, line 47: The Merge action remains enabled even when the Remove Members action is enabled, so the context menu starts out with ➖ followed by ➕. Do we need to have both actions enabled at the same time? Users shouldn’t need to add redundant memberships very often, but when they do have to, they can still add it the old-fashioned way with the way selected.

data/core.yaml
```diff
@@ -328,6 +333,12 @@ en:
       conflicting_relations: These features can't be merged because they belong to conflicting relations.
       paths_intersect: These features can't be merged because the resulting path would intersect itself.
       too_many_vertices: These features can't be merged because the resulting path would have too many points.
+    members_remove:
+      title: Remove Members
```
1ec5, line 337: Remove Members doesn’t look like an operation that undoes what Merge does. How about Unmerge? Or maybe make it look like a version of the Detach operation.

modules/actions/members_remove.js
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+        let memberIndices = [];
+        for (let i = 0; i < relation.members.length; i++) {
+            if (entityIDs.indexOf(relation.members[i].id) > 0) {
+                memberIndices.push(i);
+            }
+        }
+
+        graph = actionDeleteMembers(relation.id, memberIndices)(graph);
```
1ec5, line 30: If a feature is a member of the relation multiple times, this operation completely removes all its memberships from the relation. This makes it less useful for dealing with incorrect redundant memberships, though it would make more sense if the operation were called something like Detach.

1ec5 (commented):
The redesign of the member/membership list adds important functionality, but I’m concerned that non-expert users will experience some difficulty learning how it works on their own. Additionally, consider some additions to the Relations section of the help documentation.

So far, this PR focuses on the sidebar, but the map is still underutilized. The map could highlight the endpoints of each gap and any locations where the member ways’ directions collide.

modules/ui/sections/raw_membership_editor.js
```diff
@@ -495,6 +469,36 @@ export function uiSectionRawMembershipEditor(context) {
...
+            .classed('form-field-button', true)
+            .attr('title', t('icons.download'))
+            .call(svgIcon('#iD-icon-load'))
+            .on('click', downloadMembers);
+
+        items.selectAll('button.members-download')
+            .classed('hide', d => {
+                const graph = context.graph();
+                return d.relation.members.every(m => graph.hasEntity(m.id));
+            });
```
1ec5, line 513: Regardless of the handle design, it’s still quite tedious to fix the order in a relation that’s all jumbled up. A button to sort the whole relation would give the user a good starting point to refine further: https://github.com/openstreetmap/iD/issues/7498#issuecomment-3478076945.

modules/ui/sections/raw_member_editor.js
1ec5: If we do make the rows look like jigsaw puzzle pieces, they could make a subtle but satisfying click when you snap together pieces that had been disconnected: #5821. 😏

svg/iD-sprite/icons/icon-grab-connects-both.svg
matkoniecz: +1, my brain interpreted them as decorational pattern. Maybe fewer larger ones can be viable? Like JOSM has?

emil813-cmd:
There should also be an indication for when two adjacent member entries represent _the same_ object.

The Vespucci app already shows when adjacent members are linked. Consider looking at its implementation for inspiration.

Relation types are very different from one another. Gap-finding and merging may mean different things for different types, and there's already some sort of Merge➕ defined for Areas and Multipolygons, which _might_ conflict with the operation that you are suggesting.

- Instead of "Merge➕", perhaps another word like "Add⭐" should be used for simple membership addition.

The **Issues** menu warns when a Multipolygon Relation has gaps. It could be enhanced to offer to sort the member entries and/or to try to fill the gaps.

In addition to Error🟥 and Warning🟨 issues, one could add green "Suggestion🟩" notices that offer sorting and other functionality even when there's nothing strictly wrong with the selected object. Their functionality should probably depend on the Relation type. (E.g. Waterway Relations may have _side_stream_ members that could really mess with sorting). But some sort of default implementation might work for most types.

Question: Are there currently any Relation types that assign meaning to the _order_ of membership entries?

PR #11707, opened 2025-12-18
Adjust housenumber and floor widths for Taiwan addresses

### LINKED ISSUE #11622: Floor field too wide, housenumber too narrow

Priyanka-Kandelkar:
Adjusted housenumber and floor field widths for Taiwan addresses to prevent housenumber truncation and better match real-world usage.

Closes #11692

### DIFF
--- a/data/address_formats.json
+++ b/data/address_formats.json
@@ -132,7 +132,16 @@
       ["postcode", "city", "district"],
       ["street+place"],
       ["housenumber", "floor", "unit"]
-    ]
+    ],
+      "widths": {
+          "postcode": 0.3,
+          "city": 0.5,
+          "district": 0.5,
+          "street+place": 1,
+          "housenumber": 0.7,
+          "floor": 0.2,
+          "unit": 0.2
+      }
   },
   {
     "countryCodes": ["jp"],


1ec5:
> Closes https://github.com/openstreetmap/iD/pull/11692

#11692 is a different pull request attempting to fix the same bug, #11622. This pull request is a bit more extensive, so there are some visual differences compared to #11692. We might need feedback from a Taiwan mapper such as @jidanni about which layout would be better.

#11692 | #11707
----|----
 |

jidanni:
Maybe @Supaplextw also has opinions. All I'm concerned about is allow people to write 15A, etc. too if needed.

jidanni:
Maybe @Supaplextw also has opinions. All I'm concerned about is allow people to write 15A, etc. too if needed.

jidanni:
( @1ec5 actually in this area users shouldn't be messing with addresses in the first place. https://github.com/openstreetmap/iD/issues/11708 .)

tyrasd:
b06496780f555b23afed0b8996afc1f2b2ba9041 has a more pragmatic approach to fix this: `addr:floor` contains similarly short values like `addr:housenumber` and ?`addr:unit`, thus it should always be rendered equally narrow like those other subfields. There's no need to add a manual exception for Taiwan for this field. Instead, this commit added a proper default width for `addr:floor`.

PR #11713, opened 2025-12-20
Fix inconsistent squaring between ways and nodes

### LINKED ISSUE #11580: Square a way and square the nodes of that way don't give the same result
### URL

https://master.apis.dev.openstreetmap.org/edit?way=4307298815#map=19/-21.349230/55.562029

1. follow the link above
2. square the building : result is bad
3. cancel the square, select the nodes of the way (with, from the selected way, the keys CTRL + the downward arrow)
4. on the top right node, right click and square all the selected nodes : result is perfect

If should be great to have the same result when we select the way than when we select all the nodes.

_No response_

2.37.3

_No response_

adityayadav-dev:
Fixes #11580

### Problem
When squaring a building by selecting the whole way, the result can be distorted.
However, squaring the same building by selecting all of its nodes produces a clean,
correct rectangle.

This means two different squaring actions on the same geometry lead to different results,
which is confusing for users.

### Cause
Squaring a way and squaring selected nodes were using different orthogonalization paths.
The node-based approach produces better and more stable results.

### Solution
This PR makes squaring a way reuse the same vertex-based orthogonalization logic that is
used when squaring selected nodes. This ensures consistent behavior regardless of whether
the user squares a way or its nodes.

### Result
- Squaring a way now produces the same result as squaring all its nodes
- No change to existing node-based squaring behavior

### Testing
Tested locally using the dev OSM API with way `4307298815`.
Verified before/after behavior to confirm the results are now identical.

### Before

https://github.com/user-attachments/assets/8b443561-2e89-41f8-9ee4-665078fa1083

### After

now It is not distorting the selected area and giving same output on square and nodes 
image

### DIFF
--- a/modules/operations/orthogonalize.js
+++ b/modules/operations/orthogonalize.js
@@ -28,9 +28,24 @@ export function operationOrthogonalize(context, selectedIDs) {
         if (entity.type === 'way' && new Set(entity.nodes).size > 2 ) {
             if (_type && _type !== 'feature') return null;
             _type = 'feature';
-            return actionOrthogonalize(entityID, context.projection);
 
-        // square a single vertex
+            var graph = context.graph();
+            var nodes = graph.childNodes(entity);
+
+            // Apply vertex-based orthogonalization to all nodes
+            return function(graph, t) {
+                nodes.forEach(function(node) {
+                    var action = actionOrthogonalize(
+                        entityID,
+                        context.projection,
+                        node.id
+                    );
+                    if (!action.disabled(graph)) {
+                        graph = action(graph, t);
+                    }
+                });
+                return graph;
+            };
         } else if (geometry === 'vertex') {
             if (_type && _type !== 'corner') return null;
             _type = 'corner';


tordans:
@matkoniecz @tyrasd I propose to close this PR. I don't see the Use Case for squaring based on selected notes vs. building and the test video does not even show the issue.

matkoniecz:
https://www.openstreetmap.org/way/144745279 in regular API, 4307298815 is in dev version

matkoniecz:
> and the test video does not even show the issue.

testing it on https://www.openstreetmap.org/way/144745279 reveals that selecting nodes and squaring them in fact works differently than squaring way

which seems buggy

tyrasd:
> selecting nodes and squaring them in fact works differently than squaring way

In this case, the individual corners are squared one after each other, which is slightly different from the regular operation on a way which squares all corners "simultaneously".

> I propose to close this PR. […] test video does not even show the issue.

:+1:

PR #11714, opened 2025-12-22
fix typo

k-yle:
github launched a new UI which is overly dramatic about typos in untouched files:

image

so let's fix this one to make the warning disappear


1ec5 (approved):
Thanks for fixing the tpyo.

PR #11718, opened 2025-12-22
Fix Mapillary parent/child photo overlay toggling

### LINKED ISSUE #11671: Mapillary Data panel logic wrong
It is great that the user is allowed to turn these on independently,

Image

But the menu needs to be redesigned.

You see it is breaking a rule of logic: if I turn off the master switch in the fuse box, the subordinate switches should lose power too.

Anyway, just make a box called Mapillary, and inside have three independent switches.

krishnan05:
### Problem
The Mapillary Data panel violates expected parent/child toggle behavior (#11671).
Disabling the Mapillary master checkbox does not disable its subordinate overlays, leading to an inconsistent and confusing UI state.

### Screenshots

**Before**
Parent Mapillary toggle disabled, but child overlays remain enabled:
Before

**After**
Disabling the Mapillary parent toggle correctly disables all child overlays:
After

### Solution
This change enforces correct parent/child toggle logic for Mapillary overlays:
- Disabling the Mapillary parent checkbox now disables all Mapillary child overlays.
- Enabling or disabling child overlays updates the parent checkbox state accordingly.
- Child overlays remain independently configurable when the parent is enabled.

### Scope
- UI logic only (Photo Overlays / Mapillary section)
- No changes to rendering, data loading, or Mapillary integration

### Testing
- Toggled Mapillary parent checkbox on/off
- Verified all Mapillary child overlays follow parent state
- Verified parent state updates when child overlays are toggled individually

Fixes #11671

### DIFF
--- a/modules/ui/sections/photo_overlays.js
+++ b/modules/ui/sections/photo_overlays.js
@@ -13,7 +13,10 @@ export function uiSectionPhotoOverlays(context) {
     let _savedLayers = [];
     let _layersHidden = false;
     const _streetLayerIDs = ['streetside', 'mapillary', 'mapillary-map-features', 'mapillary-signs', 'kartaview', 'mapilio', 'vegbilder', 'panoramax'];
-
+    const _mapillaryChildLayerIDs = [
+    'mapillary-map-features',
+    'mapillary-signs'
+];
     var settingsLocalPhotos = uiSettingsLocalPhotos(context)
         .on('change',  localPhotosChanged);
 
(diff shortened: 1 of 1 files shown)


tordans:
This is not https://github.com/openstreetmap/iD/issues/11671#issuecomment-3675526585 is it?

krishnan05:
> This is not [#11671 (comment)](https://github.com/openstreetmap/iD/issues/11671#issuecomment-3675526585) is it?

You’re right — this PR addresses only the parent/child toggle logic issue described in #11671, not the broader UI restructuring discussed later in the thread.

The goal here was to fix the inconsistent “master switch” behavior without changing the existing panel structure. The broader questions around how photo and detection providers are grouped and presented go beyond a simple bug fix and likely need a more deliberate UX and architectural discussion, including panel structure and state ownership.

It may be useful to discuss possible directions or scope this into follow-up work, depending on what maintainers think makes the most sense.

PR #11651, opened 2025-12-07
Added Morocco to phone_formats.json

ilias52730:
I used the format +212 5 12 34 56 78:

- Prefix: `+212` (Country code).
- Type: Used `5` (landlines) instead of `6` or `7` (mobile), as this is more relevant for businesses and POIs on OpenStreetMap.
- Spacing: Uses the 2-digit grouping standard, following the French format on which it's based.
- Pattern: `12 34 56 78` used as a standard sequential placeholder.

Source for the format: Apart from local knowledge, [this official ITU document](https://www.itu.int/dms_pub/itu-t/oth/02/02/T02020000900031PDFE.pdf) with the contact details of the Moroccan telecommunications agency.
image

Feel free to ask if you need any more info.

PR #11733, opened 2025-12-30
Don't suggest values from Taginfo for `addr:*` tags

k-yle:
This was [reported](https://discord.com/channels/413070382636072960/926020366927790130/1455322775492493526) on the osm discord. 

Same as #10287, but for [`addr:*`](https://osm.wiki/Key:addr) instead of [`name:*`](https://osm.wiki/Key:name). Taginfo suggestions aren't useful for either tag.

We already [filter out values with an uppercase letter](https://github.com/openstreetmap/iD/blob/b36341dfa347578fe0add2f6c61b6c35124f2ab5/modules/services/taginfo.js#L99), which is why the dropdown only includes some popular CJK characters 😆

image

In a future PR we could use the same suggestions as the address picker, but that would quite a fair bit of refactoring

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased (2.38.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* Don't suggest values from Taginfo for `addr:*` tags ([#11733], thanks [@k-yle])
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
@@ -45,6 +46,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :bug: Bugfixes
 * Fix some gpx/geojson properties not visible, such as numbers or complex data structures ([#11636], thanks [@k-yle])
 #### :earth_asia: Localization
+* Add Moroccan phone number formats ([#11651], thanks [@ilias52730])
 * The Languages field shows language names in your preferred language. ([#11699], thanks [@Razen04])
 #### :hourglass: Performance
 #### :mortar_board: Walkthrough / Help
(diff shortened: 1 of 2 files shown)


k-yle:
[image](https://taginfo.osm.org/tags/addr%3Asuburb=%E5%AE%AE%E5%9F%8E%E9%87%8E%E5%8C%BA#map)

it seems like these suggestions have already caused a few misoperations. I doubt that [`addr:suburb=宮城野区`](https://taginfo.osm.org/tags/addr%3Asuburb=%E5%AE%AE%E5%9F%8E%E9%87%8E%E5%8C%BA#map) also exists [in germany](https://osm.org/way/1157487582) and [england](https://osm.org/way/186806647)

CHANGELOG.md
```diff
@@ -38,13 +38,15 @@ _Breaking developer changes, which may affect downstream projects or sites that
...
 #### :sparkles: Usability & Accessibility
+* Don't suggest values from Taginfo for `addr:*` tags ([#11733], thanks [@k-yle])
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 * Don't error on features with a sole `note` tag ([#11522])
 #### :bug: Bugfixes
 * Fix some gpx/geojson properties not visible, such as numbers or complex data structures ([#11636], thanks [@k-yle])
 #### :earth_asia: Localization
+* Add Moroccan phone number formats ([#11651], thanks [@ilias52730])
```
tordans, line 49: This is not part of the PR is it?

k-yle: yeah, I tried to push this to #11651 before merging it, but evidently i did something wrong.

just remembered. could put it into a separate PR if that's preferred

tyrasd: @k-yle IMO there's no need for dedicated PRs for trivial changelog-only changes like this. Feel free to commit those directly to _develop_.

tyrasd (approved):

PR #11734, opened 2025-12-31
Fix merge deleting buildings when using Lasso selection

### LINKED ISSUE #9328: Can create invalid geometries with node merge.

### URL

1. Click on "area"

Before merging:

After merging nodes:

```
    <create>
    </create>
    <modify/>
    <delete if-unused="true"/>
</osmChange>
```

2.22.0

Chrome

### LINKED ISSUE #11726: Merging deletes instead of merging

Ankitgkp:
Related - #11726
Updates actionMergePolygon to ignore nodes included in the selection (e.g., from Lasso tool). Previously, the presence of nodes caused mergePolygon to disable itself, causing the editor to fall back to inappropriate actions (like mergeNodes) which resulted in buildings being deleted. This change ensures that selecting buildings with the Lasso tool correctly merges them into a multipolygon.

### DIFF
--- a/modules/actions/connect.js
+++ b/modules/actions/connect.js
@@ -18,7 +18,7 @@ import { utilArrayUniq, utilOldestID } from '../util';
 //   https://github.com/openstreetmap/josm/blob/mirror/src/org/openstreetmap/josm/actions/MergeNodesAction.java
 //
 export function actionConnect(nodeIDs) {
-    var action = function(graph) {
+    var action = function (graph) {
         var survivor;
         var node;
         var parents;
@@ -73,7 +73,7 @@ export function actionConnect(nodeIDs) {
     };
 
 
-    action.disabled = function(graph) {
+    action.disabled = function (graph) {
         var seen = {};
         var restrictionIDs = [];
         var survivor;
@@ -83,6 +83,32 @@ export function actionConnect(nodeIDs) {
 
         // Select the node with the oldest ID as the survivor.
         survivor = graph.entity(utilOldestID(nodeIDs));
+
+        // 0. disable if merging would destroy a tagged closed way (e.g., building)
+        for (i = 0; i < nodeIDs.length; i++) {
+            node = graph.entity(nodeIDs[i]);
+            var parentWays = graph.parentWays(node);
+            for (j = 0; j < parentWays.length; j++) {
+                way = parentWays[j];
+                if (way.isClosed() && way.hasInterestingTags()) {
+                    // Check if this way would become degenerate after merge
+                    var testWay = way.update({});
+                    for (k = 0; k < nodeIDs.length; k++) {
+                        if (nodeIDs[k] === survivor.id) continue;
+                        if (testWay.contains(nodeIDs[k])) {
+                            if (testWay.areAdjacent(nodeIDs[k], survivor.id)) {
+                                testWay = testWay.removeNode(nodeIDs[k]);
+                            } else {
+                                testWay = testWay.replaceNode(nodeIDs[k], survivor.id);
+                            }
+                        }
+                    }
+                    if (testWay.isDegenerate()) {
+                        return 'would_destroy_feature';
+                    }
+                }
+            }
+        }
 
         // 1. disable if the nodes being connected have conflicting relation roles
         for (i = 0; i < nodeIDs.length; i++) {
@@ -132,8 +158,8 @@ export function actionConnect(nodeIDs) {
             if (!relation.isComplete(graph)) continue;
 
             var memberWays = relation.members
-                .filter(function(m) { return m.type === 'way'; })
-                .map(function(m) { return graph.entity(m.id); });
+                .filter(function (m) { return m.type === 'way'; })
+                .map(function (m) { return graph.entity(m.id); });
 
             memberWays = utilArrayUniq(memberWays);
             var f = relation.memberByRole('from');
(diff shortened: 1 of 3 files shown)


k-yle (changes requested):
did you test your changes? the problem described in #11726 is not affected by the code in this file.

#11726 describes an inconvenience with `actionMergeNodes`, not `actionMergePolygon`

Ankitgkp:
@k-yle When lasso-selecting buildings and pressing C to merge, only nodes are selected
(not ways). This causes actionMergeNodes → actionConnect to merge all nodes,
making building ways degenerate and deleting them.
Add check in actionConnect.disabled() to return 'would_destroy_feature' when
merging would make a tagged closed way (e.g., building) degenerate.

modules/actions/connect.js
```diff
@@ -84,6 +84,32 @@ export function actionConnect(nodeIDs) {
...
         survivor = graph.entity(utilOldestID(nodeIDs));
 
+        // 0. disable if merging would destroy a tagged closed way (e.g., building)
+        for (i = 0; i < nodeIDs.length; i++) {
+            node = graph.entity(nodeIDs[i]);
+            var parentWays = graph.parentWays(node);
+            for (j = 0; j < parentWays.length; j++) {
+                way = parentWays[j];
+                if (way.isClosed() && way.hasInterestingTags()) {
+                    // Check if this way would become degenerate after merge
```
k-yle: thanks, this is better. but – the code is checking whether the way is degenerate _before_ the merge, not _after_. Therefore, it doesn't work.

You would need to run the action itself, and check the graph after the action. But this is a bit unconventional, no other action does this.

k-yle: thanks, this is better. but – the code is checking whether the way is degenerate _before_ the merge, not _after_. Therefore, it doesn't work.

You would need to run the action itself, and check the graph after the action. But this is a bit unconventional, no other action does this. 

something like this maybe? (untested)
```ts
const newGraph = action(graph);
const parentWays = nodeIDs
    .map(id => graph.entity(id))
    .flatMap(node => graph.parentWays(node));

if (parentWays.some(way => way.hasInterestingTags() && !newGraph.hasEntity(way.id))) {
    return 'would_destroy_parent';
}
if (parentWays.some(way => way.isDegenerate())) {
    return 'would_degenerate_parent';
}
```

k-yle (commented):
Since you're now talking about degenerate ways, I'm guessing you want to solve #9328 at the same time as #11726 ?

if so, I would suggest first creating some test cases [in here](https://github.com/openstreetmap/iD/blob/eea86eb4c138595cebcf32ec179b6b75ad6b2f6a/test/spec/actions/connect.js#L233), so that we're aligned about the goal:

```js
it('allows uninteresting parent ways to be deleted by the operation', () => { // baseline 
    const graph = iD.coreGraph([
        iD.osmNode({ id: 'a' }),
        iD.osmNode({ id: 'b' }),
        iD.osmWay({ id: '-', nodes: ['a', 'b'] }),
    ]);
    expect(iD.actionConnect(['a', 'b']).disabled(graph)).toBe(false);
});

it('refuses to connect nodes if an interesting parent way would be deleted', () => { // #11726
    const graph = iD.coreGraph([
        iD.osmNode({ id: 'a' }),
        iD.osmNode({ id: 'b' }),
        iD.osmWay({ id: '-', nodes: ['a', 'b'], tags: { highway: 'residential' } }),
    ]);
    expect(iD.actionConnect(['a', 'b']).disabled(graph)).toBe('would_destroy_parent');
});

it('refuses to connect nodes if any parent ways would become degenerate', () => { // #9328
    const graph = iD.coreGraph([
        iD.osmNode({ id: 'a', loc: [0, 0] }),
        iD.osmNode({ id: 'b', loc: [2, 0] }),
        iD.osmNode({ id: 'c', loc: [1, 1] }),
        iD.osmNode({ id: 'd', loc: [2, 2] }),
        iD.osmNode({ id: 'e', loc: [0, 2] }),
        iD.osmWay({ id: '-', nodes: ['a', 'b', 'c', 'd', 'e', 'a'], tags: { building: 'yes' } }),
    ]);
    expect(iD.actionConnect(['b', 'd']).disabled(graph)).toBe('would_degenerate_parent');
});
```
the last example is:

Image

modules/actions/connect.js
Ankitgkp: Thanks for the guidance! I've updated actionConnect.disabled() following your suggested approach:
Implementation:

- Run `action(graph)` first to get **newGraph**

- Check if any parent ways with interesting tags were deleted → return 'would_destroy_parent'

- Check if any parent ways became degenerate (sDegenerate()) → return 'would_degenerate_parent'

- Check for self-intersection (repeated non-endpoint nodes) → return 'would_degenerate_parent'

and Test cases added to `test/spec/actions/connect.js`

```
it('allows uninteresting parent ways to be deleted by the operation', function() {
    var graph = iD.coreGraph([
        iD.osmNode({id: 'a'}),
        iD.osmNode({id: 'b'}),
        iD.osmWay({id: '-', nodes: ['a', 'b']}),
    ]);
    expect(iD.actionConnect(['a', 'b']).disabled(graph)).to.be.not.ok;
});

it('refuses to connect nodes if an interesting parent way would be deleted', function() {
    var graph = iD.coreGraph([
        iD.osmNode({id: 'a'}),
        iD.osmNode({id: 'b'}),
        iD.osmWay({id: '-', nodes: ['a', 'b'], tags: { highway: 'residential' }}),
    ]);
    expect(iD.actionConnect(['a', 'b']).disabled(graph)).to.eql('would_destroy_parent');
});

it('refuses to connect nodes if any parent ways would become degenerate', function() {
    var graph = iD.coreGraph([
        iD.osmNode({id: 'a', loc: [0, 0]}),
        iD.osmNode({id: 'b', loc: [2, 0]}),
        iD.osmNode({id: 'c', loc: [1, 1]}),
        iD.osmNode({id: 'd', loc: [2, 2]}),
        iD.osmNode({id: 'e', loc: [0, 2]}),
        iD.osmWay({id: '-', nodes: ['a', 'b', 'c', 'd', 'e', 'a'], tags: { building: 'yes' }}),
    ]);
    expect(iD.actionConnect(['b', 'd']).disabled(graph)).to.eql('would_degenerate_parent');
});
```
**All 26 tests pass (23 existing + 3 new).**

matkoniecz:
> AssertionError: expected [ 'n-51', 'n-50' ] to deeply equal [ 'n-50', 'n-51' ]

matkoniecz:
I propose to close this as an abandoned PR

tyrasd:
closing as stale

PR #11741, opened 2026-01-04
Avoid applying bridge/tunnel tags to buildings in crossing_ways

### LINKED ISSUE #11624: Tag this as higher does nothing it seems to the user
"House crosses service road."
We choose "tag this as higher".
But on the map the lines don't change.
It is just like we picked "tag this as lower."
Sure, maybe in the info panel something changes.
But the map stays the same.

Shyam-123pandey:
Ensure makeChangeLayerFix only applies bridge/tunnel tags
to valid linear features (highways, railways, waterways).

Buildings now receive layer-only adjustments, and any
existing bridge/tunnel tags are removed to prevent
invalid OSM tagging.

Also removes redundant mode re-entry and aligns
indentation with surrounding helper functions.

### DIFF
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -1,5 +1,4 @@
 preset-icon-container/* highways */
-
 /* defaults */
 .preset-icon .icon.tag-highway.other-line {
     color: #fff;
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -1,15 +1,19 @@
-preset-icon-container/* highways */
+/* preset-icon-container highways */
+
 /* defaults */
 .preset-icon .icon.tag-highway.other-line {
     color: #fff;
     fill: #777;
 }
+
 path.line.casing.tag-highway {
     stroke: #444;
 }
+
 path.line.stroke.tag-highway {
     stroke: #ccc;
 }
+
 
 /* wide highways */
 path.line.shadow.tag-highway {
--- a/css/30_highways.css
+++ b/css/30_highways.css
@@ -1,6 +1,3 @@
-/* preset-icon-container highways */
-
-/* defaults */
 .preset-icon .icon.tag-highway.other-line {
     color: #fff;
     fill: #777;
@@ -13,7 +10,6 @@ path.line.casing.tag-highway {
 path.line.stroke.tag-highway {
     stroke: #ccc;
 }
-
 
 /* wide highways */
 path.line.shadow.tag-highway {
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -32,7 +32,6 @@ export function validationCrossingWays(context) {
         return way;
     }
 
-
     function hasTag(tags, key) {
         return tags[key] !== undefined && tags[key] !== 'no';
     }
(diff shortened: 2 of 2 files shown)


css/30_highways.css
```diff
@@ -1,17 +1,20 @@
...
+
 path.line.casing.tag-highway {
     stroke: #444;
 }
+
 path.line.stroke.tag-highway {
     stroke: #ccc;
 }
 
+
```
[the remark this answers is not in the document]
matkoniecz: LLM is right here, can you drop this changes?

modules/validations/crossing_ways.js
```diff
@@ -794,7 +793,11 @@ export function validationCrossingWays(context) {
         return fix;
     }
 
-    function makeChangeLayerFix(higherOrLower) {
+    // =============================
+// FIXED: makeChangeLayerFix
```
matkoniecz: comment debris should not be present in actual PR

css/30_highways.css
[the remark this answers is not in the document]
Shyam-123pandey: okay sir

modules/validations/crossing_ways.js
Shyam-123pandey: okay sir

matkoniecz (commented):
is it LLM output?

have you reviewed patch proposed here and tested it?

css/30_highways.css
```diff
@@ -1,13 +1,12 @@
-preset-icon-container/* highways */
-
-/* defaults */
 .preset-icon .icon.tag-highway.other-line {
     color: #fff;
     fill: #777;
 }
+
```
matkoniecz, line 5: what is the reason for this edit?

modules/validations/crossing_ways.js
```diff
@@ -794,7 +793,11 @@ export function validationCrossingWays(context) {
...
+// =============================
+
+      function makeChangeLayerFix(higherOrLower) {
```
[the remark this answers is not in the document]
matkoniecz: LLM review is right here

matkoniecz:
> Avoid applying bridge/tunnel tags to buildings in crossing_ways

is it fixing some bug or adding a feature? Is this PR changing iD behavior in any way?

RAytoun:
Why is iD Editor creating bad tagging. Too many inexperienced mappers are using this to justify fixing the overlapping buildings with a layer tag on one of the buildings. As a result we now have a Map Paint Style that has to identify buildings with the layer tag so that we can remove it and fix the overlapping buildings by adjusting the geometry.

**Can we please remove the suggestion that overlapping buildings should be tagged as higher or lower.**

matkoniecz:
> Can we please remove the suggestion that overlapping buildings should be tagged as higher or lower.

iff there is such suggestion then it would be better to report it as an issue, if not reported already

bryceco:
To explain the confusion: This PR fixes bugs introduced in PR https://github.com/openstreetmap/iD/pull/11737. 

These fixes should be added to the original PR and this PR closed. 

The original PR is still a hot mess though, as it fixes a layer issue by adding bridge/tunnel tags.

matkoniecz:
> This PR fixes bugs introduced in PR https://github.com/openstreetmap/iD/pull/11737

and as it was never merged it makes no sense to open it here

> The original PR is still a hot mess though

I bet that it is mostly LLM generated (PR creator also posted LLM comments across various issues)

css/30_highways.css
Shyam-123pandey: may be sir at the time  css checking , i pushed one space

Shyam-123pandey:
> is it LLM output?
> 
> have you reviewed patch proposed here and tested it?

yes

PR #11749, opened 2026-01-07
A couple of strings fixes/typos

Porkepix:
These are a couple of strings fixes or typos, but several notes must be made:
- For the pure English spelling, English isn't my native language and while the sentence seemed completely wrong to me when I read it, it might be me who's completely mistaken and missing some English skills here, which results in me trying to correct what I think is a mistake while it isn't.
- Regarding the unbreakable spaces, having lines ending with "Tip:" and then a new line isn't correct, which is the reason for the replacement. I didn't look in the file (or if there are other files?) for other occurrences for colons as the fact it's a YamL file make it quite difficult to only look for colons as it'd basically be every line. Maybe it'd be needed to look for every second or more colon or something along?
- About the replacement of three dots by ellipses is probably the one I'm the most unsure about. It's a mistake often made due to how difficult it is to produce the ellipsis on some OS, but here in the context of something loading it might be completely preferred to have three dots to simulate a loading with one dot appearing after another.

Finally, I'm not sure what kind of markup is used here, and couldn't find the answer in the code (I don't practice JS, maybe it's a common one for this ecosystem?) but saw several markups with pairs or curly braces, so maybe ellipses or unbreakable spaces should be done that way too?

Let me know if I'm wrong about some (or even all!) of these changes and if I should adapt (drop commits) the PR accordingly!

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -2270,7 +2270,7 @@ en:
       add_name: "Let's pretend that you have local knowledge of this cafe, and you know its name. **Add a name for the cafe.**"
       add_reminder: "All entered information should be objective and verifiable."
       add_close: "The feature editor will remember all of your changes automatically. **When you are finished adding the name, press the {button} button or `{esc}` to close the feature editor.**"
-      reselect: "Often points will already exist, but have mistakes or be incomplete. We can edit existing points. **Select the cafe you just created.**"
+      reselect: "Often points will already exist, but have mistakes or are incomplete. We can edit existing points. **Select the cafe you just created.**"
       update: "Let's fill in some more details for this cafe. You can change its name, add a cuisine, or add an address. **Change the cafe details.**"
       update_close: "**When you are finished updating the cafe, press the {button} button or `{esc}` to close the feature editor.**"
       rightclick: "You can {rightclick} right-click on any feature to see the *edit menu*, which shows a list of editing operations that can be performed.{br}A right-click might be the same as a control-click or two-finger click. **Right-click to select the point you created and show the edit menu.**"
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -477,8 +477,8 @@ en:
             single_node: Divide this line into two at this point.
             multiple_node: Divide this line at these points.
           multiple:
-            single_node: "Divide all lines at this point. Tip: To limit this operation to a specific line, select both the line and point before performing the split."
-            multiple_node: "Divide all lines at these points. Tip: To limit this operation to a specific line, select the line as well as the points before performing the split."
+            single_node: "Divide all lines at this point. Tip: To limit this operation to a specific line, select both the line and point before performing the split."
+            multiple_node: "Divide all lines at these points. Tip: To limit this operation to a specific line, select the line as well as the points before performing the split."
         area:
           single:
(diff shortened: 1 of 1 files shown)


data/core.yaml
```diff
@@ -583,7 +583,7 @@ en:
   zoom_in_edit: Zoom in to edit
   login: Log In
   logout: Log Out
-  loading_auth: "Connecting to OpenStreetMap..."
+  loading_auth: "Connecting to OpenStreetMap…"
```
matkoniecz, line 586: why "..." would be a mistake?

data/core.yaml
```diff
@@ -2270,7 +2270,7 @@ en:
       add_name: "Let's pretend that you have local knowledge of this cafe, and you know its name. **Add a name for the cafe.**"
       add_reminder: "All entered information should be objective and verifiable."
       add_close: "The feature editor will remember all of your changes automatically. **When you are finished adding the name, press the {button} button or `{esc}` to close the feature editor.**"
-      reselect: "Often points will already exist, but have mistakes or be incomplete. We can edit existing points. **Select the cafe you just created.**"
+      reselect: "Often points will already exist, but have mistakes or are incomplete. We can edit existing points. **Select the cafe you just created.**"
```
matkoniecz, line 2273: Why this change is needed?

(I asked actual native speaker for help and they claimed that "be incomplete" is correct form and "are incomplete" is having grammatical problem in this sentence)

Porkepix: Well, as stated in my message, I'm not native speaker so I can be mistaken; looking at your message it seems to be the case (even though it seems weird to be, because "points" and "mistakes" are under plural form, so it looked to me like the verb had to be in a plural form too).

EDIT: Funnily enough, I tried to check both forms against grammar checkers - randomly found on the Internet - and some detects no issues with both forms, some states original form is right or want to correct my form with it, some others state my correction is right and suggest it when provided the original form. Looks like none of them agree. But on the other hand, I can't check which of these rely on LLMs and would be prone to come up with its own things based off nothing nowadays…

data/core.yaml
Porkepix: It seemed to me the ellipsis symbol was more appropriate, unless you display points one by one to show something's ongoing. Am I wrong in that regard?

data/core.yaml
matkoniecz: can you then change back to original?

tyrasd:
changing the e.g. `...` ellipsis to `…` is not improving the strings enough to warrant invalidating the existing translations. let's keep them as they are for now

Porkepix:
> changing the e.g. `...` ellipsis to `…` is not improving the strings enough to warrant invalidating the existing translations. let's keep them as they are for now

Sorry, I completely forgot about the PR.
It's not only about ellipsis, but also using non-breakable spaces, as otherwise we had split lines at some places in the UI that were not fitting.

I can remove the commit regarding English wording if you want, but I have to admit I don't know what/who to trust on that one. I got opposite feedback from other people on the matter when I asked (saying the correction was more fitting), and grammar tools online can answer it's correct or a mistake for both version, and it becomes hard to tell which ones are correct at this age of LLMs outputting things out of pure stats.

PR #11750, opened 2026-01-07
render special oneway arrows for the `priority` direction

k-yle:
this is just an experiment. [bidirectional ways](http://osm.wiki/Tag:oneway=alternating) can sometimes have one direction which has priority. Typically signposted as [](https://commons.wikimedia.org/wiki/File:Vienna_Convention_road_sign_B5-V1.svg) and [](https://commons.wikimedia.org/wiki/File:Vienna_Convention_road_sign_B6.svg),  and tagged as [`priority=forward`/`backward`](https://osm.wiki/Key:priority).

The same concept exists for railways: the interlocking can support [wrong-road running](http://osm.wiki/Key:railway:bidirectional), but there might still be a [preferred direction](http://osm.wiki/Key:railway:preferred_direction).

we could render this to map it easier to map:

image

alternatives:

image
image

real example:

image

currently it looks quite chaotic, i'm not really sure how to do this well...

### DIFF
--- a/modules/osm/tags.js
+++ b/modules/osm/tags.js
@@ -214,7 +214,10 @@ export const osmOneWayForwardTags = {
         'spillway': true,
         'stream': true,
         'tidal_channel': true
-    }
+    },
+    'railway:preferred_direction': {
+        'forward': 'forward',
+    },
 };
 export const osmOneWayBackwardTags = {
     'conveying': {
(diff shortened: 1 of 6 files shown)


matkoniecz:
it looks fine to me?

yes, it looks a bit confusing but situation is in fact confusing

matkoniecz:
maybe give up on expressing priority?

matkoniecz:
Maybe try sizes but with larger difference in size?

Maybe try colour with also line on gray side in gray? Maybe more grayish?

PR #11752, opened 2026-01-07
feat: implemented full-width mobile sidebar with close button

### LINKED ISSUE #11739: Make sidebar on narrow/mobile screens full-width modal
### Description

On mobile, there's no point keeping the main UI pane visible when the sidebar is open  - you can't use any of it and it's just reducing the amount of space available for the sidebar's contents instead

Instead, when the sidebar is opened, it should act as a full-width modal, and the close button should hide it.

### Screenshots

E.g. instead of this:

Image

Do this this instead:

Image

ak8abhinay:
This PR fixes the issue #11739 improves the mobile experience by turning the sidebar into a full-width modal on screens smaller than 850px.

# Changes
  Fixes #11739
- **UI:** Sidebar now takes 100% width on mobile devices.
- **Logic:** Added a custom floating 'X' close button in `modules/ui/sidebar.js`.
- **CSS:** Added media queries in `css/80_app.css` to hide the map when the modal (sidebar) is active and takes full width.
- 
### Testing
- Verified using Chrome DevTools.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -866,31 +866,36 @@ a.hide-toggle {
         height: 20px;
         fill: currentColor;
     }
+
     .sidebar-close-button {
         position: absolute;
-        top: 10px;
-        right: 10px;
-        z-index: 9999; /* Increased to stay above all inspector elements */
-        padding: 10px;
-        background: white; /* Changed from transparent for better visibility on mobile */
-        border-radius: 50%; /* Makes it look like a clean circular button */
+        top: 0;
+        right: 0;
+        width: 40px;
+        height: 40px;
+        background: transparent;
         border: none;
+        border-radius: 0;        
+        outline: none;           
+        -webkit-appearance: none; 
+        appearance: none;  
         cursor: pointer;
         display: flex;
         align-items: center;
         justify-content: center;
-        box-shadow: 0 2px 5px rgba(0,0,0,0.2); /* Helps it pop against map data */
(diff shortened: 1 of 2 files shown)


matkoniecz:
Have you tested how it works on mobile devices? Can you attach screenshot how it works?

> UI: Sidebar now takes 100% width on mobile devices.

what about tablets?? On large screens it seems to not be helpful

matkoniecz:
BTW, #11753 seems to be doing the same

ak8abhinay:
Current behaviour :

image


After implementation :


image






> what about tablets??


iPad Air :


image


iPad mini : 


image

matkoniecz:
Again, is it intentional that mobile devices with huge screens (tablets) will be also affected? Per PR title they will be

css/80_app.css
```diff
@@ -824,6 +824,80 @@ a.hide-toggle {
...
+    }
+
+    /* sidebar fill the entire screen */
+    .ideditor.sidebar-modal-open .sidebar {
+        position: fixed; 
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+        width: 100% !important;
```
matkoniecz, line 842: `!important` is a CSS code smell

is it really impossible to avoid it?

ak8abhinay: > `!important` is a CSS code smell

You're right.
I'll look into removing the !important and use selector that works

PR #11753, opened 2026-01-07
[Fix] Modal-like sidebar layout on mobile screens

### LINKED ISSUE #11739: Make sidebar on narrow/mobile screens full-width modal

waqas2714:
## What changed
The sidebar now behaves like a modal on small screens to improve usability on mobile devices.

## Details
- The sidebar is toggled off by default on mobile screen sizes.
- It can be opened using the **Inspect** button, which toggles the sidebar.
- Additional toggle buttons are added to the headers of the feature list and entity editor, allowing mobile users to easily navigate back to the map view.

Closes #11739  
Relates to #9815

## Screenshots
Sidebar closed on mobile

Sidebar open as modal on mobile

## Testing
- Tested locally with `npm start`
- Verified behavior using browser mobile emulation
- Confirmed desktop layout remains unchanged

## Checklist
- [x] Tested locally
- [x] No unrelated changes

### DIFF
--- a/modules/ui/sidebar.js
+++ b/modules/ui/sidebar.js
@@ -28,6 +28,10 @@ export function uiSidebar(context) {
     var _wasData = false;
     var _wasNote = false;
     var _wasQaItem = false;
+    var isMobile = (
+        window.matchMedia &&
+        window.matchMedia('(max-width: 640px)').matches
+    );
 
     // use pointer events on supported platforms; fallback to mouse events
     var _pointerPrefix = 'PointerEvent' in window ? 'pointer' : 'mouse';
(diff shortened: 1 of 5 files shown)


matkoniecz:
#11752 seems to be doing the same

modules/ui/sidebar.js
```diff
@@ -28,6 +28,10 @@ export function uiSidebar(context) {
     var _wasData = false;
     var _wasNote = false;
     var _wasQaItem = false;
+    var isMobile = (
```
matkoniecz, line 31: is it possible to define it in one place? The same conditions are tested in [‎modules/ui/init.js](https://github.com/openstreetmap/iD/pull/11753/changes#diff-883e12dd828b2eff13e3704a7b017fccd7cc280ba906b99f1afdf37de6922777) and we do not want to end with just one of them changed in future

PR #11759, opened 2026-01-10
fix locale-specific wikibase properties not being used

### LINKED ISSUE #8450: iD element image not loading correctly
Hi there.

Recently, I noticed that images from iD's elements are not loading correctly depending on language.
The image showed is always from the generic image, not the language limited image, which uses the [P26 property](https://wiki.openstreetmap.org/wiki/Property:P26) in the wiki.
For example: historic=pillory has a [specific image for Portuguese](https://wiki.openstreetmap.org/wiki/Item:Q6103), but iD shows the standard image:
https://i.imgur.com/ofKcznm.png

What's the issue here?

Regards.

k-yle:
Closes #8450

The documentation image for `historic=pillory` when using `locale=pt` now looks like this:

image

instead of this (the [default for all other languages](https://osm.wiki/Item:Q6103#P28)): 
image

The bug was caused by `localeSitelinks.includes(...)` always returning false, because the array is always empty


k-yle:
uh oh, it looks like i misoperated this and commited 3fd72418564dabadc5c47e37fc558f2a6f5e25cd directly to the develop branch 😬 @tyrasd should i revert that commit directly?

k-yle:
closing since this is completely broken. added `develop` to my `~/.gittemplates/hooks/pre-hook` safeguard to prevent this from happening again

PR #11773, opened 2026-01-14
fix remove double-escaping of & in background layer names (#9098) 

### LINKED ISSUE #9098: Background names are double-escaped
Screenshot:

Screenshot 2022-05-06 16 33 34

e.g. `StratMap CapArea, Brazos &amp; Kerr Imagery (Natural Color 2021)`.

The layer name contains a `&` character; iD apparently escapes this value then escapes it again in the process of looking up a localized name for the layer.

Kaushik4141:
Removed lodash escape() calls from background_source.js The localizer's
t.append() uses D3's  .text() which already handles special characters
safely causing pre-escaped &amp; to display literally
Fixes #9098
image


tyrasd (approved):
Thanks!

PR #11774, opened 2026-01-14
 Adding reflection axis preview on Flip Long/Short hover (#11769)

### LINKED ISSUE #11769: Add visual indicator for flip axis
### Description

**Current situation:** The flip operator either acts on the "long" or "short" axis. The user is required to work out which axis these refer to before selecting the correct operation.

**Issue:** This requires the user to be able to visualise and understand what these options mean. Although seasoned editors may understand, I think it's probably not the most intuitive action (I often end up selecting the wrong one myself). Additionally, in cases where the axis lengths aren't hugely different, or where the polygon is a complex shape, it can be hard to know which is the long and which is the short axis.

**Feature idea:** To improve usability, it would be nice if the editor could visually indicate which axis is being referred to when hovering over the Flip Long and Flip Short actions. A similar feature is already in place for the Disconnect action which highlights the nodes related to that object.

**Suggestion:** A few ideas of what this might look like:
- The flip axis could be highlighted using a method similar to the disconnect action
- The colour of the flashing padding around the flip axis could be changed (from red to something else)
- The flashing padding could be removed, except for the flip axis

### Screenshots

_No response_

Kaushik4141:
closes #11769
Display a dashed line indicating the flip axis when hovering over
Flip Long or Flip Short menu items Helps users visualize the
transformation before applying it

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :sparkles: Usability & Accessibility
 * Don't suggest values from Taginfo for `addr:*` tags ([#11733], thanks [@k-yle])
 #### :scissors: Operations
+* Display reflection axis on the map while hovering the reflection operations in the edit menu ([#11774], thanks [@Kaushik4141])
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 * Don't error on features with a sole `note` tag ([#11522])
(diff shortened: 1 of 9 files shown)


tyrasd (approved):
Thanks, this is a useful addition to make it clear what the operation is going to do.

I've refactored it slightly to remove the duplicated code in the `action` and `operation` modules, and generalized it a bit to allow the functionality to be reused by other operations in the future (a PR will be coming soon).

PR #11781, opened 2026-01-16
tag values with wiki page may be documented as bad

### LINKED ISSUE #5460: Stop treating all tag values with OSM Wiki page as recommended

matkoniecz:
fixes #5460 again (admittedly, iD is not suggesting `highway=unsurfaced` again so it is less bad than it used to be)

redo of #5461

amends a2cacaaf2456de4c8443a7b5dab0ac3393be5deb

### DIFF
--- a/modules/services/taginfo.js
+++ b/modules/services/taginfo.js
@@ -97,7 +97,7 @@ function filterValues(allowUpperCase, key) {
         if (!allowUpperCase &&
             !(key === 'type' && d.value === 'associatedStreet') &&
             d.value.match(/[A-Z*]/) !== null) return false;  // exclude uppercase letters
-        return d.count > 100 || d.in_wiki; // exclude rare undocumented tags
+        return d.count > 100; // exclude rare tags
     };
 }
 


matkoniecz:
from reading a2cacaaf2456de4c8443a7b5dab0ac3393be5deb some tests also should be changed

matkoniecz:
on first look I am unsure how to invert this test

> FAIL  test/spec/services/taginfo.js > iD.serviceTaginfo > #values > includes unpopular values with a wiki page
> AssertionError: expected spy to have been called with arguments null, [ { value: 'party', title: 'A place for partying' } ]

tordans:
So we have a way to check "tag used less than 100 but is documented in wiki and approved (AKA not "in use")"
If possible, this would be better by exposing approved (hopefully well documented) tags still. 

---

In general I think this is a good change.

matkoniecz:
> and approved (AKA not "in use")"

"de facto" is also a good value

matkoniecz:
though I would go toward dropping that suggestions, especially as iD preset get more complete

modules/services/taginfo.js
```diff
@@ -97,7 +97,7 @@ function filterValues(allowUpperCase, key) {
         if (!allowUpperCase &&
             !(key === 'type' && d.value === 'associatedStreet') &&
             d.value.match(/[A-Z*]/) !== null) return false;  // exclude uppercase letters
-        return d.count > 100 || d.in_wiki; // exclude rare undocumented tags
+        return d.count > 100; // exclude rare tags
```
k-yle, line 100: To update the test case, you would just need to change [line 271](https://github.com/matkoniecz/iD/blob/0aaedf74b93c067bd210b688be3267cdf1240f67/test/spec/services/taginfo.js#L271):

```diff
  expect(callback).to.have.been.calledWith(
-   null, [{'value':'party','title':'A place for partying'}]
+   null, []
  );
```

And update the test title [a few lines above](https://github.com/matkoniecz/iD/blob/0aaedf74b93c067bd210b688be3267cdf1240f67/test/spec/services/taginfo.js#L259C13-L259C55).

matkoniecz: https://github.com/openstreetmap/iD/pull/11794/files did this, I am fine with their PR being merged

k-yle: oh I didn't see that, thanks

k-yle:
merged via #11794

PR #11783, opened 2026-01-17
Country names will be pretty instead of ISO codes just like the language field

### LINKED ISSUE #7123: Country should be a combo field

Razen04:
Implemented the combo field of country to show the localized country names similar to the language field.

Fixes #7123 

## Recording:

https://github.com/user-attachments/assets/1e265338-4271-424b-a4c4-ef0d3d530556

Adding `if` statements like this for all type of iso codes won't scale, if somehow for specific fields we can have specific metadata then for those we can generalise or maybe there is a better way to do it? Also target looks same as country, if that should also be fixed then please let me know.


modules/ui/fields/combo.js
```diff
@@ -232,6 +246,33 @@ export function uiFieldCombo(field, context) {
           return options;
         }
 
+        // Get dropdown list for country key
+        if (field.key === 'country') {
```
1ec5: > Adding `if` statements like this for all type of iso codes won't scale, if somehow for specific fields we can have specific metadata then for those we can generalise or maybe there is a better way to do it? Also target looks same as country, if that should also be fixed then please let me know.

That’s a good point. `target=*` also expects an ISO&nbsp;3166-1 alpha-2 country code, just like `country`. Of [the keys are set to country codes](https://wiki.openstreetmap.org/wiki/Country_code#ISO_3166-1), I think id-tagging-schema only has fields for `country=*` and `target=*`. (Technically, it also has `addr:country=*`, but iD [doesn’t expose](https://github.com/openstreetmap/iD/blob/c0a783165ad73ccfc63880c90a7eac63cd0b8d44/data/address_formats.json) it as part of the Address field.)

Since it’s just two keys, I think special-casing is acceptable, but just in case, we could have the special case rely on a constant in [tags.js](https://github.com/openstreetmap/iD/blob/c0a783165ad73ccfc63880c90a7eac63cd0b8d44/modules/osm/tags.js). That would make it easier to push the logic into schema-builder and id-tagging-schema if the need arises.

modules/ui/fields/combo.js
1ec5, line 1: The selected value is displayed in monospaced font, as if it’s a raw tag value. It should look similar to a value of a more normal keyword-based field:

Nhà ở của đại sứ in blue variable-width font; Hoa Kỳ in black monospaced font.

modules/ui/fields/combo.js
```diff
@@ -232,6 +246,33 @@ export function uiFieldCombo(field, context) {
...
+          const localeCode = localizer.localeCode();
+          const regionNames = new Intl.DisplayNames(localeCode, { type: 'region' });
+
+          let options = codes.map((c) => {
+            const name = regionNames.of(c); // all the codes will be converted to pretty names like: IN=India
+            return {
+              key: c,
+              value: name,
+              title: name,
+              display: selection => selection.text(name)
```
1ec5: Each menu item’s tooltip is the same as the label:

Các Tiểu Vương quốc Ả Rập Thống nhất in a tooltip.

Normally the tooltip is the raw tag value:

branch_embassy in a tooltip.

1ec5 (commented):
Thanks, this looks pretty close.

Razen04:
Fixed all the issue:
1. Country field keys are centralized in tags.js
2. Names will not be monospaced
3. Tooltip will show raw tag values

Btw I have a question, can there be multiple countries? If yes then we should make this render as chips like how the languages were selected and rendered to allow multiple countries. Otherwise this should be fine

1ec5:
> Btw I have a question, can there be multiple countries? If yes then we should make this render as chips like how the languages were selected and rendered to allow multiple countries. Otherwise this should be fine

You’re probably referring to the `multiCombo` field type. Technically, the raw `country=*` and `target=*` tags can be set to a semicolon-delimited list of country codes. For example, [this embassy](https://www.openstreetmap.org/node/12234937959) is a mission to three different countries and the European Union. However, this is a bit of an edge case. The id-tagging-schema fields corresponding to these keys are configured as simple combo boxes that only accept one value at a time. In the release version of iD, if you set the value to `BE;NL;LU;EU`, it’ll treat that as a raw unrecognized value (in monospaced font).

Let’s just make sure you can still set the value to such a list without any errors. Since the same file handles `multiCombo` fields too, there’s a good chance that your changes will just work if id-tagging-schema turns one of these fields into a `multiCombo` in the future.

modules/ui/fields/combo.js
```diff
@@ -232,6 +247,33 @@ export function uiFieldCombo(field, context) {
...
+            };
+          });
+
+          options.sort((a, b) => {
+              return a.value.localeCompare(b.value, localeCode);
+          });
+
+          return options;
+        }
```
1ec5, line 331: The new code block has a lot of similarity to the language code functionality in #11699. We end up having to convert a code to a localized name in six different places total. Is there any way to consolidate them?

How about factoring out a helper function that generates a valid tag values for languages, returning an object mapping raw tag values to localized names. Do the same for generating the list of valid regions. Then this function only needs to call the appropriate helper method, convert the return value into a sorted array of objects to put in `options`, and add the `other` option if necessary. You’d be able to reuse the helper functions in `displayValue()` and `renderValue()`.

modules/ui/fields/combo.js
```diff
@@ -232,6 +247,33 @@ export function uiFieldCombo(field, context) {
...
+          const localeCode = localizer.localeCode();
+          const regionNames = new Intl.DisplayNames(localeCode, { type: 'region' });
+
+          let options = codes.map((c) => {
+            const name = regionNames.of(c); // all the codes will be converted to pretty names like: IN=India
+            return {
+              key: c,
+              value: name,
+              title: c, // the tooltip should show the raw-tag value
+              display: selection => selection.text(name)
```
1ec5: Not sure if you noticed, but in #11699 you made use of `addComboboxIcons()`. This function adds an icon to a menu item if the field’s definition supplies a set of icons. Flag icons could improve the usability or attractiveness of the Country and Target fields. In particular, `country=*` also powers the Flag Country field in the Flagpole (`man_made=flagpole`) preset. Flag icons would be a stopgap for openstreetmap/id-tagging-schema#1757.

I’m not sure of the best way to supply all the required icons. If that’s too challenging, a simpler alternative would be to insert a flag emoji before the name. This would only be for display, of course. We wouldn’t want it to affect the raw tag value or the sort order.

This is only a nice-to-have that we could track separately without blocking this PR.

modules/ui/fields/combo.js
```diff
@@ -655,7 +698,8 @@ export function uiFieldCombo(field, context) {
         var showsValue = value => !isMixed && value && !(field.type === 'typeCombo' && value === 'yes');
         var isRawValue = value => showsValue(value)
             && !stringsField.hasTextForStringId(`options.${value}`)
-            && !stringsField.hasTextForStringId(`options.${value}.title`);
+            && !stringsField.hasTextForStringId(`options.${value}.title`)
+            && !osmIsoCountryKeys.has(field.key); // the country key should not be raw-valued
```
1ec5, line 777: Any idea why we didn’t need to do this for languages in #11699? Is it because the Languages field is a `multiCombo` field?

1ec5 (commented):
Remember to add an entry to the changelog so you get some credit for this improvement!

modules/ui/fields/combo.js
Razen04: There are `emojiFlags` available in features, we can use that.

modules/ui/fields/combo.js
Razen04: Yeah I should create helper functions, I didn't realize that I was reusing things at a lot of places, I should do that.

modules/ui/fields/combo.js
1ec5: `emojiFlags` is a good idea. If that doesn’t work for some reason, another possibility would be to grab the icon associated with the `man_made=flagpole` `country=*` preset in name-suggestion-index.

modules/ui/fields/combo.js
```diff
@@ -123,6 +124,13 @@ export function uiFieldCombo(field, context) {
           }
         }
 
+        if (osmIsoCountryKeys.has(field.key) && tval) {
+            const localeCode = localizer.localeCode();
+              const regionNames = new Intl.DisplayNames(localeCode, { type: 'region' });
+              const name = regionNames.of(tval);
```
k-yle: This will crash iD if the tag has an invalid value, such as `country=USA`.

[`Intl.DisplayNames#of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/of) throws an Error for invalid input, regardless of the [`fallback`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/DisplayNames#fallback) option.



Therefore, we need to wrap this block in a `try…catch` statement. This would be a good opportunity to refactor these 5 duplicate blocks into a helper function like Minh suggested.

k-yle (changes requested):

modules/ui/fields/combo.js
k-yle: flag emojis are intentionally not supported on Windows, so if we choose this option, maybe we need to check if `utilDetect().os !== 'win'`?

Otherwise Windows users will just see the raw "[regional indicator symbols](https://en.wikipedia.org/wiki/Regional_indicator_symbol)" like 🇺&shy;🇸

1ec5: Oh, I wasn’t aware of the situation on Windows! Since this isn’t critical functionality, I agree that hiding it from Windows users would be fine.

In case you’d like to pursue a solution that works on Windows too, here’s an example of loading information from a preset. It gets the name of the preset, but you could make another helper function that gets the preset’s icon instead.

https://github.com/openstreetmap/iD/blob/a3990df0af8582172c3d2821b1f6441206fb6c78/modules/util/utilDisplayLabel.js#L20-L23

Since there are hundreds of countries to add flags for, we’d want to make sure that matching so many tag combinations at once wouldn’t noticeably impact performance when rendering this field. (We end up hitting this code even when hovering over an unselected embassy or flagpole feature.)

modules/ui/fields/combo.js
Razen04: Yes, languages has a chip rendering path because it is multiCombo whereas country is single combo field.

modules/ui/fields/combo.js
Razen04: Using `regionNames.of(tval)` coulg give error but if we only ran it on the codes coming from `countryCoder`, can there still be a possiblity of error? Like countryCoder can contain invalid codes?

k-yle: correct, currently the [country-coder](https://github.com/rapideditor/country-coder) library always provides valid [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) or [UN M49](https://en.wikipedia.org/wiki/UN_M49) codes. 

So this crash can only occur if a user creates an invalid tag using the raw tag editor, for example `country=A`. 

To be safe, lets move these 4 lines into a helper function with a `try...catch` block, so that we can be certain that there's no risk.

modules/ui/fields/combo.js
Razen04: If I try to add icon support since emojis won't work on Windows, I will have to add all the country icons, right? Or is there already icons present in iD? If we have to take the first route, I don't think it is worth it and the emoji will be fine. Do let me know what should I do.

Razen04:
Things done:
- Did the refactor for both language and country.
- Added emoji flags for country dropdown, if this needs to be icon do let me know.
- Some codes like FX: France don't have flags so they don't show any flag and just the name.
- Updated `CHANGELOG.md`

modules/ui/fields/combo.js
1ec5: Both of these approaches are reasonable:

* Display an emoji from country-coder instead of an image. Hide the emoji from Windows users. Seems easier.
* Display an image from a matching preset (from name-suggestion-index, but you don’t have to worry about it). Seems a little more difficult but more consistent and cross-platform.

~~And the flags are optional anyways – if you’d prefer, we can ticket out the idea and merge this PR without the flags.~~ Oh, I see you already implemented one of these options!

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 * Add Moroccan phone number formats ([#11651], thanks [@ilias52730])
 * The Languages field shows language names in your preferred language. ([#11699], thanks [@Razen04])
 * When viewing the help information for tags or presets, use locale-specific properties if available ([#11760], thanks [@k-yle])
+* Show country names in your preferred language and country flag emoji in the Country field dropdown ([#11783], thanks [@Razen04])
```
1ec5, line 57: You’ll need to define the footnote `#11783` at the bottom of the section, like the other PRs. (`@Razen04` is already defined, so that’s already taken care of.)

modules/ui/fields/combo.js
```diff
@@ -68,6 +69,57 @@ export function uiFieldCombo(field, context) {
             .join(';');
     }
 
+    // cache for language and region maps
+    let languageMap = null;
+    let regionMap = null;
+    // flag to prevent displaying emojis on a windows machine
```
1ec5: Documentation comments explain _what_; most other comments explain _why_ or _how_.

```suggestion
    // Windows intentionally does not support emoji flags
```

modules/ui/fields/combo.js
```diff
@@ -68,6 +69,57 @@ export function uiFieldCombo(field, context) {
...
+
+    // Helper to get region code map
+    const getRegionMap = () => {
+      if (regionMap) return regionMap;
+      const localeCode = localizer.localeCode();
+
+      const regionNames = new Intl.DisplayNames(localeCode, { type: 'region' });
+      const features = countryCoder.borders.features;
+
+      regionMap = new Map();
```
1ec5: A `Map` is technically correct, but some of the code using this variable is unnecessarily complex compared to what it would be if you use a simple `{}` object. The usual reason for a `Map` is when the keys need to be something other than strings. (Either way, I’d rename the variable to something like `regionsByCode` to avoid confusion with anything related to 🗺️ maps.)

modules/ui/fields/combo.js
```diff
@@ -232,6 +291,29 @@ export function uiFieldCombo(field, context) {
           return options;
         }
 
+        // Get dropdown list for country key
+        if (osmIsoCountryKeys.has(field.key)) {
+          const countryMap = getRegionMap();
+
+          let options = Array.from(countryMap.entries()).map(([c, data]) => {
```
1ec5: If `regionMap` becomes an object, then this could become something simpler like:

```js
Object.entries(countryMap).map(([code, {name, flag}]) => {
  …
});
```

modules/ui/fields/combo.js
```diff
@@ -68,6 +69,57 @@ export function uiFieldCombo(field, context) {
...
+      if (regionMap) return regionMap;
+      const localeCode = localizer.localeCode();
+
+      const regionNames = new Intl.DisplayNames(localeCode, { type: 'region' });
+      const features = countryCoder.borders.features;
+
+      regionMap = new Map();
+
+      for (const feature of features) {
+        const code = feature.properties.iso1A2;
```
1ec5, line 105: > Some codes like FX: France don't have flags so they don't show any flag and just the name.

Yes, this is a known issue: rapideditor/country-coder#25. country-coder’s maintainer says we should fall back to 🇫🇷 from the `FR` entry: https://github.com/rapideditor/country-coder/issues/157#issuecomment-3090553135. Look up the value of the `country` property, which is `FR` in this case.

modules/ui/fields/combo.js
```diff
@@ -232,6 +291,29 @@ export function uiFieldCombo(field, context) {
           return options;
         }
 
+        // Get dropdown list for country key
+        if (osmIsoCountryKeys.has(field.key)) {
+          const countryMap = getRegionMap();
+
+          let options = Array.from(countryMap.entries()).map(([c, data]) => {
+            const label = (showEmojiFlags && data.flag) ? `${data.name} ${data.flag}` : data.name;
```
1ec5: The emojis will probably look cleaner if they’re all aligned at the beginning of the label instead of appearing at the end. Hopefully putting the emoji at the beginning doesn’t affect the user’s ability to search for the item in the menu.

We have [some special styling for icons](https://github.com/openstreetmap/iD/blob/4c70f69ba894cf6a019ecdbdcbd0cba2415e43aa/modules/ui/fields/combo.js#L455-L462) that normally kicks in if you call `addComboboxIcons()`. But that function only uses images from the existing preset. Instead, you could make a separate version of the function that takes a text string (the emoji flag), then inserts spans that will match the style rules. It would look something like this, but instead of calling `svgIcon()` you can append the emoji to the span as text:

https://github.com/openstreetmap/iD/blob/4c70f69ba894cf6a019ecdbdcbd0cba2415e43aa/modules/ui/fields/combo.js#L455-L463

I don’t know for sure if that would work, but if it doesn’t, then at least it would look cleaner if you use this plain text label with the emoji at the beginning.

Razen04: I did try emoji at the beginning but that affected the search ability basically not selecting the word searched cause it was not finding the exact match due to emoji at the beginning. This is the reason I went with emoji at the end.

modules/ui/fields/combo.js
Razen04: Thanks will keep this in mind.

modules/ui/fields/combo.js
Razen04: I will try to create a different version.

Razen04:
## Updated the following things:
- Used object instead of map for better code readability.
- Used <span> class to place the emojiFlags to utilise the exisiting CSS for styling.
- Searching and Selection works fine.
- Updated `CHANGELOG.md`.

## Video:

https://github.com/user-attachments/assets/63dbcf77-cdbb-4aca-b3d6-a2ef72bbba81

modules/ui/fields/combo.js
```diff
@@ -69,56 +69,71 @@ export function uiFieldCombo(field, context) {
...
+      regionByCodes = {};
 
       for (const feature of features) {
         const code = feature.properties.iso1A2;
-        const flag = feature.properties.emojiFlag;
+        let flag = feature.properties.emojiFlag;
+        // if the flag is not present like for 'FX' code, we will look for the corresponding country flag for that code
+        if (!flag) {
+            if (feature.properties.country) {
+                  flag = countryCoder.feature(feature.properties.country).properties.emojiFlag;
```
1ec5: Nit: The usual practice in this codebase is two-space indents, but it looks like your editor wound up doing four spaces, leading to some confusing indentation. Also, you can also combine the two conditionals.

modules/ui/fields/combo.js
```diff
@@ -69,56 +69,71 @@ export function uiFieldCombo(field, context) {
...
         }
       }
 
-      return regionMap;
+      return regionByCodes;
     };
 
-    // Helper to get languageMap
+    // Helper to get languageByCodes
     const getLanguageMap = () => {
```
1ec5: Also rename these constants to match the constants being populated. This function is more substantial than just accessing an existing variable, so `buildLanguages` would be more accurate.

modules/ui/fields/combo.js
```diff
@@ -69,56 +69,71 @@ export function uiFieldCombo(field, context) {
...
+    }
+
     // cache for language and region maps
-    let languageMap = null;
-    let regionMap = null;
-    // flag to prevent displaying emojis on a windows machine
+    let languageByCodes = null;
+    let regionByCodes = null;
+    // windows does not support emoji flags
     const showEmojiFlags = utilDetect().os !== 'win';
```
1ec5: This should be above `addFlagIcon` where it’s being used.

modules/ui/fields/combo.js
```diff
@@ -69,56 +69,71 @@ export function uiFieldCombo(field, context) {
             .join(';');
     }
 
+    // adds emoji flags to country dropdown and input
+    function addFlagIcon(selection, name, flag) {
+        if (showEmojiFlags && flag) {
+            selection.insert('span', ':first-child')
+                .attr('class', 'tag-value-icon')
+                .text(flag);
```
1ec5: The emoji is misaligned:

The UN flag is offset to the upper left, leaving a large gap between it and the name for the United Nations.

The `tag-value-icon` class is designed for an image. We’re basing this code on an image-based icon that’s wrapped in `<span class="icon">`, but wrapping this text in that class won’t have the desired effect, because text can’t just be scaled by setting its width and height.

https://github.com/openstreetmap/iD/blob/585921e43fe5982d7d220514313fd64158b525ce/css/80_app.css#L1860-L1864

We’ll need a similar style rule like:

```css
.tag-value-icon .emoji {
    font-size: 24px;
    line-height: 24px;
    text-align: center;
}
```

1ec5 (commented):
Thanks for your patience. You’re pretty close, just some minor stylistic things before we’re ready to land the feature.

Razen04:
Sure I will update it.

Razen04:
I have fixed the styling but had to add:

```css
.combobox-option {
 display: flex !important;
 align-items: center !important;
}
```

cause the alignment wasn't good without this but I don't really know if this will affect something else somewhere in a negative way. Please let me know if this wasn't supposed to do.

css/80_app.css
```diff
@@ -1862,6 +1862,14 @@ input.date-selector {
     height: 21px;
     margin: auto;
 }
+.combobox-option {
+    display: flex !important;
+    align-items: center !important;
+}
```
1ec5: This seems to collapse the space between the preset name and relation name in the [relation membership](https://ideditor.netlify.app/#locale=en&map=20.43/39.28222/-84.28403&id=w38227315) dropdown menu (under the Relations section). On the other hand, you’re also fixing a preexisting issue affecting the Street/Place text box in the Address field, in countries where that field sets either `addr:street=*` or `addr:place=*`, such as [Ireland](https://ideditor.netlify.app/#locale=en&map=19.00/52.92168/-1.40503):

Before | After
----|----
Road Route US:OH:HAM | Road RouteUS:OH:HAM
Vertically misaligned place icon | Vertically aligned place icon

Flexbox layout seems to be amenable to the custom formatting that the Street/Place text box uses:

https://github.com/openstreetmap/iD/blob/585921e43fe5982d7d220514313fd64158b525ce/css/80_app.css#L2219-L2231

The relation membership dropdown menu assumes the menu item labels are inline text, but the `<b>` tag will collapse its trailing space if it’s a flexbox item:

https://github.com/openstreetmap/iD/blob/585921e43fe5982d7d220514313fd64158b525ce/modules/ui/sections/raw_membership_editor.js#L291-L300

You might try to scope the flexbox usage more tightly, audit all the other combo boxes to make sure they don’t rely on spaces to pad elements, or find a different way to align the icon and text.

1ec5 (commented):
Besides fixing the styling regression and resolving the merge conflict in the changelog, please take a moment to update the indentation in the parts of combo.js that you modified in this pull request. Thank you. That’s all the feedback I have, but I’ll let @k-yle weigh in again.

css/80_app.css
Razen04: Okay, let's see what I can do to scope it more tightly. Maybe add a special klass to them and target those

Razen04:
I added a special klass `.has-icon` to the options in dropdown for country and targeted them and also kept the address fix. I checked some combobox and they seem fine to me but do let me know if you see any issue.

The indentation issue is really annoying cause I can't even use the editor format document feature and I keep forgetting about fixing them, my apologies for this, if there are still somewhere do let me know, I have fixed it best as per my knowledge.

css/80_app.css
```diff
@@ -1857,11 +1857,20 @@ input.date-selector {
     padding-left: 0;
     padding-right: 11px;
 }
+.combobox-address-street-place .combobox-option,
+.combobox-option.has-icon {
```
1ec5, line 1861: It’s always a bit unfortunate when we have to explicitly scope a style rule by a particular field. But I’ve had you iterate on this emoji enough, even though it wasn’t in the feature request’s original scope, so in fairness we can track a more robust fix as tail work.

1ec5 (approved):

css/80_app.css
Razen04: Sure I understand the explicit scoping concern. If it is fine for now, we can move on with this.

modules/ui/fields/combo.js
k-yle: image

bit late, but I want to check if it's intentional that invalid values get the blue styling? typically these would get the black 'raw' styling.

k-yle: @copilot there's a bug on this line, open a new PR and make the following change:

```suggestion
            && !(osmIsoCountryKeys.has(field.key) && value in buildCountry());
```

Razen04: Yes it should get the raw styling, cause it is a valid thing but not conventional. Thanks for pointing it out and fixing it.

PR #11794, opened 2026-01-21
fixes rare tag value auto suggestions (fix #11782)

### LINKED ISSUE #11782: Do not suggest low-use tags just because they have wiki page
Image

suggesting this tag solely because it has wiki page makes no sense

see https://community.openstreetmap.org/t/deprecate-leisure-swing-used-6-times-worldwide/140407

previously reported as #5460

previously fixed in #5461

reintroduced in a2cacaaf2456de4c8443a7b5dab0ac3393be5deb

Kaushik4141:
Removed d.in_wiki from the filter condition for tag values in modules/services/taginfo.js
Updated test/spec/services/taginfo.js to ensure unpopular values with a wiki page are now excluded from suggestions
image

Fixes #11782 and replaces #11781

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 
 #### :sparkles: Usability & Accessibility
 * Don't suggest values from Taginfo for `addr:*` tags ([#11733], thanks [@k-yle])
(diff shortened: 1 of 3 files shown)


k-yle (approved):
thanks for your recent PRs ! this one looks good too

Kaushik4141:
> thanks for your recent PRs ! this one looks good too

Thanks glad it looks good to you

PR #11795, opened 2026-01-21
Fix search bar autofocus after drawing

### LINKED ISSUE #11793: Search bar no longer focused when exiting drawing mode
When you finish drawing a point, line, or area, the search bar should immediately receive keyboard focus so that the user can choose a preset. This seems to have regressed recently.

Steps to reproduce:

1. Click Point in the toolbar.
2. Click on the map.
3. Without clicking anywhere, press <kbd>M</kbd> to begin typing “museum”.

Expected result: “M” appears in the search bar, filtering the preset list.

Actual result: The move operation activates. Nothing is entered into the search bar.

Does not reproduce in v2.37.3. Reproduces in develop as of c0a783165ad73ccfc63880c90a7eac63cd0b8d44.

1-navneet:
Restore keyboard focus to the search bar after completing the drawing
of a point, line, or area so users can immediately filter presets.

Use the _newFeature flag to distinguish:
- New features from drawing: autofocus enabled
- Existing untagged vertices: autofocus disabled

Fixes #11793

### DIFF
--- a/modules/ui/inspector.js
+++ b/modules/ui/inspector.js
@@ -87,7 +87,7 @@ export function uiInspector(context) {
             wrap.style('right', '-100%');
             editorPane.classed('hide', true);
             presetPane.classed('hide', false)
-                .call(presetList.autofocus(false));
+                .call(presetList.autofocus(_newFeature));
         } else {
             wrap.style('right', '0%');
             presetPane.classed('hide', true);
--- a/modules/ui/inspector.js
+++ b/modules/ui/inspector.js
@@ -87,7 +87,7 @@ export function uiInspector(context) {
             wrap.style('right', '-100%');
             editorPane.classed('hide', true);
             presetPane.classed('hide', false)
-                .call(presetList.autofocus(_newFeature));
+                .call(presetList);
         } else {
             wrap.style('right', '0%');
             presetPane.classed('hide', true);


1-navneet:
Tested locally on develop and this branch.

**Before:** After finishing drawing, the search input did not regain focus.  
before-autofocus-after-drawing

**After:** Search input is automatically focused, allowing immediate preset filtering.
after-autofocus-after-drawing


Screenshots added above for reference.

modules/ui/inspector.js
```diff
@@ -87,7 +87,7 @@ export function uiInspector(context) {
             wrap.style('right', '-100%');
             editorPane.classed('hide', true);
             presetPane.classed('hide', false)
-                .call(presetList.autofocus(false));
+                .call(presetList.autofocus(_newFeature));
```
k-yle: thanks for the PR! This works, but it's a bit redundant, since we already called `presetList.autofocus(_newFeature)` up on [line 23](https://github.com/openstreetmap/iD/blob/1e38265451188f89b78f870a20d10f1a7606d67f/modules/ui/inspector.js#L23).

Instead, we could revert the union of 0026245c9ef7cd0023dccc6306df82d278d04b5e and 1a86bc5f1765b3a784218bcaf28345663b56af07, which—despite the commit description— was not an exact revert of the prior commit.

After running `git revert 002624 1a86bc` we get the following diff:


```suggestion
                .call(presetList);
```

1-navneet: Thanks for the clarification! I’ve removed the redundant autofocus call and updated the code to match the suggested diff.

k-yle (approved):
thanks! looks good to me now

1-navneet:
Thanks for the review and guidance!

PR #11807, opened 2026-01-23
Support point feature creation via query paramters

mjoe999:
I like the web editor. Not everyone is a JOSM warrior. I might have some data that I have collected and wish to curate in OSM using the web editor. I don't like typing. Perhaps I can create a series of URLs such as:

``
http://localhost:8080/?add_point=true&lat=57.1436&lon=-2.0970&name=Tasty&shop=bakery
``

It's a contrived example. More likely it would be a sequence of URLs with richer parameters, but you get the idea. I bootstrap myself into the web editor and can take it from there.

Tested on Chrome and Firefox.

My first iD PR. I gave read guidelines and hopefully done my best :-)

### DIFF
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -21,6 +21,7 @@ import { rendererBackground, rendererFeatures, rendererMap, rendererPhotos } fro
 import { services } from '../services';
 import { uiInit } from '../ui/init';
 import { utilKeybinding, utilRebind, utilStringQs, utilCleanOsmString } from '../util';
+import {osmNode} from '../osm';
 
 
 export function coreContext() {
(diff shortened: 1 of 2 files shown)


tyrasd:
Sorry for the late reply on this idea. 

It's an interesting idea, but there are some issues that speak against this particular issue:
* the integration on the OSM website is unlikely to forward such parameters to the editor
* the "remote controlling" would only work at startup, but for some use cases you would like to do this on the fly while editing.
* the current implementation of only supporting `add_point` is a bit too limited to warrant offering a whole new API

Here are some ideas to tackle this in a more generic way:
* implement a "remote control" API via [`postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage), such that a tool can more closely interact with an instance of iD that is running in an (e.g.) iframe. This might be useful for maproulette-stlye tools
* interact better with the OS clipboard: when copying: put a json representation of the OSM entity into the clipboard, when pasting use the clipboard content. Then, external tools can also put some custom tags/geometry into the clipboard for the user to be able to easily paste it.

PR #11809, opened 2026-01-25
Add standard keyboard scrolling to member list

1-navneet:
### Summary
Adds standard keyboard scrolling behavior to the member list so arrow and page navigation keys scroll the list instead of moving the map.

### Changes
- Enable keyboard focus on member list
- Handle ArrowUp / ArrowDown / PageUp / PageDown / Home / End
- Prevent map movement while list is focused
- Add visible focus outline for accessibility

### Before
- Arrow and page keys moved the map instead of scrolling the relations list.
before

### After
- Relations list captures keyboard focus and scrolls independently without moving the map.
after

### Notes
Improves keyboard accessibility and aligns the member list with standard sidebar scrolling behavior.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2877,6 +2877,13 @@ img.tag-reference-wiki-image {
 .section-raw-membership-editor .member-list {
     position: relative; /* required for drag-and-drop */
     padding-top: 5px;
+    overflow-y: auto;
+    max-height: 400px;
+}
+.section-raw-member-editor .member-list:focus,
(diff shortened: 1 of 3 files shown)


k-yle:
close/reopen to re-trigger PR preview

matkoniecz:
Yes, it adds nested scrolling. But why we would want nested scrolling?

tyrasd:
This change will not improve the usability of iD.

PR #11810, opened 2026-01-25
Prevent unused slots in the recent preset list.

### LINKED ISSUE #11405: Presets for motorway link, trunk link, primary link won’t show up as recently used presets in some regions
### URL

(you can test it anywhere)

1. Select a way or draw a new one.
2. Select four different presets to fill the four “recently used” slots.
    * e.g. Motorway, Trunk Road, Primary Road, Secondary Road
3. One after another, select all five link road types from the “Major Roads…” preset group.

After these steps, only the Secondary Link and Tertiary Link presets are shown in my recently used presets – Trunk Link and Primary Link occupy two “invisible” slots:
Only 2 recent presets entries: Secondary Link, Tertiary Link

* The presets Motorway Link, Trunk Link, Primary Link don’t show up as recently used presets, but they still take up slots.
* Notably, the issue doesn’t occur for me on the development version, but I was unable to find a similar issue report, so I don’t think it’s really fixed.
    * Could this be caused by some kind of `localStorage` corruption?

2.36.0

Firefox

Razen04:
Fixes #11405 

Filtering before slicing prevents the unused slots in the recent preset list, more detail about this in #11405 

If we wish to show the Motorway Link and others in the US and CA in the recent list, we can just prevent filtering the recent list based on location, I have already commented in the issue everything [here](https://github.com/openstreetmap/iD/issues/11405#issuecomment-3796540428).

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -410,9 +410,14 @@ export function presetIndex() {
 
 
   _this.defaults = (geometry, n, startWithRecents, loc, extraPresets) => {
+    const validHere = Array.isArray(loc) ? locationManager.locationSetsAt(loc) : null;
+
     let recents = [];
     if (startWithRecents) {
-      recents = _this.recent().matchGeometry(geometry).collection.slice(0, MAX_RECENTS_TO_SHOW);
+        // filtering before slicing to prevent unused slots in the recent preset list, issue #11405
+        recents = _this.recent().matchGeometry(geometry).collection
+        .filter(a => !a.locationSetID || (validHere && validHere[a.locationSetID]))
+        .slice(0, MAX_RECENTS_TO_SHOW);
     }
 
     let defaults;
@@ -421,7 +426,9 @@ export function presetIndex() {
         var preset = _this.item(id);
         if (preset && preset.matchGeometry(geometry)) return preset;
         return null;
-      }).filter(Boolean);
+      })
+        .filter(Boolean)
+        .filter(a => !a.locationSetID || validHere[a.locationSetID]);
     } else {
       defaults = _defaults[geometry].collection.concat(_this.fallback(geometry));
     }
@@ -429,11 +436,6 @@ export function presetIndex() {
     let result = presetCollection(
       utilArrayUniq(recents.concat(defaults).concat(extraPresets || [])).slice(0, n - 1)
     );
-
-    if (Array.isArray(loc)) {
-      const validHere = locationManager.locationSetsAt(loc);
-      result.collection = result.collection.filter(a => !a.locationSetID || validHere[a.locationSetID]);
-    }
 
     return result;
   };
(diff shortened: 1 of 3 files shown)


matkoniecz:
Can you add test so potential regression would be avoided?

Razen04:
Sure, I will do it rn.

Razen04:
It's really weird that Github gave attribute to some random account, I am trying to fix the commits because it contained my personal name which I don't wish to have. But Github attributed the commits to someone random person, I think I fixed it for now.

Anyways I have added the test case, never done before so maybe it is wrong do let me know that.

matkoniecz (commented):
when you select "motorway link" from major roads folder it will not appear in recent presets


yes, it is no longer blocking history slot but it is still a bug

Razen04:
That is what's supposed to happen if we are in US or maybe CA too, OSM prevents those from appearing. I think it is intentionally done

matkoniecz:
no, it is also clearly part of the same bug and unwanted behavior resulting from overcomplex area-specific presets

Razen04:
Ohhh, I thought that was intentional, fine then I will just make it not check the location maybe and bypass that. But then why is the data containing that, as mentioned in the issue by me?

matkoniecz:
randomly hiding motorway link from recently used entries is definitely not deliberate

it is side effect of having special motorway link entry for USA, maybe also other areas (which was well intentioned solution for some other problem but causing trouble due to increased complexity)

matkoniecz:
Looking at it again: does NOT entirely fix linked issue, but should work for cases where user edited one area then moved to another. So worth merging anyway, I guess?

modules/presets/index.js
```diff
@@ -406,9 +406,14 @@ export function presetIndex() {
...
   _this.defaults = (geometry, n, startWithRecents, loc, extraPresets) => {
+    const validHere = Array.isArray(loc) ? locationManager.locationSetsAt(loc) : null;
+
     let recents = [];
     if (startWithRecents) {
-      recents = _this.recent().matchGeometry(geometry).collection.slice(0, 4);
+      // filtering before slicing to prevent unused slots in the recent preset list, issue #11405
+      recents = _this.recent().matchGeometry(geometry).collection
+        .filter(a => !a.locationSetID || validHere[a.locationSetID])
+        .slice(0, 4);
```
matkoniecz: we changed how many recent presets are listed from 4 to defined constant and this code also should use it now

tyrasd: see #9545 

@ Razen04 can you please rebase this branch to the latest development branch?

matkoniecz: @Razen04 fixed ping

Razen04: Yeah let me do it

Razen04:
I have updated the code to work with `MAX_RECENTS_TO_SHOW` and also rebased it.

matkoniecz (commented):
confirmed that it works

having motorway link from folder eats you one history entry in iD version on osm.org, not here

BTW, I created https://github.com/openstreetmap/iD/issues/12044 as this problem is a bit wider, so https://github.com/openstreetmap/iD/issues/11405 can be closed

matkoniecz:
alternative fix would be to magically replace US-trunk link by regular trunk link and vice versa, but there is no reasonable way to achieve this and may be confusing

admittedly, disappearing trunk link is also confusing (this PR fixes not disappearance but failing to use one of recent entries slots)

tyrasd (approved):

PR #11817, opened 2026-01-27
persist overlay layer states across sessions fixes(#11777)

### LINKED ISSUE #11777: Locator overlay do not activate after closing previous sesion
### URL

https://www.openstreetmap.org/edit?way=1467196725#map=16/-19.19231/-68.57770

1) Just open a session, 
2) go to the locator overlay (activate or deactivate)
3) close the session with the locator activated
4) once opened the locator overlay looks like deactivated, so needs to be re-activated manually, instead of keep from the previous session.

Sometimes the same happen with the GPS overlay

As descripted above

_No response_

2.37.3

_No response_

Kaushik4141:
## changed 

save all enabled overlay IDs to `background-overlay-states` preference
Restore them on init
Added `_initComplete` flag to fix a race condition during startup

First-time users still get the default locator overlay enabled

image

https://github.com/user-attachments/assets/8a36d001-b82e-40c1-9aa9-73ec51806572

fixes #11777

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -23,6 +23,7 @@ export function rendererBackground(context) {
   let _checkedBlocklists = [];
   let _isValid = true;
   let _overlayLayers = [];
+  let _initComplete = false;
   let _brightness = 1;
   let _contrast = 1;
   let _saturation = 1;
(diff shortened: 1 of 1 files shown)


tordans:
At first glance the changes here look too complex to me. To many new if's that we need to maintain.
Ideally, the complexity of managing which settings "win" (URL vs. local storage vs. defaults) should be abstracted aways, which AFAIK is what https://github.com/openstreetmap/iD/issues/11305  is looking into.

Kaushik4141:
> At first glance the changes here look too complex to me. To many new if's that we need to maintain. Ideally, the complexity of managing which settings "win" (URL vs. local storage vs. defaults) should be abstracted aways, which AFAIK is what #11305 is looking into.

You're right that the priority logic adds complexity
I looked at PR #11239 for #11305  it currently has merge conflicts and test failures (`services.osm.on is not a function`). 
@Deeptanshu-sankhwar are you still working on it? I'd be happy to help fix the issues and get it moving, or take it over if needed.

I will close this PR for now so we can focus on the better solution in #11305

matkoniecz:
> I will close this PR for now so we can focus on the better solution in https://github.com/openstreetmap/iD/issues/11305

can you close it then?

PR #11818, opened 2026-01-27
preserve natural=coastline on outer ways during Merge operation (Fixes #11816)

### LINKED ISSUE #11816: `natural=coastline` tag should remain on outer way when multipolygon is created using "merge" operation
Similar to #9563, the fix from 62e171401b83d2805974fba987fa77bc558d67e5 should also be applied to the _Merge_ operation when it creates a multipolygon from one outer and one or more inner way(s).

Kaushik4141:
this applies the same fix logic from 62e1714 to the merge operation

## fix
Now, when coastline ways are merged into a multipolygon, the natural=coastline tag stays on the outer ways rather than being shifted to the relation, as recommended by the OSM wiki

image

## changes

Modified [modules/actions/merge_polygon.js](cci:7://file:///d:/OpenSource/iD/modules/actions/merge_polygon.js:0:0-0:0) to extract `natural=coastline` before merging tags

Added a regression test in [test/spec/actions/merge_polygon.js](cci:7://file:///d:/OpenSource/iD/test/spec/actions/merge_polygon.js:0:0-0:0) to verify that `natural=coastline` stays on the way while other tags move to the relation.

Fixes #11816


test/spec/actions/merge_polygon.js
```diff
@@ -162,4 +162,16 @@ describe('iD.actionMergePolygon', function () {
         expect(find(r, 'w3').role).to.equal('inner');
         expect(find(r, 'w4').role).to.equal('inner');
     });
+
+    it('preserves coastline tag on the outer ways when creating a multipolygon', function() {
+        graph = graph.replace(graph.entity('w0').update({ tags: { 'natural': 'coastline', 'place': 'island' }}));
```
matkoniecz: can we choose another example?

`place=island` on multipolygon is quite dubious at best and is confusing what is going on

I would suggest `leisure=nature_reserve`

modules/actions/merge_polygon.js
```diff
@@ -107,8 +107,15 @@ export function actionMergePolygon(ids, newRelationId) {
                 return m.id === way.id && m.role !== 'inner';
             }
             if (members.some(isThisOuter)) {
-                relation = relation.mergeTags(way.tags);
-                graph = graph.replace(way.update({ tags: {} }));
+                //separate area tags from line-only tags (like natural=coastline)
```
matkoniecz: 1) this comment claims that it is generally working for "line-only tags (like natural=coastline)" but it actually considers solely and `natural=coastline` and nothing else

either it should be made more general (not requested, not sure to which other tags it should apply)

or it should be phrased in less misleading way

2) would it make sense and be possible to store list of way-only tags in one place and use it here and in https://github.com/openstreetmap/iD/commit/62e171401b83d2805974fba987fa77bc558d67e5 code? So if future more tags expecting such handling would be added then it would be impossible to add them in one place only

test/spec/actions/merge_polygon.js
Kaushik4141: I’ve swapped the test example to `leisure=nature_reserve` so it’s less confusing
Thanks

modules/actions/merge_polygon.js
Kaushik4141: I’ve refactored this into a central `osmWayOnlyTags`  list in `tags.js` and updated the split action as well so now it’s consistent

matkoniecz (commented):
works as far as I can see :)

tyrasd (approved):
Thanks! Refactoring out the tag to `osm/tags.js` was a good idea to make this more maintainable in the future.

PR #11822, opened 2026-01-27
#11775 Validation: Remove layer fixes for building overlaps, encourage geometry fix instead

### LINKED ISSUE #11775: Adding higher or lower  to overlapping buildings
### URL

_No response_

In iD Editor draw one area and tag it as building=yes. Draw another area that overlaps the first area and tag it as building=yes. Select the outline of one of the buildings.
The yellow  Issues warning suggest **Tag this as higher** or **Tag this as lower**  instead of resolving the issue of the overlap.
Many inexperienced mappers click on one of those suggestions which then adds a plus layer or minus layer to the building but does not resolve the issue of the overlapping buildings.
This has resulted in the use of a Map Paint Style (layer checker) to identify the buildings with a layer tag so that we can remove the layer tag from the building and then adjust the geometry of the buildings (if necessary) so they do not overlap.
Remember that the recommendation for buildings is to use **level** instead of layer tags.

Image

Here is the result of that yellow Issue 
Buildings have +1 or -1 layer added and the main highway through the town has a -1 layer. All of this to avoid the validator showing overlapping buildings.
Image
The JOSM Validation function does not pick up that these are still overlapping buildings but it does pick up that the highway has intersections on different layers.

_No response_

_No response_

ak8abhinay:
This PR modifies the crossing_ways validation logic to prevent the iD Editor from suggesting "Tag as higher" or "Tag as lower" (which adds layer=*) when two building features overlap.

The Problem Currently, mappers often use the "Quick Fix" layer buttons to silence validation warnings for overlapping buildings. This creates "layer soup"—data where buildings are incorrectly marked as floating in the air. Per community best practices and my recommendation [2026-01-15], buildings should stay on layer=0. If they are physically stacked, the level tag is the correct way to map them.

The Fix

Logic: Updated modules/validations/crossing_ways.js to identify building-on-building overlaps and suppress the changeLayer fixes.

Education: Updated the validation message in core.yaml to explicitly guide mappers toward using the level tag instead of layer.

Validation

Verified that "Tag as higher/lower" buttons are hidden for building-building crossings.

Verified that these buttons still appear for valid cases, such as bridges or tunnels (non-building crossings)

Closes #11775

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1811,7 +1811,8 @@ en:
       aeroway-waterway:
         reference: "Aeroways crossing waterways should use tunnels."
       building-building:
-        reference: "Buildings should not intersect except on different layers."
+        message: "Buildings should not intersect. If they are vertically stacked, please use the level tag instead of layer."
+        reference: "Buildings should not intersect. If they are vertically stacked, please use the **level** tag instead of **layer**."
       building-highway:
         reference: "Highways crossing buildings should use bridges, tunnels, or different layers."
       building-railway:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1811,8 +1811,7 @@ en:
       aeroway-waterway:
         reference: "Aeroways crossing waterways should use tunnels."
       building-building:
-        message: "Buildings should not intersect. If they are vertically stacked, please use the level tag instead of layer."
-        reference: "Buildings should not intersect. If they are vertically stacked, please use the **level** tag instead of **layer**."
+        reference: "Buildings should not intersect. Please fix the geometry of the overlapping buildings."
       building-highway:
         reference: "Highways crossing buildings should use bridges, tunnels, or different layers."
       building-railway:
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1811,7 +1811,7 @@ en:
       aeroway-waterway:
         reference: "Aeroways crossing waterways should use tunnels."
       building-building:
-        reference: "Buildings should not intersect. Please fix the geometry of the overlapping buildings."
+        reference: "Buildings should not intersect. Please fix the geometry of the overlapping buildings. In rare cases of actually overlapping buildings set layer= tags."
       building-highway:
         reference: "Highways crossing buildings should use bridges, tunnels, or different layers."
       building-railway:
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -89,21 +89,36 @@ export function validationCrossingWays(context) {
             return true;
         }
 
+        // assume 0 by default; don't use way.layer() since we account for structures here
+        var layer1 = tags1.layer || '0';
+        var layer2 = tags2.layer || '0';
+
+        if (allowsBridge(featureType1) && allowsBridge(featureType2)) {
+            if (hasTag(tags1, 'bridge') && !hasTag(tags2, 'bridge')) return true;
+            if (!hasTag(tags1, 'bridge') && hasTag(tags2, 'bridge')) return true;
+            // crossing bridges must use different layers
+            if (hasTag(tags1, 'bridge') && hasTag(tags2, 'bridge') && layer1 !== layer2) return true;
+        } else if (allowsBridge(featureType1) && hasTag(tags1, 'bridge')) return true;
+        else if (allowsBridge(featureType2) && hasTag(tags2, 'bridge')) return true;
+
+        if (allowsTunnel(featureType1) && allowsTunnel(featureType2)) {
+            if (hasTag(tags1, 'tunnel') && !hasTag(tags2, 'tunnel')) return true;
+            if (!hasTag(tags1, 'tunnel') && hasTag(tags2, 'tunnel')) return true;
+            // crossing tunnels must use different layers
+            if (hasTag(tags1, 'tunnel') && hasTag(tags2, 'tunnel') && layer1 !== layer2) return true;
+        } else if (allowsTunnel(featureType1) && hasTag(tags1, 'tunnel')) return true;
+        else if (allowsTunnel(featureType2) && hasTag(tags2, 'tunnel')) return true;
+
         // don't flag crossing waterways and pier/highways
         if (featureType1 === 'waterway' && featureType2 === 'highway' && tags2.man_made === 'pier') return true;
         if (featureType2 === 'waterway' && featureType1 === 'highway' && tags1.man_made === 'pier') return true;
 
-        if (tags1.layer !== undefined && tags1.layer === tags2.layer) return false; // Warn if both have the same defined layer
-
-        const isElement1Bridge = allowsBridge(featureType1) && hasTag(tags1, 'bridge');
-        const isElement2Bridge = allowsBridge(featureType2) && hasTag(tags2, 'bridge');
-        if (isElement1Bridge !== isElement2Bridge) return true; // Either one is bridge, the other is not
-
-        const isElement1Tunnel = allowsTunnel(featureType1) && hasTag(tags1, 'tunnel');
-        const isElement2Tunnel = allowsTunnel(featureType2) && hasTag(tags2, 'tunnel');
-        if (isElement1Tunnel !== isElement2Tunnel ) return true; // Either one is tunnel, the other is not
-
-        return (tags1.layer || '0') !== (tags2.layer || '0');
+        if (featureType1 === 'building' || featureType2 === 'building' ||
+            taggedAsIndoor(tags1) || taggedAsIndoor(tags2)) {
+            // for building crossings, different layers are enough
+            if (layer1 !== layer2) return true;
+        }
+        return false;
     }
 
 
@@ -404,6 +419,7 @@ export function validationCrossingWays(context) {
                                 allowsTunnel(featureType2) && hasTag(entities[1].tags, 'tunnel');
         var isCrossingBridges = allowsBridge(featureType1) && hasTag(entities[0].tags, 'bridge') &&
                                 allowsBridge(featureType2) && hasTag(entities[1].tags, 'bridge');
+        var isBothBuildings = featureType1 === 'building' && featureType2 === 'building';
 
         var subtype = [featureType1, featureType2].sort().join('-');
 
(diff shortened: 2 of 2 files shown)


data/core.yaml
```diff
@@ -1811,7 +1811,8 @@ en:
       aeroway-waterway:
         reference: "Aeroways crossing waterways should use tunnels."
       building-building:
-        reference: "Buildings should not intersect except on different layers."
+        message: "Buildings should not intersect. If they are vertically stacked, please use the level tag instead of layer."
+        reference: "Buildings should not intersect. If they are vertically stacked, please use the **level** tag instead of **layer**."
```
matkoniecz: why? If they are in fact vertically stacked buildings then `layer` is appropriate.

`level` is for this on different levels (and BTW things on single level may be still vertically stacked)

modules/validations/crossing_ways.js
```diff
@@ -475,8 +491,12 @@ export function validationCrossingWays(context) {
                     featureType1 === 'building' ||
                     featureType2 === 'building')  {
 
-                    fixes.push(makeChangeLayerFix('higher'));
-                    fixes.push(makeChangeLayerFix('lower'));
+                    // For overlapping buildings, do not suggest changing the `layer=*` tag.
+                    // Vertically stacked buildings should use `level=*` instead.
```
matkoniecz: at least also here

data/core.yaml
matkoniecz: is wiki advising this anywhere?

ak8abhinay: the [Key:layer Wiki page](https://wiki.openstreetmap.org/wiki/Key:layer) advises for buildings in the final para :

'Ways in buildings... should be mostly described with level= instead of layer.'

After looking at the review, I've realized my current code logic is incomplete. While I've updated the message, the validation still incorrectly allows layer as a fix, which is inconsistent. As these changes were made some time ago, I’d like to take some time to properly review and adjust the implementation.

matkoniecz: > 'Ways in buildings... should be mostly described with level= instead of layer.'

it refers to ways in buildings, not to buildings themselves

modules/validations/crossing_ways.js
ak8abhinay: I added the check if (!isBothBuildings) here to specifically disable the 'Make Higher' and 'Make Lower' layer buttons for building-to-building overlaps.

By removing these buttons, we stop the editor from suggesting layer as an automated fix and encourage mappers to handle manual fix or fix the geometry. However, I've realized my current logic is a bit inconsistent since the validator still technically allows the tag, so I will  review this and come up with proper implementation

data/core.yaml
matkoniecz: I am not entirely sure is `layer=` proper solution of vertically stacked buildings*, but `level=` is surely not - as it refers to position within a building.

*this is distinct form whether pushing mapper to add fake `layer=` instead of fixing almost certainly bad geometries is appropriate - hiding "just add layers" when both are buildings makes sense to me

ak8abhinay: I understand now. I’ll re-evaluate how overlapping building features should be handled in the validator and adjust the logic accordingly.

ak8abhinay:
What I changed:

Hidden "Make Higher/Lower" buttons for building and indoor feature overlaps
layer=* no longer resolves building overlap warnings, warning stays until geometry is fixed
Updated reference message, removed "except on different layers"
Removed incorrect level=* mention from comments

data/core.yaml
```diff
@@ -1811,7 +1811,7 @@ en:
       aeroway-waterway:
         reference: "Aeroways crossing waterways should use tunnels."
       building-building:
-        reference: "Buildings should not intersect except on different layers."
+        reference: "Buildings should not intersect. Please fix the geometry of the overlapping buildings."
```
matkoniecz:
```suggestion
        reference: "Buildings should not intersect. Please fix the geometry of the overlapping buildings. In rare cases of actually overlapping buildings set layer= tags."
```

modules/validations/crossing_ways.js
```diff
@@ -89,21 +89,37 @@ export function validationCrossingWays(context) {
...
-        if (isElement1Bridge !== isElement2Bridge) return true; // Either one is bridge, the other is not
-
-        const isElement1Tunnel = allowsTunnel(featureType1) && hasTag(tags1, 'tunnel');
-        const isElement2Tunnel = allowsTunnel(featureType2) && hasTag(tags2, 'tunnel');
-        if (isElement1Tunnel !== isElement2Tunnel ) return true; // Either one is tunnel, the other is not
-
-        return (tags1.layer || '0') !== (tags2.layer || '0');
+        if (featureType1 === 'building' || featureType2 === 'building' ||
+            taggedAsIndoor(tags1) || taggedAsIndoor(tags2)) {
+             // For overlapping buildings, layer=* is not a valid resolution.
```
matkoniecz: actually, in some cases it is - where buildings truly overlap and one goes over another

https://commons.wikimedia.org/wiki/File:20200826_Cricoteka_w_Krakowie_1819_1345.jpg

ak8abhinay: I understand now that if two buildings are on different layers (as reference image), then `layer=*` is a valid tag to differentiate them. So I will update code to allow the layer tag validation when layer1 !== layer2 for buildings.

ak8abhinay: If this approach is good then I will push the code.

matkoniecz: Yes, iD should stop pushing layer= tags with easy to click buttons (as it mostly masks actual errors) but if present they should be respected

matkoniecz:
CI reports failed tests (not checked whether tests should be changed or have they found bugs)

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 3 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/spec/validations/crossing_ways.js > iD.validations.crossing_ways > ignores road crossing road on different layers
AssertionError: expected [ validationIssue{ …(14) }, …(1) ] to have a length of +0 but got 2

- Expected
+ Received

- 0
+ 2

 ❯ test/spec/validations/crossing_ways.js:154:32
    152|         createWaysWithOneCrossingPoint({ highway: 'path', layer: '-5' …
    153|         var issues = validate();
    154|         expect(issues).to.have.lengthOf(0);
       |                                ^
    155|     });
    156| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/3]⎯

 FAIL  test/spec/validations/crossing_ways.js > iD.validations.crossing_ways > flags road bridge crossing road on the same layer
AssertionError: expected [] to have a length of 2 but got +0

- Expected
+ Received

- 2
+ 0

 ❯ verifySingleCrossingIssue test/spec/validations/crossing_ways.js:74:32
     72|     function verifySingleCrossingIssue(issues, connectionTags) {
     73|         // each entity must produce an identical issue
     74|         expect(issues).to.have.lengthOf(2);
       |                                ^
     75|         expect(issues[0].id).to.eql(issues[1].id);
     76| 
 ❯ test/spec/validations/crossing_ways.js:396:9

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[2/3]⎯

 FAIL  test/spec/validations/crossing_ways.js > iD.validations.crossing_ways > flags road tunnel crossing road on the same layer
AssertionError: expected [] to have a length of 2 but got +0

- Expected
+ Received

- 2
+ 0

 ❯ verifySingleCrossingIssue test/spec/validations/crossing_ways.js:74:32
     72|     function verifySingleCrossingIssue(issues, connectionTags) {
     73|         // each entity must produce an identical issue
     74|         expect(issues).to.have.lengthOf(2);
       |                                ^
     75|         expect(issues[0].id).to.eql(issues[1].id);
     76| 
 ❯ test/spec/validations/crossing_ways.js:411:9

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[3/3]⎯


 Test Files  1 failed | 132 passed | 4 skipped (137)
      Tests  3 failed | 2256 passed | 15 skipped | 5 todo (2279)
   Start at  10:23:25
   Duration  10.05s (transform 4.60s, setup 7.10s, import 1.42s, tests 16.05s, environment 131.61s)


Error: AssertionError: expected [ validationIssue{ …(14) }, …(1) ] to have a length of +0 but got 2

- Expected
+ Received

- 0
+ 2

 ❯ test/spec/validations/crossing_ways.js:154:32



Error: AssertionError: expected [] to have a length of 2 but got +0

- Expected
+ Received

- 2
+ 0

 ❯ verifySingleCrossingIssue test/spec/validations/crossing_ways.js:74:32
 ❯ test/spec/validations/crossing_ways.js:396:9



Error: AssertionError: expected [] to have a length of 2 but got +0

- Expected
+ Received

- 2
+ 0

 ❯ verifySingleCrossingIssue test/spec/validations/crossing_ways.js:74:32
 ❯ test/spec/validations/crossing_ways.js:411:9


ERROR: "test:spec" exited with 1.
Error: Process completed with exit code 1.
```

ak8abhinay:
Fixed the crossing logic and tested locally,
All tests are passing now.

modules/validations/crossing_ways.js
```diff
@@ -283,7 +303,8 @@ export function validationCrossingWays(context) {
                 way2FeatureType = getFeatureType(taggedFeature2, graph);
 
                 if (way2FeatureType === null ||
-                    isLegitCrossing(taggedFeature1.tags, way1FeatureType, taggedFeature2.tags, way2FeatureType)) {
+                    isLegitCrossing(taggedFeature1.tags, way1FeatureType, taggedFeature2.tags, way2FeatureType,
```
matkoniecz, line 306: wait, why passing `taggedFeature1.tags` is insufficient?

modules/validations/crossing_ways.js
```diff
@@ -78,32 +78,54 @@ export function validationCrossingWays(context) {
     }
 
 
-    function isLegitCrossing(tags1, featureType1, tags2, featureType2) {
+    function isLegitCrossing(tags1, featureType1, tags2, featureType2, wayTags1, wayTags2) {
 
+        wayTags1 = wayTags1 || tags1;
```
matkoniecz, line 83: if those are wayTags1, is it going to work as expected with multipolygons?

if those are object tags not way tags, then better name would be better

(for whatever purpose we need tags1 and wayTags1)

modules/validations/crossing_ways.js
```diff
@@ -78,32 +78,54 @@ export function validationCrossingWays(context) {
...
+        wayTags2 = wayTags2 || tags2;
+        
         // assume 0 by default
-        var level1 = tags1.level || '0';
-        var level2 = tags2.level || '0';
+        var level1 =  wayTags1.level || '0';
```
matkoniecz, line 87: why we need to check wayTags1 not tags1 ?

PR #11823, opened 2026-01-27
fixing false positive for emergency=* on highways (fix#11806)

### LINKED ISSUE #11806: Validator suggests to remove emergency=yes from highway=*
### URL

https://www.openstreetmap.org/way/1334584383

Validator suggests to remove `emergency=yes` tag from `highway=*` way.

`emergency=yes` can be used as [access tag](https://wiki.openstreetmap.org/wiki/Key:emergency#As_access_tag) on highway=*.

Image

2.37.3

_No response_

Kaushik4141:
fixes #11806

the validator was incorrectly flagging emergency=* on highway=* and railway=* features
this PR adds a special case so emergency=yes/no/designated is treated as a valid access tag instead of a mismatch

image
image
image

### DIFF
--- a/modules/validations/mismatched_geometry.js
+++ b/modules/validations/mismatched_geometry.js
@@ -267,6 +267,10 @@ export function validationMismatchedGeometry() {
             // special case: buildings-as-points are discouraged by iD, but common in OSM, so ignore them
             if (primaryKey === 'building') return false;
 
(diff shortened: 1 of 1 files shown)


k-yle:
please see https://github.com/openstreetmap/iD/issues/11806#issuecomment-3907930272

PR #11831, opened 2026-01-29
Fix/raw tag editor 255 char warning

### LINKED ISSUE #10241: Tag values over 255 characters long are silently truncated in raw tag editor
### URL

https://www.openstreetmap.org/node/11829238608/history

Enter a tag value longer then 255 characters, as tag value for some key and then upload the changeset.

They value of the tag will be truncated after upload.

This easily occurs when writing complex `opening_hours=*` with comments, as common e.g. for German physicians (after I hit the 255 char limit limit the 2nd time, I report it now). You can also look at the the [German opening_hours=*](https://wiki.openstreetmap.org/wiki/DE:Key:opening_hours)-page ([machine translated](https://wiki-openstreetmap-org.translate.goog/wiki/DE:Key:opening_hours?_x_tr_sl=de&_x_tr_tl=en&_x_tr_hl=de&_x_tr_pto=wapp)), if you need some more complex rule examples.

The user should be notified about this and the actual length of the values over the 255 character limit should be displayed in some kind to help to shorten the comments or remove some whitespaces.

2.29.0

Firefox

JaiswalShivang:
Fixes #10241

## Summary
Added visual warning in raw tag editor when tag values exceed 255 characters.

## Problem
Tag values > 255 chars were silently truncated on upload. Preset fields show warnings, but raw tag editor did not.

## Solution
- Red border + highlighted background when value > 255 chars
- Real-time warning as user types (using `input` event)
- Reused `utilUnicodeCharsCount()` and `context.maxCharsForTagValue()`

## Files Changed
- [modules/ui/sections/raw_tag_editor.js] - Added warning logic
- [css/80_app.css] - Added warning styling
- [CHANGELOG.md] - Added entry

## Screenshot
image


tordans:
Your screenshot shows a red border without a tooltip and the border is cut of at the top.
That looks not great visually and I don't understand how a user will understand what is going on.

JaiswalShivang:
I will improve the ui the red box suggest that the text is gone above limits , In order to make understand to user i will add a message just below that "Tag value exceed 255 characters"

PR #11833, opened 2026-01-30
Fix and improve extract logic

### LINKED ISSUE #11830: bad extract of a swimming pool
### URL

https://www.openstreetmap.org/edit?way=1394540738#map=21/43.5739789/6.1844197

- follow the link above
- right click on the swimming pool and select "extract"
- as a result we have a node with the tag swimming_pool and an area with the tag "area=yes" + a red error "area has no description tag".

The extraction is clearly useless.

_No response_

2.37.3

Firefox

ak8abhinay:
I noticed that the current "Extract" operation can be a bit destructive. When you extract a named feature (like pool) from an area, it often wipes the original area clean, leading to "Area has no descriptive tags" validation errors.

This PR updates the tool about what it moves and what it keeps. It also ensures the tool doesn't appear for standalone objects where extraction wouldn't make sense.

Changes
actions/extract.js file : I added a list of identityKeys (like amenity, leisure, shop, etc.) that stay on the original area after extraction. This prevents the area from becoming "naked" and keeps it valid on the map.

Protecting the level tag and it is now kept on the original geometry rather than being moved to the point.

operations/extract.js file : Modified the tool's availability so it hides itself if an object has no parent ways or no other tags. This prevents users from accidentally turning a standalone building into a single point.

The new extracted node now receives a copy of the identity tags, ensuring it has the correct icon and remains searchable in the editor.

I tested everything locally and above mentioned solutions are working perfectly in the editor.

Fixes #11830

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -806,7 +806,7 @@ a.hide-toggle {
 }
 
 
-/* Sidebar / Inspector
+/* Sidebar / Inspector 
 ------------------------------------------------------- */
 .sidebar {
     position: relative;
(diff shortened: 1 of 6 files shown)


tordans:
This has unrelated changes in CSS and IMO we need to first discuss the issue more in https://github.com/openstreetmap/iD/issues/11830

ak8abhinay:
> This has unrelated changes in CSS and IMO we need to first discuss the issue more in #11830

@tordans,
Please disregard this PR. I've realized this branch accidentally included unrelated changes from a previous task, which is why the CSS changes are appearing. I’ve added my thoughts on the implementation to issue #11830,  let's discuss it there! Once we're ready, I will submit a fresh, clean PR with only the relevant file changes. Thanks!

Issue #7710, opened 2020-06-16
Don't redraw everything when changing modes


quincylvania:
Currently, iD will schedule a re-render of all features when changing modes, e.g. when selecting or deselecting something. This can make iD noticeably sluggish when editing with lots of features onscreen. We should make a point to re-render only features that need style updates.

PR #11834, opened 2026-01-30
Improve zoom/pan performance by avoiding redundant tag class and label width recomputation

### LINKED ISSUE #11832: Zoom and pan performance issues
### URL

https://www.openstreetmap.org/edit#map=18/53.382847/-1.470167

drawEditable when zooming and panning can take almost 3 seconds. I've included a trace and a screenshot. A lot of this seems to be spent calculating text widths and in `tagClasses`

Operating system: CachyOS 6.18.7-2
Browser: Chrome 144.0.7559.109 And Firefox 147.0.2

Image

[Trace-20260129T223514.json.gz](https://github.com/user-attachments/files/24949787/Trace-20260129T223514.json.gz)

2.37.3

Chrome

1-navneet:
This PR addresses the zoom and pan performance issue described in #11832.

### Problem
Profiling during zoom/pan interactions showed significant time spent repeatedly:
- Computing SVG tag class strings for entities whose tags were unchanged
- Measuring SVG text widths during label redraws

These operations were executed on every redraw, even when the underlying data did not change.

### Solution
This change reduces redundant work during redraws by:
- Reusing cached tag class strings when entity tags are unchanged
- Reusing a persistent SVG text measurer per font size to avoid repeated DOM create/remove cycles

### Result
- Reduced time spent in drawEditable during zoom/pan interactions
- Hotspots such as `tagClasses.getClassesString` no longer dominate redraw time
- No visual or behavioral changes observed

### Testing
- Tested locally using Chrome DevTools Performance profiling
- Reproduced using the same location and zoom level as the issue report
  (zoom level 18 near 53.382847, -1.470167)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,11 +58,8 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * When viewing the help information for tags or presets, use locale-specific properties if available ([#11760], thanks [@k-yle])
 * Show country names in your preferred language and country flag emoji in the Country field dropdown ([#11783], thanks [@Razen04])
 #### :hourglass: Performance
-
 * iD is now twice as fast during long editing sessions ([#11861], thanks [@k-yle])
-
 * Improve zoom/pan performance by avoiding redundant tag class recomputation ([#11834], thanks [@1-navneet])
-
 #### :mortar_board: Walkthrough / Help
 #### :rocket: Presets
 * Add dedicated styling for `highway=ladder` to make it distinguishable from `highway=steps` ([#11799], thanks [@bhavyaKhatri2703])
@@ -89,22 +86,16 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#11783]: https://github.com/openstreetmap/iD/pull/11783
 [#11861]: https://github.com/openstreetmap/iD/pull/11861
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
-
 [#11870]: https://github.com/openstreetmap/iD/issues/11870
 [#11904]: https://github.com/openstreetmap/iD/issues/11904
-
 [#11834]: https://github.com/openstreetmap/iD/pull/11834
-
 [@ilias52730]: https://github.com/ilias52730
 [@Razen04]: https://github.com/Razen04
 [@homersimpsons]: https://github.com/homersimpsons
 [@omsaraykar]: https://github.com/omsaraykar
 [@Kaushik4141]: https://github.com/Kaushik4141
-
 [@Kayd-06]: https://github.com/Kayd-06
 [@JaiswalShivang]: https://github.com/JaiswalShivang
-
-
 [@1-navneet]: https://github.com/1-navneet 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,6 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#11784]: https://github.com/openstreetmap/iD/pull/11784
 [#11794]: https://github.com/openstreetmap/iD/pull/11794
 [#11799]: https://github.com/openstreetmap/iD/issues/11799
-
 [#11783]: https://github.com/openstreetmap/iD/pull/11783
 [#11861]: https://github.com/openstreetmap/iD/pull/11861
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,11 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * When viewing the help information for tags or presets, use locale-specific properties if available ([#11760], thanks [@k-yle])
 * Show country names in your preferred language and country flag emoji in the Country field dropdown ([#11783], thanks [@Razen04])
 #### :hourglass: Performance
+
 * iD is now twice as fast during long editing sessions ([#11861], thanks [@k-yle])
+
+* Improve zoom/pan performance by avoiding redundant tag class recomputation ([#11834], thanks [@1-navneet])
+
 #### :mortar_board: Walkthrough / Help
 #### :rocket: Presets
 * Add dedicated styling for `highway=ladder` to make it distinguishable from `highway=steps` ([#11799], thanks [@bhavyaKhatri2703])
@@ -82,18 +86,27 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#11784]: https://github.com/openstreetmap/iD/pull/11784
 [#11794]: https://github.com/openstreetmap/iD/pull/11794
 [#11799]: https://github.com/openstreetmap/iD/issues/11799
+
 [#11783]: https://github.com/openstreetmap/iD/pull/11783
 [#11861]: https://github.com/openstreetmap/iD/pull/11861
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
+
 [#11870]: https://github.com/openstreetmap/iD/issues/11870
 [#11904]: https://github.com/openstreetmap/iD/issues/11904
+
+[#11834]: https://github.com/openstreetmap/iD/pull/11834
+
 [@ilias52730]: https://github.com/ilias52730
 [@Razen04]: https://github.com/Razen04
 [@homersimpsons]: https://github.com/homersimpsons
 [@omsaraykar]: https://github.com/omsaraykar
 [@Kaushik4141]: https://github.com/Kaushik4141
+
 [@Kayd-06]: https://github.com/Kayd-06
 [@JaiswalShivang]: https://github.com/JaiswalShivang
+
+
+[@1-navneet]: https://github.com/1-navneet 
 
 
 # v2.37.3
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -789,19 +789,31 @@ export function svgLabels(projection, context) {
 const _textWidthCache = {};
 export function textWidth(text, size, container) {
     let c = _textWidthCache[size];
-    if (!c) c = _textWidthCache[size] = {};
-
-    if (c[text]) {
-        return c[text];
+    if (!c) {
+        c = _textWidthCache[size] = { widths: {}, measurer: null };
     }
-    const elem = document.createElementNS('http://www.w3.org/2000/svg', 'text');
-    elem.style.fontSize = `${size}px`;
-    elem.style.fontWeight = 'bold';
+
+    if (c.widths[text]) {
+        return c.widths[text];
+    }
+
+    // Avoid repeated create/remove of SVG text nodes on every zoom/pan.
+    // The measured width depends only on text + font size/style, which
+    // are stable during zoom/pan, so caching is safe.
+    let elem = c.measurer;
+    if (!elem) {
+        elem = document.createElementNS('http://www.w3.org/2000/svg', 'text');
+        elem.style.fontSize = `${size}px`;
+        elem.style.fontWeight = 'bold';
+        elem.style.visibility = 'hidden';
+        elem.style.pointerEvents = 'none';
+        container.appendChild(elem);
+        c.measurer = elem;
+    }
+
     elem.textContent = text;
-    container.appendChild(elem);
-    c[text] = elem.getComputedTextLength();
-    elem.remove();
-    return c[text];
+    c.widths[text] = elem.getComputedTextLength();
+    return c.widths[text];
 }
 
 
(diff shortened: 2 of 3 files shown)


gralp-1:
This is certainly better but I'm still getting some 400ms+ zoom calls. The culprit seems to be indexing an entity's tags through `v = t[k]` on lines [81](https://github.com/1-navneet/iD/blob/406568bb6f0f2718e45d7cc4e182554598c8381c/modules/svg/tag_classes.js#L81), [106](https://github.com/1-navneet/iD/blob/406568bb6f0f2718e45d7cc4e182554598c8381c/modules/svg/tag_classes.js#L106) and [119](https://github.com/1-navneet/iD/blob/406568bb6f0f2718e45d7cc4e182554598c8381c/modules/svg/tag_classes.js#L119). It maybe seems like the cache is mostly missing? 
image

1-navneet:
@gralp-1,

Thanks for taking a closer look — glad to hear this is an improvement.
That’s a good catch. The cache key is currently based on the computed tag value, so if many entities have distinct tag objects or values, the hit rate may indeed be low.

I’ll take a closer look at whether we can avoid repeatedly indexing into the entity tags during redraws, or restructure the cache to operate at the entity or tag-set level instead. I’ll report back once I have a clearer picture.

1-navneet:
Thanks for the detailed profiling — that was very helpful.

I’ve pushed a follow-up commit that adds a WeakMap cache keyed by the entity’s tag object, so getClassesString() can return early without repeatedly indexing t[k] during redraws.
This reduces cache misses and further smooths zoom interactions, while keeping behavior unchanged and the changes confined to tag_classes.js.

Let me know if you’d like me to try a different cache key or gather additional traces.

tordans:
Could you add some guidance on how to test this properly? And also add some test results here. Is this a case that will be visible in [flame charts](https://developer.chrome.com/docs/devtools/performance/reference?hl=en#hide-func)?

1-navneet:
@tordans,
Thanks for the questions — happy to clarify.

How to test:
•Run npm start
•Open http://localhost:8080/#map=18/53.382847/-1.470167
•Open Chrome DevTools → Performance (Screenshots enabled)
•Record ~6–8 seconds while repeatedly zooming, then panning
•Stop the recording and inspect the Bottom-Up / Call Tree views

What to look for:
In the flame chart / bottom-up view, tagClasses.getClassesString previously appeared as a dominant hotspot during zoom redraws. With this change, its self/total time is significantly reduced and no longer dominates redraw time.

Observed results:
Zoom interactions are smoother and cache misses are reduced. getClassesString still appears in flame charts, but with much lower cost and fewer expensive calls. No visual or behavioral changes were observed.

Let me know if you’d like screenshots attached or traces collected for a specific scenario.

1-navneet:
@k-yle

modules/svg/tag_classes.js
```diff
@@ -39,6 +44,16 @@ export function svgTagClasses() {
 
 
     tagClasses.getClassesString = function(t, value) {
+        // Fast path: return cached class string for identical tags + base class.
+        var byBase = _classCache.get(t);
+        if (!byBase) {
+            byBase = new Map();
+            _classCache.set(t, byBase);
```
k-yle: currently your cache (`WeakMap<Tags, Map<string, string>>`) represents this hierarchy: `tags → oldClassName → newClassName`. 

This leads to a few issues:
- the cache hit rate will be very low, because whenever the user hovers or selects a feature, the class name is updated (`.selected` or `.hover` is added/removed)
- your cache will grow in size quickly, and store many no-op transformations: 

Image

---

To solve these issues, I think you could only cache a [weak-](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)mapping of `tags → className` (i.e. `WeakMap<Tags, string>`)

This will require some refactoring, so that we have a pure function which is simply `f(tags) → className`.

for example, something like this: 

```js
let cache = new WeakMap();

tagClasses.getClassesString = function(t, oldClassName) {
	const oldClassNamestoKeep = oldClassName.split(' ').filter(…).join(' ');
   
    let newClassNames = cache.get(t);
    if (newClassNames === undefined) {
    	newClassNames = getClassesStringPure(t); // refactor everything else into this new function
        cache.set(t, newClassNames);
    }

    return oldClassNamestoKeep + ' ' + newClassNames;
}
```

modules/svg/tag_classes.js
```diff
@@ -17,6 +17,10 @@ export function svgTagClasses() {
         'man_made', 'indoor', 'construction', 'proposed'
     ];
     var _tags = function(entity) { return entity.tags; };
+    // Cache computed class strings by tag object identity + base class.
+    // Safe for zoom/pan because tags do not change, and if tags change
+    // the graph creates new tag objects (cache miss => recompute).
+    var _classCache = new WeakMap();
```
k-yle: this needs to be in the global scope, not inside `svgTagClasses()`. otherwise the cache will get wiped on every render, since `svgTagClasses()` is frequently recreated

modules/svg/labels.js
```diff
@@ -789,19 +789,31 @@ export function svgLabels(projection, context) {
 const _textWidthCache = {};
 export function textWidth(text, size, container) {
```
k-yle, line 790: for the `textWidth` changes (which are unrelated to the other optimisation): 👍 from me, re-using a single DOM element is definitely more efficient than creating hudnreds of temporary elements whenever the map is panned/zoomed

modules/svg/tag_classes.js
```diff
@@ -39,6 +44,16 @@ export function svgTagClasses() {
 
 
     tagClasses.getClassesString = function(t, value) {
+        // Fast path: return cached class string for identical tags + base class.
+        var byBase = _classCache.get(t);
+        if (!byBase) {
+            byBase = new Map();
+            _classCache.set(t, byBase);
+        }
```
k-yle: Another thought, feel free to ignore this:

If we use a weakmap, that means two OSM entities with the same tags won't share the same cached classname.

Using the devtools's [Live Expression](https://developer.chrome.com/docs/devtools/console/live-expressions) feature[^1], and panning/zooming for a while, I found that only 5-20% is actually cached:

Image

So, this might be a rare case where `Map<JSON.stringify(tags), …>` is actually more efficient than `WeakMap<tags, …>`. The cache-hit rate seemed to increase to 70-90%, but this depends heavily on the local area


[^1]: …and some crude inline code like: `window.hit ||= 0; window.hit++;`

k-yle (commented):
this is a great start, nice work. some thoughts below:

1-navneet:
@k-yle

Thanks for the detailed feedback — that makes sense.
You’re right about the cache hierarchy and scope concerns. I’ll refactor this to:
- move the cache to module scope
- reduce it to a pure tags → className mapping
- separate base-class handling from tag-derived classes
I’ll push an updated revision shortly.

modules/svg/tag_classes.js
k-yle: actually, this isn't really worth it, if we apply https://github.com/openstreetmap/iD/pull/11834#discussion_r2793389953 then the benefit of this is too small to reliably benchmark

k-yle:
if it helps, 556fbb2e3a471008c6c1e9fb93d3d1c160a6abc4 is one way to implement the suggestions above

1-navneet:
Thanks for the pointer to 556fbb2 — that helped.

I’ve updated the implementation to:

• move the cache to module scope
• cache only tag-derived classes (tags → class list)
• keep base class handling separate from the cached values

Let me know if this matches what you were suggesting, or if you'd prefer it structured differently.

k-yle (approved):
thanks, this works really well now ! 

@gralp-1: I see you did your own profiling, do you want to take a look at the latest version of this PR?

@1-navneet: last suggestion from me, can you please update the CHANGELOG.md file?

1-navneet:
Updated the CHANGELOG as suggested and rebased onto latest develop.
Thanks again for the review!

gralp-1:
I still seem to be getting some redraw calls up to 2.5s 
[openstreetmap iD trace.gz](https://github.com/user-attachments/files/25327454/openstreetmap.iD.trace.gz)
(to see the traces in chrome, press f12 on any website, go to the performance tab and press load trace on the top bar)

gralp-1:
although these redraws seem to be spending ~50% of their time doing layout and style calculations

k-yle:
I've been testing this with [CPU Throttling set to](https://github.com/user-attachments/assets/875ffd53-5ff6-4ef9-a01c-e46c22adeca2) 'Mid Tier Mobile', and there's noticably less[^1] time spent in iD's own code (`getClassesString`), with the bottleneck now moving to reflows, as you noticed, which is tracked by #7710

for future reference, this is [the trace from my laptop](https://github.com/user-attachments/files/25333860/iD-pr11834-20260216.json.gz)


[^1]: it's hard to put an exact number on it, but it seems to be at least 30% faster (based on my very unscientific testing)

1-navneet:
Thanks for testing with CPU throttling — that’s helpful.
I’m also seeing the bottleneck shift toward layout/reflow costs now.
Since that aligns with #7710, I’m considering this PR an incremental improvement focused on reducing getClassesString overhead.
Happy to follow up separately on the reflow work if that makes sense.

1-navneet:
Hi @k-yle — just a quick follow-up on this when you have time.
The branch is rebased onto latest develop, conflicts resolved, and checks are passing.
Happy to make further adjustments if needed.

modules/svg/tag_classes.js
```diff
@@ -17,6 +17,7 @@ export function svgTagClasses() {
         'man_made', 'indoor', 'construction', 'proposed'
     ];
     var _tags = function(entity) { return entity.tags; };
+    // Cache tag-derived classes by tag object identity.
```
matkoniecz, line 19: to what this comment is referring to? line above?

I *think* (I may be wrong!) that in such case comment should be above, not below

CHANGELOG.md
```diff
@@ -82,20 +86,29 @@ _Breaking developer changes, which may affect downstream projects or sites that
 [#11784]: https://github.com/openstreetmap/iD/pull/11784
 [#11794]: https://github.com/openstreetmap/iD/pull/11794
 [#11799]: https://github.com/openstreetmap/iD/issues/11799
+
```
matkoniecz: is this empty line needed?

modules/svg/tag_classes.js
1-navneet: Moved the comment above the relevant line.

CHANGELOG.md
1-navneet: Removed the extra blank line.

CHANGELOG.md
```diff
@@ -85,17 +89,25 @@ _Breaking developer changes, which may affect downstream projects or sites that
 [#11783]: https://github.com/openstreetmap/iD/pull/11783
 [#11861]: https://github.com/openstreetmap/iD/pull/11861
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
+
```
matkoniecz: also here seems not needed as far as I see

CHANGELOG.md
```diff
@@ -85,17 +89,25 @@ _Breaking developer changes, which may affect downstream projects or sites that
...
 [#11904]: https://github.com/openstreetmap/iD/issues/11904
+
+[#11834]: https://github.com/openstreetmap/iD/pull/11834
+
 [@ilias52730]: https://github.com/ilias52730
 [@Razen04]: https://github.com/Razen04
 [@homersimpsons]: https://github.com/homersimpsons
 [@omsaraykar]: https://github.com/omsaraykar
 [@Kaushik4141]: https://github.com/Kaushik4141
+
```
matkoniecz: also here seems not needed as far as I see

CHANGELOG.md
```diff
@@ -85,17 +89,25 @@ _Breaking developer changes, which may affect downstream projects or sites that
 [#11783]: https://github.com/openstreetmap/iD/pull/11783
 [#11861]: https://github.com/openstreetmap/iD/pull/11861
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
+
 [#11870]: https://github.com/openstreetmap/iD/issues/11870
 [#11904]: https://github.com/openstreetmap/iD/issues/11904
+
```
matkoniecz: also here seems not needed as far as I see

CHANGELOG.md
```diff
@@ -85,17 +89,25 @@ _Breaking developer changes, which may affect downstream projects or sites that
...
+[#11834]: https://github.com/openstreetmap/iD/pull/11834
+
```
matkoniecz: also here seems to be new one, is it needed?

1-navneet: Cleaned up extra blank lines.
Adjusted formatting to match surrounding entries.

PR #11836, opened 2026-01-31
feat: Allow multiple custom background layers with names , added crud…

### LINKED ISSUE #8874: Multiple custom background layers
### Description

Here is an idea that only takes one minute of coding:
Instead of just:
- Custom

have:
- Custom 1
- Custom 2
- Custom 3

### Screenshots

20220104T005305

JaiswalShivang:
fixes #8874 

# Description
This PR addresses issue #8874 by allowing users to add, edit, and persist multiple custom background layers. Previously, iD only supported a single custom background layer.

# Features
- **Multiple Custom Backgrounds**: Users can now add multiple custom tile layers.
- **Custom Naming**: Added a name input field to the custom background modal. Users can name their layers (e.g., "My Drone Imagery", "Local Gov Map").
- **Persistence**: Custom layers are saved in `localStorage` as an array and persist across page reloads.
- **Management UI**: 
  - **Add**: Button to add a new custom layer.
  - **Edit**: Edit existing custom layers (name and URL).
  - **Delete**: Remove custom layers using the 'X' button.
- **Migration**: Automatically migrates existing single custom background settings to the new format.

# Video

https://github.com/user-attachments/assets/d00687ad-c59a-43c5-9c38-d2d7842a76b7

### DIFF
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -77,10 +77,34 @@ export function rendererBackground(context) {
         // Add 'None'
         _imageryIndex.backgrounds.unshift(rendererBackgroundSource.None());
 
-        // Add 'Custom'
-        let template = prefs('background-custom-template') || '';
-        const custom = rendererBackgroundSource.Custom(template);
-        _imageryIndex.backgrounds.unshift(custom);
+        // Add Custom backgrounds (support multiple)
+        const oldTemplate = prefs('background-custom-template');
+        let customTemplates = [];
+        try {
+          customTemplates = JSON.parse(prefs('background-custom-templates') || '[]');
+        } catch {
+          customTemplates = [];
+        }
+
+        if (oldTemplate && !customTemplates.some(c => c.template === oldTemplate)) {
+          customTemplates.unshift({ id: 'custom-0', template: oldTemplate, name: 'Custom (Migrated)' });
+          prefs('background-custom-templates', JSON.stringify(customTemplates));
+          // Clear old single-template pref after migration to prevent re-migration
+          prefs('background-custom-template', null);
+        }
+
+        customTemplates.forEach((customData, index) => {
+          const customId = customData.id || `custom-${index}`;
+          const customName = customData.name || `Custom ${index + 1}`;
+          const custom = rendererBackgroundSource.Custom(customData.template, customName);
+          custom.id = customId;
+          _imageryIndex.backgrounds.unshift(custom);
+        });
+
+        const emptyCustom = rendererBackgroundSource.Custom('');
+        emptyCustom.id = 'custom';
+        emptyCustom.isEmptyCustomSlot = true;
+        _imageryIndex.backgrounds.unshift(emptyCustom);
 
         return _imageryIndex;
       });
@@ -187,7 +211,7 @@ export function rendererBackground(context) {
   }
 
 
-  background.updateImagery = function() {
+  background.updateImagery = function () {
     let currSource = baseLayer.source();
     if (context.inIntro() || !currSource) return;
 
@@ -307,7 +331,7 @@ export function rendererBackground(context) {
   };
 
 
-  background.baseLayerSource = function(d) {
+  background.baseLayerSource = function (d) {
     if (!arguments.length) return baseLayer.source();
 
     // test source against OSM imagery blocklists..
@@ -379,7 +403,7 @@ export function rendererBackground(context) {
       .source(d)
       .projection(context.projection)
       .dimensions(baseLayer.dimensions()
-    );
+      );
 
     _overlayLayers.push(layer);
     dispatch.call('change');
@@ -398,7 +422,7 @@ export function rendererBackground(context) {
   };
 
 
-  background.offset = function(d) {
+  background.offset = function (d) {
     const currSource = baseLayer.source();
     if (!arguments.length) {
       return (currSource && currSource.offset()) || [0, 0];
@@ -412,7 +436,7 @@ export function rendererBackground(context) {
   };
 
 
-  background.brightness = function(d) {
+  background.brightness = function (d) {
     if (!arguments.length) return _brightness;
     _brightness = d;
     if (context.mode()) dispatch.call('change');
@@ -420,7 +444,7 @@ export function rendererBackground(context) {
   };
 
 
-  background.contrast = function(d) {
+  background.contrast = function (d) {
     if (!arguments.length) return _contrast;
     _contrast = d;
     if (context.mode()) dispatch.call('change');
@@ -428,7 +452,7 @@ export function rendererBackground(context) {
   };
 
 
-  background.saturation = function(d) {
+  background.saturation = function (d) {
     if (!arguments.length) return _saturation;
     _saturation = d;
     if (context.mode()) dispatch.call('change');
@@ -436,7 +460,7 @@ export function rendererBackground(context) {
   };
 
 
-  background.sharpness = function(d) {
+  background.sharpness = function (d) {
     if (!arguments.length) return _sharpness;
     _sharpness = d;
     if (context.mode()) dispatch.call('change');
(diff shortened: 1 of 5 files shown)


JaiswalShivang:
@tordans Sir i have created the pr for issue no 8874 , Any suggestions?

modules/renderer/background.js
```diff
@@ -187,7 +211,7 @@ export function rendererBackground(context) {
   }
 
 
-  background.updateImagery = function() {
+  background.updateImagery = function () {
```
tordans, line 214: Remove all code changes that are unrelated to your PR. All those whitespace changes should not be part of the PR unless you actually touched the code.

modules/renderer/background_source.js
```diff
@@ -52,72 +52,72 @@ export function rendererBackgroundSource(data) {
...
+    source.imageryUsed = function () {
         return _name || source.id;
     };
 
 
-    source.template = function(val) {
+    source.template = function (val) {
         if (!arguments.length) return _template;
-        if (source.id === 'custom' || source.id === 'Bing') {
+        if (source.id === 'custom' || source.id?.startsWith('custom-') || source.id === 'Bing') {
```
tordans, line 113:
```suggestion
        if (source.id?.startsWith('custom-') || source.id === 'Bing') {
```

You migrated before so this should only be the one case, right? 
Personally I would try to make the string "custom-" a CONST in one of the file that handles the code and consts for the custom feature and reference this here. It would be a bit less magical to have some string show up here.

modules/renderer/background_source.js
```diff
@@ -560,20 +560,22 @@ rendererBackgroundSource.None = function() {
...
 
-rendererBackgroundSource.Custom = function(template) {
+rendererBackgroundSource.Custom = function (template, customName) {
     var source = rendererBackgroundSource({ id: 'custom', template: template });
 
-
-    source.name = function() {
-        return t('background.custom');
+    source.name = function () {
+        return customName || t('background.custom');
```
tordans, line 567: See notes on the fallback being part of the URL.
IMO all layer need to be prefixed.
* > Custom: My Name
* > Custom: tiles.codefor.de/berlin/geoportal/luftbilder/2025…

(Cut of with `…` after some chars and with CSS ellipse cutoff (see "[trancate](https://tailwindcss.com/docs/text-overflow)")

modules/ui/sections/background_list.js
```diff
@@ -193,29 +188,47 @@ export function uiSectionBackgroundList(context) {
...
             })
             .on('change', change);
 
         label
             .append('span')
-            .each(function(d) { d.label()(d3_select(this)); });
+            .each(function (d) { d.label()(d3_select(this)); });
 
-        enter.filter(function(d) { return d.id === 'custom'; })
+        enter.filter(function (d) { return d.id === 'custom' || d.id?.startsWith('custom-'); })
```
tordans, line 200:
```suggestion
        enter.filter(function (d) { return d.id?.startsWith('custom-'); })
```

See above

modules/ui/sections/background_list.js
```diff
@@ -193,29 +188,47 @@ export function uiSectionBackgroundList(context) {
...
             )
-            .on('click', function(d3_event) {
+            .on('click', function (d3_event, d) {
                 d3_event.preventDefault();
-                editCustom();
+                d3_event.stopPropagation();
+                deleteCustom(d.id);
             })
-            .call(svgIcon('#iD-icon-more'));
+            .call(svgIcon('#iD-icon-close'));
```
tordans, line 229: Is this the delete button? We use the trash icon in other places. Is there a reason not to use it?

modules/ui/sections/background_list.js
```diff
@@ -255,31 +268,97 @@ export function uiSectionBackgroundList(context) {
...
-            chooseBackground(_customSource);
-        } else {
-            _customSource.template('');
-            chooseBackground(context.background().findSource('none'));
+            let customTemplates = [];
+            try {
+                customTemplates = JSON.parse(prefs('background-custom-templates') || '[]');
+            } catch {
+                customTemplates = [];
+            }
```
tordans, line 278: Please explain why we have to change the way we get the templates here. How did we do it before https://github.com/openstreetmap/iD/pull/11836/changes#diff-e37d93c86d6f7ed0f922b8bfad4c4ea1d9ef6b5c2643a06382adb535c0bf5520L18 and why do we do it differently now. I find try-catch blocks like this deep in the code smelly. Is this to ensure the structure of the data is fine? If so, maybe this check should be somewhere near the migration code?

modules/ui/sections/background_list.js
```diff
@@ -255,31 +268,97 @@ export function uiSectionBackgroundList(context) {
...
         }
     }
 
 
-    function editCustom() {
+    function editCustom(customId) {
+        const settingsCustom = uiSettingsCustomBackground(context)
+            .on('change', (d) => customChanged(d, customId));
+
+        if (customId && customId !== 'custom') {
```
tordans, line 311: What is this about "!== custom"? Is this about the already migrated prefs?

modules/ui/sections/background_list.js
```diff
@@ -255,31 +268,97 @@ export function uiSectionBackgroundList(context) {
...
+                } catch {
+                    customTemplates = [];
+                }
+                const existing = customTemplates.find(c => c.id === customId);
+                prefs('background-custom-name', existing?.name || '');
+            }
+        } else {
+            prefs('background-custom-template', '');
+            prefs('background-custom-name', '');
+        }
```
tordans, line 328: Now I am very confused. I thought we had all in `background-custom-templates`.  But now there is background-custom-name and background-custom-template being used and I see its used in multiple places. What is this about?

modules/ui/sections/background_list.js
```diff
@@ -255,31 +268,97 @@ export function uiSectionBackgroundList(context) {
...
-            _customSource.template('');
-            chooseBackground(context.background().findSource('none'));
+            let customTemplates = [];
+            try {
+                customTemplates = JSON.parse(prefs('background-custom-templates') || '[]');
+            } catch {
+                customTemplates = [];
+            }
+
+            if (customId === 'custom') {
```
tordans, line 280: I thought our IDs are `custom-123`?

modules/ui/sections/background_list.js
```diff
@@ -255,31 +268,97 @@ export function uiSectionBackgroundList(context) {
...
+                const newId = `custom-${Date.now()}`;
+                const customName = d.name || `Custom ${customTemplates.length + 1}`;
+                customTemplates.push({ id: newId, template: d.template, name: customName });
+                prefs('background-custom-templates', JSON.stringify(customTemplates));
+                window.location.reload();
```
tordans, line 285: Cannot image this being a good thing.

tordans (changes requested):
Hey @JaiswalShivang here is some input. The more I read the code I got the feeling that I did not understand the general idea of some code blocks or how data is supposed to flow in the app. Maybe you did not clean up some parts of the code (see inline comments on prefs and id="custom") or maybe I did not understand. If the later, please use the PR comments to explain some of the things.

## Thoughts on testing:

- I took the time to write down all the cases that we have to consider. Please take the time to test them and document your findings in text and screenshots (where needed).
- Your video shows too little.
- I want to mention again how easy it is to provide a Netlify test version…

## Thoughts on the UX:

- Seeing your video I think we should change the UX. The fact that there is this empty "Custom" line makes it more complex that it needs to be.
- Instead we can put a button below the list "Add custom background" which opens the modal and adds the line. (Remeber i18n)
- This way we only have lines that are actual layer and dont have to handle the different "add", "edit", "remove" actions in the options lines.
- Modal: Add spacings and labels to fields; see below on custom "name", make this clear in the label and rember i18n.
- Maybe in the modal we should make the two lists "Supported … tokens" HTML details-Summary UI so the headlines are there but the details are optional. The "summary" part needs some styling to make it look like a regular iD Link with hover and all, to make this work from a UX point of view.
- The delete should be the delete icon, the Edit should be a pencil icon IMO (or some other edit icon we use in the codebar).

## Thoughts on the code:

- I am always looking for ways to extract code that belongs together into a separate function and maybe file. The lines 80-100 in backgrounds.js look like enough code that could be separte.
- One sub function in there could be just handling the migration. 
- I think it could be easier to read to have the names of the pre as const (capital letters) at the start of this file. One for the old prev id, one for the new.

### About the prefs:

- About the format that we store: I think we should align this with [https://github.com/openstreetmap/iD/pull/11269/changes#diff-aa20c60efb510a16d5d0dd76ca90c3bfaa2af62f25400793807d838bf9c96182R36-R58](https://github.com/openstreetmap/iD/pull/11269/changes#diff-aa20c60efb510a16d5d0dd76ca90c3bfaa2af62f25400793807d838bf9c96182R36-R58) This PR is still WIP as well, but both are storing an array of things, so they should store it in a similar way. My first take is to store it as an (ordered) array that  has an object of url+name props. What do you need the "id: custom", "id:custom-1" for? The URL should always be unique and could be used as a key. Does length matter here? Or, can you create the IDs on the fly? We dont have to store them in the prefs in order to use them…

### About the "name":

- The "name" should IMO be optional. The UI should show a part of the URL as fallback. This will mean users dont have to spend time adding it if they dont need it. And you dont have to handle the during migration. Which is avoids handling i18n at this point (which is missing ATM). It also solves that URLs from URL dont have a name and should not have one.
-

PR #11843, opened 2026-02-03
Ensure XML elements are created without namespace in JXON

### LINKED ISSUE #11842: Unable to save edits without receiving an error message in the ID Editor
I use Safari on a MacBook Pro from 2023.  Before today, I could edit and save without trouble.  When I tried to save the first and last times today, I received the error message below.
Errors occurred while trying to save
[Cannot parse valid changeset from xml string <osm xmlns="http://www.w3.org/1999/xhtml"><changeset version="0.6" generator="iD"><tag k="created_by" v="iD 2.37.3"></tag><tag k="host" v="https://www.openstreetmap.org/edit"></tag><tag k="locale" v="en-US"></tag><tag k="imagery_used" v="Monongalia County 2023 Aerial Imagery;Esri World Imagery (Clarity) Beta;Bing Maps Aerial"></tag><tag k="source" v="Issue Messages;OSM data;aerial imagery"></tag><tag k="comment" v="I fixed issues OSM found including Fix Me messages. I left a Fix Me message when unsure and an issue was unfixable. I added a switched a fire station point to an area also."></tag><tag k="changesets_count" v="2271"></tag><tag k="warnings:almost_junction:highway-highway" v="1"></tag><tag k="warnings:impossible_oneway:highway" v="2"></tag><tag k="resolved:almost_junction:highway-highway" v="2"></tag><tag k="resolved:crossing_ways:building-highway" v="1"></tag><tag k="resolved:crossing_ways:highway-highway" v="3"></tag><tag k="resolved:close_nodes:vertices" v="1"></tag><tag k="resolved:disconnected_way:highway" v="1"></tag><tag k="resolved:outdated_tags:incomplete_tags" v="8"></tag><tag k="resolved:outdated_tags:noncanonical_brand" v="1"></tag><tag k="resolved:suspicious_name:generic_name" v="1"></tag><tag k="resolved:help_request:fixme_tag" v="4"></tag></changeset></osm>. XML doesn't contain an osm/changeset element.](https://www.openstreetmap.org/id#)
OK
I guess attaching a screenshot is the best that I can do since the URL did not reproduce the error.  I have not closed Safari since being unable to save.  A solution that allows me to save my 220 changes is desired but not required.

Sincerely,

Andrew
Dark_Valor

Image

Kayd-06:
### Description

This PR fixes a bug where Safari users were unable to save edits in the iD editor (#11842).

When saving in Safari, the browser was injecting the XHTML namespace (`xmlns="http://www.w3.org/1999/xhtml"`) into the root `<osm>` element. The OpenStreetMap API server fails to parse XML with this namespace, as it expects the elements to be in the null (default) namespace.

The error message reported by users was:
`Cannot parse valid changeset from xml string . XML doesn't contain an osm/changeset element.`

### Changes

- Modified [modules/util/jxon.js] to use `createElementNS(null, sName)` instead of `createElement(sName)`. This explicitly forces the elements into the null namespace, preventing browsers from adding their own defaults.

### Verification

- Verified that the generated XML for changesets does not contain the `xmlns` attribute.
- Ran all existing tests (`npm run test:spec`) to ensure no regressions in XML generation for other browsers.

Fixes #11842

### DIFF
--- a/modules/util/jxon.js
+++ b/modules/util/jxon.js
@@ -110,12 +110,12 @@ export var JXON = new (function () {
         oParentEl.setAttribute(sName.slice(1), vValue);
       } else if (vValue.constructor === Array) {
         for (var nItem = 0; nItem < vValue.length; nItem++) {
-          oChild = oXMLDoc.createElement(sName);
+          oChild = oXMLDoc.createElementNS(null, sName);
           loadObjTree(oXMLDoc, oChild, vValue[nItem]);
           oParentEl.appendChild(oChild);
         }
       } else {
-        oChild = oXMLDoc.createElement(sName);
+        oChild = oXMLDoc.createElementNS(null, sName);
         if (vValue instanceof Object) {
           loadObjTree(oXMLDoc, oChild, vValue);
         } else if (vValue !== null && vValue !== true) {


tordans:
Please remove all unrelated white space changes and force push.

Kayd-06:
I have removed all unrelated whitespace changes and force pushed a cleaned commit containing only the namespace fix.

k-yle (approved):
@Kayd-06 and/or @DocAndrew666 – which version of Safari and MacOS did you use to trigger this issue? 

I tested the following snippet in Safari 26.2 on MacOS 26.2, and I get the same result in all browsers (i.e. no `xmlns` tag):
```js
new XMLSerializer().serializeToString(document.implementation.createDocument('', '', null).createElement('root'))
// should return '<root />'
```

So I'm interested to know which Safari versions have this issue. I could only find [a webkit bug from 2007](https://bugs.webkit.org/show_bug.cgi?id=14835) complaining about this.

---

Anyway, LGTM because the API indeed breaks if provided an `xmlns` attribute.

But would be good to know exactly how to reproduce this...

DocAndrew666:
I have macOS Tahoe 26.2 and Safari 26.2 on my MacBook Pro.  My computer is updated and not too old from 2 or 3 years ago.

DocAndrew666:
I figured the reason why I was unable to upload each changeset.  I added
Norton AntiTrack to my computer.  While working with Norton support, it was
verified that was preventing me from uploading each changeset.  I removed
the app and am able to save edits again (without issue messages).

PR #11844, opened 2026-02-03
feat: Add real-time character limit warning tooltip for tag values

### LINKED ISSUE #10241: Tag values over 255 characters long are silently truncated in raw tag editor

JaiswalShivang:
Fixes #10241 

## Summary

Adds a visual warning tooltip that appears in real-time when users exceed the 255 character limit for tag values in the raw tag editor.

## Problem

Previously, users had no immediate feedback when entering tag values that exceeded OpenStreetMap's 255 character limit. They would only discover the issue later, leading to confusion and a poor editing experience.

## Solution

- Added a warning tooltip that appears dynamically below the value input field when the character limit is exceeded
- Tooltip displays: "only 255 characters are allowed"
- Styled with a red background and upward-pointing arrow for clear visibility
- Smooth fade-in/fade-out animation for better UX

## Changes

- `modules/ui/sections/raw_tag_editor.js` - Added warning span element and real-time input validation
- `css/80_app.css` - Added tooltip styling with arrow indicator
- `data/core.yaml` - Added translation key for warning message

## Demo

https://github.com/user-attachments/assets/ed8e976b-4ac3-4266-a3dd-29bb4c3b10d6

## Testing

1. Open iD editor and select any feature
2. In the raw tag editor, add or edit a tag value
3. Type more than 255 characters
4. Observe the warning tooltip appears below the input field
5. Delete characters to go below 255 - tooltip disappears

## Checklist

- [x] Code follows the project's coding standards
- [x] No breaking changes to existing functionality


JaiswalShivang:
Hi @tordans,
I previously uploaded the wrong code /demo for this issue. This new PR contains the correct implementation and demo.
I have also closed the previous PR to avoid confusion.
Please review when you have time. Thank you!

data/core.yaml
```diff
@@ -734,6 +734,7 @@ en:
     wiki_reference: View documentation
     wiki_en_reference: View documentation in English
     key_value: "key=value"
+    max_char_warning: "only {count} characters are allowed"
```
tyrasd: was there a reason why the existing string `inspector.max_length_reached` was not reused for this?

tyrasd (changes requested):
Did you try to use the existing module `uiLengthIndicator`? I do actually like the styling of the error case with the redish background color and underline. But the styling should be consistent between the different places where this error can occur (i.e. regular fields vs. in the raw tag editor), so if possible this should be incorporated into the existing length indicator ui widget.

JaiswalShivang:
Thanks for the feedback @tyrasd! You're right - I should use the existing uiLengthIndicator module for consistency. I'll refactor this to integrate with the existing length indicator UI instead of creating a separate warning tooltip. This will ensure consistent styling across both regular fields and the raw tag editor.

data/core.yaml
JaiswalShivang: Good catch! I can reuse inspector.max_length_reached instead of creating a new string. I'll update the implementation to use the existing translation key.

JaiswalShivang:
@tyrasd I’ve implemented the suggested changes so that each tag row now uses its own instance of uiLengthIndicator, and the character limit warning appears correctly for each input. I’ve also removed obsolete code and translation keys as discussed.

I’m attaching a short video to demonstrate the updated behavior: you’ll see that the warning now displays as expected when a tag value exceeds the character limit, and each row handles its own indicator independently. Please let me know if you have any further feedback or requests!

## Video

https://github.com/user-attachments/assets/41763ec6-4644-41da-afca-98a27e780d5c

JaiswalShivang:
@matkoniecz Kindly see the pr and suggest some changes if needed

matkoniecz:
@JaiswalShivang please do not ping me with review requests - with 200+ pull requests waiting for review across multiple iD repos it will take me some time to go through them

JaiswalShivang:
Oops, accidentally hit the close button! Reopened it right away. The PR is updated and ready for review.

PR #11848, opened 2026-02-04
Fix: implement tag retention for Extract tool to prevent area validation errors

### LINKED ISSUE #11830: bad extract of a swimming pool

ak8abhinay:
I’ve been working on a fix for the Extract Tool to resolve the data integrity issues (Red Errors) that occur when extracting tags from independent objects. (After looking at all possible edge cases I have chosen this approach to resolve the issue)

The Problem :
In the current implementation, the Extract tool is "too aggressive" when handling standalone features like swimming pools. When a user extracts a label to a Point for better visibility, the tool often strips all descriptive tags from the original geometry, leaving behind only area=yes. This immediately triggers the "Area has no descriptive tags" validation error.

The Approach: "Tag" Retention
Instead of a total extraction, I’ve implemented a "Tag Retention" logic in modules/actions/extract.js. This approach ensures that the tool remains available everywhere while protecting the validity of the underlying geometry.

Key Logic Applied:

I introduced a regex for primaryCategoryKeys (e.g., amenity, shop, leisure, historic, etc.) to identify the core "Identity" of an object.

Context: The code now checks if the area has a structural fallback like building=yes.

Scenario A (Buildings): If it’s a building, the primary tag moves to the Point because the area remains valid as a "Building".

Scenario B (Standalone Objects): If there is no building tag, the logic refuses to delete the primary tag from the area. It copies the tag to the Point but leaves it on the Area as a "Shell," preventing the Red Error.

Handling Vertical Data (Level Tags)
I have also ensured that vertical data standards are maintained.

Level Retention: The extraction logic specifically identifies the level key and retains it on the area footprint. This ensures that even if a business name is moved to a Point (to resolve "Too Close" warnings), the physical structure doesn't lose its floor data.

Summary of Changes
extract.js (Action): Added conditional deletion logic based on building presence and primary category lists.

This fix allows for the high icon visibility mappers want without the data corruption that currently triggers the validator. I'd love to get some feedback on this approach!

#Before (When the "Extraction" tool used)
imgpr#2
imgpr#off

#After (When the "Extraction" tool used)
imgpr#2testing
imgpr#testing

Fixes #11830

### DIFF
--- a/modules/actions/extract.js
+++ b/modules/actions/extract.js
@@ -9,11 +9,9 @@ export function actionExtract(entityID, projection) {
     /** @param {boolean} shiftKeyPressed */
     var action = function(graph, shiftKeyPressed) {
         var entity = graph.entity(entityID);
-
         if (entity.type === 'node') {
             return extractFromNode(entity, graph, shiftKeyPressed);
         }
-
         return extractFromWayOrRelation(entity, graph);
     };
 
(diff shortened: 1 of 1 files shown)


tordans:
In the future, please update the existing PR by force pushing. https://github.com/openstreetmap/iD/pull/11833

ak8abhinay:
@tordans, 
Thank you for the guidance. I tried to update the original PR  #11833 as requested, but I accidentally deleted the branch on my fork right after PR #11833 was closed.
Since PR #11833 was closed and even I restore the branch, I can't force push the updates back to it anymore. I’ve made sure this new PR #11848 is perfectly clean with just the 1 commit and 1 file change. Could we please use this one as the main PR instead? 
I'm sorry for the trouble.

tyrasd:
See https://github.com/openstreetmap/iD/issues/11830#issuecomment-3876238660

PR #11861, opened 2026-02-10
make iD 2x faster during long edit sessions

k-yle:
For changesets that last several hours and affect thousands of elements ([example](https://osmcha.org/changesets/177297256)), `coreHistory`'s stack quickly reaches >1GB of JSON, which makes `history.toJSON()` the perf bottleneck. 

Over 99% of time is spent inside this function after ever operation, and we call the function twice in a row for no good reason. We can just reuse the value from a few lines above

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * When viewing the help information for tags or presets, use locale-specific properties if available ([#11760], thanks [@k-yle])
 * Show country names in your preferred language and country flag emoji in the Country field dropdown ([#11783], thanks [@Razen04])
 #### :hourglass: Performance
+* iD is now twice as fast during long editing sessions ([#11861], thanks [@k-yle])
 #### :mortar_board: Walkthrough / Help
 #### :rocket: Presets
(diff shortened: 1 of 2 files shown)


1ec5 (approved):
This calls for a blog post or diary entry or something. 📉

bryceco:
It's one of those things where you're happy for the improvement, but embarrassed about the reason.

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 * The Languages field shows language names in your preferred language. ([#11699], thanks [@Razen04])
 * When viewing the help information for tags or presets, use locale-specific properties if available ([#11760], thanks [@k-yle])
 #### :hourglass: Performance
+* iD is now double as fast during long editing session ([#11861], thanks [@k-yle])
```
Dimitar5555:
```suggestion
* iD is now twice as fast during long editing sessions ([#11861], thanks [@k-yle])
```

k-yle: thanks, done

tyrasd (approved):
Ouch, this was my fault in b10efc85f698b25621810473593f621a6ebcebbf. Thanks for spotting this, @k-yle!

PR #11862, opened 2026-02-10
Fix custom background crash due to early source capture

### LINKED ISSUE #11855: Error setting custom background layer directly from Custom Background Settings dialog
iD fails to switch to a custom layer if you customize the layer before switching to it:

1. Open iD in a private browsing window or after clearing cookies. (I’m using [a location in the United States](https://ideditor.netlify.app/#background=Bing&locale=en&map=20.43/39.28272/-84.28522), where my custom background layers have coverage.)
2. Open the Background panel. Without selecting the Custom option, click on the … button next to the Custom option.
3. In the Custom Background Settings dialog, enter a URL such as `https://historical1-stg.arcgis.com/arcgis/rest/services/USGS_Historical_Topographic_Maps/ImageServer/exportImage?bbox={bbox}&bboxSR={proj}&size={width},{height}&format=png&transparent=false&f=image&time=-467026560800,1136073600000` and click OK.

The background doesn’t change. Instead, the following error appears in the console:

```
Uncaught TypeError: can't access property "template", r is null
    f background_list.js:260
    call dispatch.js:57
    d custom_background.js:103
    XY on.js:3
background_list.js:260:13
    f background_list.js:260
    call dispatch.js:57
    d custom_background.js:103
    XY on.js:3
```

https://github.com/openstreetmap/iD/blob/fa2deff9b96a9bd189ebb594c3ba586333d8a90a/modules/ui/sections/background_list.js#L260

Reproduces in the latest development version of iD in ideditor.netlify.app. Also reproduces if you have already used a different custom background layer URL in the past but switch to a new URL via the … button. I and others have been noticing this issue for years, but I never quite put my finger on it.

Kayd-06:
## Problem
The iD editor crashes with a `TypeError` when a user attempts to set a custom background URL before the "Custom" imagery source is fully initialized. 

### Root Cause
In [background_list.js], the `_customSource` variable was initialized once when the UI section was first created. Because background sources load asynchronously, `context.background().findSource('custom')` could return `null` at that moment. When [customChanged()] was later called to update a URL, it attempted to access `.template()` on that null reference, causing the crash.

## Solution
- Removed the static `_customSource` initialization.
- Modified [customChanged(d)] to resolve both the [custom] and `none` imagery sources dynamically at the time they are needed.
- Added defensive null-guards to safely handle cases where the background sources might still be unavailable, preventing any potential crash paths.

## Testing
- Added a new specification file [test/spec/ui/sections/background_list.js] with 3 regression tests covering:
  - Safe initialization when sources are missing.
  - Safe handling of custom URL changes.
  - Safe fallback to the "none" source.
- Verified that all tests pass and that the code adheres to iD's formatting style.

Fixes #11855

### DIFF
--- a/modules/ui/sections/background_list.js
+++ b/modules/ui/sections/background_list.js
@@ -167,7 +167,7 @@ export function uiSectionBackgroundList(context) {
             .sort(function(a, b) {
                 return a.best() && !b.best() ? -1
                     : b.best() && !a.best() ? 1
-                        : d3_descending(a.area(), b.area()) || d3_ascending(a.name(), b.name()) || 0;
+                    : d3_descending(a.area(), b.area()) || d3_ascending(a.name(), b.name()) || 0;
             });
 
         var layerLinks = layerList.selectAll('li')
--- a/modules/ui/sections/background_list.js
+++ b/modules/ui/sections/background_list.js
@@ -15,8 +15,6 @@ export function uiSectionBackgroundList(context) {
 
     var _backgroundList = d3_select(null);
 
-    var _customSource = context.background().findSource('custom');
-
     var _settingsCustomBackground = uiSettingsCustomBackground(context)
         .on('change', customChanged);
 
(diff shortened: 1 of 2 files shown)


1ec5 (approved):
Thank you for this fix. It looks good at a code level and seems to work in cursory testing. The integration test is thorough and exercises the case well. It’s kind of unfortunate that it requires so much scaffolding, but that’s a more general issue with the codebase.

1ec5:
dc162070af2de2a672f4f21734ff4252ee583021 adds a changelog entry for this fix. Please feel free to give yourself credit in the changelog next time!

PR #11865, opened 2026-02-12
Skip disconnected_way validation for Golf Walking Paths

### LINKED ISSUE #11863: Validation - Golf Walking Path should not have to be connected to another path
Please consider skipping the validation of Golf Walking Paths (highway=path, golf=path, foot=designated) to be connected to other paths, as they are very often (always?) just isolated paths on a huge grassfield.

Much appreciated.

Image

Kayd-06:
Golf walking paths (highway=path + golf=path + foot=designated) are often intentionally isolated and should not trigger the `disconnected_way` validation warning.

I've updated the logic in [modules/validations/disconnected_way.js] and added a new test case in `test/spec/validations/disconnected_way.js` to verify the fix.

Fixes #11863


k-yle (commented):
did you test your PR?

Kayd-06:
Yes, I have verified the fix in two ways:

1. Automated Tests: I added a new test case to `test/spec/validations/disconnected_way.js`
 specifically for the isolated golf path tags. I ran the full validation suite via `vitest`, and all 5 tests passed.

2. Manual Verification: I ran the editor locally and confirmed that:
- A way with `highway=path` + `golf=path` + `foot=designated` no longer triggers the "disconnected way" warning.
- Regular paths still correctly trigger the warning when disconnected.

k-yle:
what made you choose that specific tag combination? it's different to the tags used by the preset, so it won't fix the example mentioned in #11863

Kayd-06:
I've updated the validation to match the actual iD preset tagging and added tests covering:

- Golf walking path preset
- Golf cartpath preset
- The original reporter tag combination

Please let me know if anything else should be adjusted.

matkoniecz:
can you answer

> what made you choose that specific tag combination?

? How you ended with those tags?

tyrasd (approved):
the current version of the PR seems to work as intended. `golf=path/cartpath` will skip the validation on any way they are on, so it will not matter if the cartpath was tagged as e.g. `highway=service` or `highway=path`, etc.

PR #11871, opened 2026-02-13
Replace deprecated document.createEvent/initEvent with modern Event constructor

### LINKED ISSUE #11870: Replace deprecated document.createEvent/initEvent with modern Event constructor
## Problem

The codebase uses the deprecated `document.createEvent()` and `initEvent()` methods in 3 files:

1. **`modules/util/trigger_event.js`** — Core utility function used by `address.js`, `combobox.js`, and `changeset_editor.js`
2. **`modules/ui/init.js`** (line 553-555) — With an outdated comment: *"Use outdated code so it works on Explorer"*
3. **`modules/renderer/map.js`** (line 309-310) — Unnecessary fallback for browsers without `CustomEvent`

These methods are [deprecated by MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/createEvent) and have been replaced by the modern [`Event()` constructor](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event) which is supported by all modern browsers.

The `init.js` usage explicitly mentions Internet Explorer compatibility, but **iD no longer supports Internet Explorer** (minimum browser requirements have moved past IE).

## Proposed Solution

Replace all `document.createEvent()`/`initEvent()` calls with the modern `Event` constructor:

```js
// Before (deprecated)
var evt = document.createEvent('HTMLEvents');
evt.initEvent(type, true, true);

// After (modern)
var evt = new Event(type, { bubbles: true, cancelable: true });

### PR DESCRIPTION
**Pull request by JaiswalShivang**
## Summary

Replaces all usages of the deprecated `document.createEvent()` and `initEvent()` methods with the modern [Event()] constructor across the codebase.

Closes #11870 

## Problem

The codebase uses `document.createEvent()` and `initEvent()` in 3 files. These methods are [deprecated by MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/createEvent) and have been superseded by the [[Event()] constructor](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event).

Notably, [modules/ui/init.js] contains the comment:
> "Use outdated code so it works on Explorer"

Since iD no longer supports Internet Explorer, this legacy code can be safely modernized.

### 1. [modules/util/trigger_event.js]
Core utility function used by [combobox.js]
[address.js] and [changeset_editor.js]

- Replaced `document.createEvent('HTMLEvents')` + `initEvent()` with `new Event()`
- Replaced manual property loop with `Object.assign()`

### 2. [modules/ui/init.js](line 552-557)
- Removed outdated IE compatibility comment and 3-line event creation
- Replaced with single-line `new Event()` call

### 3. [modules/renderer/map.js](line 304-314)
- Removed unnecessary `CustomEvent` existence check and `createEvent` fallback
- Simplified to direct `new Event()` constructor

## Browser Support

The [Event()] constructor is supported by all browsers that iD currently targets:
- Chrome 15+ ✅
- Firefox 11+ ✅  
- Safari 6+ ✅
- Edge 12+ ✅

[caniuse: Event constructor](https://caniuse.com/mdn-api_event_event)

## Testing

- [x] Existing tests pass (`npm test`)
- [x] Manual testing: combobox interactions, address field changes, window resizing, and map zoom all work correctly

### DIFF
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -302,7 +302,9 @@ export function rendererMap(context) {
                 }
             }
             if (hasOrphan) {
-                var event = new Event('mouseup', { view: window });
+                var event = new Event('mouseup');
+                // Event needs to be dispatched with an event.view property.
+                event.view = window;
                 window.dispatchEvent(event);
             }
         }
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -302,7 +302,7 @@ export function rendererMap(context) {
                 }
             }
             if (hasOrphan) {
-                var event = new Event('mouseup');
+                const event = new Event('mouseup');
                 // Event needs to be dispatched with an event.view property.
                 event.view = window;
                 window.dispatchEvent(event);
--- a/modules/renderer/map.js
+++ b/modules/renderer/map.js
@@ -302,15 +302,7 @@ export function rendererMap(context) {
                 }
             }
             if (hasOrphan) {
-                var event = window.CustomEvent;
-                if (event) {
-                    event = new event('mouseup');
-                } else {
-                    event = window.document.createEvent('Event');
-                    event.initEvent('mouseup', false, false);
-                }
-                // Event needs to be dispatched with an event.view property.
-                event.view = window;
+                var event = new Event('mouseup', { view: window });
                 window.dispatchEvent(event);
             }
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :hammer: Development
 * Change package name to `@openstreetmap/id` to be able to publish releases on npm
 * Use Röntgen icon set directly from upstream npm package ([#11784], thanks [@tordans])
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: modules/renderer/map.js [hunk 1]
```diff
@@ -302,15 +302,7 @@ export function rendererMap(context) {
...
-                var event = window.CustomEvent;
-                if (event) {
-                    event = new event('mouseup');
-                } else {
-                    event = window.document.createEvent('Event');
-                    event.initEvent('mouseup', false, false);
-                }
-                // Event needs to be dispatched with an event.view property.
-                event.view = window;
+                var event = new Event('mouseup', { view: window });
```
tyrasd: Is it really allowed to specify arbitrary properties like `view` here? AFAICS, it doesn't seem to be part of the [specs](https://dom.spec.whatwg.org/#ref-for-dom-event-event) of the `Event` constructor.

Maybe it would be better to set it like before:

```suggestion
                var event = new Event('mouseup');
                // Event needs to be dispatched with an event.view property.
                event.view = window;
```

JaiswalShivang:
@tyrasd Thanks for the suggestion

PR #11872, opened 2026-02-13
Add validation to flag redundant area=yes tag (fixes #9260)

### LINKED ISSUE #9260: Feature request: (Conditionally) flag the area=yes tag as a problem prior to saving the change set
### Description

Countless times have I been flagged by OSMI in the Neis daily individual report to have left the area=yes tag when not needed i.e. redundant to requirement, mostly on landuse with the combos 

1. landuse=xyz, farmland, grass, scrub, forest etc and to include highway=services (countless times mapped at fuel stations). 
2. area=yes

Could there be an alert added (or ideally auto removal), to have a look as to whether the tag can be deleted or not.

To get an idea, my current 60 day rolling creation stands at > 11,000. Lot's of chances to miss removing one or the other in a change set.

### Screenshots

_No response_

FaisalMisbah23:
- Add validationRedundantAreaYes: warns when a way has area=yes but other tags (e.g. landuse, natural, highway=services) already imply area.
- One-click fix removes the area tag to avoid QA warnings (e.g. OSMI).
- Severity: warning (does not block save).
- Add strings in data/core.yaml and tests in test/spec/validations.


matkoniecz:
1)
This would be effectively resulting in people blindly "upgrading" tags, therefore (in my opinion) requires going with bot edit approval.

It seems it was not done.

2)
Why bother user about this incredibly minor issue?

If `area=yes` can be in general purged for such cases, then it can be removed with an automated edit.

BTW, if some community wants such bot edit (including global one) - feel free to propose it in accordance with https://wiki.openstreetmap.org/wiki/Automated_Edits_code_of_conduct and I would be happy to run it in any area outside Germany. 

Or propose it and get approval and use my code. Or write own.

I am already running https://wiki.openstreetmap.org/wiki/Mechanical_Edits/Mateusz_Konieczny_-_bot_account/remove_area%3Dyes_from_buildings_in_USA and https://wiki.openstreetmap.org/wiki/Mechanical_Edits/Mateusz_Konieczny_-_bot_account/remove_area%3Dyes_from_buildings_in_Poland

In general bottleneck here is consulting edit with community.

k-yle:
i agree with Mateusz.

There are cases where iD disagrees with other OSM software about whether a certain tag implies `area=yes`. For example, a [closed way with `man_made=pier`](https://osm.org/way/24639194) is treated as a line by iD, but treated as an area by OSM Carto.

Currently this disagreement is not a big deal, but it will probably cause chaos if iD insists that its [defintion of `areaKeys`](https://github.com/openstreetmap/iD/blob/6d0a5b304a8b468493492e1d5bcd39456746f401/modules/osm/tags.js#L69-L106) is the canonical truth

---

> Countless times have I been flagged by OSMI in the Neis daily individual report to have left the area=yes tag when not needed

I think is the fundamental issue. OSMI complains about a lot of things that are arguably not real problems. and some people leave [very aggressive comments](https://osm.org/changeset/178015595), just because this validator tool claims there's an "issue".

neis-one.org is also making this worse by penalizing people for every random validator warning... This tool seems to be closed-source, so I'm not sure where to raise an issue

matkoniecz:
> This tool seems to be closed-source, so I'm not sure where to raise an issue

I wrote to Pascal via e-mail few times. They changed some things, but I have not convinced them to drop "validators must be blindly obeyed" approach.

tyrasd:
I agree with @matkoniecz and @k-yle , let's not do this in iD.

PR #11873, opened 2026-02-14
Add Select all members button for relations (fixes #9982)

### LINKED ISSUE #9982: Function to switch to the multiselection of all members of a selected relation
### Description

Currently, when a relation is selected in the editor, one can see the members and load/navigate to each member individually. However, I'd need to be able to get to the multiselection view of all the members of that relation, since I need to apply some tags to the individual members of the relation, not the relation itself (in my specific case [Key:toll](https://wiki.openstreetmap.org/wiki/Key:toll)). 

A button near the "> Relations (123)" to accomplish this would be great, and/or a keyboard shortcut.

Related to [#3604](https://github.com/openstreetmap/iD/issues/3604#issuecomment-880087261).

### Screenshots

_No response_

FaisalMisbah23:
Adds a button in the Members section when a relation is selected. Clicking it loads any unloaded members, then switches to multiselection of all members so users can apply tags (e.g. toll) to each.

- Button in Members section switches to multiselection of all relation members
- Loads unloaded members first, then selects only loaded IDs
- Empty state message when relation has no members
- Promise error handling so loading state is always cleared

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -716,6 +716,8 @@ en:
     fields: Fields
     tags: Tags
     members: Members
+    select_all_members: Select all members
+    select_all_members_tooltip: Load and select all relation members
     relations: Relations
     features: Features
     title_count: "{title} ({count})"
--- a/modules/ui/sections/raw_member_editor.js
+++ b/modules/ui/sections/raw_member_editor.js
@@ -139,6 +139,39 @@ export function uiSectionRawMemberEditor(context) {
             });
         });
 
+        const selectAllWrap = selection.selectAll('.member-select-all-wrap')
+            .data([0]);
+        selectAllWrap.enter()
+            .append('div')
+            .attr('class', 'member-select-all-wrap form-field')
+            .merge(selectAllWrap)
+            .style('display', entity.members.length === 0 ? 'none' : null)
+            .each(function() {
+                const wrap = d3_select(this);
+                wrap.selectAll('*').remove();
+                const link = wrap
+                    .append('a')
+                    .attr('class', 'member-select-all')
+                    .attr('href', '#')
+                    .attr('title', t('inspector.select_all_members_tooltip'));
+                link.call(svgIcon('#iD-icon-layers', 'inline'));
+                link.append('span').text(t('inspector.select_all_members'));
+                link.on('click', function(d3_event) {
+                    d3_event.preventDefault();
+                    const linkEl = d3_select(this);
+                    if (linkEl.classed('loading')) return;
+                    linkEl.classed('loading', true);
+                    context.loadEntity(entity.id, function(err) {
+                        linkEl.classed('loading', false);
+                        if (err) return;
+                        const updated = context.entity(entity.id);
+                        if (!updated || !updated.members.length) return;
+                        const ids = updated.members.map(function(m) { return m.id; });
+                        context.enter(modeSelect(context, ids));
+                    });
+                });
+            });
+
         var list = selection.selectAll('.member-list')
             .data([0]);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 ##### 2026-03-02
 
 #### :sparkles: Usability & Accessibility
+* Add "Select all members" link and Ctrl+↓ shortcut for relations ([#11873], fixes [#9982], thanks [@FaisalMisbah23])
 * Show warning when attempting to paste but nothing has been copied ([#9401], thanks [@omsaraykar])
 * Don't suggest values from Taginfo for `addr:*` tags ([#11733], thanks [@k-yle])
 * Don't suggest values from Taginfo for tags with less than 100 uses, even if they're documented on the wiki ([#11794], thanks [@Kaushik4141])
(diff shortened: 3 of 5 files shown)


modules/ui/sections/raw_member_editor.js
```diff
@@ -138,6 +173,43 @@ export function uiSectionRawMemberEditor(context) {
...
+
+        var selectAllButtons = selectAllWrap.selectAll('button.member-select-all')
+            .data([0]);
+
+        var selectAllButtonsEnter = selectAllButtons.enter()
+            .append('button')
+            .attr('class', 'member-select-all')
+            .attr('type', 'button')
+            .attr('title', t('inspector.select_all_members_tooltip'))
+            .call(svgIcon('#iD-icon-check'))
```
tyrasd: This icon does not seem to exist. Did you forget to include it in the PR perhaps?

data/core.yaml
```diff
@@ -716,6 +716,9 @@ en:
     fields: Fields
     tags: Tags
     members: Members
+    members_empty: This relation has no members.
```
tyrasd: How is is possible to have a relation with no members? Maybe I'm overlooking something, but IIRC any relation is automatically deleted once it contains no members anymore.

modules/ui/sections/raw_member_editor.js
```diff
@@ -79,6 +79,41 @@ export function uiSectionRawMemberEditor(context) {
...
+        d3_event.preventDefault();
+        d3_event.stopPropagation();
+        if (!_entityIDs || _entityIDs.length !== 1) return;
+        var relation = context.entity(_entityIDs[0]);
+        if (!relation || relation.type !== 'relation' || !relation.members.length) return;
+
+        var members = relation.members;
+        var missing = members.filter(function (m) { return !context.hasEntity(m.id); });
+
+        function activate() {
```
tyrasd: is `activate` really the best wording for what this function does? 🤔  Maybe simply `select` would work?

modules/ui/sections/raw_member_editor.js
```diff
@@ -79,6 +79,41 @@ export function uiSectionRawMemberEditor(context) {
...
+            if (ids.length === 0) return;
+            context.enter(modeSelect(context, ids));
+        }
+
+        if (missing.length) {
+            var button = d3_select(this);
+            button.classed('loading', true);
+            Promise.all(missing.map(function (m) {
+                return new Promise(function (resolve) {
+                    context.loadEntity(m.id, function (err) { resolve(err); });
```
tyrasd: a single `loadEntity(relation.id …)` is sufficient here: that is already downloading the relation fully including all of its members… there's no need to individually load all missing members. It should also be nicer for the API to handle: one single request instead of potentially many tiny requests.

modules/ui/sections/raw_member_editor.js
```diff
@@ -79,6 +79,41 @@ export function uiSectionRawMemberEditor(context) {
...
+        if (!_entityIDs || _entityIDs.length !== 1) return;
+        var relation = context.entity(_entityIDs[0]);
+        if (!relation || relation.type !== 'relation' || !relation.members.length) return;
+
+        var members = relation.members;
+        var missing = members.filter(function (m) { return !context.hasEntity(m.id); });
+
+        function activate() {
+            var ids = members.map(function (m) { return m.id; })
+                .filter(function (id) { return context.hasEntity(id); });
```
tyrasd: as this is already checked before (`missing = …` above) the function is called, this seems unnecessary. The same might also apply to the `ids.length === 0` check in the next line?!

modules/ui/sections/raw_member_editor.js
```diff
@@ -148,7 +220,7 @@ export function uiSectionRawMemberEditor(context) {
 
 
         var items = list.selectAll('li')
-            .data(memberships, function(d) {
+            .data(memberships, function (d) {
```
tyrasd: please roll back this and similar unnecessary changes to unrelated code.

modules/ui/sections/raw_member_editor.js
```diff
@@ -79,6 +79,41 @@ export function uiSectionRawMemberEditor(context) {
...
+            Promise.all(missing.map(function (m) {
+                return new Promise(function (resolve) {
+                    context.loadEntity(m.id, function (err) { resolve(err); });
+                });
+            })).then(function () {
+                button.classed('loading', false);
+                activate();
+            }).catch(function () {
+                button.classed('loading', false);
+            });
```
tyrasd: if this would be just a single API request, it is not really necessary to use a promisified version of this workflow, e.g. the following would be more concise:

```js
context.loadEntity(relation.id, (err, data) => {
  button.classed('loading', false);
  if (!err) …
});
```

tyrasd (changes requested):
This is a start, but I think this needs to be refined a bit to be actually useful:

I think the styling would need to be improved, as it currently does not look like any of the other buttons in the UI



Maybe instead of a button, it could be a clickable text with an icon (similar to the [one](https://github.com/user-attachments/assets/16130013-b72b-464d-89b0-7065a134abd3) proposed in #11701). That said, I'm asking myself if there is even the need to do have an always visible button for this at all. There are only few examples I can think of when selecting all members of a relation is useful while mapping. So maybe instead of a button that is always visible, it could be put away in a small "hamburger menu" that is displayed alongside the Members section of a relation, which would contain this button at first (and the _Download all members_ button from #11701 next). Later we could add more specialized options to this menu like selecting all members with a certain `role`[^1], etc.

[^1]: which might actually be much more useful for mapping, e.g. to add the _operator_ tag to all stops of a bus relation, to quickly zoom to the _inner_ ways of a multipolygon, to fix a typo in the _street_s of an `associated_street` relation, etc.

Lastly, I think it could be worth considering to include this functionality also in the existing keyboard shortcut <kbd>Ctrl</kbd> + <kbd>↓</kbd>: When a relation is selected, pressing the keyboard shortcut would select all members, similarly to how it selects all nodes of a way when a way was selected.

PS: Please consider to use modern js with `const/let` instead of `var` for newly added or refactored code. 😊

FaisalMisbah23:
All review comments have been addressed in the latest push. Ready for another look.

tyrasd:
closing PR because of insufficient quality

PR #11875, opened 2026-02-14
Fix build:css fs.globSync error (fixes #11874)

### LINKED ISSUE #11874: build:css fails with "fs.globSync is not a function" (Node 22+)
## What happens
Running `npm run build` or `npm test` fails during the CSS build step with:

TypeError: fs.globSync is not a function
at file:///.../scripts/build_css.js:29:22

## Cause
`scripts/build_css.js` calls `fs.globSync('css/**/*.css')`, but Node’s built-in `fs` module does not provide `globSync`. That API comes from the `glob` package.

## Environment
- Node: 22+ (or whatever you use)
- Repo: latest `develop`
- Command: `npm run build` or `npm test`

## Suggested fix
Use the `glob` package for globbing (e.g. `import { globSync } from 'glob'`) instead of `fs.globSync`, and keep using `fs` for read/write. The project already has `glob` as a transitive dependency.

Happy to open a PR if maintainers are okay with this approach.

FaisalMisbah23:
Fixes #11874

`scripts/build_css.js` was calling `fs.globSync()`, which is not part of Node's built-in `fs` module. This PR:

- Uses the `glob` package for globbing (`globSync('css/**/*.css')`)
- Adds `glob` as a devDependency
- Keeps `fs` for read/write only

`npm run build:css` now completes successfully.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -74,6 +74,7 @@
         "fake-indexeddb": "^6.2.5",
         "fetch-mock": "^11.1.1",
         "gaze": "^1.1.3",
+        "glob": "^10.4.5",
         "happen": "^0.3.2",
         "js-yaml": "^4.1.1",
         "jsdom": "^28.0.0",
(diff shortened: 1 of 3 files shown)


k-yle:
please see https://github.com/openstreetmap/iD/issues/11874#issuecomment-3907779707

PR #11881, opened 2026-02-17
fix: Prevent toolbar from overlapping the right-click context menu

### LINKED ISSUE #11017: Toolbar overlaps right click menu
### URL

https://www.openstreetmap.org/edit#map=20/52.6462713/-7.2449954

( Discard all / do not save any changes after testing this issue, please )

Was #10495 

Similar issue when selecting multiple points (or certain combinations of points and connected ways) resulting in too many operations to show vertically. Top toolbar overlaps first option in this case (extend way)

1. Create an arc/line with some intermediate vertices(=points on a line)
2. set the points to be some POI, i.e. trees
3. Shift+select select multiple trees
4. right click to get editing options. first one is extract. Only bottom half of description is visible, icon is not visible.

Image

2.33.0, 2.34.0-dev

Chrome

mykh-hailo:
## What is this PR

This branch fixes popover placement when using left/right placements inside scrollable containers. The update adjusts the popover's vertical position to prevent overflow beyond the scroll container, clamps the arrow position so it remains visually aligned and within the popover bounds, and includes a short-lived debug log used during verification

Closes: #11017 

## Root Cause

- Symptom: Popovers placed at "left" or "right" could overflow the vertical bounds of their container(`over-map`) and the arrow could be mis-positioned or clipped, especially when the anchor or container was scrolled or the popover size changed dynamically.
- Root cause: coordinate-space mismatch and inconsistent measurement sources during positioning updates for left/right placements.

## What changed

- Adjusted popover/context menu positioning logic to account for the toolbar's width and safe area on the right side of the viewport.
- Added a small offset and collision-detection fallback so the context menu will open to the left when there's insufficient space on the right.
- Minor CSS tweaks to ensure z-index and pointer-events are correct so the context menu is interactable and visually above the toolbar.
- Small defensive updates to ensure behavior is stable across different viewport sizes and when the toolbar is collapsed/expanded.

### DIFF
--- a/modules/ui/popover.js
+++ b/modules/ui/popover.js
@@ -304,21 +304,42 @@ export function uiPopover(klass) {
         }
 
         if (position) {
+            popoverSelection.style('left', ~~position.x + 'px').style('top', ~~position.y + 'px');
 
-            if (scrollNode && (placement === 'top' || placement === 'bottom')) {
+            if (scrollNode) {
 
-                var initialPosX = position.x;
+                if ((placement === 'top' || placement === 'bottom')) {
 
-                if (position.x + popoverFrame.w > scrollNode.offsetWidth - 10) {
-                    position.x = scrollNode.offsetWidth - 10 - popoverFrame.w;
-                } else if (position.x < 10) {
-                    position.x = 10;
+                    var initialPosX = position.x;
+
+                    if (position.x + popoverFrame.w > scrollNode.offsetWidth - 10) {
+                        position.x = scrollNode.offsetWidth - 10 - popoverFrame.w;
+                    } else if (position.x < 10) {
+                        position.x = 10;
+                    }
+
+                    var arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
+                    // keep the arrow centered on the button, or as close as possible
+                    var arrowPosX = Math.min(Math.max(popoverFrame.w / 2 - (position.x - initialPosX), 10), popoverFrame.w - 10);
+                    arrow.style('left', ~~arrowPosX + 'px');
+                } else if (placement === "left" || placement === "right") {
+
+                    var popoverRect = popoverSelection.node().getBoundingClientRect();
+                    var scrollNodeRect = scrollNode.getBoundingClientRect()
+
+                    var initialPosY = position.y;
+
+                    if (popoverRect.bottom > scrollNodeRect.bottom - 10) {
+                        position.y -= popoverRect.bottom - (scrollNodeRect.bottom - 10);
+                    } else if (popoverRect.top < scrollNodeRect.top + 10) {
+                        position.y += (scrollNodeRect.top + 10) - popoverRect.top;
+                    }
+
+                    var arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
+                    // keep the arrow centered on the button, or as close as possible
+                    var arrowPosY = Math.min(Math.max(popoverFrame.h / 2 - (position.y - initialPosY), 0), popoverFrame.h - 30);
+                    arrow.style('top', ~~arrowPosY + 'px');
                 }
-
-                var arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
-                // keep the arrow centered on the button, or as close as possible
-                var arrowPosX = Math.min(Math.max(popoverFrame.w / 2 - (position.x - initialPosX), 10), popoverFrame.w - 10);
-                arrow.style('left', ~~arrowPosX + 'px');
             }
 
             popoverSelection.style('left', ~~position.x + 'px').style('top', ~~position.y + 'px');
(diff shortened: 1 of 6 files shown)


modules/ui/popover.js
```diff
@@ -304,21 +304,42 @@ export function uiPopover(klass) {
         }
 
         if (position) {
+            popoverSelection.style('left', ~~position.x + 'px').style('top', ~~position.y + 'px');
 
-            if (scrollNode && (placement === 'top' || placement === 'bottom')) {
+            if (scrollNode) {
 
-                var initialPosX = position.x;
+                if ((placement === 'top' || placement === 'bottom')) {
```
tyrasd: double parentheses are not necessary here:

```suggestion
                if (placement === 'top' || placement === 'bottom') {
```

modules/ui/popover.js
```diff
@@ -304,21 +304,42 @@ export function uiPopover(klass) {
...
+
+                    var initialPosY = position.y;
+
+                    if (popoverRect.bottom > scrollNodeRect.bottom - 10) {
+                        position.y -= popoverRect.bottom - (scrollNodeRect.bottom - 10);
+                    } else if (popoverRect.top < scrollNodeRect.top + 10) {
+                        position.y += (scrollNodeRect.top + 10) - popoverRect.top;
+                    }
+
+                    var arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
```
tyrasd: this `var arrow` is already defined above in the `if (placement === 'top' || placement === 'bottom')` branch. Maybe change both to `const`.

```js
const arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
```

(And since we're touching this part of the code anyway: can you also upgrade the other remaining `var`iable definitions in the two `if` branches here to `const`s?)

modules/ui/popover.js
```diff
@@ -304,21 +304,42 @@ export function uiPopover(klass) {
...
+                        position.x = scrollNode.offsetWidth - 10 - popoverFrame.w;
+                    } else if (position.x < 10) {
+                        position.x = 10;
+                    }
+
+                    var arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
+                    // keep the arrow centered on the button, or as close as possible
+                    var arrowPosX = Math.min(Math.max(popoverFrame.w / 2 - (position.x - initialPosX), 10), popoverFrame.w - 10);
+                    arrow.style('left', ~~arrowPosX + 'px');
+                } else if (placement === "left" || placement === "right") {
```
tyrasd: for strings, please use single brackets `'`.
```suggestion
                } else if (placement === 'left' || placement === 'right') {
```

modules/ui/popover.js
```diff
@@ -304,21 +304,42 @@ export function uiPopover(klass) {
...
 
-                var arrow = anchor.selectAll('.popover-' + _id + ' > .popover-arrow');
-                // keep the arrow centered on the button, or as close as possible
-                var arrowPosX = Math.min(Math.max(popoverFrame.w / 2 - (position.x - initialPosX), 10), popoverFrame.w - 10);
-                arrow.style('left', ~~arrowPosX + 'px');
+                    var popoverRect = popoverSelection.node().getBoundingClientRect();
+                    var scrollNodeRect = scrollNode.getBoundingClientRect()
```
tyrasd: missing `;` 😉 
```suggestion
                    var scrollNodeRect = scrollNode.getBoundingClientRect();
```

tyrasd (dismissed):
This looks good to me, thanks!

Just a few code linting remarks (see below) that should be easy to address.

PS: I was wondering why the top/bottom branch can get away with using the slightly simpler implementation using `offsetWidth` / `w` properties of the popover/scroll nodes, while for left/right we have to use `getBoundingClientRect`s to calculate the clamped positions. I'd assume the `getBoundingClientRect` approach would also work for the top/bottom case, and it should be more robust in case we want to use it in the future for a situation where the simple calculation might not work. Also, maybe the code could be simplified slightly when the two branches use the same calculation (only x/y need to be flipped for each)… What do you think?

mykh-hailo:
@tyrasd I made some updates on your comments on my PR.
Please check and leave your comments if any.
Thank you.

mykh-hailo:
@tyrasd I'd appreciate it if you review this PR again.

mykh-hailo:
Hi, @tyrasd .
I'd appreciate it if you check my PR and merge please.
Thank you.

mykh-hailo:
@tyrasd , @k-yle 
Can you take a looks at this PR please?

tyrasd (dismissed):
Please fix the remaining linter error. Last time, we already had a duplicate `var` definition. You overlooked the same issue again in the new code added with 4bc5a4b581457412b277d13fe3c91210a4297b19. The `popoverRect` /  `scrollNodeRect` (as well as the `const arrow`) should be defined outside of the `if` statement as it is used in both branches.

Please always run the tests (`npm test`) before committing to check that everything is ok. 🙏 

One last thing, can you please extract the value `10` that is used as the minimum margin to the screen edge into a constant variable somewhere at the start of the module `const MIN_MARGIN = 10;`.

mykh-hailo:
@tyrasd I updated my code for the lint error.
And it passed all testing.
I'd appreciate it if you check this again.

PR #11896, opened 2026-02-19
feat: Prioritize parent relations of connected features in relation dropdown menu

### LINKED ISSUE #11676: Relation dropdown menu should prioritize parent relations of connected features

bhavyaKhatri2703:
fixes #11676 
this pr uses the logic given by @k-yle in https://github.com/openstreetmap/iD/pull/11683#discussion_r2622822063 to 
prioritize parent relations of connected features in relation dropdown.


matkoniecz:
On which usecase you have tested it? Can you attach screenshot/video of testing?

bhavyaKhatri2703:
@matkoniecz 

Before :

https://github.com/user-attachments/assets/c5624166-44aa-43cc-866e-d06ed8a13a12

After : pritorizes the relations of the neighouring/parent ways 


https://github.com/user-attachments/assets/fcd5b8b1-f898-4196-aa67-6cc6652a3aeb

modules/ui/sections/raw_membership_editor.js
```diff
@@ -327,7 +327,39 @@ export function uiSectionRawMembershipEditor(context) {
                 });
             });
 
+            const connectedRelationIDs = new Set();
+            const selected = graph.hasEntity(entityID);
+
+            if (selected) {
+                let connectedWays = [];
```
matkoniecz: maybe it would be nice to also make it a set to avoid the same way potentially being added multiple times?

modules/ui/sections/raw_membership_editor.js
```diff
@@ -327,7 +327,39 @@ export function uiSectionRawMembershipEditor(context) {
...
+                        const node = graph.hasEntity(nodeID);
+                        if (node) {
+                            graph.parentWays(node).forEach(w => {
+                                if (w.id !== selected.id) connectedWays.push(w);
+                            });
+                        }
+                    });
+                }
+
+                connectedWays.forEach(way => {
```
matkoniecz, line 354: I guess we can care here only about way relation members and ignore cases where nodes are relation members, as those are quire exotic

modules/ui/sections/raw_membership_editor.js
```diff
@@ -327,7 +327,39 @@ export function uiSectionRawMembershipEditor(context) {
...
+                    graph.parentRelations(way).forEach(r => {
+                        connectedRelationIDs.add(r.id);
+                    });
+                });
+            }
+
             result.sort(function(a, b) {
+                const aConnected = connectedRelationIDs.has(a.relation.id);
```
matkoniecz: are more clear variable names possible? Admittedly I tries to find some that would not require me to drawing diagram to understand logic of this sort and failed.

matkoniecz: pr maybe just put here comment of `// if exactly one of two relations is connected to our selection - prefer it over unconnected`

or is it stating obvious and I for some reason have trouble to understand obvious here?

bhavyaKhatri2703:
@matkoniecz ive made the suggested changes, pls have a look again, thanks.

matkoniecz (commented):
looks fine to me, also in testing

I worry a bit about performance hit, I have not done profiling yet

PR #11898, opened 2026-02-20
Sidebar auto-scroll when dragging relation members near top/bottom

### LINKED ISSUE #11897: sidebar doesn't scroll when dragging relation members to the top or bottom edge
### URL

_No response_

1. Edit a relation for a long bus route
2. Try to drag a bus stop to reorder it - e.g. a bus stop you've just added, which is at the end but must be moved to somewhere in the middle
3. keep dragging, and drag to the top of the sidebar

Expected: should start scrolling to keep going up in the order

Actual: no scrolling. Have to drop, scroll, repeat drag.

_No response_

_No response_

_No response_

_No response_

FaisalMisbah23:
Fixes #11897

When dragging a relation member in the raw member editor, the sidebar now auto-scrolls when the pointer is near the top or bottom edge (within 40px), so long member lists can be reordered without manually scrolling first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 ##### 2026-03-02
 
 #### :sparkles: Usability & Accessibility
+* Sidebar auto-scrolls when dragging relation members near top or bottom edge (fixes [#11897], thanks [@FaisalMisbah23])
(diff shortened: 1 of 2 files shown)


FaisalMisbah23:
Friendly ping, any thoughts on this sidebar scroll fix? I've rebased onto the current development, tested it locally, and can provide more details/screenshots if helpful. Thanks!

PR #11899, opened 2026-02-20
Fix move (M/drag) not working after toggling Data panel with node selected (#9582)

### LINKED ISSUE #9582: It's not possible to move nodes in a specific case
### URL

https://www.openstreetmap.org/edit#map=20/42.65126/23.36784

1. Go to the data panel and disable everything except the two types of roads
2. Try to move the node with crossing tags (node/10795903750)
3. With that node selected, enable paths from the data panel
4. Try to move that node (notice that it's not possible and instead the map pans)
5. To fix it, you have to deselect the node

~~Trying to press M to move the node at step 4 also does not work~~ It works currently

_No response_

2.25.1

Firefox

FaisalMisbah23:
Fixes #9582

**Problem:** After toggling a layer or map-feature checkbox in the Data panel (e.g. enabling "Paths") with a node selected, moving the node via **M** or drag did not work—the map panned instead. Workaround was to click outside the checkbox so focus left the input.

**Cause:** Focus stayed on the checkbox (input element), so keyboard shortcuts were not handled and interaction was wrong.

**Fix:** Blur the active element after toggling in both Map Features (`map_features.js`) and Data layers (`data_layers.js`) so focus leaves the panel and M/drag work without an extra click.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Add `railway=crossing` tag when connecting two railways using the validation fix ([#9226], thanks [@paulklie])
 #### :bug: Bugfixes
 * Fix typo: `parking:left:capacity` duplicated in osmSummableTags, missing `parking:right:capacity` ([#11819], thanks [@JaiswalShivang])
+* Fix move (M / drag) not working after toggling Data panel layer or Map Features checkbox with a node selected; blur focus from checkbox so shortcuts work ([#9582])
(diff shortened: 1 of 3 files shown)


FaisalMisbah23:
Friendly ping, any thoughts on this data panel focus/move fix? I've rebased onto the current development, tested it locally, and can provide more details/screenshots if helpful. Thanks!

tyrasd:
Removing the focus would not work, as it will break using the UI input forms with a keyboard. This would need to be addressed differently.

PR #11902, opened 2026-02-20
PoC: Automatically split geographically distant edits into linked changesets\

### LINKED ISSUE #5424: Multiple changesets for widely separated regions
It seems to be relatively common to see changesets that appear to cover wide regions of the globe, but when examined actually consist of two unrelated edits in different regions. 

A random example of this may be seen here: https://osmcha.mapbox.com/changesets/63732158/

It would be nice if iD were capable of detecting these widely separated edits and encouraging them to be uploaded separately.

The detection for this could be quite coarse as the most egregious of these can be separated by 100 deg of longitude. 

My naive suggestion may be:
To run through the coordinates of the nodes of the edited nodes and ways and flag which 10 degree bands of latitude and longitude have edits and which do not (an alternative description may be: to create a histogram in which we only care whether a bar exists, not how high it is). 
If there is a band without edits between two bands that do have edits, then this is a sign that the user should be prompted as to whether they really want to upload in one go and provide a means for splitting the upload if desired.

Sembauke:
This change starts from one practical problem that is most visible with new users: a single editing session can contain unrelated edits spread far apart, and uploading them as one changeset makes review noisy, confusing, and harder to trust. New mappers often make exploratory edits in multiple places before saving, so they are the group most likely to run into this.

The goal here was to let iD separate those edits automatically, while still keeping dependent objects together so uploads remain valid.

The core of that work is a new splitter in `modules/core/changeset_splitter.js`, exported from `modules/core/index.js` as `coreChangesetSplitter`. It builds connected components from the changed entities with union-find, linking ways to their nodes and relations to their members. After that, it uses location data and extent checks to cluster components geographically. It keeps an early return for small local edits, handles deleted entities by resolving locations from changed data when graph entities are missing, and includes cycle-safe traversal for relation membership.

`modules/core/uploader.js` now uses that splitter before upload. Instead of one `putChangeset` call with everything, it uploads groups sequentially. It tracks split state and uploaded entity IDs so local history can be reverted group by group as each part succeeds. Comment text is also annotated with part information, and later parts reference the first uploaded changeset URL. This gives clearer auditability when one editing session becomes multiple uploads.

On the Save screen, `modules/ui/sections/changes.js` now renders grouped change lists when splitting occurs, rather than a single flat list. The warning message was updated to explicitly tell the mapper that edits are too far apart and will be uploaded separately. Related strings were added in `data/core.yaml`, and `css/80_app.css` adds spacing and titles for grouped sections so the layout stays readable.

The test coverage for this feature was built out in three new files: `test/spec/core/changeset_splitter.js`, `test/spec/core/uploader.js`, and `test/spec/ui/sections/changes.js`. The splitter tests cover structural integrity, geographic partitioning, deleted-entity behavior, cycle handling, and edge cases. The uploader tests cover sequential multi-group behavior and split comment linking. The UI tests cover split messaging and grouped rendering behavior.

ref: #5424

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -5483,6 +5483,16 @@ a.user-info {
     opacity: 0.5;
 }
 
+.changeset-group + .changeset-group {
+    margin-top: 10px;
+}
+
+.changeset-group-title {
+    margin: 0 0 4px 2px;
+    font-size: 12px;
+    font-weight: 600;
+}
+
 
 /* Conflict resolution
 ------------------------------------------------------- */
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -611,6 +611,8 @@ en:
     save: Upload
     cancel: Cancel
     changes: Changes
+    multiple_changesets_info: "Some edits are too far apart and will be uploaded separately as {count} changesets. Review each group below."
+    changeset_group: "Changeset {num}"
     download_changes: Download osmChange file
     errors: Errors
     warnings: Warnings
--- a/modules/core/changeset_splitter.js
+++ b/modules/core/changeset_splitter.js
@@ -1,0 +1,323 @@
+import { geoExtent } from '../geo/extent';
+
+
+function UnionFind() {
+    this.parent = {};
+    this.rank = {};
+}
+
+UnionFind.prototype.find = function(x) {
+    if (this.parent[x] === undefined) {
+        this.parent[x] = x;
+        this.rank[x] = 0;
+    }
+    if (this.parent[x] !== x) {
+        this.parent[x] = this.find(this.parent[x]);
+    }
+    return this.parent[x];
+};
+
+UnionFind.prototype.union = function(a, b) {
+    var ra = this.find(a);
+    var rb = this.find(b);
+    if (ra === rb) return;
+    if (this.rank[ra] < this.rank[rb]) { var tmp = ra; ra = rb; rb = tmp; }
+    this.parent[rb] = ra;
+    if (this.rank[ra] === this.rank[rb]) this.rank[ra]++;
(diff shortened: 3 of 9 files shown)


k-yle (commented):
this is already extremely complex, and will get even more complex once we consider split merge conflicts.

For example, if the first changeset chunk has a conflict, and while resolving that conflict, it introduces a new merge conflict with the second changeset chunk (which has to be "rebased"[^1]). I can imagine this happening for very long route relations, or relations that contain coastline segments.

---

I think a much simpler alternative would be to _teach_ new users how to edit well, instead of trying to repair the situation. 

For example, if they navigate too far away from the centroid of their current changeset, then show a warning like this which blocks futher editing:

Image

This works nicely because it allows the user to gradually increase the size of their changeset (for example, extending a bus route relation), but they can't "jump" to a completely different place


[^1]: by prompting the user _again_ to resolve the same conflict?

Sembauke:
Sure I will take another look at this and simplify it when I have time.

matkoniecz:
@Sembauke do you plan to do

> I think a much simpler alternative would be to teach new users how to edit well, instead of trying to repair the situation.

(likely in a new PR?)

Sembauke:
Hey @matkoniecz,

I do not have any plans to work on this at the moment. Feel free to open a pr to possibly fix this issue.

tordans:
I suggest we extract https://github.com/openstreetmap/iD/pull/11902#pullrequestreview-3846729880 into a new issue that someone might pick up and close this for now. (I am AFK and can not do it ATM.)

Sembauke:
@tordans Sure, I can close this. No worries.

PR #11905, opened 2026-02-22
Fix TypeError in renderValue() for unrecognized ISO country codes

### LINKED ISSUE #11904: renderValue() crashes with TypeError for unrecognized ISO country codes in combo.js
### URL

https://www.openstreetmap.org/edit#map=5/51.500/-0.100

1. Open iD editor
2. Select an entity and set a country key (e.g., `addr:country`)
3. Type an unrecognized country code not in the [buildCountry()]
4. The combo field crashes with a TypeError

**Root cause:**

In [modules/ui/fields/combo.js]
The [renderValue()] function calls a d3 selection object as a function instead of using `.text()`:

```js
if (osmIsoCountryKeys.has(field.key) && tval) {
    const data = buildCountry()[tval];
    if (data) return selection => addFlagIcon(selection, data.name, data.flag);
    return selection => selection(tval);  // ← BUG: selection is a d3 object, not callable
}

### PR DESCRIPTION
**Pull request by JaiswalShivang**
## Summary

Fixes a TypeError crash in [modules/ui/fields/combo.js] where [renderValue()] calls a d3 selection object as a function when encountering an unrecognized ISO country code.

## Bug

When a country-related field (e.g., `addr:country`) has a value not found in the [buildCountry()] map, the fallback path executes:

```js
return selection => selection(tval);

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,8 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 [#11783]: https://github.com/openstreetmap/iD/pull/11783
 [#11861]: https://github.com/openstreetmap/iD/pull/11861
 [#11862]: https://github.com/openstreetmap/iD/pull/11862
+[#11870]: https://github.com/openstreetmap/iD/issues/11870
+[#11904]: https://github.com/openstreetmap/iD/issues/11904
 [@ilias52730]: https://github.com/ilias52730
 [@Razen04]: https://github.com/Razen04
 [@homersimpsons]: https://github.com/homersimpsons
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -66,6 +66,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 * Change package name to `@openstreetmap/id` to be able to publish releases on npm
 * Use Röntgen icon set directly from upstream npm package ([#11784], thanks [@tordans])
 * Replace deprecated `document.createEvent`/`initEvent` with modern [Event] constructor ([#11870], thanks [@JaiswalShivang])
+* Fix crash in country combo field when entering unrecognized ISO country codes ([#XXXX], thanks [@JaiswalShivang])
```
1ec5: Remember to fill this in the with actual PR or issue number. Thanks!

JaiswalShivang: fixed

PR #11906, opened 2026-02-22
Fix region-specific presets in Recently Used list (#11405)

### LINKED ISSUE #11405: Presets for motorway link, trunk link, primary link won’t show up as recently used presets in some regions

Kayd-06:
Fixes #11405

## Summary

Region-specific presets (e.g. `motorway_link-US`) could occupy slots in the
“Recently Used” list but not appear in the UI when editing in a location
where that preset variant was invalid.

## Problem

`presetIndex.defaults` filtered presets by `locationSet` after selecting
recent presets. If a region-specific preset was invalid for the current
location, it was removed without attempting to replace it, which could
result in unused (“ghost”) slots in the Recently Used list.

## Solution

Update `presetIndex.defaults` to:

- Detect when a recently used preset is invalid in the current location
- Attempt to replace it with a semantically equivalent base preset using `matchTags`
- Ensure strict tag equivalence using `isEqual` (to avoid unintended preset switching)
- Filter the collection after replacement

If no equivalent preset exists, the invalid regional preset is removed.

## Tests

Added regression tests to `test/spec/presets/index.js` to verify:

- Regional variants are returned in valid regions
- Regional variants are replaced by their base preset outside their region
- Invalid regional presets are removed if no equivalent exists

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -1,4 +1,5 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
+import { isEqual } from 'lodash-es';
 
 import { prefs } from '../core/preferences';
 import { fileFetcher } from '../core/file_fetcher';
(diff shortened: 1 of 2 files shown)


matkoniecz:
in which way, if any, it is superior to #11810 ?

PR #11908, opened 2026-02-22
Fix: Prevent boundary relation members from rendering as areas

### LINKED ISSUE #11900: boundary with no blue-white way around some areas
### URL

https://www.openstreetmap.org/edit?editor=id&relation=18921893#map=17/44.153871/4.803449

- follow the link above
- this boundary relation is composed of several closed areas
- the area at the bottom of the screen is (partially) visible (blue-white way around the area) because this area is composed of several ways
- the area at the top of the screen has no blue-white way around the area because the area is composed of one closed way

_No response_

2.37.3

Firefox

1-navneet:
Fixes #11900

Boundary relation members were being rendered as filled areas.
This change excludes boundary members from the area layer so they render as lines.

Verified that buildings, landuse areas, and multipolygon relations are unaffected.

### DIFF
--- a/modules/svg/areas.js
+++ b/modules/svg/areas.js
@@ -1,8 +1,8 @@
 import deepEqual from 'fast-deep-equal';
 import { bisector as d3_bisector } from 'd3-array';
+import { svgPath, svgSegmentWay, svgRelationMemberTags } from './helpers';
 
 import { osmEntity } from '../osm';
-import { svgPath, svgSegmentWay } from './helpers';
 import { svgTagClasses } from './tag_classes';
 import { svgTagPattern } from './tag_pattern';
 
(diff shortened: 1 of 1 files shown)


modules/svg/areas.js
```diff
@@ -94,7 +94,19 @@ export function svgAreas(projection, context) {
 
         for (var i = 0; i < entities.length; i++) {
             var entity = entities[i];
-            if (entity.geometry(graph) !== 'area') continue;
+            // Skip non-area geometries
+        if (entity.geometry(graph) !== 'area') continue;
```
matkoniecz, line 98: it looks like something went wrong with indentation

modules/svg/areas.js
```diff
@@ -94,7 +94,19 @@ export function svgAreas(projection, context) {
...
+
+        // Boundary relation members should render as lines
```
matkoniecz, line 100: what we should do with sad cases of people adding say building as boundary relation member?

give up and declare it as an unsupported bad tagging?

modules/svg/areas.js
1-navneet: done

modules/svg/areas.js
1-navneet: Good point. Currently the change excludes any boundary relation member from the area layer. That would also suppress legitimate area features like buildings if they are added as relation members. I’ll refine the condition to only exclude members that are contributing to boundary geometry rather than all members.

modules/svg/areas.js
matkoniecz: this does not appear resolved to me

PR #11912, opened 2026-02-22
Fix readOnlyTags bypass in updateChangeset()

### LINKED ISSUE #11911: `readOnlyTags` protection in [updateChangeset()] is non-functional due to `indexOf` vs regex mismatch
### URL

https://www.openstreetmap.org/edit#map=2/0/0

## Summary

In [modules/ui/commit.js], the `readOnlyTags` guard in [updateChangeset()] uses `Array.indexOf()` to check if a tag key matches — but the `readOnlyTags` array contains **regex objects**, not strings. Since `indexOf` uses strict equality (`===`), a string like `"created_by"` never matches a regex like `/^created_by$/`, so the check always returns `-1` and **no tag is ever blocked**.

## Root Cause

The `readOnlyTags` array (lines 20-32):

```js
var readOnlyTags = [
    /^changesets_count$/,
    /^created_by$/,
    /^ideditor:/,
    /^imagery_used$/,
    /^host$/,
    /^locale$/,
    /^warnings:/,
    /^resolved:/,
    /^closed:note$/,
    /^closed:keepright$/,
    /^closed:osmose:/
];

### PR DESCRIPTION
**Pull request by JaiswalShivang**
## Summary

Fixes #11911 

The `readOnlyTags` guard in [updateChangeset()] ([modules/ui/commit.js] was using `Array.indexOf()` to check tag keys against an array of regex objects. Since `indexOf` uses strict equality (`===`), a string like `"created_by"` never matches a regex like `/^created_by$/`, so the check always returned `-1` and no tag was ever blocked.

## Changes

Changed the check from:
```js
if (readOnlyTags.indexOf(k) !== -1) return;
to
if (readOnlyTags.some(tag => k.match(tag))) return;

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,6 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Fix some gpx/geojson properties not visible, such as numbers or complex data structures ([#11636], thanks [@k-yle])
 * Fix error setting custom background ([#11862], thanks [@Kayd-06])
 * Fix crash when commenting/closing notes when the note is closed by another mapper in the meantime ([#8464])
-* Fix `readOnlyTags` protection bypass in changeset editor where `indexOf` was used instead of regex matching ([#11911], thanks [@JaiswalShivang])
 * Fix confusing 'Point should be a point' warning ([#11589], thanks [@k-yle])
 #### :earth_asia: Localization
 * Add Moroccan phone number formats ([#11651], thanks [@ilias52730])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Fix some gpx/geojson properties not visible, such as numbers or complex data structures ([#11636], thanks [@k-yle])
 * Fix error setting custom background ([#11862], thanks [@Kayd-06])
 * Fix crash when commenting/closing notes when the note is closed by another mapper in the meantime ([#8464])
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/ui/commit.js [hunk 1]
```diff
@@ -553,7 +553,7 @@ export function uiCommit(context) {
         Object.keys(changed).forEach(function(k) {
             var v = changed[k];
             k = context.cleanTagKey(k);
-            if (readOnlyTags.indexOf(k) !== -1) return;
+            if (readOnlyTags.some(tag => k.match(tag))) return;
```
k-yle: given that this `if` statement has been evaluating to `false` for the last 8 years, we could probably just delete this line, rather than changing the behaviour.

it's fine to run `cleanTagValue` on readonly keys, since we already do this further down in the file.

k-yle (commented):
nice find lol

modules/ui/commit.js
JaiswalShivang: Thanks for the suggestion removed the line

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 * Fix some gpx/geojson properties not visible, such as numbers or complex data structures ([#11636], thanks [@k-yle])
 * Fix error setting custom background ([#11862], thanks [@Kayd-06])
 * Fix crash when commenting/closing notes when the note is closed by another mapper in the meantime ([#8464])
+* Fix `readOnlyTags` protection bypass in changeset editor where `indexOf` was used instead of regex matching ([#11911], thanks [@JaiswalShivang])
```
k-yle: nice work finding this oddity, but the 'fix' shouldn't have any noticable impact for users or developers.

so, do we need a changelog entry for a PR that just removes dead code?

PR #11913, opened 2026-02-23
Fix taxiway connection to runway centerline validation

### LINKED ISSUE #11169: Render area:aeroway
### Description

Render aeroway mapped as area with [`area:aeroway`](https://taginfo.openstreetmap.org/keys/area:aeroway), mainly `area:aeroway=runway` for runways.
Render as area, identical to existing rendering when `aeroway=runway` is combined with `area=yes`.
(Old rendering should stay for the time being as the tag is still in transition, or might go into transition.)

See https://wiki.openstreetmap.org/wiki/Tag:aeroway%3Drunway for information

### Screenshots

Example https://www.openstreetmap.org/way/1410575668
Image

should render as
Image

`aeroway=runway` on its own should only render when mapped as way (centerline), of course.

whileds:
This PR fixes aeroway validation when runways are mapped as an area with a separate centerline.

Changes:
- Ensure taxiways connect to the runway centerline (aeroway=runway way), not just the runway area polygon.
- Prevent false “Crossing aeroways should be connected” warnings when the correct centerline connection exists.
- Update crossing aeroway validation logic to prefer runway centerlines for topology checks.
- Add/adjust tests to cover area runway + centerline cases.

Result:
- Correct topology for routing and validation.
- Area runways render visually, centerlines handle connections as intended.

Fixes #11169

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11920,6 +11920,314 @@
         }
       }
     },
+    "node_modules/netlify-cli/node_modules/@rollup/rollup-android-arm-eabi": {
+      "version": "4.52.2",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.52.2.tgz",
+      "integrity": "sha512-o3pcKzJgSGt4d74lSZ+OCnHwkKBeAbFDmbEm5gg70eA8VkyCuC/zV9TwBnmw6VjDlRdF4Pshfb+WE9E6XY1PoQ==",
+      "cpu": [
+        "arm"
+      ],
+      "dev": true,
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "peer": true
+    },
+    "node_modules/netlify-cli/node_modules/@rollup/rollup-android-arm64": {
+      "version": "4.52.2",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.52.2.tgz",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -11920,6 +11920,314 @@
         }
       }
     },
+    "node_modules/netlify-cli/node_modules/@rollup/rollup-android-arm-eabi": {
```
matkoniecz, line 11923: is it needed to mix-in dependency changes?

whileds: Thanks for checking. The package-lock.json changes was generated when running tests locally.
The code change themselves do not depend on new functionality from this dependency.

If you prefer, I can revert the package-lock.json change and keep this PR limited strictly to the validation logic and tests.

matkoniecz: > The code change themselves do not depend on new functionality from this dependency.

 then please revert changes to package-lock.json

matkoniecz:
> This PR fixes aeroway validation when runways are mapped as an area with a separate centerline.

> Fixes https://github.com/openstreetmap/iD/issues/11169

#11169 is titled "Render area:aeroway"

How changes to validation are supposed to fix issue about rendering area:aeroway?

package-lock.json
matkoniecz: But there is more fundamental confusion, see comment below

k-yle:
as Mateusz mentioned, this PR appears to so something completely different to the issue description. 

The original issue has already been fixed by https://github.com/openstreetmap/id-tagging-schema/pull/1648, so I'm going to close this

PR #11914, opened 2026-02-23
Enable more ESLint rules

hlfan:
Force big objects to be sorted and disallow chained assignment expressions.

A continuation of #11342 and #11481, and the reasoning behind #11540, which @matkoniecz questioned lately.


matkoniecz:
I may be being dim - but what is the benefit here?

hlfan:
... in objects, aka them being sorted.

modules/ui/intro/helper.js
```diff
@@ -52,19 +52,19 @@ export function helpHtml(id, replacements) {
     if (!helpStringReplacements) {
         helpStringReplacements = {
         // insert icons corresponding to various UI elements
-        point_icon: icon('#iD-icon-point', 'inline'),
```
tyrasd, line 55: I think the `no-multi-assign` does actually make the code more readable in those places, but I'm not sure about the `sort-keys` rule. For example, here it seems the replacements were grouped logically, e.g. the generic point/line/area icons "belong together" and are listed here in the order they are used in the UI. 🤷

tyrasd (dismissed):

modules/ui/intro/helper.js
```diff
@@ -50,6 +50,7 @@ var helpStringReplacements;
 export function helpHtml(id, replacements) {
     // only load these the first time
     if (!helpStringReplacements) {
+/* eslint-disable sort-keys */
```
tyrasd:
```suggestion
    /* eslint-disable sort-keys */
```

modules/ui/intro/helper.js
```diff
@@ -146,6 +147,7 @@ export function helpHtml(id, replacements) {
         help: t.html('help.title'),
         ok: t.html('intro.ok')
         };
+/* eslint-enable sort-keys */
```
tyrasd:
```suggestion
    /* eslint-enable sort-keys */
```

tyrasd (commented):
Question: Did you remove the `'no-multi-assign'` validation on purpose now, or was that an accident while force-pushing to the branch?

(also see a minor nit-picky indentation suggestion below)

hlfan:
> Question: Did you remove the `'no-multi-assign'` validation on purpose now, or was that an accident while force-pushing to the branch?

Ooh, sorry, I misinterpreted this statement of yours:

> I think the `no-multi-assign` does actually make the code more readable in those places

Re-adding now...

tyrasd (approved):

PR #11924, opened 2026-02-24
Make colour marker dot more noticeable

matkoniecz:
spotted when reviewing https://github.com/openstreetmap/iD/pull/11052/changes

this dot is nice but just tad too small

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2912,8 +2912,8 @@ img.tag-reference-wiki-image {
     margin-left: -5px;
     margin-right: 5px;
     border-style: solid;
-    border-width: 4px;
-    border-radius: 4px;
+    border-width: 6px;
+    border-radius: 6px;
     border-color: inherit;
 }
 .combobox-parent-relation .has-colour::before  {


matkoniecz:
https://pr-11924--ideditor.netlify.app/#disable_features=boundaries&map=18.56/50.05579/19.85611&background=Bing&id=w1046894398

https://www.openstreetmap.org/edit?editor=id#map=19/50.054835/19.854406

screen-2026-02-25-03-22-42

tyrasd:
with #11052, the way the colored circles are drawn has changed: The `has-colour` class is now only used in the dropdown to select a new relation to add (where the smaller circle radius seems reasonable to save space). I guess the CSS should be cleaned up to remove the now unused cases of the `has-colour` class. :see_no_evil: 

The following should do the trick to make the cicles a little bigger with the current code (when there is no `ref` tag on the route):

```css
[…].member-entity-ref-color […] {
  …
  min-height: 12px;
  min-width: 12px;
}
```

matkoniecz:
I guess I will recreate it if it will irritate me again.

PR #11932, opened 2026-02-26
Fix variable shadowing bug in utilCombinedTags sort

### LINKED ISSUE #11931: Bug: utilCombinedTags tag value sort broken due to variable shadowing
1. Select multiple entities that share the same tag key but with different values
   For example, select 5 roads: 3 with highway=residential and 2 with highway=service
2. Open the tag editor panel
3. Look at the values shown for the highway tag

## Description

In [modules/util/util.js], the [utilCombinedTags()] function (lines 396-412) is supposed to sort multi-entity tag values by frequency (most common first), then alphabetically. However, the sort callback contains a variable shadowing bug that completely breaks the frequency-based sorting.

## Expected Behavior
Values should be sorted by frequency: residential (3 occurrences) should appear before service (2 occurrences).

## Actual Behavior
Values are sorted alphabetically (residential before service), ignoring frequency entirely. This happens to look correct for this specific example, but with other tag values (e.g., surface=asphalt with 1 occurrence vs surface=gravel with 5 occurrences), asphalt would incorrectly appear first.

### Problematic Code

```js
// Lines 396-411
for (var key in tags) {
    if (!Array.isArray(tags[key])) continue;

    tags[key] = tags[key].sort(function(val1, val2) {
        var key = key; // capture  ← BUG: shadows outer key
        var count2 = tagCounts[key + '=' + val2];
        var count1 = tagCounts[key + '=' + val1];
        if (count2 !== count1) {
            return count2 - count1;
        }
        // ...falls through to alphabetical
    });
}

### PR DESCRIPTION
**Pull request by JaiswalShivang**
Fixes #11931 
In [utilCombinedTags()], the sort callback on line 401 does `var key = key` 
intending to capture the outer loop variable. Due to JavaScript's `var` hoisting, 
this actually creates a new [key] that is `undefined`, shadowing the outer one.

This means `tagCounts[key + '=' + val]` always looks up `"undefined=..."` which 
returns `undefined`, so the frequency comparison is skipped entirely. Tag values 
are sorted alphabetically instead of by frequency when multiple entities are selected.

### Changes
- Changed `var` to [let] in the `for...in` loop for proper block scoping
- Removed the broken `var key = key; // capture` line

### Why this works
`Array.sort()` runs synchronously, so the callback executes immediately while 
the loop variable still holds the correct value. With [let], each iteration 
gets its own binding — no capture needed.

### CODE REVIEW

**Comment by matkoniecz** [comment]
On which use case you have tested this?

File: modules/util/util.js [hunk 1]
```diff
@@ -393,12 +393,11 @@ export function utilCombinedTags(entityIDs, graph) {
         });
     });
 
-    for (var key in tags) {
+    for (let key in tags) {
```
k-yle: interesting find, this is indeed a logical error in the code, and the fix LGTM to me.

but it would be good if you could answer mateusz's question before we continue.

JaiswalShivang: yeah sorry i was unable to reply him cause i was working on another pr the custom background feature

JaiswalShivang:
@matkoniecz  Sir I have tested on the preview deployment by selecting multiple road segments 
at an intersection in Berlin (mix of Primary and Secondary roads).

The multi-entity tag editor renders correctly — no errors or regressions. 
Combined tag values display as expected when clicking "Multiple Values" dropdowns.

The bug itself is a classic JavaScript `var` hoisting issue. Inside the sort 
callback, `var key = key` hoists the declaration, so both sides of the 
assignment refer to the inner (undefined) variable — the outer loop 
variable is never captured. You can confirm this in any JS console:

```js
var x = 10;
(function() { var x = x; console.log(x); })(); // undefined

File: modules/util/util.js [hunk 2]
```diff
@@ -393,12 +393,11 @@ export function utilCombinedTags(entityIDs, graph) {
...
         if (!Array.isArray(tags[key])) continue;
 
         // sort values by frequency then alphabetically
         tags[key] = tags[key].sort(function(val1, val2) {
-            var key = key; // capture
```
tyrasd, line 406: I'm intrigued what this was supposed to do in the first place. @quincylvania it looks like this was added in https://github.com/openstreetmap/iD/commit/1b331bb678160b33fcb2862ae3696e1af0200df3#diff-212effbbd75b34b6d4cfba40fc7787b8b1fed506653fe5636342478ed1f12d0dR303, do you remember the reasoning behind the odd line here. //edit: was it perhaps to try to pacify a linter rule like https://eslint.org/docs/latest/rules/no-loop-func?

modules/util/util.js
tyrasd:
```suggestion
    for (const key in tags) {
```

why not `const`?

tyrasd (changes requested):
we should add a unit test for the `utilCombinedTags` function that tests whether the tags are sorted as expected.

modules/util/util.js
JaiswalShivang: Yes sir i totally agree with you, beccause the key is never assigned so changing it to const

JaiswalShivang:
@tyrasd Sir , I have
Addressed the review feedback:
- Changed `var` to `const` in the for...in loop
- Added unit test for [utilCombinedTags](iD/modules/util/util.js:328:0-414:1) frequency sorting

tyrasd (approved):

PR #11943, opened 2026-02-28
ref: simplify node occurrence check in validationImpossibleOneway

dipanshurdev:
## Description
Optimize node occurrence detection in impossible oneway validation

Improves code efficiency and readability by replacing a manual loop 
with native Array methods. The function now uses indexOf/lastIndexOf 
comparison to detect duplicate nodes in a way.

## Type of Change
- [x] Performance optimization
- [x] Code quality improvement
- [ ] Bug fix
- [ ] New feature

## Changes
- Replaced manual loop iteration with `indexOf()` and `lastIndexOf()` comparison
- Reduced from 7 lines to 1 line
- More idiomatic JavaScript using native array methods
- No behavioral changes - same validation logic

## Performance Impact
- Single algorithm pass instead of iterating array
- Leverages optimized native browser/runtime methods
- Reduces memory allocation (no counter variable)

## Testing
- [x] No functional changes - same validation behavior
- [x] Detects duplicate nodes the same way
- [x] Code is clearer and more maintainable

### DIFF
--- a/modules/validations/impossible_oneway.js
+++ b/modules/validations/impossible_oneway.js
@@ -29,14 +29,7 @@ export function validationImpossibleOneway() {
         }
 
         function nodeOccursMoreThanOnce(way, nodeID) {
-            let occurrences = 0;
-            for (const index in way.nodes) {
-                if (way.nodes[index] === nodeID) {
-                    occurrences++;
-                    if (occurrences > 1) return true;
-                }
-            }
-            return false;
+            return way.nodes.indexOf(nodeID) !== way.nodes.lastIndexOf(nodeID);
         }
 
         function isConnectedViaOtherTypes(way, node) {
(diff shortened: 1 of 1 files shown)


modules/validations/impossible_oneway.js
```diff
@@ -29,14 +29,7 @@ export function validationImpossibleOneway() {
...
         function nodeOccursMoreThanOnce(way, nodeID) {
-            let occurrences = 0;
-            for (const index in way.nodes) {
-                if (way.nodes[index] === nodeID) {
-                    occurrences++;
-                    if (occurrences > 1) return true;
-                }
-            }
-            return false;
+            return way.nodes.indexOf(nodeID) !== way.nodes.lastIndexOf(nodeID);
```
tyrasd: This would not return `true` when the `nodeID` is not in the way nodes list. So, at the very least it should be something like this

```js
            const firstIndex = way.nodes.indexOf(nodeID);
            return firstIndex !== -1 && firstIndex !== way.nodes.lastIndexOf(nodeID);
```

That said, it would be interesting to know if the new approach is any faster than the old `for` loop. Can you do some benchmarking and check, please?

PR #11944, opened 2026-02-28
render the `turnout_side` tag

hlfan:
When a node connects multiple railway ways, the code analyzes the angles between neighboring track segments to determine the turnout’s facing branch. It then applies the correct left/right offset relative to that geometry, sharing the rendering of #10303.


matkoniecz:
can you attach screenshot from test, if you tested this?

hlfan:
Sure:

ezgif-7d43bd7c4c093961

tyrasd (commented):
In general, this looks good to me. Only two suggestions:

* For "unrealistic" situations where it is really unclear where the right/left side of the switch is supposed to be (e.g. a node with 3 ways at approx. 120° angles between each other), it would be probably best to not show the direction at all. So maybe it would make sense to limit the arrow to situations where there is one well defined "straight" section only?
* IMHO it would be preferred to use `const` for all variables that are not mutated in the code after the initial assignment.

Issue #10981, opened 2025-04-22
Add selected line to relation of other selected line


tordans:
### Description

For tags, we have this nice feature that when I select multiple ways and edit the tags in the "Tags" section the change gets applied to both ways. I use this to sync tags between ways by selecting one segment that has the tags I need and one that does not. The tags for both are then visible as `*` (because they are not the same) but when I edit the value the UI shows what is present for the "source way" and applies it to the "target way".

I was trying to do the same for relations. But this part of the UI works differently:
When I select a way that has a relation and one that does not (yet)…
- … the dropdown does not show (only) the source-way relation(s) but all => it would be great if the drowdown now only showed what is there already; or at least put those that are present in one of the ways at the top or in a separate select group
- … then I re apply the relation, it gets added a second time to the source-way; which is now part of the relation twice (without warning)? => it would be great if in this case only the target-way would get the relation since the source way already has it

Example: 
<img width="850" alt="Image" src="https://github.com/user-attachments/assets/16a260de-5525-4fb4-8610-aef2e5cb3b54" />

danieldegroot2:
This issue is a duplicate of (part of) #2152 .
I would suggest merging comments there.

aditya1010-max:
I’ve opened a PR #11949  that addresses your first point about the relations dropdown.

Currently, the dropdown lists nearby relations, but it’s not clear which ones are already applied to the current selection. This PR shows that information in the UI.

Relations applied to all selected entities are marked as “Applied”
Relations applied to some selected entities are marked as “Partially Applied”

This does not prevent duplicate relation members from being created or change any relation logic. The goal is to make the existing behavior more visible rather than enforce new constraints.

More intrusive approaches (e.g. blocking duplicates or warning after selection) would likely require changes to action or membership logic and felt out of scope for me to take on in this first step.

Feedback welcome , especially on whether this direction makes sense.

<img width="957" height="579" alt="Image" src="https://github.com/user-attachments/assets/44b04962-42d0-498f-9a39-4a58933f444a" />

<img width="982" height="576" alt="Image" src="https://github.com/user-attachments/assets/99d8adb9-20c6-458a-ad86-ac2a5d64a684" />

PR #11949, opened 2026-03-01
Indicate applied and partially applied relations for multi-selection

aditya1010-max:
### Summary

This PR improves the discoverability of applied relations in the relations dropdown. While the dropdown already lists nearby relations, it is not currently clear which of those relations are applied to the current selection.

This change surfaces that information directly in the UI.

### Behavior

- Relations applied to **all** selected entities are labeled “Applied”
- Relations applied to **some** selected entities are labeled “Partially Applied”

These labels reflect existing state only and do not alter how relations are created or managed.

### Scope / Non-goals

- No changes to relation or membership logic
- Does not prevent duplicate relation members
- Does not introduce warnings or blocking behavior

More invasive approaches (such as enforcing constraints or modifying actions) are intentionally left out of scope for this PR.

This PR addresses one aspect of issue #10981.

Screenshot 2026-03-02 015423
Screenshot 2026-03-02 015439

### DIFF
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -354,9 +354,8 @@ export function uiSectionRawMembershipEditor(context) {
                 if (q && (value + ' ' + entity.id).toLowerCase().indexOf(q.toLowerCase()) === -1) return;
 
                 const count = relationCounts.get(entity.id) || 0;
-                const isCommon = count === selectedEntities.length;
-                const isPartial = count > 0 && count < selectedEntities.length;
-
+                const isCommon = selectedEntities.length > 0 && count === selectedEntities.length;
+                const isPartial = selectedEntities.length > 0 && count > 0 && count < selectedEntities.length;
 
                 result.push({
                     relation: entity,
--- a/modules/ui/sections/raw_membership_editor.js
+++ b/modules/ui/sections/raw_membership_editor.js
@@ -284,7 +284,7 @@ export function uiSectionRawMembershipEditor(context) {
         var graph = context.graph();
 
         //select entities
-        var selectedEntities = _entityIDs
+        const selectedEntities = _entityIDs
             .map(function(id) {
                 return graph.hasEntity(id);
             })
@@ -293,7 +293,7 @@ export function uiSectionRawMembershipEditor(context) {
             });
 
         // Map relationID -> number of selected entities in that relation
-        var relationCounts = new Map();
+        const relationCounts = new Map();
 
         selectedEntities.forEach(function(ent) {
             graph.parentRelations(ent).forEach(function (rel) {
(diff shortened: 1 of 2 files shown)


tordans:
@aditya1010-max before look into this…
- remember to use let/const for new code
- UI: add (a) space between word
- UI suggestion: I suggest to move the new feature to a pill/badge like look to the right of the line https://tailwindcss.com/plus/ui-blocks/application-ui/elements/badges
- please explain why this is a draft and what you plan to do before un-drafting
- please move the relevant parts of your comment on the issue here; we want the conversation about the implementation in one place (here); not scattered on issue and PR

UI:
I wonder if we want this red-green look here. Its more of a "warning", "all good" signal where this information is more of an FYI, right? We could make both just a gray FYI label… – Lets see what others say.

modules/ui/sections/raw_membership_editor.js
```diff
@@ -320,10 +352,20 @@ export function uiSectionRawMembershipEditor(context) {
...
+
+
                 result.push({
                     relation: entity,
                     value,
-                    display: baseDisplayLabel(entity)
+                    display: baseDisplayLabel(entity, flags),
+                    title: value,
+                    isCommon: flags.isCommon,
+                    isPartial: flags.isPartial
```
tordans: Dont quite see the value in having this extracted and then only used once.

```suggestion
                const count = relationCounts.get(entity.id) || 0;
result.push({
relation: entity,
value,
                    display: baseDisplayLabel(entity, flags),
                    title: value,
                    isCommon: count === selectedEntities.length,
                    isPartial: count > 0 && count < selectedEntities.length
```

aditya1010-max:
Thanks for the feedback.

I’ll update the new code accordingly and also fix the spacing issues.

I like the suggestion of moving this to a pill/badge-style element on the right. I’ll explore different implementations and update the UI based on that direction.

This PR is marked as draft to get early feedback and iterate on UI details. Before un-drafting, I plan to address the UI feedback and make sure the behavior is aligned with expectations.

Regarding the green/orange signaling, I agree it’s quite bold and doesn’t align well with the iD style. A neutral FYI-style label makes more sense here, and I’m happy to adjust based on what others think.

I’ll also take the other suggestions into consideration.

modules/ui/sections/raw_membership_editor.js
aditya1010-max: Thanks- Agreed. Since it’s only used once, I’ll simplify this.

aditya1010-max:
### SCREENSHOTS 

Screenshot 2026-03-03 165228

Screenshot 2026-03-03 165245

aditya1010-max:
I have addressed the feedback above and pushed the latest updates.

I would really appreciate any additional thoughts on the UI direction, especially around font size, spacing, background/border colors, or overall badge styling. If there are existing patterns in iD that this should align more closely with, I am happy to adjust it accordingly.

I will keep this PR in draft for a few more days while I review edge cases and look into adding tests where appropriate. If there are specific scenarios you’d like me to cover, please let me know.

I’ll follow up here once that’s complete.

aditya1010-max:
I’m marking this PR as ready for review. Happy to adjust anything if needed.

modules/ui/sections/raw_membership_editor.js
```diff
@@ -284,20 +283,52 @@ export function uiSectionRawMembershipEditor(context) {
...
                     .text(entityName);
+
+                if (flags.isCommon) {
+                    selection.append('span')
+                        .attr('class', 'relation-badge')
+                        .text('Already Applied');
+                } else if (flags.isPartial) {
+                    selection.append('span')
+                        .attr('class', 'relation-badge')
+                        .text('Partially Applied');
```
matkoniecz, line 328: this text is not translatable, it should be

modules/ui/sections/raw_membership_editor.js
```diff
@@ -284,20 +283,52 @@ export function uiSectionRawMembershipEditor(context) {
...
                     .append('span')
-                    .classed('has-colour', entity.tags.colour && isColorValid(entity.tags.colour))
+                    .classed('has-colour', entity.tags.colour && isColourValid(entity.tags.colour))
                     .style('border-color', entity.tags.colour)
                     .text(entityName);
+
+                if (flags.isCommon) {
+                    selection.append('span')
+                        .attr('class', 'relation-badge')
+                        .text('Already Applied');
```
matkoniecz, line 324: this text is not translatable, it should be

matkoniecz: @aditya1010-max have you seen this comment?

PR #11952, opened 2026-03-02
Convert `disclosure.js` to use `<details>` elements

Shrinks99:
HTML has native elements for this interaction, may as well use them!

## Changes

- Improves semantic markup by using `<details>` and `<summary>` elements in place of `h3` and `<a>`.
- Minor refactors:
   - Extracted `updateSummary()` to a helper function
   - Fixed double `_label()` call
   - Fade-out animation changes, set `<details>` to `open`

## Testing

While this PR was authored primarily with a coding agent, I've read over all of the code and tested the changes in all of the sidebars where `disclosure.js` is used.

Tested with VoiceOver on macOS.


css/80_app.css
```diff
@@ -314,6 +314,7 @@ table.tags, table.tags td, table.tags th {
 input.hide,
 textarea.hide,
 div.hide,
+details.hide,
```
tyrasd: this class is not used on the `<details>` elements, is it?

modules/ui/disclosure.js
```diff
@@ -25,59 +25,66 @@ export function uiDisclosure(context, key, expandedDefault) {
             _expanded = preference === null ? !!expandedDefault : (preference === 'true');
         }
 
-        var hideToggle = selection.selectAll('.hide-toggle-' + key)
+        var details = selection.selectAll('.disclosure-wrap-' + key)
```
tyrasd: can you use `const`/`let` for new and updated code, please?

modules/ui/disclosure.js
```diff
@@ -90,22 +97,27 @@ export function uiDisclosure(context, key, expandedDefault) {
...
-                .attr('aria-expanded', _expanded)
-                .attr('title', t(`icons.${_expanded ? 'collapse' : 'expand'}`));
-
-            hideToggle.selectAll('.hide-toggle-icon')
-                .attr('xlink:href', _expanded ? '#iD-icon-down'
-                    : (localizer.textDirection() === 'rtl') ? '#iD-icon-backward' : '#iD-icon-forward'
-                );
-
-            wrap
-                .call(uiToggle(_expanded));
```
tyrasd, line 104: at first glance, it seems the code below (lines 102 to 120) just replicated what `uiToggle` did before…

Something like this should do the same, reusing the existing toggle module:

```js
if (_expanded) {
    details.property('open', _expanded);
    contentWrap.call(uiToggle(true));
} else {
    contentWrap.call(uiToggle(false), () =>
        details.property('open', _expanded));
}
```

Yes, it needlessly sets the `hide` class, but it does also not hurt, does it? Alternatively, it might also worth refactoring the `uiToggle` module to natively work on `<details>` elements (i.e. if the `selection` is a details element, set the  `open` property instead of the `hide` class, and do the opacity transition on the non-summary child element instead.

tyrasd (commented):
In general, I think this is well worth modernizing. Thanks. 👍 

There are some (tiny) differences in the margins around the `<summary>` element, compared to how it looked before. I'd prefer if the change would remain the same as before, if possible.

In the screenshots below the new version is on the right. IMHO it looks a little bit too close to the closure content, and in general the "line height" of the `summary`/heading seems a little too narrow.





See also some inline comments regarding implementation details.

css/80_app.css
Shrinks99: 😒 There's always something....

This one actually is on me for not reviewing close enough.  TY.

Shrinks99:
Thanks for the review, good eye on the margin, was 5px off.  Fixed!

Also added `<details>` support for toggle.js by checking the parent tag type and setting `open` accordingly.

tyrasd (approved):

PR #11964, opened 2026-03-03
Fix buggy non-global string replacements

Kayd-06:
### Description
This Pull Request addresses a functional bug in [directional_combo.js](cci:7://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:0:0-0:0) and several string replacement inconsistencies across the codebase.

### Changes
- **directional_combo.js**: Fixed the [stripcolon](cci:1://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:26:8-28:9) function to use `.replaceAll(':', '')`. Previously, it only replaced the first colon, which resulted in invalid CSS classes for tag keys with multiple colons (e.g., `cycleway:left:oneway`).
- **keepRight.js**, **outdated_tags.js**, **nsi.js**: Updated several `.replace()` calls to `.replaceAll()` where global replacements were intended for consistency and robustness.

### Verification
- **Automated Tests**: Ran `npm run test:spec` and all 2244 tests passed.
- **Style Compliance**: All changes were applied surgically to maintain the project's strict original code style, indentation, and whitespace.

### DIFF
--- a/modules/ui/fields/directional_combo.js
+++ b/modules/ui/fields/directional_combo.js
@@ -25,7 +25,7 @@ export function uiFieldDirectionalCombo(field, context) {
     function directionalCombo(selection) {
 
         function stripcolon(s) {
-            return s.replace(':', '');
+            return s.replaceAll(':', '');
         }
 
 


bhousel:
> keepRight.js, outdated_tags.js, nsi.js: Updated several .replace() calls to .replaceAll() where global replacements were intended for consistency and robustness.

At least in these files the global replacements are not necessary.  You can definitely switch from `replace` to `replaceAll` if you feel like it, but it won't matter, as they are all replacing single things.

Kayd-06:
You're right that the replacements in those files are single occurrences, so the behavior is identical. I kept them as `.replaceAll()` for consistency across the codebase and to make the intent of a full replacement explicit. Happy to revert those if you'd prefer to keep the PR focused on just the [stripcolon](cci:1://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:26:8-28:9) fix in `directional_combo.js`.

matkoniecz:
@Kayd-06 which ones, if any, are buggy then?

How this bug can be reproduced?

Kayd-06:
The actual bug is in [stripcolon](cci:1://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:26:8-28:9) in [directional_combo.js](cci:7://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:0:0-0:0). 

The function uses `.replace(':', '')` which only removes the **first** colon. For tag keys with multiple colons like `cycleway:left:oneway`, the function produces `cyclewayleft:oneway` instead of the intended `cyclewayleftoneway`. 
This results in invalid CSS class names and HTML element IDs in the directional combo fields.

**To reproduce:**
1. Select a way with a directional tag that has multiple colons    (e.g., `cycleway:left:oneway`)
2. Open the tag editor - the directional combo field will generate CSS classes like `preset-directionalcombo-cyclewayleft:oneway`  instead of `preset-directionalcombo-cyclewayleftoneway`
3. The leftover colon in the class name is invalid and could  break CSS selectors targeting that element

The other changes ([keepRight.js](cci:7://file:///c:/Users/kunal/iD/modules/services/keepRight.js:0:0-0:0), [nsi.js](cci:7://file:///c:/Users/kunal/iD/modules/services/nsi.js:0:0-0:0), [outdated_tags.js](cci:7://file:///c:/Users/kunal/iD/modules/validations/outdated_tags.js:0:0-0:0)) are consistency improvements rather than bug fixes - as @bhousel noted, those replacements only match single occurrences so the behavior is identical. Happy to remove those from this PR if you'd prefer to keep it focused on the [stripcolon](cci:1://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:26:8-28:9) fix only.

tyrasd (changes requested):
please reduce the fix to the occurrences where this would actually lead to something _buggy_, e.g. the mentioned potentially invalid css classes.

Kayd-06:
@tyrasd , I have reduced the PR to only include the functional bug fix in [directional_combo.js](cci:7://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:0:0-0:0). 

I've reverted the other consistency only changes in [keepRight.js](cci:7://file:///c:/Users/kunal/iD/modules/services/keepRight.js:0:0-0:0), [nsi.js](cci:7://file:///c:/Users/kunal/iD/modules/services/nsi.js:0:0-0:0), and [outdated_tags.js](cci:7://file:///c:/Users/kunal/iD/modules/validations/outdated_tags.js:0:0-0:0). The PR now only addresses the [stripcolon](cci:1://file:///c:/Users/kunal/iD/modules/ui/fields/directional_combo.js:26:8-28:9) issue which was generating potentially invalid CSS classes.

PR #11965, opened 2026-03-03
added phone format for Moldova

Oni-DOS:
This format is the most common and recommended by ARCOM (National Regulatory Agencyin Communications of the Republic of Moldova). But i need to mention that there are also two other variants - `+373 12 123 456` and `+373 123 12345`, there are two different variants because Chisinau (the capital city) has the regional code 22 + 6 significant numbers and cellular operators have also allocated 2 numbers (6X and 7X) and also 6 significant numbers, which means `373 22 123 456` and `373 6X 123 456`, but other regions have allocated a 3 digits regional number in the range of 210 (Grigoriopol) to 557 (Slobozia) and 5 significat numbers, which means `373 210 12 345`. In that case the only valid variant will be `+37312345678` For details see: https://www.arcom.md/node/81

Is it possible to use the variant `"+373 12 123 456 / +373 123 12 345"` for this file?


matkoniecz:
Have you consulted this recommendation with local osm community?

Oni-DOS:
> Have you consulted this recommendation with local osm community?

Hi! Sure, we discussed that subject some months ago on our telegram group, no one was against that proposal.

data/phone_formats.json
```diff
@@ -42,6 +42,7 @@
   "kz": "+7 495 1234567",
   "lc": "+1-758-555-1234",
   "ma": "+212 5 12 34 56 78",
+  "md": "+37312345678",
```
tyrasd: should this not be one of the following:

```suggestion
  "md": "+373 12 123 456",
```

or

```json
  "md": "+373 123 12345",
```

or

```json
  "md": "+373 123 12 345",
```

as you mention in your pull request description?

🤔 

> Is it possible to use the variant "+373 12 123 456 / +373 123 12 345" for this file?

I would not include both: just use one of the two examples. I'm sure Moldovans will know that sometimes the area code can have two or three digits. That's fine.

Oni-DOS: > should this not be one of the following:
> 
> ```json
>   "md": "+373 12 123 456",
> ```
> 
> or
> 
> ```json
>   "md": "+373 123 12345",
> ```
> 
> or
> 
> ```json
>   "md": "+373 123 12 345",
> ```
> 
> as you mention in your pull request description?
> 
> 🤔
> 
> > Is it possible to use the variant "+373 12 123 456 / +373 123 12 345" for this file?
> 
> I would not include both: just use one of the two examples. I'm sure Moldovans will know that sometimes the area code can have two or three digits. That's fine.

Hi, you're right, actually there are more 2-digits area code phone numbers than 3-digits, so, yes we can use that as reference - `+373 12 123 456`

tyrasd (approved):

PR #11976, opened 2026-03-06
added Moldova adress format

Oni-DOS:
Hi, we discussed some time ago in Moldovan Telegram group also about this subject, there were no abstains or disagreements. All tags are widely used in OSM in Moldova and are based on [Moldova's Address System Guide](https://www.asp.gov.md/sites/default/files/documents/servicii/bunuri-imobile/Address_System_Guide_RO_actualizat.pdf), see pages 61-64.

Some clarifications: `unit` and `floor` are more used in urban area, because there are more office and retail buildings, `subdistrict` is very useful in rural area because there are a lot of villages and towns that have the same name and are located sometimes in the immediate vicinity, for example see - [town of Ghindești-Ghindești village ](https://www.openstreetmap.org/#map=15/47.85795/28.38520), [town of Mărculești-Mărculești village](https://www.openstreetmap.org/#map=15/47.87540/28.24430), [town of Dondușeni-Dondușeni village ](https://www.openstreetmap.org/#map=14/48.23216/27.61139)


data/address_formats.json
```diff
@@ -78,6 +78,14 @@
       ["housenumber", "street+place"],
       ["postcode", "city"]
     ]
+  },
+    {
+    "countryCodes": ["md"],
+    "format": [
+      ["street", "housenumber"],
+      ["postcode", "floor", "unit"],
+      ["city", "subdistrict", "district"]
```
tyrasd: from what I can see in the linked document from the Moldovan public administration and also on https://www.upu.int/UPU/media/upu/PostalEntitiesFiles/addressingUnit/mdaEn.pdf, the postcode seems to be always followed by the city/municipality name. Therefore, I think it is not ideal to include those here. Maybe put them after the housenumber, and subdistrict/district in a sparate line?

```suggestion
      ["street", "housenumber", "floor", "unit"],
      ["postcode", "city"],
      ["subdistrict", "district"]
```

That said, it seems like `addr:floor` and `addr:unit` are practically not present in OSM in Moldova (see taginfo: https://taginfo.geofabrik.de/europe:moldova/search?q=addr%3A) as of today. Do you have some real world examples for how those are used in practice? Maybe these are so rare that they might be better omitted in the form in iD (in order to discourage unnecessarily entering unnecessary values like _ground floor_ on every residential house's address)?

Oni-DOS: > discourage

Yeah, I think your proposal also looks good

tyrasd (approved):

PR #11977, opened 2026-03-06
Use primarily primary operation name in shortcut listing

### LINKED ISSUE #11557: Use uniform "Combine" name for Combine / Merge operation
Its shortcut is "C".
In the Help manual it is called "Combine (merge) selected features".
In the success / error messages it is called "Merge".

I would just call it "Combine" everywhere.

matkoniecz:


tyrasd (approved):

tyrasd (approved):
Thanks. I went even one step further: nowhere else in the UI we call the operation _combine_. The only benefit for including the "alternative" name of the operation would be to understand where the shortcut <kbd>c</kbd> is borrowed from (<kbd>m</kbd> was already taken by the _move_ operation), but IMO that's not reason enough for the added complexity here.

matkoniecz:
Hmm, I disagree and I think it was worth it for mnemonic purposes

Issue #10757, opened 2025-02-10
Adopt Rapid's sidewalk and curb mapping improvements to iD


tordans:
## Context

Rapid recently added a few improvements to make it easier to map sidewalks, crossings and curbs.

Here is a first list take from [the changelog](https://github.com/facebook/Rapid/blob/main/CHANGELOG.md):

> - https://github.com/facebook/Rapid/issues/1654, https://github.com/facebook/Rapid/issues/1671 Fix: Crash when adding curbs if there is no connected sidewalk - "this.tags is null"
> - https://github.com/facebook/Rapid/issues/1657 Fix: Don't suggest curb nodes when they are already mapped with kerb=no
> - https://github.com/facebook/Rapid/issues/1672 Fix: Unspecified curb should be kerb=yes, not kerb=unspecified
> - https://github.com/facebook/Rapid/issues/1460, https://github.com/facebook/Rapid/issues/1583, https://github.com/facebook/Rapid/issues/1619, https://github.com/facebook/Rapid/issues/1622 Add a new 'Curb Nodes' suggestion to suggest and add curb nodes
> - https://github.com/facebook/Rapid/issues/1597, https://github.com/facebook/Rapid/issues/1599 Fix: Don't suggest converting a crossing endpoint into a crossing node

## Goal

I wonder which of those improvements would be a good addition to iD. We should test them, collect feedback and specify which we would like to backport to iD.

---

There are other, related PRs in iD that we should keep in mind as well, especially https://github.com/openstreetmap/iD/pull/10524 and maybe https://github.com/openstreetmap/iD/pull/10256

---

FYI I added this as a suggestion for a GSoC Project (as a packages with #10743): https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2025/Project_ideas#iD_editor

tordans:
I looked into this a few weeks back to see if we could backport the feature to iD.
My current opinion is, that we first have to agree on the UX and the logic behind the sync (AKA which tags to add/remove/update).
The Rapid features is an auto-update that does not give feedback to the user on which tags are handled and how. I consider this a bit too much magic for iD. For iD we should look for a UX that gives more control to users. At the same time, the current System for Error/Warming/Info-Dialogues is quite manual and this feature will work best if some controlled automation is allowed. The use case I have in mind is: When I map a crossing way, then the node, then add tags to the node, I want those to be synced to the way automatically. But if I modify existing data, I will likely want to be informed in more detail on what updates take place.

[...]

PR #11982, opened 2026-03-07
Add validator for mismatched crossing tags between crossing nodes and crossing ways

### LINKED ISSUE #9574: Validate crossing vertex and ways tags
### URL

_No response_

1. Create a `highway=residential` way
2. Add a node on it and tag it as `highway=crossing + crossing=traffic_signals`
3. Create a way with 3 nodes, the middle one should connect to the previously created node
4. Tag that way with `highway=footway + footway=crossing + crossing=unmarked`
5. Note that the validator does not warn about this mistake.

_No response_

2.25.1

_No response_

aditya1010-max:
Fixes #9574

This PR adds a validator that detects mismatched `crossing=*` tags between a crossing node (`highway=crossing`) and its parent crossing way (`highway=footway` + `footway=crossing`).

If both the node and the crossing way have a `crossing=*` tag but the values differ, the validator raises a warning in the Issues panel.

This implementation is based on the discussion in #9574.
The Rapid implementation referenced in the issue, along with the discussions in #10757, served as helpful references while implementing this validator: facebook/Rapid#1260.

I tested this locally and confirmed that the validator correctly reports mismatched crossing types.

Happy to make adjustments if anything should be structured differently. I’d appreciate any feedback or suggestions.

@tyrasd @tordans , feedback welcome whenever you have time.

(Screenshot of the validator warning in the UI attached below)

### Before (No validation warning)

Reproducing the issue where the crossing node and crossing way have different `crossing=*` values, but no warning is shown.

before1

### After (Validator detects mismatch)

With the new validator, a warning is shown when the crossing node and crossing way have different `crossing=*` values.

after

### DIFF
--- a/modules/validations/crossing_vertex_tags.js
+++ b/modules/validations/crossing_vertex_tags.js
@@ -1,0 +1,61 @@
+import { validationIssue } from '../core/validation';
+import { t } from '../core/localizer';
+
+export function validationCrossingVertexTags() {
+
+    const type = 'crossing_vertex_tags';
+
+    function showReference(selection) {
+        selection
+            .append('div')
+            .attr('class', 'issue-reference')
+            .call(t.append('issues.crossing_vertex_tags.reference'));
+    }
+
+    const validation = function(entity, graph) {
+
+        if (entity.type !== 'node') return [];
+        if (entity.tags.highway !== 'crossing') return [];
+
+        const parentWays = graph.parentWays(entity);
+        const issues = [];
+
+        for (let way of parentWays) {
+
+            // check if the way is a crossing footway
+            if (way.tags.highway === 'footway' && way.tags.footway === 'crossing') {
+
+                const nodeCrossing = entity.tags.crossing;
+                const wayCrossing = way.tags.crossing;
+
+                if (!nodeCrossing || !wayCrossing) continue;
+
+                // detect mismatch
+                if (nodeCrossing !== wayCrossing) {
+
+                    issues.push(new validationIssue({
+                        type: type,
+                        subtype: 'crossing_type_mismatch',
+                        severity: 'warning',
+                        entityIds: [entity.id, way.id],
+
+                        title: t('issues.crossing_vertex_tags.title'),
+
+                        message: function() {
+                            return t.append('issues.crossing_vertex_tags.message.default');
+                        },
+
+                        reference: showReference
+                    }));
+
+                    break;   // avoid duplicate warnings
+                }
+            }
+        }
+
+        return issues;
+    };
+
+    validation.type = type;
+    return validation;
+}
--- a/modules/validations/crossing_vertex_tags.js
+++ b/modules/validations/crossing_vertex_tags.js
@@ -20,35 +20,39 @@ export function validationCrossingVertexTags() {
         const parentWays = graph.parentWays(entity);
         const issues = [];
 
+        // crossing way types based on iD presets
+        const crossingTypes = [
+            { highway: 'footway', key: 'footway' },
+            { highway: 'path', key: 'path' },
+            { highway: 'cycleway', key: 'cycleway' }
+        ];
+
         for (let way of parentWays) {
+            for (let crossingType of crossingTypes) {
 
-            // check if the way is a crossing footway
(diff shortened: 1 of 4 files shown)


modules/validations/crossing_vertex_tags.js
```diff
@@ -0,0 +1,61 @@
...
+        if (entity.type !== 'node') return [];
+        if (entity.tags.highway !== 'crossing') return [];
+
+        const parentWays = graph.parentWays(entity);
+        const issues = [];
+
+        for (let way of parentWays) {
+
+            // check if the way is a crossing footway
+            if (way.tags.highway === 'footway' && way.tags.footway === 'crossing') {
```
tordans: We have 3 types of crossing that are most relevant – see corresponding id-tagging-schema Presets for node and way
1. highway=footway+footway=crossing
2. highway=path+path=crossing
3. highway=cycleway+cycleway=crossing

tordans (changes requested):
I looked into this feature a few weeks ago but did not have time to write up my findings. I did so now in https://github.com/openstreetmap/iD/issues/10757#issuecomment-4019260997

There are open questions that we need to agree on first, IMO.

---

For this PR: Where is the part of the code that makes decisions on which tags to suggest for the "upgrade"? Is this done by the validation logic internally? So this would suggest migrating all tags?

modules/validations/crossing_vertex_tags.js
aditya1010-max: Thanks for pointing this out, the current logic is a bit too narrow just looking for highway=footway + footway=crossing. Supporting the `path=crossing` and `cycleway=crossing` cases makes sense as well since those correspond to the existing presets. I’ll update the validator logic to include those crossing types.

aditya1010-max:
> I looked into this feature a few weeks ago but did not have time to write up my findings. I did so now in [#10757 (comment)](https://github.com/openstreetmap/iD/issues/10757#issuecomment-4019260997)
> 
> There are open questions that we need to agree on first, IMO.
> 
> For this PR: Where is the part of the code that makes decisions on which tags to suggest for the "upgrade"? Is this done by the validation logic internally? So this would suggest migrating all tags?

No, I haven’t implemented the upgrade/suggestion logic yet. The current validator only detects the mismatch and reports it as a warning.

I’ll move this PR back to draft and follow up on the suggested changes step by step

aditya1010-max:
I’ve added a set of unit test cases to cover the tags and scenarios you suggested in your reference, so the expected behavior is clearer. There are likely more cases that can be added, but this should be a good starting point for implementing the validator logic

I’ll try implementing the validator logic along with the changes you suggested last week.

If there are any specific docs, patterns, or similar validators you’d recommend looking at as a reference, that would be really helpful.

I’m planning to start with a rough pseudocode approach based on these test cases and iterate from there.

modules/validations/crossing_vertex_tags.js
aditya1010-max: Added support for path and cycleway crossings alongside footway crossings, and generalized the check to handle all three types based on iD presets.
Please take a look when you have time.

matkoniecz:
CI reports failing tests

```
Error: TypeError: Cannot read properties of undefined (reading 'subtype')
 ❯ test/spec/validations/crossing_vertex_tags.js:200:26



Error: TypeError: Cannot read properties of undefined (reading 'subtype')
 ❯ test/spec/validations/crossing_vertex_tags.js:209:26



Error: TypeError: Cannot read properties of undefined (reading 'subtype')
 ❯ test/spec/validations/crossing_vertex_tags.js:219:26



Error: TypeError: Cannot read properties of undefined (reading 'subtype')
 ❯ test/spec/validations/crossing_vertex_tags.js:227:26


ERROR: "test:spec" exited with 1.
Error: Process completed with exit code 1.
```

aditya1010-max:
Thanks for pointing that out,  the CI failures are expected since I added the tests without completing the implementation yet.(this should have been done incrementally).
After working through this, I realized the scope is larger than I initially anticipated (normalization, multi-parent handling, conflict resolution, etc.), and I won’t be able to complete it to the required quality in the near term.

I’m happy to leave this PR as a draft in case it’s useful as a reference for others, or it can be closed if that’s preferred.

PR #11985, opened 2026-03-08
Fix URL validation rejecting URLs with accented characters

### LINKED ISSUE #11983: Wrong "invalid URL" warning
### URL

Image

Image

2.39.0-dev

Firefox

VaibhavSingh2006:
Fixes an issue where URLs containing accented characters
(e.g. "é" in "itinéraires") are incorrectly flagged as invalid.

The validator now encodes URLs using encodeURI before parsing
with the URL constructor, allowing proper validation of
internationalized URLs.

Example failing URL:
https://www.rando92.fr/randonner/itinéraires/

ScreenShots :

1). No issues now, with e'
Screenshot 2026-03-08 at 2 50 28 PM

2> can see the link with https://www.rando92.fr/randonner/itinéraires/
Screenshot 2026-03-08 at 2 50 36 PM


modules/validations/invalid_format.js
```diff
@@ -30,8 +30,8 @@ export function validationFormatting() {
         function isValidURL(url, strict = false) {
             try {
                 // First try strict WHATWG parsing
-                const link = new URL(url);
-                return link.href.includes(url);
+                const link = new URL(encodeURI(url));
+                return link.href.includes(encodeURI(url));
```
1ec5: Percent-escaped URIs are allowed and actually [preferred](https://wiki.openstreetmap.org/wiki/Key:website#Durability_and_machine-readability). This seems to work in your example because an already escaped URI would wind up getting overescaped and compared with another overescaped URI. But it also eliminates the ability to catch unescaped URIs that are unexpected.

1ec5: This test is [failing](https://github.com/openstreetmap/iD/actions/runs/22818217099/job/66187111019?pr=11985#step:7:461):

https://github.com/openstreetmap/iD/blob/895cd0654f185f4770bdfd3deb6148cef516c75e/test/spec/validations/invalid_format.js#L43

1ec5 (commented):
I can see how some mappers would expect to use an internationalized domain name or diacritics in the path without percent escaping, but we’d still need to warn about unescaped spaces and semicolons, which are more problematic. And if we do that, then a mixed encoded and unencoded URL could be problematic too. So we’d need to be very careful about all these cases.

modules/validations/invalid_format.js
k-yle: > But it also eliminates the ability to catch unescaped URIs that are unexpected.

Sorry, I labelled this as https://github.com/openstreetmap/iD/labels/new%20contributor%20opportunity because i thought it was clearly a bug. Please excuse my ignorance, but why are characters with diacritics considered unexpected? The wiki says that _some_ characters like `;` and ` ` should be percent-encoded, but there's no mention that this is a rule for all special characters.

It seems odd to me that we ban URLs like `https://teaomārama.school.nz`. If this really is intentional, maybe this validator needs an autofixer, to change the url to `https://xn--teaomrama-9bb.school.nz`? Otherwise it'll be near-impossible for mappers to figure out which character is banned and how to properly escape it.

1ec5: Diacritics are acceptable, especially since some browsers like Safari leave them unencoded when you copy-paste the full URL from the address bar. (Firefox escapes the full URL but not a substring of it.) My comment is indeed about those problematic characters like spaces and semicolons that can be misinterpreted. It still isn’t great that data consumers can’t depend on a consistent level of encoding to avoid over- or underencoding, but this isn’t something the validator can really solve on its own.

k-yle: okay cool, then how about we remove the check for `parsed.href === originalHref`, and only check if  `new URL()` throws an error?

```diff
-                const link = new URL(url);
-                return link.href.includes(url);
+                new URL(url); // eslint-disable-line no-new
+                return true;
```

With this change, all tests pass including a new test case for `https://exāmple.com;invalid-url;http://example.org/é嗯嗯`

VaibhavSingh2006: Thanks for the suggestion! That approach makes sense and keeps the validator simpler. 
I'm happy to update the implementation if you'd prefer that version - Please let me know ?

1ec5: I think the extra check was supposed to catch something like `https://example.com/foo;http://`, which is technically well-formed but probably an error. If the semicolon is intentional in that case, it should be escaped. We went through a bunch of iterations in #11438, so it’s possible that this turned out to be extraneous due to other checks.

VaibhavSingh2006: Thanks for the explanation! That makes sense — I can see how something like `https://example.com/foo;http://` could be unintended and why the extra check might have been added.

If those cases are already covered by other checks now, simplifying the validator to just rely on `new URL(url)` could be reasonable. But I'm happy to keep or adjust the extra check if you think it's still needed.

tyrasd (changes requested):
In 6adfc8c9aba62b440d4bc24bd51055fc2bc38085, I pushed two additional test cases which we want to not flag in the validator:

* accented character in URL
* URL with an [IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name) domain

Can you please take a look to get this covered?

VaibhavSingh2006:
Thanks for adding those tests! I'll take a look and update the validator to make sure URLs with accented characters and IDN domains are handled correctly.

VaibhavSingh2006:
Thanks! I've updated the validator and pushed a commit to support accented characters and internationalized domain names. All tests pass locally now.

modules/validations/invalid_format.js
```diff
@@ -28,17 +28,19 @@ export function validationFormatting() {
...
-         } catch {
-            if (strict) return false;
-            return /^https?:\/\/\S+$/i.test(url);
-         }
+         try {
+              const link = new URL(url);
+              const encoded = encodeURI(url);
+        return link.href.includes(url) ||
+               link.href.includes(encoded) ||
+               link.hostname === new URL(encodeURI(url)).hostname;
```
tyrasd, line 36: Can you explain your reasoning behind this condition? I mean in general the three different cases, but also specifically the last line:

```js
(new URL(url)).hostname === new URL(encodeURI(url)).hostname;
```

When would this ever be `false`?

tyrasd (changes requested):
I'm not convinced this really addresses the issue properly. See my question below. But also, we need some test cases of _invalid_ URLs we do actually want to flag.

VaibhavSingh2006:
Thanks for pointing that out. My intention with the hostname comparison was to ensure that URLs with accented characters or encoded paths still resolve to the same host when parsed, especially in cases where encodeURI changes the path but not the host. However, thinking about it more, you're right that in practice the hostname will almost always be identical and this check probably doesn’t add real validation. I’ll simplify the logic and rely on new URL(url) for validation and add clearer test cases for invalid URLs that we want to flag.

VaibhavSingh2006:
Thanks for the feedback! I’ve simplified the URL validation logic and added additional test cases for malformed URLs such as URLs containing spaces, incomplete URLs (https://), and malformed protocols (http//example.com). All validator tests pass locally.

tyrasd:
Closing PR as insufficient quality: the author was unfortunately unable to explain the logic and reasoning behind the proposed fix.

PR #11991, opened 2026-03-10
Add granular privacy controls for third-party icon sources

VaibhavSingh2006:
## Description

This PR adds more granular privacy controls for third-party icon sources.

Previously the editor had a single setting:

* **Show third-party icons**

This PR replaces that single option with separate controls so users can decide which sources they want to allow.

## Changes

* Added individual privacy options for:

  * Wikimedia icons
  * Facebook icons
* Updated the Privacy settings UI to display separate checkboxes.
* Preferences are stored using the existing `prefs` system.

## Why this change

Some users may want to allow icons from certain sources while blocking others for privacy reasons.
This change allows finer control over which third-party providers are enabled.

## Screenshots

Before:

* One checkbox for all third-party icons.

After:

* Separate checkboxes for Wikimedia and Facebook icons.

## Notes

This change only updates the UI preference system and keeps default behavior consistent with the previous implementation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,6 +27,12 @@ updates:
         patterns:
           - "@turf/*"
         applies-to: version-updates
+      eslint:
+        patterns:
+          - "@eslint/*"
+          - "eslint"
+          - "typescript-eslint"
+        applies-to: version-updates
     labels:
       - "chore-dependabot"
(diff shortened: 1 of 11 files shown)


VaibhavSingh2006:
Closing this PR because it duplicates #11992.

PR #11999, opened 2026-03-11
create a command bar UI

### LINKED ISSUE #8801: Add `command+k` action shortcut
`Command + K` is an emerging standard to access a search and action menu on websites and webapps.
See also https://css-tricks.com/kbar/

I would love to see it in iD as well.

For example, I tend to forget the shortcut for wireframe mode `W`. I would love to hit `cmd+k` and type `wire` to select "Style options: No fill (Wireframe view)".

For starters, all the actions that are documented in the help panel could be added to the command-list.

---

Update: Some more links and research
- https://kbar.vercel.app/
- https://github.com/osmberlin/osm-traffic-sign-tool/issues/6

k-yle:
Closes #8801 

When you press <kbd>Ctrl</kbd> <kbd>K</kbd>[^1], you get a 'command bar' where you can search for any iD setting, instead of clicking through the menus. This is most valuable for commands that have no keyboard shortcut.

<table>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

The design is based on the command bar in VSCode and the Chrome Dev Tools, which I use on a daily basis.

This is an early prototype to get feedback; it only has a few commands available. Adding more commands should probably be done in a follow-up PR, so that this one doesn't become too big.

[^1]: the exact shortcut needs discussion, see https://github.com/openstreetmap/iD/issues/10479#issuecomment-3316119174

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -6125,3 +6125,87 @@ textarea:focus:dir(rtl)::-webkit-resizer {
     width: 100px;
     color: var(--link-color);
 }
+
+.command-bar {
+    margin: 10% auto;
+    border: solid 1px var(--border-color);
+    border-radius: 4px;
+    width: 600px;
+    padding: 0;
+    font-size: 14px;
+    background: var(--bg-color);
+}
+
+.command-bar::backdrop {
+    background-color: #000a;
+}
+
+.command-bar footer {
+    display: flex;
+    justify-content: center;
+    padding: 8px 0 6px 0;
+    gap: 20px;
+    border-top: solid 1px var(--border-color);
+}
+
+.command-bar input {
+    width: 100%;
+    border: 0;
+    background: var(--bg-color-3);
+    padding: 8px 12px;
+    box-shadow: inset 0 0px 2px 0px var(--shadow-color);
+}
+.command-bar:focus-visible,
+.command-bar input:focus-visible {
+    outline: none;
+}
+
+.command-bar ul {
+    max-height: 450px;
+    overflow-y: auto;
+    padding: 4px;
+    margin: 0;
+    max-height: 50vh;
+}
+
+.command-bar li {
+    content-visibility: auto; /* major perf boost if supported */
+    list-style-type: none;
+    display: flex;
+    justify-content: space-between;
+    gap: 4px;
+    user-select: none;
+    cursor: pointer;
+    padding: 4px 8px;
+    border-radius: 4px;
+    align-items: center;
+    line-height: 1.2;
+    height: 30px;
+    position: relative;
+}
+.command-bar li:hover {
+    background-color: #0231;
+}
+.command-bar li.command-bar-selected {
+    background-color: var(--suggestion-highlight-color);
+    color: var(--dark-tooltip-text-color);
+}
+.command-bar li.command-bar-selected kbd {
+    color: initial;
+}
+li.command-bar-recently-used:not(:has(+ li.command-bar-recently-used))::after {
+    /* line after the last recently-used item */
+    content: "";
+    position: absolute;
+    left: 4px;
+    right: 4px;
+    bottom: 0;
+    height: 1px;
+    background: var(--border-color);
+}
+.command-bar-no-results {
+    text-align: center;
+    padding-bottom: 4px;
+    padding-bottom: 8px;
+    color: var(--passive-text-color);
+}
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1477,6 +1477,13 @@ en:
       label: "Browse files"
     no_geolocation:
       tooltip: Image without geolocation cannot be located on the map
+  command-bar:
+    key: K
+    placeholder: Run a command…
+    help:
+      navigate_list: to navigate
+      execute: to use
+      exit: to exit
   note:
     note: Note
     title: Edit note
@@ -2496,6 +2503,7 @@ en:
         save: "Save changes"
     tools:
       title: "Tools"
+      command-bar: Open the Command Bar
       info:
         title: "Information"
         all: "Toggle all information panels"
--- a/data/shortcuts.json
+++ b/data/shortcuts.json
@@ -362,6 +362,11 @@
           },
           {
             "modifiers": ["⌘"],
+            "shortcuts": ["command-bar.key"],
+            "text": "shortcuts.tools.command-bar"
+          },
+          {
+            "modifiers": ["⌘"],
             "shortcuts": ["info_panels.key"],
             "text": "shortcuts.tools.info.all"
           },
--- a/modules/behavior/paste.js
+++ b/modules/behavior/paste.js
@@ -6,6 +6,7 @@ import { operationPaste } from '../operations';
 import { t } from '../core/localizer';
 import { uiCmd } from '../ui/cmd';
 
+/** @param {iD.Context} context */
 // see also `operationPaste`
 export function behaviorPaste(context) {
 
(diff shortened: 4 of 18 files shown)


Dimitar5555:
In Firefox `Ctrl+K` selects the browser URL bar. I'm not sure if this is the default behavior.

1ec5:
It is still a thing on Firefox, though they’ve been forced to add alternative key bindings to accommodate Web applications that override <kbd>Ctrl</kbd><kbd>K</kbd> for a UI like this. The alternatives aren’t great but at least the situation is slightly better than with other shortcuts like <kbd>Ctrl</kbd><kbd>B</kbd>…

For what it’s worth, a common shortcut for this in native macOS applications is ⇧⌘O, as in “open quickly”.

tordans:
On the shortcut: Can someone look into how others handle this? Cmd+k is the one common shortcut I read about. https://kbar.vercel.app/ is all about it. All kind of docs that I use, use cmd+k. The question is, how do they handle this Firefox issue?…

1ec5:
Most don’t do anything special to accommodate Firefox. Chrome has enough market share that, for most of those applications, I’d probably just chalk it up to unfamiliarity with Firefox or a business decision that Firefox doesn’t matter. The shortcut originally came from Chrome, which was the first browser to unify the URL and search bars.

One application I know of has a reasonable compromise: Discourse Chat pops up a modal conversation chooser the first time you hit ⌘K, but the second ⌘K focuses the search bar (while leaving the chooser open). Similarly, ⌘L opens the Insert Link dialog, with the second ⌘L focusing the URL bar. The captured first keyboard shortcut trips me up sometimes but only momentarily. You can try it out on the [OpenHistoricalMap Forum](https://forum.openhistoricalmap.org/) if you’ve gone through the motions for trust level 1 or higher. I think this compromise could potentially work here, though we won’t be able to extend the same approach to other keyboard shortcuts like ⌘B that we treat as toggles.

tordans:
@k-yle thanks for starting this draft. Its funny, I was iterating on a Cursor Plan about this topic yesterday morning but did not post my results on the issue, yet. I like the registry approach you took here.

A few things from my plan that we might want to look into…

* We have quite a few toggle commands like "Toggle Mapillary". Ideally we can show the current status in the label as well like "Toggle Mapillary (Currently on)" or somthing.
* We have a few commands that are eather inactive until or need to call two command. Like "Mapillary: Only 360° Imagery" need to be inactive until "Mapillary" is on or it needs to do both, turn it on and also set the filter. I would prefer it did both.
* We have a few commands that are not well suited as a direct action but instead we could open the UI in the place where the action needs to be talken. Like "Mapillary: Filter by date" could open the side panel and scroll the UI element in view, maybe highlight it.
* I think we will want custom labels for each of the commands so we can write "Mapillary: Only 360° imagery" even though the form label is more like "Only 360° imagery".
* I thought about having the last used commands be the once we show when opened but I think this can just be session storage (not persistent). We could also not show any commands (before search), that would cleanup the UI more.
* FYI at https://tailwindcss.com/plus/ui-blocks/application-ui/navigation/command-palettes#component-69fd3cc3634e02f74717c1046d51aa76 there are a few design that we might want to look at for inspiration. However, your current design does fit nicely into the iD look.

aditya1010-max:
> Closes #8801
> 
> When you press Ctrl K[1](#user-content-fn-1-3be7a4e2025973523ea621a89fc393c3), you get a 'command bar' where you can search for any iD setting, instead of clicking through the menus. This is most valuable for commands that have no keyboard shortcut.


Thanks @k-yle  for starting this draft and for the command bar implementation!

I had actually spent some time experimenting with a small prototype for this feature earlier, after seeing the idea of kbar proposed for a previous GSoC project by @tordans . Your approach using the command registry integrates much more cleanly with the editor and seems like a solid foundation for extending the command palette.

Having this structure in place should make it much easier to extend the command palette and add more commands or improvements going forward.

For reference, here’s a quick prototype I was experimenting with earlier:

https://github.com/user-attachments/assets/58e49350-01bc-4b80-912d-ae2f8af526cd

Prototype commit (very rough and mostly exploratory):
https://github.com/aditya1010-max/iD/commit/793f344c1675fffc52eef09a22aa92847e1cdfdf

aditya1010-max:
> @k-yle thanks for starting this draft. Its funny, I was iterating on a Cursor Plan about this topic yesterday morning but did not post my results on the issue, yet. I like the registry approach you took here.
> 
> A few things from my plan that we might want to look into…


Thanks @tordans for sharing these ideas - the points about dynamic labels, command dependencies, UI-opening commands, and session-based command history are really interesting directions for evolving the command palette.

A few weeks ago I had also been experimenting with a small prototype around this feature, so it’s been exciting to see this draft PR. It seems like a solid foundation for implementing improvements you mentioned.

I was wondering if exploring some of these ideas further might make sense as a potential GSoC project. The directions you outlined already seem substantial enough to form a meaningful scope within the GSoC timeframe.

Since I’ve already been exploring this area, I’d be very interested in helping implement and refine these ideas over the summer. I was wondering if this might be a direction worth exploring for a GSoC project and I’d also be happy to adjust the scope based on feedback.

tordans:
> I was wondering if exploring some of these ideas further might make sense as a potential GSoC project. The directions you outlined already seem substantial enough to form a meaningful scope within the GSoC timeframe.

I don't think so given that this idea is not on the list for 2026 https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2026/Project_ideas

k-yle:
> (Re: keyboard shortcuts)

How about <kbd>Ctrl</kbd> <kbd>P</kbd> to match VSCode? The change of someone printing a broken layout onto a physical piece of paper is extremely unlikely, so i think we could override that shortcut without upsetting anyone?

> (Re: duplicate work) 

Sorry, I wasn't aware that anyone else was working on this 😭 I typically don't post comments like _"I'm working on it"_ for issues that only take an hour and only received 1 comment in the past year. But maybe I do that differently. 

For this feature, the tricky part is not the code or the css design, but rather the usability and the architecture, as Tobias mentioned in https://github.com/openstreetmap/iD/pull/11999#issuecomment-4044737907 - So we could easily close this PR, and use someone else's implementation, but the same usability and architecture questions will apply

k-yle:
Re: https://github.com/openstreetmap/iD/pull/11999#issuecomment-4044737907

> We have quite a few toggle commands like "Toggle Mapillary". Ideally we can show the current status in the label as well like "Toggle Mapillary (Currently on)" or somthing.

At first I wanted to static labels to be more performant - it's much faster to search a constant array of strings, vs calling 1000 functions to derive the label from the current state. But it probably doesn't matter. In this example, we could have a label that changes between `Enable Mapillary` and `Disable Mapillary` ?

>  I thought about having the last used commands be the once we show when opened but I think this can just be session storage (not persistent). We could also not show any commands (before search), that would cleanup the UI more.

I think showing recent items is pretty helpful? In VSCode I very frequently do <kbd>Ctrl</kbd> <kbd>P</kbd> -> <kbd>Enter</kbd> to repeat the last command

> FYI at https://tailwindcss.com/plus/ui-blocks/application-ui/navigation/command-palettes#component-69fd3cc3634e02f74717c1046d51aa76 there are a few design that we might want to look at for inspiration. However, your current design does fit nicely into the iD look.

Yeah, more design suggestions would be great. I only use the command bar in the Chrome Dev Tools and VSCode, no other apps. So the current design is based on my limited experience and my own personal biases. For example, 
- I personally don't like the slow animations in those demos, because imo features for powerusers should be instantaneous ?
- I think a simple text-only UI is better than a complex popup? github's command bar is the other extreme example - it's pretty chaotic, with heaps of layout shifts and contradictory messages [like _'no results'_ when there are results](https://github.com/user-attachments/assets/204851f6-8431-4305-9463-4a98307c38e9).

1ec5:
> The change of someone printing a broken layout onto a physical piece of paper is extremely unlikely, so i think we could override that shortcut without upsetting anyone?

The command list could include a shortcut to Field Papers, for those who really like mapping on dead trees. 😉

Issue #8174, opened 2020-11-06
Multi-select doesn't move point associated with line


jgscherber:
When select both a line and a node on that line, it'd be nice if a dragging action always moved the selected node, instead of any other nodes at the same location. See gif below for the "bad" behavior. The use case is largely around unhooking areas from roads. I don't want to move the road node just to get at the area node to move it.

![multiselect-point-line](https://user-images.githubusercontent.com/6207661/98384439-640d6900-2013-11eb-9fdc-6cad646e6b7f.gif)

(Although looking at the gif again, maybe the "good" behavior is: when a line is select, prioritize multi-selecting nodes on that line when they overlap with other nodes)

quincylvania:
I think this is the task behind #7984, but I'm going to leave this issue open instead since it seems more focused (and has a nice gif 😄).

AntMadeira:
In areas with landuse connected to highways (and other elements) it's a nightmare to try and detach them with iD due to this problem. Btw, this doesn't happen only on intersections, but in the entire shared lines between landuse and highways.
I'm just bumping this topic hoping it can get some attention.

Regards.

1ec5:
This issue is about what happens when part of all of the landuse area is already disconnected from the adjoining highways. If all you need to do is detach the landuse area from highways, select the area and use the Disconnect command. (You may need to zoom to fit the area if it’s too big.)

For more limited disconnections, I use this workaround:

1. Drag the road out of the way.
2. Select the previously obscured node.
3. Undo the move.
4. Use the keyboard shortcut (<kbd>M</kbd> in English) to start moving the multiple selection without interference from the road.

AntMadeira:
I'm aware of that process, but when you have to do this dozens of times in just one changeset it's a nightmare...

What I end doing many times is detach the entire landuse area and reduce it with Shiff + "-", but that's not an ideal solution, since it only works if the landuse is not connected to other elements.
Using Shift+select node+M also works sometimes, but not always.
What should happen is being able to move just the nodes from the selected element.

All this could be avoided if there was a warning when a mapper attaches landuse to highways...

Issue #2206, opened 2014-04-24
Disconnect point should select point wisely


finetjul:
When "disconnecting" a point connected to areas A and B. The newly selected/current point is undefined for the user. It might be a point of A or a point of B. It seems to depend on the way those 2 areas have been connected (a point of A got connected to a point of B or a point of B got connected to a point of A). There is no way for the user to know which point (of A or B) will be selected/current once disconnected.

I would suggest instead to select the point of the area that has been the last one to be selected. For example, a point P is connected to areas A and B :
- select area B
- disconnect point P
- move point P
  -> it moves the point of area B

slhh:
@finetjul 
> I would suggest instead to select the point of the area that has been the last one to be selected.
I'd agree, but the example seems not to to be exact. I think it should be:
A node P is connected to areas A and B :
1. select area B
2. select node P
3. disconnect
4. move the selected node
    -> it moves the node of area B which is no-longer P, but the new node generated by disconnect.

We can do this based on #3634, which is already part of my pull request #3631, by #3635 .
Area B would become the related parent in step 2 due to #3634, and this would be used in step 3 due to #3635.
This would also make disconnecting multiple nodes quite easy by continuing with hitting pgDn (Next vertex) and "D" (Disconnect) alternatingly.
Main usecase would be landuses glued to highways, which are quite common in some regions, but very nasty to handle with current iD.

@bhousel 

> I also had the thought of separating the new points away from each other by a few pixels (because you really never want the disconnected points on top of each other).

In addition to giving the user control which feature is disconnected as suggested above, it makes sense to move the new node a few pixels, but not the original one.

1ec5:
#6904 proposes an affordance for the specific case where a way is selected before disconnecting the point.

PR #12000, opened 2026-03-11
Fix Node selection at stacked vertices

### LINKED ISSUE #11314: Selecting a node in a cluster of nodes should select the node of the selected line
### Description

We have a good pattern to disconnect a node where multiple nodes connect: First select the line, then the node => the disconnect operation will apply to the node of the selected line.

I think the same logic should apply to moving a disconnected node afterwards.
But right now, it always(?) selects the "other" nodes, which I want to keep as is.

The video shows my issue. Selecting the node does not take into account the way that I selected before.

The video also shows, that apparently the logic changes when I first move some other node of the same way…

### Screenshots

https://github.com/user-attachments/assets/98ef1221-0ba5-4210-aa71-f685664edbbd

tordans:
> [!NOTE]
> _Update 2026-04-30:_ After Kyles Code Review I did a full cleanup and another round of fixes with Opus 4.8 High an tested the results manually.

This is a take at fixing https://github.com/openstreetmap/iD/issues/11314.
There are a couple of more tickets that sound similar: https://github.com/openstreetmap/iD/issues/8174, https://github.com/openstreetmap/iD/issues/2206 and other already closed issues.

I find this issue annoying again and again and using the changes here feel like a good improvement.

---

### Problem

When two vertices stack at the exact same location and **both** are part of the current selection, the touch layer put the wrong one on top, so dragging grabbed the wrong feature.

Repro: select line 1 (which shares a middle node with line 2), shift-click that node to disconnect line 1, then immediately drag the node. Line 1 stays highlighted, but line 2 moves instead.

### Root cause

Disconnecting gives the selected way a brand-new vertex stacked on top of the original shared node. Both ended up in `_currSelected` (the way's child vertices *and* the co-selected junction node), so `sortTouchByLocSelected`'s `selected vs. not-selected` comparator returned a tie and left the wrong node's hit-target on top.

### Fix

`sortTouchByLocSelected` now ranks stacked touch targets in three tiers instead of two:

1. vertex of a **selected way** (highest)
2. a merely co-selected node
3. everything else

This guarantees the selected way's endpoint stays on top and grabbable. The fix is deterministic and requires no redraw/timing changes.

### DIFF
--- a/modules/svg/vertices.js
+++ b/modules/svg/vertices.js
@@ -199,11 +199,40 @@ export function svgVertices(projection, context) {
     }
 
 
+    // Node ids of every selected way, so we can keep their vertices on top when stacked.
+    function selectedWayNodeIDs(graph) {
+        var ids = {};
+        context.selectedIDs().forEach(function(id) {
+            var entity = graph.hasEntity(id);
+            if (entity && entity.type === 'way') {
+                entity.nodes.forEach(function(nodeID) { ids[nodeID] = true; });
+            }
+        });
+        return ids;
+    }
+
+    // Sort so the selected way's vertex is last (drawn on top), so it receives the pointer
+    // when targets stack at the same location. Z-order only matters where circles overlap.
+    // A vertex of a selected way ranks above a merely co-selected node, which ranks above the
+    // rest - this keeps the right endpoint grabbable after e.g. disconnecting it. #11314
+    function sortTouchByLocSelected(items, wayNodeIDs) {
+        function rank(item) {
+            var id = item.properties.entity.id;
+            if (id in wayNodeIDs) return 2;
+            if (id in _currSelected) return 1;
+            return 0;
+        }
+        items.sort(function(a, b) {
+            return rank(a) - rank(b);
+        });
+    }
+
     function drawTargets(selection, graph, entities, filter) {
         var targetClass = context.getDebug('target') ? 'pink ' : 'nocolor ';
         var nopeClass = context.getDebug('target') ? 'red ' : 'nocolor ';
         var getTransform = svgPointTransform(projection).geojson;
         var activeID = context.activeID();
+        var wayNodeIDs = selectedWayNodeIDs(graph);
         var data = { targets: [], nopes: [] };
 
         entities.forEach(function(node) {
@@ -234,6 +263,10 @@ export function svgVertices(projection, context) {
             }
         });
 
+        // At stacked locations, put the selected way's vertex last so it ends up on top. #11314
+        sortTouchByLocSelected(data.targets, wayNodeIDs);
+        sortTouchByLocSelected(data.nopes, wayNodeIDs);
+
         // Targets allow hover and vertex snapping
         var targets = selection.selectAll('.vertex.target-allowed')
             .filter(function(d) { return filter(d.properties.entity); })
@@ -251,6 +284,7 @@ export function svgVertices(projection, context) {
                   || radiuses.shadow[3];
             })
             .merge(targets)
+            .order()  // keep DOM order in sync with the sorted data (selected vertex on top) #11314
             .attr('class', function(d) {
                 return 'node vertex target target-allowed '
                 + targetClass + d.id;
@@ -272,6 +306,7 @@ export function svgVertices(projection, context) {
             .append('circle')
             .attr('r', function(d) { return (_radii[d.properties.entity.id] || radiuses.shadow[3]); })
             .merge(nopes)
+            .order()  // keep DOM order in sync with the sorted data (selected vertex on top) #11314
             .attr('class', function(d) { return 'node vertex target target-nope ' + nopeClass + d.id; })
             .attr('transform', getTransform);
     }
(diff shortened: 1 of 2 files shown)


modules/svg/vertices.js
```diff
@@ -199,6 +210,32 @@ export function svgVertices(projection, context) {
...
+    function sortTouchByLocSelected(items) {
+        const sel = _currSelected || {};
+        items.sort((a, b) => {
+            const locA = a.properties.entity.loc;
+            const locB = b.properties.entity.loc;
+            if (locA[0] !== locB[0] || locA[1] !== locB[1]) return 0;
+            const aIn = a.properties.entity.id in sel;
+            const bIn = b.properties.entity.id in sel;
+            if (aIn !== bIn) return aIn ? 1 : -1;
+            return 0;
```
k-yle, line 227: if the `x`/`y` position is different, does the `z` order matter? 

if no, we could simplify this to

```suggestion
            return (a.properties.entity.id in sel) - (b.properties.entity.id in sel)
```

modules/svg/vertices.js
```diff
@@ -96,7 +96,18 @@ export function svgVertices(projection, context) {
...
+        /** @param {osmEntity} a
+         *  @param {osmEntity} b
+         *  @returns {number} */
+        vertices.sort((a, b) => {
+            const yOrder = sortY(a, b);
+            if (yOrder !== 0) return yOrder;
+            const aInSelected = a.id in (sets.selected || {});
+            const bInSelected = b.id in (sets.selected || {});
+            if (aInSelected !== bInSelected) return (aInSelected ? 1 : 0) - (bInSelected ? 1 : 0);
+            return a.id < b.id ? -1 : a.id > b.id ? 1 : 0;
```
k-yle, line 109:
```suggestion
            return a.id - b.id;
```

modules/svg/vertices.js
```diff
@@ -96,7 +96,18 @@ export function svgVertices(projection, context) {
             });
         }
 
-        vertices.sort(sortY);
+        // Same screen position: draw selected-way vertex on top so it receives the click (issue #11314).
+        /** @param {osmEntity} a
+         *  @param {osmEntity} b
+         *  @returns {number} */
+        vertices.sort((a, b) => {
```
k-yle, line 103: sorry if this is a dumb question, but why do we need to change the data layer? i think only the touch layer needs to change ?

modules/svg/vertices.js
```diff
@@ -234,46 +271,35 @@ export function svgVertices(projection, context) {
             }
         });
 
-        // Targets allow hover and vertex snapping
-        var targets = selection.selectAll('.vertex.target-allowed')
+        // Same stacked location: draw selected-way vertex on top so it receives the click (issue #11314).
+        sortTouchByLocSelected(data.targets);
+        sortTouchByLocSelected(data.nopes);
```
k-yle, line 276:
```suggestion
```
i think this is redundant? because we sort `combined` a few lines below

modules/svg/vertices.js
```diff
@@ -234,46 +271,35 @@ export function svgVertices(projection, context) {
...
-                  || radiuses.shadow[3];
+                return d.nope ? (_radii[d.properties.entity.id] || radiuses.shadow[3]) : (_radii[d.id] || radiuses.shadow[3]);
             })
-            .merge(targets)
+            .merge(touchItems)
+            .order()  // DOM order must match data order so selected-way vertex is on top (issue #11314)
             .attr('class', function(d) {
-                return 'node vertex target target-allowed '
-                + targetClass + d.id;
+                return 'node vertex target ' + (d.nope ? 'target-nope ' + nopeClass : 'target-allowed ' + targetClass) + d.id;
```
k-yle, line 300: what exactly is `target-nope` in the context of verticies? i'm just wondering why we need to merge the two layers

i had a quick look, and it seems to still work if we keep only these lines: 0831ecd1472f042ddc3bc80b696ec5e1514b4b37 but maybe there's some edge cases that i've missed?

k-yle (commented):
nice, this issue has annoyed me for ages !

modules/svg/vertices.js
tordans: I had Opus 4.8 High review the changes and this part is now gone.

modules/svg/vertices.js
tordans: I had Opus 4.8 High review the changes and this part is now gone.

modules/svg/vertices.js
tordans: This was still needed but quite a bit of duplication is now gone.

modules/svg/vertices.js
tordans: I had Opus 4.8 High review the changes and this part is now gone.

tordans:
@k-yle thanks for looking into this.
I rebased this PR and had a current model review the changes. I found more issues with the following case, that are now fixed:

1. Line A + Line B connect in the middle, Line C (not relevant) connects to this middle
2. Select Line A
3. Shift Right Click the Center Node (where all connect) => Disconnect => Line A stays selected visually
4. Move the center node
    Error: Line B moved
    Fix: Line A moves
    
> Test setup:
> Bildschirmfoto 2026-05-30 um 20 11 07

modules/svg/vertices.js
k-yle: has been resolved

Issue #9679, opened 2023-06-01
Fields should indicate that related conditional tags are present


1ec5:
A proper `opening_hours` field has eluded us for years (#974), not to mention a UI for conditional restrictions that would presumably be based on it (#2906). Meanwhile, there are plenty of features that have both a key and the conditional variant of that key set simultaneously, for example, `maxweight=*` and `maxweight:conditional=*`. At the very least, an unobtrusive label could appear at the bottom of a field to indicate that conditional tagging is present and can be edited using the raw tag editor.

Shyam-123pandey:
I’ve been looking into long-standing discussions around opening_hours (#974) and conditional tagging (#2906), especially for turn restrictions and other keys that support conditional variants (e.g., maxweight and maxweight:conditional).

Currently, if a feature has both a base tag and a conditional tag, the UI does not clearly indicate that conditional tagging is present unless the user opens the raw tag editor. This can make it easy to overlook or accidentally modify important conditional data.

As a potential first step (before building a full conditional UI), I’m wondering whether it would make sense to display a small, unobtrusive indicator beneath a field when a corresponding `*:conditional` tag exists. For example:

- maxweight=*  
- maxweight:conditional=*

The indicator could inform users that conditional tagging is present and editable in the raw tag editor.

If this direction seems reasonable, I’d be happy to explore a basic high-level design and proposal for how conditional awareness could be incrementally introduced into iD.

tordans:
@Shyam-123pandey, we need to define the UX of this first.
- What do we show, where exactly, and how (style)?
- What is the text we show?
- More generally: what is the message to users?

The fields in iD are meant to guide new users to contribute. So just writing "there are more tags" does not help…

I wonder if there actually is something we can do here for inexperienced users. Maybe this is and has to be a power user feature, and we have to accept we cannot properly guide users? Or maybe we can at least link to good wiki articles (which) per tag and write something helpful with reference to those…

Also: Which of the schema builder field types do we need to consider, and how do we display this info for all of them?

Related: Exposing a check_date:* https://github.com/openstreetmap/iD/issues/8884 and source:*, description:* – more or less all nested support tags.

Shyam-123pandey:
Okay Sir After thinking will respond

Shyam-123pandey:
Thank you — that’s very helpful feedback
I see two possible directions and would appreciate feedback on which (if any) aligns better with iD’s philosophy:

1) Passive awareness / protection  
[...]

Issue #9453, opened 2023-01-07
Warn users when editing a tag that has a source


Discostu36:
### Description

The validator should warn users when they edited the value of a tag that has a source tag associated with it. 

Examples:

````
highway=residential
maxspeed=30
source:maxpeed=DE:zone30
````
The user edits `maxspeed `to `50`.
````
place=city
population=50123
source:popolation=INSEE 2022
````
The user changes `population `to `50342`. 

````
highway=residential
width=6
source:width=ARCore
````
The user changes `width `to `5.5`

In each of those cases, a warning message should be shown by the validator, because it is very likely, that the source has changed. In some cases like the first example, it might also call attention to a tagging mistake. A first draft of how such a message could be, see below.

### Screenshots

![grafik](https://user-images.githubusercontent.com/30999326/211171414-ad7fabff-ae7f-4c06-81d9-b8fc9d277d82.png)

tordans:
I agree, that this is an issue in areas that are highly mapped. I have fixed miss-matches regularly when new users change the `maxspeed` but – understandably – don't see or understand the source* fields under "tags".

---

On the topic of `maxspeed`, there is the issue of the badly specified source fields. iD would either have to only support one field… or all… or wait for a cleanup of the tagging schema. https://github.com/westnordost/osm-legal-default-speeds/issues/4 shows a few of those tags and the linked video (last comment) goes into more details.

---

A good fist step to make those fields more visible would IMO be to add them as optional `moreTags` field via https://github.com/openstreetmap/id-tagging-schema. This would show them when present, but only then. `source:population` already exists ins this way https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/source/population.json. In this case it is visible as soon as population is selected (see `prerequisiteTag`). Which IMO makes sense here but I am unsure about the `maxspeed` source field; we would have to make sure the inputs are well defined for new users to pick from.

---

Broader picture: I see a pattern here of "supporting tags" that would ideally be displayed along side the main tag. Maybe some time iD can have a UI that shows them together. Right now I see
- `source:maxspeed` tag
- `check_date:maxspeed` tag – see references in https://github.com/openstreetmap/iD/issues?q=is%3Aissue+is%3Aopen+check_date
- `description:maxspeed`, `fixme:maxspeed` or `note:maxspeed` tags (text tags that are referenced to one specific tag)

tyrasd:
[...]

Issue #8884, opened 2021-10-03
Please support check_date:opening_hours and check_date:collection_times


dafadllyn:
Please add a field for `check_date:opening_hours` to presets with `opening_hours` and `check_date:collection_times` to presets with `collection_times`.

It is helpful to know when the opening hours or collection times were last checked - both for end users (to warn them that the opening hours or collection times may no longer be up to date) and for mappers (objects whose opening hours or collection times have not been checked recently can be resurveyed).

tordans:
I just closed https://github.com/openstreetmap/iD/issues/8548 as a duplicate but for added visibility there are two thought that we should track here:

- SC checks different values for a date but always tags one format, see https://github.com/openstreetmap/iD/issues/8548#issuecomment-1007845621
- There is a link which check_date values SC sets in there as well
- The issue asks the question about how to nudge users to update the check date. Or maybe it should update automatically?

tordans:
For reference: https://github.com/openstreetmap/id-tagging-schema/issues/1469#issuecomment-2694949309 links to https://github.com/OpenHistoricalMap/iD/pull/217 which shows a dedicated but integrated UI to handle source fields per Tag. This is the approach I had in mind for adding a check_date UI to iD as well.

Issue #8548, opened 2021-06-19
Ask for ``smoothness:check_date`` when adding/changing the smoothness value of a ``highway=*`` way


Dimitar5555:
It is quite hard to check when the smoothness key of a road or a path has been changed which makes it hard to know whether this data is recent or not. If it gets implemented it would be very beneficial to other mappers and possibly data users.

bryceco:
I wish the server would provide the per-tag last change date as metadata.

mmd-osm:
Tag `smoothness:check_date` appears to be quite niche: https://taginfo.openstreetmap.org/keys/smoothness%3Acheck_date#chronology

Dimitar5555:
There is ``check_date:smoothness`` which is also quite niche: https://taginfo.openstreetmap.org/keys/check_date%3Asmoothness#chronology. It looks like ``smoothness:date``(https://taginfo.openstreetmap.org/keys/smoothness%3Adate#chronology) is more widespread and could be used for that purpose.

tordans:
StreetComplete is working on a feature that will make it a lot easier to add smoothness values. It will also be used by the (relatively new) re-check-feature of SC. 

> Will the app show the quest for segments with existing smoothness values as well?
> (…) (yes,) if the segment hasn't been touched since 4 years or the `check_date:smoothness` key has a value that is older than 4 years

More: https://github.com/streetcomplete/StreetComplete/pull/3617#issuecomment-1004901246

We can expect this tag to be used more in the future. And I would love to see other editors to start using it as well. It's the best option we have right now to re-validate the data and make sure it's high quality.

westnordost:
StreetComplete looks for `smoothness:check_date`, `check_date:smoothness`, `smoothness:lastcheck`, `lastcheck:smoothness`, `smoothness:last_checked` and `last_checked:smoothness` but it always tags `check_date:smoothness` (and removes the others when updating the date).

The pattern for all re-checkable properties is the same for all keys in StreetComplete: `check_date:<key>`. This pattern is well-established: https://taginfo.openstreetmap.org/search?q=check_date%3A

First time I hear about `smoothness:date` thoug. It's not documented. Taginfo suggests it is used in Kasachstan a lot. Anyway, I'll add it to the list of keys to check.

tordans:
(Sorry, missclicked just now…)

[...]

PR #12005, opened 2026-03-12
Show and edit related subtags in the inspector (`check_date`, `note`/`description`, `source`, `conditional`, numeric, other)

tordans:
This is a draft to experiment with a UI that will allow us to expose related tags next to their field.
I [did not review the code](https://github.com/openstreetmap/iD/pull/12005#issuecomment-4060696657) in details, yet. **This PR is meant to be used to see if this is a good idea in terms of UX, UI and in general.**

## Related issues

Directly
- https://github.com/openstreetmap/iD/issues/8884
- https://github.com/openstreetmap/iD/issues/9679
- https://github.com/ideditor/schema-builder/issues/180#issuecomment-4047292267

Indirectly
- https://github.com/openstreetmap/iD/issues/8548
- https://github.com/openstreetmap/iD/issues/9453
- https://github.com/openstreetmap/id-tagging-schema/issues/1202

## Screenshots

_TODOs:_
* The icons are placeholder; I did not look for good icons, yet
* The Box of related tags is not properly styled, yet

<details><summary>Tags for testing</summary>
<p>

```
check_date:surface=2026-03-12
highway=motorway
maxspeed:conditional=60 @ wet
note:surface=bar
source:maxspeed=test
source:width=w foo
surface=stone
width=10
width:source=w bar
panoramax=123
panoramax:3=345
panoramax:99=nonono
panoramax:2=234
maxspeed=100

```

</p>
</details> 

**`source:tagname`**
* > Bildschirmfoto 2026-03-12 um 17 14 19
* > Bildschirmfoto 2026-03-12 um 17 14 23

**`tagname:1`, `tagname:2`** 
* > Bildschirmfoto 2026-03-12 um 17 14 10
* > Bildschirmfoto 2026-03-12 um 17 14 14

**`check_date:tagname`**
* > Bildschirmfoto 2026-03-12 um 17 13 57
* > Bildschirmfoto 2026-03-12 um 17 14 01

**`tagname:conditional`**
* > Bildschirmfoto 2026-03-12 um 17 13 44
* > Bildschirmfoto 2026-03-12 um 17 13 52

**Other related tags**
* > Bildschirmfoto 2026-03-12 um 17 32 13
* > Bildschirmfoto 2026-03-12 um 17 32 21
   
   This is buggy still, but the idea is to expose the "cycleway:left:surface" tag; the code about this is very rough but the idea would be to allow to give visibility to the tags in those more complex schema like the `parking:*` schema.

The general idea is that we always use the label of the field to show the indicator-buttons. For preset types that handle multiple key, the indicator would search for related tags for all those keys and just list them all.

The buttons are only there, when there are tags, so their presence in itself is an indicator.

They render in tree preset-fields: text, check-date-text, textarea which I think will be enough for the cases we will likely see.

---

What I like about this is, that it will allow use to give more visibility to some related tags without the need to do a bigger update like https://github.com/openstreetmap/id-tagging-schema/issues/1202

Possible simpler version: We could boil this down to something simpler first, which only shows indicator icons and not button in the label bar. The indicator icons could then have a tooltip that just informs users that there are related tags.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1478,6 +1478,64 @@ button.preset-reset .label.flash-bg {
     fill: var(--text-color);
 }
 
+/* Subtag category icons – same as remove/revert buttons in label row */
+.field-label .subtag-icons {
+    display: flex;
+    flex-flow: row nowrap;
+    flex: 0 0 auto;
+    align-items: stretch;
+}
+.field-label .subtag-icon {
+    display: inline-block;
+}
+.field-label .subtag-icon .icon {
+    width: 14px;
+    height: 14px;
+}
+.field-label .subtag-icon.active,
+.field-label .tag-reference-button.active {
+    background-color: var(--active-bg-color);
+}
+.field-label .subtag-icon.active .icon,
+.field-label .tag-reference-button.active .icon {
+    opacity: 1;
+}
+
+/* Wrapper around all expanded subtag fields (grouping + rounded border) */
+.subtag-expanded-outer {
+    margin-top: 0;
+    margin-left: 0;
+    margin-right: 0;
+    margin-bottom: 10px;
+    padding: 8px 10px 10px;
+    border: 1px solid var(--border-color);
+    border-radius: 6px;
+    background: var(--fill-secondary);
+}
+.subtag-expanded-outer:last-child {
+    margin-bottom: 0;
+}
+/* One section per expanded category inside the outer wrapper */
+.subtag-expanded-wrap {
+    margin-top: 8px;
+}
+.subtag-expanded-wrap:first-child {
+    margin-top: 0;
+}
+/* Each row is a full preset field (wrap-form-field) – same as sidebar */
+.subtag-expanded-row {
+    margin-top: 6px;
+}
+.subtag-expanded-row:first-child {
+    margin-top: 0;
+}
+.subtag-expanded-outer .form-field {
+    margin-bottom: 0;
+}
+.subtag-expanded-outer .wrap-form-field:last-child .form-field {
+    margin-bottom: 0;
+}
+
 .field-label .modified-icon,
 .field-label .remove-icon,
 .field-label .remove-icon-multilingual {
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -770,6 +770,13 @@ en:
     add_fields: "Add field:"
     lock:
       suggestion: 'The "{label}" field is locked because there is a Wikidata tag. You can delete it or edit the tags in the "Tags" section.'
+    subtag:
+      check_date: Additional related tag – when this was last verified (so you can check it more easily).
+      note_desc: Additional related tags – notes or descriptions (so you can check them more easily).
+      source: Additional related tag – source of this value (so you can check it more easily).
+      conditional: Additional related tag – conditional value, e.g. weather or time (so you can check it more easily).
+      numeric: Tags with numeric postfix (e.g. 1, 2, 3) for this key (so you can check them more easily).
+      other: Other additional related tags (so you can check them more easily).
     display_name_addr: "{housenumber} {streetOrPlace}"
     display_name_addr_with_unit: "{unit}, {housenumber} {streetOrPlace}"
     display_name:
--- a/modules/ui/entity_editor.js
+++ b/modules/ui/entity_editor.js
@@ -29,6 +29,12 @@ export function uiEntityEditor(context) {
     var _newFeature;
 
     var _sections;
+    /**
+     * D3 selection of the inspector body container. Kept so the preset fields
+     * section can be re-rendered when a subtag category is expanded.
+     * @type {d3.Selection}
+     */
+    let _bodySelection;
 
     function entityEditor(selection) {
 
@@ -88,19 +94,28 @@ export function uiEntityEditor(context) {
             .merge(bodyEnter);
 
         if (!_sections) {
+            const presetFieldsSection = uiSectionPresetFields(context)
+                .on('change', changeTags)
+                .on('revert', revertTags)
+                .on('expandSubtag', function() {
+                    if (_bodySelection) {
+                        _bodySelection.call(presetFieldsSection.render);
+                    }
+                });
             _sections = [
                 uiSectionSelectionList(context),
                 uiSectionFeatureType(context).on('choose', function(presets) {
                     dispatch.call('choose', this, presets);
                 }),
                 uiSectionEntityIssues(context),
-                uiSectionPresetFields(context).on('change', changeTags).on('revert', revertTags),
+                presetFieldsSection,
                 uiSectionRawTagEditor('raw-tag-editor', context).on('change', changeTags),
                 uiSectionRawMemberEditor(context),
                 uiSectionRawMembershipEditor(context)
             ];
         }
 
+        _bodySelection = body;
         _sections.forEach(function(section) {
             if (section.entityIDs) {
                 section.entityIDs(_entityIDs);
--- a/modules/ui/field.js
+++ b/modules/ui/field.js
@@ -11,6 +11,7 @@ import { uiFields } from './fields';
 import { LANGUAGE_SUFFIX_REGEX } from './fields/localized';
 import { uiTagReference } from './tag_reference';
 import { utilRebind, utilUniqueDomId } from '../util';
+import { renderSubtagIcons, renderSubtagExpanded } from './field_subtag_icons';
 
 
 export function uiField(context, presetField, entityIDs, options) {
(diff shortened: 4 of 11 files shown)


mfuhrmann:
This looks really promising. Is there some kind of test instance where this change could be applied?

tordans:
> Is there some kind of test instance where this change could be applied?

Yes, it is at https://pr-12005--ideditor.netlify.app/#disable_features=boundaries&map=20.33/52.45677/13.44966&background=Bing&id=w1126644207

This road segment shows the UI for parking and cycleway tags, for example.

mfuhrmann:
@tordans Where do I need to sign the contract? :joy: But seriously, this is a nice improvement.

matkoniecz:
> I did not review the code in details, yet. 

Is it LLM-generated or obtained in some other way?

tordans:
> Is it LLM-generated or obtained in some other way?

@matkoniecz I tried to be very clear in my description that **this PR is not yet about the code, but to discuss a UI and UX direction.** This is further indicated by the draft state of this PR which is also a signal that the code is not ready for review, yet. – To answer your question: Yes it is. (As will be all my code in the future.) Before I submit the code for review I will have reviewed each line myself and tested the PR meticulously which I will document in the PR.

tordans:
**A thought about the UI:**

Thinking about this further, I think the button position and style in this draft are not good enough. This section is about the tag itself (delete, info) and the related get kind of hidden. I also clicked delete by accident so it needs some kind of separation at the very least (which is hard to style).
> Bildschirmfoto 2026-03-14 um 16 14 25

Maybe moving the options down into a small (in text and style) line below the field will be better.
```
Fieldlabel       [D] [i]
[______]
Related fields: _Check date_, _Note_, _Conditions_, Subtags
```

Where `_Note_` and so on is a link that will show the fields below.

**A thought about the sub-fields:**

I was hoping we can get away with just the three fields types (text, text+today, multiline text) that I used above. This will be enough for the subfield-groups Checkdate, Note, Source, and most likely Number-Subtags Also the condtional tags at the current state (because we render them as text if at all). However, for the related tags like `bicycle:*` eg `bicycle:surface` we would want the specific preset field the way we specify it in the schema…

So either we reduce this PR to just show the tags and link to the "Tags" section for editing. 
Or the PR needs to be able to pick the right schema field based on an cleaned subtag-tag (eg. show the `surface` field for `bicycle:surface`).

**Info-i:**

Right now the sub-fields don't have the info-i and delete buttons, but they should have.

1ec5:
If you haven’t already, check out what OpenHistoricalMap did with subfields in OpenHistoricalMap/iD#187 and OpenHistoricalMap/iD#217. It involved more DOM-building code than I would’ve liked, but users have taken to these features quite well over the past couple years. (JOSM users have been clamoring for these iD features, fancy that.) The key to making this design work is predictability: the EDTF subfield is activated by a ➕ button within the date field’s body, because it’s only available on date fields. (It’s based on the Multilingual Name subfields.) By contrast, the source subfield is activated by a **@** button within the field’s header, because it’s available on every field.

I’m not sure that either design will scale to more than a couple possible subfields in a given field. Already I frequently mix up the ➕ and **@** buttons before the EDTF validator catches me. Sided and lane-based fields potentially complicate the UI beyond intuitiveness. Basically I expect to end up with a visual representation of [this eye-glazing table](https://wiki.openstreetmap.org/wiki/Order_of_key_parts). Given the limitations of this approach, we could use it for a few basic attributes, then turn the focus to more specialized field designs like #387 and #974.

matkoniecz:
Overall it is really great interface!

Could also `cycleway:surface` and `footway:surface` be made to fit in?

Do we want to support adding these associated tags? Probably no

> tagname:1, tagname:2

Interface for those was really,  really confusing

1ec5:
>> tagname:1, tagname:2
> 
> Interface for those was really, really confusing

OpenHistoricalMap/iD#223 implements this differently by leveraging the existing Add Field menu. As soon as you fill in the Source field, “Source 1” appears in the Add Field menu. If you select that and fill in Source 1, “Source 2” appears in the Add Field menu, and so on. The numbered Source fields are all sorted numerically next to each other. We avoided nesting them under the original field, so that it doesn’t look like one is more important than all the others. This design avoids cluttering the field’s mini toolbar with yet another control that the user has to distinguish. It also has the side benefit of subtly telling you there’s already a source when you go to add another one.

matkoniecz:
How long this test/demonstration PR should be open?

RudyTheDev:
Leaving some of my thoughts as per request coming from #12332.

I think it's a good proof-of-concept long-term proposal that could by default support all (or an easily-maintainable list of) sub-keyed fields without any knowledge about specifics. Personally, I am strongly in favour of showing more fields, especially for subkeys.

It has been my experience that mappers often omit or don't notice/fix subkeys (e.g. `maxspeed:type` as a quick example), so anything to display _at least_ what's already tagged is an improvement in my book. It's also my experience that a big part of why mappers don't tag common "easy" subkeys (like the `cycleway:surface` case) is because iD does not have these and entering them in raw tag editor is a comparatively tedious process. Most of the time, I have to search for the actual syntax, as I can never remember whether it's `surface:cycleway` or `cycleway:surface`.

Overall, this feels very much like a long-term design and implementation feature. In general, I would not expect this to get implemented quickly, nor immediately support certain things in a general way that something like `directionalCombo` can already do. Then there are things like life-cycle prefixes, which need their own special handling. Or things like lane editor. Those aren't even generic tag features, but they haven't been implemented (yet).

In other words, I don't think it should limit or replace custom fields for specific tags like the #12332 or existing `directionalCombo` plus any other custom fields that might have dedicated custom presentation like some of the examples 1ec5 mentions. You know, perfect is the enemy of good and all that. In fact, I would see custom fields as improvements to those cases where the generic version cannot do something.

I also think this would work really well in addition _to_ custom fields (without needing replace or be replaced by them) like the "bike lanes" example, as one could incorporate "additional fields/subkeys" toggles and display extra fields that aren't already shown by the custom field, but the whole left/right/both key update logic is preserved.

One concern of mine would be how this would support many of such OSM key peculiarities without the code just becoming a long list of hardcoded cases. Even in something as simple as the `cycleway:surface` I found multiple edge cases that I never even thought about. I imagine there are many things that would need to be considered for this. In other words, it seems that there is a significant distance going from "this works for these cases" to "this works for all cases nicely" and "this doesn't break for all these exceptions". It would certainly help others to see some sort of implementation plan and timeline.

In that vein, one thing I do not like is the direction of `field_subtag_preset_config.js` and parts of `subtags.ts`. As the core of the feature, these should be well-designed and seperate logic from data. I realize it's proof-of-concept and LLMed, but (barring any mention otherwise) this currently feels like taking the direction of creating a parallel field configuration scheme to tagging repo, but hard-coding it all in iD code. In my opinion, for a general solution like this, the goal should be the opposite - allow tagging maintainers and proposers to discuss and fully-utilize the subtag functionality in the best way for each key, while the iD provides the general functionality - like it is now for most of other fields. The code of those files should be addressed for long-term maintainability. Especially for iD forks that may not actually want (or have) such hard-coded fields.

As already mentioned, this also has extra complexity of what input field type to show for sub-keys, like the `bicycle:surface` needing `surface` field as base example - all of which again needs to have some sort of way to be specified generically. I think this is an important point because a user should never be presented with inconsistent inputs (i.e. surface works here, but not here).

I think the UI is heading roughly in the right direction by "nesting" subkeys visually. I do however think it needs to be more compact. Again, I realize it's proof-of-concept, so criticizing it feels unfair, but the UX _is_ basically the strength of iD and half the purpose of this feature, so proposing a strong visual solution sounds like the main priority.

I really like the idea of additional icons for specific subkeys (I also agree they should be placed under the inputs, so they do not mix with the "action buttons" especially as those change as you edit). Ideally, the icons would also appear next to the fields they reveal, so the user learns what they mean, but I digress into specifics. Although, I again have to point out how this (currently) ends up hard-coding much of this instead of being available via tagging.

I also see mentioned above concerns of scale - like what if there is a subtag of a subtag of a subtag and such. I think it's worth considering, but I also think that for common usable fields such precision is likely an overkill and in many cases it's better solved by custom fields if truly needed. My only suggestion would be to provide some sort of visual indication that there _are_ more subkeys, so the user is at least aware.

Anyway, in short, I think this is as a good core improvement direction and an essential feature for iD. I think it's great that it solves a large bulk of missing subtagging display. I think UX and a generic support of subkeys are the obvious strong parts of this, but my current early reservations are about hard-coding and maintainability.

If there's one broad observation I would mention, then it's that there is no clear "approved for implementation" status for this repo. It's a lot of work to build something like this, so _as an outsider_ I would hope maintainers would indicate whether this would even ever get integrated.

Please be aware that I am approaching this feedback more as a picky user rather than a developer, so please read my criticism as constructive.

PR #12010, opened 2026-03-14
Combo/access: standardize tooltips to Tag: key=value, add formatTag helper

tordans:
Follow up to https://github.com/openstreetmap/iD/pull/11876 and https://github.com/openstreetmap/iD/issues/10613

The current setup is a bit messy and very hard to understand/predict as a user.
* [Preview Current UI](https://ideditor.netlify.app/#map=19.00/52.78767/13.26743&disable_features=boundaries&background=Brandenburg-DOP20c&id=n-1)
* [**Preview this PR**](https://pr-12010--ideditor.netlify.app/#map=19.00/52.78767/13.26743&disable_features=boundaries&background=Brandenburg-DOP20c&id=w194206774)

This PR change the logic to …
- a tagging schema description is always(\*) right below the value-label; but only there (we had it duplicated in the tooltip in some cases) 
    ((\*) the access field handles this differently but that is OK IMO)
- a taginfo description – if present – is always only part of the tooltip; but we show it in more cases now; and we handle an edge case where the taginfo description is just the tag key, which resulted in duplicated lines.
- the actual tag is always visible in the tooltip as "Tag foo=bar" which makes it easy to predict what will be tagged (for advanced users); for the taginfo description case, we show both in separate lines.

| Tag | Before | After |
|-----|--------|-------|
| surface | surface-before | surface-after-taginfo-desc |
| incline | incline-before | incline-after |
| access | access-before | access-after |
| tunnel-raw | tunnel-raw-before | tunnel-raw-after-taginfo-desc<br>tunnel-raw-after-notaginfo |
| tunnel | tunnel-before | tunnel-after-taginfo-desc |
| tracktype | tracktype-before | tracktype-after |
| multi-raw | multi-raw-before | multi-raw-after |
| multi-trans | multi-trans-before | multi-trans-after |
| smoothness | smoothness-before | smoothness-after |


tyrasd (commented):
Is it intentional that the description is used at the same time in the hover text as well for the multi-line combo box options?



To me it seems like in this case the description could be omitted in the hover tooltip?!

tordans:
> Is it intentional that the description is used at the same time in the hover text as well for the multi-line combo box options?
> To me it seems like in this case the description could be omitted in the hover tooltip?!

Yes, my goal was to remove this duplication. I will look into it.

Side note: I though maybe this is a case where we have both, a `description` in the `id-tagging-schema` https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/parking.json as well as a Data item https://wiki.openstreetmap.org/wiki/Item:Q21762 
But the wording of the data item is different, so it is in fact a duplicated tagging schema description.

tordans:
Now fixed. 

I changed the code quite a bit because the logic was different from what I thought. We only have the taginfo data in one part of the code, so only there do we need to handle the case where possibly both is present (preset desc + taginfo desc).

> image

tyrasd (approved):
Great, thanks. Definitely a nice quality of life improvement!

maro-21:
I'm not sure if this is related to that pull request, but I'm assuming it is, which is why I'm posting here (if not, let me know, and I'll create a separate issue).

I was testing it on 2.40.0-dev version and what I can see is that the tooltip for surface appears in my language only for the first value - asphalt, but for others there are English descriptions, even though the descriptions of surfaces on data items I translated years ago and they correctly appear after adding surface=unpaved and clicking the "i" icon.

github_iD_surface_tooltip1
github_iD_surface_tooltip2
github_iD_surface_tooltip3

1ec5:
Yes, please open a new issue so we can investigate this issue. I can reproduce it in this case. At a glance, it does appear that we’ve regressed the tooltip because we only consult taginfo, which intentionally has poorer language coverage than the wiki’s data items.

Issue #3945, opened 2017-04-02
[feaure request] [usability] improve access value editor


Locke:
A user struggled to update access tags, as "motorcar" isn't visible in the upper part of the editor.

![screenshot from 2017-04-02 12 57 58_](https://cloud.githubusercontent.com/assets/53905/24586714/7cffdcb8-17a7-11e7-982a-26e57f3bdbc3.png)

Initially motor_vehicle=agricultural and motorcar=no were set. As he couldn't see motorcar in the upper editor he couldn't change it. (In later attempts he added access=yes, foot=yes, motor_vehicle=yes, bicycle=yes and horse=yes which all had no effect to the more specific tag motorcar).

I suggest to add more possible tags, e.g. everything from https://wiki.openstreetmap.org/wiki/Key:access#Transport_mode_restrictions . As just adding everything would be too much I think most keys should be hidden with a (+) icon which opens a tree.

Also, the default (grayed out "yes") needs to be updated depending on higher access tags: When a way has vehicle=no motor_vehicle=destination the default value of bicycle is still displayed as grayed out "yes", however it should be "no" as bicycle is included in vehicle. As I see it from #2213 only the value of "access" is inherited and not "vehicle".

freakyuser-osm:
I think hgv should be added as an preset to the access presets in id. This way it's easier for beginners to map truck access restrictions.

(See: https://github.com/openstreetmap/iD/issues/3946)

bhousel:
> I suggest to add more possible tags, e.g. everything from https://wiki.openstreetmap.org/wiki/Key:access#Transport_mode_restrictions . As just adding everything would be too much I think most keys should be hidden with a (+) icon which opens a tree.

I like this suggestion because it would let us support more keys, and at the same time we could drop the existing keys for things like `horse` `bicycle` `foot` that are almost never used and just take up space.



> Also, the default (grayed out "yes") needs to be updated depending on higher access tags: When a way has vehicle=no motor_vehicle=destination the default value of bicycle is still displayed as grayed out "yes", however it should be "no" as bicycle is included in vehicle. As I see it from #2213 only the value of "access" is inherited and not "vehicle".

Yes, I never heard of `vehicle` until just now.

slhh:
I think the access field should show a complete list of all present access tags of the feature. This should include any ":conditional", ":lanes", ":forward", or ":backward" key modifiers or their combinations, because we should not display misleading incomplete information.
[...]

PR #12011, opened 2026-03-14
Access field: show all tagged access keys

tordans:
Ref https://github.com/openstreetmap/iD/issues/3945

The goal of this PR is to give visibility to access-tags that are not specified as part of our default list of access-tags.
The UI will now show the default list by default (same as we have now) but also any access tag that was added in the raw tags section or elsewhere.

So this PR does not tackle the question of how we explain the complex access hierarchy to mappers; its goal is to give visibility to data that is mapped already. This will make it more likely users change the access tags (if needed) when they modify the data; and less likely users double tag (eg `motor_vehicle` and `vehicle` because one is not visible).

This is especially useful with tagging like `motor_vehicle` (our default) vs. `vehicle` for bicycle roads, for example ([Wiki](https://wiki.openstreetmap.org/wiki/DE:Tag:bicycle_road%3Dyes)).

The PR includes a sorted list of access tags derived from the Wiki ([and the traffic sign project](https://github.com/osmberlin/osm-traffic-sign-tool/blob/main/packages/traffic-sign-converter/src/signsToTags/utils/sortTags.ts)) which is used to keep the access keys in an order that is aligned with the wiki. The order reflects the wiki hierarchy of tags. But at the very least, it helps to render the tags in a stable order…
This results in horses being more visible than before, which we might not want…

The translations for the Keys are now part of iD core translations. Before, we used the id-tagging-schema translations for the 5 default access tags.

| | Before | After |
|---|---|---|
| **Regular** – [**Test-Link**](https://pr-12011--ideditor.netlify.app/#disable_features=boundaries&map=21.15/52.48110/13.45239&background=geovekst-nib&locale=en&id=w848033609) | before | after |
| **Bike Road** – [**Test-Link**](https://pr-12011--ideditor.netlify.app/#disable_features=boundaries&map=20.59/52.48273/13.44839&background=geovekst-nib&locale=en&id=w36739044) | bike-before | bike-after |
| `dog=leashed` – [**Test-Link**](https://pr-12011--ideditor.netlify.app/#disable_features=boundaries&map=20.59/52.48273/13.44839&background=geovekst-nib&locale=en&id=w114667578) | - | - |

----

Ping https://github.com/openstreetmap/id-tagging-schema/issues/2135


modules/ui/fields/access_keys.ts
```diff
@@ -0,0 +1,82 @@
...
+export interface AccessKeyEntry {
+    key: string;
+    default?: boolean;
+}
+export const ACCESS_KEYS: ReadonlyArray<AccessKeyEntry> = [
+    { key: 'access', default: true },
+    { key: 'foot', default: true },
+    { key: 'dog' },
+    { key: 'ski' },
+    { key: 'ski:nordic' },
```
tordans, line 19: We could remove low usage keys here like https://taginfo.openstreetmap.org/keys/ski%3Anordic with only ~3k usage

modules/ui/field.js
k-yle, line 1: this seems like a cool idea. do we need a button to add a value?

Something like this maybe? (to match the new design of the raw tag editor) ?
Image

data/core.yaml
```diff
@@ -2528,6 +2528,66 @@ en:
...
     # translate the letters but leave the format the same
     year_month_day: "YYYY-MM-DD"
+  fields:
+    access:
+      types:
+        access: All
+        foot: Foot
+        motor_vehicle: Motor Vehicles
+        bicycle: Bicycles
+        horse: Horses
```
k-yle, line 2296: > The translations for the Keys are now part of iD core translations. Before, we used the id-tagging-schema translations for the 5 default access tags.

just curious what the benefit is of changing this? it means that the translations unusable for other consumers of id-tagging-schema, right? because those apps only pull the translations from the tagging-schema repo, not from iD's repo

test/spec/ui/fields/access.js
k-yle, line 1: > this PR does not tackle the question of how we explain the complex access hierarchy to mappers

just thinking about the long term, it's not actually possible to model the heirachy as a tree right? because the heirachy is not a simple [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph). For example, is `bus` / `tourist_bus` / `school_bus` a subclass of `psv` or `hgv` or both or neither? The answer changes depending on your region, and will either create a loop in the graph, or an impossible tree structure

k-yle (commented):
some thoughts inline, to keep the discussion organised:

test/spec/ui/fields/access.js
matkoniecz: > just thinking about the long term, it's not actually possible to model the heirachy as a tree right? because the heirachy is not a simple [directed acyclic graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)

I think it is DAG, but may not be a tree

matkoniecz: > For example, is bus / tourist_bus / school_bus a subclass of psv or hgv or both or neither?

as hgv is heavy **goods** vehicle, neither should be subclass of hgv

matkoniecz: but rickshaw could be subset of psv and bicycle, while not all psv are subset of bicycle? So this could be tree counterexample

modules/ui/fields/access_keys.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Only covers land keys because the schema's "access" field is only used on
+ * highway/barrier/leisure presets. Waterway presets use "access_simple".
+ * @see https://wiki.openstreetmap.org/wiki/Key:access#Land-based_transportation
+ */
+
+export interface AccessKeyEntry {
+    key: string;
+    default?: boolean;
+}
+export const ACCESS_KEYS: ReadonlyArray<AccessKeyEntry> = [
```
matkoniecz, line 14: maybe ideally it would live in tagging schema repo, which exists mostly to prevent tagging issues from appearing in main iD project? And also makes it reusable by other projects?

Thought right now it would get stuck in https://github.com/ideditor/schema-builder hell, that just passed second birthday of its last release (recently I spend a lot of time to get the next release unblocked, but more work remains to enable this)

matkoniecz:
I have not looked at code at all, not tried to break it, ignored glitches.

Just display of `vehicle=` makes it beautiful.

modules/ui/fields/access_keys.ts
tordans: `id-tagging-schema` does not have a good place for lists like those IMO. And we have other lists deep inside of iD like for example https://github.com/openstreetmap/iD/issues/12035#issuecomment-4079095572. So it is not uncommon for us to have this here.

data/core.yaml
tordans: I always forget that we can use the `strings` object in the tagging schema as a home for more or less random translations… – So I should move this to https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/access.json#L16-L22, right?

modules/ui/field.js
tordans: That sounds like a good idea. We could also do it as a follow up, though, right?

On the UI side of this: Adding this line looks quite clean but it does expose the option to add more rows quite a bit. An alternative would be a `+` button next to the trash icon like we have with the `name` field. That would hide the option a bit more, IMO.

test/spec/ui/fields/access.js
tordans: I never understood the tagging enough to answer this :-D. But in any case I would vote to leave this for a follow up PR / discussion.

data/core.yaml
matkoniecz: I think so

modules/ui/field.js
k-yle: > We could also do it as a follow up, though, right?

🤷 

> An alternative would be a <kbd>+</kbd> button next to the trash icon like we have with the `name` field.

yeah that would be more compact, but i'm not sure how we would integrate the dropdown - would it appear immediately when you click the plus? like this:

image

tordans: Ah, I rememberd this being here…
> image

The idea was to click it and then a line like in your first screenshot shows.
This would be a special UI though, so yours is cleaner…

modules/ui/fields/access.js
```diff
@@ -29,7 +29,14 @@ export function uiFieldAccess(field, context) {
             .append('div')
             .attr('class', 'label preset-label-access')
             .attr('for', function(d) { return 'preset-input-access-' + d; })
+<<<<<<< HEAD
```
matkoniecz: Merge conflict

tyrasd: ooh, sorry, my bad! :grimacing: //edit: cleaned up here and in the develop branch: 4faa5c1ff68ae8116f935ead7d8289dec4a81990

tordans:
Here are two new iterations of this.
Personally I quite like the `+` click flow.

I also started moving the translations around but lets first clarify what UI direction to go and which values to offer in the translations. Marking this PR as a UI Exploration draft.

Next step:
1. Decide on a UI
2. Decide on which values to offer as keys
3. Then: Code cleanup and moving the translations to id-tagging-schema.

## Add access with a new line

https://github.com/user-attachments/assets/797efc52-bb7c-4841-8ffa-f69ea7da580e

## Add access with `+`-button

https://github.com/user-attachments/assets/9a417b79-5025-45a8-934d-05e25250be3d

matkoniecz:
> Add access with a new line

is it viable for brand new access line to appear at the bottom? Or would it require too much fiddling to achieve it?

Is there way to test both ways in action? From just looking at video I have no preference in either direction

tordans:
> > Add access with a new line
>
> Is there way to test both ways in action? 

No, ATM you would have to check out the branch and the check out the commit to test it.

> From just looking at video I have no preference in either direction

I remember from [the community chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Notes) that we agreed to use the Version with the `+` and the field right below the title (closer to the `+`). That was what I wanted to push next.

> is it viable for brand new access line to appear at the bottom? Or would it require too much fiddling to achieve it?

Don't see why not, but I think having it right below where you just clicked is easier to understand. It feels a bit more like the plus being the key-selector than an actual new line — Thinking about that, maybe I should refactor it this way and not the "add special line with just the key and then focus it right away to open the dropdown"-workaround I use now.

PR #12012, opened 2026-03-14
Render `access=unknown` as raw tag; but only the label

tordans:
Follow up to https://github.com/openstreetmap/iD/pull/12010 which introduced the description line below the label.
This PR changes the style, so that only the label is affected by the raw tag class and the description gets the regular font style.

Related to https://github.com/openstreetmap/iD/pull/12011.

A bit related to https://github.com/openstreetmap/iD/pull/12010 but should be mergable independently.

**Why…**

A while back we introduced a style difference between fields values from the preset that we promote as translated values and the raw values. IMO the `unknown` value in the access field is more like the raw tags than a promoted, translated value.

As such, we should render it differently; which this PR does.

We could also argue that we should not promote the value at all given that it is not used too often https://taginfo.openstreetmap.org/search?q=unknown#values and the field will render values other than the recommendations just fine… just not promote them.

https://wiki.openstreetmap.org/wiki/Tag:access%3Dunknown

---

Should we decide not to change rendering of the `access=unknown` I would extract the style changes here because they are a better fit IMO than what we have now. Even if https://github.com/openstreetmap/iD/pull/12010 clarifies that this combination is not used ATM, the combination is still valid from a code point of view.

---

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2486,8 +2486,11 @@ div.combobox:has(.has-description) {
     white-space: normal;
     padding-top: 2px;
 }
-.combobox .combobox-option.raw-option {
+.combobox .combobox-option.raw-option .combobox-option-label {
     font-family: monospace;
+    color: var(--text-color);
+}
+.combobox .combobox-option.raw-option .combobox-option-description {
     color: var(--text-color);
 }
 
--- a/modules/ui/fields/access.js
+++ b/modules/ui/fields/access.js
@@ -98,13 +98,17 @@ export function uiFieldAccess(field, context) {
             options.splice(options.length - 4, 0, 'dismount');
         }
 
-        var stringsField = field.resolveReference('stringsCrossReference');
+        const stringsField = field.resolveReference('stringsCrossReference');
         return options.map(function(option) {
+            const isRawKey = option === 'unknown';
+            const description = stringsField.hasTextForStringId('options.' + option + '.description')
+                ? stringsField.t('options.' + option + '.description') : undefined;
+            const title = isRawKey ? option : stringsField.t('options.' + option + '.description');
             return {
-                title: stringsField.t('options.' + option + '.description'),
-                description: stringsField.hasTextForStringId('options.' + option + '.description')
-                    ? stringsField.t('options.' + option + '.description') : undefined,
-                value: option
+                title: title,
+                description: description,
+                value: option,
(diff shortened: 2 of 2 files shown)


tordans:
**Testing**

https://pr-12012--ideditor.netlify.app/#disable_features=boundaries&map=21.15/52.48110/13.45239&background=geovekst-nib&locale=en&id=w848033609

| Before | After |
|--------|--------|
|  image |  Bildschirmfoto 2026-03-14 um 16 40 48 |

tyrasd (commented):
(please ignore, I accidentally posted this on the wrong PR)

tyrasd:
Why only for the value `unknown`? In principle, all of the listed values are raw tags (it's not very obvious from the English screenshots, but in other languages one can see that only the descriptions are translated)… So, would it not make sense to render them all in the monospace font? :thinking:

matkoniecz:
As I understand, this styling is used for unsupported tags from outside of presets. So `unknown` is in presets but inferior value so marked as such.

tordans:
> Why only for the value `unknown`? In principle, all of the listed values are raw tags (it's not very obvious from the English screenshots, but in other languages one can see that only the descriptions are translated)… So, would it not make sense to render them all in the monospace font? 🤔

Oh, I did not realize we did not translate those values. This changes things indeed. To follow the patterns we have elsewhere, the right thing would be to apply the monospace format to all of the values, then.

Or, we add translations to `id-tagging-schema`? – Actually, they already are https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/access.json
So the question is more… why are they not used, right?

> As I understand, this styling is used for unsupported tags from outside of presets. So `unknown` is in presets but inferior value so marked as such.

Yes, the logic is mixed up.

---

My take
- close this PR
- copy the style fix in a new PR because 
   >Should we decide not to change rendering of the access=unknown I would extract the style changes here because they are a better fit IMO than what we have now. Even if https://github.com/openstreetmap/iD/pull/12010 clarifies that this combination is not used ATM, the combination is still valid from a code point of view.
- does anyone want to discuss why we even have `unknown` in this list? My take is that we should remove it from the list and just show it whenever it is given as raw value (like we do with all non-suggested values) – can create a issue and or PR if this is something anyone wants to think about…
- should we look into the topic of missing translation values? I would open a PR for now…

PR #12016, opened 2026-03-15
Packages: Use es-toolkit instead of lodash

### LINKED ISSUE #12015: Package: Replace lodash with es-toolkit
There is this JavaScript Ecosystem Project that tries to modernize the ecosystem by helping packages to modernize their codebase – https://e18e.dev/docs/replacements/.

They have a list of suggestions for packages that have a more modern alternative – for example for `lodash` https://github.com/es-tooling/module-replacements/blob/main/docs/modules/lodash-underscore.md?plain=1

I looked into our usage of lodash and it looks like only use a couple of functions that are well supported by https://es-toolkit.dev/

Migration to es-toolkit should reduce our bundles size and be a bit faster in some cases.

----

The question for this issue is: Is this something worth doing that has a change to be merged?

---

It looks like right now there is no treeshaking happening(?)
> Image

Side node: We also have which could be replaced by the same es-toolkit.
> Image

tordans:
Closes https://github.com/openstreetmap/iD/issues/12015

I had a bit of time so I looked into this.
The all scripts run fine with these changes.
The docs suggest a 1 to 1 replacement.

After the change:
> image

---

We could reduce the number of changes by renaming the imports, so only the import lines change but nothing in the code…

---

The plan and execution of this change was done in Cursor and LLM supported.


tyrasd (changes requested):
Thanks for the suggestion.

I'm all good for the replacement of lodash with es-toolkit, including the slightly renamed imports: AFAICT the `_` at the start was just to make the function look more similar to how one would have used it in the past as `_.methodname(…)`. But as we did not even always stick to this naming convention, this would just make the naming more consistent again.

PS: I'd say that the dependency's package size is a negligible argument here, as the few kb plus or minus are not making a difference looking at some of the data files we load like the presets data and/or the editor layer index. But what is more important is their long term stability / maintenance, as well as the interoperability with the rest of the toolchain.

<hr>

The only thing I'm hesitant is the replacement of `fast-deep-equal`. From what I see in the commit history (cda8f657ac346b67abaa1bdf3bd2a988318ae358) it was introduced to address performance bottlenecks. And at least in some functions, it is doing some heavy lifting in some crucial code, e.g. in `difference.js`.

After doing some benchmarking (e.g. by adding es-toolkit to [fast-equals](https://github.com/planttheidea/fast-equals/tree/main?tab=readme-ov-file#benchmarks) benchmark suite), the performance of es-toolkit's `isEquals` does seem to be on par with (if anything slightly slower than) `lodash`, but is lagging behind by almost one order of magnitude behind `fast-deep-equal` (and `fast-equals` which would seem like a more modern / well maintained alternative to it).

I'm pretty sure the performance benefits are not at all relevant in many of the modules that currently use `fast-deep-equals`, but for some they might make a notable difference. So, what I would like to suggest would be to:

1. limit this PR to only replace lodash with es-toolkit, and leave the fast-deep-equals imports untouched
2. make a separate PR to replace `fast-deep-equals` with [`fast-equals`](https://www.npmjs.com/package/fast-equals) which is as fast as (or sometimes even slightly faster than) `fast-deep-equals` but generally more well maintained, appears to be able to handle more edge cases (even compared to es-toolkit's method it seems) and also comes with its own type definitions

tordans:
@tyrasd yes, happy to update this, thanks the deep dive! 

I wonder why they don't include `fast-equals` on https://e18e.dev/docs/replacements/deep-equal.html 🤔

tordans:
@tyrasd the PR is updated.

I removed all lodash isEqual cases so we can remove the other library.
```diff
- import { isEqual } from 'lodash-es';
+ import deepEqual from 'fast-deep-equal';
```

tyrasd (approved):
Thanks. I changed the previously `isEquals` from lodash to use the es-toolkit version, to keep this PR as minimal as possible. Let's think about switching that in the follow-up PR. 😊

PR #12020, opened 2026-03-15
Set 'Show Third Party Icons' unchecked by default to respect user pri…

### LINKED ISSUE #12019: Displaying icons from external servers should be disabled by default
- Open a new incognito window in your browser and log in on www.openstreetmap.org (It's important that your browser doesn't contain any cookies indicating that the editor has been launched in the past)
- Go to https://www.openstreetmap.org/edit?node=12601250398#map=19/52.503648/13.338985
- When you first launch the iD editor on an object that has [brand:wikidata] tags, iD displays an icon from an external source. But before I start editing, a message appears and iD asks me if I want to display icons from external sources... This option is selected by default, although I can deselect it. But iD has already downloaded this icon from an external server. So if iD wants to respect my privacy, this option should be unchecked by default, before the user makes a decision.

Image

VaibhavSingh2006:
Fixes #12019

Previously the "Show Third Party Icons" option defaulted to enabled due to:
prefs('preferences.privacy.thirdpartyicons') || 'true'

This change sets the default to false so that external icons are not loaded before the user explicitly enables the option.

### DIFF
--- a/modules/ui/sections/privacy.js
+++ b/modules/ui/sections/privacy.js
@@ -19,7 +19,7 @@ export function uiSectionPrivacy(context) {
 
       let thirdPartyIconsEnter = selection.select('.privacy-options-list')
         .selectAll('.privacy-third-party-icons-item')
-        .data([prefs('preferences.privacy.thirdpartyicons') || 'true'])
+        .data([prefs('preferences.privacy.thirdpartyicons') || 'false'])
         .enter()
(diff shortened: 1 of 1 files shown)


tordans:
The right issue is https://github.com/openstreetmap/iD/issues/12019

tyrasd:
closing for now until we decide on how to approach this

PR #12021, opened 2026-03-15
Packages: Replace `fast-deep-equal` with `fast-equals`

tordans:
Based on https://github.com/openstreetmap/iD/pull/12016#pullrequestreview-3950331060

> make a separate PR to replace fast-deep-equals with [fast-equals](https://www.npmjs.com/package/fast-equals) which is as fast as (or sometimes even slightly faster than) fast-deep-equals but generally more well maintained, appears to be able to handle more edge cases (even compared to es-toolkit's method it seems) and also comes with its own type definitions

This PR will likely be in conflict with https://github.com/openstreetmap/iD/pull/12016 (whatever is merged first). Will rebase it once the other one is merged…

Package comparison https://npmx.dev/compare?packages=fast-equals,fast-deep-equal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "diacritics": "1.3.0",
         "es-toolkit": "^1.45.0",
         "exifr": "^7.1.3",
+        "fast-deep-equal": "~3.1.3",
         "fast-equals": "^6.0.0",
         "fast-json-stable-stringify": "2.1.0",
         "idb-keyval": "^6.2.2",
@@ -352,7 +353,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
@@ -401,7 +401,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       }
@@ -1255,7 +1254,6 @@
       "integrity": "sha512-/g2d4sW9nUDJOMz3mabVQvOGhVa4e/BN/Um7yca9Bb2XTzPPnfTWHWQg+IsEYO7M3Vx+EXvaM/I2pJWIMun1bg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@octokit/auth-token": "^4.0.0",
         "@octokit/graphql": "^7.1.0",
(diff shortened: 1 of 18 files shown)


matkoniecz:
https://github.com/openstreetmap/iD/pull/12016 got merged, so this one needs to be updated

tordans:
FYI I rebased this against current develop.

package-lock.json
```diff
@@ -352,6 +352,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
```
tyrasd, line 355: PS: Those unnecessary `"peer": true` changes seems to originate form a [bug](https://github.com/npm/cli/issues/8690) in npm v11.6.2. Updating to a version of node with npm >= 11.6.3 fixes this.

package-lock.json
```diff
@@ -8834,6 +8857,29 @@
         "url": "https://github.com/sponsors/colinhacks"
       }
     },
+    "node_modules/netlify-cli/node_modules/@netlify/db-dev": {
```
tyrasd, line 8860: PSS: Those are also changes that are unrelated to the newly added dependency. They cause minor inconveniences such as a too large diff, and they will be cleaned-up by subsequent dependabot PRs, but are also a potentially harmful issue (see below). Those dependencies are actually only _optional_ dependencies of one of our dev dependencies (see `"optional": true,` below; newer npm versions will instead mark them as `"extraneous": true`, but that's similarly unnecessary to include in the package-lock), and should AFAICT not be included in the package-lock in general.

As mentioned above, these extraneous dependencies are actually harmful, because they will cause _Unsupported platform_ errors when a subsequent user runs `npm clean-install` on this package-lock file:

```sh
$ npm ci
npm error code EBADPLATFORM
npm error notsup Unsupported platform for @rollup/rollup-android-arm-eabi@4.52.2: wanted {"os":"android","cpu":"arm"} (current: {"os":"linux","cpu":"x64"})
npm error notsup Valid os:   android
npm error notsup Actual os:  linux
npm error notsup Valid cpu:  arm
npm error notsup Actual cpu: x64
```

Unfortunately, I did not find a "good" way to exclude those unnecessary changes when installing a new dependency via `npm install …` (neither `--omit=optional` nor specifying the same via `.npmrc` seems to work). A workaround that works is to use `npm install … --omit=dev` (which strangely does seems work for dev dependencies `npm install --save-dev … --omit=dev` 🤷 ).

At first glance this does look like a bug in npm, idk. 🤷 I did find https://github.com/npm/cli/issues/8128, but that was closed without a fix, sadly. If someone knows a way to circumvent this issue, I'd be glad for tips. //cc @k-yle: have you experienced this as well?

package-lock.json
tordans: Thanks for noticing; updating npm helped indeed.

package-lock.json
tordans: @tyrasd thanks for this fixup and wow, this is a frustrating bug(?). I am now on the latest npm (11.12.0) but still get issues whenever I `npm i`. Right now iterating on the wayback branch and the build fails…

So whenever you learn something new about this, please share :-).

tyrasd: Yeah, it's indeed annoying.

>  Right now iterating on the wayback branch and the build fails…

Could it be that you upgraded node to v25 to get the newest npm? In which case #12047 might be related.

//edit: another potential hint: after adding a new dep with `--omit=dev`, it's necessary to afterwards run `npm clean-install` again to actually get the required dev dependencies installed again for the build.

tordans: Ah, thanks, that helped.
`npm i --omit=dev @esri/wayback-core && npm clean-install`
I am on v24.14.0 (npm v11.9.0) so AFAIK its all good from this side, right?

erictheise: > `npm i --omit=dev @esri/wayback-core && npm clean-install`

Thank you for documenting this, @tordans; I've been doing a catch-up with the OpenHistoricalMap fork and I could not figure out how android os errors was suddenly preventing our GitHub actions from running.

PR #12029, opened 2026-03-16
Add 2025 hennepin county imagery

freaker2112:
Added 2025 Hennepin County orthoimagery and verified alignment with previous years, although for some reason this year's imagery seems 'blockier' than last year's in a local build of iD, even though they appear the same on [Hennepin County's Viewer](https://gis.hennepin.us/ObliqueViewer/), perhaps I'm missing a sharpening setting?

(resubmitting due to previous error in image capture season)


wolfy1339:
This should probably go to https://github.com/osmlab/editor-layer-index instead

iD then can pull from there and update the list

matkoniecz:
see https://github.com/freaker2112/iD/commits/1ade7e66bd15f5d2d58c72c53d567c1883524b15/data/imagery.json

any manual edits to this file will be lost

PR #12030, opened 2026-03-16
Fix: wrong invalid url warnings

### LINKED ISSUE #11983: Wrong "invalid URL" warning

bhavyaKhatri2703:
Fixes #11983

Removes the `link.href.includes(url)` check which falsely flagged valid URLs containing accented characters, since `new URL()` normalizes both before comparison. Also handles the case for Upper case letters.
Fix suggested by @k-yle in #11985.

Added tests for accented characters, IDN domains, and uppercase domains (as added by @tyrasd  in 6adfc8c).


k-yle (dismissed):
Previous discussion is at #11985. 👍 from me, since this fixes the false-positives.

@1ec5 and @hlfan, can you think of any false-negatives that this might cause?

hlfan:
Yes, see #12036 😅

But checking for the url schema would probably be enough in that case.

modules/validations/invalid_format.js
```diff
@@ -29,9 +29,8 @@ export function validationFormatting() {
 
         function isValidURL(url, strict = false) {
             try {
-                // First try strict WHATWG parsing
-                const link = new URL(url);
-                return link.href.includes(url);
+                new URL(url); // eslint-disable-line no-new
+                return true;
```
k-yle: okay, so based on https://github.com/openstreetmap/iD/pull/12030#issuecomment-4081574025 we'll need something like this:

```suggestion
                const link = new URL(url);
                return link.protocol.startsWith('http');
```

bhavyaKhatri2703: > okay, so based on [#12030 (comment)](https://github.com/openstreetmap/iD/pull/12030#issuecomment-4081574025) we'll need something like this:

yeah this passes the test in #12036

hlfan: Then I'd only change the return statement and leave the upper two lines as they are.

bhavyaKhatri2703: merged the #12036 test and made the changes

modules/validations/invalid_format.js
```diff
@@ -29,9 +29,8 @@ export function validationFormatting() {
 
         function isValidURL(url, strict = false) {
             try {
-                // First try strict WHATWG parsing
```
hlfan, line 32: Is this comment really that unnecessary?

bhavyaKhatri2703: well i guess we can keep it. although its a very obvious comment. But yeah ill revert.

hlfan (commented):
ICYMI I can't approve this, but I'd've put this into 2.39

k-yle (approved):

tyrasd (approved):

Issue #11995, opened 2026-03-10
Panoramax: mouse wheel doesn't work with panoramic photos


Mahabarata:
### Description

- follow this link https://www.openstreetmap.org/edit#map=19/43.586641/1.477698 in Firefox
- activate Panoramax
- select one of the panoramax
- after imagery load, try to use the mouse wheel: zoom doesn't work with panoramic photos
- nb : it works with Edge and Chrome

- follow this link https://www.openstreetmap.org/edit#map=18/43.593987/1.309068
- activate Panoramax
- select one of the panoramax
- after imagery load, try to use the mouse wheel: it works with flat photos
- activate Bing streetside
- select one of the streetside
- after imagery load, try to use the mouse wheel: it works with panoramic photos

Could you improve panoramic photos for Panoramax with Firefox ?

danieldegroot2:
Reproduced on
2.38.0-live, 2.38.0-release, 2.39.0-dev
Firefox 148.0
Windows 10
Regardless of display resolution, browser zoom, panellum size, image, sequence, high resolution toggle.
While/after zooming out (not in) using the minus button or double left click, or rotating one or more times, it is possible to zoom in (not out) using the scroll wheel temporarily. ([Example](https://ideditor.netlify.app/#disable_features=boundaries&map=19.22/52.65066/-7.25134&background=EsriWorldImageryClarity&photo_overlay=panoramax))
There are no browser console errors/warnings.
[suggested_tags:`bug-browser-specific` `streetlevel` `usability`]

anuragc0des:
Hi, I’ve opened this PR: https://github.com/openstreetmap/iD/pull/12032 to fix the issue with mouse wheel zoom not working correctly for Panoramax panoramas in Firefox.

The issue was caused by differences in how Firefox handles wheel events (deltaMode and deltaY). I’ve addressed this by normalizing the wheel delta values and implementing a custom zoom handler for consistent behavior across browsers.

Could you please review it and let me know if any changes or improvements are needed? I’d be happy to update the PR accordingly.

PR #12032, opened 2026-03-17
Fix Panoramax/ Pannellum wheel zoom smoothing for Firefox

anuragc0des:
Firefox wheel events use deltaMode and deltaY values differently than Chrome. Pannellum’s default wheel zoom behavior could feel inconsistent or jumpy in Firefox for panoramic images.

Changes

- Disabled Pannellum’s default mouseZoom
- Added a custom wheel event handler to the panorama viewer
- Normalized wheel delta values across browsers:
                Handled pixel-based (deltaMode === 0)
                Scaled line-based (deltaMode === 1)
                Scaled page-based (deltaMode === 2)

- Implemented smooth exponential zoom scaling:
              newHfov = currentHfov * 2^(deltaY / 500)

- Clamped HFOV values between 10 and 90
- Applied event.preventDefault() only within the panorama viewer to avoid interfering with global scrolling

### DIFF
--- a/modules/services/pannellum_photo.js
+++ b/modules/services/pannellum_photo.js
@@ -1,192 +1,232 @@
-import { select as d3_select } from 'd3-selection';
-import { dispatch as d3_dispatch } from 'd3-dispatch';
-import { utilRebind } from '../util';
+import { select as d3_select } from "d3-selection";
+import { dispatch as d3_dispatch } from "d3-dispatch";
+import { utilRebind } from "../util";
 
-
-const pannellumViewerCSS = 'pannellum/pannellum.css';
-const pannellumViewerJS = 'pannellum/pannellum.js';
+const pannellumViewerCSS = "pannellum/pannellum.css";
+const pannellumViewerJS = "pannellum/pannellum.js";
+const PANNELLUM_MIN_HFOV = 10;
+const PANNELLUM_MAX_HFOV = 90;
+const PANNELLUM_WHEEL_LINE_PIXELS = 18.001;
+const PANNELLUM_WHEEL_PAGE_PIXELS = 180;
 
 export async function pannellumPhotoFrame(context, selection) {
-    const dispatch = d3_dispatch('viewerChanged');
+  const dispatch = d3_dispatch("viewerChanged");
 
-    const module = {};
-    module.event = utilRebind(module, dispatch, 'on');
-    module.loadPannellum = function(context) {
-        const head = d3_select('head');
+  const module = {};
+  module.event = utilRebind(module, dispatch, "on");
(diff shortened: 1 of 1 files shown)


modules/services/pannellum_photo.js
```diff
@@ -1,192 +1,232 @@
...
+  const viewerEl = d3_select("#ideditor-pannellum-viewer");
+  viewerEl.on("wheel.pannellum", function (d3_event) {
+    if (!_pannellumViewer || !_pannellumViewer.getHfov) return;
+
+    d3_event.preventDefault();
+    d3_event.stopPropagation();
+
+    let deltaY = d3_event.deltaY;
+    if (d3_event.deltaMode === 1) {
+      deltaY *= PANNELLUM_WHEEL_LINE_PIXELS;
```
matkoniecz: this seems quite fragile and likely to not work well across multiple different OS and device types and browsers and time

anuragc0des: Thanks for pointing that out. I see how this approach could be fragile across different environments.
I’ll revisit this and try to come up with a more robust solution. Please let me know if you have any suggestions.

matkoniecz: 1) it is still hardcoding specific pixel counts, I really hope that is not the proper solution for the problem.

2) have you really tested this solution within few minutes from me making this comment to new commit appearing? Actually, have you tested this change at all? Which specific problem is being fixed?

modules/services/pannellum_photo.js
```diff
@@ -1,192 +1,247 @@
-import { select as d3_select } from 'd3-selection';
-import { dispatch as d3_dispatch } from 'd3-dispatch';
-import { utilRebind } from '../util';
+import { select as d3_select } from "d3-selection";
```
matkoniecz, line 1: needless extra reformatting (and yes, it is a common mistake made by low-end LLMs)

modules/services/pannellum_photo.js
anuragc0des: You’re right, I pushed the changes before completing proper testing, which was my mistake. I was still iterating on the approach and should have verified it first. I’ll revisit this, test it properly, and work on a more robust solution instead of relying on hardcoded values.

I’ll update the PR once I have a cleaner and well-tested fix.

modules/services/pannellum_photo.js
anuragc0des: Thanks for pointing that out. I’ll remove the unnecessary formatting changes.

tyrasd:
Besides the above mentioned deficiencies, this does not correctly identify the underlying issue causing the bug #11995.

PR #12036, opened 2026-03-17
Add test for Wikimedia Commons image URL validation

hlfan:
The Wikimedia Commons image URL fix introduced in #11499 doesn't have a test.
I added one.

### DIFF
--- a/test/spec/validations/invalid_format.js
+++ b/test/spec/validations/invalid_format.js
@@ -95,6 +95,17 @@ describe('iD.validations.invalid_format', function () {
             var issues = validate(entity);
             expect(issues).to.have.lengthOf(0);
         });
+
+        it('should suggest moving image URLs to Wikimedia Commons', function() {
+            var entity = createPointWithTags({
+                image: 'File:OpenStreetMap-Editor iD Logo.svg'
+            });
+            var issues = validate(entity);
+            expect(issues).to.have.lengthOf(1);
+            expect(issues[0].type).to.eql('invalid_format');
+            expect(issues[0].subtype).to.eql('website');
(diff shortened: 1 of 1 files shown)


k-yle (approved):
lgtm. for new code, [`expect.objectContaining`](https://jestjs.io/docs/expect#expectobjectcontainingobject) is a bit easier for test cases like this, and makes debugging easier. but it's probably better to match the existing code style here

hlfan:
Is the icon check enough or should the title or id also be checked?

k-yle:
> Is the icon check enough or should the title or id also be checked?

imo we should be using `id` for every `validationIssueFix`, but since we don't do this, the icon is the only stable ID[^1]. so i think it's fine, maybe @tyrasd can decide

[^1]: `dynamicFixes()[0].title.info.text` would probably work too, but then it'll break if [core.yaml](https://github.com/openstreetmap/iD/blob/develop/data/core.yaml) is updated.

tyrasd (approved):

tyrasd:
> Is the icon check enough or should the title or id also be checked?

Sorry, I overlooked this question before merging the PR.

Ideally, we would not only test that there is a fix offered, but instead that the fix does _actually fix_ the issue in the way we expect it to work. E.g. something like this:

```js
issues[0].fixes(context)[0].onClick(context);
const fixedEntity = context.entity(entity.id);
expect(fixedEntity.tags.image).to.be.undefined;
expect(fixedEntity.tags.wikimedia_commons).to.eql(entity.tags.image);
```

An issue `id` would be useful for cases when there are exists more than one dynamic fix, and we don't want to rely on a hardcoded order in the tests, but that's not needed here at the moment, as there's only one fix to test.

//edit: I added this now in 0f02fb23f

PR #11998, opened 2026-03-11
refactor `uiCmdSequence` into a separate function

k-yle:
prerequisite for #11999 

In that PR I want to reuse the logic that converts:
```js
{  modifiers: ["⌥"], shortcuts: ["+", "-"], separator: "," }
```
into:

> <kbd>⌥  Option</kbd> <kbd>+</kbd>, <kbd> -</kbd>

To test: check that the [keyboard shortcuts popup](https://github.com/user-attachments/assets/8879a530-79dd-4edf-833b-b55c3e6d81a4) still renders correctly.

Issue #7914, opened 2020-08-18
New preset: natural=hot_spring


maro-21:
https://wiki.openstreetmap.org/wiki/Tag:natural%3Dhot_spring

PR #12038, opened 2026-03-18
Selective autoformatting on changed code only

tordans:
Autoformatting for new TS files in iD.

<details><summary>Version of this PR from 2026 May</summary>
<p>

This is a new take to maybe resolve the issue of formatting and linting for iD.

**Lets first use this Draft PR to discuss if this approach is something that we want to look into.**
If so, there are things to check and test first, before this is ready for review…

**What we learned so far…**
_(This list is meant to show what I understand of the status quo and what was the basis for this PR.)_

- We don't want to reformat the whole codebase – https://github.com/openstreetmap/iD/pull/11313
- We have a mix of existing formatting in our code
- We don't have very clear formatting guidelines in our docs – https://github.com/openstreetmap/iD/pull/11998#pullrequestreview-3954923921
- Many of our PRs have formatting issues; some because users don't pay attention or don't review what their LLM suggest properly; some because editors have trouble to understand our guidelines
- We don't want LLM rules for this kind thing – https://github.com/openstreetmap/iD/pull/11311
- We gradually increased our Eslint rules recently to be more strict with some things
- We are on the path to improve TS coverage in the app by creating new files in TS and adding JSDocs to existing files

**What this PR explores…**

> ## Replace ESLint with Oxc: JS/TS split (lint + format)
> 
> ### Summary
> 
> Migrate iD from ESLint to [Oxlint](https://oxc.rs/docs/guide/usage/linter) + [Oxfmt](https://oxc.rs/docs/guide/usage/formatter), with a deliberate split between legacy **JavaScript** and new **TypeScript**:
> 
> - **JavaScript** (`.js`, `.mjs`, `.cjs`) — same lint posture as `develop` today; **no** auto-formatting (legacy files keep their existing style).
> - **TypeScript** (`.ts`) — same develop-derived lint base, plus **stricter** TypeScript rules; **Oxfmt** for 4-space, single-quote, semicolon formatting.
> 
> Rules and file-type behavior live in config files, not in npm script filters. CI and local workflows use the same commands.
> 
> ---
> 
> ### Motivation
> 
> - **Incremental TS migration** without reformatting ~300+ legacy JS modules.
> - **One toolchain** (Oxc) instead of ESLint + style rules that duplicated what a formatter does.
> - **CI parity with local** — contributors run the same `npm run lint` / `npm run lint:ci` as GitHub Actions.
> - **No git-diff helper scripts** — no line-scoped policy runners; Oxlint/Oxfmt work per file, with overrides by extension.
> 
> ---
> 
> ### Architecture
> 
> ```mermaid
> flowchart TB
>   subgraph scripts [npm scripts - same roots everywhere]
>     Roots["config/ scripts/ test/spec/ modules/"]
>   end
>   subgraph oxlint [oxlint.config.mjs]
>     Shared["**/*.{js,ts} — develop parity"]
>     TS["**/*.ts — stricter TS rules"]
>     JS["**/*.{js,mjs,cjs} — no-undef"]
>   end
>   subgraph oxfmt [oxfmt.config.mjs]
>     Ignore["ignorePatterns: all .js/.mjs/.cjs"]
>     FmtTS["format .ts only"]
>   end
>   Roots --> oxlint
>   Roots --> oxfmt
>   Shared --> TS
>   Shared --> JS
>   oxfmt --> Ignore --> FmtTS
> ```
> 
> | | **JavaScript** | **TypeScript** |
> |--|--|--|
> | **Formatter** | None | Oxfmt (`oxfmt.config.mjs`) |
> | **Lint base** | Migrated from `develop` `eslint.config.js` | Same shared block |
> | **Lint extra** | `no-undef` | `no-var`, `prefer-const`, `typescript-eslint` recommended, type-aware rules, etc. |
> | **Modern syntax** | Unchanged legacy style | Arrows, `const`/`let`, etc. encouraged where appropriate |
> 
> Override order in `oxlint.config.mjs` matters: shared `**/*.{js,ts}` first, then `**/*.ts` (so TS strict rules are not overridden by shared `no-var: off`).
> 
> ---
> 
> ### Commands
> 
> | Script | Purpose |
> |--------|---------|
> | `npm run lint` | Oxlint on `config/`, `scripts/`, `test/spec/`, `modules/` (also run via `npm run test`) |
> | `npm run lint:fix` | Safe/suggested Oxlint fixes |
> | `npm run format` | Write Oxfmt (TypeScript only) |
> | `npm run format:check` | Check formatting (TypeScript only) |
> | `npm run lint:ci` | `format:check` then `lint` (CI after `npm run test`) |
> 
> All scripts pass `-c oxfmt.config.mjs` / `-c oxlint.config.mjs` so `ignorePatterns` and overrides apply correctly.
> 
> ---
> 
> ### Config files
> 
> - [`oxlint.config.mjs`](oxlint.config.mjs) — migrated from `develop` ESLint via `@oxlint/migrate`; overrides for JS/TS/test/config.
> - [`oxfmt.config.mjs`](oxfmt.config.mjs) — EditorConfig-aligned; `ignorePatterns` exclude all JS variants.
> - [`docs/oxlint-migrate-report.txt`](docs/oxlint-migrate-report.txt) — rules skipped during migration (stylistic ESLint rules → Oxfmt on TS instead).
>    - [ ] Note to self: We need to delete this file before merging…
> 
> ---
> 
> ### CI
> 
> [`.github/workflows/build.yml`](.github/workflows/build.yml): after `npm run test`, run `npm run lint:ci` (full tree, not changed-files-only).
> 
> ---
> 
> ### Editor (VS Code)
> 
> - [Oxc extension](https://marketplace.visualstudio.com/items?itemName=oxc.oxc-vscode) (workspace recommendation).
> - **Format on save:** TypeScript only, `modificationsIfAvailable`.
> - **Lint:** reads `oxlint.config.mjs` per open file (stricter on `.ts`).
> 
> ---
> 
> ### Removed / not in scope
> 
> - ESLint, `eslint.config.js`, Biome, and helper scripts (`ox-changed.mjs`, `changed-lines-policy.mjs`).
> - Line-scoped lint/format on git diff hunks.
> - Enforcing `// @ts-check` or “new files must be `.ts`” in tooling (convention only, if desired in docs).
> - Re-applying ESLint stylistic rules on JS (`quotes`, `semi`, etc.) — JS stays unformatted; oxlint migration already dropped those as lint rules.
> 
> ---
> 
> ### Contributor docs
> 
> Update [`CONTRIBUTING.md`](CONTRIBUTING.md) (JavaScript and TypeScript section) — table above, pointer to config files, editor setup.
> 
> ---
> 
> ### Test plan
> 
> - [ ] `npm run lint` — exit 0 (warnings OK on legacy JS).
> - [ ] `npm run format:check` — only `.ts` files checked (~45 files today).
> - [ ] `npm run lint:ci` — passes after `npm run test`.
> - [ ] Open `.js` + `.ts` in VS Code: save formats TS only.
> - [ ] `var` in a `.ts` file fails lint; `var` in `.js` still allowed (develop behavior).
> - [ ] Arrow functions in `.ts` are allowed (no `func-style` ban).

</p>
</details>


.github/workflows/build.yml
```diff
@@ -36,6 +36,12 @@ jobs:
     - run: npm run all
     - run: npm run test
 
+    - name: Biome check (format + lint + organize imports, changed vs base)
+      if: ${{ github.event_name == 'pull_request' && matrix.node-version.isLatest }}
+      run: |
+        git fetch origin ${{ github.base_ref }} --no-tags --depth=500
+        npx biome check --changed --since=origin/${{ github.base_ref }} --no-errors-on-unmatched
```
tordans, line 43: TODO:
Reading the docs again only one of the two should be needed (`changed` or `since`).
And the docs don't mention the `git fetch` part either.

.vscode/settings.json
```diff
@@ -1,5 +1,23 @@
 {
     "i18n-ally.readonly": true,
     "i18n-ally.localesPaths": ["data"],
-    "i18n-ally.pathMatcher": "core.yaml"
+    "i18n-ally.pathMatcher": "core.yaml",
+    "editor.defaultFormatter": "biomejs.biome",
+    "editor.formatOnSave": true,
+    "editor.formatOnSaveMode": "modificationsIfAvailable",
```
tordans, line 7: Docs: https://code.visualstudio.com/updates/v1_49#_only-format-modified-text

biome.json
```diff
@@ -0,0 +1,413 @@
...
+        "defaultBranch": "develop"
+    },
+    "files": { "includes": ["**", "!!**/dist"] },
+    "formatter": {
+        "enabled": true,
+        "indentStyle": "space",
+        "indentWidth": 4,
+        "lineEnding": "lf"
+    },
+    "linter": {
```
tordans, line 16: This was migrated with the biome migration script. 
TODO: We need to review the rules and compare them to what we have now.

biome.json
```diff
@@ -0,0 +1,413 @@
...
+    },
+    "javascript": {
+        "formatter": {
+            "trailingCommas": "all",
+            "semicolons": "always",
+            "quoteStyle": "single"
+        },
+        "globals": []
+    },
+    "json": { "formatter": { "trailingCommas": "none" } },
```
tordans, line 93:
```suggestion
```

Don't think we need this; Double check…

CONTRIBUTING.md
```diff
@@ -251,19 +251,23 @@ We like when people get involved! iD is a busy project, so it helps if you first
...
+- **`npm run lint`** — Biome linter on `config/`, `scripts/`, `test/spec/`, and `modules/`. Must pass with **no errors** (warnings do not fail the build).
+- **`npm run lint:fix`** — applies lint fixes Biome can perform (including unsafe fixes).
+- **Pull requests** — CI runs Biome **format check** and **lint** only on files that differ from the PR base branch.
+- **Optional before commit:** `npm run biome:verify:staged` (format + lint staged files).
+- **Optional vs `develop`:** after `git fetch origin develop`, run `npm run biome:verify:changed`.
 
-JavaScript code should pass through [ESLint](http://eslint.org/) with no warnings.
+Install the [Biome VS Code extension](https://marketplace.visualstudio.com/items?itemName=biomejs.biome) (workspace recommendation). With the default workspace settings, format-on-save applies to modified lines when SCM context is available.
+
+Settings live in [`biome.json`](biome.json). More detail: [docs/changed-code-formatting-evaluation.md](docs/changed-code-formatting-evaluation.md).
```
tordans, line 270:
```suggestion
Settings live in [`biome.json`](biome.json).
```

package.json
```diff
@@ -1,150 +1,153 @@
 {
-  "name": "@openstreetmap/id",
```
tordans, line 2: TODO: We likely don't want to reformat files like this…

tordans (commented):
The biome.json Config that was generated is a mess; TODO: Strip this down to a clean minimum.

tyrasd:
> Selective linting and formatting only on changes

AFAICS this is not what the proposed solution would do, does it?

> --changed — When set to true, only the **files** that have been changed […]

tordans:
@tyrasd @k-yle this is a new take on this idea. Now that we are on a path to TS, we can make the split cleaner:
- keep all JS as is (linting as is; no formatting)
- but apply stricter rules on TS files (stricter linting, autoformatting)

I updated this PR accordingly and migrated the setup to the new oxlint, oxfmt toolchain that is supported by the Vite-Ecosystem and a lot nicer to work with than Prettier + Eslint and moving faster than Biome.

This PR does not include the reformatting that is needed to align the TS files but given those are only a few files, I think now is a good time to do this to keep the changes small.

tordans:
@tyrasd @k-yle this PR is updated. We discussed [on the community chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Notes_2) to introduce an allow list for files that follow strong lint and formatting rules and keep and avoid changes to any other parts of the system.

This aligns with our goal to keep those changes well contained.

This branch is a take on this idea:
- For legacy code nothing changes
- But we use new tooling (the ox* eco system)
- For TS files we have an allow list
- This PR picked recent TS files from Kyle and Martin and added them to this allow list (separate commit, blame-ignored in the next commit)

[The workflow](https://github.com/openstreetmap/iD/pull/12038/changes#diff-50bd490001029fe2ff769a2c4d6bfb8bee78dbc8ece460fd2e18882b624d5684R4-R8) is:
- New TS files => add to allow list right away and the first commit is clean (formatting, linting)
- TS migrations => first do the migration, later do the formatting and blame-ignore those commits

I see @tyrasd already put this topic on the agenda for [the next community chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_August_5,_12:00_-_13:00_(UTC)) so I hope this can help guide our decisions.

tordans:
@tyrasd in the community chat we talked about which settings we want for the autoformatted and linted TS files. 

I asked the LLM to look into the changes and explain where they come from and what we can do to change the settings. Below is this text.

I pushed one change to keep the D3-Chains in separate lines with `ignore`-comments since the text below says this is something that cannot be really adjusted with either Prettier or Oxlint. And I think it is either to read in most cases to keep the separate lines.

> # Analysis: allowlisted TypeScript reformatting (`d969fe0d8`)
> 
> Scope: commit **`d969fe0d8`** (*Reformat allowlisted TypeScript*) — Oxfmt + allowlist-scoped Oxlint autofixes on files in `oxc.allowlist.mjs` (67 files, ~+3583/−2871). The following commit only updates `.git-blame-ignore-revs`.
> 
> Tooling in play:
> 
> | Concern | Tool | Config |
> | --- | --- | --- |
> | Layout / whitespace / line breaks | **Oxfmt** (formatter, Prettier-compatible) | `oxfmt.config.mjs` |
> | Semantics / modern syntax (`var`→`const`, etc.) | **Oxlint** (linter, with `--fix`) | `oxlint.config.mjs` allowlist override |
> 
> The valuable modernization is the **linter** half (`prefer-const` / `no-var`). The painful D3 diffs are almost entirely the **formatter** half; Oxfmt (like Prettier) has no “preserve one `.foo()` per line” mode.
> 
> ---
> 
> ## 1. Change inventory (what the commit actually did)
> 
> Each block ends with the same headings:
> 
> - **Who** — formatter (Oxfmt) vs linter (Oxlint)
> - **Configurable** — can we restore / steer this via config while staying on Oxfmt/Oxlint?
> - **Alternative** — how to get the old look back (or a better modern equivalent), if we care
> - **Verdict**
> 
> ### A. Good / keep — modern binding (`prefer-const` / `no-var`)
> 
> - **~773** removed `var` usages; many became `const`, rest `let` when reassigned.
> - Aligns with modern JS and with CONTRIBUTING (“mostly Airbnb / Modern ES6”).
> 
> #### Who
> 
> **Linter** (Oxlint) — allowlist rules `prefer-const: error`, `no-var: error`.
> 
> #### Configurable
> 
> **Yes** — toggle those rules in `oxlint.config.mjs` (but we should keep them on).
> 
> #### Alternative
> 
> N/A — do not roll back; this is the valuable half of the commit.
> 
> #### Verdict
> 
> Keep.
> 
> ### B. `function(` → `function (` (space after `function`)
> 
> - **~282** anonymous/`function` expressions gained a space: `function(d)` → `function (d)`, `function()` → `function ()`.
> - Named declarations stay `function foo()` (no space between name and `(`) — that part did not change.
> 
> **Example:**
> 
> ```ts
> // before
> .map(function(n) { return n.loc; });
> // after
> .map(function (n) {
>     return n.loc;
> });
> ```
> 
> #### Who
> 
> **Formatter** (Oxfmt / Prettier convention).
> 
> #### Configurable
> 
> **No** — Oxfmt has no `spaceBeforeFunctionParen` (or equivalent). Staying on Oxfmt means accepting `function (`.
> 
> #### Alternative
> 
> - **Old look:** `// oxfmt-ignore` / `// prettier-ignore` per expression (does not scale); or leave Oxfmt for a formatter with stylistic paren control (breaks repo alignment).
> - **Better modern path:** prefer **arrow functions where feasible** via Oxlint `prefer-arrow-callback` (defaults) so pure callbacks become `(n) => …` instead of `function (n)`. Callbacks that reference `this` stay as `function` automatically (see D).
> 
> #### Verdict
> 
> Style noise; accept, or migrate anonymous callbacks to arrows where `this` is unused.
> 
> ### C. D3-style method chains collapsed onto previous / one line
> 
> - **~167** removed leading-dot chain lines across allowlisted TS; **~85** of those in `modules/svg/*` alone.
> - Short chains that fit under `printWidth` (Oxfmt default **100**; our config does **not** set it) are inlined.
> 
> **Before (classic iD / D3):**
> 
> ```ts
> layer
>     .transition()
>     .duration(250)
>     .style('opacity', 0);
> ```
> 
> **After (Oxfmt):**
> 
> ```ts
> layer.transition().duration(250).style('opacity', 0);
> ```
> 
> Same for `selectAll…data`, `enter().append()`, `exit().remove()`, `append().attr().attr()`, etc. Longer chains / multi-line callbacks still break — often differently than the old hand style.
> 
> There is no mainstream “D3 chain mode” in Prettier/Oxfmt — and **no dedicated Oxfmt issue** that asks for “always one `.foo()` per line” the way Prettier’s long-running requests do:
> 
> - Prettier: [#1565](https://github.com/prettier/prettier/issues/1565) (break chained calls), [#7884](https://github.com/prettier/prettier/issues/7884) (do not collapse existing chain breaks)
> - Oxc / Oxfmt (closest, not the same ask):
>   - [discussion #13608](https://github.com/oxc-project/oxc/discussions/13608) (RFC — `printWidth` reflow called out as painful for long method chains; comment asking to configure chains split across lines)
>   - [#19266](https://github.com/oxc-project/oxc/issues/19266) (open — Oxfmt keeps method chains tight under `printWidth`, same Prettier-style collapse)
>   - Other open Oxfmt “member chain” issues ([#19468](https://github.com/oxc-project/oxc/issues/19468), [#23852](https://github.com/oxc-project/oxc/issues/23852)) are Prettier-parity / bugfixes, not a style option for D3
> 
> #### Who
> 
> **Formatter** (Oxfmt `printWidth` heuristic).
> 
> #### Configurable
> 
> **No** for “always one `.foo()` per line”.
> 
> `printWidth` only soft-influences:
> 
> - lower (e.g. 80) → more breaks + global churn, still not true one-call-per-line
> - higher → collapsing gets worse
> 
> No chain-break option today (Prettier-parity).
> 
> #### Alternative
> 
> - **Old look, selective:** `// oxfmt-ignore` above a chain statement ([docs](https://oxc.rs/docs/guide/usage/formatter/ignore-comments.html)) — works, but comment clutter in hot SVG paths.
> - **Old look, by path:** exclude `modules/svg/**` (and other D3-heavy paths) from Oxfmt — today lint + format share `oxc.allowlist.mjs`, so this means splitting a **lint allowlist** vs **format allowlist**.
> - **Old look, everywhere:** drop Oxfmt for TS (keep Oxlint `prefer-const`) — restores hand style, loses the CONTRIBUTING format story.
> 
> `oxfmt.config.mjs` only soft-helps via `printWidth`; it cannot restore classic D3 chaining.
> 
> #### Verdict
> 
> Main readability regression for SVG/UI D3 code.
> 
> ### D. Arrow functions — mostly *not* introduced; `function` kept
> 
> - The commit did **not** mass-convert callbacks to arrows.
> - Existing arrows got Oxfmt `arrowParens: "always"` (default): `n => …` → `(n) => …`.
> - Inline `function (d) { return … }` often became multi-line blocks — still `function`, not `=>`.
> 
> #### Why D3 cares about `function` vs `=>`
> 
> D3 selection APIs (`selection.each`, many `.attr` / `.classed` / `.style` / `.on` callbacks, enter/update patterns) **call your callback with `this` bound to the DOM element**. That only works with a classic `function`:
> 
> ```ts
> .each(function (d) {
>     d3_select(this).call(d.layer); // `this` === the <g> / node D3 is visiting
> });
> ```
> 
> An arrow closes over lexical `this` (usually wrong in module scope), so the same code breaks:
> 
> ```ts
> .each((d) => {
>     d3_select(this); // wrong — not the DOM node
> });
> ```
> 
> So: **we do not need `function` for every D3 call** — only when the callback relies on D3’s `this` (or on `arguments`, etc.). Pure data callbacks (`return d.id`, filters, key functions) **can** be arrows with no behavior change. In iD both kinds exist in `modules/svg/**` and `modules/ui/**` (`d3_select(this)`, `this.parentNode.__data__`, … alongside simple `function (d) { return d; }`).
> 
> #### Does `prefer-arrow-callback` handle D3 `this` automatically?
> 
> **Yes, with defaults — you do not special-case each D3 callback.**
> 
> Oxlint implements ESLint’s [`prefer-arrow-callback`](https://eslint.org/docs/latest/rules/prefer-arrow-callback) with the same defaults (`allowNamedFunctions: false`, `allowUnboundThis: true`):
> 
> - Callback body **references `this`** → rule **leaves `function` alone** (arrow would change binding). Normal iD patterns like `d3_select(this)` / `this.parentNode` are covered; no per-call `oxlint-disable` needed.
> - Callback body **does not** use `this` / `arguments` / `super` / `new.target` and isn’t a generator → rule flags (and can autofix) to an arrow.
> 
> So enabling it globally (including `modules/svg/**`) is a workable “two concepts” policy: arrows where safe, `function` kept where `this` is used.
> 
> **Do not** set `allowUnboundThis: false` — that flags *all* `function` callbacks, including ones that need D3’s `this`.
> 
> Edge case only: if a callback somehow relied on D3’s `this` **without mentioning `this` in the body**, the rule would not see it (rare; not how iD’s SVG code looks).
> 
> #### One style everywhere vs two concepts
> 
> iD is D3-heavy, so “always arrows (modern)” and “always `function` (one safe style)” both look attractive. Pros/cons:
> 
> | Policy | Pros | Cons |
> | --- | --- | --- |
> | **One style: `function` everywhere** (incl. actions/util `map`/`filter`) | Single rule; no “does this need `this`?”; matches much legacy iD/D3 code; safer when copy-pasting SVG patterns | Feels dated vs modern TS; loses short `(d) => d.id`; fights common lint defaults; does **not** remove Oxfmt’s `function (` spacing (B) |
> | **One style: arrows everywhere** (naive / rewrite-off-`this`) | Matches generic modern JS; shorter callbacks; shrinks B’s `function (` noise | **Unsafe if you rewrite blindly** — breaks D3 `this` unless migrated (node argument / `selection.nodes()`). High cost; silent bugs easy |
> | **Two concepts via `prefer-arrow-callback` (defaults)** | Linter encodes the split: arrows when safe, `function` when `this` is referenced; works in svg/ui without path carve-outs | Codebase still shows both forms; reviewers should know *why*; don’t flip `allowUnboundThis` to `false` |
> 
> **If the team wants only one *authoring* style:** given D3 volume, standardizing on **`function` for callbacks in D3-adjacent code** (and skipping the rule) is the safer controversial choice — sameness over greenfield looks.
> 
> **Less controversial:** enable Oxlint `prefer-arrow-callback` with **defaults** (including on svg/ui). That is “one rule, two forms” enforced by the linter, not by hand-auditing every `.each`.
> 
> #### Who
> 
> **Neither converted** in this commit.
> 
> - Spacing / `arrowParens` → **Formatter**
> - Choosing arrow vs `function` → **Linter** policy (not enabled today; formatter never converts)
> 
> #### Configurable
> 
> - **Formatter:** `arrowParens: "always" | "avoid"` only (parens on sole params), not arrow-vs-function.
> - **Linter:** yes — Oxlint `prefer-arrow-callback` (keep `allowUnboundThis: true`) / optionally `func-style`.
> 
> #### Alternative
> 
> - **One D3-safe style:** skip `prefer-arrow-callback`; optionally document “prefer `function` callbacks in svg/ui (and maybe everywhere).”
> - **Two concepts (recommended if preferring arrows):** enable `prefer-arrow-callback` with defaults **globally** — it keeps `function` wherever `this` is used; no svg/ui allowlist carve-out required for that reason.
> - **Do not** set `allowUnboundThis: false`, and do not manually rewrite D3 `this` callbacks to arrows.
> 
> #### Verdict
> 
> Leaving `function` alone in the reformat was correct. For house style: either keep `function`-first, or turn on `prefer-arrow-callback` (defaults) and let it skip `this` callbacks automatically — both beat a blind “arrows everywhere” rewrite.
> 
> ### E. Other formatter / lint noise (secondary)
> 
> #### Trailing commas
> 
> - **Who:** Formatter (`trailingComma: 'all'`)
> - **Configurable:** Yes — `'all' | 'es5' | 'none'`
> - **Alternative:** Leave `'all'` (modern TS default)
> 
> #### `printWidth` wraps (args, types, ternaries)
> 
> - **Who:** Formatter
> - **Configurable:** Yes — set `printWidth` (default 100)
> - **Alternative:** Tune width; expect a reformat + blame-ignore commit
> 
> #### `if (…) return` → braces
> 
> - **Who:** Linter `curly: ['warn', 'multi-line']` (+ format)
> - **Configurable:** Yes — change/disable `curly`
> - **Alternative:** Keep; minor comment-placement quirks (`} // comment`)
> 
> #### Binary op spacing (`1+x` → `1 + x`)
> 
> - **Who:** Formatter
> - **Configurable:** No
> - **Alternative:** Accept
> 
> #### Blank-line / comment reflow
> 
> - **Who:** Formatter
> - **Configurable:** No (ignore comments only)
> - **Alternative:** Accept
> 
> ---
> 
> ## 2. What latest modern JS / tooling usually recommends
> 
> | Topic | Modern default (Prettier / Oxfmt / Airbnb-ish) | Fit for iD |
> | --- | --- | --- |
> | `const` / `let`, no `var` | Strong yes | **Keep** (already on allowlist) |
> | Trailing commas, single quotes, semis | Prettier/Oxfmt defaults (we already set these) | Fine |
> | Space in `function (` | Prettier always; Airbnb historically mixed named vs anonymous | Accept as formatter tax, or migrate to arrows where `this` unused |
> | Arrow vs `function` | Prefer arrows for *simple* callbacks **when `this` is unused**; keep `function` when `this` / generators / hoisting matter | For iD see **§D**: `prefer-arrow-callback` (defaults) does this split automatically; or stay `function`-first |
> | Method chains | Fit to `printWidth`; collapse short chains | Conflicts with D3 one-call-per-line house style |
> | Explicit `(x) =>` | `arrowParens: "always"` (Oxfmt default) | Harmless |
>

tyrasd:
> `// oxfmt-ignore`

This seems to be a dealbreaker to me, as our (particularly UI) code is using this formatting very extensively, and having lots of `// oxfmt-ignore`s all over the place seems to be super irritating. Is it not possible to just disable the respective rule completely to avoid this? Otherwise it would probably be better to not use _oxfmt_ and reduce the scope of this PR to the migration of the linter from `eslint` to `oxlint` mainly for the performance improvements benefits.

tordans:
> > `// oxfmt-ignore`
> 
> This seems to be a dealbreaker to me, as our (particularly UI) code is using this formatting very extensively, and having lots of `// oxfmt-ignore`s all over the place seems to be super irritating. Is it not possible to just disable the respective rule completely to avoid this? Otherwise it would probably be better to not use _oxfmt_ and reduce the scope of this PR to the migration of the linter from `eslint` to `oxlint` mainly for the performance improvements benefits.

@tyrasd the formatting tools only allow very minimal configs; this is not one of them.
There are other options which I asked the LLM to research below. Personally I would try oxlint + dprint which is less know but used by a few large projects (Deno Format for example). And else go for an eslint+eslint-stylistic-Setup. No autoformatter is IMO a non-ideal option…

> # Formatter options without D3 chain collapse
> 
> `// oxfmt-ignore` does not scale, and **there is no Oxfmt/Prettier switch to disable chain collapse**. Below: what is actually configurable, what tickets exist, and alternatives that keep classic D3 `foo.bar().baz()` line breaks — especially **lint-only style (C)**, which is the most we can enforce without a `printWidth` printer.
> 
> ## What is already settled
> 
> Layout of `foo.bar().baz()` is owned by a **printer** (Oxfmt/Prettier/Biome), not by a toggleable “rule”.
> 
> - **Oxfmt:** no `spaceBeforeFunctionParen`, no chain-break option, **no custom JS printer plugins** ([unsupported features](https://oxc.rs/docs/guide/usage/formatter/unsupported-features.html)). Oxfmt’s job is Prettier parity, including collapsing short chains under `printWidth` (default 100; we do not set it in [oxfmt.config.mjs](oxfmt.config.mjs)).
> - **Prettier:** same printer model. Long-standing WontFix/open: [#1565](https://github.com/prettier/prettier/issues/1565), [#7884](https://github.com/prettier/prettier/issues/7884). Community Prettier plugins cannot reasonably override the JS member-chain printer.
> - **Biome:** same philosophy; no “always break chains” option ([discussion #7914](https://github.com/biomejs/biome/discussions/7914)).
> - **Oxc tickets already checked** (closest, not a D3 mode):
>   - [discussion #13608](https://github.com/oxc-project/oxc/discussions/13608) — RFC; `printWidth` reflow called out as painful for long method chains; a comment asking to split chains across lines. Oxfmt shipped as **faster Prettier**, not as dprint-style “preserve newlines”.
>   - [#19266](https://github.com/oxc-project/oxc/issues/19266) — open; tight chains under `printWidth`.
>   - [#19468](https://github.com/oxc-project/oxc/issues/19468), [#23852](https://github.com/oxc-project/oxc/issues/23852) — Prettier-parity / bugfixes, not a style option.
> - **`printWidth: 80`** only changes *when* chains wrap. Short D3 chains like `layer.transition().duration(250).style(...)` still collapse.
> 
> So: **cannot keep Oxfmt and also keep classic D3 chains**, except by ignore comments or excluding those files from the formatter.
> 
> Current `// oxfmt-ignore` usage is already in allowlisted SVG/UI: `modules/svg/geolocate.ts`, `defs.ts`, `turns.ts`, `layers.ts`, `midpoints.ts`, `modules/ui/confirm.ts`, `fields/check.ts`, `modal_async.ts`.
> 
> ```js
>         // oxfmt-ignore
>         layer
>             .transition()
>             .duration(250)
>             .style('opacity', 0);
> ```
> 
> ## Option map (A–E)
> 
> ```mermaid
> flowchart TD
>   fork{"Drop Oxfmt because of D3 chains?"}
>   stayEslint["C2: Stay on ESLint 10 + Stylistic"]
>   keepOxlint["Keep Oxlint for speed"]
>   dropFmt["A: Oxlint only, no formatter"]
>   dprintFmt["B: Oxlint + dprint"]
>   c1["C1: Oxlint + Stylistic jsPlugins"]
>   splitAllow["D: Oxfmt except svg/ui"]
>   keepOxfmt["E: Keep Oxfmt, accept collapse"]
> 
>   fork --> stayEslint
>   fork --> keepOxlint
>   keepOxlint --> dropFmt
>   keepOxlint --> dprintFmt
>   keepOxlint --> c1
>   keepOxlint --> splitAllow
>   keepOxlint --> keepOxfmt
> ```
> 
> - **A — Drop Oxfmt, keep Oxlint.** Same idea as “reduce this PR to the ESLint → Oxlint migration.” Aligns with today’s JS policy in CONTRIBUTING (“no autoformatter for JavaScript”). Revert the reformat + `oxfmt-ignore` comments; keep allowlist-scoped `no-var` / `prefer-const`. Lowest conflict *if* we still want the Oxlint speed win; no quotes/EOF enforcement unless we add C1.
> - **B — dprint** (only mainstream *printer* that preserves author line breaks in member chains). See **B in depth** below. Must use `dprint-plugin-typescript`, **not** the Oxc wrapper plugin (that is Oxfmt inside dprint and still collapses chains). Still pairs with Oxlint (or could pair with ESLint; unusual).
> - **C — Lint-only style** (no printer). **C1** = keep Oxlint, load Stylistic via `jsPlugins`. **C2** = **stay on ESLint** (do not migrate to Oxlint) and add `@stylistic/eslint-plugin` on the existing ESLint 10 flat config. If Oxfmt is off the table, C2 is the coherent “one toolchain” answer: we do not need Oxc just for a faster linter.
> - **D — Split allowlists.** Format `modules/actions/**`, `core`, `util`, tests with Oxfmt; exclude `modules/svg/**` and `modules/ui/**`. Incomplete: actions already have multi-line `.map().reduce()` / `.filter().map()` chains ([modules/actions/split.ts](modules/actions/split.ts), `join.ts`, …). Two contributor stories. Does not need ignore comments in UI if UI is simply not formatted.
> - **E — Keep Oxfmt, accept collapse.** Conflicts with extensive D3 one-call-per-line style in UI/SVG. Not viable if `oxfmt-ignore` is unacceptable.
> 
> Custom Oxfmt plugins: **not an option**. Custom Oxlint/ESLint rules: **yes**, and that *is* C.
> 
> ## B in depth — dprint
> 
> Links: [dprint.dev](https://dprint.dev/), [github.com/dprint/dprint](https://github.com/dprint/dprint), [TypeScript plugin docs](https://dprint.dev/plugins/typescript/), [dprint-plugin-typescript](https://github.com/dprint/dprint-plugin-typescript).
> 
> dprint is a **pluggable formatting platform** (Rust CLI + sandboxed Wasm plugins), not a Prettier clone. Official plugins expose real knobs; the TypeScript plugin’s default is to **leave author line breaks in member and binary expressions** because “newlines can convey meaning or grouping.” That is the D3-relevant difference vs Oxfmt.
> 
> For iD we would use **`dprint-plugin-typescript`**. Do **not** use [`dprint-plugin-oxc`](https://github.com/dprint/dprint-plugin-oxc) — that is an adapter that runs **Oxfmt inside dprint**, so it still collapses short chains under `printWidth`. Boshen already said Oxc would offer a dprint Wasm plugin for existing dprint users ([RFC #13608](https://github.com/oxc-project/oxc/discussions/13608)); that wrapper exists and is the wrong plugin for this problem.
> 
> ### History
> 
> - **2019-06:** [dprint/dprint](https://github.com/dprint/dprint) created by [David Sherret](https://github.com/dsherret) (`dsherret`).
> - **2020-01/02:** Deno replaces Prettier with dprint for `deno fmt` ([denoland/deno#3818](https://github.com/denoland/deno/issues/3818), PR #3820). Motivation: Rust speed + more configuration than Prettier, written by Sherret (then on the Deno team).
> - **2020-06:** [dprint-plugin-typescript](https://github.com/dprint/dprint-plugin-typescript) split out; formats JS/TS via the **SWC** parser (not Oxc).
> - **2020–today:** `deno fmt` is still “a built-in code formatter based on dprint” ([Deno docs](https://docs.deno.com/runtime/reference/cli/fmt/)). Deno itself stays opinionated (`gofmt`-like); the standalone CLI is the configurable one.
> - **2023-10:** [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) adopts dprint for all types files ([discussion #65993](https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/65993), [PR #66235](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66235)) — chosen for performance on a huge repo and because it is “essentially `deno fmt`.” Still the DT formatter today.
> - **2026:** CLI still shipping (latest checked: **0.55.2**, 2026-07-14). TypeScript plugin still shipping (e.g. **0.96.1**, 2026-05-20). A draft PR exists to port the TS plugin from SWC to Oxc ([dprint-plugin-typescript#798](https://github.com/dprint/dprint-plugin-typescript/pull/798), June 2026) — parser swap, not an Oxfmt-style printer.
> 
> ### Maintenance status
> 
> - **Bus factor is high.** GitHub contributor counts on `dprint/dprint`: `dsherret` ~1227 commits vs next ~13. README: “I do a lot of this development in my spare time” + [sponsor](https://dprint.dev/sponsor) ask for commercial users. Sherret’s profile is now **Prev @denoland** (no longer a Deno-employed maintainer).
> - **Not abandoned.** Releases through mid-2026 (Wasmtime plugin runtime, npm specifiers, per-file overrides, LSP completions in 0.55.0). TypeScript plugin has hundreds of spec tests and frequent point releases.
> - **Compared with Oxfmt:** Oxfmt is part of **Oxc / VoidZero** — company-backed, same team as Oxlint, one VS Code extension, Prettier-compat as a product goal. dprint is a mature volunteer/platform project with production users (Deno, DefinitelyTyped) but a single primary maintainer. For iD, Oxlint can stay either way; switching the *formatter* to dprint means a second vendor and a second editor extension ([dprint.dprint](https://marketplace.visualstudio.com/items?itemName=dprint.dprint) instead of `oxc.oxc-vscode` for format-on-save).
> 
> ### Stability
> 
> - **Mature CLI, opinionated-elsewhere printer.** Six years in Deno production; DT since 2023. Not alpha in the Oxfmt-alpha sense.
> - **Not Prettier-stable.** Output will **not** match Oxfmt/Prettier. First adoption is a unique reformat (blame-ignore), not a no-diff swap. Config surface is large; some Deno folks have called TS-plugin options “experimental” when asked to expose all of them on `deno fmt` ([deno#19806](https://github.com/denoland/deno/issues/19806)).
> - **TS plugin issue count is high** (~200+ open on `dprint-plugin-typescript`) — expected for a configurable printer, but more edge-case risk than “Prettier parity tests at 100%” (Oxfmt’s claim).
> - **VS Code / CI:** `dprint check` / `dprint fmt` are the equivalents of `oxfmt --check` / `oxfmt`. Incremental cache is on by default. Pin plugin Wasm URLs (or npm `@dprint/typescript`) in config, same as we pin `oxfmt` in package.json.
> - **What we would gain vs Oxfmt for iD:** preserve D3 chains; `functionExpression.spaceAfterFunctionKeyword` (can keep `function(`); `quoteStyle: preferSingle`; indent 4; trailing commas; EOF/LF. `preferSingleLine` must stay **false** (default) or it *would* collapse chains like Oxfmt. `memberExpression.linePerExpression: true` only forces one-call-per-line when the chain is *already* multi-line — still preserve-first, not ESLint-style enforce.
> 
> ### Speed (with reference to Oxfmt)
> 
> Both are “fast enough” for iD’s allowlist (~67 TS files). Speed should not decide this; chain behavior should.
> 
> - **Cold / full-repo JS+TS:** Oxfmt is a **native Rust** printer on the Oxc AST. dprint’s TS plugin is **Rust compiled to Wasm** and run in a sandbox (Wasmer historically; Wasmtime as of dprint 0.55). Oxc’s own RFC answer to “Why not dprint?” was: CLI + Wasm plugins vs “we want to offer a better formatter,” plus they would wrap Oxfmt as a dprint plugin ([#13608](https://github.com/oxc-project/oxc/discussions/13608)). Official Oxfmt marketing: **~30× Prettier, ~3× Biome** on an initial run **without cache** ([Oxfmt beta post](https://oxc.rs/blog/2026-02-24-oxfmt-beta.html)). Those benches do not headline dprint.
> - **Warm / repeat local runs:** dprint has **file-hash incremental caching by default** ([dprint.dev](https://dprint.dev/) “incremental caching”). Oxfmt **does not** as of 2026-04/07 ([oxc#21585](https://github.com/oxc-project/oxc/issues/21585), [discussion #23640](https://github.com/oxc-project/oxc/discussions/23640)). Sherret has shown cases where dprint looks *drastically* faster than Oxfmt for that reason (cache hits, not a faster printer). CI that formats the whole tree every time still favors Oxfmt’s cold path; `format-on-save` / “format dirty files” favors dprint’s cache (dprint 0.55 also added `--dirty`).
> - **Practical for iD:** neither will be the bottleneck next to tests/build. Choosing dprint is a **style/architecture** trade (preserve chains, extra tool, maintainer risk), not a performance win over Oxfmt.
> 
> ### Cost of choosing B
> 
> - Second toolchain beside Oxlint (Oxc extension stays for lint; dprint extension for format).
> - Full reformat of allowlisted TS to dprint’s printer (different from the Oxfmt commit already on the branch — would revert Oxfmt layout, then dprint).
> - Does **not** *force* one `.foo()` per line if someone writes a chain on one line; combine with C’s `newline-per-chained-call` on svg/ui if we want enforcement.
> - Bus-factor / spare-time maintainer vs VoidZero-backed Oxfmt.
> 
> ### When B beats C
> 
> Use dprint if the team still wants a real printer (indent, wrapping of long args, format-on-save that reprints the file) **and** will not accept Oxfmt’s chain collapse. Use C if “quotes / EOF / semis / no printer” is enough and we want zero second formatter.
> 
> ## C in depth — lint-only style instead of a formatter
> 
> ### What Oxlint can do natively (not enough)
> 
> Oxlint’s `style` category is **idiomatic JS**, not a formatter. Native allowlist rules already cover the valuable half: `prefer-const`, `no-var`, `curly`. There are **no** native Oxlint rules for `quotes`, `semi`, `eol-last`, `indent`, or `newline-per-chained-call`. Oxc’s position: layout belongs in Oxfmt; leftover style comes from ESLint Stylistic via JS plugins ([oxc#11435](https://github.com/oxc-project/oxc/issues/11435), [JS plugins docs](https://oxc.rs/docs/guide/usage/linter/js-plugins.html)).
> 
> So “use Oxlint instead of Oxfmt” only gets semantics unless we add `@stylistic/eslint-plugin`.
> 
> ### How to wire Stylistic (C1 vs C2)
> 
> Same **safe rule subset** either way. Do **not** enable `stylistic.configs.recommended` / `customize()` wholesale — those turn on `indent` and become a second pretty-printer.
> 
> - **C1 — Keep Oxlint; Stylistic as jsPlugin:** `jsPlugins: ['@stylistic/eslint-plugin']` and `oxlint --fix`. One CLI, Oxc editor extension, jsPlugins still **alpha**. This only makes sense if we still want Oxlint’s speed for the existing develop-parity rules.
> - **C2 — Stay on ESLint; add Stylistic there:** restore the ESLint 10 + `typescript-eslint` stack this PR removed, and add `@stylistic/eslint-plugin`. One ecosystem, no Oxlint, no Oxfmt. Details below.
> 
> Do not confuse `@stylistic/eslint-plugin` with `typescript-eslint`’s `configs.stylistic` (that preset is semantic TS style like `array-type`, not quotes/EOF).
> 
> ### C2 — Stay with ESLint (do not migrate to Oxlint)
> 
> If we drop Oxfmt, **Oxlint is optional**. This PR’s pitch was Oxc as a pair (faster lint + faster Prettier-compatible format). Without the formatter, the remaining win is lint speed. C2 takes the other fork: **keep ESLint as the linter**, and put the D3-safe style rules in the same tool.
> 
> **What was already in this repo:** the pre-Oxlint tree was **already** newest-era ESLint, not an `.eslintrc` leftover.
> 
> - Commit `4dce8ba66` (*Replace ESLint with Oxlint and Oxfmt…*) deleted [eslint.config.js](eslint.config.js). Parent already had:
>   - `eslint` `^10.0.0`, `@eslint/js` `^10.0.1`, `typescript-eslint` `^8.56.0`
>   - Flat config via `tseslint.config(...)` + `js.configs.recommended` + `tseslint.configs.recommended`
>   - A handful of **deprecated core** style rules (`linebreak-style: unix`, `keyword-spacing`, `block-spacing`, `brace-style`, `arrow-spacing`, `func-call-spacing`) and **`indent: off`**
>   - **No** `@stylistic/eslint-plugin`
> - CHANGELOG already notes the earlier “Upgrade eslint config to version 9” step; v10 was the last ESLint on `develop` before Oxlint.
> 
> So C2 is **not** an eslintrc → flat migration, and **not** “ESLint only for quotes beside Oxlint.” It is: **revert the Oxlint/Oxfmt swap**, keep `eslint.config.js` (or rewrite it with `defineConfig` from `eslint/config`), and **add Stylistic** for the layout rules ESLint itself deprecated ([Stylistic migration](https://eslint.style/guide/migration)).
> 
> **Packages (current, 2026):**
> 
> - `eslint@^10` — current major ([v10.0.0, Feb 2026](https://eslint.org/blog/2026/02/eslint-v10.0.0-released/)). eslintrc is gone; Node `>=20.19` (iD already `>=22`). Optional nicety: `import { defineConfig } from 'eslint/config'` ([docs](https://eslint.org/docs/latest/use/configure/configuration-files)) instead of `tseslint.config(...)`; both are valid on v10.
> - `@stylistic/eslint-plugin@^5.10.0` — current **stable** (`latest`); peer `eslint ^9 || ^10`. **Do not** take `6.0.0-beta.*` for iD unless we explicitly want breaking quote/option defaults ([v6 upgrade](https://eslint.style/guide/upgrade)).
> - Keep `@eslint/js` + `typescript-eslint` as they were (recommended / TS parser / vitest plugin). Move deprecated core style rule names (`keyword-spacing`, `linebreak-style`, `brace-style`, …) to `@stylistic/*` equivalents. Add the rest of the **safe subset** (`quotes`, `semi`, `eol-last`, `comma-dangle`, …). Leave `indent` off.
> 
> **Config shape** (restore + extend, not a style-only sliver):
> 
> ```js
> import { defineConfig } from 'eslint/config';
> import js from '@eslint/js';
> import stylistic from '@stylistic/eslint-plugin';
> import tseslint from 'typescript-eslint';
> // …globals, vitest, as in the deleted eslint.config.js
> 
> export default defineConfig([
>   js.configs.recommended,
>   ...tseslint.configs.recommended,
>   {
>     files: ['**/*.{js,ts}'],
>     plugins: { '@stylistic': stylistic },
>     rules: {
>       '@stylistic/quotes': ['error', 'single', { avoidEscape: true }],
>       '@stylistic/semi': ['error', 'always'],
>       '@stylistic/eol-last': ['error', 'always'],
>       '@stylistic/linebreak-style': ['error', 'unix'],
>       // …safe subset; no @stylistic/indent, no stylistic.configs.recommended
>     },
>   },
> ]);
> ```
> 
> Allowlist-stricter `no-var` / `prefer-const` can stay as ESLint overrides on the same paths [oxc.allowlist.mjs](oxc.allowlist.mjs) used, without Oxlint.
> 
> **Scripts / CI:** `npm run lint` / `lint:fix` back to `eslint` (as before `4dce8ba66`). No `oxlint`, no `oxfmt`, no Oxc VS Code formatter. ESLint extension `--fix` on save.
> 
> **What we give up vs A/C1:** Oxlint’s speed (the original reason to migrate the linter). For iD that is real on whole-repo `modules/` + `test/spec`, but it is a *choice*, not a requirement of dropping Oxfmt.
> 
> **What we keep vs this PR as-is:** one tool contributors already knew; Stylistic’s native `--fix`; no jsPlugins alpha; D3 chains never go through a printWidth printer.
> 
> **Why C2 rather than “Oxlint + ESLint style-only”:** two linters is the worst of both worlds (ESLint deps *and* Oxc, two configs, two editor extensions). If Oxfmt is gone, pick **either** Oxlint (A or C1) **or** ESLint (C2).
> 
> ### Safe subset (max without the D3 issue)
> 
> These autofix **punctuation/whitespace in place**. They do not reprint a statement to fit `printWidth`, so they will not turn
> 
> `layer\n    .transition()\n    .duration(250)`
> 
> into `layer.transition().duration(250)`.
> 
> - `quotes`: `'single'` + `avoidEscape: true` (matches [oxfmt.config.mjs](oxfmt.config.mjs) `singleQuote: true`)
> - `semi`: `'always'`
> - `eol-last`: `'always'` (final newline)
> - `linebreak-style`: `'unix'`
> - `comma-dangle`: `'always-multiline'` (trailing commas only when already multi-line — will **not** wrap a one-line array just to add a comma)
> - `no-trailing-spaces`
> - `no-multiple-empty-lines` (e.g. `max: 2`, `maxEOF: 0`)
> - `comma-spacing`, `keyword-spacing`, `space-infix-ops` (`1+x` → `1 + x`), `object-curly-spacing`, `array-bracket-spacing`
> - `space-before-function-paren`: can **restore** `function(` vs Oxfmt’s forced `function (` (anonymous vs named is configurable)
> 
> Already covered without a printer: `curly`, `prefer-const`, `no-var` (Oxlint in A/C1; ESLint overrides in C2).
> 
> Editor: turn off Oxfmt format-on-save in [.vscode/settings.json](.vscode/settings.json). C1: Oxc `--fix` on save. C2: ESLint `--fix` on save (correctness + Stylistic). CONTRIBUTING: TS is lint-fixed, not pretty-printed (same idea as JS).
> 
> ### Optional: *enforce* D3 chains (still not Oxfmt)
> 
> [`@stylistic/newline-per-chained-call`](https://eslint.style/rules/newline-per-chained-call) is the opposite of Oxfmt: it **inserts** newlines after calls. The rule’s own docs use D3 as the motivating example.
> 
> - Default `ignoreChainWithDepth: 2` allows short `a.b()` / two-call chains; longer chains must break.
> - This does **not** collapse existing breaks.
> - It **will** expand existing one-liners (`selection.selectAll('g').data(x).enter()` → one call per line). That is a house-style *enforcement* choice, not “preserve what the author wrote”.
> - Scope it to `modules/svg/**` + `modules/ui/**` if we do not want promise/array chains in `modules/actions/**` rewritten.
> 
> This is the only realistic way to *require* classic D3 layout in CI without a D3-aware printer.
> 
> ### Do not enable (these reflow or fight D3)
> 
> - `indent` / `indent-binary-ops` — closest thing to a formatter; slow; historically buggy on TS; huge diffs; still does not give Prettier-quality wrapping. Skip unless we accept “ESLint as formatter”.
> - `max-len` — wants wrapping; same class of pain as `printWidth`.
> - `function-paren-newline`, `function-call-argument-newline`, `object-property-newline`, `array-element-newline`, `operator-linebreak`, `implicit-arrow-linebreak` — line-break printers in disguise.
> - Stylistic `recommended` preset (bundles indent + many of the above).
> 
> ### What C will never do (honest ceiling)
> 
> Without a printer we will **not** get: wrapping long argument lists / types / ternaries; consistent indent of whole files; comment reflow; Oxfmt’s `function (` spacing (unless we *want* that via `space-before-function-paren`); one command that makes every file look “pretty”.
> 
> **Max without the given issue:** the safe Stylistic subset + optionally `newline-per-chained-call` on svg/ui, whether that runs on **ESLint (C2)** or **Oxlint jsPlugins (C1)**. Quotes, semis, EOF, LF, trailing commas on multi-line, operator spacing, no trailing whitespace — D3 chains left alone (or forced *more* broken, never collapsed). C2 does that without introducing Oxc at all.
> 
> ## C + D?
> 
> Usually **worse than C alone**.
> 
> - D still runs Oxfmt on actions/core/util → those files still lose short chain breaks (`arr.map().filter()`), and contributors learn two format stories.
> - C on svg/ui *plus* Oxfmt elsewhere means two engines (quotes/indent/EOF may not match) and a split [oxc.allowlist.mjs](oxc.allowlist.mjs) (lint allowlist vs format allowlist).
> - D is only attractive if we **keep Oxfmt** for non-UI files and refuse Stylistic. If we are already doing C, drop Oxfmt: then pick **C2 (stay ESLint)** or **A/C1 (keep Oxlint)**, not both.
> 
> Conditional D: only if we want a real printer on “boring” TS **and** are willing to accept some chain collapse there.
> 
> ## What C1 / C2 would change in the repo
> 
> - Revert the Oxfmt reformat + all `// oxfmt-ignore`.
> - **C1:** keep Oxlint; add `@stylistic/eslint-plugin` via `jsPlugins`; drop Oxfmt from CI.
> - **C2:** revert Oxlint/Oxfmt (restore `eslint.config.js`, `eslint` / `typescript-eslint` deps, ESLint editor integration); add `@stylistic/eslint-plugin@^5`; rename deprecated core style rules to `@stylistic/*`; add the safe subset; leave `indent` off.
> - Either way: `--fix` on a couple of svg files should leave D3 chains intact.
> - `newline-per-chained-call` is a separate, explicit follow-up (it expands one-liners).
>

matkoniecz:
> cannot keep Oxfmt and also keep classic D3 chains

is this PR making sense without Oxfmt?

If yes, can you change it to be without Oxfmt?

tordans:
> > cannot keep Oxfmt and also keep classic D3 chains
> 
> 
> 
> is this PR making sense without Oxfmt?
> 
> 
> 
> If yes, can you change it to be without Oxfmt?

In my opinion, the last two comments show that we are still discussing how we want to set up the tooling. In my [last comment](https://github.com/openstreetmap/iD/pull/12038#issuecomment-5327666212), I posted suggestions and I'm waiting for input on those to update the PR.

matkoniecz:
> There are other options which I asked the LLM to research below. Personally I would try oxlint + dprint which is less know but used by a few large projects (Deno Format for example). And else go for an eslint+eslint-stylistic-Setup.

have you looked into viability of oxlint + dprint or is it (for now) only LLM-checked?

tordans:
> > There are other options which I asked the LLM to research below. Personally I would try oxlint + dprint which is less know but used by a few large projects (Deno Format for example). And else go for an eslint+eslint-stylistic-Setup.
> 
> have you looked into viability of oxlint + dprint or is it (for now) only LLM-checked?

What makes you assume that the research is invalid? No, I did not invest more time into this, yet. I am happy to – just like I did a couple of times before – but I do need Martin's signal that another option is something that he will consider.

matkoniecz:
> What makes you assume that the research is invalid?

I would not say invalid, rather that it is unreviewed (and quite verbose)

PR #12040, opened 2026-03-18
Add foot=yes assumption on highway=living_street

### LINKED ISSUE #12035: highway=living_street displays "not specified" for foot access in access panel
Image

it can show yes or designated by default, right?

matkoniecz:
fixes #12035

may be argued that `designated` is better, at least in some areas but `yes` is definitely better than "Not specified" that strongly encourages overtagging

other modes are sadly less obvious

### DIFF
--- a/modules/ui/fields/access.js
+++ b/modules/ui/fields/access.js
@@ -132,6 +132,9 @@ export function uiFieldAccess(field, context) {
                 foot: 'yes',
                 motor_vehicle: 'no'
             },
+            living_street: {
+                foot: 'yes'
+            },
             cycleway: {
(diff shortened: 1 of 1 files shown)


matkoniecz:
screen-2026-03-18-13-08-59


even if `yes` for foot access is wrong, then it is less wrong than "Not Specified".

matkoniecz:
Started https://community.openstreetmap.org/t/is-transit-really-banned-for-cyclists-and-others-on-highway-living-street/143153/ - if cleared it could enable safe worldwide defaulting to `yes` also for cyclists

PR #12041, opened 2026-03-18
refactor dark mode switcher, to simplify the css compilation process 

k-yle:
this PR removes a compile-time script for duplicating CSS, and replaces it with a few lines of JS.

This simplifies the build process, which would make it easier to switch to a faster bundler in the future, as discussed on slack recently.

tested with:
 - `?theme=light`
 - `?theme=dark`
 - no `theme` parameter, and using the devtools to override the theme:

image

&nbsp;

cc @hlfan

### DIFF
--- a/css/10_colors.css
+++ b/css/10_colors.css
@@ -65,8 +65,7 @@
     --error-icon-color: #f00;
 }
 
-@media (prefers-color-scheme: dark) {
-    .ideditor {
+.ideditor.theme-dark {
         color-scheme: dark;
         --bg-color: #212529;
         --bg-color-2: #2b3035;
@@ -126,5 +125,4 @@
         --error-border-color: #800;
         --error-text-color: #e33;
         --error-highlight-text-color: #e66;
-    }
 }
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -419,7 +419,11 @@ export function coreContext() {
 
   /* Container */
   let _container = d3_select(null);
-  let _theme;
+
+  const themeQuery = window.matchMedia?.('(prefers-color-scheme: dark)');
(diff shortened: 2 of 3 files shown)


hlfan:
Doesn't make this the `.theme-light` class unnecessary?

tyrasd:
> switch to a faster bundler

Could you please elaborate on this? The css building is not an issue AFAICS, taking only a couple milliseconds. Also most of the rest of the bundling process is not slow: Only the creation of the `dist:svg:mapillary:signs` sprite is taking significant amounts of time (presumably simply because there is a very large number of icons to bake into the sprite).

tordans:
> > switch to a faster bundler
> 
> Could you please elaborate on this? 

Context: https://osmus.slack.com/archives/C0754VCK2Q7/p1773582681957799 
I was experimenting with the new Vite 8 and looking for feedback.

hlfan:
Also this contrasts a principle from https://github.com/openstreetmap/iD/pull/11308#issue-3320933523:

> Making the theme respond anytime results in variables like `--bg-filter` and `--bg-blend-mode` to not need to watch for media query changes in JS.

k-yle:
> > Also this contrasts a principle from https://github.com/openstreetmap/iD/pull/11308#issue-3320933523:
>
> Making the theme respond anytime results in variables like `--bg-filter` and `--bg-blend-mode` to not need to watch for media query changes in JS.

this part i don't quite understand, since the theme is still responsive to `prefers-color-scheme` changes

but in any case, lets just close this an revisit it in a few years when [CSS's `if`-statements](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/if) are widely supported

hlfan:
> the theme is still responsive to `prefers-color-scheme` changes

It should be as responsive as CSS-only, but in my testing, JS sometimes took a little longer. Idk why.

PR #12042, opened 2026-03-18
delete unused files

k-yle:
deleted a few random files that are no longer used, or completely out of date. 

one image was worth keeping, added it to the corresponding section of `ARCHITECTURE.md`


docs/coding_standards.txt
```diff
@@ -1,26 +0,0 @@
-Coding standards and advice for iD
```
k-yle, line 1: this is 14 year old advice, which has been superseeded by `CONTRIBUTING.md`

PR #12059, opened 2026-03-20
Start refactoring UI modules to typescript

tyrasd:
In order to avoid oversights such as 0241c7653d4a93692f59b7aa1021242e63a3de72 from slipping through (incomplete) unit tests as well as manual testing, it is probably necessary to port the frontend modules over to typescript.

As a proof of concept here is how this might look like for one of the UI code:
* [x] `modules/ui/sections/map_style_options.js`
* [x] `modules/ui/fields/check.ts`
* [x] `modules/ui/cmd_sequence.ts` in #11998

Notably, for `d3.select` one needs to include the respective type hints, e.g. `selection.selectAll<HTMLUListElement, any>(ul.…)`, and interfacing to existing not-type annotated modules sometimes require an explicit case to `any` (e.g. in `uiSectionMapStyleOptions(context: any)`, or `(uiTooltip() as any).…`), at least as long as we don't port them over as well.

Feedback welcome!

### DIFF
--- a/modules/globals.d.ts
+++ b/modules/globals.d.ts
@@ -9,7 +9,14 @@ declare global {
   declare var after: typeof afterEach;
   declare var VITEST: true;
 
-  declare type Tags = { [key: string]: string | undefined };
+  declare type EntityID = string;
+
+  declare type TagKey = string;
+  declare type TagValue = string;
+  declare type TagValueUpdate = string | undefined;
+  declare type Tags = { [key: TagKey]: TagValue };
+  declare type TagsMulti = { [key: TagKey]: TagValue | TagValue[] };
+  declare type TagsUpdate = { [key: TagKey]: TagValueUpdate };
 
   /**
    * A class method that acts as both a getter and a
--- a/modules/globals.d.ts
+++ b/modules/globals.d.ts
@@ -38,7 +38,7 @@ declare global {
   declare namespace d3 {
     export type Selection<T = any> = import('d3').Selection<
       T,
-      unknown,
+      any,
       unknown,
       unknown
     >;
--- a/modules/globals.d.ts
+++ b/modules/globals.d.ts
@@ -9,7 +9,7 @@ declare global {
   declare var after: typeof afterEach;
   declare var VITEST: true;
 
-  declare type Tags = { [key: string]: string };
+  declare type Tags = { [key: string]: string | undefined };
 
   /**
    * A class method that acts as both a getter and a
--- a/modules/ui/fields/check.ts
+++ b/modules/ui/fields/check.ts
@@ -18,10 +18,10 @@ export { uiFieldCheck as uiFieldOnewayCheck };
 export function uiFieldCheck(field: any, context: iD.Context) {
     const dispatch = d3_dispatch('change');
     let options = field.options;
-    let values: (string | undefined)[] = [];
+    let values: TagValueUpdate[] = [];
     let texts: ReturnType<typeof t.append>[] = [];
 
-    let _tags: Tags;
+    let _tags: TagsMulti;
 
     let input:    d3.Selection<HTMLInputElement>  | d3.Selection<null> = d3_select(null);
     let text:     d3.Selection<HTMLSpanElement>   | d3.Selection<null> = d3_select(null);
@@ -29,8 +29,8 @@ export function uiFieldCheck(field: any, context: iD.Context) {
     let reverser: d3.Selection<HTMLButtonElement> | d3.Selection<null> = d3_select(null);
 
     let _impliedYes: boolean;
-    let _entityIDs: string[]  = [];
-    let _value: string | undefined;
+    let _entityIDs: EntityID[]  = [];
+    let _value: TagValueUpdate;
 
 
     var stringsField = field.resolveReference('stringsCrossReference');
@@ -130,7 +130,7 @@ export function uiFieldCheck(field: any, context: iD.Context) {
         input
             .on('click', function(d3_event) {
                 d3_event.stopPropagation();
-                var t: Tags = {};
+                var t: TagsUpdate = {};
 
                 if (Array.isArray(_tags[field.key])) {
                     if (values.includes('yes')) {
@@ -186,7 +186,7 @@ export function uiFieldCheck(field: any, context: iD.Context) {
     };
 
 
-    check.tags = function(tags: Tags) {
+    check.tags = function(tags: TagsMulti) {
 
         _tags = tags;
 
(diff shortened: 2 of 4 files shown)


modules/ui/sections/map_style_options.ts
```diff
@@ -21,42 +24,47 @@ export function uiSectionMapStyleOptions(context) {
...
             .data([0]);
 
         container2.enter()
             .append('ul')
             .attr('class', 'layer-list layer-visual-diff-list')
             .merge(container2)
-            .call(drawListItems, ['highlight_edits'], 'checkbox', 'visual_diff', toggleHighlightEdited, function() {
-                return context.surface().classed('highlight-edited');
-            });
+            .call(drawListItems, ['highlight_edits'], 'checkbox', 'visual_diff', toggleHighlightEdited, () =>
```
tordans, line 33: It would be great to put this into the coding style. Do we use arrow functions or not; and if, when… – Or is it OK, when we mix it up randomly…

In all my other projects I am all in on arrow functions inline and for helper. But for iD I tried to always use the other pattern to keep it the same…

tordans (commented):
+1 to push the migration a bit more!

I hope we will find a good way to reduce the `any` cases. I assume @k-yle will have some ideas here.

modules/ui/sections/map_style_options.ts
tyrasd: +1 for recommending arrow functions for anonymous inline functions. Specifically with d3 there are a few cases where regular `function`s are still needed, e.g. to access `this` in the callback of `.each(function(d) { d3_select(this).… }`.

tyrasd: > put this into the coding style

This is actually already covered by the linked code style guidelines for Javascript, e.g. https://github.com/airbnb/javascript?tab=readme-ov-file#arrow-functions and similarly for the recommendation for `const/let` instead of `var`: https://github.com/airbnb/javascript?tab=readme-ov-file#references

modules/ui/sections/map_style_options.ts
```diff
@@ -1,18 +1,21 @@
...
+export function uiSectionMapStyleOptions(context: any) {
+
+    const section = (uiSection('fill-area', context) as any)
         .label(() => t.append('map_data.style_options'))
         .disclosureContent(renderDisclosureContent)
         .expandedByDefault(false);
 
-    function renderDisclosureContent(selection) {
-        var container = selection.selectAll('.layer-fill-list')
+    function renderDisclosureContent(selection: Selection<HTMLElement, any, any, any>) {
```
k-yle: there's a helper in [modules/globals.d.ts](https://github.com/openstreetmap/iD/blob/develop/modules/globals.d.ts) which makes this a bit less verbose:
```suggestion
    function renderDisclosureContent(selection: d3.Selection<HTMLElement>) {
```
or even
```suggestion
    function renderDisclosureContent(selection: d3.Selection) {
```

modules/ui/sections/map_style_options.ts
```diff
@@ -1,18 +1,21 @@
...
 
 import { t } from '../../core/localizer';
 import { uiTooltip } from '../tooltip';
 import { uiSection } from '../section';
 
-export function uiSectionMapStyleOptions(context) {
+export type MapStyle = 'wireframe' | 'area_fill' | 'highlight_edits';
 
-    var section = uiSection('fill-area', context)
+export function uiSectionMapStyleOptions(context: any) {
```
k-yle:
```suggestion
export function uiSectionMapStyleOptions(context: iD.Context) {
```
we have definitions for this already, but it's blocked by #11589 (which adds defintions for the last 20% of attributes)

k-yle (commented):
it's awesome to have an official endorsement of #8651 🎉

I have a (closed source) fork of iD where most of the core has been rewritten in TypeScript, so I'm happy to send more PRs to backport this work

modules/globals.d.ts
```diff
@@ -38,7 +38,7 @@ declare global {
   declare namespace d3 {
     export type Selection<T = any> = import('d3').Selection<
       T,
-      unknown,
+      any,
```
tyrasd, line 48: do you think this is a good idea @k-yle ? Without it I did get _x is not assignable to y_ / _Type 'unknown' is not assignable to type '…'_ errors when trying to use this type as the parameter definition of the `drawListItems` method: Because we set `.data([0])` to a number, the type checker knows that the `container` is of type `Selection<HTMLUListElement, number, …, …>`, but that is not compatible to `Selection<any, unknown, unknown, unknown>`.

Since we often actually not care about the actual type of the bound data to a selection, it seems like `any` is reasonable for this shorthand type?

modules/globals.d.ts
```diff
@@ -9,7 +9,7 @@ declare global {
   declare var after: typeof afterEach;
   declare var VITEST: true;
 
-  declare type Tags = { [key: string]: string };
+  declare type Tags = { [key: string]: string | undefined };
```
tyrasd: also here: //cc @k-yle : I'm not sure this is superbly elegant, but a lot of front end code uses a `tags` object where setting a particular entry's value to `undefined` is interpreted as _this tag should be deleted_.

modules/globals.d.ts
k-yle: yeah I agree - I think we actually made this exact change in a different PR 😆 (can't remember which one)

`unknown` is unhelpfully strict here

modules/ui/fields/check.ts
k-yle, line 1: I think we should delete `check.js` in the same commit right? otherwise the git blame will be lost

modules/globals.d.ts
k-yle: it's tricky because this will make it less convenient for other files where `undefined` is never used as a value. maybe we need 3 different interfaces for the 3 different cases?

```ts
{ [key: string]: string }
{ [key: string]: string | undefined }
{ [key: string]: string | string[] }
```

e.g. `check.js` uses the 3rd case right? arrays are used when there's a multi-selection with different values

modules/ui/fields/check.ts
tyrasd: indeed, sorry, I must have forgot to do that.

modules/globals.d.ts
tyrasd: Good point, for multi-selections the value can be a `string[]` too[^1]. E.g. as checked in `Array.isArray(_tags[field.key]`.

But `check.ts` also uses the explicit `undefined` version also, see line 48: `values = [undefined, 'yes'];`, later this is assigned to the updated tags in `t[field.key] = values[…];`.

But I think the two/three cases are luckily mutually exclusive: `undefined` as a value is only allowed when updating the tags of objects, and `string[]` is only occurring as the input of the ui fields. I committed 70d97e511 with a potential way to divide this up.


[^1]: PS: Honestly, I would consider the way multiselections are encoded as these `key=<string | string[]>` construct as very suboptimal if not to say hacky. This makes handling them correctly quite cumbersome in places where the tags are not trivially transformable to fields (e.g. [here](https://github.com/openstreetmap/iD/blob/e2cf8e979499f7c70b4bbac1a30e326baf4722fc/modules/ui/fields/directional_combo.js#L135)). Maybe after we have this all as typescript, this can be refactored with confidence and without too much pain to `Tags[]`.

k-yle: 💯 yeah definitely, i think 70d97e5 is a good solution

PR #12060, opened 2026-03-20
Fix spacebar creating duplicate nodes during way/area drawing

### LINKED ISSUE #12051: Using spacebar for drawing way/areas can result in multiple nodes with the same coordinates
### URL

_No response_

1. Start drawing a line or an area
2. Move the mouse cursor
3. Press `Space` a few times
4. Finish drawing the line/area
5. Use the lasso select tool to select the nodes which you created using the space bar
6. *Notice there is more than 1 node*

https://github.com/user-attachments/assets/8df44cb9-c689-4324-aa4c-8cef24db4395 

 _Originally posted by @Dimitar5555 in [#9155](https://github.com/openstreetmap/iD/issues/9155#issuecomment-4034112570)_

2.38 and 2.39

Firefox

JaiswalShivang:
Fixes #12051

### Description
When drawing a way or area, pressing Space multiple times (or holding it down) could create multiple stacked nodes at the same or nearly identical coordinates. This happened because the browser fires repeated `keydown` events during key-repeat, and the existing `_disableSpace` guard would re-enable itself once the mouse moved more than 12 pixels.

### Root Cause
The [space()] function in [behavior/draw.js] relied solely on a `_disableSpace` flag that could be reset by mouse movement, allowing key-repeat `keydown` events to slip through and create extra nodes.

### Fix
Added a `_spaceHeld` flag that tracks whether the spacebar is physically held down:
- Set to `true` on the first `keydown` event — blocks all subsequent key-repeat events
- Reset to `false` only on `keyup` — ensures exactly one node per press-release cycle

This preserves the ability to rapidly tap Space to place multiple intentional nodes, while preventing accidental duplicates from key-repeat or holding the key.

### Testing
- `npm run test:once` passes
- ESLint passes with zero warnings

### DIFF
--- a/modules/behavior/draw.js
+++ b/modules/behavior/draw.js
@@ -12,6 +12,7 @@ import { utilFastMouse, utilKeybinding, utilRebind } from '../util';
 
 var _disableSpace = false;
 var _lastSpace = null;
+var _spaceHeld = false;
 
 
 export function behaviorDraw(context) {
@@ -191,6 +192,10 @@ export function behaviorDraw(context) {
         d3_event.preventDefault();
         d3_event.stopPropagation();
 
+        // ignore key-repeat events while space is held down
+        if (_spaceHeld) return;
+        _spaceHeld = true;
+
         var currSpace = context.map().mouse();
         if (_disableSpace && _lastSpace) {
             var dist = geoVecLength(_lastSpace, currSpace);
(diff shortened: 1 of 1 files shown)


JaiswalShivang:
Hi @tyrasd, I find that this pr is closed. Could sir let me know if there's anything I should change or if this approach is no longer needed? I'd be happy to fix it!

tordans:
FYI:  Fixed via  https://github.com/openstreetmap/iD/commit/c4bba05b4759c0f56eb79288c9fa4766752fe5d6

tyrasd:
This and the fact that the proposed fix with `_spaceHeld` is essentially only duplicating the existing logic and not actually fixing the issue #12051. :shrug:

PR #12070, opened 2026-03-21
fix: preserves all tags when re-selecting the same preset

### LINKED ISSUE #12068: Reselecting the same preset deletes all known tags
Simple repro: draw way, select "Foot Path" preset (`highway=footway`), add `surface=asphalt`, choose new preset, select "Foot Path" again, the `surface` tag is deleted.

You can do this to any feature - reselect the same preset and it wipes out all known tags.

2.39.2

k-yle:
Closes #12068

### DIFF
--- a/modules/actions/change_preset.js
+++ b/modules/actions/change_preset.js
@@ -25,7 +25,7 @@ export function actionChangePreset(entityID, oldPreset, newPreset, skipFieldDefa
                     .forEach(key => preserveKeys.push(key));
             }
 
-            if (oldPreset) {
+            if (oldPreset && (oldPreset.id !== newPreset.id)) {
                 // 'field-keys' are keys used by fields (different to the keys used by preset itself)
                 const oldPresetFieldKeys = [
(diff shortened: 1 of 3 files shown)


modules/actions/change_preset.js
```diff
@@ -25,7 +25,7 @@ export function actionChangePreset(entityID, oldPreset, newPreset, skipFieldDefa
                     .forEach(key => preserveKeys.push(key));
             }
 
-            if (oldPreset) {
+            if (oldPreset && (oldPreset.id !== newPreset.id)) {
```
tyrasd, line 28: At first glance, I was wondering if it were not slightly more elegant to include this logic on line 17 (depending on whether one would consider a preset as implicitly being also it's own sub-preset), but I think it is also fine to just skip this branch here for this case, as it seems unlikely that the lack of contents in `preserveKeys` would produce further problems in the future when handling this case.

tyrasd (approved):

Issue #12067, opened 2026-03-20
Change the way we show the field value description


tordans:
### Description

Based on feedback in https://github.com/openstreetmap/iD/issues/12065#issuecomment-4099396630
Follow up to https://github.com/openstreetmap/iD/pull/11876

**Goal:**
- For experienced mappers, show a minimal dropdown like we did "before" with just the (translated) values
- For mappers that want more inline help, show the custom id-tagging-schema description nearby the value

---

**How to solve it…**

Continued from https://github.com/openstreetmap/iD/issues/12065#issuecomment-4099396630

> > option in the settings
> 
> Yeah, that would be a decent idea for a customizing option. Could you please open a new ticket with that particular feature request?

I would rather like to look for a way to toggle the details off somewhere in the preset space. One idea was to only show the additional line once the info-i was activated once and use this action as an indicator that the user wants more context. But that is rather indirect. 
The second idea from https://github.com/openstreetmap/iD/pull/11876 also goes in this direction but I don't like the interaction with those separate `(i)`'s. 
We could look into something like this
> <img width="381" height="138" alt="Image" src="https://github.com/user-attachments/assets/f64fb3b4-b030-467a-bdf0-605c295ea150" />

1ec5:
Two different ℹ️ buttons in the same field would get confusing quick. If this new toggle only applies to the immediate field or only temporarily, then maybe we could add a sticky item at the bottom of the dropdown for toggling it. But if it applies globally or persistently, it should go in the Preferences pane, which is quite underused.

That said, every preference comes with a maintenance/documentation/cognitive burden, so we should make sure it’s truly necessary in the long term. In this case, I wonder if a similarly prominent but less persistent presentation would better serve both audiences: https://github.com/openstreetmap/iD/issues/12065#issuecomment-4099603470.

tyrasd:
I'm still not convinced we necessarily need to change this. The benefit of the current solution is that one can actually see all the different values at a glance, which is especially useful when choosing a value of a field/tag one is not yet familiar with (which can also benefit experienced mappers). With the also proposed tooltip solution one needs to slowly scrub through all potential options to get to see the descriptions.

> For experienced mappers, show a minimal dropdown like we did "before" with just the (translated) values

I also don't fully understand this argument. Why would it be better for experienced mappers to omit the descriptions? Yes, it looks a little bit differently than before, so one's muscle memory might need to be readjusted about where exactly to click, but I don't think that should be an argument here.

[...]

Issue #12065, opened 2026-03-20
changes in 2.39: the values of the access key take up a lot of space


maro-21:
I don't like this change:

<img width="372" height="468" alt="Image" src="https://github.com/user-attachments/assets/47a82d7c-0ea3-4e07-a399-31c5ce5f9ed5" />

----
I don't think it's an improvement; it just makes things harder for editors.
First, the descriptions of these values appear when you hover over them, so having the same text twice is unnecessary.
Second, because the descriptions are so long, they don't fit on the screen, so to select the "private" value, I have to scroll, which is annoying because this wasn't the case before.
Third, this isn't a practice used in other fields in the editor.

I request that the previous version be restored.

tyrasd:
Showing the additional text help specifically new users to understand the different concrete meanings of the raw access values (not everyone does know the wiki by heart), and especially helps users who's native language is not English.

> hover […] having the same text twice

this was addressed in #12010

> Second, because the descriptions are so long, they don't fit on the screen,

Perhaps this would be improved if the dropdown box for the access values would simply use the full available width, and not stick to the half-width input field size.

maro-21:
> Showing the additional text help specifically new users to understand the different concrete meanings of the raw access values (not everyone does know the wiki by heart), and especially helps users who's native language is not English.

But the meanings of values are in the tooltips!

I understand that you want to help new users. But there are also experienced users (I've been using iD for 10 years) who know the tags by heart and could type them in manually. And the recent change makes editing more difficult. I'd like you to keep in mind that not everyone is a programmer and has a 30-inch monitor where everything is visible and there are no scroll bars.

Maybe there could be an option in the settings so that each user could set it however they want?

I don't understand why these tag descriptions were added below the values, even though there are tooltips with the same text. And why is it for access tag fields while other fields don't have it.

tyrasd:
> But the meanings of values are in the tooltips! […] even though there are tooltips with the same text. 

As mentioned above: it is not intentional that the description is both in the dropdown and the tooltip, and those are already not shown anymore in the current development version of iD.

> know the tags by heart […] makes editing more difficult

did you consider using the raw tag editor in this case, as typing the access values in there is probably even faster / less obstructed.

> option in the settings

Yeah, that would be a decent idea for a customizing option. Could you please open a new ticket with that particular feature request?

tordans:
> Could you please open a new ticket with that particular feature request?

[...]

PR #12074, opened 2026-03-22
Improve combobox options with info tooltip and visual caret indicator

aditya1010-max:
### Summary

This PR enhances combobox options by adding contextual information and improving visual clarity.

* Adds an info icon (`ℹ️`) to options that include a description
* Displays the description via tooltip on hover (and touch for mobile)

---

### Background

This change builds on the discussions in #12065 and the issue raised in #12067.

While descriptions are useful, displaying them inline for every option can make the list harder to scan and introduce visual clutter. This approach keeps the information accessible while presenting it in a more lightweight and discoverable way through tooltips.

---

### Notes on UX

This implementation aims to balance clarity and simplicity:

* New mappers benefit from having descriptions available via tooltips
* Experienced users are not overwhelmed with additional text in the list
* Interaction remains consistent with existing combobox behavior (no nested dropdowns or new click actions)

---

### Future Consideration

It may be worth exploring a user preference or setting to control how much contextual information is shown (e.g. always visible vs on-hover), depending on user experience level.

---

### Desktop and Mobile

On desktop devices, users can hover over the info icon to view additional context via a tooltip.  
On touch devices, users can long-press the info icon to reveal the description.

This approach reduces visual clutter—especially on smaller screens—while still providing access to contextual information on demand.

Video -

https://github.com/user-attachments/assets/f01db33d-8273-468c-99bb-a0cbc387f6ad

Image(large screen device)- 
image
Image(small screen device)- 
image

### Testing

* Verified tooltip behavior on hover and touch
* Ensured no changes to option selection behavior
* Confirmed layout remains consistent with existing combobox styles

### Feedback

Would appreciate feedback on the interaction design, especially the tooltip behavior across desktop and touch devices.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1868,10 +1868,65 @@ input.date-selector {
     display: flex;
     align-items: center;
 }
-.combobox-option.has-icon .combobox-option-label {
+.combobox-option.has-icon {
     display: flex;
     align-items: center;
 }
+
+.combobox {
+  scrollbar-gutter: stable;  /*helped remove the jitter due to "i" and scrollbar overlap*/
+}
+
+.combobox-option-info .icon {
+    flex-direction: column;
+    align-items: flex-start;
+}
+
+.combobox-option-info {
+  width: 18px;
+  height: 18px;
+  border-radius: 50%;
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  border: 1px solid var(--border-color);
+  position: relative;
+  left: 7px;
+
+  cursor: pointer;
+
+  transition: 
+    background-color 0.2s ease,
+    border-color 0.2s ease,
+    transform 0.15s ease;
+}
+
+.combobox-option-info:hover {  /*for laptop, pc*/
+  border-color: #ffffff;
+  background-color: #258292;
+  transform: scale(1.05);
+}
+
+.combobox-option-info.is-active {  /*same as above class but for mobile devices*/
+  border-color: #ffffff;
+  background-color: #1e6a77;
+   transform: scale(1.03);   /* slightly less aggressive */
+}
+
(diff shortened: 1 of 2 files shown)


tordans:
Quick feedback: If we go down the tooltip route, I don't see any reason to add another avoidance to show the tool tip. Hovering over the options line is enough to trigger the tooltip. The tooltip needs to be positioned outside the options area so it doesn't overlap with the label (unless no space, then above?). 
We also need to rework the PR I did and re-merge the two kinds of tooltip (manual translate and taginfo) to streamline the behavior. The testing doc needs to be similar to what I did in that PR.

tyrasd:
I'm still not convinced we need to change this in the first place (#12067), and besides the proposed solution here is not the kind of UX patter we would use for this anyway.

aditya1010-max:
> Quick feedback:

Thanks for the feedback and the UI improvements.
Agree that reworking your PR and merging the tooltip implementations (manual translate and taginfo) is the more important direction right now to streamline the behavior.
I’ll also go through the testing documentation you mentioned and learn from the structure used there.

aditya1010-max:
> I'm still not convinced we need to change this in the first place (#12067), and besides the proposed solution here is not the kind of UX patter we would use for this anyway.


Thanks for the clarification — that makes sense.

On reflecting more about the UX and the discussion in #12067, I agree that keeping the current approach simple and easily scannable is likely the better direction here.

I had misinterpreted the earlier comment in #12065 about screen size more broadly in terms of improving UI for smaller screens, which is why I focused on the mobile/touch interaction aspect. I see now that the main concern here is preserving visibility and making it easy to scan all options at once, which the current approach already handles well.

I’ll keep this in mind when thinking about similar UI changes going forward.

PR #12085, opened 2026-03-24
Remove copyright year in LICENSE.md

matkoniecz:
Updating it to the current year or replacing with range also would be an option

See https://github.com/openstreetmap/id-tagging-schema/pull/2111


matkoniecz:
maybe we should contact LWG?

tyrasd (approved):
> maybe we should contact LWG?

IMO this is so trivial, we don't need to jump through the extra hoops.

PR #12095, opened 2026-03-24
Fix "false" tooltip on access fields (fixes #12064)

### LINKED ISSUE #12064: bug in 2.39: 'false' when hovering over the access field
When hovering over the access field which is filled in, there is "false".
Similar to https://github.com/openstreetmap/iD/issues/8289

Image

Ashish-Ranjan-jpg:
This PR fixes a bug where access fields (e.g., [access](cci:1://file:///c:/Users/HP/OneDrive/Desktop/opensource/iD/modules/ui/fields/access.js:13:4-69:5), `motor_vehicle`, `foot`) show a literal string `"false"` in their tooltip when a single value is set.
### The Problem
In [modules/ui/fields/access.js](cci:7://file:///c:/Users/HP/OneDrive/Desktop/opensource/iD/modules/ui/fields/access.js:0:0-0:0), the logic for the `title` attribute used a logical `&&` expression that returned a boolean `false` if the tag value was not an array (i.e., not a mixed selection). D3 then rendered this boolean as the string `"false"` in the HTML `title` attribute.
### The Solution
Updated the logic to return `undefined` instead of `false` in those cases. Returning `undefined` correctly instructs D3 and the browser not to render a tooltip for single-selection tags, which matches the behavior of other fields in the iD editor.
### Visuals
| Before | After |
| :---: | :---: |
| Screenshot 2026-03-24 152230 | Screenshot 2026-03-25 011033 |
### Testing
- **Local Verification**: Tested in the browser at `http://localhost:8080`.
- **Single Selection**: Confirmed the "false" tooltip no longer appears.
- **Mixed Selection**: Confirmed the tooltip still correctly shows the list of unique values when multiple objects are selected.
- **Linting**: Ran `npx eslint modules/ui/fields/access.js` and confirmed the file is clean.

fixes #12064

### DIFF
--- a/modules/ui/fields/access.js
+++ b/modules/ui/fields/access.js
@@ -290,7 +290,7 @@ export function uiFieldAccess(field, context) {
                     || new Set(getAllPlaceholders(tags, accessField)).size > 1;
             })
             .attr('title', function(accessField) {
-                return tags[accessField] && Array.isArray(tags[accessField]) && tags[accessField].filter(Boolean).join('\n');
+                return (Array.isArray(tags[accessField]) && tags[accessField].filter(Boolean).join('\n')) || undefined;
             })
             .attr('placeholder', function(accessField) {
                 let placeholders = getAllPlaceholders(tags, accessField);
--- a/modules/ui/fields/access.js
+++ b/modules/ui/fields/access.js
@@ -290,7 +290,7 @@ export function uiFieldAccess(field, context) {
                     || new Set(getAllPlaceholders(tags, accessField)).size > 1;
             })
             .attr('title', function(accessField) {
-                return (Array.isArray(tags[accessField]) && tags[accessField].filter(Boolean).join('\n')) || undefined;
+                return Array.isArray(tags[accessField]) ? tags[accessField].filter(Boolean).join('\n') : undefined;
             })
             .attr('placeholder', function(accessField) {
                 let placeholders = getAllPlaceholders(tags, accessField);
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -916,7 +916,15 @@ export function uiFieldCombo(field, context) {
                 .classed('mixed', function(d) {
                     return d.isMixed;
                 })
-                .attr('title', d => d.isMixed ? t('inspector.unshared_value_tooltip') : (!['yes', 'no'].includes(d.state) ? d.state : undefined))
+                .attr('title', function(d) {
+                    if (d.isMixed) {
+                        return t('inspector.unshared_value_tooltip');
+                    }
+                    if (!['yes', 'no'].includes(d.state)) {
+                        return d.state;
+                    }
+                    return undefined;
+                })
 
                 .classed('negated', d => d.state === 'no');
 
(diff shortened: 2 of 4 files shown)


modules/ui/fields/access.js
```diff
@@ -290,7 +290,7 @@ export function uiFieldAccess(field, context) {
                     || new Set(getAllPlaceholders(tags, accessField)).size > 1;
             })
             .attr('title', function(accessField) {
-                return tags[accessField] && Array.isArray(tags[accessField]) && tags[accessField].filter(Boolean).join('\n');
+                return (Array.isArray(tags[accessField]) && tags[accessField].filter(Boolean).join('\n')) || undefined;
```
tordans:
```suggestion
                return Array.isArray(tags[accessField]) ? tags[accessField].filter(Boolean).join('\n') : undefined;
```

tordans:
Could you link / check the code for other fields eg the combo field and compare this part of the code. We should use the same style for both.

Ashish-Ranjan-jpg:
Checked other fields and they seems to follow the same styling. Updated the radio.js to use `undefined` instead of `null`, and refactored the title attribute handling in the combo field to match the ternary pattern. Everything else appears consistent with the existing pattern.
I’ll commit these changes shortly.

Ashish-Ranjan-jpg:
@tordans I’ve addressed the suggested things, could you please take another look on it.

modules/ui/fields/combo.js
```diff
@@ -916,15 +916,8 @@ export function uiFieldCombo(field, context) {
...
-                .attr('title', function(d) {
-                    if (d.isMixed) {
-                        return t('inspector.unshared_value_tooltip');
-                    }
-                    if (!['yes', 'no'].includes(d.state)) {
-                        return d.state;
-                    }
-                    return null;
-                })
+                .attr('title', d => d.isMixed ? t('inspector.unshared_value_tooltip') : (!['yes', 'no'].includes(d.state) ? d.state : undefined))
```
tordans: We don't use arrow functions (yet); not sure if we want to start here and now.
(There is discussion on if/when to use them in the future in a different thread.)
For me ternary operators work great of one level but the code before with two conditions was easier to read/understand.

package-lock.json
```diff
@@ -97,7 +97,7 @@
         "svg-sprite": "2.0.4",
         "typescript": "^5.9.3",
         "typescript-eslint": "^8.56.0",
-        "vite": "^8.0.1",
+        "vite": "^7.3.1",
```
tordans: Looks like we are on vite 8 now, interesting https://github.com/openstreetmap/iD/blob/develop/package.json#L141
But this package lock change should not be in the PR AFAIK and IMO.

tordans (commented):
Can only comment some more opinion on this. I don't do full reviews and cannot do merges here.

modules/ui/fields/combo.js
Ashish-Ranjan-jpg: Changed the arrow function back to the regular functions .
Yes, using ternary operator there made it bit uneasy to read, so changed it back with two conditions.
Please let me know if everything is ok now.

package-lock.json
Ashish-Ranjan-jpg: Staged the package-lock file unintentionally, reset it back to upstream/develop, now it is not in the "files changed" of this pr.

PR #12099, opened 2026-03-25
Side indicators on the map for `directionalCombo`s

### LINKED ISSUE #11950: Field-specific side indicator
### Description

[left/right stuff](https://wiki.openstreetmap.org/wiki/Forward_%26_backward,_left_%26_right) is highly confusing (not saying that tagging schema is bad, I see no better solution)

I think that iD can be smarter here - my idea is to highlight side when user is editing side-specific field.

When user edits left side of sidewalk, then left side of road may get transparent highlight on left side, with matching colour marking on left field. The same for right side. See mockups below.

(there is also potential for showing arrows for forward/backward directions, if side highlight gets implemented and it works well this could be the next one)

### Screenshots (mockup)

left side is active:
Image

right side is active:
Image

[iD show sides.zip](https://github.com/user-attachments/files/25696990/iD.show.sides.zip)

tordans:
A take to resolve https://github.com/openstreetmap/iD/issues/11950

> [!NOTE]
> **This is a draft PR meant to discuss styling and interaction.** 
> The code is generated and checked only a litte; not reviewed yet and not meant to be reviewed.

---

Right now, it is quite hard to add and review direction input fields on the map. One has to find the small image arrow – that is sometimes hidden below the label name – and then extrapolate the left|right side based on that.

The idea of this change is, to give a visual indicator on the map when using the specific side.

There are a few cases out there that use a similar style…
* https://github.com/zlant/parking-lanes
* https://a-b-street.github.io/speedwalk

<details><summary>Example Screenshot from Speedwalk</summary>
<p>

Bildschirmfoto 2026-02-02 um 16 39 40

</p>
</details> 

But this solution suggests to only show the side that we are selecting ATM and only when we are working on this specific field/row.

---

### Blue arrows

The idea is to reuse the color we use for the newly added helper lines from https://github.com/openstreetmap/iD/pull/11774

Bildschirmfoto 2026-03-25 um 06 38 08
Bildschirmfoto 2026-03-25 um 06 38 13

### Red arrow

The idea here is to re-use the red color that we use for the "high is highlighted" and just add a "we look at this direction for this field" icon.

Bildschirmfoto 2026-03-25 um 06 40 48
Bildschirmfoto 2026-03-25 um 06 40 51

### Red select indicator (and label)

The idea here is to reuse the selection indicator below the line and move it to be bigger on the side that we indicate. (Ideally this would still show on the "other" side and just make the indicator side larger; here it moves.)

This also shows how we might show the key+value as an additional label. (This is not fleshed out, yet…)

Bildschirmfoto 2026-03-25 um 06 50 59
Bildschirmfoto 2026-03-25 um 06 51 03

### Other options?

https://github.com/openstreetmap/iD/pull/10303 does something with side, but there is no style we can reuse here, IMO. Other than using the gray color, maybe. But I don't think this will help here. We use the gray indicators for things that are visible on the map all the time; the blue and red is for things that happen on indicator.

--- 

The Code will handle `forward|backward` as well by translating it to `rigth|left`. Does this sound right and what field could we test this on? https://github.com/openstreetmap/iD/pull/12099/changes#diff-4fa4b7fd7b8bbe73f21424cef0b08d14b77f072153d2f744c5728e47f1585569R33-R36

### DIFF
--- a/css/20_map.css
+++ b/css/20_map.css
@@ -13,6 +13,8 @@
 .layer-osm path.sided-marker-coastline-path { fill: #77dede; }
 .layer-osm path.sided-marker-barrier-path   { fill: #ddd; }
 .layer-osm path.sided-marker-man_made-path  { fill: #fff; }
+.layer-osm path.directionalcombo-marker-left-path,
+.layer-osm path.directionalcombo-marker-right-path { fill: #7092ff; }
 
 /* IE/Edge rule for <use> marker style */
 .layer-osm path.viewfield-marker-path {
@@ -243,8 +245,13 @@ text {
 
 .onewaygroup path.oneway,
 .viewfieldgroup path.viewfield,
-.sidedgroup path.sided {
+.sidedgroup path.sided,
+.directional-combo-group path.directionalcombo {
     stroke-width: 6px;
+}
+
+.directional-combo-group path.directionalcombo {
+    opacity: 0.95;
 }
 
 text.arealabel-halo,
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1664,6 +1664,23 @@ input.date-selector {
     display: table-cell;
     width: auto;
 }
+.form-field ul.rows.rows-table li.labeled-input > .preset-label-directionalcombo .directionalcombo-label-arrow {
+    width: 13.2px;
+    height: 13.2px;
+    margin-left: 0.35em;
+    vertical-align: middle;
+    transform: rotate(var(--indicator-rotation, 0deg));
+    transform-origin: center;
+    opacity: 1;
+}
+.form-field ul.rows.rows-table li.labeled-input > .preset-label-directionalcombo .directionalcombo-label-arrow path {
+    fill: var(--passive-text-color);
+    transition: fill 120ms ease-in-out;
+}
+.form-field ul.rows.rows-table li.labeled-input.is-active-indicator > .preset-label-directionalcombo .directionalcombo-label-arrow path { fill: #7092ff; }
+.form-field ul.rows.rows-table li.labeled-input > .preset-label-directionalcombo .directionalcombo-label-arrow.directionalcombo-label-arrow-hidden {
+    display: none;
+}
 
 
 /* Field - Structure
--- a/modules/core/context.js
+++ b/modules/core/context.js
@@ -336,6 +336,28 @@ export function coreContext() {
   context.selectedIDs = () => (_mode && _mode.selectedIDs && _mode.selectedIDs()) || [];
   context.activeID = () => _mode && _mode.activeID && _mode.activeID();
 
+  /**
+   * @typedef {{
+   *   side: 'left' | 'right',
+   *   entityIDs: string[]
+   * }} DirectionalComboIndicatorState
+   */
+
+  /** @type {DirectionalComboIndicatorState | null} */
+  let _directionalComboIndicator = null;
+  context.directionalComboIndicator = function() {
+    return _directionalComboIndicator;
+  };
+  /**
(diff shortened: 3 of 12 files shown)


matkoniecz:
Thanks for looking into it!

My thoughts: in the first and second option these markings seems a bit detached from way itself. Not entirely sure how intuitive it will be.

In the third option these overlaps between segments are a bit annoying, but overall I really, really like on-the-map view.
For label - it probably should be translated label, not raw tag.
For sidedness indicator in tag menu: this red background sadly looks like an error indicator.

tordans:
Either as a follow up or as part of this PR I would like to add a similar logic to the forward/backward fields, eg `direction=forward|backward`, `*:direction=forward|backward`
> image

tordans:
Quick notes from a talk with @tyrasd at [a meetup](https://wiki.openstreetmap.org/wiki/FOSSGIS_2026/OSM-Samstag):
- First one one best
- Remove `forward|backward` logic; not used ATM and complex so only apply to `left|right`

tyrasd:
@tordans: would you like to keep this "prototype" PRs open, or would it make sense to boil this down to a shorter feature-request issue and eventually a "proper" implementation PR? :blush:

tordans:
> would you like to keep this "prototype" PRs open, or would it make sense to boil this down to a shorter feature-request issue and eventually a "proper" implementation PR? 😊

I did not find time to review this after our short talk https://github.com/openstreetmap/iD/pull/12099#issuecomment-4154791143 but I think it should be possible to move this from a exploration to something that can be properly reviewed in the next iteration. My plan is to implement the changes we talked about as a prototype. And when we agree that this is good from a UX/UI point of view, do a proper code cleanup and then mark this as ready for review.

tordans:
**This is the next iteration of this UX and UI exploration.**

Changes/Feature:
- Use blue arrows, no labels
- The arrows are visible while working on the field (label, input, open dropdown)

**New exploration:**
StreetComplete hat this nice Feature that is shows way in the Answer-UI in the direction of the selected way. This makes it really easy to align the form field with the map.
This version explores how we can replicate this for our fields:
- It shows a direction arrow next to the label (always, in gray)
- This arrow is blue when hovered, so this is our mental link to the arrows on the map
- The error use the same main direction that the arrows on the map point to
- The arrows now have a small shaft to make them easier to understand in this label use case

There are edge cases with this feature, that we need to consider: 
- A circle or very curved line does not have a direction to show; in those cases we don't show anything next to the label
- But we use the visible part of the line for this, so long ways with curves still get an error when zoomed in

**Preview:**

https://pr-12099--ideditor.netlify.app/#background=Bing&disable_features=boundaries&locale=en&map=21.17/52.28430/13.53946&id=w26957907

**Video:**

https://github.com/user-attachments/assets/e9e0cd9a-e99b-473f-bab3-41171a09cae0

(Note: In the video the arrows at the end are oriented the wrong way, that is fixed already.)

**Next steps**

- Do the map arrows look good now?
- What do we think about the label arrows?

Personally I like the label arrows but the do make the code more complex…

RudyTheDev:
> A circle or very curved line does not have a direction to show; in those cases we don't show anything next to the label

I think these should still show the arrows on the map, even if the field arrow isn't shown. Otherwise, as your video demo shows, this won't work for roundabouts and such. Personally, I rather not have field arrows at all than have roads that "fail" to show arrows due to curvature.

tordans:
> I think these should still show the arrows on the map, even if the field arrow isn't shown. Otherwise, as your video demo shows, this won't work for roundabouts and such.

That is a bug. The map should also show the arrows; only the label arrows are conditional.

PR #12102, opened 2026-03-25
show invalid values in radio fields and standard tooltip for radio field

### LINKED ISSUE #12082: should radio fields show tags that are set to an "invalid" value?
For `check`box fields, we do show the value of the respective tag in the case where it does not fit to one of the defined values (typically `yes` or `no`). This is useful for tags where there are special values that are interpreted basically the same as `yes`, but have a slightly different meaning like `covered=arcade`

Similarly, this can help for cases where the tag has a typo, or otherwise invalid/outdated value:

<hr>

For `radio` fields, however, the behaviour is different: when a non-matching value is present, the field is rendered as if the tag was not present at all. In the screenshot below assume the feature is tagged with a hypothetical tag `lifeguard=outpost`, but it would also look like this for a tag with a typo `lifeguard=Tower`.

There should be at least a visual representation that the tag is present, but does not match the expected values (e.g. as a grayed out state and a corresponding tooltip / hover text).

Alternatively, we could do this like the `check` fields: Just add the unexpected value as a temporary option, like in this mock-up:

bhavyaKhatri2703:
fixes #12082 
This PR allows showing invalid/unexpected values in radio fields as a temporary option.

Preview:

https://github.com/user-attachments/assets/63a96ee6-52fd-458e-89b5-943778a7c7e1

### DIFF
--- a/modules/ui/fields/radio.js
+++ b/modules/ui/fields/radio.js
@@ -5,6 +5,7 @@ import { presetManager } from '../../presets';
 import { t } from '../../core/localizer';
 import { uiField } from '../field';
 import { utilArrayUnion, utilRebind } from '../../util';
+import { formatTag } from './tag_title';
 
 
 export { uiFieldRadio as uiFieldStructureRadio };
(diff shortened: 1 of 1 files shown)


tordans:
Can we align this with the title tag changes that we merged recently. The tooltip for all options should be 'Tag key=value' and if available a description (taginfo or tagging schema)

Also, could you show in screenshot how this would looks with the monospace font for the not pre-defined value.

bhavyaKhatri2703:
> Also, could you show in screenshot how this would looks with the monospace font for the not pre-defined value.

image

bhavyaKhatri2703:
> Can we align this with the title tag changes that we merged recently. The tooltip for all options should be 'Tag key=value' and if available a description (taginfo or tagging schema)

added a commit for this:

screenshot

tyrasd:
I implemented this slightly differently (using proper d3 data binding patterns instead of the `wrap.selectAll('label.unknown-radio-option').remove();` workaround which could be fragile in the future) in 58b22fd20, and separately cherry-picked the added tooltips commit (68836fd0bf35e585001ff22c75bec1fb2bc15d0e) as cb414f5a9.

Issue #9733, opened 2023-07-06
Helper features for direction value (orthogonal direction, toward nearest point at line, etc.)


PeterBrodersen:
### Description

tl;dr: Entering directions numerically as degrees leads to guesses, trial-and-error. Many directions could be derived by existing features or by use of aerial imagery and simply pointing in the correction direction.

More detailed:
As directions are usually parallel or orthogonal with existing features it would be helpful to use derived values from these features as options.

At the moment it is hard to figure out whether a chosen direction matches existing features' directions (e.g. sides of a building where the direction is not specified as a tag, but could be derived from the two points for that line segment). Currently the [direction](https://wiki.openstreetmap.org/wiki/Key:direction) feature simply accepts numerical input as well as increment/decrement in 5 degree steps. It is not easy to know the correct direction in advance.

A workaround is to move the new feature next to the existing feature and changing the direction manually and simply eyeballing whether the the circle sector ⌔ lines up nicely. But this is cumbersome and usually requires trial-and-error when entering a specific direction.

This might break the simplicity principle of iD. However these are mainly suggestions for ease of use and no more complicated in the user interface than adding a new context menu feature. iD can already derive that information for existing features and we should make use of that.

## Using existing non-connected lines
Several features usually face specific other features, e.g. benches that often face a path, road or pond, or face directly away from the side of a building.

Some features usually run in parallel with specific other features, such as a [traffic sign mapped as a separate node](https://wiki.openstreetmap.org/wiki/Key:traffic_sign#As_a_separate_node).

## Placements directly on walls and other ways
A [plaque](https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dplaque) or other feature hanging on a wall would usually face in an orthogonal direction. Based on the type of feature the default suggestion could be in the direction e.g. facing away or into/inside a building wall/way.

This would simply be a suggestion of (calculated way direction ± 90 degrees). For a corner it might be the average of the two directions (again, facing inward or outward based on the circumstances).

# Implementation
Following the simplicity principle of iD this should not be more advanced than e.g. moving a node.

My suggestion is to add a "Set direction" option when right clicking a node.

* The direction should be facing the user's pointer (making it easy to set a correct direction based on aerial imagery as opposed to guesstimate degrees).
* If the pointer hovers over a point the direction should face the point.
* If the pointer hovers over a line the direction should face the nearest point on that line segment.

[...]

PR #12104, opened 2026-03-25
Add direction dial for numeric `*:direction` fields

### LINKED ISSUE #5530: Better angle field editor for direction tag
Perhaps a circle you can click next to the field that sets the angle in that direction for you, then adjust further with the arrows as needed. Something like this (although probably on the other side now that I see it):

mockup

tordans:
Closes https://github.com/openstreetmap/iD/issues/5530
Ping https://github.com/openstreetmap/iD/issues/9733, https://github.com/openstreetmap/id-tagging-schema/pull/2081

**This PR adds a dial control to the numeric **direction** fields.**
_Features:_
- Users can drag the dial to change the value in steps of 1 (regardless of the specified [`increment` property](https://github.com/ideditor/schema-builder?tab=readme-ov-file#increment)
- Users can hold down shift to change values only based on the `increment` (in the video that is 5)
- Clicking into the dial changes the value to where the user clicked
- The state is synced between map, fields, dial

Site note: The `increment` features works differently on the dial than on the `+/-` buttons. The buttons increase in +/- steps so 3 becomes 3+5=7. The dial jumps to either 5 or 10. — I think this inconsistency is OK. We keep the backward compatibility for the buttons and the dial works more natural like this IMO.

_Video:_

https://github.com/user-attachments/assets/c52a89dc-7eab-4326-a694-c7a179729e85

----

This code is created using LLM tools. I reviewed it and iterated on it for a while to get into a shape that I find easy to understand and that has a clean split between the existing features and the new additions.

Regardless, I will keep this as a draft first so we can **discuss the UX and possible edge cases** that I missed, first. But if anyone want to, the code can be reviewed from my point of view.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1504,43 +1504,27 @@ button.preset-reset .label.flash-bg {
 }
 
 .form-field-has-direction-dial {
-    display: grid;
-    grid-template-columns: minmax(0, 1fr) auto;
-    grid-template-areas:
-        "label dial"
-        "input dial";
-    align-items: stretch;
-    column-gap: 0;
-}
-
-.form-field-has-direction-dial > .field-label {
-    grid-area: label;
-    border-radius: 4px 0 0 0;
-}
-
-.form-field-has-direction-dial > .form-field-input-wrap {
-    grid-area: input;
-    border-radius: 0 0 0 4px;
+    /* Keep hook class for direction fields; layout is handled in input row. */
 }
 
 .form-field-has-direction-dial .direction-dial-wrap {
-    grid-area: dial;
     display: flex;
     flex-flow: row nowrap;
     flex: 0 0 auto;
-    width: min-content;
+    width: 32px;
     border: 1px solid var(--border-color);
     border-left: 0;
-    border-radius: 0 4px 4px 0;
-    background: var(--bg-color-2);
+    border-top-width: 0;
+    border-radius: 0;
+    background: var(--bg-color);
     justify-content: center;
     align-items: center;
-    padding: 3px;
+    padding: 2px;
 }
 
 .direction-dial {
-    width: 55px;
-    height: 55px;
+    width: 28px;
+    height: 28px;
     cursor: crosshair;
     outline: none;
 }
@@ -1553,7 +1537,8 @@ button.preset-reset .label.flash-bg {
 }
 
 @media (hover: hover) {
-    .direction-dial:not(.disabled):hover .direction-dial-ring {
+    .direction-dial:not(.disabled):hover .direction-dial-ring,
+    .direction-dial:not(.disabled).dragging .direction-dial-ring {
         stroke: var(--border-color);
     }
 }
@@ -1562,13 +1547,12 @@ button.preset-reset .label.flash-bg {
     stroke: var(--border-color);
 }
 
-.form-field-has-direction-dial .field-label .tag-reference-button {
-    border-top-right-radius: 0;
-    border-bottom-right-radius: 0;
+.form-field-has-direction-dial .form-field-input-wrap > button:last-of-type {
+    border-bottom-right-radius: 4px;
 }
 
-.form-field-has-direction-dial .form-field-input-wrap > button.increment:last-of-type {
-    border-bottom-right-radius: 0;
+.form-field-has-direction-dial .form-field-input-wrap > .form-field-button {
+    padding: 0;
 }
 
 .direction-dial-inner {
@@ -1581,7 +1565,8 @@ button.preset-reset .label.flash-bg {
 }
 
 @media (hover: hover) {
-    .direction-dial:not(.disabled):hover .direction-dial-center-dot {
+    .direction-dial:not(.disabled):hover .direction-dial-center-dot,
+    .direction-dial:not(.disabled).dragging .direction-dial-center-dot {
         fill-opacity: 1;
     }
 }
@@ -1599,7 +1584,8 @@ button.preset-reset .label.flash-bg {
 }
 
 @media (hover: hover) {
-    .direction-dial:not(.disabled):hover .direction-dial-line {
+    .direction-dial:not(.disabled):hover .direction-dial-line,
+    .direction-dial:not(.disabled).dragging .direction-dial-line {
         stroke-opacity: 1;
     }
 }
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1548,7 +1548,7 @@ button.preset-reset .label.flash-bg {
 .direction-dial-ring {
     fill: none;
     stroke: transparent;
-    stroke-width: 2;
+    stroke-width: 1;
     pointer-events: none;
 }
 
@@ -1558,7 +1558,6 @@ button.preset-reset .label.flash-bg {
     }
 }
 
-.direction-dial:not(.disabled):focus .direction-dial-ring,
 .direction-dial:not(.disabled):focus-visible .direction-dial-ring {
     stroke: var(--border-color);
 }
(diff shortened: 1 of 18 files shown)


tordans:
One thing I will likely change is: The direction indicator in the dial should IMO be dark gray like the undo, info-i etc. button until the user hovers it. This will tune the UI element back a bit and align it better with how the other buttons/interaction elements work.

tordans:
_Testing: Multiple values_ ✅

The dial is deactivated in this case. The cursor on hover on the dial is the "cannot do anything" cursor.
> image

---

_Test-Link:_ https://pr-12104--ideditor.netlify.app/#disable_features=boundaries&map=20.01/54.20815/10.42527&background=Bing&locale=en&id=n-1,n-2 => Create a traffic sign for example.

tordans:
One question that was raised during development was, if we should extend the fields like https://github.com/openstreetmap/id-tagging-schema/blob/main/data/fields/direction.json with [`minValue`](https://github.com/ideditor/schema-builder?tab=readme-ov-file#minvalue), `maxValue` properties to make sure users only put propper values in. The code guards against wird values, but still we have those properties so any reason not to use them? – I would create a PR in the tagging schema if agreed.

modules/ui/direction_field_dial.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Returns a function called on each `i(selection)` pass: updates the dial shell for `fieldRoot`
+ * and returns `sync` to run after `i.tags(…)`.
+ */
+export function createDirectionFieldDialMount(
+    enabled: boolean,
+    options: DirectionFieldDialMountOptions
+): (fieldRoot: FormFieldRootSelection) => SyncDirectionDial {
+    if (!enabled) {
+        return function mountDisabled(fieldRoot: FormFieldRootSelection) {
```
tordans, line 38: Unsure about the way the code is split out from `input.js`. A function that returns a function is not ideal from a complexity point of view. But it does help to keep the changes to `input.js` small. Another option would be to inline this in `input.js` and look into extracting a bit of helper code out where possible.

css/80_app.css
```diff
@@ -1503,6 +1503,127 @@ button.preset-reset .label.flash-bg {
     border-radius: 0;
 }
 
+.form-field-has-direction-dial {
```
1ec5, line 1506: As currently designed, the dial occupies the right end of the field’s header bar, insetting the bar to the left. This is a clever way to maximize the dial’s height while minimizing the overall field’s height. However, I’m concerned that inconsistent placement of the header bar could lead users to misclick. #12005 would add more buttons to the bar, making consistency even more desirable.

Since the dial can apparently track the cursor outside of its bounds, maybe we don’t really need to make the dial that big. But if we do, I think we could get away with just making the field taller. It would give us space for a Reverse Direction button similar to the one in the One Way field. I hit <kbd>V</kbd> all the time on directional features. For example, I’ll often map a traffic sign and duplicate it to represent the same sign facing the other side of the road.

tordans: > However, I’m concerned that inconsistent placement of the header bar could lead users to misclick. 
> …
> Since the dial can apparently track the cursor outside of its bounds, maybe we don’t really need to make the dial that big. But if we do, I think we could get away with just making the field taller.

We can try this, but I do not expect it to look great.
* If the dial is just the height of the input, I expect it to be too small to feel like a nice UI; the shift-case would not look good either.
* Making the field bigger in height did look weird
* My ideal from `12005`: I think we need to use a different place for those Buttons anyway (see https://github.com/openstreetmap/iD/pull/12005#issuecomment-4060712673) and also the direction field should not have those extra options (its not something that usually has a side, subtags or similar. Maybe source.

> It would give us space for a Reverse Direction button similar to the one in the One Way field. I hit V all the time on directional features. For example, I’ll often map a traffic sign and duplicate it to represent the same sign facing the other side of the road.

Like the idea. But we might want to track this separately. One idea I have is to remove the undo button, which AFAIR we only have for this field(?) and use that space for a new button. Or an inline "labe/button" in the field like the "reverse"-text-button for `oneway`.

tordans:
Quick notes from a talk with @tyrasd at [a meetup](https://wiki.openstreetmap.org/wiki/FOSSGIS_2026/OSM-Samstag):
- Maybe remove undo option which we only (?) have in this field; frees up space
- No new idea on a better position / layout – stick with what we have in the preview now
- Add support for ranges like `120-210`; First click set the line (one value), then `option`+drag to change the cone size; Drag without `option` moves the values as is with different from-to values
- Leave cone  on map as is; will have many edge cases if we try to sync the dial style with the map

tyrasd:
for the UI it would be advantageous to keep the info- :information_source:  button aligned on the right hand edge of the field. Maybe the field height could just be increased slightly to fit the "compass" under the :information_source: button?

tordans:
**Here is an update on this UI:**
I converted this into a draft because this is now in UI-discovery phase again. I did not review the code, just the UX and UI. Please provide feedback on those parts of the PR.

**Main changes:** 
- The Dial is now in the input field line
- I gets bigger when users leave the dial space

I did not like the increase in height for the input; but I think the dial increase is a good workaround. 
To trigger the increase you have to click the dial, then move the cursor outside the tiny circle (which happens by itself when using it), then it increases.

This also supports ranges, now. However, looking at the data I wonder if that is actually a good thing to add in terms of code complexity. It does work quite nicely once one has the shortcuts figured out. But the data does not show any relevant usage numbers on ranges: https://taginfo.openstreetmap.org/keys/direction#values
I suggest to remove this feature again an table it for a later addition.

The Info-i section now also has a "how to use this" line to explain the interaction.

**Video:**

https://github.com/user-attachments/assets/0b1f1235-70a1-42dc-a45e-1024d222133b

**Screenshots:**

* regular 
    1 dial regular
* with ticks - shift
   2 dial shift - ticks
* big  with ticks (shift pressed) - the non ticks version is missing as a  screenshot
  3 dial big with ticks

ranges:
* 4 dial big with range
* 5 dial big with range and ticks


* info-i
   info-i

Zaneo:
Does this change the behavior for objects where `forward` or `backward` is the preferred option instead of a degrees number?

tordans:
@Zaneo Please use the preview at https://pr-12104--ideditor.netlify.app/ to test this

Zaneo:
Cool seems to use the correct behavior when the node is on a way ( does not
use dial).

1ec5:
> This also supports ranges, now. However, looking at the data I wonder if that is actually a good thing to add in terms of code complexity. It does work quite nicely once one has the shortcuts figured out. But the data does not show any relevant usage numbers on ranges: https://taginfo.openstreetmap.org/keys/direction#values
> I suggest to remove this feature again an table it for a later addition.

Agreed, I don’t know of any tagging scheme that needs angle ranges. Besides, the map doesn’t render a cone for the angle when it’s set to a range like that.

> big with ticks (shift pressed)

I wonder if we could get away with something less bold. When you rotate an area on the map that’s about the same size as the compact version of this control, the area responds to mouse movements without needing the extra visual feedback of a big dial. Essentially, it’d be what you have, except that the big dial would be transparent, relying on the tick mark in the small dial to convey the current direction. (By the way, it isn’t possible to rotate a point-with-cone on the map: #12341.)

tordans:
**A new iteration to test from a UX and UI point of view:**

1. The range UI is removed again
2. The size increase is smaller but still there; I find it very helpfull…
3. Now the `R` rotate action/shortcut work when a point has a `direction` tag - https://github.com/openstreetmap/iD/issues/12341 - the video shows the different messages in the status bar at the end (cut of a bit)
4. Code is cleaned up a bit… – _Note_ In the screenrecording there are decimal places; I fixed those.

https://github.com/user-attachments/assets/f6541116-8b1e-4e08-b5cc-feb12e777241

danieldegroot2:
> I don’t know of any tagging scheme that needs angle ranges

@1ec5 at least cameras, viewpoints and bincolars schemas have the option to use ranges as further detailing the area as opposed to the direction which it is viewing.
for cameras it is usually unknown or derivable and rarely added because staying longer at such POI is impossible or questionable.
for viewpoints they are short stops along a route, it may be hard to answer solely based on zoomed in imagery and consumer apps do not allow editing it. also it is slightly subjective for larger viewing platforms, but it is not a big issue.
for binoculars they may have limited/unlimited rotation, in addition to having a limited viewing scope (which is not tagged).

tordans:
@danieldegroot2 when looking at what we support, we are closer to "what is documented" and "what is used than "what is possible in theory". So we need some numbers on tag usage, examples, wiki entries as reference…

---

FYI: My next step for this PR is, to split it in two. One for just the shortcut that allows to rotate the direction with `R`; one for the UI. The shortcut will likely land first and might very well be enough for the time being as it covers most use cases I am looking form. (FYI: This feature is already in the latest GoMap Demo.)

tordans:
I extracted the `R` shortcut part into https://github.com/openstreetmap/iD/pull/12673
I hope this much smaller feature can be merged first - and faster.
We can still do this here, but TBH I think we have more important feature to merge first…

PR #12107, opened 2026-03-26
Fix: remove localized and prefix-based tags when switching presets (#12075)

### LINKED ISSUE #12075: Handle special fields like `localized` names in tag removal logic
More follow-ups for #11696: 

* [ ] for fields of type `localized`: When changing a feature from a preset that had a `name` field (of type `localized`) and one changes it to a preset without a corresponding name field, only the base `name` tag is removed during the tag upgrade, but any localized names are preserved. Example: `craft=key_cutter name=foo name:en=bar` when changed to a recycling bin will end up as `amenity=recycling recycling_type=container name:en=bar`). This should consistently remove either all names or none.
* [ ] for fields of type `multiCombo` it does not remove the old tags when the new preset does not have a field for the corresponding tag(s). Example: `amenity=recycling recycling_type=container
 recycling:plastic=yes recycling:=invalid` ends up as `craft=key_cutter recycling:plastic=yes` (note that only the `recycling:=invalid` was removed because that's how the tag key prefix is defined in the field's `key` property). This should remove all corresponding tags with a key that matches the field's prefix.
* [ ] `directionalCombo` fields that implicitly also support `:both`
* more?

aditya1010-max:
Fix #12075

## Description

When switching presets, iD removes only the **base field keys**, but does not remove related tags belonging to the same namespace.

### Localized fields 

Only the base key (`name`) is removed, while language variants remain.

**Example:**

Before:
```
craft=key_cutter
name=foo
name:en=bar
name:hi=baz

```
After switching to recycling:
```
amenity=recycling
recycling_type=container
name:en=bar   ❌
name:hi=baz   ❌
```

### Prefix-based fields (multiCombo / key: 'prefix:')

Tags sharing a prefix are not removed.

**Example:**

Before:
```
amenity=recycling
recycling:plastic=yes
recycling:glass=yes
recycling:paper=yes
```
After switching to key cutter:
```
craft=key_cutter
recycling:plastic=yes   ❌
recycling:glass=yes     ❌
recycling:paper=yes     ❌
```

---

### Why this happens

The current implementation only removes **exact field keys**, but does not account for:

* localized namespaces (name:*)
* prefix-based namespaces (recycling:*)

As a result, related tags remain after the preset change.

## Solution

This PR expands the removal logic to handle full tag namespaces:

- Identify fields present in the old preset but not preserved in the new preset
- For each removed field:

  * **Localized fields (`type: localized`)**

    * remove: `name` and all `name:*`
  * **Prefix-based fields (`key` ends with `:`)**

    * remove: all tags matching the prefix (`recycling:*`)
- Collect all keys and remove them using `utilObjectOmit`

### Additional improvements:

* Use 'Set' for efficient lookups
* Cache `Object.keys(tags)` to avoid repeated iteration
* Filter fields using `matchGeometry(geometry)` for consistency with existing logic

## Demo Video(After)-

https://github.com/user-attachments/assets/930c4cb7-93fa-4b02-928b-c6f38a29ccb2

## Test

Added two tests to validate the behavior:

**Localized fields test**

  * Confirms `name` and all `name:*` tags are removed

**Prefix-based fields test**

  * Confirms all `recycling:*` tags are removed
  * Ensures unrelated tags are preserved

### Review appreciated, especially on:

- overall approach and correctness of the logic
- potential optimizations or edge cases
- test coverage and mocking strategy used in the test file

---

### DIFF
--- a/modules/actions/change_preset.js
+++ b/modules/actions/change_preset.js
@@ -26,15 +26,56 @@ export function actionChangePreset(entityID, oldPreset, newPreset, skipFieldDefa
             }
 
             if (oldPreset && (oldPreset.id !== newPreset.id)) {
-                // 'field-keys' are keys used by fields (different to the keys used by preset itself)
-                const oldPresetFieldKeys = [
+                const oldFields = [
                     ...oldPreset.fields(loc),
                     ...oldPreset.moreFields(loc)
-                ].flatMap(f => f.allKeys());
+                ].filter(f => f.matchGeometry(geometry));
 
-                // field-keys used by the old preset but not the new preset
+                const oldPresetFieldKeys = oldFields.flatMap(f => f.allKeys());
+
                 const fieldKeysToRemove = utilArrayDifference(oldPresetFieldKeys, preserveKeys);
-                tags = utilObjectOmit(tags, fieldKeysToRemove);
+                const fieldKeysToRemoveSet = new Set(fieldKeysToRemove);
+
+                const expandedKeysToRemove = new Set();
+                const tagKeys = Object.keys(tags);  // Cache all existing tag keys once to avoid repeated Object iteration
+
+                // Only consider fields that are actually being removed
+                const fieldsToRemove = oldFields.filter(field => {
+                    const keys = field.allKeys();
+                    return keys.some(k => fieldKeysToRemoveSet.has(k));
+                });
+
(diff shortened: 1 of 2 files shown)


tyrasd:
Replaced by #12218 because this logic is quite important to get right

PR #12111, opened 2026-03-26
fixed bug in `canSumTags` logic: correctly determine numeric tag values

### LINKED ISSUE #12110: merging features with non-numeric `step_count` can lead to NaN values

* have two ways tagged as `highway=steps`, one with `step_count=1`, one with `step_count=no`
* merge the to features
* the resulting value for the `step_count` field is now `NaN`

JaiswalShivang:
Fixes #12110 

### Description

In `modules/actions/join.js`, the `canSumTags()` helper function has a misplaced parenthesis that causes the `isFinite()` check on `tagsA[key]` to be completely bypassed.

`tagsA[key] && isFinite(tagsB[key])` is evaluated first — producing a boolean

### Current Code (Line 211-215)

```js
function canSumTags(key, tagsA, tagsB) {
    return osmSummableTags.has(key) &&
        isFinite(tagsA[key] &&
        isFinite(tagsB[key]));
}

### DIFF
--- a/test/spec/actions/join.js
+++ b/test/spec/actions/join.js
@@ -611,40 +611,37 @@ describe('iD.actionJoin', function () {
         expect(graph.entity('-').tags.step_count).to.equal('42');
     });
 
-    it('does not produce NaN when the first way has a non-numeric summable tag', function () {
-        let graph = iD.coreGraph([
+    it('returns conflicting_tags when the first way has a non-numeric summable tag', function () {
+        var graph = iD.coreGraph([
             iD.osmNode({ id: 'a', loc: [0, 0] }),
             iD.osmNode({ id: 'b', loc: [1, 0] }),
             iD.osmNode({ id: 'c', loc: [4, 0] }),
             iD.osmWay({ id: '-', nodes: ['a', 'b'], tags: { highway: 'steps', step_count: 'many' } }),
             iD.osmWay({ id: '=', nodes: ['b', 'c'], tags: { highway: 'steps', step_count: '10' } })
         ]);
-        graph = iD.actionJoin(['-', '='])(graph);
-        expect(graph.entity('-').tags.step_count).not.to.equal('NaN');
+        expect(iD.actionJoin(['-', '=']).disabled(graph)).to.equal('conflicting_tags');
     });
 
-    it('does not produce NaN when the second way has a non-numeric summable tag', function () {
-        let graph = iD.coreGraph([
+    it('returns conflicting_tags when the second way has a non-numeric summable tag', function () {
+        var graph = iD.coreGraph([
             iD.osmNode({ id: 'a', loc: [0, 0] }),
             iD.osmNode({ id: 'b', loc: [1, 0] }),
             iD.osmNode({ id: 'c', loc: [4, 0] }),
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by matkoniecz** [comment]
Can you add test covering this code?

**Comment by matkoniecz** [comment]
Though I can confirm that fix works: 

```
highway=steps
step_count=bbb
```

and

```
highway=steps
step_count=111
```

merge is now being blocked

rather than giving `step_count=NaN`

**Comment by JaiswalShivang** [comment]
@matkoniecz Done, added tests for non-numeric summable tag values.

File: test/spec/actions/join.js [hunk 1]
```diff
@@ -611,6 +611,42 @@ describe('iD.actionJoin', function () {
...
+    it('does not produce NaN when the first way has a non-numeric summable tag', function () {
+        let graph = iD.coreGraph([
+            iD.osmNode({ id: 'a', loc: [0, 0] }),
+            iD.osmNode({ id: 'b', loc: [1, 0] }),
+            iD.osmNode({ id: 'c', loc: [4, 0] }),
+            iD.osmWay({ id: '-', nodes: ['a', 'b'], tags: { highway: 'steps', step_count: 'many' } }),
+            iD.osmWay({ id: '=', nodes: ['b', 'c'], tags: { highway: 'steps', step_count: '10' } })
+        ]);
+        graph = iD.actionJoin(['-', '='])(graph);
+        expect(graph.entity('-').tags.step_count).not.to.equal('NaN');
```
matkoniecz: I though rather about testing modified function directly - or to confirm that merge has not happened at all (`.not.to.equal('NaN');` is not great as other bogus results are also possible)

JaiswalShivang: ok thanks for the suggestion

JaiswalShivang:
@matkoniecz sir updated the tests

tyrasd (approved):

matkoniecz (commented):
Looks good to me (note: I have not run tests and I am away from PC right now)

tyrasd:
The fix looks good. Good catch!

There seem to be luckily not too many affected map features resulting from this bug: https://taginfo.openstreetmap.org/tags/step_count=NaN

PS: @JaiswalShivang please don't use the solution of the bug as the issue/PR titles and descriptions, even if you already know the cause of the bug and how it has to be fixed: Instead, always include the minimal steps required to reproduce the bug in the issue, and always use a description of the bug as the issue title. PSS: one final nit-pick: you might want to take a look at how markdown [formatting](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) works on github. You seem to sometimes mistake `` ` `` with `[`. :wink:

k-yle:
there's still 21 examples of [`step_count=NaN`](https://taginfo.openstreetmap.org/keys/step_count) in osm, all of which were caused by iD. should we do a manual cleanup? someone on the osm discord just noticed

matkoniecz:
> should we do a manual cleanup? 

Yes. I have it on my TODO list (together with 3000+ other things, so feel free to look into this if anyone wants)

(I removed some)

PR #12114, opened 2026-03-27
Add comment explaining meaning of fields

matkoniecz:


tyrasd:
note to self for later: we should also add type definitions for setting variables like this one.

PR #12115, opened 2026-03-27
render labels (and icon) for animal=horse_walker features

### LINKED ISSUE #11975: Make icon appear for new preset animal=horse_walker always
A new preset https://github.com/openstreetmap/id-tagging-schema/pull/1857 was merged, but it needs something to be done here so that the icon always appears. Currently the icon only appears if the tag `animal=horse_walker` is acompanied by some other tag like `building=*` or `leisure=*`, example here: https://ideditor.netlify.app/#disable_features=boundaries&map=21.00/50.76960/-2.20871&background=MAPNIK&id=w1347874758

The reason why this is so and what needs to be done was discussed here a bit:
https://github.com/openstreetmap/id-tagging-schema/pull/1857#issuecomment-3677536591 ff

matkoniecz:
fixes #11975

### DIFF
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -46,7 +46,7 @@ export function svgLabels(projection, context) {
         ['area', 'craft', '*', 12],
         ['area', 'tourism', '*', 12],
         ['area', 'camp_site', '*', 12],
-        ['area', 'animal', '*', 12],
+        ['area', 'animal', 'horse_walker', 12],
         ['point', 'aeroway', '*', 10],
         ['point', 'amenity', '*', 10],
         ['point', 'building', '*', 10],
--- a/modules/svg/labels.js
+++ b/modules/svg/labels.js
@@ -46,6 +46,7 @@ export function svgLabels(projection, context) {
         ['area', 'craft', '*', 12],
         ['area', 'tourism', '*', 12],
         ['area', 'camp_site', '*', 12],
+        ['area', 'animal', '*', 12],
         ['point', 'aeroway', '*', 10],
         ['point', 'amenity', '*', 10],
         ['point', 'building', '*', 10],


matkoniecz:
screen-2026-03-27-08-11-09

modules/svg/labels.js
```diff
@@ -46,6 +46,7 @@ export function svgLabels(projection, context) {
         ['area', 'craft', '*', 12],
         ['area', 'tourism', '*', 12],
         ['area', 'camp_site', '*', 12],
+        ['area', 'animal', '*', 12],
```
tyrasd:
```suggestion
        ['area', 'animal', 'horse_walker', 12],
```

shouldn't we limit this to the specific tag in question?

matkoniecz: I think that if any other `animal` preset has defined icon it also should be shown, and if it has it not defined this change has no effect.

Note that other keys also have catch-all `*`

tyrasd: yes, but we might also not. What other presets do (potentially) exist that use the `animal` key as their primary tag?

matkoniecz: animal=school
animal=cemetery

https://taginfo.openstreetmap.org/keys/animal#values

tyrasd: > animal=school

you mean [`amenity=animal_training`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Danimal_training)?

> animal=cemetery

this is only used 200 times globally and poorly documented, so it's not detrimental to omit this value for now.

> https://taginfo.openstreetmap.org/keys/animal#values

Almost all of the remaining values seem to be cases where the `animal` tag is used as an attribute (e.g. to indicate which kind of animals are bread at a particular `amenity=animal_breeding` facility). Those should definitely not be labeled. Also note that none of the tags are listed on https://wiki.openstreetmap.org/wiki/Map_features – IMHO only primary tags should be considered for labeling.

The conceptual problem with using a wildcard for tags that are a mix of attribute tags and (in a few cases) primary tags, is that it might lead to the case where a mistagged feature such as `name="foo cafe" + animal=no` will be labeled as if it were correctly tagged.

matkoniecz: It would not be getting icon anyway,  so it can also serve as an indicator of problematic data 

Not sure which effect is greater

I guess I will change it,  though it would be a bit unique case among other listed keys

matkoniecz:
Note: amendment is not tested

matkoniecz:
Tested, works

tyrasd (approved):

PR #12118, opened 2026-03-27
Fix incorrect sort comparator in crossing_ways

### LINKED ISSUE #12119: Incorrect sort comparator in crossing_ways.js
### Summmary

The comparator used in `crossing_ways.js` for Array.prototype.sort() returns boolean values (`true`/`false`) instead of numeric values.

Since JavaScript coerces these to `1` and `0`, the comparator does not return negative values when required, which may lead to incorrect or unstable sorting behavior.

This does not follow the expected comparator contract defined in the JavaScript specification.

### Steps to reproduce

1. Inspect the comparator used in `crossing_ways.js`
2. Observe that `Array.prototype.sort()` is using a comparator that returns boolean values (`true`/`false`)
3. In JavaScript, these values are coerced to `1` and `0`, which does not satisfy the requirement for negative/zero/positive return values

### Example

```js
[3, 2, 1].sort((a, b) => a < b);

🔹 Expected behavior

```md
The comparator should return:
- a negative value when `a < b`
- a positive value when `a > b`
- 0 when equal

This ensures correct and stable sorting behavior as per JavaScript specification.

nisshchayarathi:
Fixes: #12119 
The current Array.prototype.sort() comparator in crossing_ways.js returns boolean values (true/false), which are coerced to 1/0.

This can lead to incorrect and unstable sorting behavior because the comparator never returns negative values, which are required to correctly order elements in JavaScript's sort algorithm.

### Before
Comparator returned boolean values:
- true → 1
- false → 0

### After
Comparator now returns proper numeric values:
- negative (a < b)
- positive (a > b)
- zero (equal)

This PR updates the comparator to return proper numeric values (-1, 0, 1), ensuring consistent and correct ordering.

✅ Fixes unstable sorting  
✅ Aligns with JavaScript sort specification

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -376,22 +376,6 @@ export function validationCrossingWays(context) {
 
 
     function createIssue(crossing, graph) {
-
-        // use the entities with the tags that define the feature type
-        crossing.wayInfos.sort(function(way1Info, way2Info) {
-            var type1 = way1Info.featureType;
-            var type2 = way2Info.featureType;
-            if (type1 === type2) {
-                var label1 = utilDisplayLabel(way1Info.way, graph);
-                var label2 = utilDisplayLabel(way2Info.way, graph);
-                return label1 > label2 ? 1 : label1 < label2 ? -1 : 0;
-            } else if (type1 === 'waterway') {
-                return 1;
-            } else if (type2 === 'waterway') {
-                return -1;
-            }
-            return type1 > type2 ? 1 : type1 < type2 ? -1 : 0;
-        });
         var entities = crossing.wayInfos.map(function(wayInfo) {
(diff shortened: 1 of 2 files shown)


matkoniecz:
As I understand, problem here is theoretical though fixing it would be fine. 

Have you tested to confirm that code continues to work?

nisshchayarathi:
@matkoniecz yes i have tested the code continues to work

tyrasd:
I don't see why these entries need to be sorted alphabetically in the first place. IMO that's not needed and should be removed completely. The idea behind the special casing for `waterway` seems to enforce that waterway features are always the first (or last) entry in the validation message. Maybe it sounds better in English if you say "Stream crosses Building" instead of the other way around (or vice versa). But as that does seem very language-specific, I think it is not worth keeping around.

nisshchayarathi:
Makes sense

PR #12129, opened 2026-03-29
fix: correct loop increment in move-limiting logic

### LINKED ISSUE #12128: Incorrect loop increment in move-limiting logic
### Description
The move-limiting logic contains an issue in an inner for loop where the wrong variable is being incremented. This can lead to incorrect iteration behavior and potentially skip or mis-handle elements during move restriction checks.

1.Open the iD editor
2.Perform an action that triggers move-limiting (e.g., moving features with constraints)
3.Observe inconsistent or incorrect behavior in how movement is restricted

### Expected Behavior
The loop increments the wrong variable, which may:
Skip elements
Cause incorrect validations
Lead to inconsistent move-limiting results

### Proposed Fix
Update the inner loop to increment the correct variable. I can submit a PR with the fix and a small regression test to prevent this issue in the future.

nisshchayarathi:
Fixes #12128

## Description
This PR fixes an issue in the move-limiting logic where the inner loop was incrementing the wrong variable, leading to incorrect iteration behavior.

## Changes
- Corrected the loop increment variable

## Testing
- Verified behavior manually
- Ran existing test suite to ensure no regressions

### DIFF
--- a/test/spec/actions/move.js
+++ b/test/spec/actions/move.js
@@ -71,4 +71,57 @@ describe('iD.actionMove', function() {
         expect(loc[0]).to.be.closeTo( 1.440, 0.001);
         expect(loc[1]).to.be.closeTo(-2.159, 0.001);
     });
+
+    it('limits delta when movedPath and unmovedPath intersect twice', function() {
+        // Use a simple planar projection so we can reason about intersections in 2D.
+        var planar = function(p) { return p; };
+        planar.invert = function(p) { return p; };
+
+        // moved way is a single segment (b -> d)
+        var b = iD.osmNode({ loc: [0, 0] });
+        var d = iD.osmNode({ loc: [10, 0] });
+        var moved = iD.osmWay({ nodes: [b.id, d.id] });
+
+        // Unmoved zigzag way contains b as an interior node.
+        // With tryDelta [5,5], the moved segment becomes y=5 from x=5..15.
+        // That segment intersects the zigzag in two places: at (5,5) and (10,5).
+        var u0 = iD.osmNode({ loc: [-10, 10] });
+        var u1 = iD.osmNode({ loc: [10, 10] });
+        var u2 = iD.osmNode({ loc: [10, 0] });
+        var zigzag = iD.osmWay({ nodes: [u0.id, b.id, u1.id, u2.id] });
+
+        // Unmoved vertical way contains d as an interior node.
+        // With tryDelta [5,5], the moved endpoint would cross this way, so x movement should be limited.
+        var v0 = iD.osmNode({ loc: [10, -10] });
+        var v1 = iD.osmNode({ loc: [10, 10] });
+        var vertical = iD.osmWay({ nodes: [v0.id, d.id, v1.id] });
+
+        var graph = iD.coreGraph([
+            b, d, moved,
+            u0, u1, u2, zigzag,
+            v0, v1, vertical
+        ]);
+
+        // Explicitly verify the scenario: with the attempted delta, the moved segment
+        // intersects the unmoved zigzag in exactly two points.
+        var movedPath = [b.loc, d.loc].map(function(p) { return [p[0] + 5, p[1] + 5]; });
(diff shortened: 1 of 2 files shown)


nisshchayarathi:
@tordans @tyrasd LGTM

matkoniecz:
1) pinging maintainers is not needed
2) have you tested is it working add intended by testing whether program runs as intended by using specific code?

Note that tests also passed before, so this specific functionality is not tested at all - so tests passing does not indicate much

nisshchayarathi:
@matkoniecz the var j is defined and in for loop i is getting incremented obviously it needs to be fixed

matkoniecz:
Yes, but fix needs to be tested.

In some cases multiple bugs were masking each other, and fixing just one of them would make situation worse.

Also, testing may reveal more problems with code here. If such mistake exists then surely something went quite wrong.

tyrasd (changes requested):
This does seem like an actual mistake in the for loop variable. See https://github.com/openstreetmap/iD/commit/9b200cf514b9b5e232f3b71aa7eeb8c7913f115d#diff-e1e4ae664eecbe2722b4542db88173ce69c14f939d89784c0ad32a61dc0d6451R296

Luckily, there are not too many cases where `hits` will have more than one entry, so that's probably why it wasn't causing much troubles in practice.

However, as mentioned by @matkoniecz: we should definitely add a unit test that checks that the method is working as intended. For this we need to create a test case where `movedPath` and `unmovedPath` meet in two points.

test/spec/actions/move.js
```diff
@@ -71,4 +71,57 @@ describe('iD.actionMove', function() {
         expect(loc[0]).to.be.closeTo( 1.440, 0.001);
         expect(loc[1]).to.be.closeTo(-2.159, 0.001);
     });
-});
+
+    it('limits delta when movedPath and unmovedPath intersect twice', function() {
```
tyrasd, line 75: This test does not fail in the current implementation (without the bug fix). Please construct a test that actually fails in the current development branch, and is fixed in the PR.

nisshchayarathi: sir I checked 
 it fails without the fix: the buggy loop ,increments the outer i inside the inner loop, so it skips the later intersection object that actually constrains the x movement.
 
Screenshot 2026-03-31 153254

nisshchayarathi: If I am missing something please suggest

RudyTheDev:
I manually analyzed the scenario to see how this would actually trigger.

"Problem" 1 here is that the loop is always using `hits[0]` (i. e. `j==0`) to calculate `_delta`. However, it doesn't matter which point is used, because it calculates `_delta` from variables independent of the hit.

"Problem" 2 is that if `if (geoVecEqual(hits[0], end)) continue;` triggers, then `_delta` might not be calculated at all. However, as far as I can tell, it is impossible to actually trigger this condition in a real-world scenario. This could theoretically happen if the line hasn't moved or has been returned to the start position - but then the delta would be 0 anyway. Or in a very convoluted layout where the hit is also exactly the end point - but that's basically impossible to reproduce due to floating point precision before the heat death of the universe.

Since `_delta` is only ever overwritten, while the actual hit doesn't matter for calculation, then all the old code did was do a bunch more redundant loops, (re)setting delta.

So anyway, the actual "fix" would be to just:

```
if (hits.length) {
    var edge = geoChooseEdge(unmovedNodes, end, projection);
    _delta = geoVecSubtract(projection(edge.loc), start);
    break;
}
```

matkoniecz:
> So anyway, the actual "fix" would be to just:

can you open PR for that, if it is needed?

RudyTheDev:
> can you open PR for that, if it is needed?

@matkoniecz - #12453.

I was hoping the author would do this, but I don't suppose they will given this was all from an LLM.

matkoniecz:
> I don't suppose they will given this was all from an LLM.

it is kind of repeated event

tyrasd:
closing in favour of #12453

PR #12131, opened 2026-03-29
fix: ensure comparator returns 0 for equal values in address field so…

### LINKED ISSUE #12130: Comparator does not return 0 for equal values in address field sorting
### Description

The sort comparator used in the address field does not correctly handle equality. It currently returns -1 even when a.value === b.value, which violates the expected behavior of JavaScript’s Array.prototype.sort comparator.

1.Open the iD editor
2.Select or create a feature (node/way)
3.Go to the Address field UI
4.Ensure there are multiple address entries with the same value
(e.g., duplicate street names or similar entries)
5.Observe how they are sorted

### Problem

When a.value === b.value, the comparator still returns -1 instead of 0. This can lead to:

unstable or incorrect sorting behavior
inconsistent ordering of equal elements

### Expected Behavior

The comparator should:

return 1 when a.value > b.value
return -1 when a.value < b.value
return 0 when values are equal

nisshchayarathi:
Fixes #12130 

## Description
This PR fixes an issue in the address field sorting comparator where equal values were not handled correctly. The comparator always returned -1, even when values were equal.

## Changes
- Updated comparator to return 0 when values are equal

## Testing
- Verified sorting behavior manually
- Ensured no regressions in address field UI

### DIFF
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -187,7 +187,7 @@ export function uiFieldAddress(field, context) {
         return utilArrayUniqBy([
             ...enclosingAddresses,
             ...nearPointAddresses
-        ], 'value').sort((a, b) => a.value > b.value ? 1 : -1);
+        ], 'value').sort((a, b) => a.value === b.value ? 0 : (a.value > b.value ? 1 : -1));
     }
 
 


matkoniecz:
> Verified sorting behavior manually

Can you post screenshot showing bug? I am confused which interface part is affected and how

PR #12134, opened 2026-03-29
Improve upload explanation message to clarify visibility and refresh behavior

### LINKED ISSUE #12113: Tell the user directly that they need to reset their browser cache after editing
### Description

The user now sees the following after submitting changes:

In my experience of watching the local community chat, newbies still ask when their changes will be displayed on the map.

I think we need to give a more explicit hint. Something like:

> Your changes should appear on OpenStreetMap within a few minutes. Be sure to reload the tab [without cache](https://wiki.openstreetmap.org/wiki/Browser_cache#How_To_Clear). 
> It may take longer for maps elsewhere to receive updates. [Details](https://wiki.openstreetmap.org/wiki/FAQ#I_have_just_made_some_changes_to_the_map._How_do_I_get_to_see_my_changes.3F)

Shyam-123pandey:
This updates the upload explanation message shown after saving changes to make it clearer for new contributors.
Previously, the message stated that changes would be visible on all maps, which could be misleading and did not provide guidance on what to do if changes were not immediately visible.
This change:
- Adds a clearer expectation about timing ("within a few minutes")
- Suggests refreshing the page (including cache)
- Clarifies that other maps may take longer to update
This aims to reduce confusion frequently observed among new users regarding when their edits appear.

fixes: #12113

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -604,7 +604,7 @@ en:
     retry: Retry
   commit:
     title: Upload to OpenStreetMap
-    upload_explanation: "The changes you upload will be visible on all maps that use OpenStreetMap data."
+    upload_explanation: "Your changes should appear on OpenStreetMap within a few minutes. If you don’t see them, try refreshing the page (you may need to clear cache). It may take longer for other maps to update."
     upload_explanation_with_user: "The changes you upload as {user} will be visible on all maps that use OpenStreetMap data."
     request_review: "I would like someone to review my edits."
     request_review_info: "Unsure about something? Invite an experienced mapper to check your work once it's live."
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -605,7 +605,7 @@ en:
   commit:
     title: Upload to OpenStreetMap
     upload_explanation: "Your changes should appear on OpenStreetMap within a few minutes. If you don’t see them, try refreshing the page (you may need to clear cache). It may take longer for other maps to update."
-    upload_explanation_with_user: "The changes you upload as {user} will be visible on all maps that use OpenStreetMap data."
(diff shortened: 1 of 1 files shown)


Shyam-123pandey:
Hello Sir please review this PR.

matkoniecz:
> Hello Sir please review this PR.

Any PR made needs review, explaining this is not needed

matkoniecz:
how next texts look like in practice?

data/core.yaml
```diff
@@ -604,8 +604,8 @@ en:
     retry: Retry
   commit:
     title: Upload to OpenStreetMap
-    upload_explanation: "The changes you upload will be visible on all maps that use OpenStreetMap data."
-    upload_explanation_with_user: "The changes you upload as {user} will be visible on all maps that use OpenStreetMap data."
+    upload_explanation: "Your changes should appear on OpenStreetMap within a few minutes. If you don’t see them, try refreshing the page (you may need to clear your browser cache). It may take longer for other maps to update."
```
matkoniecz, line 607: losing "visible on all maps that use OpenStreetMap data" is sad

matkoniecz (commented):
looking at https://github.com/openstreetmap/iD/issues/12113 it seems to be modifying different text than requested

data/core.yaml
Shyam-123pandey: yes, `commit.upload_explanation` gets replaced by `commit.upload_explanation_with_user` once user details are loaded, so updating only one of them would lead to inconsistent messaging (and possibly a brief flicker of the new text before it’s overwritten).
I’ve updated both `upload_explanation` and `upload_explanation_with_user` to keep the behavior consistent for all users.
Let me know if the wording looks good or if you’d prefer any adjustments.

Shyam-123pandey:
> looking at #12113 it seems to be modifying different text than requested

Hi Sir, according to mine investigation i putted at right place, but i can't be able to test because to test this i have save but at localhost for authentication purpose giving error, so i didn't able to save the edit, unable to test, 
can u help to test this locally?

PR #12136, opened 2026-03-29
Fix: Ctrl+Shift+I should not trigger Ctrl+I

### LINKED ISSUE #10817: CTRL+SHIFT+I should not trigger CTRL+I

ahir-u:
This fixes an issue where shortcuts that do not require Shift (e.g. Ctrl+I) are still triggered when Shift is pressed. This bug does not happen on Max (with cmnd).

### Fix
Always enforce exact matching of the shift modifier.

### Result
- It explicitly checks for the status of Shift to match (both false and true)
- Ctrl+I no longer triggers on Ctrl+Shift+I
- Shift-specific shortcuts still work
- Caps Lock behavior unaffected

Fixes #10817

### DIFF
--- a/modules/util/keybinding.js
+++ b/modules/util/keybinding.js
@@ -22,7 +22,7 @@ export function utilKeybinding(namespace) {
         for (const binding of bindings) {
             if (!binding.event.modifiers.shiftKey) continue;  // no shift
             if (!!binding.capture !== isCapturing) continue;
-            if (matches(d3_event, binding, true)) {
+            if (matches(d3_event, binding)) {
                 binding.callback(d3_event);
                 didMatch = true;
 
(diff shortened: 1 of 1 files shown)


matkoniecz:
CI logs:

```
 ❯ test/spec/util/keybinding.js (10 tests | 1 failed) 16ms
       ✓ returns self 2ms
       ✓ adds a binding for the specified bare key 2ms
       ✓ adds a binding for the specified key combination 1ms
       × matches the binding even when shift is present 8ms
```

either it is bug in the main code that needs to be fixed or it is bug in tests

matkoniecz:
@ahir-u Do you plan to finish it?

ahir-u:
Yes, didn't have time for it yet. I am a little confused, it should be this way:
  Shift + I = I
  Shift + Ctrl + I ≠ Ctrl + I 
correct?

tyrasd:
See https://github.com/openstreetmap/iD/issues/10817#issuecomment-4378305598

PR #12138, opened 2026-03-29
Avoid warning for websites having no as value

### LINKED ISSUE #12137: limit URL validation to only specific tags containing `website` in the key
Sometimes one might want to indicate that a certain feature does not have a website, where using `no` as a value could make sense. iD should not warn in such cases that the protocol is missing.

Similarly, maybe there also exist some tags with `website` in the key that are not expected to be an actual website? I found [`description:website`](https://taginfo.openstreetmap.org/keys/description%3Awebsite#values), but maybe that's a tagging mistake.

bhavyaKhatri2703:
fixes #12137

### DIFF
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -123,6 +123,7 @@ export function validationFormatting() {
             if (!tag) return null;
             const value = tag.trim();
             if (!value) return null;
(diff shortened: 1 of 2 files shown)


tyrasd:
#12137 has not been sufficiently discussed. Closing for now

Issue #12109, opened 2026-03-26
Make circles symetrical


Mahabarata:
### Description

- create an almost squared surface with 4 nodes
- select que circle function
- look at the number of nodes = 19
- create another surface, bigger or smaller, change it to a circle: the circle has still 19 nodes

problems:
- 19 nodes means that the circle is not symetrical
- a circle is normally defined with a center and a radius. With an asymetrical circle, it is **impossible** to know the position of the center and the length of the radius (create a line from a node of the circle to another one: you will never have a diameter of the circle).

To have a good 3D rendering of some buildings, it is important to place elements at their good position. To know where is the center of a circle is very important when a building has a circular part.

Look at this building https://www.openstreetmap.org/edit?editor=id&relation=5441490#map=21/44.7723417/-0.6437028 and its F4 result in the screenshot section.
It is made with 3 concentrical circles. It is important to know the position of each center to be able to move the circles at their good location.

My issue = when creating a circle, could you create a circle with **20** nodes instead of 19 ?

With 20 nodes:
- create a line from a node to its opposite = **this line will be a diameter** = you can know the diameter length and then the radius of the circle
- create a second line from another node to its opposite: it is another diameter, which means that **the 2 lines intersect at the perfect center of the circle** : now you can move the circle and the 2 lines to position the circle at its perfect location (the center of a squared building for example = a mosque with its dome).

### Screenshots

![Image](https://github.com/user-attachments/assets/b5538fba-fa8c-4147-8db9-f9a298bfc18d)

DujaOSM:
Additionally, I often use "make circular" to create ovals (such as athletic tracks, hippodromes, or church buildings). Having an odd number of points in the starting circle is suboptimal, since I can't make the oval fully symmetrical. 

Also, it would be nice to always have a "clockish" number of points: 12, 18, 24 etc. so that the circle can be easily stretched or divided in all directions.

tyrasd:
fixed in #12139: the number of vertices is now always an even number.

PR #12139, opened 2026-03-30
dynamic number of nodes for circularized way, depending on size of circularized feature

### LINKED ISSUE #8389: Polygon rounding creates too many nodes
When rounding closed ways on iD with the 'O' shortcut, iD creates a standard circular polygon with number of nodes that seems vary between 19 - 36. This has been causing problems in mapping, especially when newbies add many circular huts: adding 19 nodes for an hut which is 1-2 meters long adds unwanted complexity to the data causing problems with download of OSM data, especially when they are many.

On JOSM, the 'O' shortcut makes polygons circular keeping constant the number of nodes of the polygons. Shouldn't be this method be preferred instead of dropping 19 nodes each time? It is possible to validate this on JOSM, but wouldn't be better to create rounded polygons with less nodes, or keep the number of nodes constant?

Screenshot from 2021-03-10 09-37-37

tyrasd:
calculate number of points from a target segment length, such that
* small circles are generated using fewer nodes (minimum: 12)
* large circles are generated using more nodes (maximum: 36)
* intermediate size circles end up with a similar number of nodes then previously (~20 nodes for a single lane roundabout)

addresses #12109 and fixes #8389

before / after:

 

 

 

 

* [x] adjust/add tests
* [x] add to changelog

### DIFF
--- a/modules/actions/circularize.js
+++ b/modules/actions/circularize.js
@@ -1,5 +1,5 @@
 import { median as d3_median } from 'd3-array';
-
+import { geoLength } from 'd3-geo';
 import {
     polygonArea as d3_polygonArea,
     polygonHull as d3_polygonHull,
@@ -9,12 +9,11 @@ import {
 import { geoVecInterp, geoVecLength } from '../geo';
 import { osmNode } from '../osm/node';
 import { utilArrayUniq } from '../util';
-import { geoVecLengthSquare } from '../geo/vector';
+import { radiansToMeters } from '../ui/panels';
 
 
-export function actionCircularize(wayId, projection, maxAngle) {
-    maxAngle = (maxAngle || 20) * Math.PI / 180;
-
+export function actionCircularize(wayId, projection) {
+    const _maxSegmentLength = 4;
 
     var action = function(graph, t) {
         if (t === null || !isFinite(t)) t = 1;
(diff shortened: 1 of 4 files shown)


tordans:
This looks a lot better. 

I would argue we can reduce the number further by two nodes each (or one).
Overly precise geometries cause more work than imprecise once. So the default should IMO be closer to the later.

modules/actions/circularize.js
```diff
@@ -244,22 +251,31 @@ export function actionCircularize(wayId, projection, maxAngle) {
...
+            var actualDist = geoVecLength(actualPoint, centroid);
             var diff = Math.abs(actualDist - radius);
             //compare distances with epsilon-error (5%)
-            if (diff > 0.05*radius) {
+            if (diff > 0.05 * radius) {
                 return false;
             }
         }
 
+        const radiusM = radiansToMeters(geoLength({ type: 'LineString', coordinates: [
```
tordans: There is probably a reason but just from reading the diff it felt weird to have this logic duplicated.

tyrasd: right, this should probably be consolidated into a helper function. :+1:

tyrasd:
> I would argue we can reduce the number further by two nodes each (or one).

In my testing, 8 is definitely too few nodes for the result to be considered a circle, and 10 nodes also looked "not circular enough" (but it could be that I'm used too much to the previous number of 19 nodes). I wanted to make the number even for symmetry reasons (and #12109), so 9 or 11 are probably also not ideal.

I'd propose to use the current settings and tweak them (if needed) after having a bit of experience with them?

tordans:
Agreed. This is a good improvement to before so +1 to using this.

PR #12141, opened 2026-03-31
treat landuse=basin as water areas on map

matkoniecz:
followup to https://github.com/openstreetmap/id-tagging-schema/pull/1955

now that these areas are not deprecated, it is useful to indicate that these are not actually landuse areas (used on entire basin infrastructure area) but water areas (used only where basin holds water)


matkoniecz:
worked

screen-2026-03-31-10-38-27

PR #12151, opened 2026-03-31
use the `new` operator to instantiate coreGraph & osmEntity classes

k-yle:
~~blocked by #12150. after that PR is merged, we need to rebase this one before merging.~~

as discussed in https://github.com/openstreetmap/iD/pull/12093#issue-4127245650 and https://github.com/openstreetmap/iD/pull/12093#issuecomment-4117706960, we need to a mass update which adds the `new` operator to everywhere that instatiates the 5 main classes. This will allow us to migrate these files to ES classes in the future.

The PR updates `osmNode`, `osmWay`, `osmRelation`, `osmChangeset`, `coreGraph`. All other classes are used in <5 files, so they can be done later.

copying the original message here: 

> It [#12093] raises a tricky question: 
>
> [ES classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) must be instantiated with the [`new`  operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new) (`new coreGraph()` instead of `coreGraph()`). That gives us two options:
> 1. do a mass update to replace `coreGraph(` with `new coreGraph(` , but I think we [try to avoid](https://github.com/openstreetmap/iD/pull/10823#issuecomment-2694433828) mass updates?
> 2. [wrap the class defintion](https://github.com/openstreetmap/iD/blob/7ed11ae842d2273ac0886cf4166c2189cf006c7e/modules/core/graph.ts#L323) in a helper function which allows us to use it without `new`. Then we don't have to update any other files. 
>
> This PR includes option 2 as seperate commit. It's a bit odd, but i can't think of any other solution. 
> 
> RapiD [chose option 1](https://github.com/bhousel/Rapid/commit/8bbbb441). Thanks to [git-blame-ignore-revs](https://docs.gitlab.com/user/project/repository/files/git_blame/#ignore-specific-revisions), it's possible to hide the mass-update commit from the git history (locally and on GitHub). 
> 
> Not sure which approach is better. But it's a problem that we'll need to tackle somehow


k-yle:
@copilot please carefully check every file in this PR, to confirm that this PR only adds the `new ` operator, and does nothing else.

k-yle:
huh, that was actually useful. thanks @copilot

tyrasd:
#12150 is merged now

k-yle:
thanks, rebased now

tyrasd (approved):

Issue #10238, opened 2024-05-10
Show last update date of ELI below Layer list


tordans:
ATM, it is way to hard to understand how the pieces of ELI + iD play together.

One way to fix this are more regular updates to ELI like https://github.com/openstreetmap/iD/issues/9730 and https://github.com/openstreetmap/iD/issues/8466 suggest.

But maybe we can start with just showing the age of the data in the sidebar? So users understand why they don't see updates?

Something like
> Editor Layer Index last updated 2024-01-01

<img width="451" alt="Bildschirmfoto 2024-05-10 um 10 10 02" src="https://github.com/openstreetmap/iD/assets/111561/5b600622-0803-4086-aa42-6403ca736540">

andrewharvey:
I think that's only really useful for someone digging into ELI and iD, for a regular user it doesn't really help to tell them the index was last updated. It would be better to just move ahead with #9730 / #8466

tyrasd:
I would agree with @andrewharvey that it's probably most useful to always fetch a fresh version of ELI.

PR #12152, opened 2026-03-31
fetch imagery directly from ELI for instant updates

### LINKED ISSUE #8466: Regular releases with updated ELI
Currently, the imagery index is only updated with each release of iD.  This is ok when there are regular releases of iD., but the last version of iD was released on March 3th.  The latest commit to the develop branch was a month ago on March 18th, thus from the outside, it seems unlikely that there will be a new release soon. 

The last week without a commit to ELI was in September of last year. Personally, I perceive a duration of more than a month as too long for a change in ELI to propagate to the OSM homepage. 

I would suggest considering a weekly or bi-weekly patch release of iD with an updated version of ELI.

See also the discussion on openstreetmap-website:  https://github.com/openstreetmap/openstreetmap-website/issues/3179

### LINKED ISSUE #9730: Fetch imagery weekly
### Description

Currently, iD only fetches imagery from [editor-layer-index](https://https://github.com/osmlab/editor-layer-index) before each minor release. Since the last release that updated imagery was [over four months ago](https://github.com/openstreetmap/iD/releases/tag/v2.25.0), many sources that have been added/updated since then are not available for those that use iD. It would be nice to have the imagery list updated independent of iD releases similar to [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema), [osm-community-index](https://github.com/osmlab/editor-layer-index), and [name-suggestion-index](https://github.com/osmlab/name-suggestion-index).

Since it looks like a prior attempts to version the index itself weren't successful, I propose creating a new repository that fetches imagery from ELI weekly. This repository would function in a similar way to how [name-suggestion-index](https://github.com/osmlab/name-suggestion-index) uses the [nsi-collector](https://github.com/ideditor/nsi-collector) repo to periodically collect common brands/operators from the OSM planet file.

[Here's a proof of concept I made](https://github.com/arch0345/eli-collector)

See also:
* #4994
* #7425
* #8466

Edit: Started writing this yesterday and didn't notice #9729 which seems to touch on the same issue

k-yle:
blocked by either https://github.com/osmlab/editor-layer-index/pull/2927 or https://github.com/osmlab/editor-layer-index/pull/2530, depending on what option the ELI maintainers prefer.

---

This PR makes iD use the latest version of editor-layer-index, no need to wait for the next release anymore. 

- Closes #9730
- Closes #8466
- makes https://github.com/osmlab/editor-layer-index/issues/2625 much easier to implement
- fixe&shy;s https://github.com/osmlab/editor-layer-index/issues/2904 from iD's side (since we don't commit `imagery.json` into our git repo anymore)
- removes the need for https://github.com/openstreetmap/iD/issues/10238


config/id.js
```diff
@@ -9,6 +9,8 @@ const wmfSitematrixCdnUrl = ENV__ID_WMF_SITEMATRIX_CDN_URL
   || 'https://cdn.jsdelivr.net/npm/wmf-sitematrix@{version}/';
 const nsiCdnUrl = ENV__ID_NSI_CDN_URL
   || 'https://cdn.jsdelivr.net/npm/name-suggestion-index@{version}/';
+const eliCdnUrl = ENV__ID_ELI_CDN_URL
+  || 'https://cdn.jsdelivr.net/gh/osmlab/editor-layer-index@gh-pages/';
```
k-yle, line 13: if https://github.com/osmlab/editor-layer-index/pull/2927 is accepted, then this line needs to change to:

```suggestion
  || 'https://cdn.jsdelivr.net/npm/@openstreetmap/editor-layer-index@{version}/';
```

or alternatively, if https://github.com/osmlab/editor-layer-index/pull/2530 is accepted, then this needs to change to: 

```suggestion
  || 'https://cdn.jsdelivr.net/gh/osmlab/editor-layer-index@v1/';
```


currently it's using `gh-pages` just to make the PR demo work, it can't be merged as-is.

modules/renderer/background.js
```diff
@@ -30,17 +31,40 @@ export function rendererBackground(context) {
...
         if (_imageryIndex) return _imageryIndex;
 
+        // if the same ID exists in manual_imagery, prefer that
+        // over the original from ELI.
+        const overridenIds = new Set(manualSources.map(s => s.id));
+
+        /** @type {import("geojson").Feature[]} */
+        const merged = [
+          ...sources.features.filter(s => !overridenIds.has(s.properties.id)),
+          ...manualSources.map(source => ({
```
hlfan: What do you think of being a bit more verbose with the source origin?

```suggestion
        const sources = [
          ...eliSources.features.filter(s => !overridenIds.has(s.properties.id)),
          ...manualSources.map(source => ({
```

k-yle: thanks, good idea. done now

config/id.js
hlfan: > it can't be merged as-is.

If the eli maintainers [don't believe in software releases](https://github.com/osmlab/editor-layer-index/pull/2927#issuecomment-4640798658), which has been stated [twice](https://github.com/osmlab/editor-layer-index/pull/2530#issuecomment-2993631598), maybe it can be merged as-is?

Then also https://github.com/openstreetmap/iD/labels/waitfor-upstream would be unnecessary.

tordans: > > it can't be merged as-is.
> 
> If the eli maintainers [don't believe in software releases](https://github.com/osmlab/editor-layer-index/pull/2927#issuecomment-4640798658) which has been stated [twice](https://github.com/osmlab/editor-layer-index/pull/2530#issuecomment-2993631598)

It's only one person, though. [Recently active maintainer](https://github.com/osmlab/editor-layer-index/graphs/contributors?from=25.4.2026) where more open to it. But the governance around ELI is frustrating.

Which is part of the reason why I just released https://npmx.dev/package/@osm-editor-kit/maplibre-editor-layer-index as a take to work around this. The idea is, to use that to do automated, checked, weekly releases base on ELI and to make it easy to use that in other projects. I plan to create a DRAFT to use this in iD at some point. I consider this a workaround until ELI has something of it's own (the Maplibre/ReactMapGL-Part of the package would stay, though).

PR #12153, opened 2026-03-31
Fix geoLineIntersection when intersection occurs at segment start (u = 0)

nisshchayarathi:
## Problem

`geoLineIntersection` incorrectly returns `null` when the intersection lies at the start of the second segment (`uNumerator === 0`).

This happens because the function uses a truthy check:

```js
if (uNumerator && denominator)
```

Since `uNumerator` can legitimately be `0` for valid intersections, this condition incorrectly filters out those cases.

---

## Fix

Update the condition to check only `denominator`, which correctly handles parallel/colinear cases while allowing `u = 0` intersections:

```js
if (denominator)
```

---

## Impact

* Correctly detects intersections occurring at segment endpoints
* Fixes false negatives where valid intersections were previously missed
* Improves robustness of geometry utilities used across the editor

---

## Regression Test

Added a test case in `test/spec/geo/geom.js`:

* `a = [[0,0],[10,0]]`
* `b = [[5,0],[5,10]]`

Expected intersection: `[5,0]`

* Before fix: returned `null`
* After fix: returns `[5,0]`

---

## Verification

```bash
npm run -s test:once -- test/spec/geo/geom.js
npx -s eslint modules/geo/geom.ts test/spec/geo/geom.js
npm run -s test:typecheck
```

All checks pass.

### DIFF
--- a/modules/geo/geom.ts
+++ b/modules/geo/geom.ts
@@ -189,7 +189,7 @@ export function geoLineIntersection(a: Vec2[], b: Vec2[]) {
     var uNumerator = geoVecCross(geoVecSubtract(q, p), r);
     var denominator = geoVecCross(r, s);
 
-    if (uNumerator && denominator) {
+    if (denominator) {
         var u = uNumerator / denominator;
         var t = geoVecCross(geoVecSubtract(q, p), s) / denominator;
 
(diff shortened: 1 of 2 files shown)


tyrasd:
as long as you can't show how this would cause unwanted effects in practice, I assume this is by design and changing it would result in unexpected side effects.

PR #12159, opened 2026-04-01
preset search - match when preset text is in term

### LINKED ISSUE #9105: "private swimming pool" is not finding anything, even "swimming pool" preset

matkoniecz:
currently only when search terms are within preset text or text is close then matches are found

as result for example "private swimming pool" was not matching "swimming pool preset"

fixes #9105

Though it looks like sorting may be needed?

screen-2026-04-01-21-13-58

### DIFF
--- a/modules/presets/collection.js
+++ b/modules/presets/collection.js
@@ -65,6 +65,14 @@ export function presetCollection(collection) {
       return index === 0;
     }
 
+    function presetInSearchTerm(presetText) {
+      if (presetText.length === 0) {
+        return false;
+      }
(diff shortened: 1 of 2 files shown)


matkoniecz:
note: it may end adding some extra irrelevant or bizarre matches, but putting them at the end of list seems better than failing to match anything

matkoniecz:
I also though about splitting search term and doing word-by-word search, like

```
    if (value.includes(' ')) {
      const words = value.split(' ').filter(Boolean);
      const perWordResults = words.map(word => _this.search(word, null, loc).collection);
    }
```

tyrasd (approved):

PR #12163, opened 2026-04-02
Do not render 'none' after selecting none layer

### LINKED ISSUE #12154: Copyright strings not ready for Dark Mode
The copyright string for background "None" should just be "", not "none".

Image

By the way, these copyright strings are shown in white letters,
but the maps they are shown upon are usually white also,
that's why I never noticed them until today when I selected "None",
at least when using OSM website's Dark Mode.

aditya1010-max:
Fixes #12154

This happens because the fallback logic uses the source id ("none") when no attribution is available. The change skips rendering attribution for the "none" imagery source, preventing placeholder text from appearing in the UI. This aligns with expected behavior, where no imagery results in no attribution being shown.

After (Light Mode):
image

After (Dark Mode):
image

### DIFF
--- a/modules/ui/attribution.js
+++ b/modules/ui/attribution.js
@@ -16,9 +16,9 @@ export function uiAttribution(context) {
       .attr('class', klass)
       .merge(div);
 
-
+    const NO_IMAGERY = 'none';
     let attributions = div.selectAll('.attribution')
-      .data(data.filter(d => d.id !== 'none'), d => d.id);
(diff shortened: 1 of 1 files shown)


tordans:
For my taste this fix is too hidden and indirekt. A way to tie the two places in the code together could be to put the sting in a CONST and use this in both places.

tyrasd:
implemented without hardcoded hacks in cc94974f5

PR #12167, opened 2026-04-06
change how the flip operation detects non-numeric `direction` fields

k-yle:
see https://github.com/openstreetmap/id-tagging-schema/pull/2109#pullrequestreview-4049853428

now we hardcode `number` instead of `combo`/`radio`, so make it easier to switch between those two.


matkoniecz:
note that one direction PR was already merged, see https://github.com/openstreetmap/id-tagging-schema/pull/2081

k-yle:
yeah, so therefore we can test this by creating a vertex with `traffic_sign=yes` and pressing <kbd>v</kbd> which should automatically add `direction`

tyrasd (approved):

PR #12176, opened 2026-04-07
fix crash when restoring changes

### LINKED ISSUE #12087: The editor can't restore my changes
### URL

https://www.openstreetmap.org/edit

I don't know, ~~but I *suspect* that it happened when I reloaded the browser window and managed to open some menus before the "You have unsaved changes" prompt appeared.~~

After reloading my browser window, the iD-editor can't restore my changes. After pressing "Restore my changes", it just keeps showing the spinning loading symbol.

I've tried logging out and back in, to no avail.
I only had about 80 changes, so it shouldn't be a matter of the amount of data.

I was making little changes in Chrome on Android - which isn't really supported, but it has never had this bug.
I don't think that I can access the cookies on this phone. Should I try resetting them to just before the bug happened?

2.39.5

Other (please specify)

k-yle:
Closes #12087

caused by a typo in #11347

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -74,6 +74,7 @@ export default tseslint.config(
       'no-promise-executor-return': 'error',
       'no-proto': 'error',
       'no-prototype-builtins': 'off',
+      'no-restricted-globals': ['error', 'message'],
       'no-restricted-properties': 'error',
       'no-return-assign': 'off',
       'no-return-await': 'error',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :white_check_mark: Validation
 * Treat most aerialways as part of the routable network when checking other ways' connectivity ([#9406])
 #### :bug: Bugfixes
+* Fix crash when restoring changes ([#12176], thanks [@k-yle])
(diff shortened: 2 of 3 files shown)


eslint.config.js
```diff
@@ -74,6 +74,7 @@ export default tseslint.config(
       'no-promise-executor-return': 'error',
       'no-proto': 'error',
       'no-prototype-builtins': 'off',
+      'no-restricted-globals': ['error', 'message'],
```
k-yle, line 77: i added this so the CI will catch such issues in the future. not sure if we want to add just `message`, or all of the confusing ones, or none. 

i personally use [confusing-browser-globals](https://npm.im/confusing-browser-globals) to [ban all the global variables with generic names](https://github.com/k-yle/eslint-config-kyle/blob/61a0d87e9bcff8e15453276481fb362fcc8d4fdb/rules/commonRules.js#L100-L107).

k-yle:
@tyrasd i'm not sure if we should publish a patch release for this? if yes, it would be good to include the bugfix for #12030 too

tyrasd (approved):
Thanks! 🙇

eslint.config.js
tyrasd: good idea to extend this to all potentially mistakable globals. can you open a separate pr for that, please?

tyrasd:
yeah, a patch release for this probably makes sense. I'll take a look at #12030 

//edit: released as v2.39.6

Issue #9065, opened 2022-04-05
Place boundaries cancel out administrative boundaries in iD


ahallification:
### URL

https://www.openstreetmap.org/edit#map=16/44.2616/-73.3120

### How to reproduce the issue?

With editing open, view the border between NY and VT.

- Under Map Features, uncheck 'Boundaries'. The town boundaries (Essex<>Westport & Charlotte <> Ferrisburgh) disappear. The border between NY <> VT does not disappear.
- Under Map Features, check 'Boundaries', but uncheck 'Other Features'. The town boundaries cited above stay, but the NY <> VT border disappears. 

I believe this is a new issue, as the mapping community is moving towards not including the `boundary=administrative` tag on ways, but instead depending on tagging relations with `boundary=administrative`. The manifestation of the error appears to be rare, as this particular border has a 'place' relationship. The New England relationship appears to negate any of the other relationships that include this border. This is a personal theory and someone with a more technical understanding may find that there is some other cause. Keeping the `boundary=administrative` tagging on the ways prevents this error from manifesting. 

### Which iD Editor versions do you see the issue on?

Released version at openstreetmap.org/edit, RapiD version at mapwith.ai/rapid

### Which browsers are you seeing this problem on?

Chrome

DujaOSM:
As far as I reverse-engineered it, "Boundary" check box only checks if `boundary` tag is present on the way and displays it accordingly. It does not check if the way participates in boundary relations, as it should.

A similar issue has been relatively recently solved with landuse multipolygons. Filtering can get tricky for "multi-purpose" ways (ab)used both as administrative boundaries and for other uses, but it would be beneficial to have at least the basic use case right.

tyrasd:
another related case is when a landuse polygon has tagged outer ways, e.g. `barrier=fence`. The fence is then also hidden when only the landuse category is disabled.

//edit: the commonality seems to be that both place and barrier features are in the "other" category in iD.

PR #12177, opened 2026-04-07
Fix/simplify how features are sorted into "other" category

tyrasd:
addresses #9065:
* _boundaries_ also includes all ways that are a member of a boundary relation
* ways that don't match any non-"other" category inherit the categories of all their parent relations
* _other_ includes only the remaining tagged ways that don't match any other category

todos:
* [ ] handle case where feature is member of one relation (e.g. type=route) which does not match any category by itself
* [ ] test extensively
* [ ] add unit test for boundary ways
* [ ] add unit test for other category

### DIFF
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -2,7 +2,7 @@ import { dispatch as d3_dispatch } from 'd3-dispatch';
 
 import { prefs } from '../core/preferences';
 import { osmEntity } from '../osm';
-import { osmLanduseTags, osmLifecyclePrefixes } from '../osm/tags.js';
+import { osmIsInterestingTag, osmLanduseTags, osmLifecyclePrefixes } from '../osm/tags.js';
 import { utilRebind } from '../util/rebind';
 import { utilArrayGroupBy, utilArrayUnion, utilQsString, utilStringQs } from '../util';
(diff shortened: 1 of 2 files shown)


matkoniecz:
> boundaries also includes all ways that are a member of a boundary relation

that is unchanged from the current state, right?

https://www.openstreetmap.org/edit?editor=id&note=5520889#map=19/50.100233/20.380926

screen-2026-09-19-10-35-53

(I have first enabled boundaries only, then selected it)

It is also not in other...

screen-2026-09-19-10-36-38

matkoniecz:
> ways that don't match any non-"other" category inherit the categories of all their parent relations
> other includes only the remaining tagged ways that don't match any other category

I also tried to find behavior changes and failed.

Do you have example location where this PR has effects? Has it been since redone in other PR?

matkoniecz:
Or is it expected that situation remains the same but code is simpler? With list listing where relevant code is applicable?

matkoniecz:
From linked issue: go to https://www.openstreetmap.org/edit#map=16/44.26160/-73.31200 and deselect other features

Compare https://pr-12177--ideditor.netlify.app/#map=16.52/44.26248/-73.31230&background=Bing&id=w991395273

PR #12178, opened 2026-04-07
restrict url/website validation to specific tags

### LINKED ISSUE #12137: limit URL validation to only specific tags containing `website` in the key

bhavyaKhatri2703:
fixes #12137 

This pr restricts the url validator to specific tags like:
` 'website',
            'url',
            'contact:website',
            'contact:url',
            'source:website',
            'source:url',
            'image',
            'wikimedia_commons',
            'brand:website',
            'operator:website',
            'network:website',
            'website:en',
            'website:fr',
            'website:menu',
            'post_office:website',`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 * Treat most aerialways as part of the routable network when checking other ways' connectivity ([#9406])
+* Only consider well known tags containing URLs during website validation ([#12178], thanks [@bhavyaKhatri2703])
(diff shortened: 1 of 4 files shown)


tyrasd (approved):
`wikimedia_commons` did not belong to this list. Also, I moved the variable to the `osm/tags` module, so that it's easier to reuse/find in the future.

PR #12182, opened 2026-04-08
refine validating of wikimedia_commons links

### LINKED ISSUE #12366: URL validator unhelpful when `image=File:` matches `wikimedia_commons`
### Description

I was editing a map and I noticed that many [stolpersteine](https://wiki.openstreetmap.org/wiki/Tag:memorial%3Dstolperstein) in my area were reporting invalid URLs. The error message simply says:

> **XYZ has an invalid URL**
> URLs must start with "http://" or "https://".

But when I look at `website=*`, it's always a valid URL with `https://` protocol. I was always brushing these issues off as invalid.

Today, I realized that it's not the `website=*` that's invalid, it's the `image=File:` (linking to a Wikimedia Commons file by name)! It would be nice if the error message reflected that.

### Possible solution

We could have an error message specify tag name, e.g.

   > **XYZ has an invalid `image` URL**
   > URLs must start with "http://" or "https://".

<details>
<summary>Discarded solution</summary>

We could have a special case for this type of error

> **XYZ has an invalid `image` URL**
>
> ```diff
> -image=File:XYZ
> +wikimedia_commons=File:XYZ
> ```

</details>

### Screenshots

Image

tyrasd:
follow-up to #11499 / #12036:

* show preview of tag upgrades of the validation fix (also for normal website updates)
  
* suggest simplifying full URL to compact wikimedia_commons tag value
  
* improved reference text for `image` -> `wikimedia_commons` warning and change to _suggestion_ validation type
  
* fix validation of a few edge cases (e.g. image tag with semicolon separated values, and only a part matches the wikimedia commons syntax; or when wikimedia_commons file name contains `%` character)
* //edit: also: do not suggest to fix URLs with a domain name without a TLD (see below unexpected fix on the left, and on the right the new behaviour)
   ->


data/core.yaml
```diff
@@ -1636,6 +1636,11 @@ en:
         message: '{feature} has an invalid URL'
         message_multi: '{feature} has multiple invalid URLs'
         reference: 'URLs must start with "http://" or "https://".'
+      wikimedia_commons:
+        message: '{feature} has an invalid Wikimedia Commons link'
```
1ec5: Most people associate links with URLs. How about we call it a page title? That would be the technical term for it, whether it’s a file description page, gallery page, or category page.

tyrasd: Good point. Or maybe `… has an invalid Wikimedia Commons tag` for the osm-specific technical term? What do you think would be best?

1ec5: Sure, that works. I think we already use the term tag in other validator warning titles anyways.

hlfan:
Tried it a bit - https://www.openstreetmap.org/node/5318105690 causes this to concat multiple messages.

tyrasd:
> Tried it a bit - https://www.openstreetmap.org/node/5318105690 causes this to concat multiple messages.

Not sure I can replicate this issue. Yes, there are two warnings, but the first one is not about the URL… Do you mean something else?

hlfan:
Aah so the error is in the german message translation, not with this code. Sorry for the false alert.

tyrasd:
PS: @hlfan do you remember the reasoning behind the [_All split parts valid, but whole value still invalid_](https://github.com/openstreetmap/iD/commit/5ac1ae3fd582ae16d4c909fcdb2967904a0543a2#diff-b5e9cb58af5a1fca50cf418558961efe59fdf302b358dbb4b50192576182abd7R136-R143) branch added with #11499? At least now, after tweaking to the `isValidURL` check, it is not covered by any of our test cases (anymore?) and I cannot think of any way to trigger it. //edit: I removed the branch now in f44ea35, but maybe there is a reason to still have it?

hlfan:
I think that originated from https://github.com/openstreetmap/iD/pull/11438#discussion_r2373393551. Perhaps @1ec5 can comment on that.

hlfan:
Maybe there could be a separate message if both the `image` and the `wikimedia_commons` keys have valid Wikimedia Commons references because some mappers can't decide what to pick.

tyrasd:
> I think that originated from [#11438 (comment)](https://github.com/openstreetmap/iD/pull/11438#discussion_r2373393551). Perhaps @1ec5 can comment on that.

I see, however then this must have been buggy from the start, as `!invalidParts.length` would not catch the case for an URL like `https://example.com/foo;bar` (where there _is_ one invalid part).

I guess we _could_ think about how to actually properly handle those cases:
* `https://example.com/foo;bar` is not currently not auto-fixable as the second part does not look like a valid URL (missing a TLD)
* but `https://example.com/foo;bar.png` would be interpreted like a two-URL case where the fix would look like `https://example.com/foo;https://bar.png`. I hope the newly added tag-diff preview will make is obvious that the proposed fix is not to be applied in such cases.
* technically, we could add an extra fix for such cases, e.g. `https://example.com/foo;bar.png` -> `https://example.com/foo%3Bbar.png`
* but then it would also show up for `https://example.com;example.net` as `https://example.com%3Bexample.net`, which might be confusing if not worded well

tyrasd:
> Maybe there could be a separate message if both the `image` and the `wikimedia_commons` keys have valid Wikimedia Commons references because some mappers can't decide what to pick.

Do you mean like in the following example?
```
wikimedia_commons=File:foo.png
image=File:bar.png
```
to suggest
```
wikimedia_commons=File:foo.png;File:bar.png
```

but as far as I understood the semicolon-is-ambiguous issue is even more pronounced for the `wikimedia_commons` tag, and should therefore be avoided altogether, isn't it?

hlfan:
That, or preferably drop one of the links in that case.

matkoniecz:
wait, is the consensus to fix `image=File:bar.png` ones?

last time I looked into it (wanted to make a bot edit) there was, sadly, no clear consensus

matkoniecz:
(and if there is consensus it would make far more sense to run a bot edit rather than asking people to do bot job manually)

tyrasd:
> wait, is the consensus to fix `image=File:bar.png` ones?
> 
> last time I looked into it (wanted to make a bot edit) there was, sadly, no clear consensus

Ooh, I did not know that the `image` tag also [allows](https://wiki.openstreetmap.org/w/index.php?title=Key:image&oldid=2958203#Usage) wikimedia `File:*` values:

> There are at least three common styles of formatting the link:
> *  a Wikimedia Commons filename (formatted as `File:image.jpg`) - although this is commonly tagged as `wikimedia_commons=*`
> * […]

Looks like we should remove that validation until there is consensus about this in the community. Or at least make tweak the wording of the message and use the less aggressive "blue" validation type.

//edit: PS:

> (and if there is consensus it would make far more sense to run a bot edit rather than asking people to do bot job manually)

IMHO there is still merit, as this validation would let mappers discover that the tagging should use one tag over the other. In a "magic" bot edit this is not really the case.

matkoniecz:
> Ooh, I did not know that the image tag also [allows](https://wiki.openstreetmap.org/w/index.php?title=Key:image&oldid=2958203#Usage) wikimedia File:* values:

well, there is definitely no consensus for treating it as a valid value

but last time I checked there was no consensus to fix it either

¯\\_ (ツ)_/¯

tyrasd:
I changed the type of the `image` to `wikimedia_commons` validation to a _suggestion_ and tweaked the wording slightly such that keeping it remains acceptable, see screenshot in original post.

test/spec/validations/invalid_format.js
```diff
@@ -129,21 +129,92 @@ describe('iD.validations.invalid_format', function () {
...
+            var issues = validate(entity);
+            expect(issues).to.have.lengthOf(1);
+            expect(issues[0].subtype).to.eql('website'); // still an invalid URL in the tag
+            const fixes = issues[0].dynamicFixes(context);
+            expect(fixes).to.have.lengthOf(0);
+        });
+
+        it('should propose to remove URL from Wikimedia Commons tag', function() {
+            var entity = createPointWithTags({
+                'wikimedia_commons': 'https://commons.wikimedia.org/wiki/File:OpenStreetMap-Editor_iD_Logo.svg#mw-jump-to-license'
```
hlfan, line 206: A loop with more URLs would be nice to check for percent decoding capability:

https://commons.wikimedia.org/wiki/File:Ajout_noeud_sur_extrimité_ligne.jpg
https://commons.wikimedia.org/wiki/File:Editeur_iD_pour_OSM,_interface_en_français,_juillet_2021.png
https://commons.wikimedia.org/wiki/File:OSM_검토_요청_복사.png

maro-21:
> Ooh, I did not know that the `image` tag also [allows](https://wiki.openstreetmap.org/w/index.php?title=Key:image&oldid=2958203#Usage) wikimedia `File:*` values:

The Wiki neither allows nor disallows it. The Wiki simply describes the methods used.

PR #12186, opened 2026-04-09
split `osmEntity.id` into a new class `osmIdManager`

### LINKED ISSUE #10909: Refactor osmEntity

k-yle:
Another prerequisite to #10909

Currently `osmEntity.id` is a static property with various odd methods. We can decouple this by moving those methods into a seperate class. 

Simplifying `osmEntity` will make it much easier to refactor this weird class in the future 

Closes #10909

### DIFF
--- a/modules/core/history.js
+++ b/modules/core/history.js
@@ -12,6 +12,7 @@ import {
     utilArrayDifference, utilArrayGroupBy, utilArrayUnion,
     utilObjectOmit, utilRebind, utilSessionMutex
 } from '../util';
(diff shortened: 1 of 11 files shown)


tyrasd (commented):
👍 

Would you mind renaming the new class/singleton to `osmIdManager`? The consecutive capital letter are irritating me a little bit. 😅 Or was that name like that on purpose?

k-yle:
haha both looked weird to me in a codebase full of the word _iD_ 

but yeah I agree, changed now

PR #12188, opened 2026-04-09
Refactor hash methods to accept callbacks

hlfan:
`window.history.replaceState` is called in many places with an identical need for object <-> querystring conversion.

A new exposed `update` method can remove quite a bit of duplicate code.

Collecting these actions also increases modularity for things like dispatching events (https://github.com/openstreetmap/openstreetmap-website/pull/6811).

### DIFF
--- a/modules/behavior/hash.js
+++ b/modules/behavior/hash.js
@@ -5,31 +5,71 @@ import { select as d3_select } from 'd3-selection';
 import { geoSphericalDistance } from '../geo';
 import { modeBrowse } from '../modes/browse';
 import { modeSelect, modeSelectNote } from '../modes';
-import { utilObjectOmit, utilQsString, utilStringQs } from '../util';
+import { utilQsString, utilStringQs } from '../util';
 import { utilArrayIdentical } from '../util/array';
 import { utilDisplayLabel } from '../util/utilDisplayLabel';
 import { localizer, t } from '../core/localizer';
 import { prefs } from '../core/preferences';
 
+function getNewHash(arg) {
+    const original = utilStringQs(window.location.hash);
+    const update = typeof arg === 'function' ? arg(original) : arg;
+    if (!update || typeof update !== 'object') return;
+
+    const updated = { ...original, ...update };
+    Object.keys(update)
+        .filter(key => update[key] === null || update[key] === undefined)
+        .forEach(key => delete updated[key]);
+
+    return '#' + utilQsString(updated, true);
+}
+
+/**
+ * Updates the URL hash by applying a partial patch.
+ *
+ * Keys with nullish values will be removed from the hash.
+ *
+ * @param {(?Object<string, any>|function (Object<string, any>): Object<string, any>)} updater Either
+ * - a plain object of key/value pairs to merge into the hash, or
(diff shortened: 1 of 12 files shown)


modules/behavior/hash.js
```diff
@@ -100,29 +96,34 @@ export function behaviorHash(context) {
...
+     */
+    function updateHashIfNeeded(func) {
+        if (!func || typeof func !== 'function' || context.inIntro()) return;
 
-        var latestHash = computedHash();
-        if (_cachedHash !== latestHash) {
-            _cachedHash = latestHash;
+        const updated = func(utilStringQs(_cachedHash ?? window.location.hash));
+        Object.keys(updated).forEach(key => {
+            if (!updated[key]) delete updated[key];
```
tyrasd: > falsey values will be removed from the hash.

I could see the case where we might want to explicitly set a hash parameter to an empty string (e.g. `…&disable_features=&…`) to convey that this particular property is explicitly an empty "list", contrary to the absence of the parameter which would mean to use the default setting for this property.

I do assume this was changed from the previous explicit `delete` in order to make using the method more convenient. What about we only remove parameters when the property is explicitly set to `null` or `undefined`?

That way one can still have the convenience to assign it to variables that have either a value or are undefined/null, but might contain an empty strings. E.g. as in:

```js
    updateHash(hash => ({
      ...hash,
      background: id,
      overlays: o || null,
      offset: notableOffset ? `${x},${y}` : null
    }));
```

hlfan: I've thought about that. First I switched from nullish to falsely to make joining empty arrays less awkward, but the argument for explicity is very valid.
tbf, eslint discouraged me with the `eqeqeq` rule complicating the nullish value check.

modules/behavior/hash.js
```diff
@@ -147,11 +148,12 @@ export function behaviorHash(context) {
...
         if (mapArgs.length < 3 || mapArgs.some(isNaN)) {
             // replace bogus hash
-            updateHashIfNeeded();
+            updateHashIfNeeded(computeHashUpdate);
 
         } else {
+            const newHash = '#' + utilQsString(computeHashUpdate(utilStringQs(window.location.hash)), true);
             // don't update if the new hash already reflects the state of iD
-            if (_cachedHash === computedHash()) return;
+            if (_cachedHash === newHash) return;
```
hlfan: With the multiple instanciations `_cachedHash` is no longer behaving Singleton-like.
Hence the hash caching issues.

Could this be stored in a new context property?

hlfan:
I've moved `updateTitle` out of the exposed function to be able to call `behaviorHash.patch({ photo: null });` without requiring the `context`.

tyrasd:
@hlfan it seems like you're continuing to find to improvements for this functionality. :+1: let me know once you're happy for this to be reviewed/merged.

hlfan:
I'm happy about a review anytime @tyrasd!

I think it is now where I wanted to get it to in terms of modularity after noticing the singleton behaviour bug.
Just needed to work around my apparent chore brain quota. 😉

tyrasd (approved):

Issue #12142, opened 2026-03-31
Support playground= features in map rendering


matkoniecz:
this ideally would be done with some general though to make playground features distinguishable from general one (playground tunnel should not look like road tunnel)

maybe even playground= features can be made to be distinguishable from each other and in some way related to full-sized things (maybe tunnel can sort-of look like full tunnel? zip-wire share some styling with aerialway but be readily distinguishable?)

admittedly it is not a high-priority task

but should be nicely self-contained and likely requires no coding, just configuring some css
see https://github.com/openstreetmap/iD/pull/12141/changes for a much simpler PR doing some map rendering styling

### Description

Currently `playground=tunnel_tube` `playground=zipwire` lines, `playground=structure` areas and so on get no special rendering

in fact such objects (unless mapped as nodes) are impossible to be distinguished from untagged objects or ones tagged by gibberish like `hdjhdhdhjd=dsfhsfh`

see say https://www.openstreetmap.org/way/908479256/history#map=19/52.474675/13.443175

### Screenshots

<img width="390" height="406" alt="Image" src="https://github.com/user-attachments/assets/8f480ac4-574e-4632-ac7d-b0ee34750d4f" />

which of lines is play tunnel, which is zipwire and which is untagged?

<img width="765" height="243" alt="Image" src="https://github.com/user-attachments/assets/00745203-19b1-417c-9dae-32b0a123371b" />

matkoniecz:
for icons see https://github.com/openstreetmap/id-tagging-schema/pull/2120#issuecomment-4160781542 and related tasks

icons come from iD tagging schema

Issue #11189, opened 2025-07-05
Different map styles ("lenses")


tordans:
This issues is part of https://github.com/openstreetmap/iD/issues/11168. 

This issue will be updated with ideas and solutions on how to introduce the advanced user features in `kaligrafy/iD` to the broader user group of iD users. 

Right now **this issue is a stub** with meeting minutes from a discussion with @kaligrafy, @tyrasd and @tordans who meet online to talk about #11168.  There are many issues and some solutions for this topic already. I will try to collect some to allow us to get a broader pictures and discuss next steps.

---

**Goals**
* Introduce map styles that allow to **efficiently map one theme**
   For example side walk mapping, street side parking.
* Use map styles to get live **quality assurance** while mapping
   Eg. Highlight when one or a set of required tags are missing.

Adding map style capabilities will allow for accessibility features as well es features like dark mode along the road.

---

**Context**
The fork at `kaligrafy/iD` has custom map styles that are used by the trained mappers that use this fork. It allows them to extract a lot of information just from the style.
The style highlights
- data on maxspeed
- bike infrastructure
- buildings types
- highlight objects that do not follow a embedded minimum field requirements

The goal of this ticket is to explore how those advanced user features can be added to iD without mudding the experience for new and unexperienced users.

**Different kind of map styles**
- Styles that replace the current style completely
   (This is what `kaligrafy/iD` does, for example)
- Styles that add a set of layers below (or above) the current map style
   (This is what the "Highlight unsaved features" in iD does, for example)
- Styles could disable editing of features as part of the style
   (Eg. disable building and borders when editing side walks)
- Styles could hide or "gray out" features that as part of the style
   (Eg. Gray out buildings, POI when editing maxspeed)

---

### Research: What do we already have…

(This list does not have to be complete. Its more about getting a bigger pictures to better scope the feature.)

**iD Issue this Topic**

https://github.com/openstreetmap/iD/issues/6616

- A unique idea for a name for this feature in iD: "Lenses"
   - JOSM calls it "Map Paint Styles" [🔗](https://github.com/openstreetmap/iD/issues/6616#issuecomment-508982604) 
- A list of ideas what to highlight for iD [🔗](https://github.com/openstreetmap/iD/issues/6616#issuecomment-511659058) 
- An idea on where to expose this in the UI, once it is ready for every day use [🔗](https://github.com/openstreetmap/iD/issues/6616#issuecomment-508254031) 

**Highlight unsaved edits**
[...]

PR #12196, opened 2026-04-11
Add styling for playground features

aditya1010-max:
Addresses #12142

Adds distinct styling for `playground=*` features to make them visible and distinguishable from generic geometries.

Includes styling for:
- `playground=zipwire`
- `playground=tunnel_tube`
- `playground=structure`

Colors were chosen to be visually distinct yet consistent with related features. Happy to adjust them based on feedback.

---

### After

Map rendering

---

### Preset panel

Zipwire preset

Tunnel tube preset

Structure preset

Structure area example


tyrasd:
My first thought: We should only use one single color for all playground features. And as blue and green are already used for water/natural features a lot, I think it's best to use another color. The proposed shade of pink should work fine.

If needed, for some special features like the zipline and tunnel features mentioned above, we could distinguish them by altering the stroke and/or dashing (e.g. in order to create analogies to road tunnels / bridges).

matkoniecz:
> We should only use one single color for all playground features

What about allowing at least different shades of pink?

aditya1010-max:
> My first thought: We should only use one single color for all playground features. And as blue and green are already used for water/natural features a lot, I think it's best to use another color. The proposed shade of pink should work fine.
> 
> If needed, for some special features like the zipline and tunnel features mentioned above, we could distinguish them by altering the stroke and/or dashing (e.g. in order to create analogies to road tunnels / bridges).


Yeah, colors wouldn’t work well here, since they’re already used for things like fountains, trees, and shrubs, which are commonly present in playgrounds and wouldn’t blend well with playground features.
For tunnel we can use dashed lines, for zipline i think we can keep it as plain lines.

tordans:
I suggest we split up the effort. The initial issue and IMO most important one is to be able to see that lines / areas with `playground=*` are something else than any other unstyled feature.

Implementation detail: Is there a way to add this without adding all tags to the class "registry" but just the key? I assume adding all those tags will add to the performance cost that our rendering has.

The second phase could be to see how to better style specific `playground=*` features.
Personally, I don't see the need for this at this point; I see it more as something for #11189

aditya1010-max:
> Implementation detail: Is there a way to add this without adding all tags to the class "registry" but just the key? I assume adding all those tags will add to the performance cost that our rendering has.
> 


I think we can try implementing this, as it would be beneficial, especially from a performance perspective. 
We can treat playground=* as key-based (adding only tag-playground and skipping value-based classes), which should avoid adding multiple entries to the class registry. Value-specific styling could then be handled separately (e.g., via attributes). 
This idea could potentially be extended to other tags where value-based classes add limited benefit.

matkoniecz:
Do we want say `playground=ajanahdhdbdu` line (or with other gibberish) also become styled?

aditya1010-max:
Yes , the base .tag-playground style would apply to any playground=* value, which helps distinguish these features from unstyled ones. More specific styling would only apply to known values, example-

.tag-playground { ... }
.tag-playground[data-playground="zipwire"] {
  /* zipwire-specific styling */
}

Unknown values would simply fall back to the base style.

aditya1010-max:
This plan follows the approach suggested by @tordans above, with additional details to clarify the steps. It is divided into two logic steps and two styling steps for easy implementation

#### a. Update classification logic

Modify `tag_classes.js` to treat `playground=*` as a key-based classification:

* Add only `tag-playground` and skip generating `tag-playground-*` value-based classes

---

#### b. Apply common playground styling

Define a single, consistent style to make playground features visually distinct from unstyled features:

Example:
```css
.tag-playground { ... }
```
---

#### c. Expose `playground` values as attributes (Phase 2)

Add logic in the rendering layer to expose `playground` values as attributes

(e.g., `data-playground="zipwire"`).

---

#### d. Enhance styling in a separate layer

Introduce more specific styling based on `playground` values, using attribute-based selectors instead of introducing additional classes.

Example:

```css
.tag-playground[data-playground="zipwire"] {
  /* zipwire-specific styling */
}
```
---

aditya1010-max:
For now, this adds special styling only for playground=structure (area only).
Other playground=* features continue to use the base tag-playground styling, which already makes them visually distinct from unstyled features.

After for playground=* (except special CSS presets):
Screenshot 2026-04-17 202114

After for playground=structure (area only)
Screenshot 2026-04-17 211511
 
How it blends:
image


Before:
Screenshot 2026-04-17 202229

After:
Screenshot 2026-04-17 202242

aditya1010-max:
| Area | Line |
|------|------|
|  |  |

aditya1010-max:
> Do we want say `playground=ajanahdhdbdu` line (or with other gibberish) also become styled?

I get the concern about unintentionally styling arbitrary values. One option would be to treat the base class as the default style, and then apply additional styling only for specific presets. That would give us flexibility to fine-tune individual presets without over-styling everything by default. However, that approach also limits the benefits of this architecture. 
As for gibberish values, those are ultimately a data-quality issue rather than a styling one. Personally, i think if someone intends to add invalid data, they can do so regardless of how styling is implemented, Styling should remain robust and predictable, and not be responsible for determining data validity

aditya1010-max:
I just noticed that the current styling for playground=area seems to clash with landuse=industrial, so I’ll revert it to the base class styling. I think it’s better to keep this PR focused on introducing the architecture and a base style.
 
For the styles used in this PR, since these styles will apply across multiple tags, feedback and opinions are appreciated.

PR #12210, opened 2026-04-16
add confusing-browser-globals to lint config

k-yle:
follow up to https://github.com/openstreetmap/iD/pull/12176#discussion_r3046112076

this is one of two possible ways to prevent bugs like this. 

the other solution is to continue with the TS migration. happy to close this PR and focus on that instead


tyrasd (approved):
> the other solution is to continue with the TS migration

let's do both 😉

PR #12212, opened 2026-04-16
rename `osmEntity.key` to `osmIdManager.key`

k-yle:
part 7 of #10909.

this is the last PR for the preëmptive work ! after that is should be easy to convert `osmEntity` it to an ES6 Class. i might try this tonight.

---

in this PR: we replace a static function `osmEntity.key(...)` with a class method, so that:
- it matches every other class method
- we don't need to import `osmEntity` is as many files
- it removes another random thing which was attached to this abstract class, to make it easier to refactor this file in the next PR

### DIFF
--- a/modules/svg/points.js
+++ b/modules/svg/points.js
@@ -3,7 +3,7 @@ import { clamp } from 'es-toolkit/compat';
 import { select as d3_select } from 'd3';
 
 import { geoScaleToZoom } from '../geo';
-import { osmEntity } from '../osm';
+import { osmIdManager } from '../osm';
 import { svgPointTransform } from './helpers';
 import { svgTagClasses } from './tag_classes';
 import { presetManager } from '../presets';
@@ -40,7 +40,7 @@ export function svgPoints(projection, context) {
     function fastEntityKey(d) {
         const mode = context.mode();
         const isMoving = mode && /^(add|draw|drag|move|rotate)/.test(mode.id);
-        return isMoving ? d.id : osmEntity.key(d);
+        return isMoving ? d.id : osmIdManager.key(d);
     }
 
 
--- a/modules/svg/points.js
+++ b/modules/svg/points.js
@@ -3,7 +3,6 @@ import { clamp } from 'es-toolkit/compat';
 import { select as d3_select } from 'd3';
 
 import { geoScaleToZoom } from '../geo';
-import { osmEntity } from '../osm';
 import { svgPointTransform } from './helpers';
 import { svgTagClasses } from './tag_classes';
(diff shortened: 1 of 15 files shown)


modules/svg/points.js
```diff
@@ -38,9 +37,10 @@ export function svgPoints(projection, context) {
     // Avoid exit/enter if we're just moving stuff around.
     // The node will get a new version but we only need to run the update selection.
     function fastEntityKey(d) {
+        if (typeof d === 'number') return d;
```
k-yle, line 40: i had to add this line, because there was an minor "bug": this function was being passed a number sometimes, which caused `osmEntity.key()` to return  `"undefinedv0"`.

same reason for adding the `instanceof` guard added in the file above.

tyrasd: > this function was being passed a number sometimes

I tried to find out how/when this would be the case, but could not find any reason why it would ever be anything else than an osm entity. Any hints about how to trigger this "bug"?

tyrasd (commented):
> it matches every other class method

I believe that the method was implemented as a static method on purpose to make using it with d3's `.data` easier: `.data(…, osmEntity.key)`. I'm actually not sure if it is better suited as a instance method or for as a static function somewhere. Speaking of it: I see that this method is currently also present in the new [`osmIdManager`](https://github.com/openstreetmap/iD/blob/e8c4642/modules/osm/id_manager.ts#L34-L36), but never used.  Maybe it can also live there (either by renaming the manager to `osmEntityManager` to reflect the more generic usage, or factoring it out to a simple exported function `getOsmEntityD3Key` that only does that)? Anyway, one of the two methods should be removed.


Another thing: The naming of this method as `key` was IMHO quite unfortunate, as the association with the OSM term "tag key" is potentially confusing. Here it is not as quite as bad, but there are some other modules (e.g. `ui/fields/combo.js`) where it is super perplexing that the `key` property is assigned to a tag _value_: when in half of that line, `key` is meant to be an index for d3, and in the other half on is talking about key-values pairs of a tag. 🤷 

As we're touching all of the occurrences of this method here anyway: what do you think about renaming this to something a little less confusing. Maybe `index` could work? Or `d3_key`??

tyrasd:
small ping @k-yle :arrow_up: :blush:

k-yle:
sorry, i missed this notification amongst the dependabot spam 😆

I've simplified this to use the identical method from `osmIdManager`. then we can deal with the `undefinedv0` issue in a seperate PR

k-yle:
conflicts resolved

tyrasd (approved):

PR #12213, opened 2026-04-16
icons: use target symbol for geolocate

### LINKED ISSUE #11602: Use target symbol for Show My Location icon
> Target Symbol: This icon, common on Android and in web applications like Google Maps, often appears as a circle with a dot in the center ~~and four lines extending outwards~~, resembling a target or crosshair. Tapping it typically re-centers the map view to your current location.

This would be much more familiar than the current arrow pointing northeast.

Same goes for OSM.ORG website, which should be updated too.

Adi-gitX:
## Summary
- replace the dedicated geolocate sprite with a target-style symbol
- keep the change scoped to the existing  asset used by the Show My Location button and flash message

## Testing
- not run (SVG asset change only)

Fixes #11602.


tyrasd:
as mentioned on https://github.com/openstreetmap/iD/issues/11602#issuecomment-3572412906: if anything, this should display a different icon depending on the platform (iOS: keep current icon, other OS: maybe crosshair is better).

But I would also like to keep this in sync with the icon on the osm website (https://github.com/openstreetmap/openstreetmap-website/issues/1494) as the menu is designed to look similar to the main website on purpose. As long as it is not changed there as well, I think keeping the current icon is better.

PR #12218, opened 2026-04-18
include all tags from special fields when removing tags during preset change

### LINKED ISSUE #12075: Handle special fields like `localized` names in tag removal logic

tyrasd:
fixes #12075, replaces and closes #12107, see also #11696

includes special handling for the following fields:
* `localized`
* `multiCombo`
* `directionalCombo` fields

also refactors two similar function `allTags` function that do the same thing essentially.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Standardize tooltips in combo boxes: always show (full) raw tag, don't duplicate the already rendered titles, and show descriptions from wiki/taginfo where available ([#12010], thanks [@tordans])
 * Use full width for the dropdown box for the values of the `access` field ([#12065])
 * Preserve the order of options when a combo field accepts both static options as well as autosuggestions from taginfo and display additional tag values from taginfo as raw-options only ([#12117])
+* When removing tags while changing presets: include all tags associated with `localized`, `multiCombo` and `directionalCombo` fields ([#12075], [#11696])
 #### :hammer: Development
 * Replace `sinon` with `vitest`'s built in spy/mock library ([#12058])
 * Add type annotations to `context.js` module ([#11589], thanks [@k-yle])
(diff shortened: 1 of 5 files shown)


matkoniecz:
Retagging 

```
leisure=park
name=aaaa
name:af=afrikans
```

to bus stop keeps `name` and removes `name:af`

Is it intentional? I think that either both should be removed or both should be kept.

Both presets display `name:af` if tagged

matkoniecz:
Retagging recycling bin to recycling center will remove marked `recycling:` tags if any

I created https://github.com/openstreetmap/id-tagging-schema/issues/2214 as it would make sense to show these at least in moreFields anyway, at least I think so (and it would fix this unexpected and unwanted, I think, behavior)

matkoniecz:
A bit offtopic here: but maybe tests for retagging should also mention that discardable tags will be discarded?

I guess we do not need to test case where presets have discardable tags as it is not supposed to happen.

------------------

Also for https://github.com/openstreetmap/iD/blob/develop/test/spec/actions/change_preset.js and not sure where to report - but retagging `shop=butcher butcher=poultry` to other shop type correctly purges `butcher` tag. But it seems not covered by any test.

Though `amenity=bench bench=aaa` will preserve `bench` tag.

tyrasd:
> retagging `shop=butcher butcher=poultry` to other shop type correctly purges `butcher` tag. But it seems not covered by any test.
> Though `amenity=bench bench=aaa` will preserve bench tag.

The fact that `butcher` was removed in the first example was caused by the same bug as reported in #12253 (fixed with de28bd79). The bench example did not show this behaviour because the preset does not have a (e.g. `multicombo`) field which  triggered the bug.

matkoniecz:
https://github.com/openstreetmap/iD/pull/12218#issuecomment-4314204446 - is it expected to be fixed now?

Issue #12187, opened 2026-04-09
Highlight building + demolished:building as an error


boothym:
Based on this topic: https://community.openstreetmap.org/t/how-to-get-editors-to-notice-that-the-removal-of-a-feature-is-indicated-by-the-demolished-prefix/

A mapper wrongly added building=yes to an area already tagged as demolished:building=yes - but as the OP in that topic points out, iD (or JOSM) does not alert users to this error.

**Therefore, iD should highlight the building + demolished:building combination as an error** (and perhaps other lifecycle prefixes as well?).

tyrasd:
This does not only affect buildings, but in principle any feature with a lifecycle prefix tag.

But yes: This might be a contender for a validation message in iD. It could offer the user to choose which status is the correct one.

bhavyaKhatri2703:
i had some questions regarding implementing this validation -

1. What should be the severity for this validation , warning or error?
2. What additional lifecycle tags can be included in this list apart from:
`'demolished', 'removed', 'razed', 'disused', 'abandoned'`?

PR #12220, opened 2026-04-19
Add validation for lifecycle prefix tags conflict

bhavyaKhatri2703:
Addresses #12187.

This PR adds validation for lifecycle prefix tags and provides an option to remove one of the tags.
Example:
building:destroyed,  building

image


modules/osm/tags.js
```diff
@@ -433,3 +431,15 @@ export const osmUrlKeys = new Set([
     'website:menu',
     'post_office:website',
 ]);
+
+export const osmLifecycleConflictPrefixes = new Set([
```
RudyTheDev, line 435: What is this list based on? The prefixes documented on https://wiki.openstreetmap.org/wiki/Lifecycle_prefix are slightly different.

Why isn't this just using `osmLifecyclePrefixes`, why a separate list?

RudyTheDev:
What about something like `disused:leisure=stadium` + `leisure=park` - these are not mutually exclusive.

Or what about something like `disused:amenity=parking` + `disused:parking=surface` - this would retain `disused:parking` value.

This should really have unit tests for common scenarios/examples and edge cases.

tordans:
The next, planned step for lifecycle tags is https://github.com/openstreetmap/iD/pull/10844
IMO any improvements should be based on the progress over there as an improvement to what is already planned.

PR #12227, opened 2026-04-21
Allow more dependabot PRs at the same time

matkoniecz:
to prevent "no point in merging them, bot will open 4 new ones anyway" reaction - just let it open all

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,7 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 20
     cooldown:
       semver-minor-days: 7
       semver-patch-days: 7


Dimitar5555:
While we are at it, is it possible to force it to group multiple updates in a single PR? These notifications are quite annoying and there is no point for me to see them, so it would be nice if it was one every few days instead of (potentially) 20 every few days. Unfortunately, I can't even block dependabot, since it is a bot...

matkoniecz:
https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups-- seems like a tool for that but I have not looked into it (maybe creating issue for that would make sense? especially if you would be potentially interested in implementing a fix)

tordans:
According to https://docs.renovatebot.com/bot-comparison/ there are groups and also we could maybe have to run monthly or weekly only.

tyrasd:
Since a few weeks, we use groups to bundle updates logically (i.e. only one PR for all fontawesome updates whenever they release a new version instead of 4 independent PRs). Similarly, we have set the maximum update to weekly (for non-security related updates).

Yes, we could also bundle everything into a single PR per week, but that would make reviewing those much trickier.

tyrasd (approved):

PR #12236, opened 2026-04-21
fix dead links by replacing by IA links

matkoniecz:
replacement by live links from that site or its replacement would be nicer


matkoniecz:
I am 100% fine with this not being listed in changelog

tyrasd (approved):
thx!

PR #12243, opened 2026-04-22
Fix unable to delete way after removing it from multipolygon relation

### LINKED ISSUE #12232: cannot delete line after creating multipolygon and removing from relation
steps to reproduce:

* create an area
* select a node of the area, right click and choose the _split_ operation
* select one of the resulting two ways
* remove it from the multipolygon relation
* try to delete it

one gets the following error message which does not make sense:

Image

JustinYoo37:
Summary:
This PR fixes a bug where a line/way could not be deleted after creating a multipolygon and then removing that way from the relation.

Fixes #12232

Problem:
After creating a multipolygon, removing a member way from the relation should allow that way to be deleted normally. Instead, the way remained in a blocked state and could not be deleted.

Solution:
Updated the relation/member cleanup logic so that once a way is removed from the multipolygon relation, it is treated as an independent feature again and can be deleted normally.

### DIFF
--- a/modules/core/graph.ts
+++ b/modules/core/graph.ts
@@ -204,8 +204,6 @@ export class coreGraph {
         parentRels = parentRels || this._parentRels;
 
         var type = entity && entity.type || oldentity && oldentity.type;
-        let removed: string[] = [];
-        let added: string[] = [];
         let i: number;
 
         if (type === 'way') {   // Update parentWays
@@ -214,6 +212,8 @@ export class coreGraph {
             entity = <osmWay | undefined>entity;
             oldentity = <osmWay | undefined>oldentity;
 
+            let removed: string[] = [];
+            let added: string[] = [];
             if (oldentity && entity) {
                 removed = utilArrayDifference(oldentity.nodes, entity.nodes);
                 added = utilArrayDifference(entity.nodes, oldentity.nodes);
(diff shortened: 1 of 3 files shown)


matkoniecz:
CI:

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/spec/core/graph.js > iD.coreGraph > #parentRelations > updates after removing a way from a multipolygon (#split area → remove member)
TypeError: iD.actionDeleteMembers is not a function
 ❯ test/spec/core/graph.js:547:24
    545|             var idx = relation.members.findIndex(function(m) { return …

    546|
    547|             graph = iD.actionDeleteMembers(relation.id, [idx])(graph);
       |                        ^
    548|
    549|             expect(graph.parentRelations(graph.entity(wayRemoved))).to…

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

```

tyrasd:
wow. I've rarely seen a "fix" dodging the real cause of the reported bug as well as in this case. /s

see f8481d3ec67280d277cac8861e0e947a53a9405c for what actually had caused the regression

Issue #12239, opened 2026-04-21
outdated claim in ARCHITECTURE about d3 not supporting something


matkoniecz:
https://github.com/openstreetmap/iD/blob/develop/ARCHITECTURE.md

> (This is functionality [not yet provided](https://github.com/mbostock/d3/issues/894) by d3's own behaviors.)

links issue that is now closed as implemented in d3

matkoniecz:
maybe iD can now use that d3 thing?

tyrasd:
good catch! 🙇 

> maybe iD can now use that d3 thing?

yeah, but IMO keeping the explicit `.off` method is more readable. The implementation can continue to list the event handlers individually or use the shorthand `.on('.zoom', null)` if it should deem reasonable. No need to refactor the existing code.

PR #12244, opened 2026-04-22
Update outdated D3 behavior note in ARCHITECTURE.md

aditya1010-max:
Addresses #12239

Updates outdated D3 behavior note in ARCHITECTURE.md.

D3 now supports namespaced event listener removal (e.g. `selection.on(".zoom", null)`).


aditya1010-max:
I think this reads better now without referencing the previous limitation. Let me know if you'd prefer a different wording.

tyrasd:
replaced by ffbbd51… no need to get too technical here.

PR #10612, opened 2024-12-19
Improve Modal Layout: Group Shortcuts and Buttons with Responsive Scrolling. Solves #10591

Deeptanshu-sankhwar:
Closes #10591 

1. Changes in `uiConfirm`:
Added a `content-wrapper` div to group the `message-text` and `buttons` sections under a single parent. This improves layout structure and enables scoped scrolling behavior.

2. Changes in `uiShortcuts`:

Introduced a `content-wrapper` div to group the `shortcuts list` and `buttons` sections. This ensures consistent scrolling behavior and dynamic layout within the modal.

3. CSS Updates:

Applied scrolling behavior `(overflow-y: auto)` to the `content-wrapper` for both modals.
Added `max-height: calc(90vh - 100px)` to dynamically allocate space for the header and footer, ensuring responsiveness across various screen sizes.

4. Improved Structure and Responsiveness:

Scoped scrolling within the `content-wrapper` while keeping the header and footer static.
Ensured better organization and a clean layout to support future extensions.

PR #12254, opened 2026-04-24
Fix shortcuts modal scrolling: constrain scroll to content area and preserve header scroll behavior

### LINKED ISSUE #10591: Limit scrollbar height in overlays
### URL

_No response_

Limit the height of the scrollbar in the (i.e. Custom Layer and Keyboard Shortcuts) overlay to the bottom of the header, as it is for other menu tabs.

image

image

image

2.30.4, 2.31.0-dev, Rapid 2.4.3

Chrome

isabelladesi:
**Closes #10591**

This change fixes the scrollbar behavior in the Keyboard Shortcuts modal. Previously, the scrollbar extended into the header area because scrolling was applied to the entire modal content container. This caused the header to appear visually inconsistent and not remain fixed during scrolling.

**This fix:**
- Restricts scrolling to the content wrapper instead of the full modal container
- Keeps the header fixed at the top of the modal
- Prevents the scrollbar from overlapping the header

Additionally, this change preserves expected scrolling behavior by forwarding wheel events from the header to the scrollable content area. This ensures users can still scroll the modal even when hovering over the header.

**Testing**
- Resized the browser window to force overflow
- Verified scrollbar is limited to the content area
- Verified header remains fixed during scroll
- Verified scrolling works when hovering over both content and header

No regressions observed in the test suite (`npm test` passed).

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -5352,6 +5352,21 @@ img.tile-debug {
     max-width: 1050px;
 }
 
+.modal-shortcuts .content {
+   display: flex;
+   flex-direction: column;
+   max-height: 90vh;
+   overflow-y: hidden;
+}
+
+
+.modal-shortcuts .wrapper {
+   flex: 1 1 auto;
+   min-height: 0;
+   overflow-y: auto;
+}
+
+
 .modal-shortcuts .modal-section:last-child {
     padding: 10px 15px 20px 15px;
     min-height: 275px;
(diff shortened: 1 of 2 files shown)


mxdanger:
How would you say this differs from #10612 and what makes this approach better?

matkoniecz:
I propose to close it as abandoned

tyrasd:
> Restricts scrolling to the content wrapper instead of the full modal container

I'm not convinced this is the optimal solution here: when the screen estate is limited, we don't want to waste it by having the header of the modal always being displayed.

PR #12258, opened 2026-04-27
Fix tag search to also check addTags when searching by key=value

### LINKED ISSUE #12179: tag search fails to find amenity=school, education=school works (tag search should also check addTags)
### URL

_No response_

1. select point creation
1. click on map
1. go to feature search
1. search for `amenity=school`, `education=school` - only the second works

see https://github.com/openstreetmap/id-tagging-schema/blob/245e730ec86de017ebce8512e8bdc60d26a93fe2/data/presets/education/school.json#L43

will make PR after https://github.com/openstreetmap/iD/pull/12159 is processed

2.39.6

Firefox

yorty001:
Fixes #12179
Tag search was not checking addTags when matching key=value queries. This meant searching amenity=school would not find School Grounds, even though amenity=school is defined in its addTags.
Fixed by also checking addTags in the leadingTagKeyValues search logic in collection.js.

### DIFF
--- a/modules/presets/collection.js
+++ b/modules/presets/collection.js
@@ -170,11 +170,15 @@ export function presetCollection(collection) {
     // matches key=value to preset.tags
     let leadingTagKeyValues = [];
     if (value.includes('=')) {
-      leadingTagKeyValues = searchable.filter(a => a.tags &&
-          Object.keys(a.tags).some(key => key + '=' + a.tags[key] === value))
-        .concat(searchable.filter(a => a.tags &&
-          Object.keys(a.tags).some(key => leading(key + '=' + a.tags[key]))));
-    }
+    leadingTagKeyValues = searchable.filter(a => a.tags &&
+      Object.keys(a.tags).some(key => key + '=' + a.tags[key] === value))
+    .concat(searchable.filter(a => a.tags &&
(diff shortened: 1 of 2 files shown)


matkoniecz (commented):
adding also test to https://github.com/matkoniecz/iD/blob/develop/test/spec/presets/collection.js covering this change would be a good idea

yorty001:
Just checking whether there’s anything else needed for this PR.

modules/presets/collection.js
```diff
@@ -170,11 +170,15 @@ export function presetCollection(collection) {
...
-    }
+    leadingTagKeyValues = searchable.filter(a => a.tags &&
+      Object.keys(a.tags).some(key => key + '=' + a.tags[key] === value))
+    .concat(searchable.filter(a => a.tags &&
+      Object.keys(a.tags).some(key => leading(key + '=' + a.tags[key]))))
+    .concat(searchable.filter(a => a.addTags &&
+      Object.keys(a.addTags).some(key => key + '=' + a.addTags[key] === value)))
+    .concat(searchable.filter(a => a.addTags &&
+      Object.keys(a.addTags).some(key => leading(key + '=' + a.addTags[key]))));
+}
```
matkoniecz, line 181: is this intend change intentional?

PR #12276, opened 2026-04-30
Ensure only clickable qaItem icons show the pointer cursor (fixes #9073)

### LINKED ISSUE #9073: qa header icon has pointer cursor
### URL

_No response_

1. Enable a qa issues layer
2. Select a qa issue
3. Hover over the header icon (the icon to the left of the header) - the cursor is a pointer, indicating there may be a link.
4. Click on the [qa header icon](https://github.com/openstreetmap/iD/blob/ad67f81cfde96bfa17488ccb14b51a8655f81681/css/80_app.css#L2758). Nothing happens.
(icon is highlighted in blue in the screenshot for convenience)

image

_No response_

Chrome

Quantum-Cucumber:
Fixes #9073. Currently the qaItem icons in the issue view sidebar have `cursor: pointer` applied. 

This seems to be due to the css for qaItem icons on the map (which _are_ clickable). This change differentiates between the two icon uses based on the `.target`

---

I could not see anywhere else the qaItem icons are used so I believe this covers both use cases (sidebar & map).


tyrasd:
Thanks for investigating. I think you are indeed correct in that the cursor icon was only accidentally applied to the header section, instead of only for the map icon.

PR #12300, opened 2026-05-07
Install `@vitest/eslint-plugin`

FloEdelmann:
I don't know if this is desired; let me know if it isn't.

This PR installs [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `recommended` config and fixes all violations.

The biggest change is from Chai style to expect style. The main advantage is consistency: e.g. for testing that a value is not undefined, there were previously 4 different patterns in use (`.not.to.be.undefined`, `.to.be.not.undefined`, `.to.not.be.undefined` and `.not.to.eql(undefined)`), now there is only one (`.toBeDefined()`), enforced by the ESLint rules. I split up every single pattern update into a single commit, so it is as easy as possible to review.

I'm happy to split this up into smaller PRs if preferred.


test/spec/ui/sections/raw_tag_editor.js
```diff
@@ -28,8 +28,8 @@ describe('iD.uiSectionRawTagEditor', function() {
 
 
     it('creates input elements for each key-value pair', function () {
-        expect(element.selectAll('input[value=highway]')).not.to.be.empty;
-        expect(element.selectAll('input[value=residential]')).not.to.be.empty;
+        expect(element.selectAll('input[title=highway]').size()).toBeGreaterThan(0);
+        expect(element.selectAll('input[title=residential]').size()).toBeGreaterThan(0);
```
FloEdelmann, line 32: I think the fact that these assertions passed before the change was a coincidence; as `element.selectAll('input[value=XXX]')` resolved to empty selections, but changing the searched attribute to `title` then worked.

k-yle (commented):
if this idea is accepted, we should probably uninstall all chai related packages at the same time, right?

eslint.config.js
```diff
@@ -144,13 +149,21 @@ export default tseslint.config(
...
         'vi': 'readonly'
       }
     },
     rules: {
+      ...vitest.configs.recommended.rules,
+      'no-unused-expressions': 'off',
       'no-unused-vars': 'warn',
-      'no-unused-expressions': 'off'
+      'vitest/expect-expect': ['error', {
+        assertFunctionNames: ['expect', 'doMatch', 'dontMatch', 'verifySingleCrossingIssue']
```
tyrasd, line 155: I guess at a later point we probably should refactor those 3 custom assert functions with custom _extended expects_ (https://vitest.dev/api/expect.html#expect-extend).

tyrasd (commented):
I'm actually quite indifferent in this. Yes, vitest "recommends" Jest-style assertions, but as far as I can tell, it also officially supports chai style asserts. Both are fine in my opinion.

I guess, the `@vitest/eslint-plugin` does only support for Jest-style asserts, right? And there is no equivalent linter that accepts the chai style approach?

If so, adding the linter (which does have a benefit to the project) forces us to change over / do this refactoring once. I think that's acceptable and a net benefit for the project.

FloEdelmann:
> Yes, vitest "recommends" Jest-style assertions, but as far as I can tell, it also officially supports chai style asserts. Both are fine in my opinion.

:+1:
 
> I guess, the `@vitest/eslint-plugin` does only support for Jest-style asserts, right?

Exactly.

> And there is no equivalent linter that accepts the chai style approach?

Not that I'm aware of.

PR #12301, opened 2026-05-08
fix regression of piste:type (and building:part) rendering

### LINKED ISSUE #12297: piste:type=* ways lose ski-trail rendering in v2.40.0
### URL

https://www.openstreetmap.org/edit#map=18/70.634273/29.698802

1. Open iD editor using **v2.40.0.**
2. Navigate to an area containing ski-related ways tagged with `piste:type=*` (e.g. nordic, downhill, etc.).
3. Use aerial imagery with snow-covered terrain/background.
4. Observe the ski trail rendering on the map.
5. Compare the rendering to previous iD versions.

Result:
Ski/piste ways no longer use the previous distinct styling and instead appear as plain, low-contrast lines. On white or snow-covered aerial imagery, they become very difficult to see and can blend into the background, making ski infrastructure significantly less visually distinguishable than in earlier versions.

Image 

Image

2.40.0

Chrome

matkoniecz:
fixes #12297

followup to #10256

### DIFF
--- a/css/50_misc.css
+++ b/css/50_misc.css
@@ -8,7 +8,7 @@ path.line.shadow.tag-man_made-pipeline,
 path.line.shadow.tag-natural-tree_row,
 path.line.shadow.tag-roller_coaster-track,
 path.line.shadow.tag-roller_coaster-support,
-path.line.shadow.tag-piste {
+path.line.shadow.tag-piste_type {
     stroke-width: 16;
 }
 path.line.casing.tag-aerialway,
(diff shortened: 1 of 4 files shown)


css/50_misc.css
```diff
@@ -105,11 +105,11 @@ path.line.casing.tag-aerialway {
 }
 
 
-/* pistes */
-path.line.stroke.tag-piste {
+/* piste_types */
```
matkoniecz:
```suggestion
/* pistes */
```

matkoniecz:
note: https://pr-12301--ideditor.netlify.app/#disable_features=boundaries&map=18.70/70.63414/29.69900&background=Bing&id=w1507471180 is NOT fixed

no idea why

`piste_type=nordic` is in fact rendered as expected - and I expected `piste:type` to be transformed into this value

matkoniecz:
@k-yle do you have an idea what is going wrong here?

tyrasd:
d3e2c007a1b176e8d1865946eaeb88b340bedf58 was the culprit

tyrasd (approved):

PR #12305, opened 2026-05-08
simplify input field focussing logic

### LINKED ISSUE #12299: Clicking sources combo box after typing changeset comment does not immediatelly show the dropdown
As per title.

Repro: Make some edits. Click Save. Type something in Changeset Comment textbox. Without unfocusing, click on the source "Add..." combobox. Expect it to dropbox instantly to show common options like before. Instead, it gets focused, but does not drop down. Requires another click to drop down. There's also a short delay now, including any subsequent dropdowns on additional source combo boxes.

Did not observe this behaviour with any other similar combo boxes, for example restaurant name -> cuisines.

2.40.0

tyrasd:
Fixes #12299.

It would be important to test thoroughly if this introduces any new regressions regarding various combo fields and their interaction with the dropdowns, autosuggestions, other fields, etc.


modules/ui/fields/combo.js
```diff
@@ -749,7 +746,7 @@ export function uiFieldCombo(field, context) {
             _combobox
                 .on('accept', function() {
                     _input.node().blur();
-                    _input.node().focus();
```
tyrasd, line 752: this apparently didn't do anything previously 🤷

modules/ui/fields/combo.js
```diff
@@ -598,9 +598,6 @@ export function uiFieldCombo(field, context) {
                 }
                 t[field.key] = context.cleanTagValue(arr.filter(Boolean).join(';'));
             }
-
-            window.setTimeout(function() { _input.node().focus(); }, 10);
```
tyrasd, line 602: removed because the focus is already set by the other event handlers (e.g. `accept`), and this would sometimes wrongly focus the previously selected field's input element when clicking directly into another combo field's input box, causing strange flickering of the _focused_ highlight styling.

modules/ui/combobox.js
```diff
@@ -57,6 +57,7 @@ export function uiCombobox(context, klass) {
             .on('keyup.combo-input', keyup)
             .on('input.combo-input', change)
             .on('mousedown.combo-input', mousedown)
+            .on('mouseup.combo-input', mouseup)
```
tyrasd, line 60: this replaces the manual adding/removing of the event handler in the `mousedown` handler. I think we can trust browsers that the mouseup event will not be fired without a corresponding mousedown event. 🤷

tyrasd: also, this was the main fix for the regression reported in #12299: As `initCombo` will reset all event handlers when re-rendering the field (e.g. as a consequence of a `change` event following another field's input element's `blur` event), the `mouseup.combo-input` event handler could also be removed without it being immediately added again as expected. Doing this in the constructor works robustly in those cases as well.

modules/ui/combobox.js
```diff
@@ -446,7 +444,12 @@ export function uiCombobox(context, klass) {
             if (!d) {
                 d = _fetched[val];
             }
-            dispatch.call('accept', thiz, d, val);
+
+            if (val !== '') {
```
tyrasd, line 448: makes sure that the `accept` event is not triggered when a user tabs out of the input field without entering anything, in order to navigate to the next UI field.

PR #12306, opened 2026-05-08
make `contact:*` fields clickable if they contain usernames or full URLs

### LINKED ISSUE #11120: Handle url->id mapping, id->url click linking for key:contact=*

k-yle:
Closes https://github.com/openstreetmap/id-tagging-schema/pull/1019#issuecomment-4324879974
Closes https://github.com/openstreetmap/iD/issues/11120 via an alternative approach
Replaces https://github.com/openstreetmap/iD/pull/11124

Goal: to make both of these cases support clickable links (since both tagging styles are valid):
- `contact:instagram`=[`tinyssandwichbar`](https://instagr.am/tinyssandwichbar)
- `contact:instagram`=[`https://instagr.am/tinyssandwichbar`](https://instagr.am/tinyssandwichbar)

image

The solutions is to change social media fields like `contact:instagram` from `type=url` to `type=identifier`. 

Fields with `type=identifier` are already clickable if the `pattern` matches. But they are not clickable if the value is a full URL. This PR fixes that limitation.

The OSM-website has already implemented this approach, for every key supported by [tag2link](https://github.com/JOSM/tag2link)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased (2.41.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* Make tags like `contact:instagram` clickable if they contain a plain username, or a full URL ([#12306], thanks [@k-yle])
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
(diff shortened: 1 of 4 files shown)


k-yle:
the CI is failing on master, not related to this PR

tyrasd (approved):
👍

PR #12307, opened 2026-05-08
replace aes-js library with webcrypto

k-yle:
this was originally suggested [on the osm discord](https://discord.com/channels/413070382636072960/782592770849505281/1494310403159887973), but it turns out that person is working on RapiD, not iD. so this PR backports the idea.

---

like #11634, this reduces the bundle size by about 0.1MB. 

AES-CTR is natively supported by WebCrypto, and support by all browsers [since 2020](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/decrypt). 

So the question is: do we need to support 6-year-old browsers? 
- we already use modern browser features with [no error handling...](https://github.com/openstreetmap/iD/pull/11783#discussion_r2715935768), which means iD already crashes in 5-year-old browsers
- our [browserslist config](https://github.com/openstreetmap/iD/blob/223f0f2c46147f9fe63987b35ea1bd0a9c8667f4/package.json#L147-L149) means that we only support 2-year-old browsers, with a few expections.


bhousel:
Probably worth doing, but part of why I hesitated switching this out in Rapid was because it switches the decryption from sync to async.  It's still probably a good idea, but I didn't want to audit all the places that the code is used.

k-yle:
thanks bryan, yeah it was it bit cumbersome to make it fit into the synchronous code, but thankfully it's only used in those 2 places

tyrasd (approved):
yeah, that should be fine. btw: osm-auth also [relies](https://github.com/osmlab/osm-auth/blob/8135c73a44f689dc66ee610e4613d320fb2d8e54/src/osm-auth.mjs#L608) on webcrypto being available.

tyrasd:
@k-yle: would you like to add this to the changelog as well?

k-yle:
sure, done

PR #12308, opened 2026-05-08
migrate miscellaneous files to typescript

k-yle:
originally I was just going to edit `util/*` (since future PRs depend on these files), but I added some other files that are trivial to convert


modules/util/keybinding.ts
```diff
@@ -91,30 +106,30 @@ export function utilKeybinding(namespace) {
...
         selection.on('keydown.capture.' + namespace, capture, true);
         selection.on('keydown.bubble.' + namespace, bubble, false);
         return keybinding;
     }
 
     // was: keybinding.off()
-    keybinding.unbind = function(selection) {
-        _keybindings = [];
+    keybinding.unbind = function(selection?: d3.Selection) {
+        _keybindings = {};
```
k-yle, line 132: this was presumably a typo, that happens to still works

modules/util/keybinding.ts
```diff
@@ -141,16 +156,16 @@ export function utilKeybinding(namespace) {
 
 
     // Add one or more keycode bindings.
-    keybinding.on = function(codes, callback, capture) {
+    keybinding.on = function(codes: string[], callback: Binding['callback'], capture: boolean) {
         if (typeof callback !== 'function') {
             return keybinding.off(codes, capture);
         }
 
-        var arr = utilArrayUniq([].concat(codes));
```
k-yle, line 149: this shallow clone is redundant since `utilArrayUniq` does not mutate its input

modules/util/object.ts
```diff
@@ -1,26 +1,24 @@
-export function utilObjectOmit(obj, omitKeys) {
-    return Object.keys(obj).reduce(function(result, key) {
+export function utilObjectOmit<T extends object, K extends keyof T>(obj: T, omitKeys: K[]): Omit<T, K> {
```
tyrasd, line 1: this could be replaced by es-toolkit's [`omit`](https://es-toolkit.dev/reference/object/omit.html)?

//edit: this is probably also true for a lot of the methods in `array.ts` which also exist identically in es-toolkit. Perhaps it's better to migrate them to typescript as proposed, and replace them with es-toolkit's methods in a later step separately.

modules/util/object.ts
```diff
@@ -1,26 +1,24 @@
...
     }, {});
 }
 
-/**
- * @template T
- * @typedef {{ [key: string]: { [value: string]: T } }} TagDictionary<T>
- */
+export interface TagDictionary<T> {
+    [key: string]: {
+        [value: string]: T
```
tyrasd:
```suggestion
    [key: TagKey]: {
        [value: TagValue]: T
```

modules/util/dimensions.ts
```diff
@@ -1,11 +1,13 @@
...
+
+function refresh(selection: d3.Selection, node: HTMLElement) {
     var cr = node.getBoundingClientRect();
     var prop = [cr.width, cr.height];
     selection.property('__dimensions__', prop);
     return prop;
 }
 
-export function utilGetDimensions(selection, force) {
+export function utilGetDimensions(selection: d3.Selection, force?: boolean) {
```
tyrasd: shouldn't this return `Vec2` explicitly instead of `any`?

```suggestion
export function utilGetDimensions(selection: d3.Selection, force?: boolean): Vec2 {
```

and above:

```ts
function refresh(selection: d3.Selection, node: HTMLElement): Vec2 {
    var cr = node.getBoundingClientRect();
    var prop: Vec2 = [cr.width, cr.height];
```

tyrasd (commented):
Thanks! 🤗 

Only very partial review of the PR, but see a few details I noticed already below.

modules/util/dimensions.ts
k-yle: good point, i missed that. fixed now

modules/util/object.ts
k-yle: yeah i agree, shall we do this in a separate PR?

modules/osm/deprecated.ts
```diff
@@ -1,7 +1,7 @@
 export type DataDeprecated = { old: Tags; replace?: Tags }[];
 
 /** @param {Tags} tags @param {DataDeprecated} dataDeprecated */
-export function getDeprecatedTags(tags: Tags, dataDeprecated: DataDeprecated) {
+export function getDeprecatedTags(tags: Tags, dataDeprecated: DataDeprecated): DataDeprecated {
```
tyrasd, line 4: This might be a little over-pedantic, but I do prefer to be explicit with method return types whenever it make sense. This is mostly to avoid errors in the future when refactoring the respective function: When the return type is explicit, it will be immediately obvious when one accidentally extends the scope of the type (e.g. by writing an inadvertent `return false;` or just `return;` somewhere).

For the more complex (e.g. UI) modules with more elaborate return types I think we can omit it and rely on the automatic type inferring, as it would not add much clarity to those methods.

modules/ui/confirm.ts
```diff
@@ -3,37 +3,37 @@ import { uiModal } from './modal';
...
 
     section.append('div')
         .attr('class', 'modal-section message-text');
 
-    var buttons = section.append('div')
+    const buttons = section.append('div')
         .attr('class', 'modal-section buttons cf');
 
 
-    modalSelection.okButton = function() {
```
tyrasd, line 23: This only worked as long as `modalSelection` was `any` (see added return type annotation in `modal.js` below).

I guess the whole approach of just piggy-backing a custom method onto the selection object is a bit questionable in the first place, but let's think about that in the future, shall we?

modules/util/array.ts
```diff
@@ -105,15 +102,16 @@ export function utilArrayFlatten<T>(a: T[][]): T[] {
 //     4: [{type: 'Dog', name: 'Spot'}],
 //     5: [{type: 'Cat', name: 'Tiger'}, {type: 'Dog', name: 'Rover'}]
 //   }
-export function utilArrayGroupBy<T>(a: T[], key: keyof T | ((item: T) => string)) {
-    return a.reduce<Record<string, T[]>>(function(acc, item) {
-        var group = (typeof key === 'function') ? key(item) : <string>item[key];
+export function utilArrayGroupBy<T>(a: T[], key: keyof T): Record<string, T[]>;
+export function utilArrayGroupBy<T, K extends string | number | symbol>(a: T[], key: (item: T) => K): Record<K, T[]>;
+export function utilArrayGroupBy<T, K extends string | number | symbol>(a: T[], key: keyof T | ((item: T) => K)): Record<K, T[]> {
```
tyrasd, line 107: this is a bit tricky to correctly represent: when used with a callback method, it works as expected (the returned record has the key type of the function's return value). But if it's just a key of the supplied inputs, I did not find a way to be the right type.

Technically one could supply the methods with inputs like:

```ts
var pets = [
    { type: 1, name: 'Spot' },
    { type: 2, name: 'Tiger' },
    { type: 2, name: 'Rover' },
    { type: 3, name: 'Leo' }
];
```

Where `utilArrayGroupBy(pets, 'type')` will return a record with key of type `number` (compare `utilArrayGroupBy(pets, p => p.type)`).

I tried something like this:

```ts
export function utilArrayGroupBy<K extends string | number | symbol, T extends <any, K>>(a: T[], key: keyof T): Record<K, T[]>;
```

but that does not correctly narrow down the result's record's key's type, and would even prevent mixed-type inputs to be supplied (as the `pets` example above).

Sooo… I believe it should be fine to "bodge" this to `Record<string, T[]>` for the non-callback case (at runtime everything will be indexed by string anyway, so it doesn't matter too much). But maybe at some point we could drop that overload completely and refactor our code to only use the more typesafe callback variant?

modules/util/dimensions.ts
```diff
@@ -1,29 +1,29 @@
...
-    var cr = node.getBoundingClientRect();
+function refresh(selection: d3.Selection<HTMLElement>, node: HTMLElement): Vec2 {
+    const cr = node.getBoundingClientRect();
     const prop: Vec2 = [cr.width, cr.height];
     selection.property('__dimensions__', prop);
     return prop;
 }
 
-export function utilGetDimensions(selection: d3.Selection, force?: boolean): Vec2 {
+export function utilGetDimensions(selection: d3.Selection<HTMLElement>, force?: boolean): Vec2 {
```
tyrasd, line 10: is `selection` is just a dumb `d3.Selection`, later on `selection.node()` will be just `any`, which isn't really helpful

tyrasd (commented):
I finally got around to take a look at this. I tweaked it a bit: mostly being more explicit with function return type, converting some code to let/const/arrow functions and adding d3's selection's element base type where needed. Please see comments below pointing out the more relevant changes.

Let me know what you think. 😊

modules/ui/confirm.ts
k-yle: yeah, agreed. my current approach has been to minimise the git diff as much as possible.

so i haven't been changing cases like this, which require involve modifying heaps of code.

but maybe i shouldn't follow that approach too strictly...

k-yle:
thanks, those changes looks great !

Issue #9429, opened 2022-12-15
Note tool deselects itself when turning off data layer or when moving while the data layer is off


danieldegroot2:
### How to reproduce the issue?

1. select the Note tool
2. disable the OSM data layer
3. the Note tool deselects itself (the tool is no longer highlighted in blue and the cursor changes to a regular cursor), just like the node, way, area tools do when this is done (although these actually are disabled entirely).

1. enable the data layer
2. the note tool can be selected and when moving (dragging) around the map, the note tool stays selected.
3. disable the data layer again, the note tool will deselect itself as before
4. the note tool can be re-selected
5. now move the map
6. the note tool will deselect itself again

This makes it difficult to submit notes when the data layer is off and the notes tool remains available.
This may have been an oversight when adding the toggle.

Browser: Chrome, though outdated.

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.23.2

### Which browsers are you seeing this problem on?

Chrome

PR #12316, opened 2026-05-09
fix: don't deselect non-note objects when toggling notes layer

### LINKED ISSUE #12304: Object is deselected when disabling notes layer
### URL

openstreetmap.org/edit / Any

Object is deselected when disabling notes layer, without note selected.

1. Select an object, ie. a building.
2. Enable notes layer in data menu (line+area icon in right sidebar). Do not select a note.
3. Disable notes layer. The building is deselected.
This behavior is understandable only if a note is selected, if at all.

more or less latest version of below browsers on Windows 10.

N/A (result is identical to default view)

Released version at openstreetmap.org, development version at netlify, older versions at tasking managers.

2.40.0-live, 2.41.0-dev, earlier (at least 2.34)

Chrome, Firefox, Microsoft Edge

---

Related #9429

See also #10959.

deepak-s-git:
Fixes #12304

When disabling the notes layer...the editor was unconditionally entering 
browse mode....which deselected any currently selected object even if it 
wasn't a note...This fix only enters browse mode when disabling the notes 
layer if a note is actually selected..

### DIFF
--- a/modules/ui/sections/data_layers.js
+++ b/modules/ui/sections/data_layers.js
@@ -58,7 +58,14 @@ export function uiSectionDataLayers(context) {
             layer.enabled(enabled);
 
             if (!enabled && (which === 'osm' || which === 'notes')) {
-                context.enter(modeBrowse(context));
+                var modeId = mode && mode.id;
+                var isNoteMode = modeId === 'add-note' || modeId === 'select-note' || modeId === 'drag-note';
+
+                if (which === 'osm' && !isNoteMode) {
+                    context.enter(modeBrowse(context));
+                } else if (which === 'notes' && isNoteMode) {
+                    context.enter(modeBrowse(context));
+                }
             }
         }
     }


tyrasd (commented):
as we also want to fix the "mirrored" version of this bug #9429, I'm wondering if there's a more elegant way to approach this in `modeBrowse` directly?

deepak-s-git:
Updated the fix to also address #9429....now checks the current mode before entering browse mode
so note related modes are preserved when toggling the OSM data layer....and non-note selections are preserved when toggling the notes layer

tyrasd (dismissed):
what's up with all those unrelated edits like `function(x)` -> `function (x)`? 😒 Please clean up the PR to only include changes that are necessary for the fix.

deepak-s-git:
Sorry for the unrelated changes in the previous commit.....My editor's format on save was auto formatting the file...Didn't notice it last time 😁....

Have fixed it noww...

danieldegroot2:
Looks good to me.

Issue #12313, opened 2026-05-09
Document `maprules` param and make it work (again)


tordans:
Initial title, but [see update](https://github.com/openstreetmap/iD/issues/12313#issuecomment-4411702280)
> ~~Retire MapRules from the codebase?~~

https://github.com/openstreetmap/iD/pull/12312 made me have a look again if MapRules is around still.
Last time I checked years ago I was not able to get it working…

Nowadays I dont find a website about it at all.

@maxgrossman you worked on it last 7 years ago https://github.com/radiant-maxar/maprules/commits/develop/ - what is your take on this? Does it still work; or do you consider it worth keeping around and reviving?

I still wonder if parts of this feature can be used to allow custom presets - See item in https://github.com/openstreetmap/iD/issues/11168.

Links:
* https://wiki.openstreetmap.org/wiki/MapRules does not add much details
* https://github.com/openstreetmap/iD/pull/5617 is the PR that added the feature ~8 years ago

tordans:
FYI I am looking into this some ATM and my current take is: Lets document this feature propery without the external service. And fix the few issues that are currently in the codebase related to this like https://github.com/openstreetmap/iD/pull/12312

It looks like this feature might actually be useful when documented properly. The JSON can be stored in a public gist and used in iD.

Will update here once I tested it fully, likely suggest a PR with docs changes and small fixups.

danieldegroot2:
Was https://github.com/radiant-maxar/maprules/issues/72
See also https://github.com/hotosm/tasking-manager/issues/1592

tyrasd:
we decided on the June [community chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_10th_June_10:00_%E2%80%93_11:00_(UTC)) to remove the maprules functionality for now.

Issue #11168, opened 2025-06-29
📂  Backport advanced user features from `kaligrafy/iD`


tordans:
Pierre-Léo @kaligrafy runs a modified fork of the discontinued iD-V3 Prototype in Québec, Canada. He reached out to us to see which features can be backported to iD.

This ticket is a wrapper ticket to collect more detailed tickets that discuss certain features.

The fork can be tested at https://projets.chaire.transition.city/#background=GeodesieQuebec&disable_features=boundaries&map=17.58/45.40780/-73.94573 (but is meant for trained users for specific local projects, only).

In general, the features in kaligrafy's fork are advanced user features / power user features that aim to speed up the mapping process:

**Focus topics:**
- [ ] Draw and select presets faster. The fork uses the iD-V3 favorites and shortcuts for this.
   See https://github.com/openstreetmap/iD/issues/11170 for more
- [ ] Custom map styles for visual quality assurance (QA) and visual feedback. The fork has a custom map style that allows trained users to see data on maxspeed (data), minimum field requirements (QA/completeness), bike infrastructure (data/QA), buildings types (data/QA)
    See https://github.com/openstreetmap/iD/issues/11189 for more

**Additional topics:**
- [ ] Custom presets for special tag combinations
   See also…
   - https://github.com/openstreetmap/iD/issues/5049
   - https://github.com/openstreetmap/iD/issues/3470
   - Consider GoMap, EveryDoor, SC https://github.com/openstreetmap/iD/issues/3470#issuecomment-3017818694
   - See also https://github.com/tordans/editor-settings-schema
- [ ] Line tools to draw curved roads more easily
- [ ] Area tools to follow along a existing area to connect landuse areas easily
- [ ] Workarounds to handle imagery alignment issues

Pierre-Léo and his team plan to contribute PRs for those parts that we can find a good fit for in iD.

---

This topic was discussed outside this repo, eg. during the iD Community meetup https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2025#Community_Chat_2025-06-04_9:00_UTC

danieldegroot2:
> Custom presets for special tag combinations

See also https://github.com/openstreetmap/iD/issues/6170#issuecomment-538874723 regarding maprules

PR #5617, opened 2018-12-17
maprules service

maxgrossman:
integrates remote custom presets & validation rules generated by [maprules](https://github.com/radiant-maxar/maprules)

PR #12317, opened 2026-05-09
Explore and fix `maprules` URL param and explore adding `presets` URL param

### LINKED ISSUE #5049: Replace or merge iD data at runtime

tordans:
Year ago I looked at the options that https://github.com/openstreetmap/iD/pull/5617 promised but sadly never managed to get it working back then. (And apparently 4 years ago [I investigated it again](https://github.com/radiant-maxar/maprules/issues/72).)

Today inspired by https://github.com/openstreetmap/iD/issues/12313#issuecomment-4411702280 I took another look and I was surprised to get it to work quite easily. But also, that the `maprules` "API" does something different (today), than I expected. It is "just" about the rules/validations; it does not touch presets at all (anymore; it used to [according to the old PR](https://github.com/openstreetmap/iD/pull/5617))

---

This draft is a **pure exploration** of the feature. The code is – like the messy commit messages show – not reviewed yet and mostly LLM. My goal is to get feedback on the basic idea – what and how we could like to have in iD. And once that is clear, cleanup and review the code.

---

## Testing

https://pr-12317--ideditor.netlify.app/#presets=https://gist.githubusercontent.com/tordans/549a328ccff34963192899113c73d35a/raw/f575e37198d51350050f661d00e6e02f8461d1f3/bench-presets.json&maprules=https://gist.githubusercontent.com/tordans/549a328ccff34963192899113c73d35a/raw/f575e37198d51350050f661d00e6e02f8461d1f3/bench-rules.json&map=20.00/52.47414/13.44576&disable_features=boundaries&locale=en&background=Berlin-2024&id=n12518261746

(Note to self: Hosting on GitHub Gist means, the URL-Hashes need to be update on every change.)

## **Video:**

https://github.com/user-attachments/assets/e80a57ab-e092-454e-9596-f775a4f8caad

## **Notes on the exploration of `maprules`**

**What it does:**
- This add validation rules to the list of validations

**Status Quo:**
- Right now in iD, the feature is broken; but fixing the translation issue is quite easy

**My take:**
Personally I don't see that big of an impact for this but it is also the part that is already there. We should either clean it up (remove it) or fix it up (which is not too much work).

**Follow up:**
- ATM the rules are not represented in the Rules filter; which we might want to add later.

## **Notes on the exploration of `presets`**

I then looked into if we can follow the same idea and allow custom presets from external, CORS enabled URLs. And that turned out to be not too much code to add as well.

**What it does:**
- This allows to write preset that follow the tagging schema and merge them into the current presets
- The custom presets then behave like any other preset would

**My take:**
This could be a quick and low code change way to enable users to manage custom presets. We have this idea on the list at https://github.com/openstreetmap/iD/issues/11168 but did not discuss it in the way explored here. In the long run, I still think that https://github.com/tordans/editor-settings-schema/tree/main/presets-schema would be the better solution (storing custom presets in the user profile settings on osm.org so they can be shared between sessions and editors); but that is harder to do and this solution opens [the campaigning angle](https://github.com/openstreetmap/iD/issues/12313#issuecomment-4412385495) and is quicker to build.

**Follow up:**
- We might want to add a custom style to those (different color in list and map?)
- We might want to add a group to the preset selection overview list for the custom presets?

## **Hosting the json files**

I explored hosting the json files that iD reads as Github Gist https://gist.github.com/tordans/549a328ccff34963192899113c73d35a
That works nicely when used with the raw file option. It is tedious though, because updates require to change the URL.
Campaigns will likely want to host their files elsewhere, but Netlify and such make this kind of thing super easy nowadays. To be clear, I don't think it needs a service like https://github.com/radiant-maxar/maprules for this to work. I anything, I would say we should create clear docs so users can ask an LLM to generate the JSON files (and provide good errors for debugging).

## **Docs**

In order to explore what is there and how to use it, I asked Cursor LLM to write docs for the feature, which is part of the changes in this PR. I think docs like this would round up this feature quite nicely by being clear about presets for all (via the tagging-schema) but also explain the options to set up custom presets.

---

This would also partially close https://github.com/openstreetmap/iD/issues/5049

### DIFF
--- a/modules/services/maprules.js
+++ b/modules/services/maprules.js
@@ -1,8 +1,20 @@
 import { json as d3_json } from 'd3-fetch';
 
+import { actionChangeTags } from '../actions/change_tags';
+import { t } from '../core/localizer';
+import { validationIssue, validationIssueFix } from '../core/validation';
 import { osmAreaKeys as areaKeys } from '../osm/tags';
 import { utilArrayIntersection } from '../util';
-import { validationIssue } from '../core/validation';
+
+
+/** Keys on a selector that are not tag-condition checks (see #filterRuleChecks). */
+var SELECTOR_META_KEYS = {
+    geometry: true,
+    error: true,
+    warning: true,
+    suggestion: true,
+    fixes: true
+};
 
 
 var buildRuleChecks = function() {
@@ -148,9 +160,10 @@ export default {
     filterRuleChecks: function(selector) {
         var _ruleChecks = this._ruleChecks;
         return Object.keys(selector).reduce(function(rules, key) {
-            if (['geometry', 'error', 'warning'].indexOf(key) === -1) {
-                rules.push(_ruleChecks[key](selector[key]));
+            if (SELECTOR_META_KEYS[key]) {
+                return rules;
             }
+            rules.push(_ruleChecks[key](selector[key]));
             return rules;
         }, []);
     },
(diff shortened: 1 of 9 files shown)


tordans:
@kaligrafy it would be great to get your thought [on this](https://pr-12317--ideditor.netlify.app/#presets=https://gist.githubusercontent.com/tordans/549a328ccff34963192899113c73d35a/raw/f575e37198d51350050f661d00e6e02f8461d1f3/bench-presets.json&maprules=https://gist.githubusercontent.com/tordans/549a328ccff34963192899113c73d35a/raw/f575e37198d51350050f661d00e6e02f8461d1f3/bench-rules.json&map=20.00/52.47414/13.44576&disable_features=boundaries&locale=en&background=Berlin-2024&id=n12518261746).

modules/services/maprules.js
```diff
@@ -194,6 +245,57 @@ export default {
 
     // adds from mapcss-parse selector check...
     addRule: function(selector) {
+        var hasError = typeof selector.error === 'string' && selector.error.length > 0;
+        var hasWarning = typeof selector.warning === 'string' && selector.warning.length > 0;
+        var hasSuggestion = typeof selector.suggestion === 'string' && selector.suggestion.length > 0;
+        var messageKeyCount = (hasError ? 1 : 0) + (hasWarning ? 1 : 0) + (hasSuggestion ? 1 : 0);
+
+        if (messageKeyCount === 0) {
+            console.warn(  // eslint-disable-line no-console
```
matkoniecz, line 254: maybe we should disable `no-console` rule altogether? Rather having such rubbish comments?

matkoniecz:
Is there a reason for doing it as a parameter rather than a setting?

I guess that parameter makes easier to share it

modules/services/maprules.js
tordans: Personally I like it because it means your get a reminder to remove your sometimes very spammy debug logging and have to make an explicit decision.

tordans:
> Is there a reason for doing it as a parameter rather than a setting?
> I guess that parameter makes easier to share it

What would be the alternative? To have an upload field where users upload the JSON file and then configure the editor in this way? – This could be a workflow and also both can be possible at the same time The URL version has the advantage to be shareable and bookmarkable, though.

For this exploration, though, I did not look into this at all but just wanted to see what the existing (broken) feature was all about and how we can maybe revive and improve it without too much change.

matkoniecz:
> What would be the alternative? To have an upload field where users upload the JSON file and then configure the editor in this way? – This could be a workflow and also both can be possible at the same time The URL version has the advantage to be shareable and bookmarkable, though.

I thought about setting in UI where user sets such paramater

tyrasd:
see https://github.com/openstreetmap/iD/pull/12397#issuecomment-4717205986: for now we decided to get rid of the (defunct) maprules code, with the potential to revisit it later.

PR #12329, opened 2026-05-11
fix: restore building:part stroke width in partial fill mode....

### LINKED ISSUE #12326: `building:part` fills no longer thinner than `building`
### URL

https://www.openstreetmap.org/edit#map=21/52.5090170/13.3744665

_No response_

`building:part` partial fill (stroke width) used to be thinner than `building`. This is no longer the case making it harder to edit geometry.

_No response_

v2.40.0

_No response_

deepak-s-git:
Fixes #12326

In v2.40.0....the special-case handling that converted `building:part` to `building_part` for CSS class names was removed......but the deeper issue is that `building` appears before `building:part` in the primaries array....so features tagged with both `building=yes` and `building:part=yes` get classified as `tag-building` instead of `tag-building_part`.....causing 
the partial fill stroke to render at the same thickness as a regular building

Fix: move `building:part` before `building` in the primaries array so it takes priority....restoring the thinner stroke width for building parts in partial fill mode.

### DIFF
--- a/modules/svg/tag_classes.js
+++ b/modules/svg/tag_classes.js
@@ -4,10 +4,10 @@ import { osmPathHighwayTagValues, osmPavedTags, osmSemipavedTags, osmLifecyclePr
 
 export function svgTagClasses() {
     const primaries = [
-        'building', 'highway', 'railway', 'waterway', 'aeroway', 'aerialway',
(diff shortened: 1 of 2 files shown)


tyrasd:
this had already been fixed in #12301 (d3e2c007a1b176e8d1865946eaeb88b340bedf58)

PR #12339, opened 2026-05-14
fix more false positives from the mismatched_geometry validator

### LINKED ISSUE #8885: Gate incorrectly flagged "Gate should be a line, not a point"
image
In the wiki the examples use nodes.
image
image

### LINKED ISSUE #12235: mismatched_geometry validator uses the wrong message for presets that only allow vertex+line
While #11590 solved 1 bug with the `mismatched_geometry` validator, there is another bug in the opposite situation: a preset that support `vertex` + `line`, but not `point`.

Current behaviour:

Image

Expected (same message as `golf=tee`):

Image

2.39.6

k-yle:
Closes #8885, Closes #12235

Sadly no one sent an LLM PR to solve this issue, despite the  https://github.com/openstreetmap/iD/labels/new%20contributor%20opportunity label, so here's my "manual" solution 😆

#### 1. standalone point with `barrier=gate`

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>Image</td>
<td>image</td>
</tr>
</table>

#### 2. closed line with `junction=yes` 

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>Image</td>
<td>image</td>
</tr>
</table>

also added several unit tests for existing situations that were untested.


modules/validations/mismatched_geometry.js
```diff
@@ -438,8 +471,8 @@ export function validationMismatchedGeometry() {
         var vertexPoint = vertexPointIssue(entity, graph);
         if (vertexPoint) return [vertexPoint];
 
-        var lineAsArea = lineTaggedAsAreaIssue(entity);
-        if (lineAsArea) return [lineAsArea];
+        var areaAsLine = areaTaggedAsLineIssue(entity);
```
k-yle, line 474: i did not change this code, just fixed the confusing function name. it handles the opposite situation

PR #12340, opened 2026-05-14
Add invert filter for background imagery

kalucky0:
I've been doing some mapping lately in areas where the aerial imagery is pretty outdated, so I've had to rely on geodetic plans. The problem is those plans are mostly white, which makes it super hard to see what I'm drawing and just starts to hurt my eyes after a while.

I started adding `filter: invert(1);` through the browser dev tools and I figured I'd just build it in as a feature instead of doing it every time. I know the contributing guide says to open an issue for discussion first, but this felt like such a small, specific addition that I just went for it.

I tried to keep the changes as minimal as possible and stuck to the existing coding style in the files I was editing so I didn't mess up the formatting.

image

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -836,6 +836,7 @@ en:
     contrast: Contrast
     saturation: Saturation
     sharpness: Sharpness
+    invert: Invert
     minimap:
       description: Show Minimap
       tooltip: Show a zoomed out map to help locate the area currently displayed.
--- a/modules/renderer/background.js
+++ b/modules/renderer/background.js
@@ -28,6 +28,7 @@ export function rendererBackground(context) {
   let _contrast = 1;
   let _saturation = 1;
   let _sharpness = 1;
+  let _invert = 0;
 
 
   function ensureImageryIndex() {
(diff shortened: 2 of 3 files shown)


tyrasd (changes requested):
I would rather prefer to this implemented as a simple checkbox (default unchecked) as intermediate values for `invert()` don't make sense, at least for the given use case.

andrewharvey:
Should this be limited to only having an effect on `category: "map"` layers? It doesn't make much sense on satellite/aerial imagery as all the colours are inverted.

In the case where a satellite/aerial background is active, should this option then apply to the overlays instead? As generally the overlays would be a light colour to contrast the generally dark satellite/aerial, but in some cases you might have bright satellite/aerials and want to invert the overlays on top.

For example here the white lines are the overlay which are clear over the darker water areas, but not so clear over the whiter beach and buildings.

image

matkoniecz:
> It doesn't make much sense on satellite/aerial imagery as all the colours are inverted.

maybe if someone maps ice sheets?

matkoniecz:
In general I would be fine with merging it also without adding such pile of features.

@kalucky0 do you plan to change it into checkboxed version?

PR #12343, opened 2026-05-15
Render short radio field options inline on one row

FloEdelmann:
Related to https://github.com/openstreetmap/id-tagging-schema/issues/2049 (CC @owiecc as the author of that issue).

When all options of a radio field fit side-by-side in the available width, render them horizontally as equal-width cells instead of stacking them vertically. This saves vertical space for compact option sets like "Yes / Limited / No".

The available/needed size is measured at runtime (so it is locale-aware) and updated via `ResizeObserver`, there is no hardcoded option count or pixel threshold. If `ResizeObserver` is not available, the default vertically stacked layout remains.

Test object: https://pr-12343--ideditor.netlify.app/#map=20.49/48.09836/11.78720&background=Bing&id=n1377621859

### Screenshot

grafik

### Screencast

https://github.com/user-attachments/assets/fce3c04f-b15a-4fa8-aa8c-3c6af80ed11a


tordans:
I think this is a good idea. It would be interesting to see if we can get this working just with css and container queries. Maybe an LLM can test this out for us...

One other thing: we recently prepared to have the radio buttons work with icons same as the dropdowns. But that should work just as well for this case.

FloEdelmann:
> It would be interesting to see if we can get this working just with css and container queries. Maybe an LLM can test this out for us...

I don't think that suffices, and Claude Sonnet 4.6 agrees:

> **No — container queries aren't sufficient here.** Here's why:
>
> The core logic is: *switch to horizontal layout only when `maxLabelWidth × labelCount ≤ containerWidth`*. Both `maxLabelWidth` and `labelCount` are **dynamic per-field** (they depend on the specific options and their text). CSS container queries only support **fixed breakpoints** — there's no way to express "apply this style when the container is wide enough to fit all N labels at their natural widths" without hardcoding a pixel value.
>
> A CSS-only flex approach with `flex-wrap: wrap` and `min-width: max-content` on labels would cause wrapping into multiple partial rows — not the clean vertical→horizontal toggle that's desired.
>
> To get CSS-only behavior that looks correct, you'd need a fixed `min-width` breakpoint per element, which would require either:
>
> - A hardcoded threshold (doesn't adapt to label content), or
> - Inline styles set per-field (back to JS anyway)
>
> The `ResizeObserver` + JS measurement approach on this branch is the right tool for this job.

css/80_app.css
```diff
@@ -2095,6 +2095,30 @@ input.date-selector {
...
+    overflow: hidden;
+}
+.form-field-input-radio.one-line > label {
+    flex: 1 1 0;
+    min-width: 0;
+    width: auto;
+}
+.ideditor[dir='ltr'] .form-field-input-radio.one-line > label:first-of-type,
+.ideditor[dir='rtl'] .form-field-input-radio.one-line > label:last-of-type {
+    border-radius: 0 0 0 4px;
```
k-yle: (question, not suggestion)

[CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Logical_properties_and_values) have been supported by all browsers for over 5 years, but we don't use them yet.

we could remove all the duplicate CSS blocks for RTL by switching to _logical properties_ (such as [`border-start-start-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/border-start-start-radius)) Is that something iD wants to do eventually?

modules/ui/fields/radio.js
```diff
@@ -45,6 +45,19 @@ export function uiFieldRadio(field, context) {
             .append('span')
             .attr('class', 'placeholder');
 
+        enter.each(function() {
+            if (typeof ResizeObserver === 'undefined') return;
+            const element = this;
+            const observer = new ResizeObserver(() => {
+                if (!element.isConnected) {
+                    observer.disconnect();
```
k-yle: i assume this block is a sufficient alternative to unsubscribing in `wrap.exit()` ?

FloEdelmann: If I understand correctly, `wrap.exit()` returns all elements that are to be removed _right now_ (i.e. an empty selection in this function where the radio field is constructed) and is not an event handler that will be called before the element will be removed.

That said, `if (!element.isConnected)` is certainly rather hacky and I'm open to better suggestions :slightly_smiling_face:

css/80_app.css
FloEdelmann: Done in [`a52929d`](https://github.com/openstreetmap/iD/pull/12343/commits/a52929d0530303e7d5adbb0ff5847eac89b4ec36).

modules/ui/fields/radio.js
tyrasd: Yeah, `.exit()` does not work here, as the whole field/entity-editor is removed from the DOM somewhere higher up in the DOM tree.

>  is certainly rather hacky

I was pondering about a more elegant way to do this… 

One downside of the current implementation is that when the inspector is hidden due to an OSM feature being de-selected, the inspector is briefly in a hidden state (`.inspector-hidden` / `display: none;`), where the `element.isConnected` is still `true`, but the observer will not trigger further resize events because the element is not changing size when passing from `display: none` to _disconnected_. Thus, we would potentially accumulate observers and non-garbagecollectable DOM elements over time. It's probably a negligible memory-leak, but still not ideal.

I guess the "iD way" to do implement this is be to look for size changes on the _sidebar_ module itself, and re-render the whole _inspector_ on every resize. -> see 519e37ffb

tyrasd (commented):
I like this a lot, thanks! 

@FloEdelmann would you be so kind to test my added changes in 519e37f and append a short changelog entry for this new feature (don't hesitate to attribute yourself btw!).

FloEdelmann:
Thanks for pushing your improvement! I've tested it and it works fine :slightly_smiling_face: 

I've added a changelog entry.

tordans:
This is a good improvement. Tested in https://pr-12343--ideditor.netlify.app/#disable_features=boundaries&map=20.00/52.47389/13.44515&background=Bing&id=n5503778828

---

One follow up idea: Also use a 2 col layout when applicable for cases like
> image

PR #12345, opened 2026-05-15
Fix radio field crash with nested option strings

FloEdelmann:
When a `radio` field uses `stringsCrossReference` to reference another field whose option strings are structured as `{ title, description }` objects (rather than flat strings), rendering the field throws:

```
TypeError: callback.apply is not a function
```

This happens because `tInfo` returns the raw `{ title, description }` object in `texts`, and `t.append` then calls `selection.call(object)` — but D3's `selection.call()` expects a function.

The `combo` field already handles this correctly via its `getLabelId` helper, which checks for a `.title` sub-key first. The `radio` field was missing equivalent logic.

This issue was exposed by [openstreetmap/id-tagging-schema#2268](https://github.com/openstreetmap/id-tagging-schema/pull/2268), which changes the `parking_entrance` field from type `combo` to type `radio`. The `parking` field it references via `stringsCrossReference` uses nested `{ title, description }` option strings.

### DIFF
--- a/modules/ui/fields/radio.js
+++ b/modules/ui/fields/radio.js
@@ -66,7 +66,12 @@ export function uiFieldRadio(field, context) {
 
         enter
             .append('span')
-            .each(function(d) { strings.t.append('options.' + d, { 'default': d })(d3_select(this)); });
+            .each(function(d) {
+                const labelId = strings.hasTextForStringId('options.' + d + '.title')
+                    ? 'options.' + d + '.title'
+                    : 'options.' + d;
+                strings.t.append(labelId, { 'default': d })(d3_select(this));
+            });
 
         labels = labels
             .merge(enter);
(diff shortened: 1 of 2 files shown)


FloEdelmann:
Yay, issue **`#12345`** :partying_face: :smile:

tordans:
This will not happen anymore after https://github.com/ideditor/schema-builder/pull/281, right?

tyrasd:
> This will not happen anymore after https://github.com/ideditor/schema-builder/pull/281, right?

No, this fix is still relevant also when we have pre-resolved strings: If a `radio` field references strings from a `combo` field with "nested" strings (i.e. separate `title` and `description` properties), the `radio` field still needs to explicitly choose the `.title` string instead of the simple "non-nested" string that radio fields usually have.

* [x] We might want to document how `radio` fields interpret strings in the schema-builder docs: https://github.com/ideditor/schema-builder/#strings

tyrasd (approved):

Issue #12337, opened 2026-05-14
Upgrade to name-suggestion-index v7.2


Dimitar5555:
NSI has a new minor release which seems to have a few potentially breaking changes. It would be nice if iD uses the newer version so we use newer data.

<details><summary>Changelog</summary>

> # 7.1.YYYYMMDD
> ##### 2026-Apr-27
> * ⚠️  Bump to [`@rapideditor/location-conflation`](https://github.com/rapideditor/location-conflation) v3 — downstream consumers must upgrade simultaneously
> * ⚠️ `Matcher.buildLocationIndex(data: NsiData, loco?: LocationConflation)` now expects a `LocationConflation` v3 instance.
>   * Consumers who call `matcher.buildLocationIndex(data, loco)` should now pass in their own configured `LocationConflation` instance so indexing and lookups share the same cache
>   * This eliminates the need for downstream apps (e.g. Rapid) to monkey-patch the matcher to use a shared `LocationConflation` cache!
>   * Removed the `which-polygon` dependency — location lookups are now handled inside `LocationConflation` v3.
>   * Note that the `loco` param is optional — if omitted, the `Matcher` can create its own local `LocationConflation` resolver.
> * New: Preset generating functions are now exported from `name-suggestion-index` as pure, in-memory functions
>   * `buildIDPresets(data: NsiData, opts: BuildIDPresetsOptions): BuildIDPresetsResult` - Generates iD presets JSON
>   * `buildJOSMPresets(data: NsiData, opts: BuildJOSMPresetsOptions): XMLBuilder` - Generated JOSM presets XML
>   * These accept NSI data, source iD presets, wikidata, dissolutions) as arguments — no file I/O, no console output, no input mutation
>   * Suitable for downstream apps (e.g. Rapid) that fetch NSI data on-the-fly and want to generate presets in the browser
>   * `dist.ts` now wraps these functions for the file-writing CLI behavior (we still provide the presets under `dist/presets/` as before)
> * New: Project now provides a complete set of TypeScript types to cover all data structures and file shapes.
> 

</details>

tordans:
Those breaking changes sound to me this should get a major version increase?

https://github.com/osmlab/name-suggestion-index/blob/main/CHANGELOG.md

tyrasd:
> should get a major version increase?

[...]

PR #12349, opened 2026-05-17
chore: update name-suggestion-index to v7.2

bjornstar:
Removed the hack which should no longer be necessary according to [the changes in v7.1](https://github.com/osmlab/name-suggestion-index/blob/main/CHANGELOG.md#71yyyymmdd)

Recommend @bhousel to take a look.

Tested locally and got good, locally appropriate suggestions.


tyrasd:
Thanks. v7.2 also provides the new method `buildIDPresets`, which is used with 25af7c9. PS: see https://github.com/openstreetmap/iD/issues/12337#issuecomment-4466659578 for a potential further improvement regarding this method.

PR #12353, opened 2026-05-18
Add high-resolution toggle for Mapilio pano viewer

sezerbozbiyik:
- Adds a “High Resolution” toggle to switch Mapilio pano images between 1080p and 2080p.
- Refactors the image load flow into a reusable loadTheImage function after getImageData.
- Keeps attribution details (captured by + timestamp) intact alongside the new option.


tyrasd (changes requested):
it would be nice if the checkbox would stay active during the mapping session as long as it is not manually unchecked. See the behaviour of the similar checkbox on the panoramax photo service.

sezerbozbiyik:
hi @tyrasd,

Thanks for the suggestion! I’ve updated the Mapilio HD checkbox to persist across the mapping session

tyrasd (approved):

tyrasd:
merged as ba55f72a1

PR #12359, opened 2026-05-19
Add "flow_direction" to directed tags.

### LINKED ISSUE #12161: Use flow_direction for waterway error detection
### Description

Currently, the Warning message
> "\<Subject\> flows against/\{away from\} a connected waterway"

is controlled by the [oneway Key](https://wiki.openstreetmap.org/wiki/Key:oneway), even though that Key actually relates to _traffic_ direction.

Thus, one may be tempted to add "oneway=no" when the water flows two ways or in an unknown direction, despite this being objectively the incorrect tag.

Please let the Warning messages and the displayed arrows be controlled by the [flow_direction Key](https://wiki.openstreetmap.org/wiki/Key:flow_direction) instead, when it comes to waterways specifically.

Waterways should function as having "flow_direction=forward" by default, which would be overruled by tagging flow_direction and setting it to a different value. If flow_direction is set to an undefined value like "unknown", then the waterway should behave the same as a highway without the oneway Key, i.e. give no errors and have no arrows.

In theory, flow_direction Key should perhaps be modified to use the same values as the oneway Key.

### Screenshots

_No response_

emil813-cmd:
Adds flow_direction to directed tags (like 'oneway' and 'conveying').

### DIFF
--- a/modules/osm/tags.ts
+++ b/modules/osm/tags.ts
@@ -239,10 +239,16 @@ export const osmOneWayBiDirectionalTags: TagDictionary<boolean> = {
         'reversible': true,
     },
     'flow_direction': {
-        'both' : true,
-        'unknown' : true,
-        // instead of BiDirectional, 'unknown' might fit better under a new "UnDirectional" category
+        'both' : true
     },
+};
+export const osmOneWayNoTags: TagDictionary<boolean> = {
+    'oneway': {
+        'no': true
+    },
+    'flow_direction': {
+        'unknown' : true
+    }
 };
 export const osmOneWayTags = merge(
     osmOneWayForwardTags,
--- a/modules/osm/tags.ts
+++ b/modules/osm/tags.ts
@@ -214,7 +214,10 @@ export const osmOneWayForwardTags: TagDictionary<boolean> = {
         'spillway': true,
         'stream': true,
         'tidal_channel': true
-    }
+    },
+    'flow_direction': {
+        'forward' : true,
+    },
 };
 export const osmOneWayBackwardTags: TagDictionary<boolean> = {
     'conveying': {
@@ -222,6 +225,9 @@ export const osmOneWayBackwardTags: TagDictionary<boolean> = {
     },
     'oneway': {
         '-1': true,
+    },
+    'flow_direction': {
+        'backward' : true,
     },
 };
 export const osmOneWayBiDirectionalTags: TagDictionary<boolean> = {
@@ -231,6 +237,11 @@ export const osmOneWayBiDirectionalTags: TagDictionary<boolean> = {
     'oneway': {
         'alternating': true,
         'reversible': true,
+    },
+    'flow_direction': {
+        'both' : true,
+        'unknown' : true,
+        // instead of BiDirectional, 'unknown' might fit better under a new "UnDirectional" category
     },
 };
 export const osmOneWayTags = merge(
--- a/modules/osm/way.js
+++ b/modules/osm/way.js
@@ -3,7 +3,7 @@ import { geoArea as d3_geoArea } from 'd3-geo';
 import { geoExtent, geoVecCross } from '../geo';
 import { osmEntity } from './entity';
 import { osmLanes } from './lanes';
-import { osmTagSuggestingArea, osmRightSideIsInsideTags, osmRemoveLifecyclePrefix, osmOneWayBiDirectionalTags, osmOneWayBackwardTags, osmOneWayForwardTags, osmOneWayTags } from './tags';
+import { osmTagSuggestingArea, osmRightSideIsInsideTags, osmRemoveLifecyclePrefix, osmOneWayBiDirectionalTags, osmOneWayBackwardTags, osmOneWayForwardTags, osmOneWayTags, osmOneWayNoTags } from './tags';
 import { utilArrayUniq, utilCheckTagDictionary } from '../util';
 import { osmIdManager } from './id_manager';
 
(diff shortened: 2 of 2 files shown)


modules/osm/tags.ts
```diff
@@ -235,6 +241,11 @@ export const osmOneWayBiDirectionalTags = {
         'alternating': true,
         'reversible': true,
     },
+    'flow_direction': {
+        'both' : true,
+        'unknown' : true,
+        // instead of BiDirectional, 'unknown' might fit better under a new "UnDirectional" category
```
tyrasd: would it not be better to just omit `unknown` then? then, no arrows are shown at all, which is more true to the actual meaning of the tag.

https://wiki.openstreetmap.org/wiki/Key%3Aflow_direction does not include the value `unknown` either, which would also be an argument against including it.

emil813-cmd: No, this tag is meant to be used on waterways which inherently have arrows displayed.
I'm not satisfied with using oneway=no, since it describes *traffic* and doesn't have an "unknown" value.

I don't know what it'd take to make flow_direction able to just remove arrows, but after some testing how oneway and "conveying" work, this seems like the best option.

emil813-cmd: https://github.com/openstreetmap/iD/blob/develop/modules/osm/way.js#L144
I've now found out why `oneway=no` removes arrows entirely. It has been hardcoded into both `isOneWayForwards`, `isOneWayBackwards`, `isBiDirectional` and `isOneWay`.

I don't find this to be an elegant solution, so I don't want to replicate that.

tyrasd: Valid point. Implementing a more elegant solution is relatively straightforward (it was just not necessary, as `oneway=no` was the only tags with this connotation until now). I added this to this branch in 15b209978

That said, I'm still very hesitant to merge this as is: The tag value `unknown` is neither documented on the [current wiki page](https://wiki.openstreetmap.org/w/index.php?title=Key:flow_direction&oldid=2947958#Values) nor was it mentioned in the [original proposal](https://wiki.openstreetmap.org/wiki/Proposal:PipelineExtension). Such a tag must first be discussed in the general community, then properly documented in the wiki, and only then implemented in editors.

There are two options now:
1. start discussion about the new tag `flow_direction=unknown`, and after a grace period, merge this PR
2. amend the PR to omit the `flow_direction=unknown` tag for now (then it can be merged sooner), and start the discussion about the new tag in parallel (with a follow-up PR to complete the implementation at a later point)

emil813-cmd: Value "unknown" is defined for many other keys with the same meaning, so it shouldn't be a problem. Alternatively, as you suggested in the first comment, "unknown" doesn't need to be defined if unfamiliar values of flow_direction were to default to directionlessness.

I am looking forward to checking out your changes. Will they appear in https://ideditor.netlify.app/ soon?

The tags interact somewhat strangely as is. flow_direction and the like need to be able to overrule the direction inherent to waterway=ditch. Have you somehow given lower priority to the direction of "feature type" keys like waterway and aerialway?

(The reason that *I* am the one who invented the flow_direction=unknown combination is that I've been placing waterways in forests in Denmark using Terrain Shadow Map, and even with a height map in a separate web pane, many ditches are too flat and/or short to determine direction.)

modules/osm/tags.ts
```diff
@@ -232,6 +238,17 @@ export const osmOneWayBiDirectionalTags: TagDictionary<boolean> = {
...
+        'both' : true
+    },
+};
+export const osmOneWayNoTags: TagDictionary<boolean> = {
+    'oneway': {
+        'no': true
+    },
+    'flow_direction': {
+        'unknown' : true
+    }
```
matkoniecz, line 251:
```suggestion
    }
```

as this tag is undocumented, see https://wiki.openstreetmap.org/wiki/Special:WhatLinksHere/Tag:flow_direction%3Dunknown

emil813-cmd: The Tag ele=45 is also undocumented. The value "unknown" is defined for many other Keys. Its meaning is obvious.
And it would be much better to use than oneway=no, which has both the wrong Key (traffic direction) and Value (no direction, rather than unknown).

I am not confident enough to make edits to the wiki myself.

k-yle:
since you're the inventor of the `flow_direction` tag, can you please help me understand whether this tag should be used on waterslides, or if `oneway` should be used? the wiki doesn't explain this

modules/osm/tags.ts
matkoniecz: > The Tag ele=45 is also undocumented

1) it has no special support in iD or iD tagging schema

2) there is difference between key with range of generic values and one with few enum-type values

3) there are fields where `unknown` is unwanted/poor idea/different value is used for such meaning

emil813-cmd: I know. But the use and need of flow_direction=unknown is obvious, and I dare not create create a wiki page myself.

emil813-cmd:
I didn't invent the `flow_direction` Key, I merely started using the `flow_direction=unknown` Tag.

On waterslides, the substrate (water) and traffic (people in swimwear) always move in the same direction, so I think the more common Key `oneway` should be used. But [attraction=water_slide](https://wiki.openstreetmap.org/wiki/Tag:attraction%3Dwater_slide) already defines a default direction (forward), so there will rarely be a need for either Key.

(I can't visualize a waterslide where people move against the water or where the direction changes, but tell me if you know of one.)

modules/osm/tags.ts
matkoniecz: You can make a simple thread asking on forums what they ask about this tagging idea and whether anyone would be against documenting it

emil813-cmd:
Speaking of the differences between `flow_direction` and `oneway`, suppose you have a `waterway=stream` Way **X** where you are only allowed to sail *with* the flow direction, and a `side_stream` **Y** parallel to it where you are only allowed to sail *against* the flow direction. I don't know if this scenario is realistic.

The iD-editor doesn't handle traffic and water flow separately, so it will get confused. The type of confusion will depend on the order of priority between the directions of `waterway=stream`, `oneway` and `flow_direction`.

(It is obviously not worth separating traffic and substrate flow in error-detection just for this hypothetical issue. Way **Y** should probably just be excluded from direction-checking.)

PR #12373, opened 2026-05-22
fix: add target=_blank to Read more link in custom background settings

### LINKED ISSUE #12370: "Read more" link under 'Custom Background Settings' not working
### URL

https://www.openstreetmap.org/#map=19/0.000000/0.000000

1. Go to: https://www.openstreetmap.org/#map=19/0.000000/0.000000
2. Hit button "Edit"
3. Hit button "Background Settings"
4. Hit button "Edit custom background"
5. Hit blue link "Read more"

Expected behavior: opening new browser tab to address "[https://wiki.openstreetmap.org/wiki/FAQ#What_images_and_maps_may_I_use_to_make_maps_from?](https://wiki.openstreetmap.org/wiki/FAQ#What_images_and_maps_may_I_use_to_make_maps_from?)"

Actual behavior:
a) under Firefox: no action
b) under Chrome: crashed window, see recording under section "Screenshot(s) or anything else?"

Custom Background Settings / Read more:

Image

Chrome browser behavior:

Image

2.40

Chrome

Aryannaik-max:
Fixes #12370

## Problem
The "Read more" link in Custom Background Settings had no `target` 
attribute. When iD runs inside the iframe on openstreetmap.org, the 
link tries to open inside the iframe which is blocked by policy, 
so nothing happens.

## Fix
Added `target="_blank"` and `rel="noopener noreferrer"` to all links 
generated by `marked()` in the custom background settings dialog.


modules/ui/settings/custom_background.js
tyrasd:
```suggestion
<a href="${t('settings.custom_background.instructions.license_faq_url')}" target="_blank">${t.html('settings.custom_background.instructions.license_faq')}</a>
```

tyrasd (changes requested):
this seems much more complicated than needed. why not directly set the target in the link supplied markdown? see suggestion below

modules/ui/settings/custom_background.js
matkoniecz: @Aryannaik-max ping?

Aryannaik-max: Hey, sorry for the delayed response. I applied the suggested fix, thanks for the ping @matkoniecz and for the cleaner suggestion @tyrasd!

PR #12374, opened 2026-05-22
Render cyclist waiting aid side arrow

RudyTheDev:
This PR allows the side arrow to render for cyclist waiting aids when unambiguous, i.e. `direction` is given:

cwa side cases

This makes mapping the `side` less confusing, especially since waiting aids use a different meaning of "side".

Original PR where rendering exception occurred #10303. Some relevant side-related comments and discussion at #10128. In short: the proposal used a non-standard meaning of `side`; it could've been changed, but wasn't; now waiting aids have unique meaning of `side`.

---

The [comment in code](https://github.com/openstreetmap/iD/commit/0a351fa93f1736b39ba193bbc0eafd28d14365dc#diff-9e0fb29b62a2af32e56479f07a425bdd4995d191046479b79b7cd797b838c8d9L100) previously implied that it is impossible to determine the side arrow because the proposal's use of `side` is ambiguous. However, that is not technically correct. The `side` here means "relative to cyclist" (unlike other features where it is "relative to way"). So it is not ambiguous if the direction of the waiting aid is known. This is basically what this PR implements - if the `direction` is known, then the `side` is not ambiguous, therefore the arrow can be rendered. The caveat is to flip it around if the `direction=backward`. The way's actual direction is ignored/irrelevant.

The code previously also assumed that this wouldn't be ambiguous for one-way ways - but it is actually ambiguous in those cases too (e.g. road with `oneway=yes`, `oneway:bicycle=no` and waiting aid in the "opposite" direction). In fact, the code always showed the side arrow based on the way's direction, regardless if the `direction` was actually set, leading to a case where it would be wrong:

cwa side on oneway wrong

Thankfully, this would only be wrong in that specific rare condition, so I doubt there is any significant number of mistagged features due to this.

I think the intent here was that iD could use way directionality like it does for other features that use `side`, but that assumption doesn't really work. Like, the meaning of `side` is not wrong for waiting aids - it's different. Anyway, I removed this way "sidedness" assumption based on the way being one-way. I think it's best that the mapper just specifies `direction` in those rare cases so nothing has to be assumed. I don't think iD should hard-code (any more) assumptions.

Also, technically the case where `side=both` is also unambiguous, because `direction` doesn't matter - it's both sides however you look at it. So I included this case too (although the preset doesn't even show that option). There are apparently cases like that, e.g. [example](https://www.openstreetmap.org/node/11932470578) - [mapillary](https://www.mapillary.com/app/?pKey=1287710125863818&focus=photo). So it will render `side=both` pretty much always:

cwa both sided

No LLM code used in this one.

### DIFF
--- a/modules/osm/node.js
+++ b/modules/osm/node.js
@@ -97,18 +97,17 @@ const prototype = {
         } else {
             // generic side tag
 
-            // unfortunately, the proposal for highway=cyclist_waiting_aid used
-            // an ambiguous definition for `side`, which basically makes the tag
-            // useless in the situation where the parent way (the cycleway) is
-            // bidirectional. It's impossible for software to determine which
-            // direction the mapper is referring to…
-            const hideSideTag = (
+            // for `highway=cyclist_waiting_aid`, `side` (i.e. from cyclist's perspective) will be ambiguous
+            // unless the `direction` is explicitly specified (even on one-way roads).
+            const isSideTagAmbiguous = (
                 this.tags.highway === 'cyclist_waiting_aid' &&
-                resolver.parentWays(this).some(way => !way.isOneWayForwards())
+                this.tags.side !== 'both' &&
+                this.tags.direction !== 'forward' &&
+                this.tags.direction !== 'backward'
             );
 
             const sideTag = SIDE_TAGS.map(key => this.tags[key]).find(Boolean);
-            if (SIDES.has(sideTag?.toLowerCase()) && !hideSideTag) {
+            if (SIDES.has(sideTag?.toLowerCase()) && !isSideTagAmbiguous) {
                 rawValues.push({
                     type: 'side',
                     value: sideTag?.toLowerCase(),
@@ -196,6 +195,12 @@ const prototype = {
 
                 const isSide = type === 'side' && SIDES.has(v);
 
+                // In case of `highway=cyclist_waiting_aid`, if the `side` is to be shown, then
+                // it will be explicitly specified by the `direction` from the cyclist's perspective (not way's perspective)
+                const isSideFlipped = isSide &&
+                    this.tags.highway === 'cyclist_waiting_aid' &&
+                    this.tags.direction === 'backward';
+
                 // string direction - inspect parent ways
                 const lookBackward =
                     (this.tags['traffic_sign:backward'] || v === (isSide ? 'left' : 'backward') || v === 'both' || v === 'all');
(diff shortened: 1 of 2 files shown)


RudyTheDev:
A bunch of examples around the intersection cycleways: https://pr-12374--ideditor.netlify.app/#map=20.00/52.20125/20.93626

matkoniecz:
at least for me this direction arrow is confusing

is it showing where aid is located? or is showing direction of cyclist that would use it? something else?

RudyTheDev:
It's showing both. The preset specifies this:

image

Regardless, this was already implemented before - I didn't change its appearance, only fixed the logic for this case. It's already implemented on all features using `side`.

image

modules/osm/node.js
```diff
@@ -97,18 +97,17 @@ const prototype = {
...
-            // direction the mapper is referring to…
-            const hideSideTag = (
+            // for `highway=cyclist_waiting_aid`, `side` (i.e. from cyclist's perspective) will be ambiguous
+            // unless the `direction` is explicitly specified (even on one-way roads).
+            const isSideTagAmbiguous = (
                 this.tags.highway === 'cyclist_waiting_aid' &&
-                resolver.parentWays(this).some(way => !way.isOneWayForwards())
+                this.tags.side !== 'both' &&
+                this.tags.direction !== 'forward' &&
+                this.tags.direction !== 'backward'
```
k-yle, line 106: what do you think of [this case](https://pr-12374--ideditor.netlify.app/#map=21.80/-43.53295/172.68405&background=LINZ_NZ_Aerial_Imagery&id=n13012943874)? (`direction=both` + `side=right`)

- Is it logically invalid?
- or does it mean that the handrail in on _your_ right-hand side in both directions? If yes, that would mean that `side=right` = `side=both` 😆

modules/osm/node.js
k-yle, line 1: In parrallel, I think we should update id-tagging-schema and add a prerequisite tag for the `side` field:

```json
  "prerequisiteTag": {
    "key": "direction",
    "values": ["forward", "backward"]
  },
```

This will avoid confusion about the rendering logic, since impossible situations will not be allowed by iD's UI.

this is blocked by https://github.com/openstreetmap/id-tagging-schema/pull/2255

k-yle (commented):
thanks for the PR! I think this is a nice improvement 

_(although i really regret not fixing the poor tag definition when we realised this issue [2 years ago](https://github.com/openstreetmap/iD/issues/10128#issuecomment-1965268713) - now it's way too late)_

modules/osm/node.js
RudyTheDev: Hmm, I can see how this can be considered _technically_ logically valid - whichever direction you use, the handrail is always on _your_ right indeed.

But that's not enough information for visuals. Like, there's no way to disambiguate and render combinations of side arrows for two-directional features without additional assumptions. Like, for one, we would have to consider the driving side of the way, which is by the way [left-hand side here](https://kartaview.org/details/1377634/738/track-info), so `side=right` = `side=inner` or something...

image

This feels like a lot of consideration for a very niche application. Personally, I would just not bother rendering any side arrow stuff for such cases for now. As a mapper, I would be mapping these as 2 nodes.

tyrasd (approved):

PR #12387, opened 2026-05-26
allow typos for showing suggestions in combo dropdowns

### LINKED ISSUE #8802: Combo field should suggest values that almost match the input
If I type `desiganted` into the Allowed Access field or `portugese` into the Cuisine field, the suggestions should include near matches such as `designated` and `portuguese` based on an edit distance to known values. This might be difficult for combo fields that get their suggestions from taginfo, but predefined `options` from id-tagging-schema could still participate in the same kind of fuzzy matching as presets in the search bar or fields in the More Fields dropdown.

https://github.com/openstreetmap/iD/blob/fc6c71160c6378fa16e70436304a3e3d6fdb30e2/modules/presets/collection.js#L130-L132

Fuzzy matching would keep id-tagging-schema from having to maintain an index of likely typos, as in openstreetmap/id-tagging-schema#266.

tyrasd:
fixes #8802

the number of allowed typos depends on the length of the entered string:

* very short strings (up to 4 characters) are only matched exactly to not spam the dropdown list with too many completely unrelated matches
* medium length strings (5-9 chars) allow one character difference (either 1 removed, added or differing character)
* longer strings allow one additional typo for each additional 5 characters


RudyTheDev:
I did encounter a funky case:

softball

I guess the sorting does not take into account the number of "typos" for results, even for a correct value.

I suppose there would be a bunch of these random examples for all sorts of combinations:

tapas

shed

tyrasd:
Strictly speaking, this behaviour is not new (screenshot from current dev branch):



But yes, it is greatly alleviated by the fact that with typos more strings will match.

Currently, the sorting does not take anything into account: It's just the order of the options as defined in the tagging schema (or as returned by taginfo if there are no options defined in the preset field). While typing, iD just filters out any options which definitely not match the given input. Now we do potentially get more results in the list when allowing typos, which I see can be confusing. :thinking: Yes, we could change the order dynamically depending on the user input, but that could lead to the list changing order in unexpected ways as well. I'm not sure what would be better.

Maybe what we could do instead is: if the input text does match an entry fully, or if there are any exact substring matches, then do not consider typos for the filter. That way:

* when entering `tapa`, it will only list `Tapas`
* when entering `Softball`, it will only list `Softball`
* when entering `Shed` it will list `Shed`, `Cowshed` (and a few exact substring matches from taginfo)
* when entering `portugese` it will list `Portuguese` 
* when entering `desiganted` it will list `designated`
* when entering `tapan` it will list both `Japanese` and `Tapas`

Would that be better?

RudyTheDev:
> .. Yes, we could change the order dynamically depending on the user input, but that could lead to the list changing order in unexpected ways as well.

I think from user's perspective once you start typing custom text, you no longer care about some pre-defined order, but want the best result first.

> Maybe what we could do instead is: if the input text does match an entry fully, or if there are any exact substring matches, then do not consider typos for the filter.

Do you think this would not exclude some legitimate "typo results"? Like if I typed "trial" for a "trail map" but the result is "trial chamber" and no option for anything with "trail"? Realistically, I struggle to think of any real examples though even with 4 letter threshold.

In any case, I don't think this is a big issue at all and any extra consideration for better match sorting is a bonus to what is otherwise a really useful feature.

tyrasd:
I tried it again today and sorting the results turned out to be a little trickier to implement than expected, and did not "feel" usability-wise in my opinion (but it could well be that I'm just very used to how this previously worked).

I just now tweaked the limit of when a typo is allowed to 5 input characters, so that tags with only 3 matching characters do not match, which IMHO produced too many false positives.

Issue #10780, opened 2025-02-17
rendering of tracks depends on tag adding order


tyrasd:
(this was originally reported on https://osmus.slack.com/archives/CBK3JLUJU/p1739751098393849 by @rivermont )

It's slightly confusing that the rendering style of tracks with `tracktype=*` and `surface=*` can depend on the order the tags were added to the entities:

> all the tracks have surface=asphalt and tracktype=grade2; the one added last determines the rendering

![Image](https://github.com/user-attachments/assets/8b398bdf-b892-4e68-b92d-b637731d553b)

youssefelzedy:
Which road should appear in this case (asphalt or grade 2), or will an error occur?

youssefelzedy:
if you can tell me about this issue more, I want to work on it @tyrasd  @rivermont

tordans:
I don't see any reason why we should not use the styling that iD uses for existing ways (when a way is not edited). The issue here is, that editing a way results in inconsistent styles. So the goal is to make those styles consistent by always look like a way that was not edited (but has the same tags).

youssefelzedy:
Yes, you are right. I also did not understand the issue correctly; I will work on it. Thanks for your response @tordans .

youssefelzedy:
Is there any way (value in the osmway's object) to tell me if this osmway has changed before or not I searched a lot if any but I did not find it. so I now start to think about to add an option for this and I want know from you if I want to make this how to make it?

tyrasd:
check out where and how `tag-unpaved`/`tag-paved` is applied to the feature's css classes.

youssefelzedy:
![Image](https://github.com/user-attachments/assets/a40e9548-7e1a-4910-86ee-cb0924481513)
in `tag_classes.js` file 

I got if the primary is highway and not any of the path highway tag values or if the primary is aeroway
and in the case above the `surface=asphalt and tracktype=grade2` surface will be "paved", because k in osmPavedTags(surface=asphalt) returns true.

youssefelzedy:
Regarding the arrangement of Tags, depending on primaries and secondaries, they are placed in a fixed order, as you show in these figures.

![Image](https://github.com/user-attachments/assets/3ed1518c-df71-4f7b-9ded-8aefefeb595b)

![Image](https://github.com/user-attachments/assets/56de76af-d2a5-4b84-a527-363435af7b92)

[...]

PR #12392, opened 2026-05-27
Add track grade-based styling

RudyTheDev:
This PR adds different style rendering to `highway=track` based on their `tracktype`:

grade previews

This is similar to how Karto or OsmAnd or others render it - but here specifically changes only the casing dash pattern for `grade2`-`grade5` and "ungraded". `grade1` remains as is - matching current paved road style. Ungraded has an dot-dash pattern to indicate it's unknown.

The current track dash pattern is roughly between proposed 3 and 4, which is sort of the median of grade values.

current compare

The actual values are like this:

```
paved/grade1:   solid (this is current use)
track grade2:   10, 4
track grade3:   6, 4
unpaved roads:  4, 4 (this is current use)
track grade4:   4, 6
track grade5:   2, 8
track ungraded: 7, 5, 3, 5
```

I experimented with a bunch of different layouts and ended up with the example above, which is roughly a linear progression from paved to grade5. It's basically a couple pixel difference, so there isn't much leeway for uniqueness. I played around with some ideas, but I don't think there's a lot of different ways to do it better. Changing main line/stroke looks weird and clashes with life-cycle rendering. Changing colors is not obvious what they mean and is way too distracting. Casing is already used for surface differentiation, so that seems to work best here. Adding line dash pattern looks like it's a crossing or railway or something, but it could also be an option. Perhaps the casing could be a bit lighter and more contrasted. And the dash patterns could be different.

I did not add unique low-level zoom variants, because I think it works fine as-is. I don't think adding even more dash patterns makes anything more legible. If anything, tracks are mostly found in wide expanses of forests and fields, so having them distinguishable at lower zooms is arguably even more useful.

low zoom

Implementation-wise I added a bit of code to detect an ungraded track, so that the CSS doesn't need an excessive amount of cases and conditions and can specify it in one block.

There are a bunch of other variants and combinations that might be of interest:

other combos

There are technically a few dozen different combinations of tags. But I didn't even attempt to come up with unique variants for these, because most of them are either wrong or ambiguous and don't have an obvious expected rendering. Basically, the above examples cover most valid common cases that I can think of and the rest are incorrect or incomplete tagging.

The only part I'm not so sure about is `embankment` rendering:

embanked grading

These look weird to me at first glance. But then again, embankments look weird on other roads too:

other embanked

There isn't an exact style for embankment color and pattern that I can see - they seem to be based directly on the casing pattern. I believe this is a rendering limitation? In any case, I don't know that there is anything I can do about it. I'm only concerned here because tracks are commonly embanked, so this might be worth considering further.

No LLM code.


RudyTheDev:
Here's an example location with a bunch of differently-graded tracks: https://pr-12392--ideditor.netlify.app/#map=16.00/53.1318/13.0282

tyrasd:
This looks good, thanks.

Maybe `tracktype=grade1` should get the thin outline in the same color as the other tracktypes, to make it consistent across all grades?

For the unknown state, I found the pattern looks a little bit too much "chopped". I found that `stroke-dasharray: 7, 3, 3, 3, 3, 3;` looks a little bit better while still conveying the meaning of _“could be any grade”_.

> embankment rendering

Yeah, that rendering really not ideal (in general). See #12396 for a possible improvement on that, which should also benefit this PR. Let me know what you think.

tyrasd:
> There are a bunch of other variants and combinations that might be of interest

Yeah… there's also the case of e.g. `surface=paved + tracktype=grade2` (which could make sense maybe, e.g. really poor condition concrete surface or so). Should we not just simplify this even further, and for `highway=track` only look at the `tracktype` for rendering? That might also fix #10780 if I see it correctly.

RudyTheDev:
> Maybe `tracktype=grade1` should get the thin outline in the same color as the other tracktypes, to make it consistent across all grades?

Hmm, perhaps? I left it alone because while it would be consistent between track grades, it would be inconsistent with other paved roads. I guess here's some comparisons:

black vs white

If it's okay to change, then I agree being cased white is better to make it more consistent.

> Yeah… there's also the case of e.g. `surface=paved + tracktype=grade2`

Yeah, that's the sort of semi-ambiguous stuff I mean. Like, in theory we could also pick color based on surface but dash pattern based on grade. I tried it briefly, but you really have to "push" dash patterns and casing color/thickness to make it visible enough:

black paved patterns

And I'm not sure this is the way to go.

> Should we not just simplify this even further, and for `highway=track` only look at the `tracktype` for rendering?

If it's fine to have grade1 ignore generic surface styling, then I suppose we might as well have tracks only styled by grading.

> For the unknown state, I found the pattern looks a little bit too much "chopped". I found that `stroke-dasharray: 7, 3, 3, 3, 3, 3;` looks a little bit better while still conveying the meaning of _“could be any grade”_.

Yeah, that looks good too. My intention here was to make it really obvious that it's missing a grade, so being "choppy" was intentional. May be a little less "compressed" because it kind of blurs together at a quick glance?

choppy track patterns

> Yeah, that rendering really not ideal (in general). See #12396 for a possible improvement on that, which should also benefit this PR. Let me know what you think.

Yeah, that would be ideal if embankments were a separate "layer". I'll comment there. I won't touch anything embankment-related for this PR then.

RudyTheDev:
Updated preview for white casing variant:

white casing

In-code tracks are now handled separately and do not get surface-based classes, but only work off grade-based styling.

Speaking of, here are a couple example of slightly lighter casing:

casing color

I am thinking perhaps `#DDD` might be a bit better for readability? It didn't matter much before, but background aerial of vegetation can really mess with visibility if one wants to distinguish the grade quickly. (I suppose tracks themselves could be made darker, but I think the main problem here is aerial itself and not the contrast within the track.)

tyrasd:
thanks for the update.

> casing color

agree, #DDD looks like the best for now. 

> in theory we could also pick color based on surface but dash pattern based on grade. […] And I'm not sure this is the way to go.

That seems like trying to display too much information at once. I _think_ tracktype is the most relevant for tracks, so let's limit it to that for the rendering.

> inconsistent with other paved roads.

true, I did not think about that at first. Yes, there's also merit to show a grade1 track the similarly to a (paved) `highway=service` or `highway=unclassified`. I guess by changing the casing for all tracktypes to the bright color we did make _that_ inconsistent. :see_no_evil: //cc @k-yle , @tordans : what's your opinion regarding this?

RudyTheDev:
Here is a comparison for reference:

current

tyrasd:
I changed the styling of regular roads to always use a dark casing, and distinguish paved from unpaved only via the dashing. The thinking would be: dark casing (e.g. `highway=service`, `unclassified`, etc.) is for general purpose traffic roads, while light casings are for ways in the fields, nature, and similar contexts (`highway=track`, but applies also to `path`, etc.).

The styling of bridges and tunnels are also adjusted (e.g. dashing applied to the stroke instead of casing on bridges) to give them a more streamlined appearance. Screenshot has from the top to bottom: cutting, tunnel, normal, bridge, embankment; and from left to right: paths, unclassified paved/unpaved, tracks 1-5 and unspecified tracktype.

image

one last note: it is very hard to see which tracktype a particular track has when it is on a very light surface (e.g. compacted, grade2 in the screenshot below):



That's actually not a new problem, and the only simple way to approach it would be to use a slightly darker casing color (maybe sand brown or light gray), but any of those would be at the cost of reduced contrast in other situations (e.g. actual sand, or slightly darker gravel). :shrug:

RudyTheDev:
> dark casing .. is for general purpose traffic roads, while light casings are for ways in the fields, nature, and similar contexts

That looks good, I think that is a more consistent approach overall.

> very hard to see which tracktype a particular track has when it is on a very light surface

Yeah, I also don't think there's any good solution here. As you say - nothing new - whatever colors one picks, there is always some scenario where they "disappear".

> The styling of bridges and tunnels are also adjusted (e.g. dashing applied to the stroke instead of casing on bridges)

I suppose that's an interesting idea, I didn't even think of doing something like that to make bridges/tunnels more consistent casing-wise as an uninterrupted "structure". But the reason I didn't mess with stroke is because that's part of the life-cycle stuff. So if stroke is changed, then one has to consider something like when `abandoned=yes` is set as it has "interesting" cases:

image

I don't know how much this matters for tracks, there aren't that many examples. But any casing/stroke patterns do normally get lost once abandoned is applied. Except for bridges/tunnels now. I guess the same issue is then also true for other paved/unpaved roads:

image

tyrasd:
> `abandoned=yes`

oh no, I didn't think about that additional combination. :see_no_evil: 

I would argue that the whole lifecycle rendering should be reworked in general, as:
* it currently only works on discouraged tags like `abandoned=yes` or `highway=abandoned`, 
* proper lifecycle prefixed tagging (`abandoned:highway=…`) is only using the fallback styles
* maybe should be better styled completely differently (e.g. much more de-emphasized compared to the current renderings).

Luckily, I believe for tracks it is actually not very common to use lifecycle tagging. And I would argue that this PR does not significantly worsen the styling in that regard. So, let's put that aside for now.

RudyTheDev:
> I would argue that the whole lifecycle rendering should be reworked in general

Yeah, I agree life-cycle stuff really needs its own styling _on top_ of the existing styles long-term. With may be some exceptions like road construction. Though I suppose we need to wait for something like #10844 for that.

PR #12396, opened 2026-05-28
render embankments and cuttings with triangle markers

tyrasd:
Using the same markers as cliffs, coastlines, retaining walls, etc. which intuitively indicate the "downward/outside" direction of the respective features. This replaces using wide casings to indicate embankments and cuttings, which are arguably unintuitive for this property and also applied somewhat inconsistently as they can be influenced by casing properties coming from other tags (e.g. dashing of unpaved roads)

In addition, this also supports single-sided embankments/cuttings, which are used to map cases where e.g. a road mapped as a dual carriageway that share a common embankment.

### DIFF
--- a/css/50_misc.css
+++ b/css/50_misc.css
@@ -360,30 +360,6 @@ path.line.casing.tag-location-underwater {
 }
 
 
-/* embankments / cuttings */
-path.line.shadow.tag-embankment,
-path.line.shadow.tag-cutting {
-    stroke-width: 28;
-}
-path.line.casing.tag-embankment,
-path.line.casing.tag-cutting {
-    stroke-opacity: 0.5;
-    stroke: #000;
-    stroke-width: 22;
-    stroke-dasharray: 2, 4;
-    stroke-linecap: butt;
-}
-
-.low-zoom path.line.shadow.tag-embankment,
-.low-zoom path.line.shadow.tag-cutting {
-    stroke-width: 14;
-}
-.low-zoom path.line.casing.tag-embankment,
-.low-zoom path.line.casing.tag-cutting {
-    stroke-width: 10;
-}
-
-
 /* Surface - unpaved */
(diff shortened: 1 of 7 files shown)


Dimitar5555:
Will it work well with `highway=cycling_waiting_aid + side=*`?

tyrasd:
I think it is acceptable:

RudyTheDev:
I definitely support this change. Current rendering of embankments/cuttings based on the casing/edge is very subpar. It shouldn't be changing with road type, surface or dash pattern like now. Plus, having side/direction for markers is a big bonus.

I think on low zooms the arrows should be closer to the way.

My concern is that these blend into the background if they are not on some light aerial like asphalted highways in the example. Many embankments are paths and tracks and railways and those are rarely in places with light background. So examples like this:

embank as line old
embank on the triangle

I guess because it was casing-based before, it incidentally used contrasting colors from the road styles. But now it's the same style for everything. I am not sure what the best solution here would be. May be major roads = black, other ways = white? Or may be have a border (it that even possible)?

I guess most extreme example would be a transition like this:

track from light to dark

As a comparison, currently separate embankments are extremely white, so they are never invisible:

extra white embankment

tyrasd:
> blend into the background

true, on darker forests it really is hardly visible. My intention with the darkish gray was to make them not stand out too much, but they should not be completely blending in with the background also.

>  separate embankments

true, we probably should adjust those to make the colors match with the embankment styling on roads.

<hr>

I tried something with a very thin outline around the triangles to give them some contrast. Here are some relatively tricky cases:

matkoniecz:
maybe use two colours? With light fill and dark border or opposite?

tyrasd:
The last screenshot does already use two tones of gray (see the triangles in the shadowy part of the road), to provide a bit more contrast. But maybe inverted is actually better:

PR #12397, opened 2026-05-28
delete `field.overrideLabel`

k-yle:
it seems to be dead code. there are no other search results in [org:openstreetmap](https://github.com/search?q=org%3Aopenstreetmap%20overrideLabel&type=code), [org:osmlab](https://github.com/search?q=org%3Aosmlab%20overrideLabel&type=code), [org:ideditor](https://github.com/search?q=org%3Aideditor%20overrideLabel&type=code), [org:rapideditor](https://github.com/search?q=org%3Arapideditor%20overrideLabel&type=code). so i think this can be safely deleted

### DIFF
--- a/modules/presets/field.js
+++ b/modules/presets/field.js
@@ -39,10 +39,8 @@ export function presetField(fieldID, field, allFields) {
     return _this;
   };
 
-  _this.title = () => _this.overrideLabel || _this.resolveReference('label').t('label', { 'default': fieldID });
-  _this.label = () => _this.overrideLabel ?
-      selection => selection.text(_this.overrideLabel) :
-      _this.resolveReference('label').t.append('label', { 'default': fieldID });
+  _this.title = () => _this.resolveReference('label').t('label', { 'default': fieldID });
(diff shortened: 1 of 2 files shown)


tyrasd:
this was originally added with #5617, the [maprules specification](https://github.com/radiant-maxar/maprules/blob/master/maprules.spec.md) does not mention it (anymore?), but there are some [examples](https://github.com/radiant-maxar/maprules/blob/develop/maprules.apidocs.md#examples-5) which do include it. See also #12317 

Shall we discuss what to do with the whole maprules generally in the upcoming community chat (either fixing it or removing it completely) and leave this code as is until then?

tyrasd:
we talked about this last week in the [community chat](https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_10th_June_10:00_%E2%80%93_11:00_(UTC)), the decision was to drop the defunct maprules implementation for now, and talk about potentially reintroduce (parts?) of it again in a follow-up issue if the community thinks them to be useful.

tyrasd (approved):
as mentioned: let's drop this now

PR #12401, opened 2026-05-31
Fix tests for node 26

### LINKED ISSUE #12047: tests failing on nodejs v25
On my machine, the unit tests are failing randomly when executed using node version 25.

I see a lot of of `localStorage quota exceeded` warnings in the console, including an error from node that supposedly `--localstorage-file` is pointing to an invalid file. 🤷 I don't see where we set this ourselves, so it might be something a library is doing and/or a new/changed setting of node itself. 

```
 FAIL  test/spec/services/osm.js > iD.serviceOsm > API capabilities > #imageryBlocklists > updates imagery blocklists
TypeError: this.userDetails is not a function
 ❯ Object.userChangesets modules/services/osm.js:758:14

[…]

 FAIL  test/spec/ui/sections/background_list.js > iD.uiSectionBackgroundList > customChanged handles missing custom source gracefully
AssertionError: expected +0 to equal 1

[…]
```

//edit: see https://github.com/openstreetmap/iD/actions/runs/23302592596/job/67767600914#step:7:86 for a failing test run

brianstrauch:
Closes #12047.

When running `brew install node` (defaults to v26) followed by `npm test` I get quite a few failures.
* Added a localstorage file to the test configuration to get around the lack of local storage in node 26.
* It looks like IndexedDB is used in place of localstorage in production, which was introduced by #11141.
* Bonus: While debugging, I noticed that the in-memory storage fallback (`let s = {}`) is flawed since all of our tests expect `null`, not `undefined` if an item doesn't exist in storage.


modules/core/preferences.js
```diff
@@ -7,7 +7,7 @@ try { _storage = localStorage; } catch {}  // eslint-disable-line no-empty
 _storage = _storage || (() => {
   let s = {};
   return {
-    getItem: (k) => s[k],
+    getItem: (k) => s[k] || null,
```
brianstrauch:
```suggestion
    getItem: (k) => s[k] === undefined ? null : s[k],
```

vitest.config.ts
```diff
@@ -1,9 +1,15 @@
...
+import path from 'node:path';
 import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
   test: {
     css: true,
     environment: 'jsdom',
+    execArgv: [
+        '--localstorage-file',
+        path.resolve(os.tmpdir(), `vitest-${process.pid}.localstorage`),
```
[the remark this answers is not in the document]
brianstrauch: This should be fine, this is the purpose of `$TMPDIR`

modules/core/preferences.js
```diff
@@ -7,7 +7,7 @@ try { _storage = localStorage; } catch {}  // eslint-disable-line no-empty
 _storage = _storage || (() => {
   let s = {};
   return {
-    getItem: (k) => s[k],
+    getItem: (k) => s[k] === undefined ? null : s[k],
```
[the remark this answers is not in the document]
brianstrauch: Better yet, looks like Javascript has an `in` operator: https://stackoverflow.com/questions/1098040/checking-if-a-key-exists-in-a-javascript-object. Updating...

modules/core/preferences.js
k-yle, line 14: since iD has already implemented its own localStorage polyfill, we don't actually need to use node's built-in one.

So an alternative solution would be this one-line-fix:
```suggestion
})();
localStorage ||= _storage;
```

(I'm not saying this is a better solution, I just noticed that the code already exists so we could just use it)

compared to the vitest config solution, this would also prevent hundreds of temporary files building up every time we run the unit tests

brianstrauch: Apologies since I'm new to Javascript, but are you considering lines 7-14 as the polyfill? If so yes, I agree that we only need one fallback. I'll make the change but let me know if you meant something else!

k-yle: yes exactly

k-yle (approved):
thanks, lgtm. final question is whether the CI should run tests on node26 too

tyrasd (commented):
🤔 

From what I can see, the crashes are actually coming from osm-auth which does not properly fall back to a polyfill when the localStorage object is `undefined`. See https://github.com/osmlab/osm-auth/pull/152 for a proper fix for that.

Until then we could do the workaround of hacking[^1] our polyfill into the `globalThis` object, or alternatively we could just disable the experimental webstorage support in `vitest.config.ts`:

[^1]: which could be slightly dangerous as we only implement the subset of the API we need, but others might actually call, say `localStorage.length` potentially.

```
    execArgv: [
        '--no-experimental-webstorage'
    ],
```

Which would be slightly nicer, as then we'd also not get the false positive warnings about missing `--local-storage-file` parameter in the logs.

What do you think?

PR #12406, opened 2026-06-02
Fix multi-combo field remove and undo buttons not working

### LINKED ISSUE #12590: Can't delete payments and currencies with the trash bin icon
When I add any payments, then I can't delete it. Clicking trash bin icon doesn't work. The same with currencies.

RudyTheDev:
Closes #12590

This PR fixes a bug with `multiCombo` preset fields for sub-keyed values, such as "Fuel Types" (e.g. `fuel:diesel`), where the "remove" and "undo" buttons would not work/show. This currently affects 19 fields like `fuel_multi`, `payment_multi`, `diet_multi` etc.

**Currently**:

before base

Pressing the trash can does nothing:

before delete

Making any changes does not show an undo arrow:

before add

before manual remove

**Fix**:

after base

Pressing trash can deletes all the sub-tags and shows undo arrow:

after remove

Modifying key value shows undo arrow:

after modify

Undo arrow is shown when making individual additions (i.e. new key not in original):

after add one

Undo arrow is shown when making individual deletions (i.e. key exists only in original):

after remove one

For all the above, undo now properly restores not just current keys, but also the original keys.

Seems to work for multi-selection fine as well including mixing new/existing elements.

In technical details: this fixes a primary bug in `uiField` where `allKeys()` wrapper was not passing `tags` to `presetField.allKeys`, thus not accounting for "dynamic" preset keys. Specifically, in `uiFieldCombo` in `.tags` where it did tag collection in `field.keys =  ...` for `multiCombo` variant, this value was never used. The consequence is that tags like `fuel:diesel` would not be considered by `uiField` functions, namely `remove`, `revert`, `isModified` and `tagsContainFieldKey`. Furthermore,  this fixes a secondary bug that `isModified` and `revert` did not consider that the original entity may have had keys matching the field that no longer exist. In other words, `allKeys(_tags)` (even with tags given) would not consider the `allKeys(original.tags)`. Since any selected entity triggers either modified flag or adds the keys to revert list, then multi-select also works.

No LLM code, only some auto-complete.


matkoniecz:
> npm error 404 Not Found - GET https://registry.npmjs.org/-/npm/v1/attestations/marked@18.0.4 - Not found

maybe close/reopen of PR to try retriggering CI build?

RudyTheDev:
> maybe close/reopen of PR to try retriggering CI build?

Ah, I didn't know I could do that, thanks for the tip. I assumed I had to push another commit or have a repo member rerun the workflow.

matkoniecz:
> have a repo member rerun the workflow.

actually they also would need to close/reopen or push commit (maybe push trash commit and force push previous state)

RudyTheDev:
I added tests for checking the modified state and undo and remove operations for the various ways the relevant keys can be modified.

Currently, they pretty much all fail:

tests before

With fixes applied:

tests after

I used LLM to suggest initial test syntax, because there were no exact existing examples and I'm not quite sure what the best way for arranging everything is to make least assumptions. In particular, I am not quite sure what the best way to simulate clicking undo/remove is, because there seems to be some editor wiring that is not really present in the tests. Asking LLM to do this just sends it off the rails adding tons of code. There basically isn't a field UI undo test example or test helper code that I could use. So I'm essentially mocking `entity_editor` functions which only really check that the field provides the right keys for reverting. Which I suppose is enough for the immediate logic and underlying issue, but it feels a bit too hard-coded. That said, a lot of tests do these kinds of things, so I dunno.

tyrasd (approved):

tyrasd:
>  I am not quite sure what the best way to simulate clicking undo/remove is. […] So I'm essentially mocking entity_editor functions which only really check that the field provides the right keys for reverting.

Yeah, for code that relies on the correct interaction between different components (here: the field and the entity editor), we would need proper UI integration tests, which we don't have at the moment. Anything else is possibly too much effort to simulate in a unit test environment and probably also not very maintainable.

I think your solution is fine, and should be able to test the most important aspects at play. Thanks

Issue #11039, opened 2025-05-09
Support Extension Access to coreContext in Production


julcnx:
### Description

Browser extensions that enhance the iD editor—such as the Strava Heatmap extension—rely on access to `coreContext` in order to interact with native iD functions and provide deeper editor integration.

Currently, `coreContext` is only exposed in [development](https://github.com/openstreetmap/iD/blob/3ba82cb473d83dc1045f64735ecec5b5bbbcae17/index.html#L47), but not in production. As a result, extensions like Strava Heatmap must resort to fragile workarounds, hooking into the page load process and the `window.iD` property to capture the `coreContext` reference.

Exposing `coreContext` in production would simplify extension development by avoiding these overrides, and would also allow power users and testers to experiment with the API directly from the console.

k-yle:
Coming from #1392 - I realise this issue is focused on browser extensions, but if you just want access to `coreContext` for spontaneous debugging, [`queryObjects(Rapid.Context)`](https://developer.chrome.com/docs/devtools/console/utilities#queryObjects-function) works in RapiD, but not iD (since we [don't use ES classes](https://github.com/openstreetmap/iD/issues/10909))

k-yle:
coming from https://github.com/openstreetmap/iD/pull/12411#discussion_r3387891560 - that PR would expose `window.iD` in production, solving this issue. Is this a good idea or should we implement it differently? 

Alternative proposal: iD's coreContext already has a 1:1 relation with a [`containerNode`](https://github.com/openstreetmap/iD/blob/2544603aceccf6446ecaef4f19ea7861d2b2a8c9/modules/core/context.js#L457). JS allows you to attach arbitrary objects to DOM elements. So we could attach a property called `.iD` to the root node. It's a one-line change on iD' side:

```diff
    context.containerNode = function(val) {
      if (!arguments.length) return context.container().node();
      context.container(d3_select(val));
+     val.iD = context;
      return context;
    };
```

anyone who wants to hack around with iD can access the context via `element.iD`, if you have permission to get the `element` node somehow. that could be as simple as `document.querySelector('....').iD`

---

cc'ing @deevroman from [better-osm-org](https://github.com/deevroman/better-osm-org) and @julcnx from [strava-heatmap-extension](https://github.com/julcnx/strava-heatmap-extension). not sure if any other third-parties are using iD internals?

tyrasd:
[...]

PR #12411, opened 2026-06-06
Migrate to Vite 8

tordans:
As discussed no Slack, this is a take on migrating to Vite 8.

I tested the app a bit an all looks nice. Changing light<>dark mode on a system level still works.

- Replace esbuild + standalone CSS concat with Vite (`vite.config.ts`, `main.ts` entry)
  - Dev: `npm start` runs `build:data` then Vite HMR (no separate JS watch/server)
  - Prod: `dist/iD.js` / `dist/iD.min.js` / `dist/iD.css` with stable names for embedders
- Move env config from esbuild `define` to `import.meta.env` (`config/env.ts`, `config/id.ts`)
- Keep PostCSS plugin for `.ideditor` scoping + dark-mode duplication; Lightning CSS for autoprefix
- Hand-maintain `dist/index.html` (minimal diff: inline loader → `<script type="module">`)
- Merge Vitest into `vite.config.ts`; broaden `tsconfig` to `**/*.ts`
- PR preview deploys from `dist/`; add `netlify.toml`
- Bump Vite ecosystem deps to latest (vite, vitest, postcss, browserslist, rollup-plugin-visualizer)
- Suppress Vite warning for test data imports
- Simplify build and dev scripts

## Follow-up migrations (align with Vite 8 best practices)

- [ ] Adopt [Vite's `public/` directory](https://vite.dev/guide/assets#the-public-directory) for static assets (`img/`, `data/`) so `url(img/...)` resolves natively, removing the `img → dist/img` symlink and `/dist` resolve alias. This will make things like https://github.com/openstreetmap/iD/pull/12411/changes#r3368185484 a lot easier
- [ ] Drop the custom `scripts/build_css.ts` PostCSS step by moving `.ideditor` scoping + dark-mode handling into source CSS (or native Lightning CSS features). (ref [discussion r3368169678](https://github.com/openstreetmap/iD/pull/12411/changes#r3368169678))
- [ ] Replace the `dist:mapillary` / `dist:pannellum` copy scripts with native Vite dynamic imports / code-splitting; address the ~1.9 MB chunk-size warning via `manualChunks`.
- [ ] Reassess `browserslist-to-esbuild` for `build.target`: Vite 8 uses Rolldown/Oxc, so confirm this is the right adapter or move to Vite's native browser-target syntax once Rolldown reads browserslist directly. – Upstream issue https://github.com/rolldown/rolldown/issues/9152


config/env.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Client env values from import.meta.env (Vite envPrefix: 'ID_').
+ * Used by config/id.ts for CDN URLs, API connection, and feature flags.
+ */
+export const idPresetsCdnUrl: string | null = import.meta.env.ID_PRESETS_CDN_URL ?? null;
```
tordans: Note: We override the Vite default `envPrefix` for public env.* to `ID_*` which is why this works.

config/envs.js
tordans, line 1: Same idea, now in `config/env.ts`. We could preserve the names `ENV_*` but given that `config/id.ts` reassigns many of them anyways, this looks nicer to me.

config/id.ts
tordans, line 1: Nearly each line was touched, so I migrated this to TS as a byproduct.

css/entry.css
tordans, line 1: Vite and [lightning](https://vite.dev/guide/features#lightning-css) handles the CSS now.

css/entry.css
```diff
@@ -0,0 +1,18 @@
+/**
+ * Single CSS entry for Vite. Order matches previous build_css.js (sorted by filename).
+ */
```
tordans:
```suggestion
```

We don't need this.

scripts/build_css.ts
tordans, line 1: This is one of the more interesting changes…
Here is some explanation summary from Cursor:

> With Vite, `scripts/build_css.ts` is not a separate build step anymore — it is a **Vite plugin** (`idCssPlugin`) that applies two transforms iD still depends on, which Vite/Lightning CSS do not provide on their own.
> 
> ## What it does
> 
> When Vite processes any file in the `css/` tree (via `css/entry.css` → `@import` chain in `main.ts`), the plugin runs PostCSS with two plugins:
> 
> ### 1. Scope all styles under `.ideditor`
> 
> Most source CSS is written with bare selectors:
> 
> ```3:12:css/50_misc.css
> path.line.shadow.tag-aerialway,
> path.line.shadow.tag-attraction-summer_toboggan,
> ...
> ```
> 
> At runtime, iD mounts into a container with class `ideditor` (set in `context.js`). The plugin uses `postcss-prefix-selector` to turn `.foo` into `.ideditor .foo` so styles do not leak onto the host page when iD is embedded. Selectors that already start with `.ideditor` are skipped via the exclude regex.
> 
> ### 2. Make OS dark mode work with explicit theme choice
> 
> Authors write dark styles the natural way:
> 
> ```68:70:css/10_colors.css
> @media (prefers-color-scheme: dark) {
>     .ideditor {
> ```
> 
> iD also lets users pick **light**, **dark**, or **system** in settings (`theme-light` / `theme-dark` classes on the container). The custom `duplicate-from-media` plugin rewrites those `@media (prefers-color-scheme: dark)` blocks:
> 
> - Original rules get `:not(.theme-light)` — explicit light theme ignores OS dark preference
> - Duplicated rules get `.theme-dark` — explicit dark theme applies dark styles even on a light OS
> 
> Without this, “use dark theme” in iD settings would not work reliably when the OS preference disagrees.
> 
> ## Why it exists as its own file
> 
> | Concern | Handled by |
> |---|---|
> | Bundle JS, HMR, output `iD.css` | Vite |
> | Autoprefixer / browser targets | Lightning CSS (`css.transformer: 'lightningcss'`) |
> | `.ideditor` prefix + theme duplication | `scripts/build_css.ts` |
> 
> The pre-Vite setup had a standalone `scripts/build_css.js` that did the same PostCSS work. After the Vite migration, that logic became a plugin so it runs **in dev and production** on the same CSS pipeline — no duplicate “build CSS then build JS” step.
> 
> ## What it deliberately does *not* do
> 
> The file header says it only handles prefix + dark-mode. Autoprefixer was removed from PostCSS here because Lightning CSS already covers that via `browserslist` in `vite.config.ts`.
> 
> ---
> 
> **In short:** iD’s CSS is authored as if it were page-global (short selectors + `@media (prefers-color-scheme: dark)`). `build_css.ts` adapts that legacy style to iD’s embedding model and theme API. Until someone rewrites all `css/*.css` with explicit `.ideditor` prefixes and `.theme-dark` rules by hand, this plugin is required.
> 
> A follow-up from the migration plan was to eventually drop the PostCSS dark-mode duplication once CSS is simplified — that would be a large CSS refactor, not a Vite config tweak.

package.json
```diff
@@ -17,17 +17,15 @@
...
-    "build:css": "node scripts/build_css.js",
+    "build": "run-s build:data build:app",
+    "build:app": "vite build",
     "build:data": "shx mkdir -p dist/data && node scripts/build_data.js",
-    "build:stats": "node config/esbuild.config.min.js --stats && esbuild-visualizer --metadata dist/esbuild.json --filename docs/statistics.html && shx rm dist/esbuild.json",
-    "build:js": "node config/esbuild.config.js",
-    "build:js:watch": "node config/esbuild.config.js --watch",
-    "clean": "shx rm -f dist/esbuild.json dist/*.js dist/*.map dist/*.css dist/img/*.svg",
+    "build:stats": "BUILD_STATS=true vite build",
+    "stats:view": "npm run build:stats && npx -y serve docs -p 4174",
```
tordans: We could remove this. But how do you view those results? Manually?

package.json
```diff
@@ -17,17 +17,15 @@
...
   "scripts": {
     "all": "run-s clean build dist",
-    "build": "run-s build:css build:data build:js",
-    "build:css": "node scripts/build_css.js",
+    "build": "run-s build:data build:app",
+    "build:app": "vite build",
     "build:data": "shx mkdir -p dist/data && node scripts/build_data.js",
-    "build:stats": "node config/esbuild.config.min.js --stats && esbuild-visualizer --metadata dist/esbuild.json --filename docs/statistics.html && shx rm dist/esbuild.json",
-    "build:js": "node config/esbuild.config.js",
-    "build:js:watch": "node config/esbuild.config.js --watch",
```
tordans, line 25: This is simplified. Vite handles the minification (and source map) and JS build.

package.json
```diff
@@ -39,11 +37,8 @@
     "imagery": "node scripts/update_imagery.js",
     "lint": "eslint config scripts test/spec modules",
     "lint:fix": "eslint scripts test/spec modules --fix",
-    "start": "run-s start:watch",
-    "start:single-build": "run-p build:js start:server",
-    "start:watch": "run-p build:js:watch start:server",
-    "start:server": "node scripts/server.js",
-    "test": "npm-run-all -s lint build test:typecheck test:spec",
+    "start": "npm run build:data && vite",
+    "test": "run-s lint build test:typecheck test:spec",
```
tordans, line 41: This is simplified as well. Vite does a lot of the reload and watch internally.

tsconfig.json
```diff
@@ -12,5 +12,5 @@
         "resolveJsonModule": true,
         "types": ["vitest/globals", "vitest/jsdom"]
     },
-    "include": ["modules", "test"]
+    "include": ["**/*.ts"]
```
tordans: We can simplify this.

vite.config.ts
```diff
@@ -0,0 +1,64 @@
...
+import { visualizer } from 'rollup-plugin-visualizer';
+import { idCssPlugin } from './scripts/build_css';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const isBuildStats = process.env.BUILD_STATS === 'true';
+
+export default defineConfig(() => ({
+  base: './',
+  envPrefix: 'ID_',
```
tordans, line 18: See `config/env.ts`.

vitest.config.ts
tordans, line 1: Now inline in `vite.config.ts` as `test` prop.

scripts/build_css.ts
```diff
@@ -0,0 +1,59 @@
...
+    name: 'id-css',
+    enforce: 'post',
+    transform(css, id) {
+      if (!id.endsWith('.css')) return null;
+      if (id.includes('html-proxy') || id.includes('inline-css')) return null;
+
+      const cssStr = typeof css === 'string' ? css : String(css ?? '');
+      // Skip when Vite has already wrapped CSS in a JS module (dev HMR)
+      if (cssStr.startsWith('import ') && cssStr.includes('__vite__')) return null;
+
```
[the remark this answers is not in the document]
tordans: We have a special setup here that already makes `url(img/...)` work in dev and production — no CSS rewrite needed.

**Step by step:**

1. **`npm start`** runs `npm run build:data && vite` (see `package.json`).

2. **`build:data`** runs `node scripts/build_data.js`, which creates a symlink at the repo root:
   ```js
   img → dist/img
   ```
   (see `scripts/build_data.js` lines 47–54)

3. **CSS** uses paths like `url(img/cursor/grab.png)` — written for the deployed layout where `iD.css` and `img/` sit side by side at the site root (`foo.com/iD.css`, `foo.com/img/...`), not under `/dist/`.

4. **In dev**, Vite serves the app from `/` (`index.html` at project root). Those `url(img/...)` paths resolve to `/img/...` on the dev server. Because of the symlink, `/img/...` maps to `dist/img/...` on disk, so cursors/loaders load correctly.

5. **In production**, Netlify deploys the *contents* of `dist/` as the site root (`--dir=dist`), so `dist/img/` becomes `foo.com/img/` and `dist/iD.css` becomes `foo.com/iD.css` — same relative layout the CSS expects.

So we don't need to change anything here. The `img → dist/img` symlink is the dev bridge; production already matches the CSS paths after deploy.

vite-env.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  readonly ID_SHOW_DONATION_MESSAGE: string | undefined;
+}
+
+interface ImportMeta {
+  readonly env: ImportMetaEnv;
+}
+
+declare global {
+  interface Window {
+    context?: iD.Context;
```
k-yle: if this is whitelisted as a global, i would mark this as `@deprecated` or something, so that we don't accidentally use it.

```suggestion
    /** @deprecated */
    context?: iD.Context;
```

or alternatively, mark this as a banned global variable in eslint.config.js (see #12210)

types/ambient.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'postcss-prefix-selector';
```
k-yle: it would be better to import [@types/postcss-prefix-selector](https://npm.im/@types/postcss-prefix-selector) rather than silencing the warning

main.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Entry point for the iD editor (Vite dev and build).
+ * In dev, Vite serves this and transforms on the fly; in build, this is bundled to dist/iD.min.js.
+ */
+import './css/entry.css';
```
k-yle: I'm guessing this import magic will be converted into a `<link rel="stylesheet" />` tag?

If so, then so then this only applies to the standalone build? because the OSM website will load `iD.css` and `iD.js` directly, right?

config/env.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Client env values from import.meta.env (Vite envPrefix: 'ID_').
+ * Used by config/id.ts for CDN URLs, API connection, and feature flags.
```
k-yle: can we just delete this file? with a modern bundler we don't need a file that reëxports everything. 

and `import.meta.env.*` is type-safe already

package.json
```diff
@@ -17,17 +17,15 @@
...
-    "build": "run-s build:css build:data build:js",
-    "build:css": "node scripts/build_css.js",
+    "build": "run-s build:data build:app",
+    "build:app": "vite build",
     "build:data": "shx mkdir -p dist/data && node scripts/build_data.js",
-    "build:stats": "node config/esbuild.config.min.js --stats && esbuild-visualizer --metadata dist/esbuild.json --filename docs/statistics.html && shx rm dist/esbuild.json",
-    "build:js": "node config/esbuild.config.js",
-    "build:js:watch": "node config/esbuild.config.js --watch",
-    "clean": "shx rm -f dist/esbuild.json dist/*.js dist/*.map dist/*.css dist/img/*.svg",
+    "build:stats": "BUILD_STATS=true vite build",
```
k-yle: this won't work on windows, can we use `cross-env` (or whatever the modern replacement is - `shx` and `shelljs` are _both_ already installed)

or can we use a CLI argument instead?

tsconfig.json
k-yle: we need to include JS files too (since definitions can be imported between both files).

could we just keep this line as-is ?

vite.config.ts
```diff
@@ -0,0 +1,64 @@
...
+  },
+  resolve: {
+    alias: {
+      '/dist': path.resolve(__dirname, 'dist'),
+    },
+  },
+  css: {
+    transformer: 'lightningcss',
+    lightningcss: {
+      targets: browserslistToTargets(browserslist()),
```
k-yle: the only place where browserlist is used is here, for CSS. 

do we need to link the browserslist to the JS somehow? vite uses [`build.target`](https://vite.dev/config/build-options#build-target) instead.

or - should we delete the browserlist config and rewrite it using vite's syntax? that might be easier

index.html
```diff
@@ -28,37 +27,6 @@
...
-      document.getElementsByTagName('head')[0].appendChild(newScript);
-
-
-      function checkScript() {
-        if (typeof iD === 'undefined' || !iD.utilDetect().support) {
-          container.innerHTML = 'Sorry, your browser is not currently supported. Please use another <a href="https://github.com/openstreetmap/iD#basics">browser</a> or <a href="https://wiki.openstreetmap.org/wiki/Editors">editor</a> to contribute to the map.';
-          container.style.padding = '20px';
-
-        } else {
-          var context = iD.coreContext()
```
k-yle, line 49: all this code is [copy-pasted (with modifications) in the osm-website repo](https://github.com/hlfan/openstreetmap-website/blob/c9e551faf81f094d315059010998c4138709e92f/app/assets/javascripts/id.js#L14). 

so any changes needs to be coordinated with them. 

we could solve #11039 at the same time by exposing `context` differently - I've commented at https://github.com/openstreetmap/iD/issues/11039#issuecomment-4669703906

main.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  const assetBase = import.meta.env.DEV ? 'dist/' : '';
+  const context = iD.coreContext()
+    .assetPath(assetBase)
+    .containerNode(container);
+  if (!context) return;
+
+  window.context = context;
+  window.id = context;
```
k-yle: previously, this was only available in debug mode. now it's available in production, so people might start using it.

I think we need to find a conclusion to #11039 first, before people start relying on this hack

k-yle (commented):
this is just a very quick review, following the meeting today. i haven't tried it out yet

css/entry.css
tordans: Site note: We need to list each css file manually now. Before we had this auto loading. I think that is fine because for me that is what I would expect anyways.

tsconfig.json
tordans: Done in 63f5d9ae2 — restored modules + test (so allowJs still covers JS cross-imports) and added **/*.ts for root/config/scripts TS files. Also added exclude for node_modules and dist.

vite-env.d.ts
tordans: Done in 4e951ca07 — added `@deprecated `JSDoc on window.context / window.id in vite-env.d.ts and added context to no-restricted-globals in eslint.config.js.

types/ambient.d.ts
tordans: Done in 651de0444 — added @types/postcss-prefix-selector and removed types/ambient.d.ts.

main.ts
tordans: Yes — Vite extracts CSS to dist/iD.css (cssCodeSplit: false). Embedders load iD.css and iD.min.js separately; the CSS import is a build-pipeline entry, not a runtime dependency for embedders.

After af34007e4 the CSS import lives in lib.ts (library entry), not main.ts. main.ts / bootstrap.js is bootstrap-only and does not import CSS.

config/env.ts
tordans: Done in cd8db8d4e — removed config/env.ts and read import.meta.env.ID_* directly in config/id.ts. Typings stay in vite-env.d.ts (ImportMetaEnv).

package.json
tordans: Documented in f72c7ea8c — npm run stats:view now builds with `--mode stats` and serves docs/statistics.html at http://127.0.0.1:4174. Also hinted in CONTRIBUTING.md.

vite.config.ts
tordans: Thanks, I was not aware that Vite does not read this anymore.

Done in 1ebcc6b3b — package.json browserslist is now the single source of truth:
- CSS: browserslistToTargets(browserslist()) → Lightning CSS
- JS: browserslistToEsbuild() → build.target

> Vite doesn't auto-read browserslist for Lightning CSS (unlike old PostCSS/autoprefixer), so explicit wiring is still required on both sides — but there's only one config to edit.

This also removes the maintained node versions from the browserslist query (browser bundle only) because it is not really needed there and would require manual filtering before passing it to the helper plugin.

index.html
tordans: Here is a different way to do this – what do you think?

4e951ca07 — window.context / window.id are only set when import.meta.env.DEV (same as the old standalone dev index.html; prod dist/index.html never exposed them).

af34007e4 — split library vs app bootstrap so iD.min.js no longer self-initializes:
- `lib.ts` → builds `iD.min.js` + `iD.css` (the library; sets `window.iD`). This is what embedders load.
- `main.ts` → now bootstrap-only, builds a tiny `bootstrap.js` (0.64 kB) loaded only by the standalone page.

`iD.min.js` (1.9 MB, self-contained) + `bootstrap.js` (references the global iD, does not re-bundle the library) + iD.css. This restores the pre-Vite contract where `iD.min.js` **doesn't self-initialize**, directly resolving the highest-severity review finding.

---

It looks like we will need something along these lines to not run into this issue (before the changes described above):
> `dist/iD.min.js` now self-boots. Any embedder that loads it and has a `#id-container` will double-init; embedders that called `iD.coreContext()` themselves now get an extra context.

main.ts
tordans: It is now wrapped to only run for dev ENV.

tordans (commented):
@k-yle thanks for the first review; I pushed an updated and added comments per answer.

> ### Updates since review (commits on top of 02b5d776e)
> 
> **Review feedback addressed:**
> - `cd8db8d4e` — remove `config/env.ts`, use `import.meta.env` directly in `config/id.ts`
> - `651de0444` — `@types/postcss-prefix-selector` instead of ambient shim
> - `899826ef9` — `vite build --mode stats` (Windows-safe) instead of `BUILD_STATS=true`
> - `63f5d9ae2` — `tsconfig`: `modules` + `test` + `**/*.ts`
> - `1ebcc6b3b` — single `browserslist` → CSS (Lightning) + JS (`build.target`); drop `maintained node versions`
> - `4e951ca07` — `window.context`/`window.id` dev-only + `@deprecated` + ESLint ban
> - `27c1c7205` — remove `css/entry.css` header comment
> - `f72c7ea8c` — document `npm run stats:view` in CONTRIBUTING
> - `af34007e4` — split library (`lib.ts` → `iD.min.js` + `iD.css`) from standalone bootstrap (`main.ts` → `bootstrap.js`). Restores pre-Vite embedder contract: `iD.min.js` does not self-initialize.
> - `23a0cd55b` — unit test for `idCssPlugin` (`.ideditor` scoping + dark-mode duplication)

Also a thing: The Vite way would be to not have the index.html hand written. But switching causes issues, so I did not look into it further for now.

vite-env.d.ts
```diff
@@ -0,0 +1,30 @@
...
+declare global {
+  interface Window {
+    /** @deprecated Debug-only. Use container DOM `.iD` once #11039 is resolved. */
+    context?: iD.Context;
+    /** @deprecated Debug-only. */
+    id?: iD.Context;
+  }
+}
+
+export {};
```
k-yle: the top half of this file won't work if you add any `import`/`export` statements. [ambient modules](https://www.typescriptlang.org/docs/handbook/modules/reference.html#ambient-modules) are really weird. 

so this `declare global` block needs to be removed, and merged into the existing global.d.ts file


p.s. it'd be good to also add this:

```ts
interface ViteTypeOptions {
 strictImportMetaEnv: boolean;
}
```

vite-env.d.ts
```diff
@@ -0,0 +1,30 @@
+/// <reference types="vite/client" />
```
k-yle: should this go in tsconfig.json -> `compilerOption.types` for consistency with the other 2 references, which are already defined in that file?

index.html
k-yle: i think that's a nice solution, although i would change
```diff
- bootstrap: path.resolve(__dirname, 'main.ts')
+ index: path.resolve(__dirname, 'index.html')
```

then we can delete `dist/index.html` and gitignore it. it becomes an auto-generated file, instead of a version-controlled copy

package.json
k-yle: yeah, you can just double click the .html file to open it in your browser

main.ts
k-yle: okay cool, then i guess there's no need to solve #11039 first

vite-env.d.ts
matkoniecz: @tordans

tordans: It is gone in the latest https://github.com/openstreetmap/iD/pull/12411/changes#diff-2d0c1425364be067f68ba13edc401ae775b507cb530d583dd46a38d81e90f0b3

tordans:
@k-yle here is an overview of the updates since https://github.com/openstreetmap/iD/pull/12411#pullrequestreview-4491967772

### Updates since that review

**Review feedback addressed:**
- `f344c4c9c` — move `window.context` / `window.id` typings into `modules/globals.d.ts`; add `ViteTypeOptions.strictImportMetaEnv`; put `vite/client` in `tsconfig` `compilerOptions.types`; drop `stats:view` (open `docs/statistics.html` after `build:stats`)
- `7e0590ea0` — use root `index.html` as the Vite input; delete/gitignore checked-in `dist/index.html` (now generated; standalone entry is `index.js`)

**Also:**
- Rebased onto current `upstream/develop` (conflicts resolved; dropped `happen`, kept develop’s vitest `execArgv`, pinhead, TS 6, etc.)
- `a78af3eff` — treat empty `ID_API_CONNECTION*` env strings as unset (`|| null`), matching pre-Vite behavior
- `cb7455f0d` — bump Vite ecosystem (`vite` 8.2, `vitest` 4.1.10, related tooling)
- `ad4f3034c` / `9eaf6283c` — fix CI typecheck (`containerNode` GetSet cast) and vite.config import extension for native configLoader + `tsc`

PR #12421, opened 2026-06-08
Copy crossing island tag to crossing nodes

KTibow:
### Description, Motivation & Context

When iD detects a road/path crossing and suggests creating a missing crossing node, it already copies known crossing details from the crossing way to the generated `highway=crossing` node. This currently includes the crossing type and `crossing:markings`.

This extends that behavior to also copy `crossing:island` from the crossing way to the generated crossing node.

This keeps refuge island information with the crossing node when iD automatically creates or repairs the crossing point, matching the existing behavior for crossing markings.

### Testing

- `npm run lint -- --quiet` passed.
- `npm run test:once -- test/spec/validations/crossing_ways.js` passed its assertions: `83 passed`, including the new `crossing:island` test.
- That focused spec still emitted the same unhandled `SyntaxError: "undefined" is not valid JSON` errors from `history.migrateHistoryData` during `coreContext` setup on my laptop. This matches the baseline local test issue seen in unrelated existing specs.
- Verified in preview environment: correctly copies crossing island tag over
image

### Disclosure

Generated with GPT-5.5 with human oversight.

### DIFF
--- a/modules/validations/crossing_ways.js
+++ b/modules/validations/crossing_ways.js
@@ -168,10 +168,13 @@ export function validationCrossingWays(context) {
                         return {};
                     }
                     if (['marked', 'unmarked', 'traffic_signals', 'uncontrolled'].indexOf(pathFeature.tags.crossing) !== -1) {
-                        // if the path is a crossing, match the crossing type and markings
+                        // if the path is a crossing, match the crossing type and details
                         var tags = { highway: 'crossing', crossing: pathFeature.tags.crossing };
                         if ('crossing:markings' in pathFeature.tags) {
                             tags['crossing:markings'] = pathFeature.tags['crossing:markings'];
+                        }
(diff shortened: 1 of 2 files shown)


Dimitar5555:
The linear way might cover more than one carriageway, so it would have `crossing:island=yes` across the whole length. But each `hw=crossing` node would not have such island, so `crossing:island=yes` on the node would be incorrect. See the example situation below.

image

KTibow:
That's true but I'd expect it to be rather uncommon and easily manually fixable (remember the add point action is manually triggered). Rapid also already does this (which I'm not using because it's broken on my GPU).

In that case, I'd also argue that that specific crossing would be better mapped as a crossing + a foot path + another crossing, especially when the current way's `crossing:island=yes` is mostly wasted (algorithms care about crossing nodes, not crossing ways).

PR #12425, opened 2026-06-09
migate osmEntity to ES6 classes

### LINKED ISSUE #10909: Refactor osmEntity

k-yle:
> [!warning]
> Do not squash this PR when merging. That will destroy the git history, it must be 2 separate commits

Closes #10909

This is the final part - the previous 7 PRs have split out most of the changes. unfortunately this PR is still quite big, and there's no way to shrink it even more. 

<table>
<tr>
<td>Old Model</td>
<td>New Model</td>
</tr>
<tr>
<td>

```mermaid
flowchart LR
    osmEntity -->|produces| osmNode
    osmEntity --> osmWay
    osmEntity --> osmRelation
    osmEntity --> osmChangeset
   
    osmNode -->|is subclass of| osmEntity[osmEntity]
    osmWay --> osmEntity
    osmRelation --> osmEntity
    osmChangeset --> osmEntity
```

</td>
<td>

```mermaid
flowchart LR
    classDef new fill:#0f0
    
    createEntity -->|produces| osmNode
    createEntity --> osmWay
    createEntity --> osmRelation
    createEntity --> osmChangeset
   
    osmNode -->|is subclass of| OsmAbstractEntity
    osmWay --> OsmAbstractEntity
    osmRelation --> OsmAbstractEntity
    osmChangeset --> OsmAbstractEntity
```

</td>
</tr>
</table>

why?
- in the old system, `osmEntity` is both an abstract class and a utility function to create a new instance. This creates a confusing situation where you can effecively instantiate an abstract class without declaring whether it's a node,way, or relation
- the old system is way too complicated to be modelled in typescript
- immutability was enforced at runtime only (if `iD.debug = true`). now it's also enforced at compile-time (using `readonly` attributes).
- ES6 classes support [go-to-definition](https://code.visualstudio.com/docs/editing/editingevolved) in your IDE, and are generally easier to work with.
- this is the last blocker before we can use typescript anywhere else in the core code


k-yle:
conflicts resovled, and the `EntityID` concept from #12450 has been incorporated

k-yle:
as a test, i rebased a few folders from that fork of iD where almost everything is TS. See: https://github.com/openstreetmap/iD/commits/189acb5b2e18d623cb6d7c9995843ea0727a935e - this branch has `actions`, `operations`, and `svg`.

everything passes, so i'm pretty confident that there's no flaws with the defintions in this PR

PR #12429, opened 2026-06-10
provide change event on hash behavior

### LINKED ISSUE #12426: Emit a custom hash change event
### Description

Since #12188, the uses of `history.replaceState` have been collapsed into one reused method, but since it still uses replaceState, no HashChangeEvent is dispatched natively.
I already gave implementing @k-yle's advice from https://github.com/openstreetmap/openstreetmap-website/pull/6811#issuecomment-4213218699 a shot, but couldn't make it work.
However, I can share my accompanying test suite addition.

### Screenshots

```javascript
describe('iD.patchHash', function () {
    it('dispatches a hash-change event', function () {
        const spy = fn();
        d3.select(window).on('hash-change', spy);
        iD.patchHash({foo: 'bar'});
        expect(spy).to.have.been.called;
    });
});
```

tyrasd:
closes #12426, unblocks https://github.com/openstreetmap/openstreetmap-website/pull/6811

@hlfan: on the osm website, this should allow you to hook into it as follows:

```js
id.ui().hash.on("change.embed", function() {
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :hammer: Development
 * Remove redundant software dependencies to reduce the amount of the code that is bundled with iD ([#11634], [#12307], thanks [@k-yle])
 * Update name-suggestion-index to v7.2 ([#12337], thanks [@bjornstar])
+* Introduce location hash `change` events ([#12429])
 
 [#8848]: https://github.com/openstreetmap/iD/issues/8848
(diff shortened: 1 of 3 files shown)


hlfan:
Aha so I've been foiled by async again. Good to know. 😅

tyrasd:
yeah, properly testing this was/is a bit tricky, as the other tests might or might not alter the hash simultaneously (for performance reasons, we're not running each test in a separate sandboxed DOM, so there really is only one single `location.hash` for all tests to share :sweat_smile: ).

PR #12436, opened 2026-06-12
show titles and descriptions when searching for wikidata items

### LINKED ISSUE #7460: Add wikidata description to dropdown list items
I was adding wikidata to an object, however when doing so there were a lot of items with the same name - meaning I had to mouseover each one to get the description to appear.

If there is more than one wikidata item of the same name, could the description be added so users can differentiate between them? E.g. "Green Hill Q3314551 - Census-designated place in Wilson County Tennessee"

Untitled-1

Would it even be possible to prioritise the geographically closest items in the list?

k-yle:
I know #11876 is controversial, but i think it would be super helpful for the `*:wikidata` fields, where it's common to have many results with the same name.

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

The same cannot be done for the `wikipedia=*` input field, due to API limitations, see [phab:T241437](https://phabricator.wikimedia.org/T241437) for context.

### DIFF
--- a/modules/ui/fields/wikidata.js
+++ b/modules/ui/fields/wikidata.js
@@ -162,7 +162,7 @@ export function uiFieldWikidata(field, context) {
                         .attr('lang', item.display.label.language)
                         .text(item.display.label.value),
                     description: item.display.description && item.display.description.value,
-                    title: item.aliases,
+                    title: item.aliases.join(', '),
                     terms: item.aliases
                 };
             });
--- a/modules/ui/fields/wikidata.js
+++ b/modules/ui/fields/wikidata.js
@@ -161,7 +161,8 @@ export function uiFieldWikidata(field, context) {
                         .attr('class', 'localized-text')
                         .attr('lang', item.display.label.language)
                         .text(item.display.label.value),
-                    title: item.display.description && item.display.description.value,
+                    description: item.display.description && item.display.description.value,
+                    title: item.aliases,
                     terms: item.aliases
                 };
             });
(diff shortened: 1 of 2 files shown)


k-yle:
this also makes iD's UI more similar to wikimedia's own UI:

<table>
<tr>
<td>wikidata.org</td>
<td>wikipedia.org</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

modules/ui/fields/wikidata.js
```diff
@@ -161,7 +161,8 @@ export function uiFieldWikidata(field, context) {
                         .attr('class', 'localized-text')
                         .attr('lang', item.display.label.language)
                         .text(item.display.label.value),
-                    title: item.display.description && item.display.description.value,
+                    description: item.display.description && item.display.description.value,
```
tyrasd, line 164: We should make this also be rendered via a d3 function (similar to `display`), in order to set the correct language metadata on the `span`. But let's do that separately in a follow-up PR.

tyrasd (approved):
Yes, it's perfect for this, thanks!

1ec5:
> The same cannot be done for the `wikipedia=*` input field, due to API limitations, see [phab:T241437](https://phabricator.wikimedia.org/T241437) for context.

That MediaWiki feature request was declined. Would any of [these alternative APIs](https://phabricator.wikimedia.org/T241437#6138793) be workable for us?

modules/ui/fields/wikidata.js
k-yle: done in #12485

PR #12449, opened 2026-06-15
show field label (or tag key) in invalid URL validation message

tyrasd:
addresses another point raised in #12366: sometimes, the tag diff is not available to indicate which tag is actually causing an _invalid URL_ warning (e.g. when the value does not look like a URL at all, or there are multiple invalid URLs in the semicolon-separated tag value). This shows the field name (or tag key if no field is rendered for the particular tag) in the message:

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1637,8 +1637,8 @@ en:
         message_multi: '{feature} has multiple invalid email addresses'
         reference: 'Email addresses must look like "user@example.com".'
       website:
-        message: '{feature} has an invalid URL'
-        message_multi: '{feature} has multiple invalid URLs'
+        message: '{feature} has an invalid URL in {where}'
+        message_multi: '{feature} has multiple invalid URLs in {where}'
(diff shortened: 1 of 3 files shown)


data/core.yaml
```diff
@@ -1637,8 +1637,8 @@ en:
         message_multi: '{feature} has multiple invalid email addresses'
         reference: 'Email addresses must look like "user@example.com".'
       website:
-        message: '{feature} has an invalid URL'
-        message_multi: '{feature} has multiple invalid URLs'
+        message: '{feature} has an invalid URL in {where}'
```
k-yle, line 1640: would _`[…] in the 'Image' field`_ be easier to understand than _`[…] in Image`_? 

To me, the current phrasing sounds like there's a something wrong _inside_ the image file

Likewise, _`Point has an invalid URL in URL`_ might be confusing, since the raw tags are not visible by default. maybe _URL_ should be a hyperlink which scrolls down to the raw tag section and opens that section if it's collapsed?

Issue #12442, opened 2026-06-13
Note actions do not appear when opened using hash


danieldegroot2:
### How to reproduce the issue?

Load note via hash, for example

https://www.openstreetmap.org/edit?note=3143577#map=18/51.088700/4.237000

https://ideditor.netlify.app/#map=18.00/51.08870/4.23700&disable_features=boundaries&background=DV_OrthoPhoto&id=note/3143577

<sub>At time of writing the example is an open note; it was also tested with other open notes.</sub>

The note comment input box and buttons do not appear. The user is unable to comment or close/reopen the note.

Chrome 149.0.7827.115

### Screenshot(s) or anything else?

it might appear as if it is off the screen in this case but it is not there anyways.
Zooming out the browser does not resolve this issue.
Clicking on a different note and then this note again does, however.
<sub>as a workaround the user can 'go to the note on osm.org' of course<sub>

Example with browser zoomed out to show there is no comment input box at all.

<img width="3837" height="1669" alt="Image" src="https://github.com/user-attachments/assets/af4ff64e-3cbe-4092-935e-2bd35425c0de" />

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit, Development version at ideditor.netlify.app

### What version numbers does this issue affect?

2.40-live, 2.41-dev

### Which browsers are you seeing this problem on?

Chrome

PR #12450, opened 2026-06-15
[typescript] make osm id handling symetric, explicit with return types

tyrasd:
Previously `fromOSM` expected a number for the `id`, but `toOSM` returned a string.

We should be consistently using numbers for the id part wherever it makes sense. This prevents bugs like #12442 which are caused by iD internally sometimes storing the id of a thing as a string, but the OSM API returning it as a number (when using JSON endpoints), which then in turn causes statements like `_note.id === context.selectedNoteID()` to be false just because of mismatching types.

### DIFF
--- a/modules/osm/id_manager.ts
+++ b/modules/osm/id_manager.ts
@@ -1,4 +1,8 @@
-export type FeatureType = 'node' | 'way' | 'relation';
+export type OsmType = 'node' | 'way' | 'relation';
+/**
+ * Combined internal representation of the OSM type and OSM id of an entity.
+ */
+export type EntityId = string;
 
 /**
  * All newly created features need an ID, so this singleton
@@ -13,29 +17,29 @@ class OsmIdManager {
         relation: -1,
     };
 
-    fromOSM(type: FeatureType, id: number) {
+    fromOSM(type: OsmType, id: number): EntityId {
         return type[0] + id;
     }
 
-    toOSM(id: string) {
+    toOSM(id: EntityId): number {
         var match = id.match(/^[cnwr](-?\d+)$/);
         if (match) {
-            return match[1];
+            return +match[1];
         }
-        return '';
+        return NaN;
     }
 
-    type(id: string) {
-        return <FeatureType>(
+    type(id: EntityId): OsmType {
+        return <OsmType>(
             { c: 'changeset', n: 'node', w: 'way', r: 'relation' }[id[0]]
         );
     }
 
-    key(entity: iD.OsmEntity) {
+    key(entity: iD.OsmEntity): string {
         return entity.id + 'v' + (entity.v || 0);
     }
 
-    newId(type: FeatureType) {
+    newId(type: OsmType): EntityId {
         return this.fromOSM(type, this.next[type]--);
     }
 }
--- a/modules/osm/relation.js
+++ b/modules/osm/relation.js
@@ -6,7 +6,7 @@ import { geoExtent, geoPolygonContainsPolygon, geoPolygonIntersectsPolygon } fro
 import { osmIdManager } from './id_manager';
 
 /**
- * @typedef {{ type: import('./id_manager').FeatureType; id: string; role: string }} RelationMember
+ * @typedef {{ type: import('./id_manager').OsmType; id: string; role: string }} RelationMember
  */
 
 /**


modules/osm/id_manager.ts
```diff
@@ -1,4 +1,8 @@
-export type FeatureType = 'node' | 'way' | 'relation';
+export type OsmType = 'node' | 'way' | 'relation';
```
tyrasd, line 1: This is just a small nitpick. But in my model the feature is the augmented internal representation of an OSM entity, while this type describes more the OSM API / data model side of things.

modules/osm/id_manager.ts
```diff
@@ -13,29 +17,29 @@ class OsmIdManager {
...
 
-    toOSM(id: string) {
+    toOSM(id: EntityId): number {
         var match = id.match(/^[cnwr](-?\d+)$/);
         if (match) {
-            return match[1];
+            return +match[1];
         }
-        return '';
+        return NaN;
```
tyrasd, line 29: To check if this does not potentially cause side effects somewhere? I hope we don't rely on this method returning literally `''`. 🙈

modules/osm/id_manager.ts
```diff
@@ -13,29 +17,29 @@ class OsmIdManager {
...
     }
 
-    type(id: string) {
-        return <FeatureType>(
+    type(id: EntityId): OsmType {
+        return <OsmType>(
```
tyrasd, line 33: This is a bit ugly. if one calls the method with an invalid / messed up `id` (e.g. `osmIdManager.type('foo')`), it will actually return `undefined`, which might trigger bugs further along.

I think changing the return type to `OsmType | undefined` is also a but ugly, and unnecessary in almost all cases. Maybe a small assert here and throwing a runtime exception if it fails would be sufficient. (also applicable above to the `toOSM` method.

modules/osm/id_manager.ts
```diff
@@ -1,4 +1,8 @@
-export type FeatureType = 'node' | 'way' | 'relation';
+export type OsmType = 'node' | 'way' | 'relation';
+/**
+ * Combined internal representation of the OSM type and OSM id of an entity.
+ */
+export type EntityId = string;
```
k-yle, line 5: in the future we could make this stricter: 

```suggestion
export type OsmTypeShort = 'c' | 'n' | 'w' | 'r';
export type EntityId = `${OsmTypeShort}${number}`;
```

we could also do this now, but it might cause some false-positive errors during the migration. 


should be easy to update #12425 to use this approach

modules/osm/id_manager.ts
k-yle: yeah, this is quite risky. it will definitely break this line:

https://github.com/openstreetmap/iD/blob/6dcb8190fa1cce4471da3d4efbea846aea4c07ab/modules/osm/entity.js#L94

and it will presumably also break JXON.

to be safe, maybe it's better to keep this as a `string`[^1] for now? after #12425 it should be a bit safer to refactor


[^1]: or `` `${number}` `` if you prefer, it [accepts the negative sign](https://www.typescriptlang.org/play/?target=10#code/C4TwDgpgBAysBOBLAdgc0QM0RAJgOQFcBbAIwnigF4oADAEgG9liz4BfGqAHygHJeAUEN4BqAAy8oAZwCGwRFKwQpsBCnRL8LcgG4BvALQTpchUpVwkaTNi2ld+47PmLsFtdc2F78PYbEAdE6mrsqqVhq23qx+AIwQwS7m4eo2uNEOvLEyiWZuKZ5R2r6OQRBGks55YZapXsV+EJUhybWF6Q36zUn5bZEdPnpAA)

modules/osm/id_manager.ts
k-yle: https://github.com/openstreetmap/iD/pull/12450/changes#r3426788861 would make this typecast safe

if we trust the input, then we can trust that the typecast is safe, so we don't need ` | undefined`

modules/osm/id_manager.ts
matkoniecz: @tyrasd Can you make it back to '' and open issue to refactor it separately?

@k-yle Or maybe it is safe now that #12425 was merged?

matkoniecz:
>        × converts a way to jxon 27ms

is failing already, so I guess that `''` should be restored or at least this migration fixed?

PR #12453, opened 2026-06-16
Apply a proper intended logic fix for incorrect loop increment variable

RudyTheDev:
The original issue report #12128 points out that the inner loop is using the outer loop's counter variable in `move.js`'s `limitDelta` function. The PR fix #12129 indeed corrects this. However the author didn't explain what the observed problem was.

So this is the kind of scenario we are talking about, where a way is connected to another way and we attempt to move it:

snap1

As we move it without intersecting the attached way, everything is fine:

snap2

But as soon as we touch the attached way, the moved way is snapped back to attached way, only preserving movement along the attached way:

snap3

Just to emphasize the point, this is one very specific edge case that this whole function is considering. There are tons of ways all of this behaves weirdly that have nothing to do with this function and loop:

current behaviour

Anyway, after the fix, the behavior is exactly the same as it was before. I explain this [in the comment on that PR](https://github.com/openstreetmap/iD/pull/12129#issuecomment-4583397670). The short version that the calculation always results in the same delta, regardless what intersections are present. In other words, this method potentially finds the first conflicting intersection (i.e. not a connected point) and if so - snaps back.

It is almost impossible to reproduce this "issue" as a user-facing bug, which is why no one ever saw or reported it. (Since the original issue and PR are all AI slop, the reported "Observe inconsistent or incorrect behavior in how movement is restricted" is nonsense.) The only way to reproduce this would be to have a convoluted synthetic test that will never happen in real-world.

So in essence this PR fixes some redundant sub-optimal looping in the code and makes the code's intention more clear.

### DIFF
--- a/modules/actions/move.js
+++ b/modules/actions/move.js
@@ -327,10 +327,12 @@ export function actionMove(moveIDs, tryDelta, projection, cache) {
             var unmovedPath = unmovedNodes.map(function(n) { return projection(n.loc); });
             var hits = geoPathIntersections(movedPath, unmovedPath);
 
-            for (var j = 0; i < hits.length; i++) {
-                if (geoVecEqual(hits[j], end)) continue;
+            if (hits.length) {
+                // snap delta back to the edge we are attached to, so we only move along the edge,
+                // not away from it since that causes intersection(s)
                 var edge = geoChooseEdge(unmovedNodes, end, projection);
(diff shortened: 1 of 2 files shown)


tyrasd (approved):
Thanks for the in-deepth analysis, this helped a lot to understand the issue and fix.

I did also add a test for this behaviour.

PR #12455, opened 2026-06-17
introduce two-column layout for radio fields

tyrasd:
Follow-up to #12343, specifically https://github.com/openstreetmap/iD/pull/12343#issuecomment-4602885272:

If the options of a radio field do not fit into a single line, but they are short enough for a two-column layout, use it:

Odd numbered options get a full line for the last option (even allowing to fit longer texts if needed):

Slightly unrelated, but this also tweaks the way the single-line layout is decided: now, it does not force the options to all have the same width, instead, if there is enough space to fit all options in a single line, it will do it by using all the available space. This is most prominent in languages with longer words for some options, like here:

For aesthetically reasons, we could potentially use an evenly-spaced single-line layout if all entries do fit even in that case, but I did not bother to implement this yet. Let me know if you think if this is worth it or not. Below is the comparison between the old and new rendering for this case:


matkoniecz:
> For aesthetically reasons, we could potentially use an evenly-spaced single-line layout if all entries do fit even in that case, but I did not bother to implement this yet. Let me know if you think if this is worth it or not. Below is the comparison between the old and new rendering for this case:

even spacing looks far neater, not sure how much effort and extra code it would take to implement

tyrasd:
not very. it does now switch between equally spaced single line layout and "condensed" single line layout as you narrow down the sidebar width:


https://github.com/user-attachments/assets/10503446-c1c1-489a-866b-a6772eed8e5f

PR #12458, opened 2026-06-17
Add flowerbed flower pattern

RudyTheDev:
This adds a pattern to `landuse=flowerbed`, which is a subtly different pattern to grass. Currently, flowerbeds don't have a pattern.

flowerbed vs grass

I also tried alternatives with the flowers being bigger and denser and more random, but everything ended just being noisy and distracting. For example, denser:

flowerbed vs grass denser

I didn't try changing colors as I think that will create more confusion than anything.


tyrasd:
cool, thanks, I think it looks great :+1: would it maybe a good idea to use this pattern also for `leisure=garden`?

RudyTheDev:
> would it maybe a good idea to use this pattern also for `leisure=garden`?

I don't actually think so. The reason being that a lot of flowerbeds are mistagged as gardens and this was mainly because the tag didn't exist but then also Carto not rendering it. And mappers still do it and there is still some debate as to exact meaning of those. I don't have any quantitative analysis, but it's easy to find locations with "gardens" that are arguably not really gardens once you start looking at these in detail, for example:

image

I guess what I'm saying is that they should probably be distinct patterns. (Even garden icon is _slightly_ misleading.) I guess similar points to discussion in [openstreetmap-carto#4564](https://github.com/openstreetmap-carto/openstreetmap-carto/issues/4564). I realize though that this matters significantly less for iD. May be I'm overthinking it.

PR #12463, opened 2026-06-18
remove the 'chai' library

k-yle:
vitest supports chai-style assertions, so we don't need to explicitly install the package ourself


tyrasd:
slightly related: #12300

PR #12466, opened 2026-06-18
remove the 'alif-toolkit' library

k-yle:
Follow up to #8741 - this code was disabled for modern browsers 5 years ago. now we don't even support 5 year old browsers[^1].

as far as I can tell, the text-shaping bugs reported in #6018 have indeed been solved by the rendering engine.

cc'ing @mapmeld since this is all your work

[^1]: (tangentially discussed in #12307)


mapmeld:
It makes sense to me; the only problem would be if Safari has fixed their
rendering issue.  If I remember this right, Safari Webkit is the only
renderer on iOS because even other browsers (did? still do?) use it

If the issue is I need to clean up dependencies for alif-toolkit , then I
can look at it next week

tyrasd:
> the only problem would be if Safari has fixed their rendering issue

To me it doesn't seem so, also not on MacOS. Screenshot from the example shown in https://github.com/openstreetmap/iD/pull/6018#issue-418126075 (near 36.17752/44.02082) on Safari 27:

this branch:
image

vs. how it looks like in current `develop`:
image

vs. Firefox (PR 12466):
image

vs. Firefox (`develop`):
image

I assume this means that we still need the fix and thus cannot remove `alif-toolkit`?

k-yle:
oh, typical safari... lets close this then and wait another few years 😆

 this is [webkit bug #266832](https://bugs.webkit.org/show_bug.cgi?id=266832) right?

mapmeld:
If someone wants to research fixing this longtime SVG textPath issue in Safari/WebKit, I chatted to a coding agent and created a commit ( https://github.com/WebKit/WebKit/commit/fceb2cb0ed ) but I couldn't make enough RAM and disk space on my Mac to build WebKit (seriously!). So I can't fully vouch for the code, but it did find and change the character-by-character rendering, and I asked about memory issues in C++.

I believe that a minimal (not debug) build is something like `Tools/Scripts/build-webkit --release --jobs=2 --no-unified-builds --make-args="DEBUG_INFORMATION_FORMAT=dwarf"`

k-yle:
Didn’t they fix this in https://github.com/WebKit/WebKit/pull/68085 , or was that fix only for indic scripts?

mapmeld:
@k-yle Excellent find! I downloaded the Safari Technology Preview from https://webkit.org/downloads/ and checked the examples I have on https://mapmeld.com/text-path-problems/
It worked for Latin (é) and an Indic alphabet, but the Arabic example did not use text-shaping or RTL order

PR #12470, opened 2026-06-19
Enhance OSM API error messages with specific HTTP status handling

### LINKED ISSUE #12255: Improve OSM API error feedback in uiStatus (Resolve TODO)
### Description

I noticed a TODO: nice messages for different error types in modules/ui/status.js. I would like to implement this by adding logic to show more specific messages based on HTTP error codes (like 500, 503, etc.) instead of the generic error message currently used.
Code Reference: https://github.com/openstreetmap/iD/blob/develop/modules/ui/status.js#L52

### Screenshots

_No response_

Kayd-06:
## Description
This PR addresses an existing `TODO` in `modules/ui/status.js` regarding the display of nicer error messages based on different error types from the OpenStreetMap API. 

Previously, any non-rate-limited and non-connection-switched error would fall back to a generic "Unable to reach the OpenStreetMap API" message. This update introduces specific, user-friendly localization strings for common HTTP error codes.

## Changes Made
- Added specific HTTP error translations to `data/core.yaml` (e.g., 400 Bad Request, 401 Unauthorized, 403 Forbidden, 404 Not Found, 502 Bad Gateway, 503 Service Unavailable, and 504 Gateway Timeout).
- Imported `localizer` into `modules/ui/status.js`.
- Modified the status update function to check `err.status`. If a specific `error_XXX` localized message exists, it displays that; otherwise, it gracefully falls back to the original generic error message.

### Testing Instructions:
1. Trigger a network failure or mock an API response to return one of the handled HTTP statuses (e.g., 403 or 503).
2. Verify that the application's bottom-right status bar displays the user-friendly specific message instead of the generic one.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -596,13 +596,13 @@ en:
   osm_api_status:
     message:
       error: Unable to reach the OpenStreetMap API. Your edits are safe locally. Check your network connection.
-      error_400: "Bad Request: The server could not understand your request."
-      error_401: "Unauthorized: Please ensure you are logged in."
-      error_403: "Forbidden: You don't have permission to perform this action."
-      error_404: "Not Found: The requested data could not be found."
-      error_502: "Bad Gateway: The OpenStreetMap API is experiencing issues."
-      error_503: "Service Unavailable: The OpenStreetMap API is temporarily down for maintenance."
-      error_504: "Gateway Timeout: The OpenStreetMap API is taking too long to respond."
+      error_400: "Bad Request: The server could not understand your request. Your edits are safe locally."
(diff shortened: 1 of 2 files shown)


data/core.yaml
```diff
@@ -596,6 +596,13 @@ en:
   osm_api_status:
     message:
       error: Unable to reach the OpenStreetMap API. Your edits are safe locally. Check your network connection.
```
matkoniecz, line 598: "Your edits are safe locally." is missing from new error descriptions

is it intentional?

Kayd-06: Thanks! 
It was not intentional. I've just pushed a commit that adds the missing 'Your edits are safe locally.' text to all the new error messages.

PR #11850, opened 2026-02-05
Added the feature for custom backgrounds

JaiswalShivang:
# feat: Multiple custom background layers with CRUD operations
**Fixes #8874**
## Summary
This PR enables users to add, edit, name, and delete multiple custom background layers in the iD editor. Previously, only a single custom background was supported.
## Demo Video

https://github.com/user-attachments/assets/17fb1419-5ef3-4152-bff4-fc8339af677a

> **Note:** The video demonstrates adding, editing, naming, and deleting multiple custom backgrounds.
### User-Facing Changes
- **Multiple Custom Backgrounds**: Users can now add unlimited custom background layers
- **Dedicated Section**: New "Custom Backgrounds" section in the Background pane
- **Named Backgrounds**: Each custom background can have an optional name
- **URL Fallback**: Backgrounds without names display a truncated URL (e.g., "Custom: tiles.example.com/berlin/…")
- **Add/Edit/Delete**: Full CRUD operations with intuitive icons (pencil for edit, trash for delete)
- **Delete with Confirmation**: Confirmation dialog to prevent accidental data loss
- **"Custom:" Prefix**: All custom backgrounds are clearly labeled with "Custom:" prefix
- **Data Persistence**: All custom backgrounds persist across sessions
### Technical Changes
1. **Data Format**: Custom backgrounds stored as JSON array in `background-custom-templates` preference:
   ```json
   [
     { "id": "custom-1", "name": "My Tiles", "template": "https://..." },
     { "id": "custom-2", "name": "Berlin 2025", "template": "https://..." }
   ]

PR #12473, opened 2026-06-20
Add Lens support: import and apply custom CSS lenses

### LINKED ISSUE #11189: Different map styles ("lenses")

kaligrafy:
May close or partially close #11189

## Summary
Adds a **Lens** section to the **Map data** pane so users can import a custom CSS file and have it restyle the map. "Lens" is the name proposed for this feature in #11189; it also avoids any overlap with the existing light/dark `context.theme()`.

- Pick the built-in default lens or import a `.css` file, kept in `localStorage`.
- The active lens's CSS is injected into the page via a dedicated `<style>` element (replaced on lens change).
- OSM keys the lens styles via `tag-{key}[-{value}]` selectors are emitted as classes on map elements (through `svgTagClasses`), so a lens can style features by tag/value. Keys are matched against each entity's real tags (`:` normalized to `_`); computed tokens (status, wikidata, surface categories) are skipped.
- The lens is applied at startup and re-applied with a redraw whenever the selection or stored lenses change.

Only the English source strings are added to `data/core.yaml` (other locales come from Transifex).

## Overriding core styles (cascade layer)
Core CSS is wrapped in `@layer ideditor` at build time, and the lens CSS is injected **outside** any layer. Since unlayered rules always beat layered ones, a lens overrides core feature styling at **any** specificity — no `!important` and no need to mirror core selector shapes. The cascade is otherwise preserved, so interactive states (`.selected`, `.hover`, …) still apply.

Side effect to flag: when iD is **embedded**, an unlayered stylesheet on the host page would now take precedence over iD's core styles. In practice this is negligible (host CSS does not target `path.line.stroke.tag-*`), but noting it for visibility.

## Scope
This is the CSS-import subset of #11189 (styles that replace/augment the current style). It does not lock in a styling language — it reuses iD's existing `tag-*` classes — and does not address the editing-behavior categories (disable/grey-out features) discussed there.

## Security
Imported CSS is sanitized: `@import` rules and external `url()` references are stripped (only inline `data:` URIs survive) to prevent remote fetches and data exfiltration. A "import only trusted lenses" warning is shown in the UI.

## Test plan
- [x] Unit tests for CSS tag-key extraction, `:`/`_` disambiguation, and sanitization (`test/spec/core/lenses.js`).
- [x] Manual: import a CSS file that styles features by tag (e.g. `path.line.stroke.tag-highway-cycleway`, `path.area.fill.tag-building-school`); matching features get styled, and reverting to the default lens clears both the classes and the injected CSS.


matkoniecz:
Is it necessary for user to trust whoever provides CSS? Presumably there are ways to include malicious payload there, right?

kaligrafy:
> Is it necessary for user to trust whoever provides CSS? Presumably there are ways to include malicious payload there, right?

Good catch, I will see if there are packages or simple code to detect malicious code in CSS

matkoniecz:
I suspect that it is rather call to add "use only secure sources" reminder or doing it via curated gallery

I suspect that "is it a malicious CSS" checking is only tiny easier than "is it a malicious code"

kaligrafy:
Added a sanitazion step for @import and urls inside the uploaded css. Also add a warning to validate css before importing. c7519224e3fdcd643076d9859301d08805a2d3d8

tordans:
@kaligrafy two quick thoughts

**Naming**
How about we stick with "lenses" for this? It is a bit unconventional but it has the advantage to scope this feature distinctively and it is something that is around in the iD discussion space for a while; we might want to wait for Martin's input though

**Lens creation / Lens debugging**
I there a good way and place to show kind of linting errors or maybe "which classes are applied in viewport" somewhere? And maybe also "which classes are available"? I wonder how I would find out which classes I can actually use. And once I have a theme, if they actually work… – apart from trial and error, reading source code/inspector output and visually comparing results.

tordans:
@kaligrafy could you share some example themes that we can use to test this?
Following up on my last message: I am not sure how I would build one right now; I could not find the classes using the browser inspector on the map element.
>image
>
> [Location](https://pr-12473--ideditor.netlify.app/#disable_features=boundaries&map=20.00/52.47425/13.44577&background=Bing)

https://github.com/openstreetmap/iD/issues/11189 has a view ideas on lenses/themes that we should try out, for example…
- highlight building types
- highlight maxspeed data
- "highlight objects that do not follow an embedded minimum field requirements"
- disable editing for selected features via the style
- gray out features to provide focus to a top, like the pedestrian mode [that Rapid explored
](https://github.com/facebook/Rapid/pull/1394#issuecomment-2091620754)
- something like a color blind mode – [see Rapids take](https://github.com/facebook/Rapid/pull/1230)

tordans:
**Move from Preferences to "Map Data"?** I think the "Map Data" Panel is a better fit for this looking at the panels we have. We use the Preferences for personal settings like privacy (and favorite presets/shorcuts) which dont have a better panel. But this is very much about how the map data is presented and potentially filtered.
image

tordans:
**Shortcuts?** Does this already have a shortcut to toggle the custom, last selected lens and the default lens? Like the b-Shortcut that toggles between the last two map backgrounds, for example.

**Multiple themes?** We know from things like the background select, that it is not long until we need a way to not just have one thing but multiple. Maybe we should have this from the start and make the UI a list with "Default" and add theme. The UI we where discussing in https://github.com/openstreetmap/iD/pull/11850#issuecomment-4060676668 would work here as well, IMO. (Unfortunately this thread is a bit messy, will try to clarify this…)

kaligrafy:
I will rename to lenses and provide screenshots. The imported css should use the same format as the css files upstream. I will provide examples too. I can also move it to the Map data section. The required classes are added automatically on css file read to reduce unneeded classes.

kaligrafy:
- [x] moved to Map Data
- [x] renamed to lenses
- [x] fix css layering (needed too many !important to work)

kaligrafy:
Here is a sample CSS file to test:
[example-lens.css](https://github.com/user-attachments/files/29208671/example-lens.css)

kaligrafy:
And a screenshot video: 

https://github.com/user-attachments/assets/e4d22899-f417-4a63-a745-809d27add66f

kaligrafy:
The choice of using CSS for now, instead of MapCSS or JSON like Rapid is to get the feature up and running faster, and then after discussions, we could accomodate different solutions.

kaligrafy:
> **Shortcuts?** Does this already have a shortcut to toggle the custom, last selected lens and the default lens? Like the b-Shortcut that toggles between the last two map backgrounds, for example.
> 
> **Multiple themes?** We know from things like the background select, that it is not long until we need a way to not just have one thing but multiple. Maybe we should have this from the start and make the UI a list with "Default" and add theme. The UI we where discussing in [#11850 (comment)](https://github.com/openstreetmap/iD/pull/11850#issuecomment-4060676668) would work here as well, IMO. (Unfortunately this thread is a bit messy, will try to clarify this…)

For now, shortcuts for lenses are not implemented, but could easily be after the main preset shortcut systme is merged.

PR #12480, opened 2026-06-22
stop using yes when nothing was selected in typeCombo (bad PR)

### LINKED ISSUE #11618: Clandestinely adding trade=yes
Add a preset `shop=trade`.
Expand the list of types... but don't choose anything. iD will clandestinely add the tag `trade=yes`.

matkoniecz:
fixes #11618

### DIFF
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -148,10 +148,6 @@ export function uiFieldCombo(field, context) {
         });
         if (found) return found.key;
 
-        if (field.type === 'typeCombo' && !dval) {
-            return 'yes';
-        }
-
         return restrictTagValueSpelling(dval) || undefined;
     }
 


matkoniecz:
it fixes in fact https://github.com/openstreetmap/iD/issues/11618 and I see no obvious problems with it

but not fully sure I have not missed anything

tyrasd:
>  I see no obvious problems with it

`typeCombo` is (mainly?) used for the primary tag of generic fallback presets, which should never be set to nothing, because then the previously preset will jump back to the global fallback preset for a completely untagged feature.

For example: when adding a feature with the generic Craft preset adds `craft=yes` as the feature's primary tag. Not choosing anything in the _Type_ column should not remove the `craft` tag. In this PR, this is broken.

I think in the above mentioned issue, the `trade` field should just be a regular `combo` field (instead of `typeCombo`) if the fallback to `yes` is not intended.

matkoniecz:
whops, now I get what is happening here

thanks for catching this!

should have grepped for this interface type

now https://github.com/openstreetmap/id-tagging-schema/pull/2362 has a proper fix (and confirmed it has no extra consequences, this field does not appear elsewhere)

PR #12481, opened 2026-06-23
don't add field's default value when changing preset and both before&after preset have the field 

### LINKED ISSUE #12472: Don't add building=yes to indoor=room area when changing the category

When you change the category of a POI that is combined with indoor=room, like e.g. https://www.openstreetmap.org/way/346580249 to a different category like e.g. restaurant, then `building=yes` is automatically added although it should not be.

2.40.0

tyrasd:
fixes #12472

todo:
* [ ] add unit test
* [ ] changelog


matkoniecz (changes requested):
per description of PR: unit tests are missing

PR #12483, opened 2026-06-23
unify diacritics stripping logic

k-yle:
We have 3 different code blocks for stripping diacritics, one of which uses a library that is [a bit lim](https://github.com/osmlab/name-suggestion-index/issues/5017#issuecomment-817343126)[ited](https://github.com/andrewrk/node-diacritics/issues/21) and [odd](https://github.com/andrewrk/node-diacritics/issues/36). 

we don't need this library anymore, since all browsers can do this natively [since 2020](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape#browser_compatibility).


1ec5 (approved):
This should work at least as well as before, at least in Latin/Cyrillic/Greek text.

matkoniecz:
Would it make sense to add a test? Maybe for something that used to fail and now works?

Or is it too much for such minor change?

I guess it can be tested by confirming that "Xrt InstXllXtiXn" does not work in preset search while "Ąrt Instąllątión" still works? (really fake example, sorry)

matkoniecz (commented):
seems to work from cursory checking

tyrasd (approved):

PR #12485, opened 2026-06-23
wrap wikidata description with <span lang=...>

k-yle:
followup to https://github.com/openstreetmap/iD/pull/12436#discussion_r3403493247

### DIFF
--- a/modules/ui/combobox.js
+++ b/modules/ui/combobox.js
@@ -411,10 +411,14 @@ export function uiCombobox(context, klass) {
                     } else {
                         labelSpan.text(d.value);
                     }
-                    if (d.description) {
+                    if (typeof d.description === 'string') {
                         sel.append('span')
                             .attr('class', 'combobox-option-description')
                             .text(d.description);
+                    } else if (typeof d.description === 'function') {
+                        sel.append('span')
+                            .attr('class', 'combobox-option-description')
+                            .call(d.description);
                     }
                 });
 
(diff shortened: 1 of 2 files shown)


matkoniecz:
what would be practical effects here? RTL/LTR handling? Anything else?

also, any idea how can I look at HTML of popup? Firebug from Firefox (however it is named now) et all has problem - when I try to select element popup disappears

probably unrelated  to this change or earlier ones, but why "Conrada w Krakowie" fails to find https://www.wikidata.org/wiki/Q123909561 ?

k-yle:
this helps with the [han unification problem](https://en.wikipedia.org/wiki/Han_unification), and it makes life easier for browser translation extensions. so it's just a best-practice.

To actually test it, you would need a wikidata item with:
- a `jp` description; and
- no `zh` description; and
- the `jp` description contains a character that [renders differently depending on the language](https://en.wikipedia.org/wiki/Han_unification#Examples_of_language-dependent_glyphs), such as `入` 
- and iD's UI language set to `zh`

(in that example, `jp` and `zh`could be replaced with any combination of CJK languages. but i can't find a single example)

k-yle:
> probably unrelated to this change or earlier ones, but why "Conrada w Krakowie" fails to find [wikidata.org/wiki/Q123909561](https://www.wikidata.org/wiki/Q123909561) ?

yeah that's unrelated and i have no idea why their search doesn't match substrings like that. presumably for performance reasons 🤷

tyrasd (approved):

PR #12500, opened 2026-06-28
note that Tor is fine and iD works with it

matkoniecz:
As far as I know no special support should be needed for iD.

I also tested it and editing works fine.

Note that https://www.openstreetmap.org/user/new gives "Unfortunately we are not currently able to create an account for you automatically.

Please contact support for assistance - we will try and deal with the request as quickly as possible." but as noted it is out of scope for iD.

(not sure whether Tor should be listed here at all, considered also removing it)


tyrasd:
> not sure whether Tor should be listed here at all, considered also removing it

I would tend slightly towards just removing the line, as I don't think tor is too relevant for the typical usage of iD

matkoniecz:
Maybe listing Tor Browser in Browser Compatibility would make more sense? Or should it be also skipped?

tyrasd:
>listing Tor Browser in Browser Compatibility 

yeah, that makes sense to me :+1:

matkoniecz:
moved around

tyrasd (approved):

PR #12505, opened 2026-06-29
do not describe Mercator meters as meters

### LINKED ISSUE #9792: Wrong offset values in meters
### URL

https://www.openstreetmap.org/edit#map=18/46.43038/30.71566

During imagery alignment the offset in meters shows incorrect values comparing to the real imagery offset. 

Here is how to reproduce it:
1. Select any imagery (I used Bing), reset offset. Objects layer could be switched off for clarity.
2. Move map to some contrast object and measure its width with the ruler and any other program (see results from Google Earth). 

As you can see, the width of this ice rink is correctly shown as 40 meters so ruler is correct.

3. Set offset to "-40, 0". We expect moving of the object left and its right edge should be aligned with the left edge of the ruler. But here is what we see:

4. Try offset "40, 0" expecting the opposite: right edge of the object should be aligned with the left edge of the ruler. The result:

As you can see, it is not 40 m but about 27 m, 1.45 size smaller.
I suspect that latitude and longitude are mixed up in the parameters of the functions which convert degrees delta to meters. Why I think so? There is the same problem in JOSM which I noticed after I applied offset from offset database. iD don't use offset database but there is conversion of degrees to meters. Here is the example of possible error:

Point 1 (lat, lon): (46.429716,  30.714416)
Point 2 (lat, lon): (46.429716,  30.714942)
dLon = 30.714942 - 30.714416 = 0.000526
Let's calculate distance using `geoLonToMeters()`
```js
0.000526 * (TAU * EQUATORIAL_RADIUS / 360) * Math.abs(Math.cos(46.4297 * (Math.PI / 180))) = 40.2
```
which is correct. Now let's use `geoLatToMeters`
```
0.000526 * (TAU * POLAR_RADIUS / 360)/40 = 58.5
```
which is exactly the same offset we need to enter to get correct imagery offset.

So I suspect that `geoLatToMeters()` was used internally for map **longitude** offset and `geoLonToMeters() was used for map **latitude** offset.

This bug makes it hard to set the same offset to work with different editors.

See above.

2.26.2

Chrome

matkoniecz:
fixes #9792

(should I also change label to purge translations?)

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -847,7 +847,7 @@ en:
       description: Show Location Panel
       tooltip: Show coordinates and regional details.
     fix_misalignment: Imagery Offset
-    offset: "Drag anywhere in the gray area below to adjust the imagery offset, or enter the offset values in meters."
+    offset: "Drag anywhere in the gray area below to adjust the imagery offset, or enter the offset values."
     offset_label: "Adjust Imagery Offset"
     nudge:
       top: "nudge top"


matkoniecz:
screen-2026-06-29-15-33-09

PR #12511, opened 2026-06-30
Avoid splitting surrogate pairs when shortening address labels

mahirhir:
The address-label shortening loop in `modules/svg/labels.js` trims one character at a time using `substring` with a UTF-16 `.length` count:

```js
name = `${name.substring(0, name.replace(/…$/, '').length - 1)}…`;
```

`.length` counts UTF-16 code units, so `length - 1` can land in the middle of a surrogate pair. When the label ends in a character outside the BMP, that character is two code units, and `substring` keeps only the first half. The result is a lone surrogate that renders as `�`.

This is reachable for real address points. `utilDisplayName` resolves the label from `name`, then `addr:housename`, then the house number, and both `name` and `addr:housename` are free text. A house or building name ending in an emoji, or in a CJK Extension B ideograph such as 𠮷 (U+20BB7, which appears in Japanese names), hits this case once the label is wide enough to be trimmed.

The fix uses `Array.from(...).slice(0, -1)` to drop the last code point instead, the same approach `utilUnicodeCharsTruncated` already takes in `modules/util/util.ts`. Strings within the BMP, including all BMP CJK, are unchanged.

Checked with node:

```
"ビル𠮷"      old: "ビル\uD842…"  (lone surrogate)   new: "ビル…"
"漢字テスト"   old: "漢字テス…"                        new: "漢字テス…"  (unchanged)
```

I did not add a test, since this runs inside the canvas label-rendering loop. `utilUnicodeCharsTruncated`, which uses the same technique, is covered in `test/spec/util/util.js`.


matkoniecz:
> the same approach utilUnicodeCharsTruncated already takes in modules/util/util.ts

can `utilUnicodeCharsTruncated` be used here directly?

mahirhir:
Not as a drop-in for the whole loop — that block shortens the label one step at a time by measured pixel width (textWidth) until it fits under 36, not to a fixed character count, so utilUnicodeCharsTruncated can't replace the loop itself. But it is the right tool for the per-step truncation, which is where the surrogate split happens. I can change the line to: `var base = name.replace(/…$/, ''); name = utilUnicodeCharsTruncated(base, Array.from(base).length - 1) + '…';` That reuses the helper and drops a whole code point each iteration instead of a UTF-16 unit. Want me to update the PR to that?

modules/svg/labels.js
```diff
@@ -344,7 +344,7 @@ export function svgLabels(projection, context) {
                     var renderAs = renderNodeAs[entity.id];
                     if (renderAs.geometry === 'vertex' && zoom < 17) continue;
                     while (renderAs.isAddr && width > 36) {
-                        name = `${name.substring(0, name.replace(/…$/, '').length - 1)}…`;
+                        name = `${Array.from(name.replace(/…$/, '')).slice(0, -1).join('')}…`;
```
1ec5: > A house or building name ending in an emoji

The approach you took is probably good enough, but FYI it would still potentially split an emoji sequence and some other situations involving combining characters. The most correct solution would be [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) with a `granularity` of `grapheme`.

tyrasd:
> Checked with node: […] "ビル𠮷"  

I am honestly having issues reproducing the issue. I tried a couple of different browsers (Firefox, Chrome, Edge, Safari) and operating systems (Linux, Windows, Mac), and none show the described behavior. I assume this is because on most default fonts the placeholder glyph for an invalid unicode character is as wide or wider then a CJK glyph. So, I guess we're kind of lucky that the naive substring approach still works in most cases. :shrug:

tyrasd (changes requested):
As pointed out above: reusing the existing method seems like a better approach here. I'd go with something like:

```js
                    let length = utilUnicodeCharsCount(name);
                    while (renderAs.isAddr && width > 32) {
                        name = utilUnicodeCharsTruncated(name.replace(/…$/, ''), length--) + '…';
```

That way we can improve the implementation of `utilUnicodeCharsTruncated` independently (e.g. to use `Intl.Segmenter`) at some point in the future if required.

tyrasd (approved):

PR #12523, opened 2026-07-03
move power points from point layer to power layer

### LINKED ISSUE #8100: Power poles and layers
If I turn on "Power Features" I can see all the power features, including power poles and towers, but to add one, I have to turn on "POI" layers, it's strange.

matkoniecz:
fixes #8100

### DIFF
--- a/modules/renderer/features.js
+++ b/modules/renderer/features.js
@@ -103,7 +103,7 @@ export function rendererFeatures(context) {
         100);
 
     defineRule('points', (tags, geometry) =>
-        geometry === 'point' && !isAddressPoint(tags, geometry) && tags.power !== 'no',
+        geometry === 'point' && !isAddressPoint(tags, geometry) && (!tags.power || tags.power === 'no'),
         200);
 
     defineRule('traffic_roads', function isTrafficRoad(tags) {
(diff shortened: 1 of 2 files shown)


matkoniecz (commented):
for some reason has no effect at all

matkoniecz:
actually, before I will change tests so they pass in LLM fashion...

should be point power features actually be removed from point POI layer?

tyrasd:
I would say so. It does seem only logical, and fixes usability issues such as #8100.

I guess an argument could be made that if we change/fix it for power features, we should treat other feature types consistently, e.g.:
* water point features like `natural=spring`
* highway point features like traffic signs
* railway point features like stations
* indoor point features
* and so on

matkoniecz:
> and fixes usability issues such as https://github.com/openstreetmap/iD/issues/8100.

I though about changing logic in https://github.com/openstreetmap/iD/blob/develop/modules/renderer/features.js#L497 instead

If I treat it correctly currently if something matches multiple groups and any group is hidden such preset hides

maybe instead if matches multiple groups then if any group is visible then preset continues to be visible?

Compare https://www.openstreetmap.org/edit?editor=id&node=11281393359#map=20/50.0975248/20.0250373 map data: power generator point remains visible even with all groups except power one being disable

but adding more is blocked

screen-2026-07-10-18-07-09

PR #12525, opened 2026-07-05
Use Oxford comma to align with US English

ghost:


matkoniecz (commented):
PR seems to be correct

sadly US really uses serial comma

tyrasd:
In the help texts we (mostly) use the oxford comma, but I do still think the omission in the tooltip of the map headings is so minor that it does not need to be changed. And it prevents additional unnecessary work of translators.

I'm happy to be convinced otherwise, but the given (or non-given) description in this PR doesn't suffice.

data/core.yaml
1ec5, line 1: Looks like we already have some Oxford commas in the codebase, so this would be more consistent:

https://github.com/openstreetmap/iD/blob/31fb8eca363d18f20d4c2918efec1b8251fe7c64/data/core.yaml#L1076

PR #12536, opened 2026-07-08
migrate actions to typescript (part 1)

k-yle:
I'll split these into several PRs, this first part is nice and easy

### DIFF
--- a/modules/core/graph.ts
+++ b/modules/core/graph.ts
@@ -1,4 +1,5 @@
 import { debug, type EntityId } from '../index';
+import type { NodeId, RelationId, WayId } from '../osm/id_manager';
 import type { osmNode } from '../osm/node';
 import type { osmRelation } from '../osm/relation';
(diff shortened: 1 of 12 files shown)


modules/core/graph.ts
```diff
@@ -31,10 +32,18 @@ export class coreGraph {
     this.frozen = !mutable;
   }
 
+    hasEntity(id: NodeId): osmNode | undefined;
+    hasEntity(id: WayId): osmWay | undefined;
+    hasEntity(id: RelationId): osmRelation | undefined;
```
k-yle, line 37: this is very handy, so that we don't need typecasts everywhere. it figures out the correct type based on the `EntityId`

tyrasd (approved):
a little annoying that git does not detect the file renames for the very short actions, but I guess those are trivial enough that the loss of linked history is not too relevant.

PR #12546, opened 2026-07-10
delete cross-referencing code, since it's now done at compile-time

### LINKED ISSUE #12094: Cease confusing and no longer needed and not documented implicit inheritance of `fields` and `moreFields` in presets
See https://github.com/ideditor/schema-builder/issues/204

See https://github.com/openstreetmap/id-tagging-schema/issues/1813

See https://github.com/openstreetmap/id-tagging-schema/pull/1855

iD now can stop implicit inheritance as it is no longer needed or useful

Benefits include: removal of code from iD, not using undocumented functionality, reduction of confusion.

k-yle:
all this code was moved to https://github.com/ideditor/schema-builder/pull/281, so we can delete it from the runtime code


tyrasd (approved):
merci!

Issue #12531, opened 2026-07-07
Link wikidata when NSI proposes to add brand:wikidata=


matkoniecz:
to test: open say https://ideditor.netlify.app/#background=Bing&disable_features=boundaries&map=20.98/38.74519/-75.15476&id=w488118123 and remove `brand:wikidata`

iD will encourage user to add `brand:wikidata` without making clear what is the meaning of this alphanumeric pattern

I think that relevant Wikidata page should be linked to get at least chance for user to verify claimed match (sometimes NSI is wrong!)

related to #6517

EDIT:

<img width="379" height="714" alt="Image" src="https://github.com/user-attachments/assets/b6dc26aa-8a91-46d9-b6dc-671d57453e2b" />

Gaurav775-Git:
Hi @matkoniecz  I understand the issue - users are being asked to add brand:wikidata tags without understanding what they mean. I'd like to help by displaying a human-readable label with a link to the Wikidata page. Before I start, could you point me to where in the codebase the Wikidata suggestions are rendered? I've found the modules/ui/presets/ directory but want to make sure I'm looking in the right place.

matkoniecz:
I have no idea which part of code should be modified and how (If I would know I would probably make PR myself)

Gaurav775-Git:
OK @matkoniecz  Thanks for the reply. I'll dig into the code and try to find where these Wikidata suggestions are coming from. I'll keep you posted once I figure out the approach.

Gaurav775-Git:
I looked into this. When you type Q5122694 into wikidata.org/wiki/Q5122694, it shows "Citizens Financial Group" - so the code does link to real info.

The problem is users just see the random code and have no idea what it means or if it's correct.

I think we should:


approach-1 

( Fetch the brand name from Wikidata

Show it with the code ("Citizens Financial Group (Q5122694)")

Make it clickable to the Wikidata page

Or add a small button next to the code that says "Verify" or "View on Wikidata" - clicking it would open the link directly)

<img width="1318" height="871" alt="Image" src="https://github.com/user-attachments/assets/0cabb957-5756-4f41-bcaf-87351bf6b56e" />

<img width="1846" height="877" alt="Image" src="https://github.com/user-attachments/assets/31f19ece-796b-4830-bf25-918cf3aa7ce6" />

approach-2 

(add one redirect link under the link "view document' to make it verify directly by pressing on that 'verify link')

[...]

PR #12551, opened 2026-07-11
Linkifying the qId of wikidata on sidepanel

Gaurav775-Git:
image

This is the ui implementation of link placement in other part .

https://github.com/user-attachments/assets/997d5560-928f-430f-9822-0c785cb4b143

and this is how it works.

I have done changes in raw_tag_editor.js and in outdated_tags.js to implement this .

### DIFF
--- a/modules/actions/add_entity.js
+++ b/modules/actions/add_entity.js
@@ -1,5 +1,0 @@
-export function actionAddEntity(way) {
-    return function(graph) {
-        return graph.replace(way);
-    };
-}
--- a/modules/actions/add_entity.ts
+++ b/modules/actions/add_entity.ts
@@ -1,0 +1,8 @@
+import type { Action } from '../core/history';
+import type { OsmEntity } from '../osm/abstract-entity';
+
+export function actionAddEntity(way: OsmEntity): Action {
+    return function (graph) {
+        return graph.replace(way);
+    };
+}
--- a/modules/actions/add_member.ts
+++ b/modules/actions/add_member.ts
@@ -1,8 +1,13 @@
+import type { coreGraph } from '../core/graph';
+import type { Action } from '../core/history';
+import type { osmWay, RelationId } from '../osm';
 import { osmJoinWays } from '../osm/multipolygon';
+import type { osmRelation, RelationMember } from '../osm/relation';
 import { utilArrayGroupBy, utilObjectOmit } from '../util';
 
+interface MemberWithIndex extends RelationMember { index?: number };
 
-export function actionAddMember(relationId, member, memberIndex) {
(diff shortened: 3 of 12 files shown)


modules/ui/sections/raw_tag_editor.js
```diff
@@ -278,6 +278,24 @@ export function uiSectionRawTagEditor(id, context) {
...
+                if (d.key && d.key.endsWith(':wikidata') && val && /^Q\d+$/.test(val)) {
+                    wrap.append('a')
+                        .attr('class', 'wikidata-link')
+                        .attr('href', 'https://www.wikidata.org/wiki/' + val)
+                        .attr('target', '_blank')
+                        .attr('rel', 'noopener')
+                        .attr('title', 'wikidata.org/wiki/' + val)
+                        .call(svgIcon('#iD-icon-out-link'));
+                }
+            });
```
1ec5: As explained in https://github.com/openstreetmap/iD/issues/12531#issuecomment-4924348929, the change you’re implementing here is unrelated to #12531 and was previously rejected.

modules/validations/outdated_tags.js
```diff
@@ -272,5 +273,22 @@ export function showTagDiffReference(selection, reference, tagDiff) {
...
+            el.append('a')
+                .attr('class', 'tagDiff-wikidata-link')
+                .attr('href', 'https://www.wikidata.org/wiki/' + qid)
+                .attr('target', '_blank')
+                .attr('rel', 'noopener')
+                .text(qid);
+        } else {
+            el.text(d.display);
+        }
+    });
```
1ec5: Can we get a screen recording of _this_ feature? Thanks!

Gaurav775-Git: as clearly we can see the redirect link placement in between the delete and the drop down arrow.
And it make's the interaction better as before the QID was not making any sense but this can make it to do action .
And the position of the link is also different form the previous issue .

https://github.com/user-attachments/assets/eab0f1f0-c0ec-437b-85fd-fbc3d0eb691a

1ec5: Your screen recording does not show anything related to this file, or to the feature requested in #12531. Instead, it shows the changes you made to raw_tag_editor.js.

modules/ui/sections/raw_tag_editor.js
matkoniecz: I thought description was clear, just in case I added now screenshot of relevant interface

modules/validations/outdated_tags.js
Gaurav775-Git: okay , got it and here is the recording of that feature .


https://github.com/user-attachments/assets/a3e855d8-6090-41ad-b461-e9a22f06ef18

PR #12558, opened 2026-07-13
migrate actions to typescript (part 2)

k-yle:
this one is fairly straightforward too, and the diff is pretty minimal


modules/actions/disconnect.ts
```diff
@@ -49,7 +56,7 @@ export function actionDisconnect(nodeId, newNodeId) {
     };
 
 
-    action.connections = function(graph) {
+    function getConnections(graph: coreGraph) {
```
k-yle, line 59: this does not need to a public API, nothing else is using it

tyrasd (approved):

PR #12559, opened 2026-07-13
migrate actions to typescript (part 3)

k-yle:
this one is also very simple

### DIFF
--- a/modules/actions/reverse.ts
+++ b/modules/actions/reverse.ts
@@ -1,6 +1,17 @@
+import type { coreGraph } from '../core';
+import type { Action } from '../core/history';
+import type { EntityId, OsmEntity, osmWay } from '../osm';
 import { presetManager } from '../presets';
 
-/*
+export interface ActionReverse extends Action {
+    entityID(): EntityId;
+}
+
+export interface ReverseOptions {
+    reverseOneway?: boolean;
+}
+
+/**
 Order the nodes of a way in reverse order and reverse any direction dependent tags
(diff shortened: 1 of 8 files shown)


modules/actions/straighten_nodes.ts
```diff
@@ -28,8 +32,8 @@ export function actionStraightenNodes(nodeIDs, projection) {
     }
 
 
-    var action = function(graph, t) {
-        if (t === null || !isFinite(t)) t = 1;
+    const action: Action = function(graph, t) {
+        if (t === null || t === undefined || !isFinite(t)) t = 1;
```
k-yle, line 36: this has no impact because `!isFinite(t)` was already checking for `undefined`. it's just more explicit

modules/actions/reverse.ts
```diff
@@ -150,7 +161,8 @@ export function actionReverse(entityID, options) {
     }
 
     /** @returns {false | string} - returns false or the name of the direction key */
```
tyrasd:
```suggestion
    /** @returns returns false or the name of the direction key */
```

no need to duplicate the type in the docstring, is there?

tyrasd (commented):
Just for clarity, since we have the dedicated types for tag keys and values: why not use them (as in https://github.com/tyrasd/iD/commit/83aaf155be10059334c27176e1540f13401c3043)?

modules/actions/reverse.ts
k-yle: good point, fixed

tyrasd (approved):

PR #12560, opened 2026-07-13
migrate the svg renderer to typescript (part 2)

k-yle:
see inline comments

### DIFF
--- a/modules/globals.d.ts
+++ b/modules/globals.d.ts
@@ -47,6 +47,11 @@ declare global {
         // custom overload so that `Object.keys(Record<T, …>)` returns `T[]`
         keys<T>(o: T extends Record<infer K, unknown> ? [K] extends [string] ? T : never : never): (keyof T)[];
     }
+
+    interface ParentNode extends Node {
+        /** used internally by d3 to store the values passed to `.data()` */
+        __data__: any;
+    }
 }
 
 export {};
--- a/modules/osm/qa_item.ts
+++ b/modules/osm/qa_item.ts
@@ -1,5 +1,16 @@
+import type { Vec2 } from '../geo/vector';
+
+type Props = Omit<QAItem, 'loc' | 'service' | 'itemType' | 'id'>
+
 export class QAItem {
-  constructor(loc, service, itemType, id, props) {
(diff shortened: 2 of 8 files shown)


modules/osm/qa_item.ts
```diff
@@ -16,11 +27,11 @@ export class QAItem {
     }
   }
 
-  update(props) {
+  update(props: Props) {
     // You can't override this initial information
     const { loc, service, itemType, id } = this;
 
-    Object.keys(props).forEach(prop => this[prop] = props[prop]);
+    Object.assign(this, props);
```
k-yle, line 34: this is a runtime change, because typescript didn't like the old syntax. it behaves exactly the same

modules/svg/helpers.ts
```diff
@@ -89,6 +101,8 @@ export function svgMarkerSegments(projection, graph, dt, shouldReverse = () => f
             type: 'LineString',
             coordinates: coordinates
         }, projection.stream(clip({
+            polygonStart() {},
+            polygonEnd() {},
```
k-yle, line 105: [according to the d3 docs](https://d3-wiki.readthedocs.io/zh-cn/master/Geo-Streams/#:~:text=Stream%20listeners%20must%20implement%20several%20methods), we "must" implement all methods. hence the no-op function

modules/globals.d.ts
```diff
@@ -47,6 +47,11 @@ declare global {
         // custom overload so that `Object.keys(Record<T, …>)` returns `T[]`
         keys<T>(o: T extends Record<infer K, unknown> ? [K] extends [string] ? T : never : never): (keyof T)[];
     }
+
+    interface ParentNode extends Node {
```
tyrasd, line 51: this isn't used anywhere currently, is it? 🤔

k-yle: `ParentNode` is a built-in global type (like [`Node`](https://developer.mozilla.org/en-US/docs/Web/API/Node)), so this is just adding the `__data__` property to the existing built-in global. 

`__data__` itself is used in various files

tyrasd (approved):

Issue #10380, opened 2024-08-03
Add background version grouping


hlfan:
### Description

Add a system for grouping versions or variants of layers in a single line. This idea builds on facebook/Rapid#1388 and #8826, further decluttering the background selection.
While this feature is oriented towards individually indexed versions of background layers, it may also be used for variants of maps, similar to WMS layers.

### Screenshots

![Berlin](https://github.com/user-attachments/assets/6693fb78-85c5-4e2a-bf1f-425e84363404)
![Vorarlberg](https://github.com/user-attachments/assets/d806bd24-b4ec-49e0-a63b-dfaa2ae9d21b)

tordans:
There is a lot of discussion on this topic in https://github.com/openstreetmap/iD/pull/8826

Changes in this part of the UI have quite a few side effects that need to be considered.

---

On the idea of moving years to a dropdown: Personally, I prefer a one dimensional list of image layers which I can more easily scan and which allow for a faster change between layers.

1ec5:
> On the idea of moving years to a dropdown: Personally, I prefer a one dimensional list of image layers which I can more easily scan and which allow for a faster change between layers.

OpenHistoricalMap/iD#181 implemented an inline set of radio buttons for each Esri Wayback layer. It is very handy for flipping back and forth, but mappers have been complaining that the sheer number of options takes up too much space and makes non-Wayback imagery undiscoverable. It would get only worse once OHM adds additional temporal layers. OpenHistoricalMap/issues#806 tracks replacing the OHM implementation with Rapid’s as a stopgap. Long-term, I think something like the slider proposed in https://github.com/OpenHistoricalMap/iD/pull/181#issuecomment-2253177422 would be much more usable.

PR #12561, opened 2026-07-13
group background layers with multiple (e.g. yearly) variants

tyrasd:
Tries to automatically detect background layers that are just  together
let user select the version they want in a dropdown. Supports either what looks like a date (or date range, e.g. `2014/15`), or DTM/DSM for hillshade layers.

https://github.com/user-attachments/assets/70af85a5-6779-4666-8052-4c89c9792e96

UI-wise this would tie in well with #11780

some example regions:

 

 

 

### todo:

* [ ] sometimes the matched date part is including extra (punctuation and whitespace) characters which should probably be stripped in the selection:
   
* [ ] overly aggressive detection as in


tordans:
Quick note before I forget: The star icon probably has to change based on the selection
>image
> vs. 
> image

tordans:
The spacing looks a bit off in some details. Around the dropdown arrow it looks good (maybe even a few px too much), but some cases between text and dropdown look too narrow and there should be a space like on the right of the inner dropdown also on the left. Or something like this.

> image

tyrasd:
The margins are now hopefully a little better adjusted for my test system (i.e. Firefox). It might need further tweaking for other browsers/OS combinations to look good everywhere.

> The star icon probably has to change based on the selection

I thought so as well at the start, but it's not so staight-forward: hiding the star when the respective variant is not currently selected could be a little confusing, as we're also not hiding the start while other background sources are active. Maybe graying it out could work, but I'm wondering if it is still worth the effort and even beneficial for the following reasoning: When the most recent version from a particular provider is the _best_ available source, and I manually actively select an older version of the same provider (for whatever reason), it still is the best available (group of) layer(s) for me, otherwise I would have kept mapping with the most recent version. Does that make sense?

modules/ui/sections/background_list.js
```diff
@@ -159,14 +163,46 @@ export function uiSectionBackgroundList(context) {
         });
     }
 
+
     function drawListItems(layerList, type, change, filter) {
-        var sources = context.background()
+        const sources = [];
+        const dateLikeRegex = /(.+?\s\(?)((?:[-\/, 0-9]{2,}|DTM|DSM|DOM)+\s*)(\)?(?:\s|$).*)/;
```
k-yle:
```suggestion
        const dateLikeRegex = /(.+?\s\(?)((?:[-\/, 0-9]{2,}|DTM|DSM|DOM|DEM)+\s*)(\)?(?:\s|$).*)/;
```

&shy;+ `DEM` as well?

modules/ui/sections/background_list.js
```diff
@@ -159,14 +163,47 @@ export function uiSectionBackgroundList(context) {
         });
     }
 
+
     function drawListItems(layerList, type, change, filter) {
-        var sources = context.background()
+        const sources = [];
+        const dateLikeRegex = /(.+?\s\(?)((?:(?:\d{2,}|[-\/, ])+|DTM|DSM|DOM|DEM)+\s*)(\)?(?:\s|$).*)/;
```
hlfan, line 169: The current regex doesn't match DGM (for Gelände), and I think a somewhat looser check could go a long way in resolving #10380:
```suggestion
        const dateLikeRegex = /(.+?\s\(?)((?:(?:\d{2,}|[-\/, ()])+|\w{1,3}\b)+\s*)(\)?(?:\s|$).*)/;
```

PR #12567, opened 2026-07-14
migrate the extract action & operation to typescript

k-yle:
this one is a bit annoying. In #9816, i used the 2nd argument for the new parameter, but it should ideally have been the 3rd argument. 

We want all actions and operations to have the same method signature, so the 2nd argument must be reserved for _`t`_.

This PR:
 - moves `shiftKeyPressed` to the 3rd argument
 - converts both the action and the operation to typescript, in the same PR

I tested the extract option and confirmed that it still works. (1. via <kbd>E</kbd>, 2. via <kbd>↑ Shift</kbd> <kbd>E</kbd>, and 3. via the `mismatched_geometry` validator)

### DIFF
--- a/modules/core/localizer.js
+++ b/modules/core/localizer.js
@@ -33,7 +33,7 @@ export {
 
 /** @typedef {unknown} Translations */
 
-/** @typedef {any} TRenderLocalizedText */
+/** @typedef {any} LocalizedTextRenderer */
 
 //
 // coreLocalizer manages language and locale parameters including translated strings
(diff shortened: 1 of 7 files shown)


modules/core/localizer.js
```diff
@@ -16,6 +16,8 @@ export {
     _t as t
 };
 
+/** @typedef {any} TAppend */
```
tyrasd:
```suggestion
/** @typedef {any} LocalizedTextRenderer */
```

maybe?

k-yle: sure, done

tyrasd (approved):

Issue #12548, opened 2026-07-11
Remember left sidebar width


Oivo35:
### Description

The iD editor does not remember the width setting of the left sidebar, unlike other settings such as recently used presets, backgrounds or map features. It would be beneficial when working with long values or relation names.

### Screenshots

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/e147608e-671b-4a14-933b-5fb17c1d4e21" />

Gaurav775-Git:
@Oivo35 This is a great quality improvement! I'd like to work on this.

My plan:

1. Find where the sidebar resizing is handled in the UI code
2. Save the width to localStorage when the user drags the divider
3. Restore the saved width when the page loads
4. Follow the existing preference pattern used for other settings

Gaurav775-Git:
So ,I have found the code where the sidebar ui is present (modules/ui/sidebar.js) and having this issue .

<img width="883" height="153" alt="Image" src="https://github.com/user-attachments/assets/539ffd63-5c91-4a12-bcbe-608c82af2720" />
<img width="576" height="147" alt="Image" src="https://github.com/user-attachments/assets/cd4e03a2-fe08-4503-ae9f-e412fdea47cd" />

Default width: Line 45 — hardcoded as '33.3333%'
Min width: 240px (line 39)
Max width: 400px initially, expanded to 85% during drag (line 76)
DOM setup: modules/ui/init.js:141-144 creates the sidebar div and calls ui.sidebar

Now working on the implementation part to fix this .

Gaurav775-Git:
@tyrasd Done with the implementation and testing part , just have a look on this then I will raise the PR .

Before modifying the side_bar width feature.
https://github.com/user-attachments/assets/d91b2488-cc03-4a36-a39b-d1f165a5fce1

After modifying the side_bar width feature.
https://github.com/user-attachments/assets/9c6cb2a8-bf23-417b-bde4-5d731ca2e195

PR #12581, opened 2026-07-17
Feature/sidebar width (Remember left sidebar width  #12548)

Gaurav775-Git:
Remember left sidebar width
 #12548
The iD editor does not remember the width setting of the left sidebar, unlike other settings such as recently used presets, backgrounds or map features. It would be beneficial when working with long values or relation names.
so i made this to save size locally and work .

Before fix.

https://github.com/user-attachments/assets/a49f9664-9c4b-465e-8614-94bd6f382e78

After fix .
https://github.com/user-attachments/assets/13b73104-59c3-4756-b8a4-1473623d1fc4

### DIFF
--- a/modules/ui/sidebar.js
+++ b/modules/ui/sidebar.js
@@ -40,10 +40,11 @@ export function uiSidebar(context) {
         var dragOffset;
 
         // Set the initial width constraints
+        var savedWidth = prefs('sidebar-width');
         selection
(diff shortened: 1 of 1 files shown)


matkoniecz:
main problem here (like with other similar features) is that user cannot do anymore "I broke something, so I should reload page to get defaults back"

PR #12586, opened 2026-07-18
Fix child node selection of ways repeating node selection

RudyTheDev:
Per title.

Repro - draw a triangle (closed way) and select it:

before triangle

Selection is this:

before way

Do a Ctrl+Down and observe new selection:

before children

Note that there are 4 nodes selected instead of 3.

After the fix:

after children

I added unit tests for the child selection operation. Before/currently:

child-node-select-fix before

After fix:

child-node-select-fix after

Disclaimer: Used LLM to suggest unit test syntax. (Seems it used the same syntax as `add_point.js`.)


tyrasd (approved):
Thanks! I only changed it to reuse the existing `utilArrayUniq` method, which is IMHO slightly more readable than the array-to-set-to-array approach.

And thanks for the extensive unit tests, including for the existing behaviour! :+1:

tyrasd:
PS: I could have sworn that I've seen an issue for this, but cannot find it anymore; perhaps I'm just misremembering…

PR #12600, opened 2026-07-22
convert pannellum to typescript

k-yle:
this is nice and easy, we just need to install the @&shy;types package

### DIFF
--- a/modules/services/plane_photo.ts
+++ b/modules/services/plane_photo.ts
@@ -1,31 +1,34 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
-import { zoom as d3_zoom, zoomIdentity as d3_zoomIdentity } from 'd3-zoom';
+import { zoom as d3_zoom, zoomIdentity as d3_zoomIdentity, type D3ZoomEvent } from 'd3-zoom';
 import { utilSetTransform, utilRebind } from '../util';
+import type { PhotoFrame } from './pannellum_photo';
+import type { Vec2 } from '../geo/vector';
+import type { coreContext } from '../core';
 
 
-export async function planePhotoFrame(context, selection) {
+export async function planePhotoFrame(context: coreContext, selection: d3.Selection<HTMLDivElement>) {
     const dispatch = d3_dispatch('viewerChanged');
 
-    const module = {};
+    const module: PhotoFrame = function() {};
     module.event = utilRebind(module, dispatch, 'on');
 
     let _photo;
-    let _imageWrapper;
-    let _planeWrapper;
-    let _viewerDimensions = [];
-    let _photoDimensions = [];
-    const _imgZoom = d3_zoom()
+    let _imageWrapper: d3.Selection<HTMLDivElement>;
+    let _planeWrapper: d3.Selection<HTMLDivElement>;
+    let _viewerDimensions: Vec2 = [] as any;
+    let _photoDimensions: Vec2 = [] as any;
+    const _imgZoom = d3_zoom<HTMLDivElement, unknown>()
         .on('zoom', zoomPan)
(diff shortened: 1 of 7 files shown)


modules/services/plane_photo.ts
```diff
@@ -1,31 +1,33 @@
 import { dispatch as d3_dispatch } from 'd3-dispatch';
-import { zoom as d3_zoom, zoomIdentity as d3_zoomIdentity } from 'd3-zoom';
+import { zoom as d3_zoom, zoomIdentity as d3_zoomIdentity, type D3ZoomEvent } from 'd3-zoom';
 import { utilSetTransform, utilRebind } from '../util';
+import type { PhotoFrame } from './pannellum_photo';
+import type { Vec2 } from '../geo/vector';
 
 
-export async function planePhotoFrame(context, selection) {
+export async function planePhotoFrame(context: iD.Context, selection: d3.Selection) {
```
tyrasd: I think we should avoid the `any`-typed selections like this one. For example, this hides that on line 54, `_imgZoom.transform` can only be used on `ZoomedElementBaseType` selections (for whatever reason). In this particular case, it is a bit silly as that is only an alias to `Element`, but in other cases it might actually a requirement that make sense and should be enforced by the type checker.

Maybe we should change `d3.Selection` to default to `Selection<T = HTMLElement>` as that is what we most often start from in the UI code of iD, and then manually adjust to the respective narrower (or wider) types if needed: here, e.g. `let _planeWrapper: d3.Selection<Element>;` + later on `_planeWrapper = selection.append<Element>('div')`?

PS: After fiddling around a bit, I think we really should change the `Selection` to default to `HTMLElement`, but as that will require a cascade of followup-fixes unrelated to this PR, let's do that in a separate step and leave it here as e.g.:

```suggestion
export async function planePhotoFrame(context: iD.Context, selection: d3.Selection<HTMLElement>) {
```

modules/services/plane_photo.ts
```diff
@@ -99,7 +101,7 @@ export async function planePhotoFrame(context, selection) {
 
     /**
      * Hides the photo frame if shown
-     * @param {*} context the HTML wrap of the frame
+     * @param context the HTML wrap of the frame
```
tyrasd: `context` here is quite confusing. maybe we should just change it to `selection` as everywhere else? or at least `viewerContext` as in `pannellum_photo.ts`…

modules/services/plane_photo.ts
```diff
@@ -1,31 +1,33 @@
...
-    function zoomPan(d3_event) {
+    function zoomPan(d3_event: D3ZoomEvent<Element, unknown>) {
         let t = d3_event.transform;
         _imageWrapper.call(utilSetTransform, t.x, t.y, t.k);
     }
 
-    function loadImage(selection, path) {
-        return new Promise((resolve) => {
+    function loadImage(selection: d3.Selection, path: string) {
+        return new Promise<d3.Selection>((resolve) => {
```
tyrasd:
```suggestion
    function loadImage(selection: d3.Selection<HTMLImageElement>, path: string) {
        return new Promise<d3.Selection<HTMLImageElement>>((resolve) => {
```

then `naturalWidth/Height` are safely accessible from the `selection.node()!` on line 127.

tyrasd (dismissed):
See below. I'm sorry for being so pedantic with the d3-related typing, but I do really think that without precise typing of our selections, we do throw out half of the benefits typescripts gives us when it comes to type-safety of the front-end code. Let's invest this little bit of extra time and effort to get this right from the start, shall we?

k-yle:
sounds good, shall we proceed with #12603 first?

k-yle:
suggestions applied, and rebased after #12603

tyrasd (approved):

PR #12599, opened 2026-07-22
convert localizer to typescript

k-yle:
see inline comments

PR #12601, opened 2026-07-22
convert localizer to typescript class

tyrasd:
continues #12599 with some additional improvements (I think)

* typescript class instead of "iD" module (might not really be any better, but it was easy to do while at it and looks a bit cleaner IMHO)
* even more precise typing, requiring less use of manual casting (e.g. `as …`)
* rectifies largely inconsistent indentation across the file (some methods used 2 spaces, others 4). Hiding whitespace-only changes will be useful when reviewing :wink: 
* addresses https://github.com/openstreetmap/iD/pull/12599/changes#r3630879989 by replacing the "manual" wildcard index (`'locales_index_*'`) with template literal types for the locale index and translation data files (see `LocaleIndexKey`, `LocaleDataKey`).
* fixes a few minor edge cases that were previously not covered and other minor additional improvements

@k-yle: can you please give this a look and let me know what you think?

I've set the base branch of this PR to `kh/localizer` for now to facilitate easier review, but plan to eventually merge it directly into develop together with #12599 of course.


modules/presets/field.js
```diff
@@ -22,7 +22,7 @@ export function presetField(fieldID, field) {
   };
 
   _this.t = (scope, options) => t(`_tagging.presets.fields.${fieldID}.${scope}`, options);
-  _this.t.all = (scope, options) => t.all(`_tagging.presets.fields.${fieldID}.${scope}`, options);
+  _this.t_all = (scope, options) => localizer.t_all(`_tagging.presets.fields.${fieldID}.${scope}`, options);
```
tyrasd, line 25: I changed this one (and the `t.html` variant) from the direct exported method to having to use it via the `localizer` because it is relatively obscure, not [documented](https://github.com/openstreetmap/iD/wiki/How-to-Use-iD's-Localizer-Module) and should not really be used outside of the presets implementation.

modules/core/localizer.ts
```diff
@@ -1,187 +1,220 @@
...
-    [id: string]: string | string[] | d3.Selector | undefined;
-}
+type Replacements = {
+    [key: string]: number | string | d3.Selector | undefined
+} & {
+    default?: string,
+    prefix?: string,
+    suffix?: string
+};
+type ReplacementsSimple = {
```
tyrasd, line 57: dedicated argument type for the different variants of the `t.*` methods: for example `t()` does not accept a selector function, while for the `t_all` the `default` attribute must be an array instead of a single string.

modules/core/localizer.ts
```diff
@@ -1,187 +1,220 @@
...
+    [key: string]: number | string | undefined
+} & {
+    default?: string
```
tyrasd, line 60: also add `default` (and `prefix` / `suffix` above) as their own special properties in the Replacements dicts

modules/core/localizer.ts
```diff
@@ -1,187 +1,220 @@
...
-        _loadPromise = undefined;
-        return localizer;
-    };
+        this._loadPromise = undefined;
+        return this;
+    }
 
 
-    var _loadPromise: Promise<void> | undefined;
+    private _loadPromise?: Promise<this>;
```
tyrasd, line 156: this allows returning `this` in the `ensureLoaded` promise to directly work with the object. Not really needed here as the class is always used as a singleton (outside of tests), but it would make sense to introduce this across all modules with an async initializations.

modules/core/localizer.ts
```diff
@@ -1,187 +1,220 @@
...
 
-        return _loadPromise = Promise.all(filesToFetch.map(key => fileFetcher.get(key)))
-            .then(results => {
-                _dataLanguages = results[0] as LanguagesJSON;
-                _dataLocales = results[1] as LocalesJSON;
+        return this._loadPromise = Promise.all([
+                fileFetcher.get('languages'),
+                fileFetcher.get('locales'),
+                ...filesToFetch.map(key => fileFetcher.get(key))
+            ]).then(([languages, locales, ...indexes]) => {
```
tyrasd, line 182: separating the languages/locales files from the `filesToFetch` list, the ts typing is even more clever: `languages`, `locales` and `indexes` all arrive correctly typed from the result of the `Promise.all` here. 🤓

modules/core/localizer.ts
```diff
@@ -1,187 +1,220 @@
...
             })
             .then(() => {
-                updateForCurrentLocale();
+                this._updateForCurrentLocale();
+                return this;
             })
-            .catch(err => console.error(err));  // eslint-disable-line
+            .catch(err => {
+                console.error(err);  // eslint-disable-line
+                throw err;
```
tyrasd, line 212: Not actually "consuming" the exception here would allow users of the module to implement their own fallback/fail logic. again (as above): probably not needed here, but could be a useful pattern for other modules.

modules/core/localizer.ts
```diff
@@ -1,187 +1,220 @@
...
+            });
     };
 
     // Returns the locales from `requestedLocales` supported by iD that we should use
-    localizer.localesToUseFrom = (supportedLocales: { [locale: string]: unknown }) => {
+    localesToUseFrom(supportedLocales: Record<string, any>) {
```
tyrasd: I was pondering much too long whether `unknown` or `any` is better here. 🙈 From the wording, I like `any` (as in _anything_) better, but maybe there is an argument to prefer `unknown`, which I can't think of right now?

modules/core/localizer.ts
```diff
@@ -1,187 +1,222 @@
...
+        pct: number;
+    }
+};
+
+type TranslationStrings = { [key: string]: any | TranslationStrings };
+export type Translations = {
+    [key: string]: TranslationStrings,
+} & {
+    languageNames?: Record<string, string>,
+    scriptNames?: Record<string, string>
```
tyrasd, line 32: add separate type for the properties on the root level of the translation files

modules/core/localizer.ts
```diff
@@ -1,187 +1,222 @@
...
+export type LocalesJSON = {
+    [localeCode: string]: {
+        /** right-to-left or left-to-right text direction */
+        rtl: boolean;
+        /** the percent of strings translated; 1 = 100%, full coverage */
+        pct: number;
+    }
+};
+
+type TranslationStrings = { [key: string]: any | TranslationStrings };
```
tyrasd, line 27: changed this from `[key: string]: string` because for presets this can also be a `string[]`, or generally, a `Record` when the text has plural forms. I tried to do something clever with union types, but it did not make it easier where it is actually consumed, so I left it as `any` for now. Let me know if you know a way to make it more elegant.

modules/core/localizer.ts
```diff
@@ -318,104 +352,105 @@ export function coreLocalizer() {
...
         if (stringsKey.toLowerCase() === 'en-us') stringsKey = 'en';
-        let localeString: any = _localeStrings && _localeStrings[scopeId] && _localeStrings[scopeId][stringsKey];
 
-        while (localeString !== undefined && path.length) {
-            localeString = localeString[path.pop()!];
+        let localeStringTree: TranslationStrings = this._localeStrings && this._localeStrings[scopeId] && this._localeStrings[scopeId][stringsKey];
+        while (localeStringTree !== undefined && path.length > 1) {
+            localeStringTree = localeStringTree[path.pop()!];
         }
+        let localeString = localeStringTree?.[path.pop()!];
```
tyrasd, line 360: Here, I separated the tree-walking from assigning the final value to `localeString` to avoid having to use the `any` type for the result. It end up being of type `any` anyway (see comment on line 27 above). 🙃 But at least the tree-walking itself can be done typesafe this way.

modules/core/localizer.ts
```diff
@@ -318,104 +352,105 @@ export function coreLocalizer() {
...
-                        // format numbers for the locale
-                        return value.toLocaleString(locale, {
-                          style: 'decimal',
-                          useGrouping: true,
-                          minimumFractionDigits: 0
+                if (typeof localeString === 'string') {
+                    let parts: (string | { key: string })[] = [localeString];
+                    for (let key in replacements) {
+                        const token = `{${key}}`;
+                        // replace every instance of `{key}` with a placeholder object {key: 'key'}
```
tyrasd, line 394: No idea why this used the `regex` to split the string originally (it's my code, but cannot remember any reason for it, lol). Simplified it now (and should also be more robust in case we ever want to use replacement tokens with not-regex-safe characters.

modules/core/localizer.ts
```diff
@@ -318,104 +352,105 @@ export function coreLocalizer() {
...
+                            }
+                        }
+
+                        return value;
+                      }
 
-                  return part;
-                });
+                    return part;
+                  }).filter(Boolean);
```
tyrasd, line 425: I don't think this will ever be triggered, but it's in principle a fail-safe against the case where a replacement object has a key set explicitly to `undefined`. 🤷

modules/core/localizer.ts
```diff
@@ -587,14 +625,14 @@ export function coreLocalizer() {
      * Returns a function that formats a floating-point number in the given
      * locale.
      */
-    localizer.floatFormatter = (locale: string) => {
+    floatFormatter(locale?: string) {
         if (!('Intl' in window && 'NumberFormat' in Intl &&
               'formatToParts' in Intl.NumberFormat.prototype)) {
-            return (number: number, fractionDigits: number) => {
+            return (number: number, fractionDigits?: number) => {
```
tyrasd, line 631: `fractionDigits` are optional, see line below

modules/core/localizer.ts
```diff
@@ -660,6 +698,15 @@ export function coreLocalizer() {
             return parts && parts[1] && parts[1].length || 0;
         };
     };
-
-    return localizer;
 }
+
+const _mainLocalizer = new coreLocalizer(); // singleton
+const _t = (stringId: string, replacements?: ReplacementsSimple, locale?: string) => _mainLocalizer.t(stringId, replacements, locale);
```
tyrasd, line 704: Because of the ts class, it is now slightly more annoying to provide those exports in their previous form (for backwards compatibility), but it works luckily

modules/ui/fields/check.ts
```diff
@@ -24,10 +24,10 @@ export function uiFieldCheck(field: any, context: iD.Context) {
 
     let _tags: TagsMulti;
 
-    let input:    d3.Selection<HTMLInputElement>  | d3.Selection<null> = d3_select(null);
-    let text:     d3.Selection<HTMLSpanElement>   | d3.Selection<null> = d3_select(null);
-    let label:    d3.Selection<HTMLLabelElement>  | d3.Selection<null> = d3_select(null);
-    let reverser: d3.Selection<HTMLButtonElement> | d3.Selection<null> = d3_select(null);
+    let input:    d3.Selection<HTMLInputElement>  = d3_select(null!);
```
tyrasd: `d3_select(null)` always returns an empty selection, which is compatible with any `d3.Selection<T>`, so using `null!` seems to be the more elegant way to model this in ts (as `d3.Selection<null>` does not really make sense conceptually).

modules/ui/intro/building.js
```diff
@@ -165,12 +165,12 @@ export function uiIntroBuilding(context, reveal) {
         var onClick = function() { continueTo(addHouse); };
 
         revealHouse(house, helpHtml('intro.buildings.retry_building'),
-            { buttonText: t.html('intro.ok'), buttonCallback: onClick }
+            { buttonText: localizer.t_html('intro.ok'), buttonCallback: onClick }
```
tyrasd, line 168: As above: `t.html` is heavily discouraged (at least outside of use within a markdown sandbox like here), so I removed it from the convenience direct export.

The rest of the files from here until the `test` folder are only replacements of `t.html` if I remember correctly.

modules/globals.d.ts
```diff
@@ -40,7 +40,7 @@ declare global {
       unknown
     >;
 
-    export type Selector = (selection: Selection) => void;
+    export type Selector = <T extends HTMLElement>(selection: Selection<T>) => void;
```
tyrasd, line 43: this makes sure, it can only be used on a properly typed d3 selection, and avoids another implicit fallback to `any` for the omitted generic type of the `Selection` supplied to the e.g. `t.append` method.

modules/core/localizer.ts
```diff
@@ -208,96 +243,95 @@ export function coreLocalizer() {
...
+
+        this._usesMetric = this._localeCode.slice(-3).toLowerCase() !== '-us';
     }
 
 
     /* Locales */
     // Returns a Promise to load the strings for the requested locale
-    localizer.loadLocale = (locale: string, scopeId: string, directory?: string) => {
-
+    loadLocale(locale: string, scopeId: string, directory: string) {
```
tyrasd, line 283: I don't think we should actually allow `directory` to be omitted, as that would only potentially lead to strange things (like loading a file from `undefined/en.min.json`). I adjusted the test case instead.

tyrasd (commented):
See notes inline below. PS: Some of the comments are two or three lines off it seems like – I had force-pushed a small change while still having the partial review pending. I hope it is still clear what the comments refer to.

modules/ui/fields/check.ts
k-yle:
```suggestion
    let input  = d3_select<HTMLInputElement, any>(null!);
```

not that it matters, but this is slightly more concise

modules/core/localizer.ts
k-yle: `unknown` is usually better, because it forces you to replace it with a stricter defintion if you ever need to access the value ([example](https://www.typescriptlang.org/play/?#code/DYUwLgBAhghAXBASiAxgewE4BMA8BnMDASwDsBzAGmhIE8A+AbgCgoA6EADygFsAHUVmDQBlQqTIAKAJTMmoSAFd4SVJlwFi5KgpIBrEmgDuJRkwXsufAUNGbJMoA))

k-yle (approved):
nice, this solves a lot of the problems that i couldn't figure out!

PR #12603, opened 2026-07-22
change default type of d3.Selection to HTMLElement

tyrasd:
See https://github.com/openstreetmap/iD/pull/12600#discussion_r3632482162 for rationale / benefits.

An alternative would have been to use `BaseType` as the fallback instead of `HTMLElement`, but I think this is more useful in most cases, and whenever we do use a selection of something else (e.g. of `document`, `EnterElement`, etc.), it makes sense to state them explicitly.

(blocked for now by #12601 / #12599, ~~and #12600 to avoid potential conflicts~~)

### DIFF
--- a/modules/svg/geolocate.ts
+++ b/modules/svg/geolocate.ts
@@ -11,7 +11,7 @@ import type { SvgLayer } from './layers';
 export type WithLoc = Pick<osmNode, 'loc'>;
 
 export function svgGeolocate(projection: Projection): SvgLayer {
-    var layer = d3_select<SVGGElement, 0>(null!);
+    var layer: d3.Selection<SVGGElement> = d3_select(null!);
     var _position: GeolocationPosition;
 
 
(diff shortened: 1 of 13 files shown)


modules/svg/geolocate.ts
```diff
@@ -11,7 +11,7 @@ import type { SvgLayer } from './layers';
 export type WithLoc = Pick<osmNode, 'loc'>;
 
 export function svgGeolocate(projection: Projection): SvgLayer {
-    var layer = d3_select<SVGGElement, 0>(null!);
+    var layer: d3.Selection<SVGGElement> = d3_select(null!);
```
k-yle, line 14: i tried to a find an alternative solution that works with this:

```suggestion
    var layer = d3_select<SVGGElement, 0>(null!);
```

but i couldn't figure it out

k-yle (approved):

PR #12604, opened 2026-07-23
Add a split tool for touch editing

koharachan:
## Summary
- add a dedicated split tool to the top toolbar for touch and pen editing
- show it only while split is available in select mode on touch/mobile contexts
- keep touch feedback stable even if the operation mutates selection state while running

## Testing
- npm run test:once -- test/spec/ui/tools/split.js
- npx eslint modules/ui/tools/split.js modules/ui/top_toolbar.js modules/ui/tools/index.ts test/spec/ui/tools/split.js
- git diff --check

### DIFF
--- a/modules/ui/tools/index.ts
+++ b/modules/ui/tools/index.ts
@@ -2,4 +2,5 @@ export * from './modes';
 export * from './notes';
 export * from './save';
 export * from './sidebar_toggle';
+export * from './split';
 export * from './undo_redo';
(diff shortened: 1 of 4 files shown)


matkoniecz:
> add a dedicated split tool to the top toolbar for touch and pen editing

please post screenshot of this in action

koharachan:
图片
图片
图片

matkoniecz:
Is action menu unavailable on touch devices? If yes, we should fix THAT, rather than giving access to only one specific action.

matkoniecz:
Iff action menu is unavailable or its access is hidden on touch devices, then maybe putting there this right-click menu would make sense?

matkoniecz:
@koharachan

PR #12606, opened 2026-07-23
Extract address tags from address-only buildings

### LINKED ISSUE #12447: Extract command to remove address from building and create an address node
### Description

In many countries, building areas are mapped with address tags added directly to the area, except in cases where a building has more than one address for different sections or entrances.

Currently when trying to use the Extract command on a building with only address tags and no business/POI info, nothing happens. I think it would be beneficial in this case for the command to remove any "addr:" tags from the building and add these tags to a new address node.

This is useful in situations where buildings were previously tagged with on address but are later found to have multiple addresses, saving the time of manually deleting address tags and creating a new node with these tags.

### Screenshots

A building currently has address tags directly on the area:

Image

Pressing "E" now moves (doesn't copy) the address tags to a node if no other POI-related tags (like amenity, shop, craft, office, etc.) are present...

Image

...making it quicker to add other address nodes to the building

Image

koharachan:
Generated using CHATGPT 5.5, with 70% completion rate.
Closes #12447.\n\nThis updates the Extract operation so address-only buildings can be extracted into address nodes. For buildings that also have point/POI tags, the existing behavior is preserved and address tags remain on both features.\n\nTests run:\n- npm run build:data\n- npm run build:js\n- npm run test:typecheck\n- npx eslint modules/actions/extract.ts modules/operations/extract.ts test/spec/actions/extract.js test/spec/operations/extract.js\n- npx vitest run --no-isolate test/spec/actions/extract.js test/spec/operations/extract.js

### DIFF
--- a/modules/actions/extract.ts
+++ b/modules/actions/extract.ts
@@ -10,6 +10,44 @@ export interface ActionExtract extends Action<boolean> {
     getExtractedNodeID(): NodeId;
 }
 
+const keysToCopyAndRetain = ['source', 'wheelchair'];
+const keysToRetain = ['area'];
+const buildingKeysToRetain = ['architect', 'building', 'height', 'layer', 'nycdoitt:bin', 'ref:GB:uprn', 'ref:linz:building_id'];
+
+function isAddressKey(key: string) {
+    return /^addr:.{1,}/.test(key);
+}
+
+function isBuildingKey(key: string) {
+    return buildingKeysToRetain.indexOf(key) !== -1 ||
+        /^building:.{1,}/.test(key) ||
+        /^roof:.{1,}/.test(key);
+}
+
+export function actionExtractMovesAddressTags(entity: OsmEntity) {
+    var isBuilding = (entity.tags.building && entity.tags.building !== 'no') ||
+        (entity.tags['building:part'] && entity.tags['building:part'] !== 'no');
+
+    if (!isBuilding) return false;
+
+    var hasAddressTags = false;
+    for (var key in entity.tags) {
+        if (entity.type === 'relation' && key === 'type') continue;
+        if (keysToRetain.indexOf(key) !== -1) continue;
+        if (keysToCopyAndRetain.indexOf(key) !== -1) continue;
+        if (isBuildingKey(key)) continue;
+
+        if (isAddressKey(key)) {
+            hasAddressTags = true;
+            continue;
+        }
+
+        return false;
+    }
+
+    return hasAddressTags;
+}
+
 export function actionExtract(entityID: NodeId, projection: Projection): ActionExtract {
 
     var extractedNodeID: NodeId;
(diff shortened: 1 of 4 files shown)


matkoniecz:
Is LLM generation involved here in some way? If yes, in which way - and in future please disclose it. If no: how you ended with newlines written as \n ?

Have you tested this change? Please post before/after screenshots.

koharachan:
yes,test photo Please wait a moment.

koharachan:
图片
图片

koharachan:
ChatGPT worked on analyzing issues, dealing with relevant documents, and initiating PRs. I used both the ChatGPT app and GPT 5.5.

matkoniecz:
Then, please, add clear disclosure at the beginning of this PR and other making clear that it is LLM-generated content and to which degree it is

koharachan:
ok

modules/actions/extract.ts
```diff
@@ -10,6 +10,44 @@ export interface ActionExtract extends Action<boolean> {
     getExtractedNodeID(): NodeId;
 }
 
+const keysToCopyAndRetain = ['source', 'wheelchair'];
+const keysToRetain = ['area'];
+const buildingKeysToRetain = ['architect', 'building', 'height', 'layer', 'nycdoitt:bin', 'ref:GB:uprn', 'ref:linz:building_id'];
+
+function isAddressKey(key: string) {
+    return /^addr:.{1,}/.test(key);
```
k-yle, line 18: if we are trying to define _'all address tags'_, then we also need to consider tags like `source:addr` and some `ref:*` tags.

We already have exceptions like this for buildings (for example `nycdoitt:bin`). 

Based on a quick [check on taginfo](https://taginfo.openstreetmap.org/keys/addr%3Ahousenumber#combinations), we would at least need to add `nysgissam:nysaddresspointid`, `ref:ruian:addr`, `ref:linz:address_id`, `ref:RS:kucni_broj`). I only checked tags with >1 million uses, there are likely others.

k-yle: i also noticed that we already hardcode tags like `nysgissam:nysaddresspointid` in a different part of the codebase. 

ideally this logic would be unified, since we're currently missing some tags like `osak:identifier`, which results in a worse user expierence [in denmark](https://kyle.kiwi/iD/#notes=ja&map=18.60/54.83605/9.36856&disable_features=boundaries&background=Geodatastyrelsen_Denmark)

koharachan: @k-yle Thanks for taking the time to fix this directly! I've reviewed the changes in "Files changed" and they look good to me.

Also, I've already added the LLM disclosure to the PR description as requested by @matkoniecz.

Let me know if there's anything else I need to do on my side before merging.

matkoniecz:
How it behaves and should behave when POI and address data and building data is mixed into one object?

Would user expect POI tags to be extracted?

koharachan:
This should extract both of them together, I’ll take a look at the code and think about it.

koharachan:
In the current develop branch, modules/operations/extract.ts serves as the UI/availability entry point for Extract. It first maps the selected object into a list of _actions. If the object doesn’t exist, lacks a “meaningful label,” or is an isolated node that doesn’t belong to any path, null is returned. For non-node objects, the code calls presetManager.match(entity, graph) and checks whether preset.geometry contains a point. If not, Extract cannot be used at all. During execution, all actions are combined into a single combinedAction, which is then recorded in the history stack. After that, the code proceeds to modeSelect(context, extractedNodeIDs). There’s one important UX consideration here: After extraction, the id simply marks the newly selected node. It doesn’t automatically proceed to the “move point immediately” phase, as originally planned.
So next, I should first complete the development of address-like key recognition, add area guards, and expand the testing.

modules/actions/extract.ts
k-yle: > @k-yle Thanks for taking the time to fix this directly! [...]

I didn't change anything, I just posted the two comments above

> [...] I've reviewed the changes in "Files changed" and they look good to me.

maybe you were reviewing your own changes? no one else has pushed to [your branch](https://github.com/koharachan/BetteriD/commits/codex/fix-extract-address-12447/)

koharachan: Maybe I misread it?

PR #12614, opened 2026-07-23
migrate turn restrictions to typescript

k-yle:
another easy one

### DIFF
--- a/modules/osm/way.ts
+++ b/modules/osm/way.ts
@@ -27,6 +27,13 @@ export class osmWay extends OsmAbstractEntity {
   declare readonly id: WayId;
   declare readonly nodes: NodeId[];
 
+    /** @deprecated hack used for turn restrictions */ declare __first?: boolean;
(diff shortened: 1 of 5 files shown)


modules/osm/way.ts
```diff
@@ -27,6 +27,13 @@ export class osmWay extends OsmAbstractEntity {
   declare readonly id: WayId;
   declare readonly nodes: NodeId[];
 
+    /** @deprecated hack used for turn restrictions */ declare __first?: boolean;
+    /** @deprecated hack used for turn restrictions */ declare __last?: boolean;
+    /** @deprecated hack used for turn restrictions */ declare __from?: boolean;
+    /** @deprecated hack used for turn restrictions */ declare __via?: boolean;
+    /** @deprecated hack used for turn restrictions */ declare __to?: boolean;
+    /** @deprecated hack used for turn restrictions */ declare __oneWay?: boolean;
```
k-yle, line 35: obviously this is not ideal, but we these properties are used, so we need to define them

tyrasd (approved):

PR #12624, opened 2026-07-27
Fix flash message overflow

### LINKED ISSUE #9072: Follow mode flash message has overflow hidden
### URL

_No response_

Follow mode flash message has overflow hidden. (ie. try to add a way over another -closed- way and make this new way follow the existing way)

image

Perhaps [flash messages](https://github.com/openstreetmap/iD/blob/ad67f81cfde96bfa17488ccb14b51a8655f81681/css/80_app.css#L4523) in general do not have overflow and default to `overflow: visible`?

_No response_

Chrome

alejandro-2g:
## Description

This pull request fixes the flash message overflow issue described in #9072.

When a long follow mode error message is displayed in a narrow browser window, the message is now allowed to wrap onto multiple lines instead of being cut off. The hide animation was also updated so multiline flash messages move completely out of view when dismissed, preventing any portion of the message from remaining visible in the footer.

Fixes #9072

## Testing

- Reproduced the follow mode flash message overflow issue in Mozilla Firefox on Linux Mint using a narrow browser window.
- Verified that long follow mode error messages wrap onto multiple lines and remain fully readable.
- Verified that flash messages disappear completely after the display duration without leaving any text visible in the footer.
- Ran `npm test`.

Test results:
- 137 test files passed
- 2362 tests passed

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -4865,22 +4865,40 @@ img.tile-debug {
     flex: 0 0 100%;
     flex-flow: row nowrap;
     justify-content: space-between;
-    height: 100%;
     position: absolute;
     right: 0;
     left: 0;
 }
+.main-footer-wrap {
+    height: 100%;
+}
 
-.footer-show {
-    bottom: 0px;
+.flash-wrap {
+    height: auto;
+    min-height: 100%;
+    bottom: 0;
+    transition: transform 75ms linear;
+}
+
+.main-footer-wrap.footer-show {
+    bottom: 0;
     transition: bottom 75ms linear;
 }
 
-.footer-hide {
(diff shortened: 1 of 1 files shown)


css/80_app.css
```diff
@@ -4980,6 +5000,9 @@ img.tile-debug {
 
 .flash-text {
     flex: 1 1 auto;
+    min-width: 0;
```
matkoniecz, line 5003: why adding this adds anything or is needed? This should be rather more than covered by `flex`

matkoniecz:
Have you generated this pull request with LLM? Because it is not disclosed and PR description is exactly matching LLM-style ones.

PR #12629, opened 2026-07-28
Add zoom controls to photo viewers

### LINKED ISSUE #8576: Add zoom-in zoom-out buttons to OpenStreetCam images
Hi! When using OpenStreetCam photo overlays it would be useful to have zoom-in and zoom-out buttons ("+" and "-"), as it is for mapillary.

subotac:
Closes #8576

This adds explicit zoom-in and zoom-out buttons to flat photo viewers while preserving their existing wheel and drag behavior.

- The shared plane photo frame covers Panoramax, Vegbilder, and local photos.
- KartaView and Mapilio use the same control with their existing zoom behaviors.
- 360° photos continue to use Pannellum's native zoom controls.
- The buttons use existing translations, expose accessible labels, and disable at the configured scale limits.

Tests cover rendering, scale changes and limits, shared plane photo frames, and switching to panorama frames.

### DIFF
--- a/css/60_photos.css
+++ b/css/60_photos.css
@@ -593,8 +593,21 @@ label.streetside-hires {
     pointer-events: initial;
 }
 
+.photo-zoom-controls {
+    position: absolute;
+    right: 14px;
+    bottom: 90px;
+    display: flex;
+    flex-direction: column;
+    gap: 2px;
+    z-index: 10;
+    pointer-events: none;
+}
+
 .photo-controls button,
-.photo-controls button:focus {
+.photo-controls button:focus,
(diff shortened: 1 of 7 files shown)


tyrasd:
example screenshot:


which looks ok, but in my opinion it would be less confusing if the buttons were not right next to the ◀ ▶ buttons. 

@subotac do you think it would be possible to move the zoom controls to the right hand side side of the viewer, similar to how the mapillary component has them:

subotac:
Yes, separating them from the previous/next buttons makes sense. I'll move the zoom buttons into a vertical control stack on the right, matching Mapillary's placement.

PR #12639, opened 2026-07-29
migrate presets and fields to typescript

k-yle:
this is a fairly straightforward one


modules/presets/preset.ts
```diff
@@ -196,7 +254,7 @@ export function presetPreset(presetID, preset, addable, allFields, allPresets) {
     }
 
     // Lookup documentation on OSM Wikibase...
-    let key = _this.originalReference.key || Object.keys(utilObjectOmit(_this.tags, 'name'))[0];
+    let key = _this.originalReference.key || Object.keys(utilObjectOmit(_this.tags, ['name']))[0];
```
k-yle, line 257: this worked because `.indexOf()`  also exists for strings.

but it's a bug, currently it would deletes tags like `n=*` or `me=*` (substring match)

tyrasd (approved):

PR #12641, opened 2026-07-29
migrate validators to typescript (part 1)

k-yle:
`validationIssue` and `validationIssueFix` are now proper ES classes. This was nice and simple, because they're already being instantiated with the `new` operator


modules/ui/sections/validation_issues.js
```diff
@@ -52,15 +52,15 @@ export function uiSectionValidationIssues(id, severity, context) {
         const withDistance = _issues.map(issue => {
             const extent = issue.extent(graph);
             return {
-                ...issue,
+                issue,
```
k-yle, line 55: ES classes can't be spread, so this part had to be tweaked

tyrasd (approved):

PR #12642, opened 2026-07-29
migrate coreContext to typescript

k-yle:
this was easier than i thought. I kept it as a function, no need to migrate it to an ES6 class. Therefore, there's no impact on tools like [strava-heatmap-extension](https://github.com/julcnx/strava-heatmap-extension) which hack around with `coreContext`.

some more comments inline

### DIFF
--- a/modules/core/context.ts
+++ b/modules/core/context.ts
@@ -1,19 +1,17 @@
-import { debounce, throttle } from 'es-toolkit/compat';
+import { debounce, throttle, type DebouncedFunc } from 'es-toolkit/compat';
 
-import { dispatch as d3_dispatch } from 'd3-dispatch';
+import { dispatch as d3_dispatch, type Dispatch } from 'd3-dispatch';
 import { json as d3_json } from 'd3-fetch';
 import { select as d3_select } from 'd3-selection';
 
 import packageJSON from '../../package.json';
-
-import { t } from '../core/localizer';
-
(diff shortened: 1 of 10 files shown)


modules/modes/select.js
```diff
@@ -30,10 +30,9 @@ import {
 
 
 export function modeSelect(context, selectedIDs) {
-    var mode = {
-        id: 'select',
-        button: 'browse'
-    };
+    var mode = {};
```
k-yle, line 33: this syntax is required so that TS can infer all the properties

modules/core/history.js
```diff
@@ -59,6 +59,7 @@ import { osmIdManager } from '../osm';
 
 /** @typedef {(context: iD.Context, selectedIDs: EntityId[]) => Operation} CreateOperation */
 
+/** @typedef {ReturnType<typeof coreHistory>} coreHistory */
```
k-yle, line 62: also temporary, until this file is migrated to TS

modules/core/context.ts
```diff
@@ -141,8 +286,8 @@ export function coreContext() {
         }
         return;
 
-      } else {
-        _history.merge(result.data, result.extent);
```
k-yle, line 145: this argument isn't used

tyrasd (approved):
awesome, thanks

PR #12647, opened 2026-07-30
Improve community language display on the post changeset screen

### LINKED ISSUE #8937: Spacing on post-changeset community links
Don't know if there's been a deliberate change, but with the new iD version I've noticed there seems to be a lot more spacing between each of the community index links, see screenshot.

Also, each one has "> More" below it, however sometimes when you click on more there is no description and it simply says "_Languages: English_" - which makes clicking on that a bit pointless. 

Where there is no extendedDescription, could the language be indicated in the main description so there is no need for the dropdown? 

The word "Languages" should probably be made singular when there is only one language as well! 😉 

id

alejandro-2g:
## Summary

This PR updates the post-changeset community section so the **More** dropdown is only shown when there is an extended description to display.

If a community only has language information, the language list is now shown directly below the description instead of inside the **More** dropdown. If there is an extended description, the **More** dropdown still appears and the language information remains inside it.

Fixes #8937

## Testing

- Ran `npm run lint`
- Ran `npm test`
- All tests passed (137 test files and 2,363 tests)

I also tried to test the change manually using my local development environment. I wasn't able to complete the post-changeset flow because the local server couldn't authenticate with the OpenStreetMap OAuth service due to a redirect URI mismatch. Because of that, I couldn't verify the UI change manually, but the project built successfully and all automated checks passed.

### DIFF
--- a/modules/ui/success.js
+++ b/modules/ui/success.js
@@ -307,8 +307,8 @@ export function uiSuccess(context) {
       .attr('class', 'community-description')
       .html(d.resolved.descriptionHTML);
 
-    // Create an expanding section if any of these are present..
-    if (d.resolved.extendedDescriptionHTML || (d.languageCodes && d.languageCodes.length)) {
+    // Create an expanding section if an extended description is present.
+    if (d.resolved.extendedDescriptionHTML) {
       selection
         .append('div')
         .call(uiDisclosure(context, `community-more-${d.id}`, false)
@@ -317,6 +317,8 @@ export function uiSuccess(context) {
           .label(() => t.append('success.more'))
           .content(showMore)
         );
+    } else {
+      showLanguages(selection);
     }
 
     let nextEvents = (d.events || [])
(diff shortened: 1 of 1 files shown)


matkoniecz:
> I wasn't able to complete the post-changeset flow because the local server couldn't authenticate with the OpenStreetMap OAuth service due to a redirect URI mismatch. 

This should work in PR preview above (test edits making garbage changes should be made at test server)

matkoniecz:
Opened https://github.com/openstreetmap/iD/issues/12651

tyrasd:
here's a screenshot how it looks like after editing somewhere in Germany, compared to how it currently looks like (on the right hand side, with only a few "show more" sections expanded).

 


To be honest, I don't particularly like the noisy repeated `Languages: <local dominant language>` which adds very little information I'd say: normally it is quite clear from the context what language the respective community interacts with, and some of the data is quite outdated it seems (e.g. the community forum has many of the languages not listed). Maybe we can further simplify this list by moving the languages to a tooltip (`title` attribute when hovering the paragraph of the respective community description)?

1ec5:
Maybe only display the Languages if it’s something other than the current locale? But this will annoy techies who use English for everything regardless of their native language.

A tooltip wouldn’t be very discoverable, but maybe a 💬 or A文 icon with a tooltip could work.

PR #12655, opened 2026-07-31
delete dead code in presetManager 

k-yle:
you can't convert a file to typescript if there's clearly broken code, so we need to either fix the bugs or delete all the dead code.

does not need a changelog entry

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -1,15 +1,13 @@
-import { dispatch as d3_dispatch } from 'd3-dispatch';
-
 import { prefs } from '../core/preferences';
 import { fileFetcher } from '../core/file_fetcher';
 import { locationManager } from '../core/location_manager';
 
-import { osmNodeGeometriesForTags, osmSetAreaKeys, osmSetLineTags, osmSetPointTags, osmSetVertexTags } from '../osm/tags';
+import { osmNodeGeometriesForTags, osmSetAreaKeys, osmSetPointTags, osmSetVertexTags } from '../osm/tags';
 import { presetCategory } from './category';
 import { presetCollection } from './collection';
 import { presetField } from './field';
 import { presetPreset } from './preset';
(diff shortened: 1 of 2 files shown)


modules/presets/index.js
```diff
@@ -348,7 +348,7 @@ export function presetIndex() {
 
 
   _this.lineTags = () => {
-    return _this.collection.filter((lineTags, d) => {
+    return _this.collection.reduce((lineTags, d) => {
```
k-yle: this never worked, `reduce` and `filter` have different arguments

modules/presets/index.js
```diff
@@ -514,7 +514,7 @@ export function presetIndex() {
   _this.getAddable = () => {
       if (!_addablePresetIDs) return [];
 
-      return _addablePresetIDs.map((id) => {
+      return [..._addablePresetIDs].map((id) => {
```
k-yle: `_addablePresetIDs` is a `Set`, and `Set`s don't have a `.map` property

modules/presets/index.js
```diff
@@ -359,8 +359,8 @@ export function presetIndex() {
 
       // if this can be a line
       if (d.geometry.indexOf('line') !== -1) {
-        lineTags[key] = lineTags[key] || [];
-        lineTags[key].push(d.tags);
+        lineTags[key] ||= {};
+        lineTags[key][d.tags[key]] = true;
```
k-yle: `lineTags[key]` is an object, not an array

tordans:
>are not used anywhere in the codebase, so this bugfix has no impact

What am I missing why we would not just remove dead code?

k-yle:
okay, done. most of this code is related to iD v3, which is why it doesn't work anymore. but i look forward to reviving some of this one day, once we clear the current PR backlog

tyrasd (approved):

Issue #2493, opened 2014-12-23
More search-ranges between "visible map" and "worldwide"


hajo4:
If "Search features"  finds "no results in visible map area",
it offers to "search worldwide".

It would be more useful to have the option to search more locally first,
e.g. depending on zoomlevel map-area*2 / *4 / *8 ...
or 1 km / 10 km / 100 km around, 
or "inside this town/region/country/continent only".

PR #12657, opened 2026-07-31
Filter combo field options by optionsLocationSet

julcnx:
This came out of a discussion on [openstreetmap/id-tagging-schema#2332](https://github.com/openstreetmap/id-tagging-schema/issues/2332), where a maintainer asked whether a single `surface` combo option (`laterite`, a tropical soil type essentially absent outside a specific belt of countries) could be scoped to only the regions where it's actually relevant, instead of appearing in the dropdown everywhere. Field-level `locationSet` already exists for hiding a whole field by region (`modules/ui/field.js`), but there's nothing equivalent for a single option inside a shared combo field like `surface`, which is used across most highway presets globally.

id-tagging-schema's schema gained an `optionsLocationSet` property to carry this (companion PR: [openstreetmap/schema-builder#349](https://github.com/openstreetmap/schema-builder/pull/349)), a map from option value to a `locationSet` object. This PR wires that data up on the iD side.

## What changed

- **`modules/presets/index.js`**: when merging a field with an `optionsLocationSet`, each option's `locationSet` gets pushed through `locationManager.registerLocationSets()` (the same call already used for whole-field/preset/category `locationSet`s), and the resulting `locationSetID`s land on `field.optionsLocationSetID`, keyed by option value.
- **`modules/ui/fields/combo.js`**: adds `optionAllowedHere(value)`, which checks the current entity's location against the option's `locationSetID` the same way `modules/ui/field.js`'s `isAllowed()` already does for whole fields (same `combinedEntityExtent()` / `locationManager.locationSetsAt()` pattern). Applied in both suggestion paths:
  - the static options list (`getOptions()`)
  - the live taginfo-backed suggestions, since taginfo has no concept of this restriction and would otherwise surface a globally-popular value (`laterite` currently has 700+ uses) regardless of where you're editing

An option with no entry in `optionsLocationSetID`, or when the entity's location can't be determined (e.g. no entity yet), is offered everywhere, unchanged from today. Typing a value directly is still always possible (`customValues`), this only affects what gets suggested, matching how whole-field `locationSet` filtering already behaves (it hides the field, but doesn't retroactively invalidate already-tagged data).

## Testing

- `modules/presets/index.js`: unit test confirming `optionsLocationSetID` resolves correctly on merge, and that fields without `optionsLocationSet` are unaffected.
- `modules/ui/fields/combo.js`: two end-to-end tests exercising the real DOM/combobox path (render the field, focus, simulate a `↓` keypress to open the dropdown, read the actual rendered `.combobox-option` elements), confirming a region-restricted option is excluded when `locationSetsAt()` doesn't include its id, and included when it does.
- Full suite passes locally (`npm run lint`, `tsc`, `vitest run`, 2367 passed).

This is the iD half of a three-repo change; a follow-up id-tagging-schema PR will apply `optionsLocationSet` to `laterite` once this and the schema-builder PR both land. Happy to adjust the approach if you'd prefer something different.

### DIFF
--- a/modules/presets/index.js
+++ b/modules/presets/index.js
@@ -96,6 +96,7 @@ export function presetIndex() {
   //}
   _this.merge = (d) => {
     let newLocationSets = [];
+    let newOptionLocationSets = [];   // [{ field, value, wrapper }]
 
     // Merge Fields
     if (d.fields) {
@@ -105,6 +106,14 @@ export function presetIndex() {
         if (f) {   // add or replace
           f = presetField(fieldID, f);
           if (f.locationSet) newLocationSets.push(f);
+          if (f.optionsLocationSet) {
+            f.optionsLocationSetID = {};
+            Object.keys(f.optionsLocationSet).forEach(value => {
+              const wrapper = { locationSet: f.optionsLocationSet[value] };
+              newLocationSets.push(wrapper);
+              newOptionLocationSets.push({ field: f, value: value, wrapper: wrapper });
+            });
+          }
           _fields[fieldID] = f;
 
         } else {   // remove
@@ -193,6 +202,11 @@ export function presetIndex() {
     if (newLocationSets.length) {
       locationManager.registerLocationSets(newLocationSets);
     }
+
+    // Copy each option's resolved locationSetID back onto its field.
+    newOptionLocationSets.forEach(({ field, value, wrapper }) => {
+      field.optionsLocationSetID[value] = wrapper.locationSetID;
+    });
 
     return _this;
   };
--- a/modules/ui/fields/combo.js
+++ b/modules/ui/fields/combo.js
@@ -5,6 +5,7 @@ import * as countryCoder from '@rapideditor/country-coder';
 
 import { fileFetcher } from '../../core/file_fetcher';
 import { localizer, t } from '../../core/localizer';
+import { locationManager } from '../../core/location_manager';
 import { services } from '../../services';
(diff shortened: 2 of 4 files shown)


tordans:
We are still discussing what the new format should look like. Latest discussion on this topic is here https://github.com/openstreetmap/id-tagging-schema/issues/2493#issuecomment-4968881391

I don't think the idea at https://github.com/openstreetmap/schema-builder/pull/349/changes#diff-52dd1cfe1fba598e780b1576304786ad0f86bbbbcb7b0c505e75ba12563075c9R120 is better.

julcnx:
Thanks for the pointer, hadn't seen #2493. Fair concern, `optionsLocationSet` as a standalone property is exactly the kind of one-off metadata addition that thread is trying to avoid piling up.

This PR's `combo.js`/`presets/index.js` plumbing only depends on `field.optionsLocationSetID` at runtime, it doesn't care what the upstream schema property is called or how it's nested. So it should adapt to whatever comes out of #2493 (e.g. living under a richer `options: [{value, locationSet}]` shape) with a small follow-up rather than a rewrite.

Happy to wait for #2493 to settle before pushing on the schema side, or to help prototype the richer-field-metadata format there if that's useful. Let me know which is more helpful.

k-yle (commented):
i agree, we should go with tobias' more flexible proposal from https://github.com/openstreetmap/id-tagging-schema/issues/2493#issuecomment-4968881331 - that would solve 3 different issues at once

PR #12658, opened 2026-08-01
fix: Respect embankment:side / cutting:side when determining sidedness

### LINKED ISSUE #12656: Embankment (embankment=yes) triangle markers don't respect embankment:side=*
### URL

https://www.openstreetmap.org/edit#map=19/47.785250/18.115583

Add `embankment=yes` and `embankment:side=right` to a way.

The triangle markers are always shown on both sides of ways with `embankment=yes`, even though the embankment is explicitly specified as one-sided using `embankment:side=*`.

The new embankment markings were implemented very recently, in #12396.

2.41.2, 2.42.0-dev

Chrome

bhavyaKhatri2703:
fixes #12656 

Fixes sidednessIdentifier() in modules/osm/way.ts so that when a way has embankment=yes/dyke/both or cutting=yes/both (which normally render both-sided triangle markers), it now also checks for a companion embankment:side=* / cutting:side=* tag. If that tag is left or right, the function returns the corresponding one-sided category (embankment-left/embankment-right/cutting-left/cutting-right) instead of the generic both-sided one, so the renderer picks the correct one-sided marker (these markers already existed in modules/svg/defs.ts, they just weren't reachable via the documented :side tag).

### DIFF
--- a/modules/osm/way.ts
+++ b/modules/osm/way.ts
@@ -184,8 +184,7 @@ export class osmWay extends OsmAbstractEntity {
             if (key in osmSidednessTags && (value in osmSidednessTags[key])) {
                 if (osmSidednessTags[key][value] === true) {
                     return key;
-                } 
-                
+                }
                 // embankment=yes/dyke/both and cutting=yes/both are both-sided
                 // by default, but a separate `embankment:side=*` or
                 // `cutting:side=*` tag can narrow it to just one side.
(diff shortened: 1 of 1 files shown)


matkoniecz:
Note that this issue was just opened and tag is quite rare. 


It may be smarter to make PRs for issues that had chance to be reviewed

tyrasd:
closing until #12656 is clarified

PR #12664, opened 2026-08-03
Delete the unused lane parsing code

wantaekchoi:
Retargeted to deleting the file, as @matkoniecz and @k-yle suggested.

Nothing reaches `osmLanes`: its only caller was `uiFieldLanes`, and no field in id-tagging-schema declares `"type": "lanes"`.

Also removes the field, its `uiFields` entry, `osmWay.lanes()`, the `modules/osm` re-export, the test file, and the `ARCHITECTURE.md` line, the appendages asked about in #12668. `osmLanes` and `uiFieldLanes` were both exported, so the changelog entry is marked breaking.


modules/osm/lanes.js
```diff
@@ -119,7 +119,7 @@ function parseMaxspeed(tags) {
     var maxspeed = tags.maxspeed;
     if (!maxspeed) return;
 
-    var maxspeedRegex = /^([0-9][\.0-9]+?)(?:[ ]?(?:km\/h|kmh|kph|mph|knots))?$/;
+    var maxspeedRegex = /^([0-9][\.0-9]*?)(?:[ ]?(?:km\/h|kmh|kph|mph|knots))?$/;
```
1ec5: The change looks fine to me, though I can’t tell where we’re even using this code. It seems to be part of some general-purpose lane parsing code. Hopefully we haven’t lost any code that was supposed to use it for something like validation. Maybe it was intended to undergird a future lane editor?

https://github.com/openstreetmap/iD/blob/3940995296b4f767451dcba2cd25edbebcd237f4/modules/osm/lanes.js#L8 https://github.com/openstreetmap/iD/blob/3940995296b4f767451dcba2cd25edbebcd237f4/modules/ui/fields/lanes.js#L14

1ec5 (approved):

modules/osm/lanes.js
matkoniecz: Note: it seems that both old and new will match `1...............2` etc.

Or `1...............2....3.3.3..2.12.2.23.2.2.2.2...1.1.1.1.1.1.1111111`

Following should fix reported problem and also multidot one, from what I see.

But not sure should we keep such dead code around in the first place.

```suggestion
    var maxspeedRegex = /^([0-9]+(\.[0-9]+)?)(?:[ ]?(?:km\/h|kmh|kph|mph|knots))?$/;
```

k-yle (commented):
yeah, i would suggest deleting this whole file, like https://github.com/openstreetmap/iD/pull/12655#issuecomment-5143237817 

and the forks of iD that already have lane editors don't reuse this code

PR #12668, opened 2026-08-03
Delete modules/osm/lanes.js as unused

matkoniecz:
closes #12664


modules/osm/lanes.js
k-yle, line 1: there's also a unit test file, and i think node.ts or way.ts imports from here?

k-yle (changes requested):

matkoniecz:
(if @wantaekchoi wants to adapt their PR to delete entire file and its appendages, feel free to do so)

wantaekchoi:
#12664 now deletes the file and the appendages @k-yle listed: the test file, `osmWay.lanes()`, both barrel exports, the field itself, and the ARCHITECTURE.md line. Thanks for the offer.

Issue #5754, opened 2019-01-24
range of directions on viewpoints


eehpcm:
I recently mapped a tourism=viewpoint and tried to specify a range on the direction.  When I deselected the object iD's internal validation silently removed the range and left only the first angle.  So direction=90-270 became, without my knowing, direction=90.  Only when I selected the object again did this behind-the-scenes modification become apparent.

It is debatable if this is correct behaviour or not.  Allowing direction to specify a range seems to have first appeared in the wiki page for tourism=viewpoint, but I can't find a proposal for tourism=viewpoint.  The range idea then appears to have been imported into the wiki page for the direction tag itself, again without a proposal or much discussion.  So it could be argued that direction ranges should not be permitted or it could be argued that OSM tags are anarchic so if somebody wants to do that then they can.

So, should iD be restricting the range?  And if so, should it do so silently or should it either refuse to let the user enter a range (refuse to accept the - sign and characters that follow it) or pop up a warning dialogue?

bhousel:
related  #4755 

We can't support the ranged values currently, but we could add support for rendering them by changing 1. the code to parse the values and 2. the code to render the viewfields.

> The range idea then appears to have been imported into the wiki page for the direction tag itself, again without a proposal or much discussion. So it could be argued that direction ranges should not be permitted or it could be argued that OSM tags are anarchic so if somebody wants to do that then they can.

Yes, it's really best to think of the OSM tagging as anarchic.  People can propose stuff and write stuff on the wiki, but that doesn't mean anybody will support it.

eehpcm:
@bhousel 

> We can't support the ranged values currently, but we could add support for rendering them by changing 1. the code to parse the values and 2. the code to render the viewfields.

Sounds like a lot of work.  Especially as it (currently) isn't rendered any differently on OSM carto if direction is specified as a single value, or as a range, or is not given at all.  In fact, I'm so old that I remember when iD didn't show the direction in the overlay.  So I'm not too worried if the overlay doesn't support it, but it's counterintuitive when iD changes things behind my back.  For example, if iD made it more obvious that it disallowed a range 0-180 then people might instead choose direction=90 instead of what iD silently gives them - direction=0.

bhousel:
> So I'm not too worried if the overlay doesn't support it, but it's counterintuitive when iD changes things behind my back. For example, if iD made it more obvious that it disallowed a range 0-180 then people might instead choose direction=90 instead of what iD silently gives them - direction=0.

[...]

PR #12673, opened 2026-08-04
Add rotate action for points with `direction`

### LINKED ISSUE #12341: Rotating a point should adjust its direction
The Rotate operation should be enabled when a point is selected and the point has a numeric `direction=*` tag (and thus a direction cone). When the Rotate operation is active, the `direction=*` tag should respond to the mouse cursor moving around the point. This would be similar to when an area is being rotated around its centroid, except that the pivot is the selected point itself. The direction cone would provide interactive feedback, similar to the dial proposed in #12104 but integrated directly into the map view.

tordans:
**Testing:** 
* standalone (bench) https://pr-12673--ideditor.netlify.app/#disable_features=boundaries&map=21.85/52.47409/13.44556&background=Bing&id=n12518261743
* vertex (camera) https://pr-12673--ideditor.netlify.app/#disable_features=boundaries&map=21.96/52.47427/13.44495&background=Bing&id=n877058722

---

This is a subset of https://github.com/openstreetmap/iD/pull/12104.
The PR only add the `R` Shortcut that we explored over there, plus some context.

The idea is: This change is a lot simpler than https://github.com/openstreetmap/iD/pull/12104 but it bringt us a long way to improve the UX for editing node `direction`s.

Changes:
- Add the `R` for nodes with `direction` tag
- Help text on the Info-i of the field
- Some other help texts modified
- During rotate, the sidebar stays on the feature (and does not switch to "Search features") – this ist IMO a UX bug we had before but we did not really notice because when rotating lines and areas I look at the map, not the sidebar. But for nodes, I want to see the direction number change during rotate, so the feature needs to stay active.

Closes https://github.com/openstreetmap/iD/issues/12341

---

Aside: This idea was also adopted to GoMap where it works great, IMO

---

🤖  the changes are build with Cursor Models and reviewed by me.

### DIFF
--- a/modules/actions/rotate_point_direction.ts
+++ b/modules/actions/rotate_point_direction.ts
@@ -1,11 +1,12 @@
 import type { Action } from '../core/history';
 import type { EntityId } from '../osm';
 import { utilRotatePointDirectionKey } from '../util/direction_field';
-import { utilWrap } from '../util';
+import { utilRetargetDirectionDegreesValue } from '../util/direction_degrees';
 
 
 /**
- * Set a node's numeric direction tag to `degrees` (OSM azimuth, clockwise from north).
+ * Set a node's direction tag so it aims at `degrees` (OSM azimuth, clockwise from north).
+ * Multi-value tags keep relative offsets; cardinals become numeric degrees.
  */
 export function actionRotatePointDirection(
     entityID: EntityId,
@@ -19,8 +20,10 @@ export function actionRotatePointDirection(
         const directionKey = key || utilRotatePointDirectionKey(entity, graph);
         if (!directionKey) return graph;
 
-        // Keep values aligned with iD direction tagging (whole degrees, [0, 360)).
-        const nextDirection = Math.round(utilWrap(degrees, 360)).toString();
+        const nextDirection = utilRetargetDirectionDegreesValue(
+            entity.tags[directionKey],
+            degrees
+        );
         if (nextDirection === entity.tags[directionKey]) return graph;
 
         const tags = Object.assign({}, entity.tags, { [directionKey]: nextDirection });
--- a/modules/actions/rotate_point_direction.ts
+++ b/modules/actions/rotate_point_direction.ts
@@ -1,0 +1,21 @@
+import type { Action } from '../core/history';
+import type { EntityId } from '../osm';
+import { utilWrap } from '../util';
+
+
+export function actionRotatePointDirection(entityID: EntityId, deltaDegrees: number): Action {
+    return function(graph) {
+        const entity = graph.hasEntity(entityID);
+        if (!entity || entity.type !== 'node') return graph;
+
+        const direction = Number(entity.tags.direction);
+        if (!isFinite(direction)) return graph;
+
+        // Keep values aligned with iD direction tagging (whole degrees, [0, 360)).
+        const nextDirection = Math.round(utilWrap(direction + deltaDegrees, 360)).toString();
+        if (nextDirection === entity.tags.direction) return graph;
+
+        const tags = Object.assign({}, entity.tags, { direction: nextDirection });
+        return graph.replace(entity.update({ tags: tags }));
+    };
+}
--- a/modules/actions/rotate_point_direction.ts
+++ b/modules/actions/rotate_point_direction.ts
@@ -1,21 +1,29 @@
 import type { Action } from '../core/history';
 import type { EntityId } from '../osm';
+import { utilRotatePointDirectionKey } from '../util/direction_field';
 import { utilWrap } from '../util';
 
 
-export function actionRotatePointDirection(entityID: EntityId, deltaDegrees: number): Action {
+/**
+ * Set a node's numeric direction tag to `degrees` (OSM azimuth, clockwise from north).
+ */
+export function actionRotatePointDirection(
+    entityID: EntityId,
+    degrees: number,
+    key?: TagKey
+): Action {
     return function(graph) {
         const entity = graph.hasEntity(entityID);
         if (!entity || entity.type !== 'node') return graph;
 
-        const direction = Number(entity.tags.direction);
-        if (!isFinite(direction)) return graph;
+        const directionKey = key || utilRotatePointDirectionKey(entity, graph);
+        if (!directionKey) return graph;
 
         // Keep values aligned with iD direction tagging (whole degrees, [0, 360)).
-        const nextDirection = Math.round(utilWrap(direction + deltaDegrees, 360)).toString();
-        if (nextDirection === entity.tags.direction) return graph;
+        const nextDirection = Math.round(utilWrap(degrees, 360)).toString();
+        if (nextDirection === entity.tags[directionKey]) return graph;
 
-        const tags = Object.assign({}, entity.tags, { direction: nextDirection });
+        const tags = Object.assign({}, entity.tags, { [directionKey]: nextDirection });
         return graph.replace(entity.update({ tags: tags }));
     };
 }
--- a/modules/modes/rotate.js
+++ b/modules/modes/rotate.js
@@ -22,7 +22,9 @@ import { operationMove } from '../operations/move';
 import { operationOrthogonalize } from '../operations/orthogonalize';
 import { operationReflectLong, operationReflectShort } from '../operations/reflect';
 
+import { utilSelectedRotatePointDirectionKey } from '../util/direction_field';
 import { utilKeybinding } from '../util/keybinding';
+import { utilWrap } from '../util';
 import { utilFastMouse, utilGetAllNodes } from '../util/util';
 
 
@@ -53,7 +55,7 @@ export function modeRotate(context, entityIDs) {
     var _prevAngle;
     var _prevTransform;
     var _pivot;
-    let _pointDirectionRotate = false;
+    let _pointDirectionKey = false;
 
     // use pointer events on supported platforms; fallback to mouse events
     var _pointerPrefix = 'PointerEvent' in window ? 'pointer' : 'mouse';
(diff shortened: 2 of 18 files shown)


tordans:
## Screenshots
* regular node 
   point regular
* node with rotate
   point rotate
* rotating in progress with info line bottom and sidebar still visible
   in progress
* help on info-`i`
   help inline
* help panel 
   help panel

tyrasd (commented):
Nice, thanks. The only thing I noticed at first while playing around with it was that the operation is using the angle relative to the mouse position at the start of the operation, instead of just snapping it to the absolute direction the mouse points to. This could be a little counterintuitive maybe (e.g. where you need to draw), but I'd say we should probably leave it like this for consistency to the rotate operation for areas and lines (where there no absolute direction really exists which the feature can be directly rotated to), but maybe we could add a modifier key like <kbd>Shift</kbd> which toggles the relative to absolute mode? Does not need to be done in this PR immediately, though. Just wanted to find out about what your opinion on this idea.

One additional quality of life improvement would be to be able to start setting a direction on such a vertex by pressing <kbd>R</kbd> even if the tag was previously absent. This could be allowed when the preset has a field for the `direction` tag in `fields` or `moreFields`.

> During rotate, the sidebar stays on the feature

Good catch! I also never noticed it until now, but it was definitely a bug.

tordans:
> using the angle relative to the mouse position at the start of the operation

Will test this more and push the other technique so we can see how much code is needed for this. I think for nodes it will feel more natural.

>One additional quality of life improvement would be to be able to start setting a direction on such a vertex by pressing R even if the tag was previously absent. This could be allowed when the preset has a field for the direction tag in fields or moreFields.

Agreed, we should do that. From the top of my head GoMap does it for `direction` and `camera:direction`.

modules/modes/rotate.js
```diff
@@ -27,15 +28,15 @@ import { utilFastMouse, utilGetAllNodes } from '../util/util';
 
 export function modeRotate(context, entityIDs) {
 
-    var _tolerancePx = 4; // see also behaviorDrag, behaviorSelect, modeMove
+    const _tolerancePx = 4; // see also behaviorDrag, behaviorSelect, modeMove
```
tordans: FYI will revert those changes of lines that are not touched by the code. My cleanup was a bit too broad here. My goal is to only update to const/let for lines that the PR actually touches.

RudyTheDev:
This won't work for surveillance camera (`man_made=surveillance`) which has `camera:direction`, not `direction`.

I was not able to rotate a point attached to a way like an information board on a building or something. Are these not supposed to work?

> One additional quality of life improvement would be to be able to start setting a direction on such a vertex by pressing R even if the tag was previously absent.

Ditto.

modules/ui/tag_reference.js
```diff
@@ -93,6 +94,20 @@ export function uiTagReference(what) {
               .call(t.append(docs.wiki.text));
         }
 
+        // Rotate (R) adjusts numeric direction=* on standalone points.
+        if (what.key === 'direction') {
+            _body
+                .append('div')
+                .attr('class', 'tag-reference-separator');
+
```
tordans: I will try to avoid this wrapper div so we only have one new class for tag-reference-shortcut-hint and consolidate the style changes there.

modules/actions/rotate_point_direction.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { EntityId } from '../osm';
+import { utilWrap } from '../util';
+
+
+export function actionRotatePointDirection(entityID: EntityId, deltaDegrees: number): Action {
+    return function(graph) {
+        const entity = graph.hasEntity(entityID);
+        if (!entity || entity.type !== 'node') return graph;
+
+        const direction = Number(entity.tags.direction);
```
k-yle: as mentioned above, `direction` is not the only tag. we already have a helper function to determine the correct direction tag based on the preset, from #9317. maybe that could be reused?

https://github.com/openstreetmap/iD/blob/3d1c9c57cf01d883a1dfe26149167a2cb1c02532/modules/actions/reverse.ts#L163-L185

modules/util/direction_field.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+
+/**
+ * Find a direction field on the entity's preset.
+ * @param numeric when true, match angle fields (`type: number`);
+ *   when false, match relative fields used by reverse (`forward`/`backward`).
+ * @returns false or the direction tag key
+ */
+export function utilDirectionFieldKey(
```
tordans, line 36: FYI Context: 

> `utilDirectionFieldKey` is the old `supportsDirectionField` helper moved out of `reverse.ts` and parameterized.
> 
> For reverse (`numeric: false`) the behavior is unchanged: same preset `fields`/`moreFields` scan, same `direction` / `*:direction` key check, same geometry filter, still only non-`number` (relative) fields.
> 
> `numeric: true` is the new path for rotate (angle fields). Key matching now goes through shared `utilIsDirectionKey`, which is equivalent for the reverse case.

tordans:
**FYI — updates since the last review** (history was rebased/amended onto `upstream/develop`)

Addressed the review feedback and a few follow-ups:

- **Direction keys:** not only `direction` — also `*:direction` (e.g. `camera:direction`), via a shared helper extracted from reverse’s old `supportsDirectionField`.
- **Vertices:** R works on way-attached nodes, not only standalone points; added `annotation.vertex` for history.
- **Absent tags:** R works when the preset has a numeric direction field even if the tag is missing.
- **Interaction:** for nodes, rotate **points at the mouse** (absolute azimuth), not relative drag like line/area rotate. After pan/zoom, the first move does not snap the tag (same idea as geometry rotate’s zero-delta).
- **Docs/hint:** updated
- **Cleanup:** shared selection/key helpers; `const`/`let` only on lines we touch; no broad `var` churn.

Compared against the old tip `ed350d56` (~25h+12h ago). Current tip is `c9942c3b5` (already on `origin/rotate-point-direction`).

* standalone (bench) https://pr-12673--ideditor.netlify.app/#disable_features=boundaries&map=21.85/52.47409/13.44556&background=Bing&id=n12518261743
* vertex (camera) https://pr-12673--ideditor.netlify.app/#disable_features=boundaries&map=21.96/52.47427/13.44495&background=Bing&id=n877058722

modules/actions/rotate_point_direction.ts
tordans: Thanks, done, details below.

modules/actions/rotate_point_direction.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Set a node's numeric direction tag to `degrees` (OSM azimuth, clockwise from north).
+ */
+export function actionRotatePointDirection(
+    entityID: EntityId,
+    degrees: number,
+    key?: TagKey
+): Action {
+    return function(graph) {
```
k-yle, line 16: for two-sided features, pressing <kbd>R</kbd> will destroy the two-sidedness. for example: `direction=120;300`
Image



i think it should either:
- preseve the sidedness - if there were 2 angles offset by 180degrees before rotating, then it should rotate both angles, still offset by 180degrees.
- or; block the operation in this situation

test/spec/util/direction_field.ts
```diff
@@ -0,0 +1,88 @@
+describe('iD.utilDirectionFieldKey / iD.utilRotatePointDirectionKey', () => {
+    beforeEach(() => {
+        const cached: any = iD.fileFetcher.cache();
+        cached.preset_fields = {
```
k-yle, line 4: currently there is exactly 1 file that overrides `preset_fields`, now we have multiple files mutating a global state.

i'm not sure if this will cause a race condition because we use [`--no-isolate`](https://github.com/openstreetmap/iD/pull/10452)? i guess we'll find out...

test/spec/util/direction_field.ts
```diff
@@ -0,0 +1,88 @@
...
+    it('shows the rotate hint only for numeric direction fields or values', () => {
+        expect(iD.utilShowsDirectionRotateHint({ key: 'direction', fieldType: 'number' })).toBeTruthy();
+        expect(iD.utilShowsDirectionRotateHint({ key: 'direction', fieldType: 'combo' })).toBeFalsy();
+        expect(iD.utilShowsDirectionRotateHint({ key: 'direction', value: '90' })).toBeTruthy();
+        expect(iD.utilShowsDirectionRotateHint({ key: 'direction', value: 'forward' })).toBeFalsy();
+        expect(iD.utilShowsDirectionRotateHint({ key: 'direction' })).toBeFalsy();
+    });
+
+    it('finds a numeric direction field on the preset', async () => {
+        await (iD.presetManager as any).ensureLoaded(true);
```
k-yle, line 46: fixed in #12683, i'll update it if this PR is merged first

modules/util/direction_field.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export function utilRotatePointDirectionKey(
+    node: OsmEntity,
+    graph: coreGraph
+): false | TagKey {
+    let plainDirectionKey: TagKey | false = false;
+
+    for (const key of Object.keys(node.tags)) {
+        if (!utilIsDirectionKey(key)) continue;
+        if (!isFinite(Number(node.tags[key]))) continue;
```
k-yle: this will reject [any cardinal value (see list here)](https://github.com/openstreetmap/iD/blob/2a5c68aa04cb4279e5af0d24f00bf7ae9b0635dd/modules/osm/node.ts#L12-L29).

Since #9141, the field's <kbd>⬆️</kbd>/<kbd>⬇️</kbd> buttons support these values, so the action should probably support it?

k-yle (commented):
very cool! sorry i have another few questions

modules/ui/tag_reference.js
```diff
@@ -93,6 +95,16 @@ export function uiTagReference(what) {
               .call(t.append(docs.wiki.text));
         }
 
+        // Rotate (R) adjusts numeric direction / *:direction on the map.
+        if (utilShowsDirectionRotateHint(what)) {
```
k-yle, line 99: i don't know where the previous comments went, but i have one more thought:

 this feels quite hacky, maybe it would be worth opening a new issue to discuss if we should start putting unrelated "tips & tricks" into this panel? 

if yes, it probably shouldn't be hardcoded for the rotate operation

tordans: My take is: Yes, we need to make a decision in this PR if we think this is helpful for the feature. 

But we should postpone the discussion of an abstraction to a follow-up issue. We could create an issue to review existing cases to see if adding such a help section would be useful. Or we could just wait for the next PR that needs to add something here and then look into an abstraction. — Building and discussing the abstraction now is IMO premature. And we should make the decision for this not based on the code – one more if is not that big of a deal – but from a user point of view.

I am open to keep or remove it either way…

tordans:
**FYI — updates since the last review** ([previous note](https://github.com/openstreetmap/iD/pull/12673#issuecomment-5200295456))

Addressed the later review feedback:

- **Two-sided / multi-value tags:** `direction=120;300` no longer collapses to a single angle. Rotate aims the first parseable segment at the mouse and applies the same delta to the rest (e.g. mouse at 90° → `90;270`), matching the field’s `+/-` behavior. Unparseable segments (`forward`, `error`, …) are left alone.
- **Cardinals:** `N` / `ne` / etc. are eligible for R and convert to numeric degrees on rotate (same recognition as the field’s `+/-` since #9141).
- **Shared helpers:** parse / has / retarget live in `modules/util/direction_degrees.ts`; eligibility + the rotate hint use the same rules.

test/spec/util/direction_field.ts
tordans: Should we address this now and refactor? Or in a follow up? Or only once we work on #10452 again?

andrewharvey:
Where `direction` is a range, eg. `direction=N-S` should it retain the angle between the start/end bearing (the range) and adjust both in unison, or should the tool be disabled?

Because currently you can enable the tool but it doesn't do anything.

k-yle:
that sounds like https://github.com/openstreetmap/iD/pull/12673#discussion_r3728139539, i guess it would just need to split at hyphens as well as semicolons

andrewharvey:
> that sounds like https://github.com/openstreetmap/iD/pull/12673#discussion_r3728139539, i guess it would just need to split at hyphens as well as semicolons

Yes, it's the same logic that could apply to both scenarios. I think best to just have it modify both numbers, so you can fine tune a range, or multiple values.

However, a range doesn't give a visual feedback in iD, so we should either fix that, or disable the rotate tool for these cases until that's fixed.

k-yle:
> However, a range doesn't give a visual feedback in iD, so we should either fix that [...]

yeah that's #5754, which should be easy and fun to implement (separately to this PR)

PR #12674, opened 2026-08-04
Add rotate action for points with `direction`  (stacked branch experiment)

### LINKED ISSUE #12341: Rotating a point should adjust its direction

tordans:
**Testing:** https://pr-12674--ideditor.netlify.app/#disable_features=boundaries&map=21.85/52.47409/13.44556&background=Bing&id=n12518261743

---

This is a subset of #12675 (formerly #12104).
The PR only adds the `R` shortcut that we explored over there, plus some context.

The idea is: This change is a lot simpler than the dial PR but it brings us a long way to improve the UX for editing node `direction`s.

Changes:
- Add the `R` for nodes with `direction` tag
- Help text on the Info-i of the field
- Some other help texts modified
- During rotate, the sidebar stays on the feature (and does not switch to "Search features") – this is IMO a UX bug we had before but we did not really notice because when rotating lines and areas I look at the map, not the sidebar. But for nodes, I want to see the direction number change during rotate, so the feature needs to stay active.

Closes #12341

🤖 the changes are build with Cursor Models and reviewed by me.

---

Aside: This idea was also adopted to GoMap where it works great, IMO

---

Replaces #12673 (recreated as a same-repo PR so it can use GitHub stacked PRs). Sorry for the noise, wanted to try out the new stacked branch feature.

---

## Screenshots
* regular node 
   point regular
* node with rotate
   point rotate
* rotating in progress with info line bottom and sidebar still visible
   in progress
* help on info-`i`
   help inline
* help panel 
   help panel

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -2919,6 +2919,13 @@ button.raw-tag-option svg.icon {
 .tag-reference-link .icon:first-child {
     margin-left: 0;
 }
+.tag-reference-separator {
+    border-top: 1px solid var(--border-color);
+    margin: 8px 0;
+}
+.shortcut-hint {
+    margin: 0;
+}
 
 img.tag-reference-wiki-image {
     float: right;
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -733,6 +733,7 @@ en:
     edit_reference: "edit/translate documentation"
     wiki_reference: View documentation
     wiki_en_reference: View documentation in English
+    direction_rotate_help: "Press {key} to adjust this value on the map."
     key_value: "key=value"
     empty: (empty)
     multiple_values: Multiple Values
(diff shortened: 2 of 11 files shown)


tordans:
Sorry for the noise, https://github.com/openstreetmap/iD/pull/12673 is still the right PR. I wanted to try out stacked branches but those don't work with forks, so I tested this… but I don't think its worth to change our standard system just for stacked branches, so I closed this again.

PR #12675, opened 2026-08-04
Add direction dial for numeric `*:direction` fields (stacked branch experiment)

### LINKED ISSUE #5530: Better angle field editor for direction tag

tordans:
Closes #5530
Ping #9733, https://github.com/openstreetmap/id-tagging-schema/pull/2081

**This PR adds a dial control to the numeric direction fields.**

Stacked on top of #12674 (rotate for points with `direction`).

_Features:_
- Users can drag the dial to change the value in steps of 1 (regardless of the specified [`increment` property](https://github.com/ideditor/schema-builder?tab=readme-ov-file#increment)
- Users can hold down shift to change values only based on the `increment`
- Clicking into the dial changes the value to where the user clicked
- The state is synced between map, fields, dial

Site note: The `increment` features works differently on the dial than on the `+/-` buttons. The buttons increase in +/- steps so 3 becomes 3+5=7. The dial jumps to either 5 or 10. — I think this inconsistency is OK. We keep the backward compatibility for the buttons and the dial works more natural like this IMO.

_Video:_

https://github.com/user-attachments/assets/c52a89dc-7eab-4326-a694-c7a179729e85

----

Replaces #12104 (rebased onto `rotate-point-direction` and recreated as a same-repo PR for GitHub stacked PRs).

This code is created using LLM tools. I reviewed it and iterated on it for a while to get into a shape that I find easy to understand and that has a clean split between the existing features and the new additions.

### DIFF
--- a/css/80_app.css
+++ b/css/80_app.css
@@ -1504,43 +1504,27 @@ button.preset-reset .label.flash-bg {
 }
 
 .form-field-has-direction-dial {
-    display: grid;
-    grid-template-columns: minmax(0, 1fr) auto;
-    grid-template-areas:
-        "label dial"
-        "input dial";
-    align-items: stretch;
-    column-gap: 0;
-}
-
-.form-field-has-direction-dial > .field-label {
-    grid-area: label;
-    border-radius: 4px 0 0 0;
-}
-
-.form-field-has-direction-dial > .form-field-input-wrap {
-    grid-area: input;
-    border-radius: 0 0 0 4px;
+    /* Keep hook class for direction fields; layout is handled in input row. */
 }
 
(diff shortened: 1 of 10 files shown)


tordans:
Sorry for the noise, https://github.com/openstreetmap/iD/pull/12104 is still the right PR. I wanted to try out stacked branches but those don't work with forks, so I tested this… but I don't think its worth to change our standard system just for stacked branches, so I closed this again.

PR #12679, opened 2026-08-04
drop unused and currently broken maprules code

tyrasd:
see https://github.com/openstreetmap/iD/issues/12313#issuecomment-4864249446

todo:
* [x] add to changelog


tyrasd:
@k-yle: can you give it a quick look? maybe I overlooked something?

test/spec/presets/index.js
```diff
@@ -401,7 +401,7 @@ describe('iD.presetIndex', function () {
         });
 
         it('makes only the external presets initially addable', function () {
-            var url = 'https://fakemaprules.io/fake.json';
+            var url = 'https://example.com/fake.json';
 
             iD.presetIndex().fromExternal(url, function(externalPresets) {
```
k-yle: this confused me a lot, because `fromExternal` doesn't exist. then i realised this whole block has `describe.skip`, so it's all dead and broken. so maybe it should all be deleted?

k-yle (approved):
i had a quick look and i think this is all good

test/spec/presets/index.js
tyrasd: good catch, I didn't even notice that :see_no_evil:

PR #12683, opened 2026-08-05
convert presetManager to typescript

k-yle:
this one is finally unblocked

### DIFF
--- a/modules/presets/index.ts
+++ b/modules/presets/index.ts
@@ -1,7 +1,12 @@
+import type { Fields, Geometry, Preset, PresetCategories, PresetDefaults, Presets } from '@openstreetmap/id-tagging-schema';
 import { prefs } from '../core/preferences';
 import { fileFetcher } from '../core/file_fetcher';
 import { locationManager } from '../core/location_manager';
-
+import type { TagDictionary } from '../util/object';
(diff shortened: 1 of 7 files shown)


modules/presets/index.ts
```diff
@@ -89,15 +136,14 @@ export function presetIndex() {
...
+    let newLocationSets: (presetField | presetPreset)[] = [];
 
     // Merge Fields
     if (d.fields) {
-      Object.keys(d.fields).forEach(fieldID => {
-        let f = d.fields[fieldID];
+      Object.entries(d.fields).forEach(([fieldID, rawField]) => {
 
-        if (f) {   // add or replace
-          f = presetField(fieldID, f);
```
k-yle, line 100: variable `f` is reassigned, so it refers to two different things depending on the line of code (object vs class). it's much easier to understand if we use a separate variable

modules/presets/index.ts
```diff
@@ -456,28 +499,17 @@ export function presetIndex() {
   };
 
 
-  function RibbonItem(preset) {
```
k-yle, line 459: this was not deleted, just moved to the top of the file

tyrasd (approved):

PR #12688, opened 2026-08-06
tweak osmPavedTags and osmSemipavedTags

matkoniecz:
better description - drop obvious "in osm", move surface=metal to expected smooth surface, restrict also semipaved to paths/roads, extend paved to paths...

(alternative would be listing also carpet, tartan and probably artificial_turf)

done in preparation of possible move of this info into iD tagging schema (see https://github.com/openstreetmap/id-tagging-schema/issues/2673 and https://github.com/openstreetmap/id-tagging-schema/issues/2623) as some changes in meaning were tempting - and I thought it would be better to be sure these are fine on iD side

### DIFF
--- a/modules/osm/tags.ts
+++ b/modules/osm/tags.ts
@@ -234,7 +234,7 @@ export const osmOneWayTags = merge(
     osmOneWayBiDirectionalTags,
 );
 
-// solid and smooth surfaces akin to the assumed default road surface in OSM
+// solid and smooth surfaces akin to the assumed default road or path surface
 export const osmPavedTags: TagDictionary<boolean> = {
     'surface': {
         'paved': true,
@@ -243,6 +243,7 @@ export const osmPavedTags: TagDictionary<boolean> = {
         'chipseal': true,
         'concrete:lanes': true,
         'concrete:plates': true,
+        'metal': true,
         'tiles': true
     },
     'tracktype': {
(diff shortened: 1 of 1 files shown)


tyrasd (approved):

modules/osm/tags.ts
```diff
@@ -234,7 +234,7 @@ export const osmOneWayTags = merge(
     osmOneWayBiDirectionalTags,
 );
 
-// solid and smooth surfaces akin to the assumed default road surface in OSM
```
k-yle, line 237: > [...] drop obvious "in osm",

that's often intentional:
-  [a "surface" in iD](https://grep.app/search?f.repo=openstreetmap%2FiD&q=context.surface) is different to [a "surface" in OSM](http://osm.wiki/Key:surface).
- [a "path" in iD](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/Paths) is also different to [a "path" in OSM](https://osm.wiki/Tag:highway=path).

modules/osm/tags.ts
```diff
@@ -259,7 +260,6 @@ export const osmSemipavedTags: TagDictionary<boolean> = {
         'sett': true,
         'paving_stones': true,
         'grass_paver': true,
-        'metal': true,
```
k-yle, line 262: huh, i learned something new. in `en-NZ`, a ["metal road"](https://docs.nzfoa.org.nz/live/nz-forest-road-engineering-manual-operators-guide/applying-aggregate-metalling/) means an unpaved road 😆

modules/osm/tags.ts
matkoniecz: yes definitely, though here I think this was not unclear?

modules/osm/tags.ts
matkoniecz: and AFAIK in Australia at least `surface=gravel` is not for gravel roads, these are `surface=compacted`

PR #12689, opened 2026-08-06
remove obliterated: prefix used 5 times worldwide

matkoniecz:
https://taginfo.openstreetmap.org/search?q=obliterated


matkoniecz:
it is also not documented at OSM Wiki from what I see

matkoniecz:
in other direction - it is probably too early for https://taginfo.openstreetmap.org/search?q=ruins and https://taginfo.openstreetmap.org/search?q=ruined

tyrasd (approved):

PR #12696, opened 2026-08-06
migrate validators to typescript (part 2)

k-yle:
another straightforward one

### DIFF
--- a/modules/validations/impossible_oneway.ts
+++ b/modules/validations/impossible_oneway.ts
@@ -6,14 +6,17 @@ import { osmFlowingWaterwayTagValues, osmOneWayTags, osmRoutableHighwayTagValues
 import { validationIssue, validationIssueFix } from '../core/validation';
 import { services } from '../services';
 import { utilCheckTagDictionary, utilObjectOmit } from '../util';
+import type { CreateValidator, Validator } from '../core/validation/models';
+import type { NodeId, osmNode, osmWay, WayId } from '../osm';
+import type { coreContext } from '../core';
 
-export function validationImpossibleOneway() {
+export const validationImpossibleOneway: CreateValidator = () => {
     const type = 'impossible_oneway';
 
     // ignore these oneway tags when checking for oneway-ness of highways
     const ignoreNonHighwayKeys = ['piste:type', 'aerialway', 'man_made', 'seamark:type', 'waterway'];
 
-    const validation = function checkImpossibleOneway(entity, graph) {
+    const validation: Validator = function checkImpossibleOneway(entity, graph) {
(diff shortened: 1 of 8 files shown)


modules/validations/impossible_oneway.ts
```diff
@@ -95,8 +98,8 @@ export function validationImpossibleOneway() {
             });
         }
 
-        function issuesForNode(way, nodeID) {
-            const isFirst = (nodeID === way.first()) ^ way.isOneWayBackwards();
+        function issuesForNode(way: osmWay, nodeID: NodeId) {
+            const isFirst = (nodeID === way.first()) !== way.isOneWayBackwards();
```
matkoniecz, line 102: wait

so if nodeID is the first in nodelist and `isOneWayBackwards` returns false, it evaluates to true as expected

but if `isOneWayBackwards` returns true, it evaluates to true when node is not first on list? So also second, third, fourth one? Maybe only last one should in such case?

modules/validations/impossible_oneway.ts
```diff
@@ -200,8 +203,8 @@ export function validationImpossibleOneway() {
                 loc: node.loc
             })];
 
-            function getReference(referenceID) {
-                return function showReference(selection) {
+            function getReference(referenceID: string) {
```
matkoniecz, line 206: maybe this reference can get its own type? Not share it with other strings?

matkoniecz (commented):
tried to be helpful but not sure have I succeeded and managed to leave review more useful than copilot one

modules/validations/impossible_oneway.ts
k-yle: i don't think that's necessary for a string which is only used in one place. we do this for things like `EntityId` because they're used 1000s of times and make inference [much more useful](https://github.com/openstreetmap/iD/blob/c2d7132065adbed05dbf980062cdbae5b36f8c45/modules/core/graph.ts#L35-L37)

modules/validations/impossible_oneway.ts
k-yle: this function is only ever given the first and last node, never midpoints, so this is safe. 

we could rename it, but i'm trying not to touch unrelated things to make these PRs easier to review. this line just needs to change because [a bitwise operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR) can only be used on numbers, not booleans.

matkoniecz: I guess comment/etc can be added separately

> this function is only ever given the first and last node, never midpoints, so this is safe.

oh right, that makes sense

tyrasd (approved):

PR #12704, opened 2026-08-09
Enable territory-aware address hints (with proper Hong Kong and Macao address formats)

### LINKED ISSUE #10906: Items in Hong Kong/Macau are wrongly getting Chinese address hints

Vectorial1024:
This supercedes the following PRs:

- Territory detection: #10914
- Format documentation: #10907

This fixes #10906.

-----------

While attempting to tackle #10906, I have previously attempted to "keep it fancy", which has been a grave mistake. the detection PR has nothing to confirm it is indeed working, and the format PR has no mechanism to demonstrate the updated formats. As such, both PRs could not be merged. I am deeply sorry about the undue progress blocking.

-----------

### Territory Detection (10914)

As guided by https://github.com/openstreetmap/iD/issues/10899#issuecomment-2745909687 , this PR enables the editor to look at the territory-level country code. It turns out the editor only looks at sovereign country codes to generate address format hints, which unfortunately means it omits the formats of Hong Kong and Macao, both non-sovereign and use their colonial address hint formats distinct from the suzerain PRC format.

### Format Documentation (10907)

While reading https://github.com/openstreetmap/iD/issues/10899#issuecomment-2745909687 , it is discovered that the editor does not currently know about Hong Kong's address format. Also, the Macao address format is inaccurate. Both should be documented/updated.

Address levels above "city" are removed because both are essentially city-states. Postcodes are omitted because both ([Hong Kong](https://en.wikipedia.org/wiki/Postal_codes_in_Hong_Kong), [Macao](https://www.travelchinaguide.com/essential/area_zip/macau.htm)) do not use them. Detailed indoor addresses are added owing to the dense urban development in both cities.

------------

Pinging past participants and interested parties: @Kovoschiz , @k-yle , @matkoniecz , @novolife, @tyrasd

### DIFF
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -371,7 +371,8 @@ export function uiFieldAddress(field, context) {
                 countryCode = t('intro.graph.countrycode');
             } else {
                 var center = extent.center();
-                countryCode = countryCoder.iso1A2Code(center);
+                // even if we do not have data for this territory/country, the editor will generate a default format for us.
+                countryCode = countryCoder.iso1A2Code(center, { level: 'territory' });
             }
             if (countryCode) {
                 _countryCode = countryCode.toLowerCase();
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -383,6 +383,8 @@ export function uiFieldAddress(field, context) {
                     countryCoder.iso1A2Code(center),
                 ];
             }
+            // country codes can be null (e.g. unclaimed ocean); we don't want that.
+            countryCodes = countryCodes.filter((item) => item !== null);
             if (countryCodes) {
                 _countryCodes = countryCodes.map((x) => x.toLowerCase());
                 updateForCountryCodes();
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -383,8 +383,8 @@ export function uiFieldAddress(field, context) {
                     countryCoder.iso1A2Code(center),
                 ];
             }
-            // country codes can be null (e.g. unclaimed ocean); we don't want that.
-            countryCodes = countryCodes.filter((item) => item !== null);
+            // country codes can be null/undefined (e.g. unclaimed ocean); we don't want that.
+            countryCodes = countryCodes.filter((item) => !!item);
             if (countryCodes) {
                 _countryCodes = countryCodes.map((x) => x.toLowerCase());
                 updateForCountryCodes();
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -23,7 +23,7 @@ export function uiFieldAddress(field, context) {
 
     var _entityIDs = [];
     var _tags;
-    var _countryCode;
+    var _countryCodes = []; // sometimes we want to check multiple address hints (e.g. HK vs CN, and CK vs NZ)
     var _addressFormats = [{
         format: [
             ['housenumber', 'street'],
@@ -196,18 +196,23 @@ export function uiFieldAddress(field, context) {
     }
 
 
-    function updateForCountryCode() {
+    function updateForCountryCodes() {
 
-        if (!_countryCode) return;
+        if (!_countryCodes) return;
 
+        // find the country with a usable format
         var addressFormat;
-        for (var i = 0; i < _addressFormats.length; i++) {
-            var format = _addressFormats[i];
-            if (!format.countryCodes) {
-                addressFormat = format;   // choose the default format, keep going
-            } else if (format.countryCodes.indexOf(_countryCode) !== -1) {
-                addressFormat = format;   // choose the country format, stop here
-                break;
+        for (var countryIter = 0; countryIter < _countryCodes.length && !addressFormat; countryIter++) {
+            var currentCountryCode = _countryCodes[countryIter];
+
+            for (var i = 0; i < _addressFormats.length; i++) {
+                var format = _addressFormats[i];
+                if (!format.countryCodes) {
+                    addressFormat = format;   // choose the default format, keep going
+                } else if (format.countryCodes.indexOf(currentCountryCode) !== -1) {
+                    addressFormat = format;   // choose the country format, stop here
+                    break;
+                }
             }
         }
 
@@ -365,18 +370,22 @@ export function uiFieldAddress(field, context) {
         var extent = combinedEntityExtent();
 
         if (extent) {
-            var countryCode;
+            var countryCodes = [];
             if (context.inIntro()) {
                 // localize the address format for the walkthrough
-                countryCode = t('intro.graph.countrycode');
+                countryCodes = [t('intro.graph.countrycode')];
             } else {
                 var center = extent.center();
-                // even if we do not have data for this territory/country, the editor will generate a default format for us.
-                countryCode = countryCoder.iso1A2Code(center, { level: 'territory' });
+                // we try to find the territory-level address format
+                // if we dn't have that, we fall back to (sovereign) country-level address format
+                countryCodes = [
+                    countryCoder.iso1A2Code(center, { level: 'territory' }),
+                    countryCoder.iso1A2Code(center),
+                ];
             }
-            if (countryCode) {
-                _countryCode = countryCode.toLowerCase();
-                updateForCountryCode();
+            if (countryCodes) {
+                _countryCodes = countryCodes.map((x) => x.toLowerCase());
+                updateForCountryCodes();
             }
         }
     }
@@ -429,8 +438,9 @@ export function uiFieldAddress(field, context) {
 
 
     function getLocalPlaceholder(key) {
-        if (_countryCode) {
-            var localkey = key + '!' + _countryCode;
+        if (_countryCodes) {
+            var firstCountryCode = _countryCodes[0];
+            var localkey = key + '!' + firstCountryCode;
             var tkey = addrField.hasTextForStringId('placeholders.' + localkey) ? localkey : key;
             return addrField.t('placeholders.' + tkey);
         }
--- a/modules/ui/fields/address.js
+++ b/modules/ui/fields/address.js
@@ -201,18 +201,14 @@ export function uiFieldAddress(field, context) {
         if (!_countryCodes) return;
 
         // find the country with a usable format
-        var addressFormat;
-        for (var countryIter = 0; countryIter < _countryCodes.length && !addressFormat; countryIter++) {
-            var currentCountryCode = _countryCodes[countryIter];
-
-            for (var i = 0; i < _addressFormats.length; i++) {
-                var format = _addressFormats[i];
-                if (!format.countryCodes) {
-                    addressFormat = format;   // choose the default format, keep going
-                } else if (format.countryCodes.indexOf(currentCountryCode) !== -1) {
-                    addressFormat = format;   // choose the country format, stop here
-                    break;
-                }
+        const defaultAddressFormat = _addressFormats.find((item) => !item.countryCodes);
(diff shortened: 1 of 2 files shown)


k-yle (changes requested):
same code review feedback as  https://github.com/openstreetmap/iD/pull/10907#pullrequestreview-2709776125 and  #10904:

this will break the logic for other "territories" such as [CK](https://r12a.github.io/app-subtags/index?lookup=CK), which should fallback to the "parent" (in that case, NZ)

Vectorial1024:
Interesting. It's very likely I have missed the previous comment on CK. Will review soon.

Vectorial1024:
Code updated to allow multi-country address hint lookup. Please review.

Vectorial1024:
Item in Hong Kong: https://pr-12704--ideditor.netlify.app/#disable_features=boundaries&map=18.42/22.38287/114.19198&background=Bing&id=n12131104122

image

Item in PRC China (Shen Zhen): https://pr-12704--ideditor.netlify.app/#disable_features=boundaries&map=17.98/22.55237/114.10701&background=Bing&id=n5768010001

image

Item in Cook Island (Avarua): https://pr-12704--ideditor.netlify.app/#disable_features=boundaries&map=18.26/-21.20729/-159.77054&background=Bing&id=n6324434485

image

Item in New Zealand (Wellington): 
https://pr-12704--ideditor.netlify.app/#disable_features=boundaries&map=20.00/-41.28980/174.77565&background=Bing&id=n4633729858

image

-------------

HK-CN case is good. However, I notice the CK-NZ case seems to have different address format. Is this working correctly?

modules/ui/fields/address.js
```diff
@@ -365,17 +370,22 @@ export function uiFieldAddress(field, context) {
...
+                countryCodes = [t('intro.graph.countrycode')];
             } else {
                 var center = extent.center();
-                countryCode = countryCoder.iso1A2Code(center);
+                // we try to find the territory-level address format
+                // if we dn't have that, we fall back to (sovereign) country-level address format
+                countryCodes = [
+                    countryCoder.iso1A2Code(center, { level: 'territory' }),
+                    countryCoder.iso1A2Code(center),
+                ];
```
k-yle, line 380: this will crash if countryCoder returns `null`. e.g. if someone adds an address to a buoy in the middle of the ocean. 

so we probably need something like this:
```suggestion
                ].filter(Boolean);
```

modules/ui/fields/address.js
```diff
@@ -196,18 +196,23 @@ export function uiFieldAddress(field, context) {
...
+            for (var i = 0; i < _addressFormats.length; i++) {
+                var format = _addressFormats[i];
+                if (!format.countryCodes) {
+                    addressFormat = format;   // choose the default format, keep going
+                } else if (format.countryCodes.indexOf(currentCountryCode) !== -1) {
+                    addressFormat = format;   // choose the country format, stop here
+                    break;
+                }
             }
         }
```
k-yle, line 213: > HK-CN case is good. However, I notice the CK-NZ case seems to have different address format. Is this working correctly?

that's because of a bug in this loop. maybe we could simplify it to something like this:

```suggestion
        let addressFormat;
        for (const code of _countryCodes) {
            addressFormat = _addressFormats.find(format =>
                format.countryCodes?.includes(code));
            if (addressFormat) break;
        }
        addressFormat ||= _addressFormats.find(format => !format.countryCodes);
```

 to test it, CK should have the same format as NZ. but HK should be different to CN.

modules/ui/fields/address.js
Vectorial1024: Done in d2412990e9ddcd329424ee0bba405e417c91cda9

modules/ui/fields/address.js
Vectorial1024: The past code was confusing, but now I can see the idea is basically `get(first available) ?? default`.

Vectorial1024: Done in 807714c62360fca18639611333ad30d2db4b5541 ; wait for deploy/test.

Vectorial1024:
New code deployed.

--------------

Item in Cook Island (Avarua): https://pr-12704--ideditor.netlify.app/#disable_features=boundaries&map=18.26/-21.20729/-159.77054&background=Bing&id=n6324434485

image

Item in New Zealand (Wellington):
https://pr-12704--ideditor.netlify.app/#disable_features=boundaries&map=20.00/-41.28980/174.77565&background=Bing&id=n4633729858

image

Item in unclaimed area (can't think of any; just make a new node):

image

Previous HK/CN cases unchanged (passed).

---------------

All cases passed; thanks for the guidance. Ready for merging.

modules/ui/fields/address.js
```diff
@@ -196,17 +196,18 @@ export function uiFieldAddress(field, context) {
     }
 
 
-    function updateForCountryCode() {
+    function updateForCountryCodes() {
 
-        if (!_countryCode) return;
+        if (!_countryCodes) return;
```
k-yle: this does nothing, since `_countryCodes` is always truthy

modules/ui/fields/address.js
```diff
@@ -365,17 +366,24 @@ export function uiFieldAddress(field, context) {
...
+                    countryCoder.iso1A2Code(center, { level: 'territory' }),
+                    countryCoder.iso1A2Code(center),
+                ];
             }
-            if (countryCode) {
-                _countryCode = countryCode.toLowerCase();
-                updateForCountryCode();
+            // country codes can be null/undefined (e.g. unclaimed ocean); we don't want that.
+            countryCodes = countryCodes.filter((item) => !!item);
+            if (countryCodes) {
```
k-yle: same as https://github.com/openstreetmap/iD/pull/12704#discussion_r3843610224, this `if`-statement's condition is always true

modules/ui/fields/address.js
```diff
@@ -196,17 +196,18 @@ export function uiFieldAddress(field, context) {
...
-        var addressFormat;
-        for (var i = 0; i < _addressFormats.length; i++) {
-            var format = _addressFormats[i];
-            if (!format.countryCodes) {
-                addressFormat = format;   // choose the default format, keep going
-            } else if (format.countryCodes.indexOf(_countryCode) !== -1) {
-                addressFormat = format;   // choose the country format, stop here
+        // find the country with a usable format
+        const defaultAddressFormat = _addressFormats.find((item) => !item.countryCodes);
+        console.assert(defaultAddressFormat !== undefined); // if undefined, then the config files are wrong.
```
k-yle, line 205: i wouldn't bother with this. if you want a concise solution: https://github.com/openstreetmap/iD/pull/12704#discussion_r3810118579

or we could add `if (!_addressFormats) return;` at the end of this block?

modules/ui/fields/address.js
Vectorial1024: Done in cbb6ca48852a8078bccfdae383d92b57a7b435eb

modules/ui/fields/address.js
Vectorial1024: Done in cbb6ca48852a8078bccfdae383d92b57a7b435eb

modules/ui/fields/address.js
Vectorial1024: The way I see this, the default value code may be clumsy, but at least we are clear there is a default value.

I would say even the proposed code is not defensive enough; what if the config array is somehow empty or is somehow malformed (i.e. no default format), and then we should detect that and stop early.

Not sure which "end of block" we are referring to. Address format lookup should continue through the array/loop until we run out of options, so that e.g. CK locations (`undefined`) can continue with NZ lookup (`exists`).

k-yle: > what if the config array is somehow empty or is somehow malformed

this would be noticed in the code review, and detected by typescript[^1], so it's impossible in practice. 

[`console.assert(x, y)`](https://developer.mozilla.org/en-US/docs/Web/API/console/assert_static) is also not a runtime assertion. It's basically just `if (!x) console.error(y)`.


[^1]: once the remaining TS PRs are merged.

Vectorial1024: Thanks for the guidance on `console.assert`.

Today I learned there is this TS PR concept in this repo; should I wait after TS PR or should I continue with this PR first? Is this repo being rewritten in TS that I was not aware of?

k-yle: yes please continue with this PR, it's unfortunate that it's been stuck for so long. This thread is my last suggestion, I think it could be merged after that.

as an example, 6888e5c99eb3682fd6c87f75c9e144dd263f6e8c is how I'd simplify it

matkoniecz: > Is this repo being rewritten in TS

and yes, it is being migrated to TS after some bugs in releases that would likely be impossible with TS

Vectorial1024: Finally got around; code changes done in e753be65c74b7e96c996e85b780a2626e6e14907

Vectorial1024:
Should be good now. Please review.

k-yle (approved):
thanks, and sorry that these PRs were stuck for over a year 😭 

in terms of code, this is ready to merge imo.

---
in terms of tagging: seems reasonable per [taginfo .hk](https://taginfo.geofabrik.de/asia:china:hong-kong/search?q=addr%3A) and [taginfo .mo](https://taginfo.geofabrik.de/asia:china:macau/search?q=addr%3A). It'd be nice to support [`addr:place:zh`](https://taginfo.geofabrik.de/asia:china:hong-kong/keys/addr%3Aplace%3Azh) too, but that's obviously out of scope.

I assume that the discussion on discord supported this change? 

cc'ing @Kovoschiz - in case you're only subscribed to the original PR

matkoniecz:
> I assume that the discussion on discord supported this change?

If I understood it correctly, at time I forwarded proposed changes as they were then: yes

It was Hing Kong Discord IIRC.

matkoniecz:
Not sure should I resubmit it (has format changed?) or ask other communities.

Vectorial1024:
I understand your concerns, but back then when experimenting/determining what address tags iD can support, I quickly found out iD doesn't have translations for all the various fancy tags that HK/MO is using. It would show sth like "no translation key for xxx".

With this, I decided to just supply a "mostly correct" format for now, and then leave the extra fancy tags as future work.

I personally don't follow the address discussion too closely, but the general idea (at least for HK) is that, due to historical reasons, there is not a single "uniform" address format for the entire HK. These non-uniformity is still pending discussion (I think), but there is a useful subset of "these will always work + iD has these" address tags.

Vectorial1024:
Also, specific to the various `addr:*:zh`/`addr:*:en` tags, given that bilingualism is globally rare, it may be passable for iD to just ignore it for the time being. So far I see even Quebec (Canada) only has French names, and does not have English names.

If iD is ever so inclined to deal with the various language extensions, it better deal with it properly as a separate issue/PR. Things are complicated.

k-yle (approved):
well as long as the HK and MO communities are happy then I think this is ready to merge. Maybe we should wait a few more days for final feedback?

Vectorial1024:
Me thinks LGTM, but we can still wait a bit, just in case.

PR #12721, opened 2026-08-11
fix `imagery_used` set to None

### LINKED ISSUE #12717: imagery_used sometimes set to None since 2.42.0 even when imagery is used
### URL

https://www.openstreetmap.org/edit#map=19/48.037934/7.015766

1. Change the background to Bing Maps Aerial (the default works fine)
2. Make any temporary modifications
3. Save to open the upload menu and look at the imagery_used tag

OSMCha filter with imagery_used=None
https://osmcha.org/changesets/187254000?filters=%7B%22imagery_used%22%3A%5B%7B%22label%22%3A%22None%22%2C%22value%22%3A%22None%22%7D%5D%7D
Already 723 since 2.42.0 release

2.42.0

Firefox

k-yle:
Closes #12717

fixes a regression from #12561, currently the imagery-source object is being [spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), which creates a new object while rendering.

and that breaks existing code which [relies on reference-equality](https://github.com/openstreetmap/iD/blob/524d01baf65bd136c3d60098831222e2299fc088/modules/renderer/background.js#L110).

typescript would prevent this, since it bans spreading class instances (as discussed in https://github.com/openstreetmap/iD/pull/12641#discussion_r3673806034). but there are a lot of blockers before this file can be converted to TS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
+* fix `imagery_used` set to None ([#12721], thanks [@k-yle])
 #### :earth_asia: Localization
 #### :hourglass: Performance
 #### :mortar_board: Walkthrough / Help
@@ -50,6 +51,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Drop code previously responsible for _maprules_ integration (which has been defunct for a while now) ([#12679])
 
 [#12679]: https://github.com/openstreetmap/iD/pull/12679
+[#12721]: https://github.com/openstreetmap/iD/pull/12721
 
 # 2.42.0
 ##### 2026-Aug-10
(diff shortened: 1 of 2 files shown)


Nekzuris (commented):
Tested locally, it's fixed 👍

tyrasd (approved):
IIRC, I wanted to avoid modifying the imagery source objects in place, as this information is only needed for the condensed rendering of the layers in the UI, and should not have side effects, ideally.

I guess doing it without using the spread operator would have been more elegant:

```js
                        sources.push({
                            source,
                            prefix,
                            suffix,
                            variants: [{ variant, source }]
                        });
// ... else ...
                sources.push({ source });
```

But in order to address the current bug, your fix is working just as well, thanks!

PS: I'm wondering if the `===` checks could be avoided by instead checking for the imagery's `.id` instead?

k-yle:
> PS: I'm wondering if the `===` checks could be avoided by instead checking for the imagery's `.id` instead?

yeah that's an alternative, but we would still need to remove the spread operator in the future, if we want to convert this file to an ES6 class:

```js
class MyClass {
    a() {}
    b = 1
}

({ ...new MyClass() })
// this prints { b = 1 }, property `a` is missing. because classes can't/shouldn't be spread
```

PR #12722, opened 2026-08-11
fix broken mapillary features / traffic signs

### LINKED ISSUE #12741: Mapillary 'Traffic signs' and 'map features' overlays no longer work
### URL

_No response_

_No response_

The Mapillary 'map features' and 'traffic signs' overlays no longer display the relevant imagery. Clicking on any of the on-screen features will display only a blank window. If Mapillary street-level imagery has been used in the session, the window will instead show the most recent Mapillary image viewed, regardless of location.

Image

Development version at ideditor.netlify.app, Released version at openstreetmap.org/edit

2.42.0

Chrome

k-yle:
Closes #12741

clicking on a traffic sign or "mapillary feature" currently results in a blank image:

image

seems to be caused by dc9d3cf5754cbf46a79cfc6a49bd7403132c7fb0 which changed the arguments (`context, imageId` -> `image`). typescript would prevent this, I opened a separate PR for that: #12725

### DIFF
--- a/modules/svg/mapillary_map_features.js
+++ b/modules/svg/mapillary_map_features.js
@@ -61,21 +61,21 @@ export function svgMapillaryMapFeatures(projection, context, dispatch) {
 
         context.map().centerEase(d.loc);
 
-        const selectedImageId = service.getActiveImage() && service.getActiveImage().id;
+        const selectedImage = service.getActiveImage();
 
         service.getDetections(d.id).then(detections => {
             if (detections.length) {
-                const imageId = detections[0].image.id;
-                if (imageId === selectedImageId) {
+                const { image } = detections[0];
+                if (selectedImage && image.id === selectedImage.id) {
                     service
                         .highlightDetection(detections[0])
-                        .selectImage(context, imageId);
+                        .selectImage(image);
                 } else {
                     service.ensureViewerLoaded(context)
                         .then(function() {
                             service
                                 .highlightDetection(detections[0])
-                                .selectImage(context, imageId)
+                                .selectImage(image)
                                 .showViewer(context);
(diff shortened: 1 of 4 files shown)


matkoniecz:
>  I opened a separate PR for that: #99999

placeholder probably was supposed to be replaced (feel free to delete or hide this comment after fix)

danieldegroot2:
Fixes #12741

modules/svg/mapillary_map_features.js
```diff
@@ -61,21 +61,21 @@ export function svgMapillaryMapFeatures(projection, context, dispatch) {
 
         context.map().centerEase(d.loc);
 
-        const selectedImageId = service.getActiveImage() && service.getActiveImage().id;
```
matkoniecz, line 64: this is unrelated refactoring, not load bearing part, right?

test/spec/services/mapillary.js
```diff
@@ -44,7 +44,7 @@ describe('iD.serviceMapillary', function() {
     describe('#reset', function() {
         it('resets cache', function() {
             mapillary.cache().foo = 'bar';
-            mapillary.selectImage(context, { key: 'baz', loc: [10,0] });
+            mapillary.selectImage({ key: 'baz', loc: [10,0] });
```
tyrasd, line 47: a bit odd that this did not cause issues here. 😅  I think it `.selectImage` doesn't really do anything here (as long as the test does not also check for the respective `getActiveImage()` result), but let's address that separately.

tyrasd (approved):
Sorry for the oversight. Yes, typescript would have been super useful to prevent this. 👍

PR #12724, opened 2026-08-11
increase the colour contrast of `--link-color`

k-yle:
as mentioned in https://github.com/openstreetmap/iD/discussions/12714#discussioncomment-17969751, in light mode `--link-color` doesn't meet the [WCAG AA colour constrast guidelines](https://en.wikipedia.org/wiki/Web_Content_Accessibility_Guidelines).

Before:
- Light Mode: `#7092ff` $`{\color{#7092ff} \blacksquare}`$ on $`{\color{#ffffff} \blacksquare}`$ `#ffffff` [fail](https://webaim.org/resources/contrastchecker/?fcolor=7092ff&bcolor=FFFFFF)
- Dark Mode: `#7b92db` $`{\color{#7b92db} \blacksquare}`$ on $`{\color{#212529} \blacksquare}`$ `#212529` [pass](https://webaim.org/resources/contrastchecker/?fcolor=7b92db&bcolor=212529)

After:
- Light Mode: `#516fd9` $`{\color{#516fd9} \blacksquare}`$ on $`{\color{#ffffff} \blacksquare}`$ `#ffffff` [pass](https://webaim.org/resources/contrastchecker/?fcolor=516fd9&bcolor=FFFFFF)
- Dark Mode: _unchanged_

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>image
 <td>image

</table>

apart from personal preference, one minor disadvantage is that it no longer matches the light blue text used on parts of the main osm website:

image

cc @hlfan for feedback since you worked on dark mode and the osm website

### DIFF
--- a/css/10_colors.css
+++ b/css/10_colors.css
@@ -12,8 +12,8 @@
     --outset-color: #fcfcfc; /* brighter than --border-color with white */
     --shadow-color: #bbb; /* mix 91.5% of --border-color with black */
     --text-color: #333;
-    --link-color: #7092ff;
-    --link-focus-color: #597be7;
+    --link-color: #516fd9;
+    --link-focus-color: #3f6ae0;
     --muted-text-color: #aaa;
     --transparent-text-color: #2223;
     --transparent-text-emphasis-color: #0007;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 # Unreleased (2.43.0-dev)
 
 #### :sparkles: Usability & Accessibility
+* Increase the colour contrast of the blue text used for links and preset fields ([#12724], thanks [@k-yle])
 * Render diameter and radius tags when a node is selected ([#9732], thanks [@k-yle])
 #### :scissors: Operations
 #### :camera: Street-Level
@@ -52,6 +53,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 
 [#9732]: https://github.com/openstreetmap/iD/pull/9732
 [#12679]: https://github.com/openstreetmap/iD/pull/12679
+[#12724]: https://github.com/openstreetmap/iD/pull/12724
 
 
 # 2.42.1


hlfan:
> apart from personal preference, one minor disadvantage is that it no longer matches the light blue text used on parts of the main osm website

I can't quite believe the bootstrap colors don't meet these guidelines; I'm pretty sure bootstrap has some link state colors we can base these on, if anyone wants to beat me to looking this up: feel free to.

tyrasd:
I believe the blue highlight color was chosen to fit the color used on osm.org on buttons. However, maybe it would be wiser to use the color the osm website for links instead? That color is a bit darker and does not rely on the button having a large background area for ok contrast?

> I can't quite believe the bootstrap colors don't meet these guidelines

Isn't the osm website overriding the standard bootstrap colors (see https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/stylesheets/parameters.scss#L5). By default the bootstrap-styled buttons would use a much more saturated blue.

For links, the osm website is using [`#24d` (`#2244dd`)](https://github.com/openstreetmap/openstreetmap-website/blob/master/app/assets/stylesheets/parameters.scss#L14-L15) in light mode and `#a9beff` in dark mode (I assume this is somehow calculated from the '#24d' color automatically, since I cannot find where it is defined in the code?).

This is how it would look like in light mode:



It's arguably quite a bit more saturated/dark (especially on the bold text for section headings), but maybe that's fine, and just needs a little bit of getting used to?

And for completeness, this is how the dark mode would look like if we would also change it to match the link color of the osm-website:

hlfan:
So the thing I was worried about is already documented:

https://github.com/openstreetmap/openstreetmap-website/blob/85bb488d5ba61a7c3fd9f786cbb9d4ec781fb64a/app/assets/stylesheets/common.scss#L404-L407

Since we probably still want a different link focus color, we can start from the `#24d` (`#2244dd` with ratio 7.14) base and let bootstraps default [tinting & shading](https://github.com/twbs/bootstrap/blob/main/site/src/content/docs/customize/sass.mdx#colors) functions guide this:
- for the [light hover color](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L456) we can take `#23a` (`#2233aa` with ratio 9.89),
- I personally would round the [bootstrap-calculated](https://github.com/twbs/bootstrap/blob/main/scss/_variables-dark.scss#L53) `#a9beff` to `#9af` (`#99aaff` with ratio 7.01)
- and the [dark focus color](https://github.com/twbs/bootstrap/blob/main/scss/_variables-dark.scss#L54) to `#abf` (`#aabbff` with ratio 8.27).

And then there's the active map control button style, which inverts the contrast relation:

maxelliotmingail-afk (commented):
The colour change itself is covered by the current green checks, but the changelog reference is off by one: it adds `[#12723]` and maps that label to `/pull/12723`. GitHub has no PR 12723 in this repository; this contribution is PR 12724.

Please change both occurrences to `[#12724]` / `https://github.com/openstreetmap/iD/pull/12724` so the release notes retain a working attribution link.

Reviewed by Max Mingail with AI assistance. Max verified the diff, both PR numbers, the nonexistent target and current check state before commenting.

k-yle:
thanks for spotting that. weird that issue number #12723 was never allocated. or maybe that issue was deleted

css/10_colors.css
```diff
@@ -12,8 +12,8 @@
     --outset-color: #fcfcfc; /* brighter than --border-color with white */
     --shadow-color: #bbb; /* mix 91.5% of --border-color with black */
     --text-color: #333;
-    --link-color: #7092ff;
-    --link-focus-color: #597be7;
+    --link-color: #516fd9;
+    --link-focus-color: #3f6ae0;
```
hlfan, line 16: Why stop at just above AA contrasts and not go for AAA contrasts?

matkoniecz: @k-yle

k-yle: mostly because the current change is already quite jarring, if you're gotten used to pale blue text in the preset sidebar, which has been around for at least 5 years, maybe longer.

for comparison:


<table>
<tr>
 <td>Before
 <td>AA
 <td>AAA
<tr>
 <td>image
<td>image
 <td>image

</table>

matkoniecz: AAA does not seem to be bad or jarring to me? If anything now that it was pointed out I get irritated by unreadability of a current version

maybe it would be in action in actual interface it would be different

but at least it should not block upgrading to AA, I guess

k-yle: i was expecting a lot of complaints given how noticable this is, but if people don't mind then i guess we could change it to AAA (which i believe [is `#2b4dc5`](https://webaim.org/resources/contrastchecker/?fcolor=2b4dc5&bcolor=ffffff)). 

maybe i'll post about this in the #iD channel on slack to see what people think

matkoniecz:
Also, maybe we should report this problem upstream (especially in light of https://github.com/openstreetmap/openstreetmap-website/blob/85bb488d5ba61a7c3fd9f786cbb9d4ec781fb64a/app/assets/stylesheets/common.scss#L404-L407) or someone(TM) can make a pull request to osm.org itself?

But in this case I think that readability is more important than color matching, even if there would be refusal to fix colors on the website itself.

But at least attempting to harmonize changes would be nice...

For now I do not see anything for https://github.com/openstreetmap/openstreetmap-website/issues?q=is%3Aissue%20state%3Aopen%20accessibility or https://github.com/openstreetmap/openstreetmap-website/issues?q=is%3Aissue%20state%3Aopen%20bootstrap%20colour or https://github.com/openstreetmap/openstreetmap-website/issues?q=is%3Aissue%20state%3Aopen%20bootstrap%20color

https://github.com/openstreetmap/openstreetmap-website/issues/2716 and https://github.com/openstreetmap/openstreetmap-website/issues/183 are kind of related

tordans:
Did not read all but AAA https://github.com/openstreetmap/iD/pull/12724#discussion_r3967395776 looks good to me so why not that…

css/10_colors.css
matkoniecz: screen-2026-09-09-15-16-06

matkoniecz:
@k-yle have you considered increasing contrast to AAA level? (maybe it was increased already, but if I read commit history correctly it was not yet)

Is there an upstream issue for insufficient contrast at osm.org? Is anyone feeling competent to create it? Should I do it?

zekefarwell:
The color change looks fine to me.  

It may not make much of a difference in this case, but I thought it worth mentioning here for future reference that a new Advanced Perceptual Contrast Algorithm (APCA) that will likely be adopted with WCAG 3.0 (currently under development). This new algorithm does a much better job matching how we actually perceive readability, factoring in text size and weight as well as color contrast.

https://apcacontrast.com/
https://www.achecks.org/apca-accessible-colour-contrast-checker/
https://www.accessibilitychecker.org/blog/apca-advanced-perceptual-contrast-algorithm/
https://accessibe.com/blog/wcag-3-point-0

k-yle:
I changed it to a colour that meets the AAA requirements: [$`{\color{#2b4dc5} \blacksquare}`$,  `#2b4dc5`](https://webaim.org/resources/contrastchecker/?fcolor=2b4dc5&bcolor=ffffff).

also posted about this on the OSM US slack and got ~~14~~ 21 👍s, so I'm withdrawing my earlier concerns

css/10_colors.css
k-yle: done, conclusion: https://github.com/openstreetmap/iD/pull/12724#issuecomment-5759564574

PR #12727, opened 2026-08-11
tidy & deduplicate geometry code

k-yle:
1. deduplicating some code from [#11774](https://github.com/openstreetmap/iD/pull/11774) which i want to re-use in [#12728](https://github.com/openstreetmap/iD/pull/12728), but i'm not going to copy-paste it for the 3rd time...
2. tidying some changes from [#12613](https://github.com/openstreetmap/iD/pull/12613): `getReflectAxis` and `useLongAxis` are specific to the reflect action, they're not re-used anywhere else.

### DIFF
--- a/modules/actions/reflect.ts
+++ b/modules/actions/reflect.ts
@@ -1,22 +1,25 @@
-import type { Vec2 } from '@rapideditor/location-conflation';
 import type { coreGraph } from '../core';
 import type { Action } from '../core/history';
-import { geoGetSmallestSurroundingRectangle, geoVecInterp, geoVecLength } from '../geo';
+import { geoGetAxis, geoVecInterp, type Vec2 } from '../geo';
 import type { Projection } from '../geo/raw_mercator';
 import { utilGetAllNodes } from '../util';
 import type { EntityId } from '../osm';
 
+export interface ActionReflect extends Action {
+    useLongAxis: GetSet<this, boolean>;
+    getReflectAxis(graph: coreGraph): [Vec2, Vec2] | undefined;
+}
 
 /* Reflect the given area around its axis of symmetry */
-export function actionReflect(reflectIds: EntityId[], projection: Projection) {
+export function actionReflect(reflectIds: EntityId[], projection: Projection): ActionReflect {
     var _useLongAxis = true;
 
 
-    var action: Action = function(graph, t) {
+    const action: ActionReflect = function(graph, t) {
         if (t === null || t === undefined || !isFinite(t)) t = 1;
         t = Math.min(Math.max(+t, 0), 1);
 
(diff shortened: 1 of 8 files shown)


matkoniecz:
> which i want to re-use in #99999

it looks like it is a placeholder

modules/actions/reflect.ts
```diff
@@ -1,22 +1,25 @@
...
 import type { Projection } from '../geo/raw_mercator';
 import { utilGetAllNodes } from '../util';
 import type { EntityId } from '../osm';
 
+export interface ActionReflect extends Action {
+    useLongAxis: GetSet<this, boolean>;
+    getReflectAxis(graph: coreGraph): [Vec2, Vec2] | undefined;
+}
 
 /* Reflect the given area around its axis of symmetry */
```
matkoniecz, line 13: this comment was here already, but from what I see it will happily flip asymmetrical objects

modules/geo/vector.ts
```diff
@@ -93,6 +93,16 @@ export function geoVecCross(a: Vec2, b: Vec2, origin?: Vec2) {
 }
 
 
+/**
+ * given a line between `o` and `b`, this function determines how far along
+ * this line point `a` is, as a percentage (0-1 unless outside the range of
```
matkoniecz, line 98:
```suggestion
 * this line point `a` is, as a percentage (0-1 if a is within the range of
```

?

matkoniecz: at least for me negations makes things less comprehensible

matkoniecz (commented):
I tested it a bit and things I tested still work

modules/geo/geom.ts
```diff
@@ -314,6 +318,37 @@ export function geoGetSmallestSurroundingRectangle(points: Vec2[]) {
...
+/**
+ * returns the endpoints of an axis of symmetry of the `points` bounding rect.
+ * By default this is the long axis, unless `useLongAxis` is false.
+ */
+export function geoGetAxis(points: Vec2[], useLongAxis = true): [Vec2, Vec2] | undefined {
+    const ssr = geoGetSmallestSurroundingRectangle(points);
+    if (!ssr) {
+        // if there are exactly 2 or 3 points, then use the outermost points
+        const hull = d3_polygonHull(points) || points;
+        if (hull.length === 2) return [hull[0], hull[1]];
```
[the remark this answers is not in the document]
matkoniecz: How can you get two coincident points into that function, from iD interface?

modules/geo/vector.ts
matkoniecz, line 98: is `a` project onto `ob` line by means of constructing line perpendicular to `ob` and crossing `a`? Or some other way? Maybe mention it in the comment?

Issue #6079, opened 2019-03-21
Replace singular walkthrough with a library of short tutorials 


quincylvania:
iD's walkthrough has a number of areas for improvement:

- It can take 10-20 minutes to complete. New mappers might be turned off by this.
- It can be too expansive for many cases. Someone volunteering at a roads-only mapathon doesn't need training on buildings or POIs.
- It can only cover so much. There are many mapping patterns and iD tricks that could be taught with a few minutes of training but don't make sense to teach everyone.
- It's framed as a getting-started lesson. If we add new content for modern best practices, it's unlikely that existing mappers will see it.

To address these, I suggest replacing the monolithic walkthrough with a library of bite-sized lessons. Each lesson would typically be around 2-5 minutes long and would cover a wide range of topics at various difficulty levels. We could develop a common framework for defining the content of the lessons so iD contributors could create their own fairly easily. Users could keep track of what lessons they've completed and be notified when new content is available so their knowledge stays up to date.

Some example lessons:
- Getting started — adding your first feature
- Working with fields and raw tags
- Using the square, straighten, and circularize operations
- Using the split, merge, and disconnect operations
- Adding multipolygons
- Using street-level imagery
- Handling validation issues
- Mapping transit stops
- Mapping indoor features
- Mapping sidewalks and accessible crossings
- Mapping roads in desert imagery
- Mapping buildings in forest imagery

matkoniecz:
How triggering them would work? Main benefit of tutorial at this moment is that one can recommend to potential mappers to go through it once iD offers it.

It would be preferable that it would be also possible after splitting it into parts.

pnorman:
Has anyone run statistics on current tutorial usage?

quincylvania:
> How triggering them would work?

@matkoniecz Similar to the current walkthrough, you could specify a specific tutorial (or possibly list of tutorials) to run automatically on launch. You could also access them through the Help pane.

> Has anyone run statistics on current tutorial usage?

@pnorman I don't know, I'm not sure if there are statistics available except the changeset tag specifying if the user had started the walkthrough.

PR #12728, opened 2026-08-11
create a 'distribute' operation

### LINKED ISSUE #12669: New feature: "distribute nodes evenly along axis"
### Description

Add a method to "distribute nodes" along an axis.

Conceptually, this is equivalent to the "Distribute Rows/Columns" in Microsoft Word and "Distribute Rows/Columns Evenly" in LibreOffice Writer. This is analogous to "Straighten" for ways, which already exists in iD.

The use case is that, in some urban areas (e.g. Hong Kong), sometimes the shop nodes are evenly spaced between two points. Having this "distribute nodes evenly along axis" feature allows speedy alignment of these shop nodes.

### Screenshots

Consider this simple diagram:

Image

There are 6 same-size buildings along a street, with each having its own distinct housenumber. Each building (at their respective ground level) contains a shop. It is easy to see that the 6 shops (at housenumber 2, 4, ...) are evenly spaced between each other.

k-yle:
Closes #12669

we already have #9020 for evenly dividing **areas**. This PR is a similar idea, but for evenly distributing **nodes**:

20260811224736_rec_-convert

The keyboard shortcut is the same as the straighten operation, but with <kbd>⇧ Shift</kbd>. In english, that would be <kbd>⇧ Shift</kbd> <kbd>S</kbd>. Mostly because i see this as an extension of the straighten operation, since distributing involves first straightening the nodes. 

use cases:
 - indoor shops
 - any vertices like a loading dock, which you surveyed by counting, but can't see the exact location on aerial imagery
 
~~to do~~:
- [x] ~~unit tests~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :sparkles: Usability & Accessibility
 * Render diameter and radius tags when a node is selected ([#9732], thanks [@k-yle])
 #### :scissors: Operations
+* Create a 'distribute' operation ([#12728], thanks [@k-yle])
 #### :camera: Street-Level
 #### :white_check_mark: Validation
 #### :bug: Bugfixes
@@ -52,6 +53,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 
 [#9732]: https://github.com/openstreetmap/iD/pull/9732
 [#12679]: https://github.com/openstreetmap/iD/pull/12679
+[#12728]: https://github.com/openstreetmap/iD/pull/12728
 
 
 # 2.42.1
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -172,6 +172,16 @@ en:
       not_downloaded:
         single: This can't be made square because parts of it have not yet been downloaded.
         multiple: These can't be made square because parts of them have not yet been downloaded.
+    distribute:
+      title: Distribute Evenly
+      description: Spread out these points evenly along a straight line.
+      annotation: Distributed {n} points.
+      disabled:
+        too_few: Distributing requires at least 3 points.
+        already_distributed: These points can't be made distributed any more evenly than they already are.
+        too_large: These points can't be distributed because not enough of them are currently visible.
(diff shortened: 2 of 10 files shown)


matkoniecz:
quite hidden, but can be useful for power users and I have no good idea for making it more discoverable

maybe some expert tutorial? mimicking regular tutorial? but not sure is it worth it

matkoniecz:
it can break geometries of areas, but regular `s` could do it anyway

k-yle:
> maybe some expert tutorial? mimicking regular tutorial? but not sure is it worth it

in general it would be nice to start a series of tips/tutorials, maybe in the #iD on slack? or alternatively, something like #6079

> it can break geometries of areas, but regular <kbd>s</kbd> could do it anyway

yeah that should probably be tracked in a different issue, if we fix it for <kbd>s</kbd> then it fixes it for this PR too

tordans:
Can we add this to the context menu? Most of the time I remember the is a power user feature for something but don't know how to get to it, so I have to look for the shortcut menu, find the action in the right tab and then apply it. The k-bar would also help here, but a context menu would be easiest IMO.

matkoniecz:
I really would not complicate right-click menu more, I would prefer to not have this feature at all over putting more there.

It is already too complex.

tordans:
> I really would not complicate right-click menu more, …
>It is already too complex.

I disagree. Users who right click know to expect a list of things. Those are not newbies. But most importantly this is a context aware menu, so this option is only there when the conditions are met… in which case it is the right thing to inform users about it.

matkoniecz:
I added it to https://wiki.openstreetmap.org/wiki/ID/Community_Chats/2026#Community_Chat_September_9%2C_16%3A00_-_17%3A00_%28UTC%29

but maybe wait till next one when @k-yle can join?

for reference: `* (Mateusz) Context menu design - should we add there more items like https://github.com/openstreetmap/iD/pull/12728`

k-yle:
> but maybe wait till next one when @k-yle can join?

feel free to discuss it without me if anyone wants. but i won't be getting up at 4am sorry 😆

PR #12734, opened 2026-08-12
fix the crossing_ways validator not working

### LINKED ISSUE #12731: Validator not showing warnings when crossings cross a roadway
### URL

https://www.openstreetmap.org/edit#map=21/26.1335352/-81.7948973

1. Draw a way that connects to at least one other footway and crosses a roadway
2. Select a crossing preset (e.g. Unmarked Crossing, Marked Crossing, etc.)
3. No warning appears

Image

4. If you add a roadway spur that crosses this new crossing, the missing warning appears

Image

5. This restores the warning on the roadway AND crossing element
Image

_No response_

2.42.0

_No response_

k-yle:
Closes #12731

caused by #12425 😭 quite ironic because this is the exact type of bug that PR was supposed to prevent. i didn't notice it because both `osmWay` and `Segment`  have a property called `.nodes`.

this caused the validator to only use a single segment from the first to the last node, ignoring all other intermediate nodes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
+* fix the crossing_ways validator not working ([#12734], thanks [@k-yle])
 #### :bug: Bugfixes
 #### :earth_asia: Localization
 #### :hourglass: Performance
@@ -50,6 +51,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Drop code previously responsible for _maprules_ integration (which has been defunct for a while now) ([#12679])
 
 [#12679]: https://github.com/openstreetmap/iD/pull/12679
+[#12734]: https://github.com/openstreetmap/iD/pull/12734
 
 # 2.42.0
 ##### 2026-Aug-10
--- a/modules/osm/way.ts
+++ b/modules/osm/way.ts
@@ -18,7 +18,7 @@ export interface Segment {
     wayId: WayId;
     index: number;
     nodes: NodeId[];
-    extent(graph: coreGraph): geoExtent | undefined;
+    extent(this: Segment, graph: coreGraph): geoExtent | undefined;
 }
 
 export class osmWay extends OsmAbstractEntity {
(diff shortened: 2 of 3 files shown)


1ec5 (approved):
Thanks for the quick fix!

tyrasd (approved):

kbzimmer:
I'm still seeing this intermittently after this fix.  Screenshot from just now:
image

jjiglesiasg:
If version 2.42.0 is still active and the Problem Remains (Multiple Validation  Errors) how these issues #12731 and #12734 are closed?

I guess Issues should be closed ONCE the active solution or fix becomes available to the Mappers. 

In the days since this error was reported.  Hundreds if not thousands of map errors, could has been included in the map data base without acknowledge of the mappers...

tordans:
Github issues are closed the code was merged. The changes can then be tested on the preview release that is linked in the readme. 
The release to osm-org or any other places where id is used is independent of the merge. This aligns with how most projects work. 
A good way to help catch issues is to use the preview release regularly to catch issues before they are released.

k-yle:
yeah, as tobias mentioned I think our workflow of issues and PRs is fine. but I do want to apologise for this bug, and think about some concrete things that we can improve on:

- to prevent similar bugs in the future: continue work on [%47](https://github.com/openstreetmap/iD/milestone/47) 
- to reduce bugs in general: maybe we could announce when a release is impending? so that mappers can beta-test the preview version before it gets released. 
- to speed up the time between releases: lets discuss it in [#12757](https://github.com/openstreetmap/iD/issues/https://github.com/openstreetmap/iD/)

tyrasd:
In addition, I made a mistake yesterday while backporting the fixes for the v2.42.1 release for the osm website. The compiled build output in the `dist` directory was not included in the release, so while the sources had the fix, the deployed version on osm.org did not actually have the fix (and also still showed v2.42.0 after v2.42.1 was already been rolled out).

https://github.com/openstreetmap/openstreetmap-website/pull/7327 will fix this shortly.

tyrasd:
> announce when a release is impending? so that mappers can beta-test

Well, we have managed to have more or less monthly [releases](https://github.com/openstreetmap/iD/releases) (as mentioned on the [readme](https://github.com/openstreetmap/iD#participate)), so it shouldn't really be necessary to announce them every 4 weeks or so.

Ideally, experienced mappers would use the [preview](https://preview.ideditor.com) generally a bit more to catch such bugs early. Maybe we should just advertise it a bit more to raise awareness?

PR #12736, opened 2026-08-12
fix the outdated_tags validator raising erroneous tag upgrade suggestions

### LINKED ISSUE #12733: "Suggested update" with nothing to change
### URL

_No response_

When editing railway stations, the editor currently suggests an update of the tags, but it is not clear what it wants to change. Two tags are mentioned, but these are already present. Accepting the suggested update does not change anything.

1. open https://www.openstreetmap.org/edit#map=24/53.96095032/11.61496227
2. click on the railway station node _Hagebök_
3. the left sidebar shows the following, despite both tags already present:
Image

4. clicking on _Accept the suggested updates_ does not change the left sidebar
5. but the button in the top right now offers to undo the changes which did not change anything (note the greyed-out _Save_ button)
Image

_No response_

2.42.0 / 2.43.0-dev

Chrome

k-yle:
Closes #12733

caused by 88f106ebd92bd1596f27aa21dc119ecba0180355, now the diff includes unchanged rows which are identified by `type="~"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
+* fix the outdated_tags validator raising erroneous errors ([#12736], thanks [@k-yle])
 #### :bug: Bugfixes
 #### :earth_asia: Localization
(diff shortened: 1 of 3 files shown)


tyrasd (approved):
Thanks! Sorry for not thinking about this side effect when refactoring this code.

PR #12737, opened 2026-08-12
allow tags like `not:*:wikidata` to have multiple values

### LINKED ISSUE #12691: Fix V'Lille preset in the same way that IDEcycle and Métrovélo
V'Lille is described with this [wikidata](https://www.wikidata.org/wiki/Q3552843) and this [wikipedia](https://fr.wikipedia.org/wiki/V%27Lille)

The bug:
- follow that link https://www.openstreetmap.org/edit?node=11268810483#map=19/50.640105/3.075108
- in the left panel, iD shows a good Feature Type
- but iD shows 1 issue: it wants to replace many good values with bad ones
- if you select the choice "tags as not", the tag not:brand:wikidata=Q3146580 is added
- in the left panel, iD wants again to replace many good values with some others bad ones
- if you select the choice "tags as not", the tag not:brand:wikidata=Q3333896 replaces the previous value
- in the left panel, iD wants again to replace many good values with... the first suggestion!

And so on : impossible to stop iD suggestions, it always wants to replace the good values with bad ones.

### LINKED ISSUE #12715: impossible to mark NSI preset as invalid when multiple `not:brand:*` entries are needed
As reported in https://github.com/openstreetmap/iD/issues/12691 :

>V'Lille is described with this [wikidata](https://www.wikidata.org/wiki/Q3552843) and this [wikipedia](https://fr.wikipedia.org/wiki/V%27Lille)
>
>The bug:
>- follow that link https://www.openstreetmap.org/edit?node=11268810483#map=19/50.640105/3.075108
>- in the left panel, iD shows a good Feature Type
>- but iD shows 1 issue: it wants to replace many good values with bad ones
>- if you select the choice "tags as not", the tag not:brand:wikidata=Q3146580 is added
>- in the left panel, iD wants again to replace many good values with some others bad ones
>- if you select the choice "tags as not", the tag not:brand:wikidata=Q3333896 replaces the previous value
>- in the left panel, iD wants again to replace many good values with... the first suggestion!
>
>And so on : impossible to stop iD suggestions, it always wants to replace the good values with bad ones.

k-yle:
Closes #12691, Closes #12715

`not:brand:wikidata` can now be a semicolon-delimited list.

since https://github.com/osmlab/name-suggestion-index/issues/12496 and https://github.com/osmlab/name-suggestion-index/issues/12497 have already been fixed and released, this can't be tested with the original example.

but you can still test it by creating a node anywhere in the world tagged with:
```ini
amenity=fast_food
name=McDonald's
not:brand:wikidata=abc
```

### DIFF
--- a/modules/services/nsi.js
+++ b/modules/services/nsi.js
@@ -7,6 +7,7 @@ import { nsiCdnUrl } from '../../config/id.js';
 // Make very sure this resolves to iD's `package.json`
 // If you mess up the `../`s, the resolver may import another random package.json from somewhere else.
 import packageJSON from '../../package.json';
+import { utilSplitAtSemicolon } from '../util/util';
 
 
 // This service contains all the code related to the **name-suggestion-index** (aka NSI)
@@ -516,10 +517,10 @@ function _upgradeTags(tags, loc) {
       if (!item) continue;
       const mainTag = item.mainTag;               // e.g. `brand:wikidata`
       const itemQID = item.tags[mainTag];         // e.g. `brand:wikidata` qid
-      const notQID = newTags[`not:${mainTag}`];   // e.g. `not:brand:wikidata` qid
+      const notQIDs = utilSplitAtSemicolon(newTags[`not:${mainTag}`]); // e.g. `not:brand:wikidata` qid
 
       if (                                        // Exceptions, skip this hit
-        (!itemQID || itemQID === notQID) ||       // No `*:wikidata` or matched a `not:*:wikidata`
+        (!itemQID || notQIDs?.includes(itemQID)) ||// No `*:wikidata` or matched a `not:*:wikidata`
         (newTags.office && !item.tags.office)     // feature may be a corporate office for a brand? - #6416
       ) {
         item = null;
--- a/modules/validations/outdated_tags.js
+++ b/modules/validations/outdated_tags.js
@@ -8,7 +8,8 @@ import { actionUpgradeTags } from '../actions/upgrade_tags';
 import { fileFetcher } from '../core';
 import { presetManager } from '../presets';
 import { services } from '../services';
-import {  utilHashcode, utilTagDiff } from '../util';
+import { utilArrayUniq, utilHashcode, utilTagDiff } from '../util';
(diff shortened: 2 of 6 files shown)


modules/services/nsi.js
```diff
@@ -516,10 +516,10 @@ function _upgradeTags(tags, loc) {
       if (!item) continue;
       const mainTag = item.mainTag;               // e.g. `brand:wikidata`
       const itemQID = item.tags[mainTag];         // e.g. `brand:wikidata` qid
-      const notQID = newTags[`not:${mainTag}`];   // e.g. `not:brand:wikidata` qid
+      const notQIDs = newTags[`not:${mainTag}`]?.split(';'); // e.g. `not:brand:wikidata` qid
```
1ec5: Do we need to strip whitespace around the semicolon just in case? I don’t expect iD to generate a space here, but we strip it for most keys.

1ec5 (approved):

modules/services/nsi.js
k-yle: sounds good, i extracted that code into a helper function and used it here

1ec5 (approved):

modules/validations/outdated_tags.js
```diff
@@ -214,7 +215,14 @@ export function validationOutdatedTags() {
       const wd = item.mainTag;     // e.g. `brand:wikidata`
       const notwd = `not:${wd}`;   // e.g. `not:brand:wikidata`
       const qid = item.tags[wd];
-      newTags[notwd] = qid;
+      if (newTags[notwd]) {
+        newTags[notwd] = utilArrayUniq([
+            ...utilSplitAtSemicolon(newTags[notwd]),
+            qid,
```
matkoniecz, line 221: what if new tag would exceed in length maximum allowed tag length?

k-yle: then it'll get silently truncated before being uploaded. this edge case affects everything in iD so i don't think it needs to be explicitly considered here

matkoniecz (approved):

tyrasd (approved):

PR #12755, opened 2026-08-18
Prevent area merges that introduce repeated interior nodes

### LINKED ISSUE #9328: Can create invalid geometries with node merge.

seanwessmith:
**AI disclosure:** This PR contains AI-generated code and tests. This review update was prepared with OpenAI Codex: it read #9328 and the review comments, inspected the merge implementation, clarified the helper name and comments, added regression tests, ran the checks below, and exercised the local editor to capture the demonstration. These checks were performed by the agent; they are not a claim of independent human review.

## What changed

Prevent area-node merges that introduce or increase a repeated interior vertex. The normal repetition of the first node at the end of a closed way is explicitly excluded from the check.

- Preserve valid merges of adjacent or consecutive area nodes, including across the ring closure.
- Keep repeated interior nodes supported for linear ways.
- Allow merges that do not worsen an already-invalid area's repeated nodes.

The connect action simulates node replacements in affected parent areas and disables the merge with the existing `paths_intersect` reason when it would introduce or increase an interior repetition.

Closes #9328.

## Demonstration

Recorded from the local iD build at `5246475`, using synthetic geometry matching #9328. The extra “Load example area” button only supplies the fixture; selection and merging use the normal editor UI. OSM loading and saving were disabled using the walkthrough controls.

1. Select the nonadjacent top-right and bottom-right nodes and press **C**: the intersection warning appears and the area stays unchanged.
2. Reload the example, select the adjacent top-left and bottom-left nodes (across the ring closure), and press **C**: the merge succeeds.

Recorded demonstration of the blocked invalid merge and allowed adjacent merge

[MP4 recording](https://github.com/seanwessmith/iD/raw/57fe7478d3e007ff84519efba71fb9c9d7331ce2/area-node-merges.mp4). Pauses between steps were shortened.

## Validation

- Focused connect/merge-node tests: 41 passed.
- Full unit suite: 2,268 passed, 9 skipped, 6 todo.
- Build, distribution build, and TypeScript checks passed.
- Changed-file ESLint passed without warnings. Full lint completed with one existing `no-console` warning in `test/spec/core/difference.js:371`.
- `git diff --check` passed.

### DIFF
--- a/modules/actions/connect.ts
+++ b/modules/actions/connect.ts
@@ -6,6 +6,37 @@ import type { EntityId, osmNode, osmRelation, osmWay, NodeId } from '../osm';
 import type { RelationMember } from '../osm/relation';
 
 
+function wouldIntroduceRepeatedAreaNode(way: osmWay, nodeIDs: NodeId[], survivorID: NodeId) {
+    if (!way.isArea()) return false;
+
+    var result = way;
+    for (const nodeID of nodeIDs) {
+        if (nodeID !== survivorID) {
+            result = result.replaceNode(nodeID, survivorID);
+        }
+    }
+
+    var originalNodes = way.isClosed() ? way.nodes.slice(0, -1) : way.nodes;
+    var originalCounts = new Map<NodeId, number>();
+    for (const nodeID of originalNodes) {
+        originalCounts.set(nodeID, (originalCounts.get(nodeID) ?? 0) + 1);
+    }
+
+    var resultNodes = result.isClosed() ? result.nodes.slice(0, -1) : result.nodes;
+    var resultCounts = new Map<NodeId, number>();
+    for (const nodeID of resultNodes) {
+        resultCounts.set(nodeID, (resultCounts.get(nodeID) ?? 0) + 1);
+    }
+
+    for (const [nodeID, count] of resultCounts) {
+        if (count > 1 && count > (originalCounts.get(nodeID) ?? 0)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+
 // Connect the ways at the given nodes.
 //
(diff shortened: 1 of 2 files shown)


modules/actions/connect.ts
```diff
@@ -6,6 +6,37 @@ import type { EntityId, osmNode, osmRelation, osmWay, NodeId } from '../osm';
 import type { RelationMember } from '../osm/relation';
 
 
+function wouldIntroduceRepeatedAreaNode(way: osmWay, nodeIDs: NodeId[], survivorID: NodeId) {
```
matkoniecz, line 9: title is confusing given first/last node is always the same for closed ways (and therefore repeating)

matkoniecz:
This is LLM generated PR, right? In such case it should be disclosed at beginning of description, with explanation how it was generated.

matkoniecz:
> This pull request is still a work in progress

please, mark it as ready for review once it is ready for review

also, can you attach recording of this change in action and attach it to original post or in a new comment?

modules/actions/connect.ts
seanwessmith: Clarified in 5246475: the helper is now `wouldIntroduceRepeatedInteriorAreaNode`, with an explicit comment that the normal first/last node repetition is excluded. I also updated the PR title and description, and added regression cases showing that a merge across the closing node is allowed while a new interior repetition is blocked. The description now starts with the AI disclosure and includes a recording of both behaviors. All 2,268 unit tests pass.

PR #12758, opened 2026-08-19
NSI: only remove existing wikidata/wikipedia tags if they match the NSI suggestion

### LINKED ISSUE #12756: Outdated tags validator removes wiki tags from protected area boundary
The protected area boundary relation for [Mendocino National Forest](https://www.openstreetmap.org/relation/2067891) should match the United States Forest Service preset from name-suggestion-index. However, iD warns that the `wikidata=Q6816692` and `wikipedia=en:Mendocino National Forest` tags are nonstandard and offers to remove them. [Changeset 186,096,983](https://www.openstreetmap.org/changeset/186096983) apparently accepted this suggestion. I can see no reason why these tags would keep an element from matching the preset.

Protected Area Mendocino National Forest looks like a common feature with nonstandard tags

k-yle:
Closes #12756

PR is draft because this should probably be discussed with the NSI maintainers, and we should start writing unit tests for NSI.

---

The bug is caused because:
1. "Mendocino National Forest" [is an _operator_ in NSI](https://nsi.guide/index.html?t=*&k=amenity&v=toilets&tt=mendocino%20national%20forest), but it's also a physical place. so `wikidata=Q6816692` and `operator:wikidata=Q6816692` are both valid.
2. also, [`r2067891`](https://osm.org/relation/2067891) matches a _different_ NSI preset. 

I can't think of any examples from my local area where this is valid, so maybe it's an NSI data issue? are the toilets really operated by the forest? the toilets are certainly not cleaned by the forest, i assume it's staff from the  _United States Forest Service_ who clean the toilets. Therefore, shouldn't the toilets use `operator=United States Forest Service` instead of `operator=Mendocino National Forest`?

--- 

this PR does not fix the even weider case of
```ini
building=office
wikidata=Q7660181
name=Sydney Trains HQ
```

which causes this suggestion:

image

because I'd like to get some unit tests written before we start fixing more of these issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
+* Only suggest deleting `wikidata`/`wikipedia` tags if they have the same value as the suggested `*:wikidata` tags from NSI ([#12758], thanks [@k-yle])
 #### :bug: Bugfixes
 #### :earth_asia: Localization
 #### :hourglass: Performance
@@ -52,6 +53,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 
 [#9732]: https://github.com/openstreetmap/iD/pull/9732
 [#12679]: https://github.com/openstreetmap/iD/pull/12679
+[#12758]: https://github.com/openstreetmap/iD/pull/12758
 
 
 # 2.42.1
--- a/modules/services/nsi.js
+++ b/modules/services/nsi.js
@@ -566,7 +566,7 @@ function _upgradeTags(tags, loc) {
     });
 
     // Replace mistagged `wikidata`/`wikipedia` with e.g. `brand:wikidata`/`brand:wikipedia`
-    if (foundQID) {
+    if (foundQID && foundQID === item.tags[item.mainTag]) {
       delete newTags.wikipedia;
       delete newTags.wikidata;
     }


matkoniecz:
> are the toilets really operated by the forest? the toilets are certainly not cleaned by the forest, i assume it's staff from the United States Forest Service who clean the toilets. Therefore, shouldn't the toilets use operator=United States Forest Service instead of operator=Mendocino National Forest?

I guess it depends on whether it is operated by United States Forest Service or specifically employees of 
Mendocino National Forest legal entity?

And I cannot really answer this one, maybe @1ec5 knows?

https://www.wikidata.org/wiki/Q6816692 seems to be about legal entity rather than physical object, similarly to https://www.wikidata.org/wiki/Q606000 (and Babiogórski Park Narodowy definitely can be operator of toilets etc)

though https://www.wikidata.org/wiki/Q6816692 is not entirely clear, and Wikidata very often mixes both in a single object

matkoniecz:
> and we should start writing unit tests for NSI.

for NSI processing in iD or external unit tests for NSI data?

matkoniecz:
> fix the outdated_tags validator trying to delete wikidata/wikipedia tags - #12758

this description is a bit confusing as it stops specifically some specific case (while allowing it to delete wikidata/wikipedia tags where it is valid)

> and we should start writing unit tests for NSI.

maybe add one (first) here? Do we have one that would confirm that obvious cases of deleting them still works? (I can and will test it manually, but guarding against regression would be great)

modules/services/nsi.js
```diff
@@ -566,7 +566,7 @@ function _upgradeTags(tags, loc) {
     });
 
     // Replace mistagged `wikidata`/`wikipedia` with e.g. `brand:wikidata`/`brand:wikipedia`
-    if (foundQID) {
+    if (foundQID && foundQID === item.tags[item.mainTag]) {
```
matkoniecz, line 569: I tried to read code to understand how this change works but I admit I failed

Is `outdated_tags` mentioned in commit referring to some NSI property or to iD function? Or to something else?

1ec5:
> I can't think of any examples from my local area where this is valid, so maybe it's an NSI data issue? are the toilets really operated by the forest? the toilets are certainly not cleaned by the forest, i assume it's staff from the United States Forest Service who clean the toilets. Therefore, shouldn't the toilets use `operator=United States Forest Service` instead of `operator=Mendocino National Forest`?

Oh, huh, I didn’t think to look for an NSI entry for the national forest. I assume the entry came in through nsi-collector from someone tagging the operator on a bunch of restrooms in OSM.

In general, a National Forest is both an legal jurisdiction and a unit within the Forest Service organization. (Some national forests are only “proclaimed forests” without an administration, being administered as part of a consolidated “administrative forest”.) I suppose it could be valid to tag the facilities as being operated by the national forest, or even more specifically, the semiautonomous ranger districts within them. I came across this example because of a plan to [clean up tagging of National Forest routes](https://community.openstreetmap.org/t/network-tagging-proposal-for-usfs-roads/97057). In principle, the constituent ways could have `operator=Mendocino National Forest`. At some point, though, it starts to feel pedantic.

> I guess it depends on whether it is operated by United States Forest Service or specifically employees of 
Mendocino National Forest legal entity?
> 
> And I cannot really answer this one, maybe @1ec5 knows?
> 
> https://www.wikidata.org/wiki/Q6816692 seems to be about legal entity rather than physical object, similarly to https://www.wikidata.org/wiki/Q606000 (and Babiogórski Park Narodowy definitely can be operator of toilets etc)

In some instances, Wikidata is distinguishing between [the national forest](https://www.wikidata.org/wiki/Q1427349) and [the agency](https://www.wikidata.org/wiki/Q108566721) that runs it, and even [ranger districts](https://www.wikidata.org/wiki/Q49332203). It might be feasible to split out more agency items, but I’m already catching flak from some Wikidata editors for doing something similar to counties, so I probably won’t be able to prioritize that work anytime soon. In any case, a National Forest isn’t a physical `landuse=forest` in the traditional OSM sense.

modules/services/nsi.js
tyrasd: @matkoniecz: 

`foundQID` is basically the OSM feature's `wikidata` tag (if NSI knows about it), here it is `Q6816692` for the national forest.
`item` is the NSI entry that was found to be matching the given OSM feature (in this case the one for the operator _United States Forest Service_ with `operator:wikidata=Q1891156`).

Going by the comment above, we would assume that the `wikidata` tag is actually mistagged when the same value is both present on the feature's `wikidata` and the matching NSI entry's tag (e.g. `operator:wikidata`).

<hr>

`outdated_tags` is the name of the validator in iD which handles both NSI tag suggestions as well as the `deprecated.json` list from id-tagging-schema.

tyrasd (approved):
If I understand it correctly, the removal of `wikidata=*` tags when the value is a valid operator was implemented as a feature, in order to fix cases where a previous mapper had accidentally used the wrong tag for the operator's wikidata value. Limiting this functionality to only kick in when the `wikidata` tag value in question is the same as the matched operator seems reasonable to me.

But let's wait for what the maintainers of NSI think about it. //cc @arch0345  @bhousel @Cj-Malone @kjonosm

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ _Breaking developer changes, which may affect downstream projects or sites that
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
+* Only suggest to remove existing `wikidata`/`wikipedia` tags if they match the matching suggestion from NSI ([#12758], thanks [@k-yle])
```
matkoniecz:
```suggestion
* Only suggest to remove existing `wikidata`/`wikipedia` tags if they are the same as  matching suggestion from NSI ([#12758], thanks [@k-yle])
```

matkoniecz: avoids confusing repetition of verb match

k-yle: or 
```suggestion
* Only suggest deleting `wikidata`/`wikipedia` tags if they have the same value as the suggested `*:wikidata` tags from NSI ([#12758], thanks [@k-yle])
```
?

matkoniecz: also fine

matkoniecz:
sad consequence is that some currently invalid will not be removed, but I have no idea how to magically avoid both problems and false negatives are better than false positives here

k-yle:
which "currently invalid" ones will no longer be removed?

matkoniecz:
sorry, after testing my claim appears to be a nonsense, I have hidden it (feel free to hide also this one)

PR #12768, opened 2026-08-22
use translated labels for relation roles

k-yle:
this is not ready for a code review, it's blocked by several other PRs. 

i'm just creating this PR to make it easier to review https://github.com/openstreetmap/id-tagging-schema/pull/1538

---

relation roles are now translated if the geometry and tags match:

image

in this case, `stop` is not translated, since it's only valid for `vertex`, not valid for `point`:

image

### DIFF
--- a/config/id.js
+++ b/config/id.js
@@ -2,7 +2,7 @@
 
 // cdns for external data packages
 const presetsCdnUrl = ENV__ID_PRESETS_CDN_URL
-  || 'https://cdn.jsdelivr.net/npm/@openstreetmap/id-tagging-schema@{presets_version}/';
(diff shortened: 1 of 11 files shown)


matkoniecz:
> it's blocked by several other PRs.

for example? (I would make reviewing them higher priority)

k-yle:
> for example? 

mostly the typescript ones from a few weeks ago: https://github.com/openstreetmap/iD/milestone/47 and https://github.com/openstreetmap/iD/pull/12727

so for now, maybe we could focus on the id-tagging-schema side

matkoniecz:
> so for now, maybe we could focus on the id-tagging-schema side

in such case I will be more useful on id tagging side

PR #12773, opened 2026-08-24
convert nsi to typescript

k-yle:
another straightforward one. 1 comment inline:


modules/services/nsi.ts
```diff
@@ -461,15 +497,6 @@ function _upgradeTags(tags, loc) {
           delete newTags[osmkey];
         }
       }
-      if (replace && replace.wikipedia !== undefined) {  // replace or delete `*:wikipedia` tag
```
k-yle, line 464: this block was deleted because it's dead code, NSI no longer conflates `*:wikipedia` tags, the typedefs don't even include `.wikipedia` anymore

tyrasd (approved):

PR #12777, opened 2026-08-24
convert taginfo to typescript

k-yle:
this one has a big diff, but the changes are all trivial

### DIFF
--- a/modules/services/taginfo.ts
+++ b/modules/services/taginfo.ts
@@ -9,15 +9,67 @@ import { allowUpperCaseTagValues } from '../osm/tags';
 import { taginfoApiUrl } from '../../config/id.js';
 
 var apibase = taginfoApiUrl;
-var _inflight = {};
-var _popularKeys = {};
+let _inflight: { [url: string]: AbortController } = {};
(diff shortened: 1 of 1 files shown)


modules/services/taginfo.ts
```diff
@@ -159,17 +213,16 @@ function request(url, params, exactMatch, callback, loaded) {
...
         .then(function(result) {
             delete _inflight[url];
-            if (loaded) loaded(null, result);
+            if (loaded) loaded(null, result as T);
         })
         .catch(function(err) {
             delete _inflight[url];
             if (err.name === 'AbortError') return;
-            if (loaded) loaded(err.message);
+            if (loaded) loaded(err);
```
k-yle, line 221: this was missed in #12684

modules/services/taginfo.ts
```diff
@@ -357,21 +410,13 @@ export default {
...
-                _taginfoCache[url] = d.data;
-                callback(null, d.data);
+                _taginfoCache[url] = d!.data;
+                callback(null, d!.data);
             }
         });
     },
-
-
-    apibase: function(_) {
```
k-yle, line 371: this is not used

PR #12778, opened 2026-08-24
migrate bing streetside to typescript

k-yle:
there's a few runtime changes in this PR, but they're all self explanatory

### DIFF
--- a/modules/services/streetside.ts
+++ b/modules/services/streetside.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-this-alias */
 import { dispatch as d3_dispatch } from 'd3-dispatch';
 import { timer as d3_timer } from 'd3-timer';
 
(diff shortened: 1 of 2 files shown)


modules/services/streetside.ts
```diff
@@ -107,18 +170,18 @@ function loadNextTilePage(which, url, tile) {
...
-      const d = {
+      const d: Bubble = {
         service: 'photo',
         loc: loc,
         key: bubbleId,
         imageUrl: bubble.imageUrl
           .replace('{subdomain}', bubble.imageUrlSubdomains[0]),
         ca: bubble.he || bubble.heading,
-        captured_at: bubble.vintageEnd,
+        captured_at: bubble.vintageEnd ? new Date(bubble.vintageEnd) : undefined,
```
k-yle, line 184: this is technically a bug fix, but it the bug isn't noticable, because both `String` and `Date` have a [`.toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) method

PR #12784, opened 2026-08-25
remove redundant CLDR overrides

k-yle:
these overrides are no longer required, since CLDR contains this information. 

to test: run `npm run build:data` and note how there is no change to `data/*`.


matkoniecz (approved):
running `npm run build:data` results in no changes, removing `nys: { nativeName: 'Nyungar' },` results in  a change

tyrasd (approved):

PR #12785, opened 2026-08-25
remove name:wlp (typo)

k-yle:
[`wlp`](https://r12a.github.io/app-subtags/index?lookup=wlp) is a typo, it should have been [`wbp`](https://r12a.github.io/app-subtags/index?lookup=wbp) which is already in CLDR (see line above).

neither of them are currently used in osm.


matkoniecz (approved):
checked code on https://en.wikipedia.org/wiki/Warlpiri_language

also checked https://taginfo.openstreetmap.org/search?q=name%3Awlp

PR #12787, opened 2026-08-26
Clean up some images

hlfan:
I'd like to transition the loader GIFs away from the GIF format. I'm open to whether SVG is the better choice.

Here you can see the old and new loading backgrounds next to each other:

This is a first stab; for the non-background loaders, some JS needs to be involved too to load the backgrounds in a way to make use of the CSS colour variables.


matkoniecz:
what is wrong with gifs? Why using JS-animations will be better?

hlfan:
I wanna reuse the CSS colour definitions, since iD has much infrastructure built around SVGs, I'd hook into that too.

Just to use that, the SVGs have to be inserted directly into the DOM and not via an image tag, nor just as a CSS background.

dist/img/arrow-icon.png
tyrasd, line 1: this file was just unused, correct?

tyrasd (commented):
Thanks, this looks good to me.

Just one question: the `background-image: url(img/mini-loader.gif);` [gif](https://github.com/openstreetmap/iD/blob/develop/dist/img/mini-loader.gif) could also be converted to an animated svg, while we're at it?

dist/img/arrow-icon.png
hlfan: afaict yes.

hlfan:
<del>Yeah the `mini-loader.gif` is also on my radar, but I haven't yet figured out which way to model it.</del> <ins>Done.</ins>

I think converting the turn restriction gifs into APNGs would fit here as well (I think animated SVGs for that are overkill).

dist/img/arrow-icon.png
hlfan: Therefore also the commit comment of 4d1f61f6246bf1ed2b07435874caef2f724d8585.

PR #12794, opened 2026-08-27
Detect goo.gl shortlinks as an incompatible source

### LINKED ISSUE #12740: Regexes for detecting Google sources and imagery should also detect equivalent goo.gl shortlinks
### Description

Currently, iD offers a helpful warning when a changeset description or the changeset or object-level `source=*` tag contains a reference to Google. However, it only detects the URLs that contain the full google.com domain, rather than the shortened `goo.gl` equivalent, e.g. `https://maps.app.goo.gl/UX2Kv2GVQR9bkWwv5`. In [a changeset yesterday that I commented on](https://osm.org/changeset/187355985), a new user expressed surprise and frustration that iD didn't warn them about submitting data that was potentially from a clearly proprietary source before uploading their changeset, as they would have then known to make sure to avoid doing so before the changeset actually touched the database. More broadly, I regularly patrol for new changesets that have Google or goo.gl in the description or source, and there are a [fair few](https://osmcha.org/?filters=%7B%22date__gte%22%3A%5B%7B%22label%22%3A%222026-08-04T05%3A00%3A00.000Z%22%2C%22value%22%3A%222026-08-04T05%3A00%3A00.000Z%22%7D%5D%2C%22source%22%3A%5B%7B%22label%22%3A%22goo.gl%22%2C%22value%22%3A%22goo.gl%22%7D%5D%2C%22hide_whitelist%22%3A%5B%7B%22label%22%3A%22Yes%22%2C%22value%22%3A%22True%22%7D%5D%7D) that use the shortened version.

I can understand not wanting to go down the slippery slope of adding checks for any incompatible source, but given all the UI and infrastructure already exists to do this for non-shortened links (and mentions) of Google, it would just be a quick regex tweak to include shortlinks as well (rather than a whole new rule). Specifically, you'd just need to modify the regex on [line 18](https://github.com/openstreetmap/iD/blob/9b18ef296a1ae04f2072d02f0be5a38e0af74455/modules/validations/incompatible_source.ts#L18) of [validations/incompatible_source.ts](https://github.com/openstreetmap/iD/blob/develop/modules/validations/incompatible_source.ts#L18) from `/(google)/i` to `/(google|goo\.gl)/i`, and probably the exception regex on [line 19](https://github.com/openstreetmap/iD/blob/9b18ef296a1ae04f2072d02f0be5a38e0af74455/modules/validations/incompatible_source.ts#L19) from

```regex
/((books|drive)\.google|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
```

to

```regex
/((books|drive)\.(google|goo\.gl)|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
```

Similarly, the [regex used by the imagery blocklist](https://github.com/openstreetmap/iD/blob/develop/modules/services/osm.js#L36) does not detect `goo.gl` links, which resulted in other changesets by other users like [this one yesterday](https://www.osm.org/changeset/187323832) directly using prohibited Google imagery to map inside iD. That could probably be fixed in [`services/osm.js`](https://github.com/openstreetmap/iD/blob/develop/modules/services/osm.js#L36) by changing the regex on [line 36](https://github.com/openstreetmap/iD/blob/9b18ef296a1ae04f2072d02f0be5a38e0af74455/modules/services/osm.js#L36) from

```regex
/.*\.google(apis)?\..*\/(vt|kh)[\?\/].*([xyz]=.*){3}.*/
```
to something like

```regex
/.*\.(google(apis)?\..*\/(vt|kh)[\?\/].*([xyz]=.*){3}.*|goo\.gl\/.+)/
```

I note this regex is duplicated on [line 323](https://github.com/openstreetmap/iD/blob/9b18ef296a1ae04f2072d02f0be5a38e0af74455/modules/renderer/background.js#L323) of [`renderer/background.js`](https://github.com/openstreetmap/iD/blob/developmodules/renderer/background.js#L323), although given there was a reason it was duplicated rather than DRYly referenced I'm not sure if that should be updated too. It may also be prudent to create a regression test case for this in [the appropriate file](https://github.com/openstreetmap/iD/blob/develop/test/spec/validations/incompatible_source.js#L51).

Thanks!

### Screenshots

Expected:

Image

Actual:

Image

kunalKumar-13:
Fixes #12740

The Google rule in the incompatible-source validator matched only the literal
string `google`, so a shortened link like
`https://maps.app.goo.gl/UX2Kv2GVQR9bkWwv5` produced no warning even though it
points straight at Google Maps.

### Scope

@1ec5 settled the open question in the issue — since Google
[discontinued](https://developers.googleblog.com/en/google-url-shortener-links-will-no-longer-be-available/)
third-party use of the shortener, surviving links are effectively Google
properties only. So this matches the whole `goo.gl` domain rather than just
`maps.app.goo.gl`.

Deliberately left out, to keep this reviewable:

- **UTM tracking parameters** — @1ec5 asked for those to be tracked in a
  separate issue.
- **Adding `sites` to the exception regex** — suggested later in the thread but
  not confirmed by a maintainer, and it is a different question (which Google
  subdomains are *acceptable* sources) from this one.

### The change

```js
regex: /(google|\bgoo\.gl\b)/i,
```

Word boundaries matter here: without them `kangoo.gl` and `goo.glass` would both
be flagged. With them, `goo.gl`, `goo.gl/abc123` and `maps.app.goo.gl/…` match
while those two do not.

All three surfaces that warn about proprietary sources — the `source` tag
validator, the changeset comment check (`ui/commit.js`) and the changeset source
check (`ui/changeset_editor.js`) — go through `getIncompatibleSources()`, so one
rule change covers all of them.

### Tests

Four cases added to `test/spec/validations/incompatible_source.js`:

| Case | Expected |
|---|---|
| `https://maps.app.goo.gl/UX2Kv2GVQR9bkWwv5` | flagged |
| `goo.gl/abc123` | flagged |
| `kangoo.gl` | not flagged |
| `goo.glass` | not flagged |

The two positive cases fail on `develop` and pass here. The two negative cases
pass either way and are there to guard the word boundaries.

Full suite: **2265 passed**, 0 failures. `lint` and `tsc` clean.

### DIFF
--- a/modules/validations/incompatible_source.ts
+++ b/modules/validations/incompatible_source.ts
@@ -15,8 +15,9 @@ const incompatibleRules = [
   },
   {
     id: 'google',
-    // `goo.gl` is Google's link shortener. Third-party use was discontinued in
-    // 2025, so surviving links effectively point only at Google properties.
+    // `goo.gl` is Google's own link shortener. Third parties have not been able
+    // to create links on it since 2019, while Google products still generate
+    // them (`maps.app.goo.gl`), so a shortlink in a source is a Google source.
     regex: /(google|\bgoo\.gl\b)/i,
     exceptRegex: /((books|drive)\.google|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
   }
--- a/modules/validations/incompatible_source.ts
+++ b/modules/validations/incompatible_source.ts
@@ -19,7 +19,7 @@ const incompatibleRules = [
     // to create links on it since 2019, while Google products still generate
     // them (`maps.app.goo.gl`), so a shortlink in a source is a Google source.
     regex: /(google|\bgoo\.gl\b)/i,
-    exceptRegex: /((books|drive)\.google|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
+    exceptRegex: /((books|drive|sites)\.(google|goo\.gl)|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
   }
 ];
 
--- a/modules/validations/incompatible_source.ts
+++ b/modules/validations/incompatible_source.ts
@@ -15,7 +15,9 @@ const incompatibleRules = [
   },
   {
     id: 'google',
-    regex: /(google)/i,
+    // `goo.gl` is Google's link shortener. Third-party use was discontinued in
+    // 2025, so surviving links effectively point only at Google properties.
+    regex: /(google|\bgoo\.gl\b)/i,
     exceptRegex: /((books|drive)\.google|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
   }
 ];
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 #### :scissors: Operations
 #### :camera: Street-Level
 #### :white_check_mark: Validation
+* Warn about `goo.gl` shortlinks as sources, alongside full Google URLs ([#12740])
 #### :bug: Bugfixes
 #### :earth_asia: Localization
 #### :hourglass: Performance
@@ -51,6 +52,7 @@ _Breaking developer changes, which may affect downstream projects or sites that 
 * Drop code previously responsible for _maprules_ integration (which has been defunct for a while now) ([#12679])
 
 [#9732]: https://github.com/openstreetmap/iD/pull/9732
+[#12740]: https://github.com/openstreetmap/iD/issues/12740
 [#12679]: https://github.com/openstreetmap/iD/pull/12679
 
 
--- a/test/spec/validations/incompatible_source.js
+++ b/test/spec/validations/incompatible_source.js
@@ -102,4 +102,12 @@ describe('iD.validations.incompatible_source', function () {
         var issues = validate();
         expect(issues).toHaveLength(0);
     });
+
+    it('does not flag a Google Sites page', function() {
+        // Google Sites pages are user-generated; Google holds no copyright
+        // interest in the content, so they are a legitimate source.
+        createWay({ amenity: 'cafe', building: 'yes', name: 'Key Largo Café', source: 'https://sites.google.com/view/key-largo-cafe'});
+        var issues = validate();
+        expect(issues).toHaveLength(0);
+    });
 });
--- a/test/spec/validations/incompatible_source.js
+++ b/test/spec/validations/incompatible_source.js
@@ -74,4 +74,32 @@ describe('iD.validations.incompatible_source', function () {
         var issues = validate();
         expect(issues).toHaveLength(0);
     });
+
+    it('flags way with a goo.gl shortlink as source', function() {
+        createWay({ amenity: 'cafe', building: 'yes', name: 'Key Largo Café', source: 'https://maps.app.goo.gl/UX2Kv2GVQR9bkWwv5'});
+        var issues = validate();
+        expect(issues).toHaveLength(1);
+        var issue = issues[0];
+        expect(issue.type).toEqual('incompatible_source');
+        expect(issue.entityIds).toHaveLength(1);
+        expect(issue.entityIds[0]).toEqual('w-1');
+    });
+
+    it('flags way with a bare goo.gl source', function() {
+        createWay({ amenity: 'cafe', building: 'yes', name: 'Key Largo Café', source: 'goo.gl/abc123'});
+        var issues = validate();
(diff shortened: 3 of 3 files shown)


kunalKumar-13:
Good catch on the date — corrected in 2be1fc8. 2025 was the deactivation date for unused links, not when third-party use ended; new third-party links stopped in 2019. The comment now states the fact the rule actually rests on: third parties can't create `goo.gl` links, while Google products still generate them (`maps.app.goo.gl`).

kunalKumar-13:
Disclosure I should have put at the top when opening this: the change and tests
here were written with AI assistance. The full suite passes locally (2265 tests,
0 failures) and the two positive tests fail on `develop`, so the regex change is
doing the work the tests claim.

Apologies for not flagging it up front — I've since seen your note on #12630 that
disclosure belongs at the start rather than the end. Happy to answer questions on
any part of it, or to close this if the project would rather not take
AI-assisted changes.

modules/validations/incompatible_source.ts
```diff
@@ -15,7 +15,10 @@ const incompatibleRules = [
   },
   {
     id: 'google',
-    regex: /(google)/i,
+    // `goo.gl` is Google's own link shortener. Third parties have not been able
+    // to create links on it since 2019, while Google products still generate
+    // them (`maps.app.goo.gl`), so a shortlink in a source is a Google source.
+    regex: /(google|\bgoo\.gl\b)/i,
```
k-yle, line 21: please take a look at https://github.com/openstreetmap/iD/issues/12740#issuecomment-5320947194 - no one has replied, but I agree that it's reasonable to allow Google Sites, since some businesses use the default domains for their website. @CAM-Gerlach looks over a lot changesets, so if it's been seen in the wild, then it should probably be allowed

k-yle (changes requested):

kunalKumar-13:
Thanks — done in 9489042.

I've added `sites` to the exception and, following the regex in [#12740](https://github.com/openstreetmap/iD/issues/12740#issuecomment-5320947194), widened the exception's domain alternation to `(google|goo\.gl)` so the shortener's subdomains are covered the same way:

```
exceptRegex: /((books|drive|sites)\.(google|goo\.gl)|google\s?(books|drive|plus))|(esri\/Google_(Africa|Open)_Buildings)/i
```

Checked the resulting behaviour:

```
allow   https://sites.google.com/view/some-business
allow   https://books.google.com/...
allow   https://drive.google.com/...
FLAG    https://goo.gl/abc
FLAG    https://maps.app.goo.gl/xyz
FLAG    https://www.google.com/maps
```

So Google Sites is allowed while the shortlink detection this PR is about still works. 12 tests pass, and reverting just the regex fails the one added test.

On the Copilot note about the date: that was corrected earlier in 2be1fc8 — the comment now says third parties have not been able to create `goo.gl` links since 2019, with 2025 being the deactivation date for existing ones.

kunalKumar-13:
@k-yle Google Sites is allowed now — `sites` is in the `exceptRegex` alongside `books` and `drive`, pushed in 9489042.

Agreed with the reasoning on [#12740](https://github.com/openstreetmap/iD/issues/12740#issuecomment-5320947194): businesses do use the default domains for their site, so a Google Sites URL is a real website rather than an incompatible source.

Ready for another look when you have time.

CAM-Gerlach:
Just a quick heads up @k-yle -- you probably noticed already and not sure whether it bothers you, but to be clear you're talking to an LLM rather than a human contributor (even before I saw the em-dash-palooza, the tone and structure made it quite obvious).

k-yle:
yeah, it's quite a weird vibe... but that's the consequence of the magic label https://github.com/openstreetmap/iD/labels/new%20contributor%20opportunity , which is like a magnet for LLMs. i'm not sure why or how it works, but it usually leads to someone sending a PR within a matter of hours

k-yle (approved):
my previous comments were addressed. i'll leave this for someone else to look at

CAM-Gerlach:
> that's the consequence of the magic label new contributor opportunity , which is like a magnet for LLMs. i'm not sure why or how it works, but it usually leads to someone sending a PR within a matter of hours

Yeah, we've had the same issue on the repos I maintain and was just talking about it yesterday with a bunch of other maintainers when deciding how to surface good first issues for cohort of new company-sponsored developers looking to contribute to our projects, so that they can easily find good first issues to get them up to speed without the LLMs snapping them up. At least on the Spyder repo, we have the label `help wanted: easy` which while not ideal in terms of messaging (they often aren't _actually_ that easy per say) we find at least anecdotally we get a lot fewer LLM contributions versus, say, with the `good first issue` label we use on the Python repo (albeit that's a higher profile target in general; even before LLMs we've always had to deal with spam PRs there).

PR #12803, opened 2026-08-27
fetch language names from the IANA registry if missing from CLDR

k-yle:
Currently this dropdown only supports languages that have translated names from [cldr-localenames-full](https://npm.im/cldr-localenames-full):

image

unfortunately, CLDR won't accept language data for many  languages. [their criteria](https://cldr.unicode.org/index/requesting-additionsupdates-to-cldr-languagepopulation-data) states:

> It is not enough that a language be “in use”. For example, within a country such as the United States, hundreds of languages are used, sometimes by fairly sizable populations, but they are not all useful additions to the CLDR supplemental data.
>
> For CLDR purposes, the language data focus on the usefulness with computer interfaces, rather than general utility as spoken languages. Data for primarily spoken languages are usually included only where the languages have official status.

There are 835 languages used in OSM `name:*` tags which don't have translated names in CLDR, and will probably never meet that criteria. 

So, this PR adds a fallback data source: the [IANA language subtag registry](https://www.iana.org/assignments/language-subtag-registry) (via [an npm package](https://npm.im/language-subtag-registry)). But we don't want all 8000 language codes. We only include them if:
 - they are used at least once in OSM[^1] (excludes values like [`name:ant`](https://r12a.github.io/app-subtags/index?lookup=ant)); and
 - they are in the IANA registry (excludes values like [`name:qpa`](https://taginfo.osm.org/keys/name:qpa)); and
 - they are not deprecated (excludes values like [`name:xtz`](https://taginfo.osm.org/keys/name:xtz) or [`name:iw`](https://taginfo.osm.org/keys/name:iw)); and
 - they are not [a macrolanguage](https://en.wikipedia.org/wiki/ISO_639_macrolanguage) (excludes values like [`name:aae`](https://taginfo.osm.org/keys/name:aae)); and
 - they are not a "collection" (excludes values like [`name:aus`](https://taginfo.osm.org/keys/name:aus)).

Using this criteria, we get 835 new languages. Importing these names is much more sustainable than expanding our existing list of 57 hardcoded additions. (see previous PRs like #9241, #10684, #10488). Therefore, this PR removes some of those hardcoded overrides which are now redundant. 

Look at the diff of [data/languages.json](https://github.com/openstreetmap/iD/compare/kh/ts-languages...kh/ts-iana?expand=1#diff-336d7bc7b9ca631470662e689b7fc22405677507414737bcc120cb321de68dd8) to see the impact.

---

draft PR because:
- [ ] it needs more discussion about whether this is actually a good idea.
- [ ] maybe we should consider pulling translated language names from wikidata? IANA's names tend to be english descriptions rather than the endonym

[^1]: determined using the taginfo API at compile-time


data/languages.json
hlfan, line 1: It would be interesting to see if this can become an untracked file.

k-yle: yeah i think that's a good idea, this topic gets brought up quite often, also for id-tagging-schema. theoretically we don't need to version-control the `./dist` folder either

matkoniecz:
> The JSON database is licensed under the [Creative Commons Zero v1.0 Universal (CC0-1.0)](https://creativecommons.org/publicdomain/zero/1.0/legalcode) license.

note that in some cases it can still cause issues (like Wikidata with their missing decision whether to respect database rights)

but we probably can use their data without problem?

k-yle:
as far as I know, showing CC-0 text in the UI is perfectly fine.

if we're allowed to display copyrighted logos from facebook via NSI, then surely plaintext from a CC-0 source is also fine?

data/languages.json
tyrasd: +1 to dropping this file, it really does not need to be versioned. Done in 5be9d6bf5

> […] theoretically we don't need to version-control the `./dist` folder either 

`/dist` is already mostly [gitignored](https://github.com/openstreetmap/iD/blob/develop/.gitignore#L11-L18), thus it only contains the files which are either static or hard to recreate (e.g. translations which require a transifex API key). I think they are fine.

We [manually include](https://github.com/openstreetmap/iD/blob/develop/RELEASING.md#update-and-tag-release-branch) the full `dist` folder in the `release` branch (and tagged releases) which would not be strictly necessary anymore as they would be available via the published versions on npm. But keeping the full builds in the release branch has the benefit that the tar balls of those releases on github always contain a fully functioning build of iD. IMHO that's worth it.

tyrasd (approved):
> it needs more discussion about whether this is actually a good idea.

In my opinion it's a very good idea, actually: OSM does support all languages, so having better labels for as many of them in the editors is definitely an improvement.

> maybe we should consider pulling translated language names from wikidata? IANA's names tend to be english descriptions rather than the endonym

Not a bad idea to include additional language names from wikidata, but I would not wait for that before merging this PR. It's already an good improvement as is and the fetching of the endonyms (and additional non-English exonyms) from wikidata can be implemented in a later iteration.

Issue #7199, opened 2020-01-03
Endless conflict check loop following conflict resolution save failure 


mikini:
Like #5114 and #7163 I also encountered an endless loop today in conflict checking on a locally deleted entity that had been updated on the remote.

This was on the current OSM production [iD 2.17.0](https://www.openstreetmap.org/edit#map=19/56.41969/8.20379).

During the original save I was presented with the conflict resolution dialogue on a way modified both locally and remotely (both were fixes to overlap in outline of way [760017529 v1](https://www.openstreetmap.org/way/760017529/history)). In the dialogue I was not able to click "Save" after choosing to keep "theirs". The button was blue and seemed active visually but to my recollection didn't react on hover and nothing happened on mousedown.

After multiple failed attempts aborting and repeating the save with no other result I went back to editing mode removing the conflicting way locally in the hopes of then being able to save the changeset of my day-long editing session (sad but true). Saving now didn't trigger the resolution dialog but ended up in the endless "Checking conflicts" loop repeatedly counting from 0 to 1246.

In retrospect I should have saved the OsmChange representation of the changes before deleting the way and had a look at my script modifiers to see if they interfered with the conflict resolution (they usually don't interfere with iD); [uMatrix](https://addons.mozilla.org/en-US/firefox/addon/umatrix/), [uBlock Origin](https://addons.mozilla.org/en-US/firefox/addon/ublock-origin/) and [Decentral Eyes](https://addons.mozilla.org/en-US/firefox/addon/decentraleyes/).

I left the conflict check for about an hour until I started poking at it again. Looking at Firefox's javascript console revealed 3 occurrences of **graph.entities[n.id] is undefined** (see screenshot below, locale is Danish).

These seem to originate from segmentWasEdited() (from [svgLines::drawTargets](https://github.com/openstreetmap/iD/blob/v2.17.0/modules/svg/lines.js#L68)) at line 72580 in asset (also see screenshot below).

Looking at the debugger didn't give further hints (I'm not that familiar with javascript debugging so didn't dare experimenting with breaking the execution with my changeset dangling in there).

I do have a screen capture with some browsing through the Developer Tools if you reckon I might have captured anything useful.

[...]

PR #12811, opened 2026-08-31
fix unit tests calling the real panoramax api & osm api

k-yle:
I was mocking network requests while working on #7199, and noticed that our unit tests are currently spamming https://api.panoramax.xyz and https://osm.org/api/capabilities.json because the mocks are misconfigured. 

unit tests should never be calling real APIs, now it's completely blocked.

### DIFF
--- a/test/spec/services/osm.js
+++ b/test/spec/services/osm.js
@@ -7,6 +7,16 @@ import { fakeServer } from 'nise';
 describe('iD.serviceOsm', function () {
     var context, connection, spy;
     var serverXHR;
+    const capabilitiesJSON = {
+        api: {
+            waynodes: { maximum: 2000 },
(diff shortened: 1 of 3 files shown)


matkoniecz:
> unit tests are currently spamming

has it been outright logged or do you need to look at your logs?

hmmm, maybe I can test it by switching off internet connection and checking whether tests still pass?

k-yle:
you can test it by changing
```diff
- fetchMock.config.fallbackToNetwork = true;
+ fetchMock.config.fallbackToNetwork = false;
```
and then tests will fail

(but yes, you can also spy on your wifi/network traffic to see the API requests)

tyrasd (approved):
Thanks!

PR #12813, opened 2026-08-31
migrate uploader to typescript

k-yle:
prerequisite for the merge conflict PR


modules/core/uploader.ts
```diff
@@ -352,9 +388,9 @@ export function coreUploader(context) {
     }
 
 
-    function didResultInConflicts(changeset) {
+    function didResultInConflicts(this: any, changeset: osmChangeset) {
         // add a changeset tag to aid reviewers
-        changeset.tags.merge_conflict_resolved = 'manually';
+        changeset = changeset.mergeTags({ merge_conflict_resolved: 'manually' });
```
k-yle, line 393: this was a bad practice i introduced in #9636 - `osmEntity`s are supposed to be immutable. This is only enforced in unit tests, not an runtime.

now TypeScript also enforces this at compile-time, so we need to clean this up

PR #12820, opened 2026-09-02
Flag multipolygon type on nodes and ways

### LINKED ISSUE #8508: Validator: Show warning for nodes or ways tagged with 'type=multipolygon'
See, title, i assume and wiki says `type=multipolygon` is intended only for relations. Please display a warning message if nodes or ways are tagged with `type=multipolygon`.

Wiki's:
* https://wiki.openstreetmap.org/wiki/Types_of_relation
* https://wiki.openstreetmap.org/wiki/Relation:multipolygon 

Similar closed JOSM validation request: https://josm.openstreetmap.de/ticket/14288

roli-lpci:
Closes #8508.

## Summary

- warn when a node or way is tagged `type=multipolygon`
- offer a one-click fix that removes only the invalid `type` tag
- leave proper multipolygon relations and unrelated `type=*` tags unchanged

The check is intentionally narrow. The issue discussion notes legitimate non-relation values such as `type=palm`, so warning on every `type` tag would introduce false positives.

## Validation

- 18 focused `mismatched_geometry` tests pass
- targeted ESLint passes
- TypeScript checking passes
- `npm run build:data` passes
- `git diff --check` passes

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1669,6 +1669,9 @@ en:
         message: "{feature} has no descriptive tags"
       relation_type:
         message: "{feature} is a relation without a type"
+    multipolygon_type_on_non_relation:
+      message: "{feature} has type=multipolygon but is not a relation"
+      reference: "The type=multipolygon tag is only valid on relations."
     mutually_exclusive_tags:
       title: Contradictory Tags
       tip: "Find features that have contradictory tags."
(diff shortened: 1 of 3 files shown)


matkoniecz:
there is also https://github.com/openstreetmap/id-tagging-schema/issues/2516 which may enable handling it more generically, but not sure is it feasible to do it this way

roli-lpci:
Thanks — I checked that route. Today the tagging-schema deprecation matcher has no geometry constraint, and iD's deprecation matching receives tags rather than the entity geometry. A schema rule for `type=multipolygon` without #2516's geometry field and corresponding iD plumbing would also match valid multipolygon relations.

I'd keep this PR narrow for now. If #2516 and the corresponding iD support land, this hardcoded check can be retired in favor of the generic rule.

Issue #8607, opened 2021-07-30
Validator: lanes tagging by using numeric values only


sun-geo:
See title and wiki:
https://wiki.openstreetmap.org/wiki/Key:lanes

I assume this applies not only to `lanes=*` rather also to:
* `lanes:forward=*`
* `lanes:backward=*`
* `lanes:both_ways=*`
* `lanes:psv=*`
* `lanes:psv:forward=*`
* `lanes:psv:backward=*`
* `lanes:bus=*`
* `lanes:bus:forward=*`
* `lanes:bus:backward=*`
* `lanes:hov=*`
* etc...

Seems there is a corresponding JOSM check, with error message: _lanes must be a positive integer number_
https://josm.openstreetmap.de/ticket/10837

tordans:
@k-yle  just double checking that the auto close from https://github.com/ideditor/schema-builder/pull/217 was correct. https://github.com/openstreetmap/iD/pull/11181#pullrequestreview-3251358960 sounds like we might need to add a validation first to close this issue?

k-yle:
oops, yes you're right, this should not have been auto-closed. 

We'll need a followup PR in this repo to implement validation and possibly UI changes

barislabs:
I’m taking the requested follow-up validator and tests. I’ll keep it scoped to numeric `lanes*` values, and I’ll identify the automated assistance in the PR.

matkoniecz:
replaced by #12827

PR #12824, opened 2026-09-03
Validate integer field values

### LINKED ISSUE #12827: Support `integer` type field by complaining on non-integer values
see https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#type

barislabs:
## Summary

- Validate values using `integer` fields from the tagging schema instead of hard-coding lane tags.
- Check applicable preset, additional, and universal fields, including alternative keys.
- Respect field geometry, location sets, and the graph snapshot being validated.
- Accept signed integers and semicolon-separated integer lists while leaving `number` fields unchanged.
- Show only one warning when multiple fields point to the same tag.

## Testing

- `npx vitest run --no-isolate test/spec/validations/invalid_format.js`: 47 tests passed
- `npm test`: 136 test files passed and 4 skipped; 2,285 tests passed, 9 skipped, and 6 todo
- `git diff --check`: passed

The full test run still reports existing warnings in unchanged `modules/core/file_fetcher.ts` and `test/spec/core/difference.js`.

Fixes #12827

## AI assistance

I used Hermes Agent (GPT-5.6) to help implement and test this change.

### DIFF
--- a/data/core.yaml
+++ b/data/core.yaml
@@ -1636,6 +1636,9 @@ en:
         message: '{feature} has an invalid email address'
         message_multi: '{feature} has multiple invalid email addresses'
         reference: 'Email addresses must look like "user@example.com".'
+      integer:
+        message: '{feature} has a non-integer value in {key}'
+        reference: 'This value must be an integer.'
       website:
         message: '{feature} has an invalid URL in {where}'
         message_multi: '{feature} has multiple invalid URLs in {where}'
--- a/modules/validations/invalid_format.js
+++ b/modules/validations/invalid_format.js
@@ -1,4 +1,5 @@
 import { t } from '../core/localizer';
+import { locationManager } from '../core/location_manager';
 import { utilDisplayLabel } from '../util/utilDisplayLabel';
 import { validationIssue, validationIssueFix } from '../core/validation';
 import { actionChangeTags } from '../actions/change_tags';
@@ -10,7 +11,7 @@ import { presetManager } from '../presets';
 export function validationFormatting(context) {
     var type = 'invalid_format';
 
-    var validation = function(entity) {
+    var validation = function(entity, graph) {
         var issues = [];
 
         function isValidEmail(email) {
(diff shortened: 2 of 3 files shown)


matkoniecz (changes requested):
this definitely should not be lane-specific

it should be more general for integer-type fields, see https://github.com/openstreetmap/id-tagging-schema/blob/main/SCHEMA.md#type


sorry for being bitten by ancient https://github.com/openstreetmap/iD/issues/8607 - I have now opened more proper https://github.com/openstreetmap/iD/issues/12827 that relies on less of hardcoding

-------------------

for code itself: I have not reviewed what LLM generated

barislabs:
Thanks, I reworked it around the tagging schema's `integer` type and removed the lane-specific matching. It now checks applicable preset, more, and universal fields while respecting geometry and location sets. Alternative keys are covered, `number` fields stay unchanged, and overlapping fields only produce one warning per tag. The focused tests and full suite pass. I also changed the PR to fix #12827.

1ec5:
By the way, I don’t think we actually have any `integer`-typed fields in id-tagging-schema yet, either in production or in proposed pull requests. That isn’t to say that there’s anything wrong with this PR, but it might not have the intended effect. The original idea was to validate `lanes=*`, which is currently a [`number`](https://github.com/openstreetmap/id-tagging-schema/blob/cc7da72b1457c59d74e223b55388988eadd22609/data/fields/lanes.json#L3) field. But if we extend this validator rule to any `number` field, we’d quickly run into situations where non-numeric values are also valid, just not well-supported by iD yet. For example, we still haven’t split out the `measurement` type yet: openstreetmap/id-tagging-schema#1988.

matkoniecz:
Thanks for pointing it out, I started with https://github.com/openstreetmap/id-tagging-schema/pull/2819

barislabs:
I noticed this was closed without a comment. The supporting schema change in openstreetmap/id-tagging-schema#2819 has now merged, and the latest commit here contains the requested schema-driven rework. Would you be open to reopening and reviewing it, or would you prefer a different direction?

tyrasd:
only if you can explain how the 🤖 emoji managed to find its way to the PR title.

barislabs:
I added them because the contributor guide says automated agents should put `🤖🤖🤖` at the end of the title to opt into the streamlined process. I took that note literally. I’ve removed them now—sorry for the confusion.

barislabs:
> only if you can explain how the 🤖 emoji managed to find its way to the PR title.

Haha I am sorry with the unprofessional title, I am testing new methodologies for mass agent control schemes. Just do not want to give the impression of this is all handwritten as a lot of agentic help was directly used at it.  This was not a great way!

PR #12833, opened 2026-09-05
Make description of what is being defined more userproof

matkoniecz:
Try to describe in way that would be clear for past-me

### DIFF
--- a/modules/ui/preset_icon.js
+++ b/modules/ui/preset_icon.js
@@ -340,11 +340,12 @@ export function uiPresetIcon() {
       .attr('src', imageURL);
   }
 
-  // Route icons are drawn with a zigzag annotation underneath:
+  // Route icons are drawn with an additional zigzag annotation underneath main icon part:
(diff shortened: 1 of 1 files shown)


modules/ui/preset_icon.js
```diff
@@ -340,11 +340,12 @@ export function uiPresetIcon() {
       .attr('src', imageURL);
   }
 
-  // Route icons are drawn with a zigzag annotation underneath:
+  // Route icons are drawn with an additional zigzag annotation underneath main icon part:
   //     o   o
   //    / \ /
   //   o   o
   // This dataset defines the styles that are used to draw the zigzag segments.
+  // create relation with tags type=route route=bus and see its icon to see it in action
```
k-yle:
```suggestion
  // create a relation with the tags `type=route` + `route=bus` to see this icon in action
```
?

k-yle (approved):

PR #12835, opened 2026-09-05
exclude icon dependencies from cooldown

matkoniecz:
it is quite unlikely that waiting extra will help, if there is some rogue code there

it is frustrating to not be able to use new icons when it is released already, and making these update PRs manually is a waste of time

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,9 @@ updates:
       semver-major-days: 7
       semver-minor-days: 7
       semver-patch-days: 7
+      exclude:
+        - "@waysidemapping/pinhead"
+        - "@enzet/roentgen"
     groups:
       fontawesome:
         patterns:


matkoniecz:
maybe I am missing some security implications again or something is wrong with syntax here?

I am not the most confident here, at least this time I correctly made it PR rather than accidentally committing directly

tordans:
> maybe I am missing some security implications again

afaik the reasons for the cooldown are valid here as well. Any of those packages could be take over and they can all hold code other than just the icons. 

We can still do this, of course, just be aware...

matkoniecz:
@tordans as I understand, delaying here works well for popular packages because it is likely that someone else will spot problem

so malicious versions may be yanked

but for this niche ones, how likely is that delay would help, if there would be successful hijacking?

tyrasd (approved):

Issue #12164, opened 2026-04-02
Overlapping ways does not report anything


homersimpsons:
### URL

https://www.openstreetmap.org/edit?editor=id&way=232175986#map=22/47.4906970/11.0775621

### How to reproduce the issue?

1. Create 2 `highway=track` (or any other highway type) line one atop of each other
2. Notice nothing is reported

### Screenshot(s) or anything else?

Here the 2 warnings are about disconnected ways.
We can observe 4 ways (two stack of two), visible thanks to the "grey connected dots"

<img width="857" height="291" alt="Image" src="https://github.com/user-attachments/assets/eaa40ca9-4f34-41d7-b1b1-4f0b830ad095" />

~~This was originally requested in #5217 and then implemented in #5830~~ Edit: Actually those were "crossing ways" and not overlapping ones (which shares at least one segment in common).

It would greatly help for some maproulette challenge such as https://maproulette.org/challenge/49865. I did changes regarding crossing ways in #10999, I'm interested to potentially work on this I did not dig yet, any guidance would be appreciated.

crossing ways triggers correctly:

<img width="912" height="293" alt="Image" src="https://github.com/user-attachments/assets/62d2ab5c-bf84-430d-86fd-9b0a3bc0b651" />

### Which deployed environments do you see the issue in?

Released version at openstreetmap.org/edit

### What version numbers does this issue effect?

2.39.5

### Which browsers are you seeing this problem on?

Firefox

1ec5:
The validation rule implemented in #5830 is still present, but it doesn’t detect a problem if the two ways are concurrent, sharing the same nodes.

homersimpsons:
> The validation rule implemented in [#5830](https://github.com/openstreetmap/iD/pull/5830) is still present, but it doesn’t detect a problem if the two ways are concurrent, sharing the same nodes.

Yes, I just understood that, so I think the best option is to actually extend the check so overlapping ones gets treated as crossing?

Maybe that is not feasible as the UI propose to connect them while here connecting would stil keep them overlapping.

The check is maybe not so hard to implement, we need to find any 2 consecutive nodes (possibly in reverse) that are part of 2 ways. If we want to go futher we can maybe detect unconnected but with overlapping nodes (this one is probably less interesting as there already is an "unconnected ways but close node" issue reported).

1ec5:
> Maybe that is not feasible as the UI propose to connect them while here connecting would stil keep them overlapping.

[...]

PR #12857, opened 2026-09-12
:sparkles: Implement overlapping ways warning

homersimpsons:
Implements #12164

image

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -152,7 +152,7 @@ export default tseslint.config(
       'no-unused-expressions': 'off',
       'no-unused-vars': 'warn',
       'vitest/expect-expect': ['error', {
-        assertFunctionNames: ['expect', 'doMatch', 'dontMatch', 'verifySingleCrossingIssue']
+        assertFunctionNames: ['expect', 'doMatch', 'dontMatch', 'verifySingleCrossingIssue', 'verifySingleOverlapIssue']
       }],
       'vitest/no-commented-out-tests': 'off',
       'vitest/no-disabled-tests': 'off',
--- a/modules/validations/overlapping_ways.ts
+++ b/modules/validations/overlapping_ways.ts
@@ -1,0 +1,64 @@
+import { t } from '../core/localizer';
+import { utilDisplayLabel } from '../util/utilDisplayLabel';
+import { validationIssue } from '../core/validation';
+import type { CreateValidator, Validator } from '../core/validation/models';
+import type { osmWay } from '../osm';
+import type { Edge } from '../actions/add_midpoint';
+
+interface Data {
+    edges: Edge[];
+    way1: osmWay;
+    way2: osmWay;
+}
+
+export const validationOverlappingWays: CreateValidator = () => {
+    var type = 'overlapping_ways';
+
+    function considerOverlap(tags: Tags) {
+        return tags.highway !== undefined && tags.highway !== 'no';
+    }
+
+    const validation: Validator = function checkOverlappingWays(entity, graph) {
+        if (entity.type !== 'way') return [];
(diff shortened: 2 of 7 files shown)


modules/validations/overlapping_ways.ts
```diff
@@ -0,0 +1,64 @@
...
+    edges: Edge[];
+    way1: osmWay;
+    way2: osmWay;
+}
+
+export const validationOverlappingWays: CreateValidator = () => {
+    var type = 'overlapping_ways';
+
+    function considerOverlap(tags: Tags) {
+        return tags.highway !== undefined && tags.highway !== 'no';
```
homersimpsons, line 18: 1. Should this also check for waterway / railway / aeroway?
2. Should this use `osmRoutableHighwayTagValues`?
3. Should this ignore `demolished` (or similar)?

modules/validations/overlapping_ways.ts
```diff
@@ -0,0 +1,64 @@
...
+        for (let i = wayNodes.length - 2; i >= 0; i--) {
+            const currentNode = wayNodes[i];
+            const currentParentWays = graph.parentWays(currentNode).filter(
+                w => w.id !== entity.id
+                 && considerOverlap(w.tags)
+            );
+            for (const prevParentWay of prevParentWays) {
+                for (const currentParentWay of currentParentWays) {
+                    if (prevParentWay.id === currentParentWay.id && prevParentWay.areAdjacent(prevNode.id, currentNode.id)) {
+                        issues.push(new validationIssue<Data>({
```
homersimpsons, line 40: I have to provide `Data` here, I could not use `undefined` nor `null`. This seems to be useful only so it can be passed into `validationIssueFix`. In a follow-up it might be interesting to propose to "split the ways".

modules/validations/overlapping_ways.ts
```diff
@@ -0,0 +1,64 @@
...
+                            type,
+                            severity: 'warning',
+                            message: () => t.append('issues.overlapping_ways.message', {
+                                feature: utilDisplayLabel(entity, graph, false),
+                                feature2: utilDisplayLabel(prevParentWay, graph, false),
+                            }),
+                            entityIds: [entity.id, prevParentWay.id, prevNode.id, currentNode.id],
+                         }));
```
homersimpsons, line 48: Note that I didn't provide a `loc`, it looks like `iD` can derive it from the `entityIds`, but maybe I should provide one so it is in between the 2 adjacent nodes?

eslint.config.js
```diff
@@ -152,7 +152,7 @@ export default tseslint.config(
       'no-unused-expressions': 'off',
       'no-unused-vars': 'warn',
       'vitest/expect-expect': ['error', {
-        assertFunctionNames: ['expect', 'doMatch', 'dontMatch', 'verifySingleCrossingIssue']
+        assertFunctionNames: ['expect', 'doMatch', 'dontMatch', 'verifySingleCrossingIssue', 'verifySingleOverlapIssue']
```
homersimpsons, line 155: I can remove `verifySingleOverlapIssue` here and inline the `expect`s in test methods.

modules/validations/overlapping_ways.ts
homersimpsons, line 1: https://github.com/openstreetmap/iD/issues/12164#issuecomment-4186381127:
> (By the way, connected overlapping ways are definitely valid and expected in OpenHistoricalMap. That fork would need to limit this validator to only check ways that coexist in the same time period.)

Should I do anything on `iD` to handle this?

matkoniecz: I would say no, presumably OHM would comment out trigger of this validator in their fork

Issue #10772, opened 2025-02-15
editor-layer-index at npm was taken down, claimed to be containing malware, npm audit gives scary warning on install


matkoniecz:
As I understand: at some point https://www.npmjs.com/package/editor-layer-index was taken down. Maybe it actually had malware.

https://github.com/openstreetmap/iD/blob/develop/package.json#L101C28-L101C69 points not to NPM but to github

>     "editor-layer-index": "github:osmlab/editor-layer-index#gh-pages",

but npm audit does npm audit things and emits scary warning linking https://github.com/advisories/GHSA-jq9w-gjvg-mxr8

![Image](https://github.com/user-attachments/assets/f32ac81a-37e0-4907-bac9-6792ae68d496)

Is it avoidable somehow? It is definitely scary for new contributors

andrewharvey:
I don't think osmlab/editor-layer-index ever published to the npmjs.com registry. I think what happened is someone else published malware into the untaken namespace to try and catch unsuspecting developers trying to npm install from the registry instead of directly from github.com/osmlab/editor-layer-index.

Then the npmjs team took over the package.

It looks like a bug in npm audit, if we never request a package from npmjs.com within package.json, why is it reporting security issues relating to it.

EDIT: specifically this bug https://github.com/npm/cli/issues/5110

tyrasd:
Oh, this is really unfortunate. Does anyone have experience with such situations? Could it be worth trying to contact npm's support (perhaps via https://support.github.com/contact/npm-name-disputes maybe??) to get this transferred? 

Alternatively, we'd have to either live with the false positive produced by `npm audit`, or rename the package to something like `@openstreetmap/editor-layer-index` (and put a dummy/NOP version in its place on npm for it) and ask existing users to switch.

matkoniecz:
> Could it be worth trying to contact npm's support (perhaps via https://support.github.com/contact/npm-name-disputes maybe??)

I can try - if you think it is a good idea, can you post that here so I will have evidence that my request is coming from the project, not another hijacker?

tyrasd:
It's probably more effective if I try, as I'm already member of all relevant npm and github organizations/repos. Or do you prefer to take care of this @andrewharvey?

andrewharvey:
> It's probably more effective if I try, as I'm already member of all relevant npm and github organizations/repos. Or do you prefer to take care of this @andrewharvey?

Nope all yours @tyrasd, agreed probably more effective coming through you.

[...]

PR #12864, opened 2026-09-14
fix package name of editor-layer-index in staging build workflow

laurelmay:
This follows on what was done in 0d07cad and f48205a for #10772. This should
fix the issue of the staging environment not pulling in the latest imagery.


tyrasd (approved):
Thanks, good catch!

PR #12865, opened 2026-09-14
Fix flash message overflow below footer bar (#9072)

### LINKED ISSUE #9072: Follow mode flash message has overflow hidden

SaifuddinM23:
### Summary of Changes
Fixes #9072.

Adds `overflow: hidden;` to `.map-footer-bar` in `css/80_app.css`. When flash messages transition offscreen via `.footer-hide` (`bottom: -100%`), they previously leaked outside the container because `.map-footer-bar` lacked `overflow: hidden`.

### Testing Done
- Ran `npm run build:css` cleanly.
- Executed `npx vitest run --no-isolate` — all 2,269 unit tests passed.


tordans:
https://github.com/openstreetmap/iD/issues/9072#issuecomment-4197526107 describes a different solution.

SaifuddinM23:
> [#9072 (comment)](https://github.com/openstreetmap/iD/issues/9072#issuecomment-4197526107) describes a different solution.

Updated! I changed `map-footer-bar` to use `min-height: 2.5em; height: auto;` so multi-line flash messages can expand the footer height instead of clipping overflow. Thanks for pointing that out!

package-lock.json
```diff
@@ -1951,9 +1951,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
```
matkoniecz, line 1955: this appears to commit unrelated changes

matkoniecz: can you remove it from the PR? Maybe redo it in a new PR as edit will be likely easier to redo than to fix this PR?

SaifuddinM23: Fixed! I have removed the changes to `package-lock.json` so this PR now only touches `css/80_app.css`. Thanks for pointing that out!

SaifuddinM23:
Updated and cleaned up. The PR now only contains the intended css/80_app.css change. Thanks for the review!

PR #12876, opened 2026-09-15
Simplify Turkish address form

w2r4:
This PR removes postcode, district and "province" fields from the address form of iD.

And these are my reasons:

---

Postcodes in Turkey exist, but:

- They are more like an "Neighbourhood ID": They can be derived from neighbourhood information alone.
- Not actively utilized in specifying addresses, therefore *almost* no one remembers nor uses a single one.

Showing an empty postcode field in iD annoyingly tempts the user to check the postal code corresponding to the neighbourhood on [the government website](https://postakodu.ptt.gov.tr/) and copy-paste it into OSM, for every addressable feature.

In OSM, some neighbourhoods are already tagged with `postal_code=`, or some cities have postal code boundary relations.

---

Districts and provinces are completely mapped as boundaries in Turkey.

`addr:*` keys on features could also be used in theory but it is a total mess right now:
- [OSM Wiki page](https://wiki.openstreetmap.org/wiki/Addresses#Turkey) recommends `addr:subdistrict` **for districts** and `addr:district` **for provinces**.
- Despite that, features in a few provinces in Turkey are tagged with `addr:district` for districts and `addr:province` for provinces.
- Despite that, majority of features follow the iD format: They are tagged with `addr:district` for districts and `addr:city` **for provinces**.

I think it would be a good start to reduce the rate of increase of these tags for improving OSM address quality in Turkey.

---

Also, turns out many new mappers are confused by the address form, resulting in inaccurate tagging. (e.g. [1](https://www.openstreetmap.org/way/1536501143), [2](https://www.openstreetmap.org/way/530620375), [3](https://www.openstreetmap.org/way/1535266886), [4](https://www.openstreetmap.org/node/13919957907)).

[A maproulette challenge for Istanbul](https://maproulette.org/browse/challenges/53570) had been also made to fix such values in `addr:city`.


matkoniecz:
Has it been consulted with Turkish community? Thread at community forums or something similar?

matkoniecz:
@w2r4 Has it been consulted with Turkish community? Thread at community forums or something similar?

w2r4:
Well, actually no.

matkoniecz:
Are you planning to do so? Not saying that it is a bad idea but there are typically clear preferences among community which address fields should be filled and which should be left implied.

w2r4:
> Are you planning to do so?

I'm sorry but I'm not feeling like doing that. If such discussion is really important, we can save time by not applying this patch.

matkoniecz:
closing as it is unclear whether community supports it

> [A maproulette challenge for Istanbul](https://maproulette.org/browse/challenges/53570) had been also made to fix such values in addr:city.

running "Standardize Istanbul Addresses" without checking with community is suspect at best

Issue #8765, opened 2021-10-24
Possibly undesirable instance of warning "Foot Path is disconnected" when connected to piers


harahu:
I am getting a "Foot Path is disconnected" warning on this way: https://www.openstreetmap.org/way/930569104

I am not sure I _should_ be getting this warning.

While the warning _is_ technically true, the path connects two floating piers that are connected to the road network. These piers can be considered "paths", and are traversable by foot, an I haven't seen precedence for also tagging floating piers as `highway=...` in addition to `man_made=pier` I haven't checked, but piers might or might not be routable depending on engine.

Should a connection through a pier be considered a valid connection by iD, or is there an alternative I can use to mend this warning?

1ec5:
This seems to be a false positive but a bit of an edge case. It would be helpful to know how many footways only connect piers like this.

> I haven't checked, but piers might or might not be routable depending on engine.

There seems to be support in some routing engines but not all:

* GraphHopper: graphhopper/graphhopper#1120
* OpenTripPlanner: opentripplanner/OpenTripPlanner#921
* OsmAnd: osmandapp/OsmAnd#9416
* [OSRM](https://github.com/Project-OSRM/osrm-backend/blob/41dda32546399f1dc12af1de41668993de44c7dc/features/foot/way.feature#L30)
* Valhalla: ?

harahu:
> There seems to be support in some routing engines but not all: [...]

Thank you for the detailed research, @1ec5!

The partial support for routing on piers is interesting. Could it be implying a weakness in the tagging system? At least, if one intend piers to be routable and considered a way, it makes sense for it to be tagged with `highway=...`, similar to how bridges are usually tied to a way of some sort. The benefit would be that it would play nicely with routing engines as they currently are implemented.

Does there exist anything like a pier that is _not_ intended for human traversal by either foot or land-based vehicle? I struggle to imagine such a case, but if it exists, one might need a mechanism for differentiating between routable and un-routable piers.

Not sure how easy it is to automate a search for instances like this. Might be better to start with piers connected to highways rather than the other way around.

1ec5:
> Does there exist anything like a pier that is _not_ intended for human traversal by either foot or land-based vehicle? I struggle to imagine such a case, but if it exists, one might need a mechanism for differentiating between routable and un-routable piers.

[...]

Issue #7757, opened 2020-06-29
Consider piers for routing disconnected ways


rivermont:
Currently if a `route=ferry` is connected to a larger network only by `man_made=pier`'s, iD throws the `Disconnected from other roads and paths` error.  
Piers are routable according to [Osmose](https://github.com/osm-fr/osmose-backend/blob/703f6bac7d56b623c193363fb22398f5f24fc93b/analysers/analyser_osmosis_highway_floating_islands.py#L42) and [the wiki](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dpier#How_to_map) (kinda).

quincylvania:
The existing behavior is intentional. Not all routers will necessarily treat piers as routable and people use the `man_made=pier` tag for all sorts of things you can't really walk on, like bridge pilings.

You can manually add `highway=footway` or `highway=service` to routable piers.

PR #12880, opened 2026-09-16
Reduce false positives in routing islands validator

tyrasd:
* small islets without ferry, but with a pier (e.g. private island at https://osm.org/go/0MZ4tNAk--?m=)
* remote locations with an airstrip or helipad (e.g. https://osm.org/go/0NV2WHVQ-?m= )
* remote villages accessible by boat without a pier (e.g. https://osm.org/go/YMWpUZw?m=)

considered tags are:

* `man_made=pier/quay` (way only)
* `waterway=access_point` (vertex only)
* `aeroway=airstrip/helipad` (vertex or area/way)

not considered are the more "major" variants of aeroways (i.e. `aerodrome`/`heliport`), as those would unlikely be found in the small locations this PR is trying to cover.

//edit: this should help reduce the [mistagging](https://www.openstreetmap.org/node/1037593184/history/2) of piers with fake `entrance` tags in order to pacify iD's validator.


modules/validations/disconnected_way.ts
```diff
@@ -162,15 +166,34 @@ export const validationDisconnectedWay: CreateValidator = () => {
...
 
+        function isConnectedWay(entity: osmWay) {
+            if (entity.tags.man_made === 'pier' ||
+                entity.tags.man_made === 'quay') {
+                // consider piers and quays as sources of traffic
+                return true;
+            }
+            if (entity.tags.aeroway === 'airstrip' ||
+                entity.tags.aeroway === 'helipad' ||
+                entity.tags.aeroway === 'runway') {
```
k-yle, line 177: for new code, should we use a [`TagDictionary`](https://github.com/openstreetmap/iD/blob/f4cad2b7b84097330a9d00ac9263bd444951af29/modules/util/object.ts#L10-L29) ? 

iD already has heaps of tags hardcoded in mysterious places, if they were all in an object then we could eventually generate a [taginfo-projects](https://github.com/taginfo/taginfo-projects) file, to document the assumptions iD makes about each tag (see 857a34f8ab003634d35f29449b33a3e6084a3e00 for an old unfinished implementation)

modules/validations/disconnected_way.ts
```diff
@@ -162,15 +166,34 @@ export const validationDisconnectedWay: CreateValidator = () => {
...
-            // entrances are considered connected - #3906
+            // entrances - #3906
             if (vertex.tags.entrance &&
                 vertex.tags.entrance !== 'no') return true;
             if (vertex.tags.amenity === 'parking_entrance') return true;
+            // aeroway features
+            if (vertex.tags.aeroway === 'airstrip' ||
+                vertex.tags.aeroway === 'helipad') return true;
+            // waterway features
+            if (vertex.tags.waterway === 'access_point') return true;
```
k-yle, line 196: thanks for including `waterway=access_point`! it would be nice to also include `leisure=slipway` (as  a node or way). 

where I live, most waterfront properties have a private boat ramp which is disconnected from the road network (it just leads into a grassed backyard)

Image

tyrasd:
I just noticed that this now somewhat contradicts/counteracts the previous implementation in #7757 / https://github.com/openstreetmap/iD/issues/8765#issuecomment-4658073509 (i.e. explicitly treating `man_made=pier` as non-routable), because now the validator cannot distinguish between a simple pier to a tiny island and a pier which is part of a ferry port. Should we limit this to piers that also have `highway=footway/path/service` on them?

modules/validations/disconnected_way.ts
tyrasd: Yeah, totally makes sense. I think I did look at the tag, but dismissed it too quick as something one would only find on a port or larger island.

PR #12881, opened 2026-09-16
Prevent clobbering typed text on delayed click (#12877)

### LINKED ISSUE #12877: Tag value field suggestions don't recognize text already typed
### URL

_No response_

1. Go to the Tags section of iD sidebar in List mode
2. Type a common tag key value in the key field (e.g. "natural")
3. _Quickly_ type a few characters for the value for the tag in the text field to the right
4. The value selection drop-down picker has a slight delay before it appears.

The mismatch in value selections occurs when...
- characters are quickly typed immediately upon clicking into the text field during the brief delay period
- characters are already in the value text field and you click to reenter the field

Note: While it may be useful for drop-down pickers for preset fields to disregard text already in the text field (e.g. to quickly change a surface tag from paved to asphalt), the raw tag editor would be more intuitive and versatile if it always recommends values based on what is already typed, only showing top values when the field is empty.

Value field with "shr" typed before drop-down appears; thus, rather than recommending the top TagInfo value "shrubbery", the drop-down behaves as if the text field is empty
Image

Current behavior when typing "s" into value field (with other values starting with "s" recommended)

Image

Current behavior when clicking into a value field where "s" has already been typed

Image

Image

2.42.2

_No response_

SaifuddinM23:
## Description

Fixes #12877.

- Prevent delayed combobox mouseup handling from overwriting text typed while the dropdown is opening.
- Add `fetchValueOnOpen()` to allow comboboxes to fetch suggestions based on existing input.
- Enable this behavior for tag keys and values in the raw tag editor.
- Preserve the existing behavior for preset comboboxes.

## Testing

- `npm run test:once` — 2272 tests passed
- `npm run lint` — passed
- Manually tested typed and existing tag values in the raw tag editor.


matkoniecz:
please, mention LLM use as the first sentence in PR description in cases where LLM generated code

SaifuddinM23:
> please, mention LLM use as the first sentence in PR description in cases where LLM generated code

I wasn't aware of this earlier when I opened the PR. I've fixed it now

Issue #3044, opened 2016-03-24
Allow reserved words to be used as keys


robybully:
If a user uses the table in the lower left to edit tags and inserts "constructor" as key iD changes the key to "constructor_1". If it is changed to "constructor" again, iD changes it to "constructor_2". If it is again changed iD changes it again to "constructor_1" and so on...

Produced errors can be seen here: http://taginfo.openstreetmap.org//search?q=constructor

My browser is Firefox 45.0 with deactivated Adblock Plus on Windows 7.

![openstreetmap-000178](https://cloud.githubusercontent.com/assets/17740158/14024430/4dae0398-f1e9-11e5-8d5a-7e191f3e29a3.png)

bhousel:
Discussed tangentially in #2896 

The raw tag control is data bound to a JavaScript object.  Other special keywords will do the same thing,  e.g. you also can't enter tags like `__proto__`, `__defineGetter__`, `hasOwnProperty`, `toString`.

We should probably change this to use a [d3.map](https://github.com/mbostock/d3/wiki/Arrays#maps).

kepta:
@bhousel 
Don't you think refactoring the code to use d3.map, would be an enormous task?

I see tags everywhere in the code, and they all are using the key/value pair directly.

bhousel:
Yes, unfortunately this is an ugly issue, and it's something that needs an audit everywhere in the code that works with OSM tags.  I guess we have 2 choices:
1. Replace tag objects with `d3.map` - or better yet, I think all the browsers we target support [Native ES6 style Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) - then replace everywhere we use dot notation with `get`/`set`/`has`/etc.
... or ...
2. Maybe use the [`Object.create(null)` trick](http://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull) to create empty objects with no properties to conflict with OSM keys.  But if we do this, we'd have to do such a significant code audit that we might as well just replace them with modern Maps.

@jfirebaugh any thoughts on what direction @kepta should take?

jfirebaugh:
Is there an advantage to using native `Map` over `Object.create(null)`? I think we should do one or the other.

bhousel:
[...]

PR #12887, opened 2026-09-16
stop using `tags.hasOwnProperty(...)`

k-yle:
following a discussion about #3044 last week - fixing the issue for [`constuctor`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor) is way too hard, but we can fix it for [`hasOwnProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty), by enabling the eslint rule and then fixing all violations.

After this PR, creating a line tagged with `hasOwnProperty=*` no longer causes a crash.


matkoniecz:
is it intentional that `hasOwnProperty=` still cannot be set?

matkoniecz (approved):

k-yle:
yes it's likely that certain parts of the editor still won't work properly for the same reason as #3044. this PR is just trying to fix the crash

tyrasd (commented):
as we use this method quite a lot and `Object.prototype.hasOwnProperty.call(…)` is pretty verbose: what if we create a helper function (e.g. `utilHasOwnProperty(obj, prop)`) for this?

tyrasd:
PS:

> fixing the issue for constuctor is way too hard

I haven't been able to reproduce the crash with a `constructor` tag? Is it still triggerable somehow?

PR #12890, opened 2026-09-17
Refresh edit menu disabled states after map redraws

### LINKED ISSUE #12754: right-click menu icon state are not updated in obscure cases
to reproduce:

- select road attached to footway
- right-click it
- open map data sidebar
- disable paths
- note that icons are not grayed out but should be (actions are still blocked)

the same error happens when you start with some action disabled and say paths got enabled making actions available

Shubham-Padkonde:
Fixes #12754.

When map feature visibility changes while the edit menu is open, its buttons retain their original disabled styling even though clicking them checks the current operation state. Refresh the existing disabled-state calculation on full map redraws, alongside menu positioning. Operations with an interrupt handler remain visually enabled.

Four regression tests cover enabling, disabling, interruptible actions, and listener cleanup. The enabling/disabling assertions fail on the original source and pass with this change.

Validation (Windows, Node 24.15):
- All four focused tests pass.
- Full suite: 2,273 assertions pass, 9 skipped, 6 todo. The run exits unsuccessfully because of two unhandled fetch-mock errors for an OSM map request. An unchanged-source run, excluding the new test file, reproduces the same two errors with 2,269 assertions passing.
- Full ESLint, TypeScript checks, and CSS/data/JavaScript build pass.
- The local application loads in the browser. The reported layer-toggle interaction is covered at the menu component's redraw boundary; no live OSM edits were made.

Prepared with OpenAI Codex assistance.


matkoniecz:
> Prepared with OpenAI Codex assistance.

can you edit description to put this warning on top?

PR #12895, opened 2026-09-18
Prevent connecting non-adjacent nodes in closed ways

### LINKED ISSUE #9328: Can create invalid geometries with node merge.

SaifuddinM23:
Hey , My solution disables the connect/merge action when attempting to connect two non-adjacent nodes of the same closed way (polygon), which previously resulted in invalid self-intersecting geometries.

 Fixes #9328

https://github.com/user-attachments/assets/c4d90d22-206e-4658-9aca-0bb03a72df06

As you can see in the above video if i try to merge the nodes a pop-up appears at the bottom which tells "These features cant be merged because the resulting path would intersect itself"


SaifuddinM23:
@matkoniecz Can you please review and take a look on my solution?

matkoniecz:
I will try, sadly number of waiting PRs is quite significant (127 waiting for review, after excluding ones waiting for something else etc). 

BTW, is cog icon in top right in review panel, allowing to ask for review from specific people is accessible to you?

If so, feel free to ask for feedback from me there (rather than ping)

screen-2026-09-19-00-00-52

matkoniecz:
Is it intentional that it is happening for linear features forming closed way and not only for areas?

As far as I know it is entirely valid for footway to form such shapes and for example when mapping park it may be entirely reasonable mapping

SaifuddinM23:
> I will try, sadly number of waiting PRs is quite significant (127 waiting for review, after excluding ones waiting for something else etc).
> 
> BTW, is cog icon in top right in review panel, allowing to ask for review from specific people is accessible to you?
> 
> If so, feel free to ask for feedback from me there (rather than ping)
> 
> screen-2026-09-19-00-00-52

I was not aware that I could request you for feedback from review panel but currently I dont see any icon or a way to ask for a review

SaifuddinM23:
> Is it intentional that it is happening for linear features forming closed way and not only for areas?
> 
> As far as I know it is entirely valid for footway to form such shapes and for example when mapping park it may be entirely reasonable mapping

I was not able to think about the edge case , I'll surely take a look how iD handles closed linear features

matkoniecz:
> I was not aware that I could request you for feedback from review panel but currently I dont see any icon or a way to ask for a review

may be restricted to maintainers and triagers... Sorry for noise then

either way: I am actively looking for PRs with missing reviews, just that we have quite a lot of them.

PR #12899, opened 2026-09-18
Size the raw tag editor's key column to fit its keys

### LINKED ISSUE #12859: Tags table should reduce width of the key column to avoid overflowing values
### Description

The tags table appears to split the key/value columns by 50:50 and is not able to be dragged by the user.

In some cases the values overflow, while the keys have ample whitespace.

We should consider that if the values overflow, to reduce the width of the key column (without causing it to overflow) to provide more space for values.

### Screenshots

Image

jonesfionn101-dotcom:
> **AI use:** the code change, the test, the CHANGELOG line and this description were written with Claude Code (an AI assistant). I'm responsible for the change and will answer review comments.

Fixes #12859

The key and value columns in the tag list were always split 50:50, so long values got cut off even when every key was short.

Each key input now gets a `size` equal to the longest key in the list. The key column starts from 50% but never grows past what the keys need (minimum 25%), and the value column takes the rest. So when all keys are short the values get more room, and when any key is long the layout is the same as today.

Lint and tests pass. There is a test in `test/spec/ui/sections/raw_tag_editor.js` for the `size` attribute. Before and after, rendered with the built iD.css:

before and after


tordans (commented):
I think those Claude commends are too long for what they explain. And they should not talk about the old code, but explain the new code of needed.

matkoniecz:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

can you put those on top? and specify what was generated (also PR description?)

hlfan:
> I think those Claude commends are too long for what they explain

...and also not accounting for this being git-blame-able.

jonesfionn101-dotcom:
Thanks for the reviews! I've shortened the comments so they only describe the new code, and moved the AI note to the top of the description, saying what it wrote.

PR #12900, opened 2026-09-19
Add railway halt area support

Geo-2695:
When users try to represent a specific railway halt using an area, support anomalies occur
halt1
halt2

This update adds support for areas


Geo-2695:
### Test
hal

tordans:
Please link/share some reference from the wiki and/or actual usage (eg with PostPass) to support the change.

matkoniecz:
Also, if it makes sense as area iD tagging schema also should be changed.

matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:railway%3Dhalt is not mentioning validity on areas

there was and still is https://community.openstreetmap.org/t/railway-station-as-an-area/104839

Geo-2695:
I understand...

PR #12906, opened 2026-09-21
Make handling of fields containing `:both` more consistent

k-yle:
follow up to https://github.com/openstreetmap/id-tagging-schema/pull/2903#discussion_r4034844618

the same bug also exists in iD. It also affects the code that decides when to render items from `moreFields`.

To test:
- create a line with `highway=residential` + `parking:both:orientation=parallel`, the field should appear
- create a line with `highway=residential` + `parking:orientation=parallel`, the field should appear


matkoniecz (approved):

tyrasd (changes requested):
Hmm… not sure this the proper fix for this "bug":

As far as I understand, the tag [`parking:orientation`](https://taginfo.openstreetmap.org/keys/parking:orientation) was previously used on `amenity=parking + parking=street_side` features, but was deprecated with the [new parking proposal](https://wiki.openstreetmap.org/wiki/Proposal:Street_parking_revision) a few years ago. It was never intended to be used on roads, where only `parking:*:orientation` is to be used.

So, for fields like this one, where the `key` already contains `:both`, we probably should not allow the version without `:both` to be allowed. If I remember correctly, that's also how this field was initially modeled and implemented, but we overlooked this possibility in when changing the logic for the inverse case in #9587 it seems. 🙈

modules/presets/field.ts
```diff
@@ -65,13 +65,14 @@ export function presetField(fieldID: string, field: Field) {
...
         // directionalCombo fields can have an additional key describing the for
         // cases where both directions share a "common" value.
         // The field also support *:both. The preset decides which field to write to.
-        const baseKey = field.key!.replace(/:both$/, '');
-        allKeys.add(baseKey);
-        allKeys.add(`${baseKey}:both`);
+        const otherKey = field.key.includes(':both')
+            ? field.key.replace(/:both(:|$)/, '$1')
+            : `${field.key}:both`;
+        allKeys.add(otherKey);
```
tyrasd, line 75:
```suggestion
        if (!field.key.includes(':both')) {
            allKeys.add(`${field.key}:both`);
        }
```

does this make sense?

k-yle:
yeah, the fact that we support `parking:orientation` in this field is a side-effect of #9587. I think that's fine, because this tagging combination should be non-existant (or at least increasingly rare), and we show the error:...

image

...which would cause the field to disappear if the user accepts the tag upgrade.

---

i saw this as a bug "fix" because this code is duplicated in 3 places, and this one is different to the other two:

https://github.com/openstreetmap/iD/blob/b08827e5e87a977587f49239fb734a126ac9ea69/modules/ui/fields/directional_combo.js#L135-L137

and 

https://github.com/openstreetmap/iD/blob/b08827e5e87a977587f49239fb734a126ac9ea69/modules/ui/fields/directional_combo.js#L91-L93

So this change just aligns field.ts with the reality of how the tag is used in directional_combo.js

modules/presets/field.ts
k-yle: this wouldn't work for `sidewalk:both`, where we need to add `sidewalk`

tyrasd: true, forget this

tyrasd (approved):
> ...which would cause the field to disappear if the user accepts the tag upgrade.

Well, ideally we don't show this deprecation message on roads: https://github.com/openstreetmap/id-tagging-schema/pull/2932

So, in fact if everything is properly tagged this "fix" doesn't change anything and just makes the handling of an "unsupported" tag a bit more consistent in the different parts of the code. That I can live with. :)

tyrasd:
PS: I removed the changelog entry, as this is really an edge case, where the description in the changelog would probably add more confusion than benefit, I think. Hope that's ok.

tordans (approved):
LGTM
In the new parking schema tagging the ":side" is required (I fought for that ;-)).
+1 to align code
+1 to cleanup the deprecation as suggested
+1 https://github.com/openstreetmap/iD/pull/12906#pullrequestreview-5276895210

